PR #5514, opened 2019-07-04
CB-1914 Create unit tests for Credential related services/validators

afarsang:
Create unit tests for Credential related services/validators

### DIFF
--- a/environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialServiceTest.java
+++ b/environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialServiceTest.java
@@ -1,0 +1,220 @@
+package com.sequenceiq.environment.credential.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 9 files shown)


environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialServiceTest.java
```diff
@@ -0,0 +1,224 @@
...
+    private SecretService secretService;
+
+    @MockBean
+    private NotificationSender notificationSender;
+
+    @MockBean
+    private CloudbreakMessagesService cloudbreakMessagesService;
+
+    static {
+        CREDENTIAL = new Credential();
```
topolyai5: What is the reason for static init block?

afarsang: no real reason, I created test data as static field and some of them needs initialization

afarsang: no, it is a mistake, I'll fix it

PR #5515, opened 2019-07-04
CB-2186. Use cloud storage objects in logging requests.

oleewere:
- consolidate telemetry related request/response models, put those as well into cloud-api, so won't be needed to duplicate on different places (core, core/distrox, environment), also uses 1 telemetry converter instead of doing similar conversions at the same places
- move unit tests as well
- remove most of the test properties from request (like access / private key from WA request ... from that point it will work in a similar way, as cm license ..with mock auth, you can use a credential file that can be read, and access/secret keys will be filled based on that). also from this point use simple map objects for providing extra fluentd or WA (CM service component) configs
- added storageLocation string input to logging request. i saw that for cloud storage requests, there was a property based storage location request, but with fluentd use-case, there is no any property so a string is enough
- remove enabled field
- set databusEndpoint in WA object if that is set globally

FT: cluster install delete etc. validated

next step: pass telemetry to CloudStack from stack, then the instance profile can be set though that


oleewere:
retest this please

oleewere:
overall what changed here: 
- using one Telemetry/Logging/WA request (re-used from cloud-api) instead of multiple ones in different modules

1. on "logging" side cloud storage specific params are re-used instead of using objects like S3Attribures (+ add a storage location string to the request);

so it looked like this in the template(s);
```json
"telemetry" : {
    "logging" : {
     "enabled": true,
    "attributes": {
        "s3" : {
          "bucket" : "oszabo"
       }
    }
    }
}
```
and now it looks like:
```json
"telemetry" : {
    "logging" : {
        "storageLocation" : "oszabo", // or s3://oszabo/vmi etc.
        "enabled": true,
        "s3" : {
          "instanceProfile" : "<blah-blah>"
       }
    }
}
```
of course make this work require to provide instance profile in the cluster stack request (as the instance profile from the s3 params are not used yet, but next step should be after that is to pass telemetry object to CloudStack object - what is used to fill aws params for cloud formation template)

2. on "WA" (workload analytics) side, main change is remove provided test parameters (access/secret key), and use similar idea as with CM license, so with mock-auth, you can provide a credential file through `altus.databus.credential.file` property. (it's needed as with mock-auth, you cannot use ums to generate an access/secret keypair) -if you would like to create your own credential file, go to iam for manowar/thunderhead dev console, create a machine user, assign a dbus role and generate an API key (with using old rsa format for secret)...in the end if you download that file you can use that for `altus.databus.credential.file`. (note: sdxId and sdxName params are still used for testing as until no crn on stack/cluster level, we cannot get the sdx id from crn)

in template it could look like this: (for testing with mock auth use the credential file, not needed for manowar dev)
```json
"telemetry" : {
    "workloadAnalytics" : {
        "enabled": true,
        "attributres" :
        {
          "sdxId" : "<id from sdx crn, second uuid in the string>",
          "sdxName": <sdx-name>
       }
    }
}
```
note if sdx name/id not used, these will be randomly generated

oleewere:
@attilapalfi92 , consider the description above, if you would like to cherry-pick this change to the other branch or a new one (as i won't be available on next week)

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

PR #5517, opened 2019-07-04
CB-2200 FreeIPA should open it's ports to the whole VPC CIDR range

lacikaaa:


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Security.java
```diff
@@ -13,9 +13,18 @@
 
     private final List<String> cloudSecurityIds;
 
+    private final boolean useNetworkCidrAsSourceForDefaultRules;
+
     public Security(@Nonnull Collection<SecurityRule> rules, @Nonnull Collection<String> cloudSecurityIds) {
```
lajosrodek: Not a big deal but why not delegate to the other constructor below as `this(rules, cloudSecurityIds, false)`, so that the logic need not be duplicated?

PR #5533, opened 2019-07-05
CDPAM-116 | Persisting public key cert to be used to configure knox t…

dvguruprasad:


core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/orchestration/ClusterProxyRegistrationHandler.java
```diff
@@ -39,18 +52,31 @@ public String selector() {
...
+                return new ClusterProxyRegistrationSuccess(request.getResourceId());
             }
+            ConfigRegistrationResponse registerResponse = clusterProxyService.registerCluster(stack);
+            Cluster cluster = stack.getCluster();
+            if (cluster.hasGateway()) {
+                LOGGER.debug("Updating Gateway for cluster {} in environment {} with public key certificate retrieved from Cluster Proxy",
+                        cluster.getId(), stack.getEnvironmentCrn());
+                Gateway gateway = cluster.getGateway();
+                gateway.setTokenCert(registerResponse.getX509Unwrapped());
+                gateway.setSsoType(SSOType.SSO_PROVIDER);
```
keyki: Do we need to update the sso type? I think this comes from the API requests.

dvguruprasad: When debugging locally we didn't see it getting populated - gets persisted as 'NONE'. Was this changed recently?

keyki: NONE is the default value if you don't provide it on the API. 
```
    "gateway": {
      "topologies": [
        {
          "topologyName": "dp-proxy",
          "exposedServices": [
            "ALL"
          ]
        }
      ],
      "ssoType": "SSO_PROVIDER"
    }
```
SDX and DistroX already sets them to `SSO_PROVIDER`

dvguruprasad: Ok. Looks like this is not getting set from the UI.

PR #5536, opened 2019-07-08
CB-2061 yarn and azure support for sdx

sodre90:


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/EnvironmentNetworkYarnParams.java
```diff
@@ -0,0 +1,47 @@
...
+import com.sequenceiq.environment.api.doc.environment.EnvironmentModelDescription;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(value = "EnvironmentNetworkYarnV1Params")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
+public class EnvironmentNetworkYarnParams {
+    @ApiModelProperty(value = EnvironmentModelDescription.AWS_VPC_ID, required = true)
```
topolyai5: If required, you should add `@NonNull` as well

environment/src/main/resources/schema/app/20190702141237_add_yarn_queue.sql
```diff
@@ -0,0 +1,9 @@
+-- // DISTX-197 Implement Logging/WXM request support on Env side
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE environment_network ADD COLUMN IF NOT EXISTS queue text;
```
topolyai5: Why do you use `text`, is `varchar2` not enough?

PR #5543, opened 2019-07-08
CB-1988 introduction of cluster shapes in SDX

lnardai:
[need to test] 
light-duty OK
medium-duty needs testing
[needs cli] 
internal API needs to send custom cluster shape

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterResponse.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterResponse.java
@@ -6,6 +6,8 @@ public class SdxClusterResponse {
 
     private String name;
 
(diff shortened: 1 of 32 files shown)


datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterResponse.java
```diff
@@ -6,6 +6,8 @@
 
     private String name;
 
+    private SdxClusterShape shape;
```
lajosrodek: Any particular reason why the field is named differently than in the request?

lnardai: Sure, it should be cluster shape.
I also need to define the correct bp-cluster-template relations, so this is really a draft as of now.

PR #5545, opened 2019-07-08
CB-1863 api name collision fix.

lnardai:
isn't there swagger def generation during PR time?


lnardai:
It's already merged in 5544

PR #5549, opened 2019-07-08
CDPSDX-595: Replacing the sensitive strings on the response to Secret…

kkalvagadda:
The intent of this change is to store sensitive information like principal and keytab in the vault and send the SecureResponses for them with the path and secret.

Testing performed
1. Invoked the API and made sure that vault is updated with the corresponding entries and the path and secret match the value that is returned by the API. I have used vault web UI for it.
2. Added test code to retrieve the value from the vault using the path and secret and made sure that they match with the values that are inserted into the vault. This test code is not committed.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
@@ -1,6 +1,8 @@
 package com.sequenceiq.freeipa.kerberosmgmt.v1;
 
 import java.util.Optional;
+import java.util.UUID;
+
 import javax.inject.Inject;
 
 import org.slf4j.Logger;
@@ -9,6 +11,9 @@ import org.springframework.stereotype.Service;
 
 import com.google.common.base.Strings;
 import com.googlecode.jsonrpc4j.JsonRpcClientException;
+import com.sequenceiq.cloudbreak.service.secret.model.SecretResponse;
+import com.sequenceiq.cloudbreak.service.secret.model.StringToSecretResponseConverter;
+import com.sequenceiq.cloudbreak.service.secret.service.SecretService;
 import com.sequenceiq.freeipa.api.v1.kerberosmgmt.model.HostRequest;
 import com.sequenceiq.freeipa.api.v1.kerberosmgmt.model.ServiceKeytabRequest;
 import com.sequenceiq.freeipa.api.v1.kerberosmgmt.model.ServiceKeytabResponse;
@@ -39,6 +44,8 @@ public class KerberosMgmtV1Service {
 
     private static final String KEYTAB_FETCH_FAILED = "Failed to fetch keytab.";
 
+    private static final String VAULT_UPDATE_FAILED = "Failed to update Vault.";
+
     private static final String EMPTY_REALM = "Failed to create service as realm was empty.";
 
     private static final String IPA_STACK_NOT_FOUND = "Stack for IPA server not found.";
@@ -54,6 +61,12 @@ public class KerberosMgmtV1Service {
     @Inject
     private FreeIpaClientFactory freeIpaClientFactory;
 
+    @Inject
+    private SecretService secretService;
+
+    @Inject
+    private StringToSecretResponseConverter stringToSecretResponseConverter;
+
     public ServiceKeytabResponse generateServiceKeytab(ServiceKeytabRequest request, String accountId) throws FreeIpaClientException {
         LOGGER.debug("Request to generate keytab for Service:{} Host:{} in Environment:{}", request.getServiceName(), request.getServerHostName(),
                 request.getEnvironmentCrn());
@@ -66,8 +79,8 @@ public class KerberosMgmtV1Service {
         hostAdd(request.getServerHostName(), ipaClient);
         String servicePrincipal = serviceAdd(request.getServiceName(), request.getServerHostName(), realm, adminUser, ipaClient);
         String serviceKeytab = getServiceKeytab(servicePrincipal, ipaClient);
-        response.setKeytab(serviceKeytab);
-        response.setServicePrincial(servicePrincipal);
+        response.setKeytab(getSecretResponseForKeytab(accountId, serviceKeytab));
+        response.setServicePrincial(getSecretResponseForPrincipal(accountId, servicePrincipal));
         return response;
     }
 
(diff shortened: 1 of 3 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -181,4 +192,30 @@ private boolean isDuplicateEntryException(FreeIpaClientException e) {
                 .filter(c -> c == DUPLICATE_ENTRY_ERROR_CODE)
                 .isPresent();
     }
+
+    SecretResponse getSecretResponseForPrincipal(String accountId, String principal) {
+        try {
+            String path = String.format("%s/%s/%s/%s-%s", accountId,
+                    "ServiceKeytab", "serviceprincipal",
+                    UUID.randomUUID().toString(), Long.toHexString(System.currentTimeMillis()));
```
jamisonbennett: Why is a timestamp needed? Doesn't the UUID already handle the time for uniqueness (ie if the UUID is a duplicate then the timestamp will also be the same) or is there timestamp useful for something else?

(Same for `getSecretResponseForKeytab()`)

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -181,4 +192,30 @@ private boolean isDuplicateEntryException(FreeIpaClientException e) {
                 .filter(c -> c == DUPLICATE_ENTRY_ERROR_CODE)
                 .isPresent();
     }
+
+    SecretResponse getSecretResponseForPrincipal(String accountId, String principal) {
+        try {
+            String path = String.format("%s/%s/%s/%s-%s", accountId,
+                    "ServiceKeytab", "serviceprincipal",
```
jamisonbennett: Shouldn't `"serviceprincipal"` be `principal`? That way the clean up can remove it by the principal name. Or do we need to have the path contain something which is specific to the env/cluster (I think thinking of clean up in cases where this succeeds but the client gets an error back and so the client doesn't know that the cleanup methods need to be called but they do know that the cluster needs to be deleted or something along those lines). We can probably handle this better with CDPSDX-515.

(Same for `getSecretResponseForKeytab()`)

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -181,4 +192,30 @@ private boolean isDuplicateEntryException(FreeIpaClientException e) {
...
+                    UUID.randomUUID().toString(), Long.toHexString(System.currentTimeMillis()));
+            String secret = secretService.put(path, principal);
+            return stringToSecretResponseConverter.convert(secret);
+        } catch (Exception exception) {
+            LOGGER.warn("Something went wrong with Secret store. ServiceKeytabResponse is not updated ", exception);
+        }
+        return null;
```
jamisonbennett: The other methods have been modified to throw `KeytabCreationException` rather than returning `null`. It seems like returning `null` might ignore the swallowed exceptions.

(Same for `getSecretResponseForKeytab()`)

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
kkalvagadda: This is the convention followed in other places where the vault is used. I followed it.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
kkalvagadda: Once we have more understanding of the cases where cleanup has to be performed we can make changes to the way is constructed based on what we find.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -181,4 +195,35 @@ private boolean isDuplicateEntryException(FreeIpaClientException e) {
                 .filter(c -> c == DUPLICATE_ENTRY_ERROR_CODE)
                 .isPresent();
     }
+
+    private String getDecodedKeytab(String encodeKeytab) {
+        byte[] decodedBytes = Base64.getDecoder().decode(encodeKeytab);
+        return new String(decodedBytes);
```
jamisonbennett: I think converting the binary data into a `String` has unintended consequences that are based on the JVM configuration. Here is the documentation for the constructor: "Constructs a new String by decoding the specified array of bytes using the platform's default charset."

I know I said that we should probably put the binary data in the vault, but that was before I knew what our the API is for `SecretService`. It looks like that isn't a great idea based on the API which takes a string.

lacikaaa: the whole KerberosMgmtV1Service handles the keytab as String. should it be switched to sth else?

lacikaaa: if it has to be binary and we can't base64 encode it for some reason, it still could be encrypted and the key stored in vault. but in the first run I think we should try the base64 way.

kkalvagadda: you are correct, it's not a good idea to store binary data in String. I will change it.

Base64 is a way to encode binary data into an ASCII character set known to pretty much every computer system, in order to transmit the data without loss or modification of the contents itself.
I think KerberosMgmtV1Service should put the base64 encoded string into the vault. It should be converted back to binary format just before it's written to a file. Bottom line, keytab should be in base64 encoded format when it is moved around.

If you have a different opinion, let me know.

lacikaaa: I meant the FreeIPAClient should return with byte array if it's not a string, and we should have only base64 encoded stuff in String then. This still should be fixed

kkalvagadda:
@lacikaaa Thanks.

PR #5552, opened 2019-07-09
CB-2311 add api method to return status of allocated database server

gergopapi2:


gergopapi2:
> I think this looks fine, and renaming / repurposing DatabaseServerAllocationOutcomeV4Response makes sense. Two suggestions:
> 
> 1. Rename the class (again) to just DatabaseServerStatusV4Response, just because nothing else starts with "External".
> 2. Hold on to this until the termination work in CB-1248 is done, and rebase on that. Then, see if the termination response API object should also be replaced with DatabaseServerStatusV4Response.

Thanks!

Comment to 2.: I am not sure if name DatabaseServerStatus would make sense, as registered databases do not have a status - or we could just extend the Status enum with a field registered to show that it is not managed by CDP.

PR #5557, opened 2019-07-09
POC - redbeams databaseservers api suggestion (never merge)

gergopapi2:
Please have a look at just DatabaseServerV4Endpoint, and within that tothe api paths. Return types are not always updated.

I basically suggest a better distinction between allocated and registered databases as their representations (entities and dtos) are I think different enough.

Another option would be to merge RegisteredDatabaseServer and AllocatedDatabaseServer as much as possible where either the registered or the allocated part of the request / response is filled in.

What is your opinion? What other options do we have?

This is a POC, never merge it.


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -56,42 +56,42 @@ DatabaseServerV4Responses list(@NotNull @QueryParam("environmentId") String envi
     DatabaseServerV4Response get(@NotNull @QueryParam("environmentId") String environmentId, @PathParam("name") String name);
 
     @POST
-    @Path("external")
+    @Path("/managed")
```
topolyai5: what does it mean: `managed`? Do you have any other _type_? Why do you want to add this to the URL?

PR #5574, opened 2019-07-10
CB-1818 Update redbeams logging

bhavanki:
Redbeams logging is improved in several ways. First, a new MDCFilter
inserts the request ID and user CRN associated with every request into
the MDC. Second, the new ControllerLogContextAspects aspect intercepts
calls to redbeams controllers and updates the MDC based on method
arguments. It is worth noting that the corresponding aspect in other
services also stores the request ID, but that is in fact redundant with
the filter.

The logback.xml file for redbeams is updated to match those of other
services. It pulls out expected values from the MDC, and so makes
log entries much more useful and relevant.

Additional logging has been added to AbstractRedbeamsTerminationAction
and RedbeamsTerminationService, to bring them in line with corresponding
logging during the creation flow.

### DIFF
--- a/redbeams/src/main/java/com/sequenceiq/redbeams/aspect/ControllerLogContextAspects.java
+++ b/redbeams/src/main/java/com/sequenceiq/redbeams/aspect/ControllerLogContextAspects.java
@@ -1,0 +1,55 @@
+package com.sequenceiq.redbeams.aspect;
+
+import java.util.Arrays;
+
+import javax.inject.Inject;
+
+import org.aspectj.lang.JoinPoint;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Before;
+import org.aspectj.lang.annotation.Pointcut;
+import org.aspectj.lang.reflect.CodeSignature;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.logger.LogContextService;
+
+@Component
+@Aspect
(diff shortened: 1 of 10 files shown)


redbeams/src/main/java/com/sequenceiq/redbeams/aspect/ControllerLogContextAspects.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(ControllerLogContextAspects.class);
+
+    @Inject
+    private LogContextService logContextService;
+
+    @Pointcut("execution(public * com.sequenceiq.redbeams.controller..*(..))")
+    public void interceptControllerMethodCalls() {
+    }
+
+    @Before("com.sequenceiq.redbeams.aspect.ControllerLogContextAspects.interceptControllerMethodCalls()")
```
meowfaceman: What's the point of interceptControllerMethodCalls here? I'm guessing just syntactic nicety, but it's not actually doing anything useful here AFAICT.

bhavanki: It's how AspectJ defines pointcuts. You can see the method call stated in the @Before annotation right afterwards. If you take a look at SecretAspects, you can see several more examples. :D

meowfaceman: I have a feeling this isn't necessary, but I don't feel that strongly about this. Check out the PipelineJobProfiler in Director. I don't think the empty method with a Pointcut is necessary.

PR #5575, opened 2019-07-10
CDPSDX-596: Set "Enable Knox Trusted Proxy Support" to true for Range…

kkalvagadda:
added configuration "ranger.authentication.allow.trustedproxy" with default value "true" to ranger admin service. This change is done in sdx, sdx-lite and sdx-medium blueprints.


keyki:
This is enabled by default for every SDX cluster
https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/ranger/RangerKnoxRoleConfigProvider.java

PR #5581, opened 2019-07-11
COMPX-333. Error while creating clusters with whitelisted set of polc…

rohithsharmaks:
Liftie managaes KMS and attache/detach autoscaller policy. 

Patch adds up KMS and IMA policy actions required for liftie.


rohithsharmaks:
Thanks @doktoric !

PR #5582, opened 2019-07-11
CB-2348 DatabaseServerConfig: add GET by Crn endpoint without environ…

gergopapi2:


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -52,25 +52,32 @@ DatabaseServerV4Responses list(@NotNull @QueryParam("environmentId") String envi
...
+    @GET
+    @Path("/crn/{crn}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = DatabaseServerOpDescription.GET_BY_CRN, produces = MediaType.APPLICATION_JSON, notes = Notes.DATABASE_SERVER_NOTES,
+            nickname = "getDatabaseServerByCrn")
+    DatabaseServerV4Response getByCrn(@PathParam("crn") String crn);
+
     @POST
-    @Path("external")
+    @Path("managed")
```
bhavanki: I know I said I'm bad at naming things, but I'll still say that I don't like "managed" here, because it doesn't say _who_ is managing - is it redbeams or something else? At least "external" means "outside", which seems clear to me is not redbeams.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -52,25 +52,32 @@ DatabaseServerV4Responses list(@NotNull @QueryParam("environmentId") String envi
...
-            nickname = "getDatabaseServer")
+            nickname = "getDatabaseServerByNameOrCrn")
     DatabaseServerV4Response getByNameOrCrn(@NotNull @QueryParam("environmentId") String environmentId, @PathParam("nameOrCrn") String nameOrCrn);
 
+    @GET
+    @Path("/crn/{crn}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = DatabaseServerOpDescription.GET_BY_CRN, produces = MediaType.APPLICATION_JSON, notes = Notes.DATABASE_SERVER_NOTES,
+            nickname = "getDatabaseServerByCrn")
+    DatabaseServerV4Response getByCrn(@PathParam("crn") String crn);
```
bhavanki: This endpoint looks perfect. My question, though: If we have this one, does "getByNameOrCrn" need to support looking up a server by CRN at all?

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
gergopapi2: I see. What about serviceManaged?

External to me is confusing, it could mean:
1) external in relation to redbeams, because the user manages the lifecycle of it, redbeams only uses it
2) external in relation to the user, redbeams manages it.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
gergopapi2: Totally agree, I just did not want to touch that. Then I will.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
bhavanki: I see you have the same confusion with "external" that I have with "managed". :)

To me, "serviceManaged" doesn't help - which service? Redbeams?

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
gergopapi2: @bhavanki would it be ok for you to do that as part of another PR? I think it would make sense to change our api to GET, DELETE, PUT by crn only, as that seems to be the common currency.

PR #5583, opened 2019-07-11
Knox topology fixes for various services

keyki:


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -201,7 +171,7 @@
     <service>
         <role>ZEPPELINWS</role>
         {% for hostloc in salt['pillar.get']('gateway:location')['ZEPPELIN_SERVER'] -%}
-        <url>ws://{{ hostloc }}:9995/ws</url>
+        <url>ws://{{ hostloc }}:{{ ports['ZEPPELIN'] }}</url>
```
risdenk: Need to keep the `/ws` on the end of this. Sorry didn't get to review this quicker.

risdenk: It should be `<url>ws://{{ hostloc }}:{{ ports['ZEPPELIN'] }}/ws</url>`

keyki: Hm, I've tested this and worked without /ws

keyki: I'll double check.

risdenk: @keyki and I checked this and looks like I was wrong and don't need the `/ws`.

PR #5584, opened 2019-07-11
CB-2148 calling redbeams for databaseserver info

lnardai:
dummy-dummy, please don't look


lnardai:
This is already merged with a another PR to an rc branch.

PR #5591, opened 2019-07-11
CB-1579 Workflow to create the FreeIPA bind accounts per cluster

lacikaaa:


core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        ServiceKeytabRequest request = new ServiceKeytabRequest();
+        request.setEnvironmentCrn(stack.getEnvironmentCrn());
+        request.setServerHostName(primaryGatewayConfig.getHostname());
+        request.setServiceName("CM");
+        request.setDoNotRecreateKeytab(Boolean.TRUE);
+        RoleRequest roleRequest = new RoleRequest();
+        roleRequest.setRoleName("hadoopadminrole-" + stack.getName());
+        roleRequest.setPrivileges(Set.of("Service Administrators"));
+        request.setRoleRequest(roleRequest);
+        return kerberosMgmtV1Endpoint.generateServiceKeytab(request);
```
kkalvagadda: Can you use api "getServiceKeytab", instead? When you use this getServiceKeytab, there is no need to pass doNotRecreateKeytab in the request.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        String principal = secretService.getByResponse(serviceKeytabResponse.getServicePrincial());
+        return new KeytabModel("CM", "/etc/cloudera-scm-server", "cmf.keytab", principal, keytab);
+    }
+
+    private ServiceKeytabResponse getServiceKeytabResponse(Stack stack, GatewayConfig primaryGatewayConfig)
+            throws Exception {
+        ServiceKeytabRequest request = new ServiceKeytabRequest();
+        request.setEnvironmentCrn(stack.getEnvironmentCrn());
+        request.setServerHostName(primaryGatewayConfig.getHostname());
+        request.setServiceName("CM");
```
kkalvagadda: There could be other use cases that would want to use "KeytabConfigurationHandler". For example MLX.

MLX would want to use KeytabConfigurationHandler to create ketyab with administrative permissions. Having hardcoded strings like "CM" would need a rewrite if it has to be reused. Can you make these API's generic so that they can be re-used?

I'm giving this comment with the limited knowledge I have. Please ignore if it doesn't make sense.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        request.setDoNotRecreateKeytab(Boolean.TRUE);
+        RoleRequest roleRequest = new RoleRequest();
+        roleRequest.setRoleName("hadoopadminrole-" + stack.getName());
```
kkalvagadda: Same comment as above.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -149,12 +160,45 @@ private String serviceAdd(String serviceName, String hostname, String realm, Str
...
+            boolean roleSetForService = service.getMemberOfRole().stream().anyMatch(member -> member.contains(roleRequest.getRoleName()));
+            if (!roleSetForService) {
+                ipaClient.addRoleMember(role.getCn(), null, null, null, null, Set.of(service.getKrbprincipalname()));
+            }
+        }
+    }
+
+    private void addPrivilegesToRole(Set<String> privileges, FreeIpaClient ipaClient, Role role) throws FreeIpaClientException {
+        if (privileges != null) {
+            Set<String> privilegesToAdd = privileges.stream().filter(privilegeName -> {
```
kkalvagadda: Can you please explain the intent of this block? 
I think member in the privilege would be roles, right. What is the logic looking for "privilegeName" in members of privilege with name "privilegeName".

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
jamisonbennett: I was going to comment elsewhere but I will add it here. There is a get and post method for /v1/kerberosmgmt/servicekeytab. The intent was that post method will always create a new keytab and if there are existing keytabs it will always invalidate them. The get method will never invalidate the existing keytabs but it will also never create a keytab.

It seems that doNotRecreateKeytab seems to enable the post method to be used to attempt to encapsulate logic that a client already can do try a "get" and if it fails then do a "post". The problem I see with hiding this logic in the FreeIPA management service is that is that it may actually lead to invalidations in some race cases. Consider what happens if this is called twice in a very short period of time. When we move to an HA environment, the time period of race condition expose is enlarged. In HA, FreeIPA operates with multimaster. The later timestamp will win. This means the second request can actually cause the regeneration of the keytab and thus cause the invalidation of the first.

So I think doNotRecreateKeytab will mislead the caller of the API and lead to situations that are not possible to guarantee (since its not supported by the underlying FreeIPA APIs).

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -75,12 +81,16 @@ public ServiceKeytabResponse generateServiceKeytab(ServiceKeytabRequest request,
...
-        String adminUser = freeIpaClientFactory.getAdminUser();
         hostAdd(request.getServerHostName(), ipaClient);
-        String servicePrincipal = serviceAdd(request.getServiceName(), request.getServerHostName(), realm, adminUser, ipaClient);
-        String serviceKeytab = getServiceKeytab(servicePrincipal, ipaClient);
+        com.sequenceiq.freeipa.client.model.Service service = serviceAdd(request, realm, ipaClient);
+        String serviceKeytab;
+        if (service.getHasKeytab() && request.getDoNotRecreateKeytab()) {
+            serviceKeytab = getExistingServiceKeytab(service.getKrbprincipalname(), ipaClient);
+        } else {
+            serviceKeytab = getServiceKeytab(service.getKrbprincipalname(), ipaClient);
```
jamisonbennett: The race condition with `doNotRecreateKeytab` that I mentioned earlier can cause this line to be executed more than once resulting in invalidation of the initial keytab.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: the main reason for this to avoid 2 calls from CB to FreeIPA. If we don't have this option, then we have to make a call with get to see if there is already a keytab for the service. If no, we can call the create. This way our call from CB is idempotent, we can call it many times.

From CB side I can't see the issue of calling this method in a short time causing race condition, but if it's an issue from other services I can rewrite it to do a call with get.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: yes, Jamison is right. This logic is in CB for generating keytab for CM, and for CM I think the right choice is CM. Do you think we should expose this variable on our API? I'm not sure about that.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: it's still internal stuff for CB. We use these to replace this script for generating keytabs: https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/sssd/template/generate_cm_keytab.j2

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
lacikaaa: good catch! That should be the roleName

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -149,12 +160,45 @@ private String serviceAdd(String serviceName, String hostname, String realm, Str
...
+                ipaClient.addRoleMember(role.getCn(), null, null, null, null, Set.of(service.getKrbprincipalname()));
+            }
+        }
+    }
+
+    private void addPrivilegesToRole(Set<String> privileges, FreeIpaClient ipaClient, Role role) throws FreeIpaClientException {
+        if (privileges != null) {
+            Set<String> privilegesToAdd = privileges.stream().filter(privilegeName -> {
+                try {
+                    Privilege privilege = ipaClient.showPrivilege(privilegeName);
```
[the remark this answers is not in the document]
lacikaaa: it would make RPC calls, but won't be duplicate as the privilegeName is different

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
keyki: Yep, this keytab should not be granted to any other applications, besides CM.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
keyki: If we try to get first then post we could encounter the same race condition, aren't we?

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
lacikaaa: I think if here we can have a race condition, we can have it with 2 calls.
2 calls come to the get, both fails as there is no keytab. Both calls the create, the latter wins. I think using this 2 call solution won't solve the race condition itself. If we want to really solve it, then we have to block the read for the service if there is a create in operation for the same service, and drop new creates until there is a running create.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: yes, see: https://github.com/hortonworks/cloudbreak/pull/5591#discussion_r302852273

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
jamisonbennett: @keyki , you are correct. The race condition exists no matter what we do. I am just suggesting that we not try to pretend there isn't a race condition. Regardless of `doNotRecreateKeytab`, the race condition is present. This API will be exposed to our internal workloads and I just don't want them to be surprised when `doNotRecreateKeytab=True` actually recreates a keytab and thus invalidates prior keytabs. I guess I am OK with adding the `doNotRecreateKeytab` flag as long as the API documentation warns that it is a "best effort" which is not guaranteed and may actually invalidate prior keytabs.

As to the short time, what happens if there is an HA scenario and a temporary network partition between the FreeIPA masters? Two post requests could be mapped to different FreeIPA servers. They are multi-master so there will be two valid responses. But after the network partition is healed, only one of the two keytabs will be usable. This network partition could potentially last longer than a "short period of time".

But I understand the desire to minimize the number of round trips requests, so can we just reword the documentation a little bit?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
kkalvagadda: I understand that this code was added to address the generation of CM keytab but there will be other cases where we would want to use KeytabConfigurationHandler to create keytab for other services.

keyki: I don't get it. How would you use this handler. This is cloudbreak code.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: ok, I will modify the docs according to your suggestion

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabConfigurationHandler.java
lacikaaa: I'm not aware of any other usecase where CB needs a keytab. If it happens we will have to rewrite this piece of code

PR #5593, opened 2019-07-11
Add error handling for gcp

tiborpopovics:


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -169,7 +169,7 @@
         insert.setPrettyPrint(Boolean.TRUE);
         try {
             Operation operation = insert.execute();
-            if (operation.getHttpErrorStatusCode() != null) {
+            if (operation.getHttpErrorStatusCode() != null && !operation.getError().isEmpty()) {
```
pdarvasi: I think there should be an "or" relation here

PR #5596, opened 2019-07-12
CB-879 set base image when CM or CDP repo provided

keyki:


lajosrodek:
Deprecated by #5598 .

PR #5600, opened 2019-07-12
CB-1792 isolated environment service integration tests

afarsang:
isolated environment service integration tests, it runs on h2 in-memory database with mocked connection (e.g. vault). Proxy, and credential tests are included

Closes CB-1792


environment/src/test/java/com/sequenceiq/environment/service/integration/JpaTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.sequenceiq.environment.proxy.repository.ProxyConfigRepository;
+import com.sequenceiq.environment.proxy.v1.ProxyTestSource;
+
+@DataJpaTest
+@EntityScan(basePackages = {"com.sequenceiq.flow.domain",
+        "com.sequenceiq.environment",
+        "com.sequenceiq.cloudbreak.ha.domain"})
+@Tag("outofscope")
+public class JpaTest {
```
topolyai5: It is a too common name. I think we will create more test with a different entity, right?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/endpoint/CredentialEndpoint.java
```diff
@@ -112,10 +112,10 @@
     CredentialPrerequisitesResponse getPrerequisitesForCloudPlatform(@PathParam("cloudPlatform") String platform,
             @QueryParam("deploymentAddress") String deploymentAddress);
 
-    @GET
+    @POST
```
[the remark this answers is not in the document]
afarsang: maybe, but it won't work if it has body

environment/src/test/java/com/sequenceiq/environment/service/integration/JpaTest.java
afarsang: now it is more likely an example how to test database interactions. I am not hundred percent sure we need that, but in the future if we have more complex DB queries (@attilapalfi92  mentioned something like these recently, didn't he?)

environment/build.gradle
```diff
@@ -90,7 +90,7 @@ dependencies {
   }
   testImplementation group: 'com.openpojo',              name: 'openpojo'
   testImplementation group: 'org.ow2.asm',               name: 'asm',                            version: asmVersion
-
+  testCompile group: 'com.h2database', name: 'h2', version: '1.3.148'
```
biharitomi: Could you please use the same indentation?

PR #5603, opened 2019-07-12
CB-2356 redbeams create databaseserver body: remove region; make subn…

gergopapi2:


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/requests/AllocateDatabaseServerV4Request.java
```diff
@@ -29,11 +29,7 @@
 
     @NotNull
     @ApiModelProperty(value = ModelDescriptions.ENVIRONMENT_ID, required = true)
```
bhavanki: Now is a good time to replace ModelDescriptions.ENVIRONMENT_ID with ENVIRONMENT_CRN.

redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
```diff
@@ -0,0 +1,26 @@
+package com.sequenceiq.redbeams.service;
+
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public class UserGeneratorService {
```
bhavanki: Refactoring ftw

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/SubnetChooserService.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
+import com.sequenceiq.cloudbreak.exception.BadRequestException;
+
+@Service
+public class SubnetChooserService {
+
+    public List<CloudSubnet> chooseSubnetsFromDifferentAzs(List<CloudSubnet> subnetMetas) {
+        if (subnetMetas.size() < 2) {
+            throw new BadRequestException("Insufficient number of subnets: at least two subnets in two different availability zones needed");
```
bhavanki: I think RedbeamsException is better here, because this condition is not because there was something (directly) wrong with the API request. This isn't a blocker for me, though. Your call.

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/SubnetChooserServiceTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
+import com.sequenceiq.cloudbreak.exception.BadRequestException;
+
+public class SubnetChooserServiceTest {
```
bhavanki: Unit test ... awesome

redbeams/src/main/java/com/sequenceiq/redbeams/converter/stack/AllocateDatabaseServerV4RequestToDBStackConverter.java
```diff
@@ -111,11 +136,26 @@ private void updateCloudPlatformAndRelatedFields(AllocateDatabaseServerV4Request
...
+    private MappableBase getNetworkFromEnvironment(DetailedEnvironmentResponse environmentResponse) {
+        if (environmentResponse.getNetwork() == null || environmentResponse.getNetwork().getSubnetMetas() == null) {
+            throw new RedbeamsException("Environment does not contain metadata for subnets");
+        }
+        List<CloudSubnet> subnetMetas = new ArrayList<>(environmentResponse.getNetwork().getSubnetMetas().values());
+        List<String> chosenSubnetIds = subnetChooserService.chooseSubnetsFromDifferentAzs(subnetMetas).stream()
+                .map(CloudSubnet::getId)
+                .collect(Collectors.toList());
+
+        return networkParameterFactoryService.createNetworkParameters(chosenSubnetIds, environmentResponse.getCloudPlatform());
```
bhavanki: The cloud platform is already available as an enum in the AllocateDatabaseServerV4Request object. It would be better to use that so that the service called here can switch on the enum instead of strings.

This does surface a problem - what happens when the cloud platform in the environment is different from the cloud platform in the request? That's something to think about as a separate item.

gergopapi2: That was exactly my concern as well and that's why I did go with the "inferior" string solution instead of a nice enum. If we digest what the env returns we are safe.

Yes, I agree, we should check if the providers do match in the request and the actual env. After that, however, we could use the enums.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/requests/AllocateDatabaseServerV4Request.java
gergopapi2: Thanks, done.

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/SubnetChooserService.java
gergopapi2: Ok, I did change this to redbeams exception. This will be mapped to INTERNAL_SERVER_ERROR. To the end user this perfectly makes sense. On the other hand, to the direct consumers of our api I am not sure if this is ok.

redbeams/src/main/java/com/sequenceiq/redbeams/converter/stack/AllocateDatabaseServerV4RequestToDBStackConverter.java
gergopapi2: I created this jira to follow up:
https://jira.cloudera.com/browse/CB-2425

PR #5604, opened 2019-07-12
CB-1923 redbeams api improvements

gergopapi2:


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -91,18 +98,32 @@ DatabaseServerV4Responses list(@NotNull @QueryParam("environmentId") String envi
...
+    @DELETE
+    @Path("/name/{name}")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = DatabaseServerOpDescription.DELETE_BY_NAME, produces = MediaType.APPLICATION_JSON, notes = Notes.DATABASE_SERVER_NOTES,
-        nickname = "deleteDatabaseServer")
-    DatabaseServerV4Response delete(@NotNull @QueryParam("environmentId") String environmentId, @PathParam("name") String name);
+            nickname = "deleteDatabaseServerByName")
+    DatabaseServerV4Response deleteByName(@NotNull @QueryParam("environmentCrn") String environmentCrn, @PathParam("name") String name);
+
+//    @DELETE
```
bhavanki: Looks like this commented-out stuff can go

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/base/DatabaseServerV4Identifiers.java
```diff
@@ -1,45 +1,27 @@
...
-
-    public String getName() {
-        return name;
-    }
-
-    public void setName(String name) {
-        this.name = name;
-    }
+    @ApiModelProperty(value = CRN, required = true)
+    private String resourceCrn;
```
bhavanki: As a separate change, not now: This should be annotated with @ValidCrn. I bet there are other places the API should use it ...

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/base/DatabaseServerV4Identifiers.java
```diff
@@ -1,45 +1,27 @@
...
-import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.sequenceiq.redbeams.doc.ModelDescriptions;
-import com.sequenceiq.redbeams.doc.ModelDescriptions.DatabaseServer;
 
 import io.swagger.annotations.ApiModelProperty;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class DatabaseServerV4Identifiers implements Serializable {
```
bhavanki: Since this class now only holds a single field, should it be eliminated?

redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/provision/AbstractRedbeamsProvisionAction.java
```diff
@@ -68,7 +69,7 @@ protected RedbeamsContext createFlowContext(FlowParameters flowParameters,
         CloudContext cloudContext = new CloudContext(dbStack.getId(), dbStack.getName(), dbStack.getCloudPlatform(), dbStack.getPlatformVariant(),
                 location, userName, userName, accountId);
         // FIXME must use CRN
```
bhavanki: This comment can go away! 😃

redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
```diff
@@ -76,7 +77,7 @@ protected RedbeamsContext createFlowContext(FlowParameters flowParameters,
             cloudContext = new CloudContext(dbStack.getId(), dbStack.getName(), dbStack.getCloudPlatform(), dbStack.getPlatformVariant(),
                     location, userName, accountId);
             // FIXME must use CRN
```
bhavanki: This one too

redbeams/src/main/java/com/sequenceiq/redbeams/repository/DBStackRepository.java
```diff
@@ -1,15 +1,19 @@
...
 import org.springframework.data.jpa.repository.JpaRepository;
 
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.redbeams.domain.stack.DBStack;
+
 @Transactional(Transactional.TxType.REQUIRED)
 public interface DBStackRepository extends JpaRepository<DBStack, Long> {
 
-    Optional<DBStack> findByNameAndEnvironmentId(String name, String environmentId);
+    Optional<DBStack> findByNameAndEnvironmentCrn(String name, String environmentId);
```
bhavanki: Does this work with the changed method name? I thought that JPA devised the query based on the method name. I could be wrong.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/database/request/CreateDatabaseV4Request.java
```diff
@@ -18,11 +16,8 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class CreateDatabaseV4Request implements Serializable {
 
-    @ApiModelProperty(ENVIRONMENT_CRN)
-    private String environmentCrn;
-
     @ApiModelProperty(DatabaseServer.NAME)
```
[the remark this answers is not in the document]
gergopapi2: Thanks, done.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/base/DatabaseServerV4Identifiers.java
gergopapi2: Jira for that: https://jira.cloudera.com/browse/CB-2452

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/base/DatabaseServerV4Identifiers.java
gergopapi2: Yes, thank you!

redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
gergopapi2: Done

redbeams/src/main/java/com/sequenceiq/redbeams/repository/DBStackRepository.java
gergopapi2: In some of the entities (where there were no handcrafted queries) I did rename the field environmentId to environmentCrn, so this should work. 

On the other hand I am thinking we should retain environmentId, just to decouple the concept of an id from its representation as a crn. So I rename it back to environmentId. And, I think it would be better to rename this field for all or none of the entities but not just some.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -45,23 +43,22 @@
...
-    @Path("{nameOrCrn}")
-    @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = DatabaseServerOpDescription.GET_BY_NAME_OR_CRN, produces = MediaType.APPLICATION_JSON, notes = Notes.DATABASE_SERVER_NOTES,
-            nickname = "getDatabaseServerByNameOrCrn")
-    DatabaseServerV4Response getByNameOrCrn(@NotNull @QueryParam("environmentId") String environmentId, @PathParam("nameOrCrn") String nameOrCrn);
+    DatabaseServerV4Responses list(@NotNull @QueryParam("environmentCrn") String environmentCrn);
 
     @GET
-    @Path("/crn/{crn}")
+    @Path("{crn}")
```
topolyai5: Use the same pattern as the other module, for example: `EnvironmentEndpoint`

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
```diff
@@ -76,7 +78,7 @@
     @Inject
     private UserGeneratorService userGeneratorService;
 
-    public Set<DatabaseServerConfig> findAll(Long workspaceId, String environmentId, Boolean attachGlobal) {
+    public Set<DatabaseServerConfig> findAll(Long workspaceId, String environmentId) {
```
topolyai5: We don't use workspaceId in microservices as well

redbeams/src/main/java/com/sequenceiq/redbeams/domain/stack/DBStack.java
```diff
@@ -33,6 +33,10 @@
     @SequenceGenerator(name = "dbstack_generator", sequenceName = "dbstack_id_seq", allocationSize = 1)
     private Long id;
 
+    @Convert(converter = CrnConverter.class)
+    @Column(nullable = false)
+    private Crn resourceCrn;
```
topolyai5: What is the reason for Crn object? We add more unnecessary conversation to the process.

redbeams/src/test/java/com/sequenceiq/redbeams/controller/v4/database/DatabaseV4ControllerTest.java
```diff
@@ -133,8 +133,8 @@ public void testRegister() {
     // public void testTestWithName() {
```
topolyai5: Could you please delete the commented lines if you don't need for these lines anymore?

redbeams/src/test/java/com/sequenceiq/redbeams/converter/spi/DBStackToDatabaseStackConverterTest.java
```diff
@@ -22,6 +13,15 @@
 import org.junit.Test;
 import org.mockito.InjectMocks;
 
+import com.sequenceiq.cloudbreak.api.endpoint.v4.common.DatabaseVendor;
```
topolyai5: Unnecessary changes

redbeams/src/test/java/com/sequenceiq/redbeams/domain/stack/DBStackTest.java
```diff
@@ -2,16 +2,16 @@
...
-import com.sequenceiq.cloudbreak.auth.altus.Crn;
-import com.sequenceiq.cloudbreak.common.json.Json;
-import com.sequenceiq.redbeams.api.model.common.Status;
-
 import java.util.Map;
 
 import org.junit.Before;
 import org.junit.Test;
 
+import com.google.common.collect.ImmutableMap;
```
topolyai5: Unnecessary changes

redbeams/src/test/java/com/sequenceiq/redbeams/flow/redbeams/provision/AbstractRedbeamsProvisionActionTest.java
```diff
@@ -4,6 +4,12 @@
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
 
+import org.junit.Before;
```
topolyai5: Unnecessary changes

redbeams/src/test/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationActionTest.java
```diff
@@ -4,6 +4,14 @@
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
 
+import java.util.Optional;
```
topolyai5: Unnecessary changes

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
gergopapi2: Yes. It remained there because of historic reasons ... I do not want toe delete it as part of this jira but we could get rid of it. @bhavanki, what do you suggest?

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
gergopapi2: Thanks.

First, EnvironmentEndpoint and e.g. SdxEndpoint use a somewhat different pattern. Namely, in sdx you have /{name} and /crn/{crn}.
Since in our case the main way to reach resources is via crn, we made the /{crn} the main way to get a resource and /name/{name the secondary. 

If you still think we should have /crn/{crn}, we can change.

redbeams/src/main/java/com/sequenceiq/redbeams/domain/stack/DBStack.java
gergopapi2: Yes, string would be easier. 
This way, however, 
1) you can be sure that the stored string can be mapped to a crn, and
2) the type of the variable tells you what this thing is (crn), not the name.

redbeams/src/test/java/com/sequenceiq/redbeams/controller/v4/database/DatabaseV4ControllerTest.java
gergopapi2: I created a jira to work on the DatabaseConfig api part.
https://jira.cloudera.com/browse/CB-2480

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
bhavanki: At this point, I'm fine getting rid of it. It's becoming more of a wart.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
bhavanki: We did discuss the path format for CRN vs name for redbeams. We looked at the environment and freeipa services, and found that they have different standards. When I look right now, environment uses /crn and /name, but freeipa doesn't use either - environments passed by CRN are passed to endpoints without "/crn". (If I recall correctly, when we looked earlier, freeipa was using /crn but not /name.)

So, we decided to not use "/crn" because the CRN fits the definition of a RESTful resource identifier, and it's architecturally better to focus on it as the primary identifier in the API. So, "/crn" would not be necessary.

PR #5512, opened 2019-07-04
CDPSDX-584: Add the implementation to create and retrieve keytabs

jamisonbennett:
Added the ability to create and retrieve keytabs. When creating a
keytab, the host and service principal are added. If they already
exist, the operation will proceed without an error.

If a keytab is created for the same service principal more than once,
only the latest keytab is usable. Prior keytabs are invalidated. The
retrieve operation can be used to retrieve an existing keytab without
invalidating the prior keytabs.

The result from the get_keytab API on the FreeIPA server was updated
to be structured similar to other FreeIPA APIs. This enables the
FreeIPA client to be used on the FreeIPA management service.

Unit tests were updated and the changes were manually tested using
the REST API on the FreeIPA management service. The keytabs were
validated against the FreeIPA server.

Co-authored-by: Jamison Bennett <jamison.bennett@gmail.com>

Closes CDPSDX-584

PR #5608, opened 2019-07-12
Backport the kerberos keytab functionality to 2.10.

jamisonbennett:
This backports the keytab changes into the 2.10 branch so that the DWX team can use them before the 2.11 sprint is complete.

Since this is a new feature being added to the RC branch, I want to call out that I have a deployment locally. The FreeIPA server was created successfully but the SDX cluster failed to provision with the reason "RuntimeException: Can not start provisioning, client error happened on Cloudbreak side: HTTP 400 Bad Request". The same error message is present when I locally deploy without this change.


lajosrodek:
Corresponding master commits: #5512 , #5549 .

PR #5613, opened 2019-07-14
CDPAM-124 | Setting dp user 'cmmgmt' as the default service user when…

dvguruprasad:


keyki:
rc-2.10 should be merged to master thus this PR is not necessary

PR #5621, opened 2019-07-15
CB-2250 - Cloudbreak selects wrong image - ignore not matching OS for…

pdarvasi:


keyki:
Do we already have a centos image in the catalog?

pdarvasi:
> Do we already have a centos image in the catalog?

It is being built now (http://build.eng.hortonworks.com:8080/view/All/job/cloudbreak-AWS-multi-packer-image-builder-salt/26/console) but it should not matter as we should select the image available.

PR #5638, opened 2019-07-17
CB-2501 Redbeams dbstack password too long

gergopapi2:


redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
```diff
@@ -23,4 +23,8 @@ public String generateUserName() {
     public String generatePassword() {
         return UUID.randomUUID().toString();
     }
+
+    public String generatePassword(int maxLength) {
+        return UUID.randomUUID().toString().substring(1, maxLength);
```
bhavanki: Should be 0, not 1. But see my overall comment on a better way to drop characters.

gergopapi2:
> The maximum length of the password should be determined by the cloudPlatform, and maybe also the database vendor, of the allocation request. Soon enough we will be working with Azure, and it may have different limitations. It'd be enough to make the user generator service take those factors as arguments.
> 
> A better way to take characters away from the UUID than just grabbing a substring:
> 
> 1. Drop all the hyphens. These don't contribute to its randomness.
> 2. Drop the first digits of the third and fourth groups. Parts of those are deterministic, so they are "less random" than the other digits (for a type 4 UUID, which this is).
> 
> https://en.wikipedia.org/wiki/Universally_unique_identifier#Format



> The maximum length of the password should be determined by the cloudPlatform, and maybe also the database vendor, of the allocation request. Soon enough we will be working with Azure, and it may have different limitations. It'd be enough to make the user generator service take those factors as arguments.
> 
> A better way to take characters away from the UUID than just grabbing a substring:
> 
> 1. Drop all the hyphens. These don't contribute to its randomness.
> 2. Drop the first digits of the third and fourth groups. Parts of those are deterministic, so they are "less random" than the other digits (for a type 4 UUID, which this is).
> 
> https://en.wikipedia.org/wiki/Universally_unique_identifier#Format

Great, thank you for the insight.
As for refactoring for the differences between cloud providers: I did not work for a nice solution as we do not know what azure will require. I think a refactor will be possible when we implement for azure.

redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
gergopapi2: Yes, I revised the implementation.

redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
```diff
@@ -13,14 +17,34 @@
...
     }
 
+    /*
+    Generates a password using a UUID.
+    Provider specific limitations:
+    - AWS: password length in cf template can be 30 characters long at most
+     */
     public String generatePassword() {
-        return UUID.randomUUID().toString();
+        Pattern pattern = Pattern.compile(UUID_EXTRACTOR_REGEXP);
```
bhavanki: Instead of compiling the pattern each time this method is called, just compile it once and make it a static final. Pattern objects are safe to share across multiple threads.

redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
```diff
@@ -13,14 +17,34 @@
...
+        String uuid = uuidGeneratorService.randomUuid();
+        Matcher matcher = pattern.matcher(uuid);
+        if (!matcher.matches()) {
```
bhavanki: How will the UUID string, which was just generated, ever not match the pattern?

redbeams/src/main/java/com/sequenceiq/redbeams/service/UserGeneratorService.java
```diff
@@ -13,14 +17,34 @@
...
+            return uuid.substring(0, MAX_LENGTH_PASSWORD < uuid.length() ? MAX_LENGTH_PASSWORD - 1 : uuid.length());
```
bhavanki: (If this remains) Don't subtract one from MAX_LENGTH_PASSWORD, or else the resulting string will be MAX_LENGTH_PASSWORD - 1 characters in length.

PR #5640, opened 2019-07-17
CB-2500 Cluster creation fails due to missing compute in host template

doktoric:


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -899,10 +899,8 @@ public Cluster recreate(Stack stack, String blueprintName, Set<HostGroup> hostGr
                 throw new BadRequestException("Ambari doesn't support resetting external DB automatically. To reset Ambari Server schema you must first drop "
                         + "and then create it using DDL scripts from /var/lib/ambari-server/resources");
             }
-            if (validateBlueprint) {
-                BlueprintValidator blueprintValidator = blueprintValidatorFactory.createBlueprintValidator(blueprint);
-                blueprintValidator.validateBlueprintForStack(blueprint, hostGroups, stackWithLists.getInstanceGroups());
-            }
+            BlueprintValidator blueprintValidator = blueprintValidatorFactory.createBlueprintValidator(blueprint);
+            blueprintValidator.validateBlueprintForStack(blueprint, hostGroups, stackWithLists.getInstanceGroups(), validateBlueprint);
```
attilapalfi92: It's little weird that this is a `validateBlueprintForStack` method and a `validateBlueprint` boolean has to be passed to it. I think a better refactor is necessary. E.g. create a simple `HostGroupValidator` class that has a `validate` method and is always called. Move the relevant code from `validateBlueprintForStack` to `HostGroupValidator`

attilapalfi92: Then of course the boolean don't have to be passed, the current logic can be kept.

PR #5649, opened 2019-07-17
DISTX-258. Workload analytics installation improvements.

oleewere:
Contains a few changes:
- move altus credential reading to auth mock side (so from that point you will be able to provide your altus secret / access keypair for auth app - so the mock will send that back if you will try to create a keypair )
- use sdx STACK crn tp gather sdx id for telemetry publisher (also store sdx crn on stack side - nullable)
- installing telemetry publisher feature can be enabled by default without providing in the request (of course using in the request can install it even if it is disabled...it makes sense as this way as you can install TP manually anyway if you go to CM UI)


oleewere:
open question: 
sdx crn is the stack crn here ... maybe would be more correct to use the real sdx crn ? (I'm not sure WXM will use sdx public api to use these in queries) ... then it would be better to pass sdxCrn with stack request (same as it's done with environmentCrn) and set it in DatalakeResource

attilapalfi92:
I'm not very familiar with these parts of the code therefore I would rather not judge it.

oleewere:
@sodre90 , sdx/distrox cluster installation finished successfully. this change does not broke actual flows so it can be merged.

PR #5652, opened 2019-07-17
DISTX-266. Reject concurrent user sync requests

handavid:
The FMS was enhanced to better track SyncOperations and reject
concurrent requests.

The syncoperation table was modified
to add the environmentList and userList columns to represent
the scope of the sync operation. The SyncOperation entity was
modified to use new AttributeConverters to hide the conversion
to/from Json.

When requested, the FMS will store the SyncOperation in the db
then check whether to accept the request or not based on other
active requests in the db.

The SyncOperationAlreadyRunningException was created and mapped
to 409 CONFLICT.


handavid:
closing this to re-open against rc-2.11

PR #5656, opened 2019-07-18
Backport Revert CDPSDX-595 and do not use vault to store keytabs to rc-2.11

jamisonbennett:
Backport Revert CDPSDX-595 and do not use vault to store keytabs to rc-2.11


jamisonbennett:
I meant to backport this to 2.11.

PR #5672, opened 2019-07-19
CB-2061 remove required=true from subnetIds

sodre90:


tamasfoldi:
It should go to rc-2.11

PR #5676, opened 2019-07-19
DISTX-40. Setup DistroX install to use fluentd with azure BLOB storage

oleewere:
Although no MSI support yet, WASB can be supported with the actual CB configs (provided account + access key)

- fix edit environment request with telemetry
- some small fixes recarding account + account keys
- extends fluent configs with azure blob storage support (no MSI yet)
- extract some functions inside FluentConfigService
- there is a bug with the azurestorage plugin (conflicting 2 version of fluentd), but after fixing that, outputs were generated to WASB - that will be fixed in a different place, and uploaded to the remote cldr ruby gem repo - edit: that is done already so upgrading the version to 1.0.1

wasb_test


core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -26,78 +34,127 @@
...
 
-            String clusterType = CLUSTER_TYPE_DISTROX;
-            if (StackType.DATALAKE.equals(stackType)) {
-                clusterType = CLUSTER_TYPE_SDX;
-            }
-            String s3bucket = null;
-            String s3LogFolderName = null;
-            boolean fluentEnabled = false;
-            String providerPrefix = DEFAULT_PROVIDER_PREFIX;
+            String clusterType = StackType.DATALAKE.equals(stack.getType()) ? CLUSTER_TYPE_SDX : CLUSTER_TYPE_DISTROX;
```
lajosrodek: Not a big deal but do we really want to name data lake folders as `"sdx"`? According to the [CDP Glossary](https://docs.google.com/document/d/1SyXqp0fhjhEmwLCSeScon-i3Lf7HKnwRpDkgU-dKPHI/edit), "Data Lake Cluster" is official term used in marketing and UI labels (even though internal CB API endpoints and CB CLI still refer to them as "sdx clusters").

oleewere: then it probably can be instead of distrox or sdx like: datalake and datahub

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -26,78 +34,127 @@
...
+            String logFolderName = Paths.get(CLUSTER_LOG_PREFIX, clusterType, clusterName).toString();
+
+            builder.withPlatform(platform)
+                    .withOverrideAttributes(
+                            logging.getAttributes() != null ? new HashMap<>(logging.getAttributes()) : new HashMap<>()
```
lajosrodek: Alternatively: `java.util.Objects.requireNonNullElse​(new HashMap<>(logging.getAttributes()), new HashMap<>())`.

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -1,103 +1,160 @@
...
+            return new BucketFolderPrefixPair(splitted[0], folderPrefix);
+        }
+        throw new CloudbreakServiceException("Storage location parameter is missing for S3");
+    }
+
+    private AzureBlobContainerParts generateAzureBlobContainerParts(String location) {
+        if (StringUtils.isNotEmpty(location)) {
+            String locationWithoutScheme = getLocationWithoutSchemePrefixes(location, WASB_SCHEME_PREFIXES);
+            String[] splitted = locationWithoutScheme.split("@");
+            String[] storageWithSuffix = splitted[0].split("/", 2);
```
attilapalfi92: Isn't it a possible NPE? Length could be 0.

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -1,103 +1,160 @@
...
+            String folderPrefix = storageWithSuffix.length < 2 ? null :  "/" + storageWithSuffix[1];
             if (splitted.length < 2) {
-                return new BucketFolderPrefixPair(splitted[0], null);
+                return new AzureBlobContainerParts(storageWithSuffix[0], folderPrefix, null);
```
attilapalfi92: Isn't it a possible NPE? Length could be 0.

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -1,103 +1,160 @@
...
+        }
+        throw new CloudbreakServiceException("Storage location parameter is missing for S3");
+    }
+
+    private AzureBlobContainerParts generateAzureBlobContainerParts(String location) {
+        if (StringUtils.isNotEmpty(location)) {
+            String locationWithoutScheme = getLocationWithoutSchemePrefixes(location, WASB_SCHEME_PREFIXES);
+            String[] splitted = locationWithoutScheme.split("@");
+            String[] storageWithSuffix = splitted[0].split("/", 2);
+            String folderPrefix = storageWithSuffix.length < 2 ? null :  "/" + storageWithSuffix[1];
```
attilapalfi92: Use empty string instead of null.

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
```diff
@@ -1,103 +1,160 @@
...
-            String[] splitted = path.toString().split("/", 2);
+            String locationWithoutScheme = getLocationWithoutSchemePrefixes(location, S3_SCHEME_PREFIXES);
+            String[] splitted = locationWithoutScheme.split("/", 2);
+            String folderPrefix = splitted.length < 2 ? null :  splitted[1];
+            return new BucketFolderPrefixPair(splitted[0], folderPrefix);
+        }
+        throw new CloudbreakServiceException("Storage location parameter is missing for S3");
+    }
+
+    private AzureBlobContainerParts generateAzureBlobContainerParts(String location) {
```
attilapalfi92: Extract this method and `generateBucketFolderPrefixPair` into a class called `CloudWhateverGenerator` and unit test these methods.
Replace whatever with the name of what's common in `BucketFolderPrefixPair` and `AzureBlobContainerParts`. E.g. to `CloudStorageDataHolderGenerator`, or something.

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
oleewere: no, if the location is splitted it will be at least an empty string

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
oleewere: no as for the String class checking length first, NPE could happen only splitted would be null but as the location cannot be empty at the start of the function

oleewere:
retest this please

core/src/main/java/com/sequenceiq/cloudbreak/fluent/FluentConfigService.java
oleewere: i keep it that way for now, as im copying the map, i'm creating a new one and passing the attributes as a map, so before the null check in Object.xxx function, it will produce an NPE already.

lajosrodek: Oops, you are right. Was indeed a silly suggestion as one should really check for `logging.getAttributes()` explicitly. 😊

PR #5677, opened 2019-07-19
CB-2567 arm template for db stack creation

gergopapi2:
This is the single server deployment of postgres (is HA).
Template for server group is still to be checked if needed, but will come later. 

I provide some readings for the template I created:

arm template for postgres:
https://docs.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/2017-12-01/servers

service endpoints: vnet to postgres connection without adding the postgres a public ip:
https://docs.microsoft.com/en-us/azure/postgresql/howto-manage-vnet-using-portal

### DIFF
--- a/cloud-azure/src/main/resources/templates/arm-dbstack.ftl
+++ b/cloud-azure/src/main/resources/templates/arm-dbstack.ftl
@@ -1,0 +1,194 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "dbServerName": {
+            "type": "string",
+            "metadata": {
+                "description": "Name of the database server resource."
+            }
+        },
+        "skuTier": {
+            "type": "string",
+            "allowedValues": [
+                "Basic",
+                "GeneralPurpose",
+                "MemoryOptimized"
+            ],
+            "metadata": {
+                "description": "The tier of the particular SKU, e.g. Basic. - Basic, GeneralPurpose, MemoryOptimized"
+            }
+        },
+        "skuFamily": {
+            "type": "string",
+            "metadata": {
+                "description": "The family of hardware."
+            }
+        },
+        "skuCapacity": {
+            "type": "int",
+            "metadata": {
+                "description": "The scale up/out capacity, representing server's compute units."
+            }
+        },
+        "skuSizeMB": {
+            "type": "int",
+            "metadata": {
+                "description": "The size code, to be interpreted by resource as appropriate."
+            }
+        },
+        "skuName": {
+            "type": "string",
+            "defaultValue": "GP_Gen5_2",
+            "allowedValues": [
+                "GP_Gen5_2",
+                "GP_Gen5_4",
+                "GP_Gen5_8",
+                "GP_Gen5_16",
+                "GP_Gen5_32",
+                "MO_Gen5_2",
+                "MO_Gen5_4",
+                "MO_Gen5_8",
+                "MO_Gen5_16",
+                "MO_Gen5_32"
+            ],
+            "metadata": {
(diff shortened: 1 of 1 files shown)


cloud-azure/src/main/resources/templates/arm-dbstack.ftl
```diff
@@ -0,0 +1,179 @@
...
+                "sslEnforcement": "[parameters('sslEnforcement')]",
+                "storageProfile": {
+                    "backupRetentionDays": "[parameters('backupRetentionDays')]",
+                    "geoRedundantBackup": "[parameters('geoREdundantLookup')]",
+                    "storageMB": "[parameters('skuSizeMB')]",
+                    "storageAutogrow": "Disabled"
+                },
+                "createMode": "Default"
+            },
+            "location": "NorthEurope",
```
bhavanki: Should be "[parameters('location')]" ?

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
```diff
@@ -0,0 +1,179 @@
...
+            "type": "securestring",
+            "minLength": 8,
+            "maxLength": 128,
+            "metadata": {
+                "description": "The administrator password for the database server."
+            }
+        },
+        "sslEnforcement": {
+            "type": "string",
+            "defaultValue": "Enabled",
```
bhavanki: What do you think about making this a Boolean value? Is it possible and straightforward to have the template engine convert true/false to Enabled/Disabled? I worry that the spelling will be wrong at some point. :)

Same for geoRedundantLookup, storageAutogrow.

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
```diff
@@ -0,0 +1,179 @@
...
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "dbServerName": {
+            "type": "string",
+            "metadata": {
+                "description": "Name of the database server resource."
+            }
+        },
+        "skuTier": {
```
jadair-cloudera: Add description metadata?

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
```diff
@@ -0,0 +1,179 @@
...
+                "GP_Gen5_16",
+                "GP_Gen5_32",
+                "MO_Gen5_2",
+                "MO_Gen5_4",
+                "MO_Gen5_8",
+                "MO_Gen5_16",
+                "MO_Gen5_32"
+            ],
+            "metadata": {
+                "description": "Azure database for PostgreSQL sku name "
```
jadair-cloudera: NIT: SKU

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
gergopapi2: Thanks, corrected.

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
gergopapi2: Yes, would be nice, I check it now. On the other hand, field "allowedValues" does a sanity check on the input string right away. The feedback comes before anything in azure would be provisioned.

gergopapi2: Ok, it was quite easy to do the boolean mapping, so fixed.

cloud-azure/src/main/resources/templates/arm-dbstack.ftl
gergopapi2: Yes, a bunch of descriptions were missing, thanks for pointing out that.

PR #5679, opened 2019-07-19
CB-2581 Fix typo in HBaseUI gateway proxy URL

joshelser:
Simple change to change the generated url from `?host=<host>?port=<port>` to `?host=<host>&port=<port>`. The initial page will load from this URL, but subsequent traversals on the proxied UI will break.


keyki:
Oh. Thank you for the fix. I'll change the base version to 2.11 so it can be rolled out to dev immediately.

keyki:
Actually, I can't simply change it. Could you please send the PR against rc-2.11?

joshelser:
> Actually, I can't simply change it. Could you please send the PR against rc-2.11?

Done!

PR #5680, opened 2019-07-20
Logging: monitor KNOX service logs (gateway)

oleewere:
add knox gateway logs for td-agent config

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/fluent/template/td-agent.conf.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/fluent/template/td-agent.conf.j2
@@ -187,6 +187,38 @@
 <source>
   @type tail
   format none
+  path /var/log/knox/gateway.log*
+  pos_file /var/log/td-agent/pos/knox-gateway.log.pos
+  tag {{fluent.providerPrefix}}.KNOX-GATEWAY
+  read_from_head true
+</source>
+<source>
+  @type tail
+  format none
+  path /var/log/knox/gateway-audit.log*
+  pos_file /var/log/td-agent/pos/knox-gateway-audit.log.pos
+  tag {{fluent.providerPrefix}}.KNOX-GATEWAY_AUDIT
+  read_from_head true
+</source>
+<source>
+  @type tail
+  format none
+  path /var/log/knox/idbroker/gateway.log*
+  pos_file /var/log/td-agent/pos/knox-idbroker-gateway.log.pos
+  tag {{fluent.providerPrefix}}.KNOX-IDBROKER_GATEWAY
+  read_from_head true
+</source>
+<source>
+  @type tail
+  format none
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/fluent/template/td-agent.conf.j2
```diff
@@ -184,6 +184,22 @@
   tag {{fluent.providerPrefix}}.HDFS-SECONDARYNAMENODE
   read_from_head true
 </source>
+<source>
+  @type tail
+  format none
+  path /var/log/knox/gateway.log*
```
lajosrodek: Not sure you were going to cover it elsewhere, but we should also have monitoring for IDBroker logs.

oleewere: @lajosrodek , is it be on the machine where you have knox gateway? (can be interesting, maybe we won't use the same instance profile for idbroker machines, that can be problematic for s3 log support).. do you know the path of those logs? then i can include those as well

lajosrodek: No, IDBroker is always deployed to a separate instance (or instances, in case of HA). See `"refName": "idbroker"` in the [non-HA](https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/defaults/blueprints/cdp-sdx.bp) and [medium-HA](https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/defaults/blueprints/cdp-sdx-medium-ha.bp) BPs.

The IDBroker identity (EC2 instance profile or MSI) will always contain the baseline one, so whatever grants are needed for fluentd logging etc. will be also applied to IDBroker nodes: see [this doc section](https://docs.google.com/document/d/1DDUpWOIaKq-lbmTEyUoYnK8NIfrlRrGN54DWIms4S4E/edit?ts=5cef1597#heading=h.st7ehb8q9pnm).

IDBroker logs are found here:
```
[root@ip-10-65-64-185 ~]# ls -l /var/log/knox/idbroker/
total 64
-rw-r--r-- 1 knox knox  1480 Jul 18 12:45 gateway-audit.log
-rw-r--r-- 1 knox knox 41028 Jul 18 12:45 gateway.log
-rw-r--r-- 1 knox knox 16087 Jul 18 12:26 knoxcli.log
drwxr-xr-x 2 knox knox     6 Jul 18 12:26 stacks
[root@ip-10-65-64-185 ~]# ls -l /var/log/knox/idbroker/stacks/
total 0
```

oleewere: ok, thanks. overall it won't hurt if i include these as well

PR #5686, opened 2019-07-22
CB-2339: upate password policy

tiborpopovics:


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/PasswordPolicyService.java
```diff
@@ -0,0 +1,77 @@
...
+import com.sequenceiq.freeipa.client.model.PasswordPolicy;
+
+@Service
+class PasswordPolicyService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PasswordPolicyService.class);
+
+    private static final String MIN_PASSWORD_LENGTH = "krbpwdminlength";
+
+    private static final Integer MIN_PASSWORD_LENGTH_VALUE = 8;
```
keyki: would be better if we make these values configurable @Value.. with the defaults

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/PasswordPolicyService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    void updatePasswordPolicy(FreeIpaClient freeIpaClient) {
+        try {
+            PasswordPolicy config = freeIpaClient.getPasswordPolicy();
+            Map<String, Object> params = getParameters(config);
+            if (!params.isEmpty()) {
+                freeIpaClient.updatePasswordPolicy(params);
+            }
+        } catch (FreeIpaClientException e) {
+            LOGGER.error("Failed to create policy.", e);
```
lacikaaa: we should throw the error here to make freeipa provision fail

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/PasswordPolicyService.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String MAX_PASSWORD_LIFETIME_DAYS = "krbmaxpwdlife";
+
+    private static final String MAX_FAILURES_BEFORE_LOCK = "krbpwdmaxfailure";
+
+    private static final String FAILURE_RESET_INTERVALL_SECONDS = "krbpwdfailurecountinterval";
+
+    private static final String LOCKOUT_DURATION_SECONDS = "krbpwdlockoutduration";
+
+    @Value("passwordpolicy.min-password-lenght")
```
keyki: does this work? Shouldn't it be freeipa.passwordpolicy..?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/FreeIpaPasswordValidator.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.freeipa.controller.exception.BadRequestException;
+
+@Component
+class FreeIpaPasswordValidator {
+
+    private static final Set<String> PATTERNS = Set.of("(?=.*[a-z]).*$", "(?=.*[A-Z]).*$", "(?=.*[0-9]).*$", ".*\\W+.*");
```
abhishekaypurohit: This should match with : https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/api/environments2/src/main/java/com/cloudera/thunderhead/service/environments2api/server/PasswordValidator.java#L16

keyki: The policy is based on JIRA https://jira.cloudera.com/browse/CB-2339. Can you please update the jira so we can make the proper change here.

PR #5693, opened 2019-07-22
CB-2418 Make more redbeams db server parameters optional

bhavanki:
The following parameters for redbeams database server allocation are
now fully optional: allocated storage, backup retention period, engine
version. The server port also remains optional, as before.

Additionally, a type error in the cloud model's DatabaseServer object
is fixed, so that null allocated storage sizes may pass fully through
the necessary data conversions.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
@@ -20,6 +20,7 @@ import com.amazonaws.services.ec2.AmazonEC2Client;
 import com.amazonaws.services.ec2.model.DescribeImagesRequest;
 import com.amazonaws.services.ec2.model.DescribeImagesResult;
 import com.amazonaws.services.ec2.model.Image;
+import com.google.common.annotations.VisibleForTesting;
 import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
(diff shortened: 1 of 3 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
```diff
@@ -117,33 +118,34 @@ private String getRootDeviceName(AuthenticatedContext ac, CloudStack cloudStack)
...
-                new Parameter().withParameterKey("StackOwner").withParameterValue(String.valueOf(ac.getCloudContext().getUserName()))
+                new Parameter().withParameterKey("StackOwner").withParameterValue(ac.getCloudContext().getUserName())
         ));
 
-        // Add the port if it's been supplied
-        Integer port = stack.getDatabaseServer().getPort();
-        if (port != null) {
-            parameters.add(new Parameter().withParameterKey("PortParameter").withParameterValue(Objects.toString(port)));
-        }
+        addParameterIfNotNull(parameters, "AllocatedStorageParameter", awsRdsInstanceView.getAllocatedStorage());
```
gergopapi2: Just a question: do all the values made optional here have a default value in aws?

bhavanki: They have default values in the CloudFormation template, actually. I think allocated storage is required by the RDS API, but the CF template will always supply something.

PR #5702, opened 2019-07-23
CB-2569 Redbeams api client without UserCrnClientRequestFilter

gergopapi2:


bhavanki:
After talking with @gergopapi2 we decided that, instead of _adding_ a second set of client code that uses a different mechanism to get the user CRN, we will _replace_ the existing client code with the code proposed here. The code here should work for any caller, and we still have only one set of client code.

lnardai:
This needs to go to rc-2.11 not master, as we want to release it as soon as possible.

PR #5707, opened 2019-07-24
DISTX-41 Implementing business metrics collection for Datahub and Dat…

cegganesh84:
…alake clusters

1. The usage event schema is defined in thunderhead repo.
2. The schema protobuf file is fetched from git in gradle and compiled using java.
3. The usage reporter interface and implementation will be used in appropriate place in this repo.
4. Build and checked that it works fine.
5. This is my first PR to this repo, will add the instrumentation after this PR is merged.
6. Add necessary steps to make sure that gradle builds properly in all environments.


keyki:
I have reverted the code now. Can you please copy the corresponding proto files rather than downloading them on-the-fly?

cegganesh84:
Hi @keyki,

Thanks for the merge and revert. Could you please give a little more color on the revert part. Is the build runs in a place where it does not have access to GitHub? Or dynamic fetching made the build unstable. 

I preferred to fetch the file instead of checking in because that way it is easier to make the changes. This file is going to be heavily edited as more use-cases are added. It is cumbersome to make changes in two places.

Thanks,
Ganesh

keyki:
Hi,

We copy the proto files here, because this is a public repository and ppl cannot build it outside of the VPN. Also I'm not sure if there are frequent changes in the proto file would break the builds randomly.

cegganesh84:
Hi @keyki,

Thanks for the reply. I agree about the public repository part. Also the idea is to use a fixed hash to fetch the file. If there is a cloudbreak related change - then change the hash and then use that new changes in the code. So it will not break builds randomly and still allows us to move to new version.

Thanks,
Ganesh

PR #5709, opened 2019-07-24
CB-2443 Introduce role based auth for Environment / FreeIPA / REDBEAMS / SDX / DISTROX services

horadla23:
- UMS related change: https://review.infra.cloudera.com/r/104348/
- tested with auth-mock (it returns true in case of every checkRight call), created an environment with freeipa and sdx, called a redbeams API endpoint successfully from browser
- deleted duplicated redbeams authz

**new module structure for authorization** (lot of refactor)
- auth-connector (not changed)
  - authorization-common (new)
    - workspace-authorization (includes new workspace-authorization-api) -> used by Cloudbreak (DistroX)
    - authorization (new) -> used by SDX, FreeIPA, Redbeams, Environment


authorization-common/src/main/java/com/sequenceiq/authorization/resource/RightUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+public class RightUtils {
+
+    public static final String DISTROX_RIGHT_PREFIX = "datahub/";
+
+    static final List<AuthorizationResource> NON_DATAHUB_RESOURCE =
```
lacikaaa: nevezd at!!!444negynegy

gregito:
fart

authorization-common/src/main/java/com/sequenceiq/authorization/resource/RightUtils.java
```diff
@@ -0,0 +1,23 @@
...
+            ImmutableList.of(AuthorizationResource.ENVIRONMENT, AuthorizationResource.FREEIPA, AuthorizationResource.DATALAKE);
+
+    private RightUtils() {
+    }
+
+    public static String getRight(AuthorizationResource resource, ResourceAction action) {
+        if (!NON_DATAHUB_RESOURCE.contains(resource)) {
+            return DISTROX_RIGHT_PREFIX + action.name().toLowerCase() + resource.getAuthorizationName();
```
lacikaaa: can we replace this ambiguous `name().lowerCase()` to a method on the enum so it would be more self describing and errorprune. thanks

authorization-common/src/main/java/com/sequenceiq/authorization/aspect/CheckPermissionsAspects.java
```diff
@@ -19,9 +19,9 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(CheckPermissionsAspects.class);
 
     @Inject
-    private PermissionCheckerService permissionCheckerService;
+    private AbstractPermissionCheckerService permissionCheckerService;
 
-    @Pointcut("execution(public * com.sequenceiq.cloudbreak.repository..*.*(..)) ")
+    @Pointcut("execution(public * com.sequenceiq..repository..*.*(..)) ")
```
attilapalfi92: As we agreed at your desk, ples make this aspect to look for subclasses of BaseWhateverRepository that we have to extend in our repos anyway, so we won't have package restrictions.

attilapalfi92: (and place back the moved repos)

PR #5719, opened 2019-07-24
CB-2148  SDX should create RBDMS instance and Databases

bbihari:


datalake/src/main/java/com/sequenceiq/datalake/flow/create/handler/RdsWaitHandler.java
```diff
@@ -0,0 +1,110 @@
...
+            LOGGER.info("Database polling failed for sdx: {}", sdxId, exception);
+            response = new SdxCreateFailedEvent(sdxId, userId, exception);
+        } catch (Exception anotherException) {
+            LOGGER.error("Something wrong happened in sdx database creation wait phase", anotherException);
+            response = new SdxCreateFailedEvent(sdxId, userId, anotherException);
+        }
+        sendEvent(response, event);
+    }
+
+    private void validForDatabaseCreation(Long sdxId, DetailedEnvironmentResponse env) {
```
sodre90: please move validation before flow start and validate it when request received

bbihari: For the validation we need the environment response and that is only available in the flow

sodre90: We have environment response here:
https://github.com/hortonworks/cloudbreak/blob/0a4b4d41adce1792d354a8d72d23aec8132027be/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java#L129

PR #5726, opened 2019-07-24
CB-2684. Improve Swagger docs for redbeams API

bhavanki:
Swagger configuration in redbeams is split out from general endpoint
configuration. The base path for the configuration is fixed so that it
includes the "/redbeams" context path for the service as a whole.

The API now documents the need for the x-cdp-actor-crn header for
authorization. Not only is this simply more accurate, but it causes the
Swagger UI to offer the option to set the header. Along with the
corrected base path, this makes the UI usable for performing redbeams
calls from a browser.

Unfortunately, the Swagger API documentation available at /v2/api-docs
is not working. My guess is that it's because the API controllers use
JAX-RS, and not Spring, for their annotations. This issue is already
overcome for other Swagger artifacts.


bhavanki:
I talked with @gergopapi2 separately - the path for it is http://localhost:8087/redbeams/swagger-ui.html without including "/api". I only figured this out through guessing.

PR #5731, opened 2019-07-25
CB-2643 Ensure that UMS cached properly

akanto:


redbeams/src/main/java/com/sequenceiq/redbeams/authorization/PermissionCheckingUtils.java
```diff
@@ -56,7 +55,7 @@ public void checkPermissionsByTarget(Object target, String userCrn, ResourceActi
             LOGGER.info("Checking permissions on " + resourceCrn + " for user " + userCrn);
             String requestId = threadBasedRequestIdProvider.getRequestId();
             LOGGER.debug("- tracking with request ID {}", requestId);
-            if (!grpcUmsClient.checkRight(userCrn, action.name(), resourceCrn, requestId)) {
+            if (!grpcUmsClient.checkRight(userCrn, userCrn, action.name(), resourceCrn, Optional.of(requestId))) {
```
keyki: We're passing the userCrn 2 times.

akanto: Actor and the User is the same

PR #5738, opened 2019-07-25
CB-2668 Add SCM workaround settings for OPSAPS-51681

lajosrodek:
Has been verified with the following builds:

* CM: 1294992
* CDH: 1287733


lajosrodek:
@akanto Good point. I will launch a datalake with the default builds (and omitting cloud storage settings) and see how it goes. Will ping back with the result.

PR #5743, opened 2019-07-25
CB-2536 - new cloudStorageRequest in CB and distroX

attilapalfi92:
In CB and DistroX the CloudStorage request now looks like this, as we agreed with @akanto and @doktoric:

```
{
  "cluster": {
    "cloudStorage": {
      "aws": {
          "s3guard": {
             "dynamoTableName": "almafa"
           }
       },
      "locations": [
        {
          "type": "RANGER-AUDIT",
          "value": "s3a://cloudbreak-group/demo/cumu-whdir/core-services/ranger/audit"
        },
        {
          "type": "HIVE-WAREHOUSE",
          "value": "s3a://cloudbreak-group/demo/cumu-whdir/core-services/ranger/audit"
        }
      ],

      "identities": [
       {
         "type": "IDBROKER",
         "s3": {
           "instanceProfile": "arn:aws:iam::980678866538:instance-profile/akanto-assumer"
         }
       },
       {
         "type": "LOG",
         "s3": {
           "instanceProfile": "arn:aws:iam::980678866538:instance-profile/akanto-log-bucket-access"
         }
       }
     ]
    }
  }
}
```

The supported cloud storages are: S3 and WASB. Other cloud storage requests are denied with HTTP 400. The SPI part is not changed yet, and only 1 identity is supported now. 

Work we still need to do:
- adding support for multiple identities in SPI
- separate validation that is currently sawed into the converters
- due to lack of time there is no separate request and response objects for `storageLocation` and `storageIdentity`
- some unit tests are needed for some of the code here


common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/CloudStorageBase.java
```diff
@@ -0,0 +1,46 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import io.swagger.annotations.ApiModel;
+
+@ApiModel
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public abstract class CloudStorageBase implements Serializable {
+
+    private S3Guard s3Guard;
```
doktoric: this field is really aws specific. as I see other fields are not really. can we somehow use a base . class here as well?

common-model/src/main/java/com/sequenceiq/common/model/CloudStorageCdpService.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public enum CloudStorageCdpService {
+
+    ZEPPELIN_SERVER,
```
doktoric: I think are comming for a josn configuration. do we need this enum ?

attilapalfi92: - contra against this: the enum has to store the same values as the json, so we have to maintain it at 2 places
- pro: it is straightforward in the code that what kind of values can be accepted in the API and what we store in the database

I think the pro argument is stronger, what do you think?

common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/CloudStorageBase.java
attilapalfi92: I'll move it inside an AwsStorageParameters class and on the API it will be called aws.

common-model/src/main/java/com/sequenceiq/common/model/CloudStorageCdpService.java
attilapalfi92: Actually this enum is even used for parsing the JSON. This makes the contra argument weaker.

PR #5747, opened 2019-07-26
DISTX-41 Implementing business metrics collection for Datahub and Dat…

cegganesh84:
…alake clusters

1. Instrument the request creation and status change events for both the type of clusters.
2. Used cluster construct primarily. Traversed to Stack and User from the cluster object.
3. Happy to change the instrumentation approach based on feedback.


cegganesh84:
@keyki I could not check why integration-test failed. Is there a way to re-kick that specific build. It will be good to keep the history for one full day, as we have folks from other timezones contributing to this repo.

cegganesh84:
@keyki @doktoric @lnardai - I am done with all the changes and made sure all the pre-commits check pass. Please take a look at this request.

cegganesh84:
Thanks @keyki

PR #5750, opened 2019-07-26
CB-1447 S3Guard configuration managed at environment level

topolyai5:


environment/src/main/java/com/sequenceiq/environment/environment/domain/Environment.java
```diff
@@ -347,4 +351,12 @@ public Tunnel getTunnel() {
     public void setTunnel(Tunnel tunnel) {
         this.tunnel = tunnel;
     }
+
+    public Json getParameters() {
```
attilapalfi92: I don't like that we start adding jsons to the relational database when it's unavoidable. Instead add a pojo.

attilapalfi92: Suggestion 2: if you can't let jsons go for some reason, than at least implement it like we did with the `EnvironmentTelemetry` here. Only this class has to know that it has a json representation inside the DB, in the code we want to do json conversions.

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDtoConverter.java
```diff
@@ -45,13 +51,27 @@ public EnvironmentDto environmentToDto(Environment environment) {
...
+            builder.withAws(getParameterAs(environment, AwsEnvironmentParamsDto.class));
+        }
         if (environment.getNetwork() != null) {
             builder.withNetwork(environmentNetworkConverterMap.get(CloudPlatform.valueOf(environment.getCloudPlatform()))
                     .convertToDto(environment.getNetwork()));
         }
         return builder.build();
     }
 
+    private <T> T getParameterAs(Environment environment, Class<T> clss) {
```
attilapalfi92: It's not a very good solution to handle jsons, do it like we did with `EnvironmentTelemetry`.

environment/src/main/java/com/sequenceiq/environment/environment/domain/Environment.java
attilapalfi92: Suggestion 1 was this: https://stackoverflow.com/a/10434692/3757113

This is better than the json.

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDtoConverter.java
attilapalfi92: Or what's even better: don't use json at all, for this single field. It's overkill and unnecessary. Use the relational database like it's designed to be used.

environment/src/main/java/com/sequenceiq/environment/environment/domain/Environment.java
bergerdenes: reworked

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDtoConverter.java
bergerdenes: reworked

PR #5751, opened 2019-07-26
CB-2710 Redbeams security groups: redbeams should create one per db

gergopapi2:
Fixing checkstyles and tests.


cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
```diff
@@ -87,7 +97,42 @@
   },
 
   "Resources" : {
-
+  <#if hasSecurityGroup>
+  <#else>
```
bhavanki: <#if !hasSecurityGroup> should work, but switching to that isn't a show-stopper for me.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
```diff
@@ -87,7 +97,42 @@
...
+                      "CidrIp": "0.0.0.0/0",
+                      "IpProtocol": "-1"
+                    }
+              ],
+              "SecurityGroupIngress" : [
+                    {
+                      "IpProtocol" : "tcp",
+                      "FromPort": 5432,
+                      "ToPort" : 5432,
+                      "CidrIp" : "0.0.0.0/0"
```
bhavanki: This is too wide open - it will allow access from the internet. Instead, access should be restricted to every subnet in the VPC.

(Same for port 22, which probably isn't necessary anyway but shouldn't hurt.)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
```diff
@@ -144,6 +145,21 @@ private String getRootDeviceName(AuthenticatedContext ac, CloudStack cloudStack)
             parameters.add(new Parameter().withParameterKey("PortParameter").withParameterValue(Objects.toString(port)));
         }
 
+        if (awsRdsInstanceView.getVPCSecurityGroups().isEmpty()) {
+            parameters.addAll(
+                    asList(
+                            new Parameter().withParameterKey("VPCIdParameter").withParameterValue(String.valueOf(awsNetworkView.getExistingVPC())),
```
bhavanki: I recommend adding a comment here that the VPC is guaranteed to be present in the stack because of NetworkParameterFactoryService. That isn't obvious.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
gergopapi2: Ah, great, just found another part of the ftl manual.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
gergopapi2: Ok, the code was not quite ok, VpcId and cidr could be missing if network is provided but sec group not in the request => I now always add the vpc parameters.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
gergopapi2: Ok, I added the VPC cidr, as discussed. Later we should address if this is good enough (change env can change the network).

gergopapi2:
> Using 0.0.0.0/0 as the CIDR for inbound security group rules is a blocker for me.

That is fixed now.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkView.java
```diff
@@ -55,4 +58,8 @@ public String getExistingIGW() {
     public String getExistingVPC() {
         return network.getStringParameter(VPC);
     }
+
+    public String getExistingVpcCidr() {
```
jadair-cloudera: NIT: I think the casing of VPC should be consistent. Either VPC or Vpc everywhere. Right now this method and the one above it are inconsistent.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
```diff
@@ -87,7 +101,35 @@
...
+                    {
+                      "CidrIp": "0.0.0.0/0",
+                      "IpProtocol": "-1"
+                    }
+              ],
+              "SecurityGroupIngress" : [
+                    {
+                      "IpProtocol" : "tcp",
+                      "FromPort": 5432,
+                      "ToPort" : 5432,
```
jadair-cloudera: Are we always setting up the DB to listen on this port? If the port is configurable elsewhere, we should respect that configuration here.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
```diff
@@ -87,7 +101,35 @@
...
+              "SecurityGroupEgress" : [
+                    {
+                      "CidrIp": "0.0.0.0/0",
+                      "IpProtocol": "-1"
+                    }
+              ],
+              "SecurityGroupIngress" : [
+                    {
+                      "IpProtocol" : "tcp",
+                      "FromPort": 5432,
```
jadair-cloudera: Shouldn't the FromPort be unconstrained?

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
gergopapi2: That's a great point. Port can be specified from the outside, I modify the template. Thanks!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkView.java
gergopapi2: Ok, fixed. Thanks.

PR #5754, opened 2019-07-26
DISTX-285. CB: support databus output for service logs and metering by Salt

oleewere:
Added dbus support on salt side: (no changes yet on java side, the telemetry usage will be refactored a bit because of https://github.com/hortonworks/cloudbreak/pull/5743)

# What is added here?
- extend some of the common values to variables in settings.sls
- in settings.sls, there are some handling of missing boolean vars, that is required because no java code change yet, that will be removed later
- td-agent.conf contents moved to input.conf
- added some missing logs to monitor: ZEPPELIN, salt master/minion/api and saltbootrap.log
- added dbus based report bundle logs, the flow for those logs are: service logs on nodes -> databus-api endpont -> logconsumer on CP -> kibana , BUT!!! as we are not owning those logs, it cannot run forever, so we are processing those logs for only 20 minutes (e.g.: to get results about cluster install failures), but fluentd is a daemon processes, so i have added a delayed restart script which executed in the background (right after fluentd started we are reconfiguring it ...sleep 20 mins then restart with the new configs) -> that is managed on the node, not from CB (as we loose any connection between these, logs will be processed to CP until things are not fixed), that's why we had the script (+ a lock to make sure we have already used dbus processing ...with deleting the lock and reapply fluent.init can re-trigger dbus log processing).
using a service restart is ok in the script as td-agent package also contains init.d script which uses systemctl when it's available.
- for the above example, i have provided a property called `dbusReportBundleDisableStop `, that is used to not stop processing logs with databus...that can be useful for QE or Dev to see service logs in kibana
- added metering input/output configs as well (that is also dbus based, but won't disappear after the delayed restart), metering events will be followed by a json file, so we will need only process that (event generator is not yet implemented but that is in progress)


orchestrator-salt/src/main/resources/salt/salt/fluent/init.sls
```diff
@@ -83,21 +84,110 @@ install_fluentd_plugins:
       - group
       - mode
 
-/etc/td-agent/td-agent.conf:
+{%- if fluent.is_systemd %}
+  service.dead:
```
lajosrodek: The file/state name is missing here, maybe`fluent_stop:` can be also used here?

PR #5761, opened 2019-07-26
CB-2684 Further improve redbeams API documentation

bhavanki:
Swagger descriptions have been defined for all redbeams endpoints,
covering each operation, its parameters, and model objects that are
not part of database stacks. (Those will come later.)

Controller descriptions have been removed, since Swagger no longer
uses them. Also, the empty DatabaseV4Controller::create methods has
been eliminated.

None of the changes affect how the API works, but only its
documentation.


redbeams-api/src/main/java/com/sequenceiq/redbeams/doc/ModelDescriptions.java
```diff
@@ -2,58 +2,79 @@
 
 public final class ModelDescriptions {
 
-    public static final String ID = "ID of the resource";
-    public static final String CRN = "CRN of the resource";
-    public static final String DESCRIPTION = "Description of the resource";
-    public static final String CREATION_DATE = "Creation date / time of the resource, in epoch milliseconds";
-    public static final String ENVIRONMENT_CRN = "Crn of the environment of the resource";
+    public static final String ALLOCATE_DATABASE_SERVER_REQUEST = "Request to allocate a new database server in a provider";
```
jadair-cloudera: NIT: This one says "Request to..." and the next one says, "Request for..." Please make consistent.

bhavanki: Will change in the next PR!

bhavanki: Actually, since a rebase is necessary, I fixed this now.

PR #5763, opened 2019-07-27
CB-2763 If multiple images were burnt on the same date, we still shou…

akanto:


doktoric:
@akanto pls check the unit tests

PR #5767, opened 2019-07-29
CB-2769 Redbeams should get default security groups from env

gergopapi2:


redbeams/src/main/java/com/sequenceiq/redbeams/converter/stack/AllocateDatabaseServerV4RequestToDBStackConverter.java
```diff
@@ -196,13 +198,17 @@ private DatabaseServer buildDatabaseServer(DatabaseServerV4Request source, Strin
         return server;
     }
 
-    private SecurityGroup buildSecurityGroup(SecurityGroupV4Request source) {
+    private SecurityGroup buildSecurityGroup(SecurityGroupV4Request source, SecurityAccessResponse securityAccessResponse) {
         SecurityGroup securityGroup = new SecurityGroup();
-        if (source == null) {
+        if (source != null) {
+            securityGroup.setSecurityGroupIds(source.getSecurityGroupIds());
             return securityGroup;
```
jadair-cloudera: I would prefer a single exit point from this method. Instead of doing a return here, could you just change the following if statement to an else if?

gergopapi2: Ok, I changed it. I also added a test.

PR #5775, opened 2019-07-29
CB-2764 DistroX UI is frozen on Mow

gregito:
CB-2764 DistroX UI is frozen on Mow


doktoric:
We should remove the current clustertemplates because we need to reregister those as well

gregito:
> We should remove the current clustertemplates because we need to reregister those as well

gonna do asap

doktoric:
We should remove the current clustertemplates because we need to reregister those as well

> > We should remove the current clustertemplates because we need to reregister those as well
> 
> gonna do asap

already added

PR #5782, opened 2019-07-30
CB-2687 Remove connector JAR URL from redbeams

bhavanki:
It is no longer possible to specify a URL for a JDBC driver JAR, as
doing so opens up a security hole where a user-provided custom driver
could perform local code execution.

The implication is that redbeams must find a JDBC driver itself. This is
already built-in for PostgreSQL, but there will need to be changes later
to support other database vendors, such as MySQL.


bhavanki:
I need to update this with a migration that removes the connector JAR URL from the database.

PR #5785, opened 2019-07-30
CB-1571: configure cm mgmt sercices to use external volume

tiborpopovics:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerMgmtSetupService.java
```diff
@@ -104,6 +107,7 @@ public void setupMgmtServices(Stack stack, ApiClient client, ApiHostRef cmHostRe
                 mgmtRoles.addItemsItem(apiRole);
             }
         }
+        cmMgmtServiceConfigLocationService.setConfigLocations(stack, mgmtRoles);
```
daszabo: Please check whether these configurations can be sent in with the initial setup request. See Line 93.

daszabo: Nvm, the telemetry was done in the same way as you did.

PR #5786, opened 2019-07-30
CB-1677. Add in Azure database allocation.

meowfaceman:
Azure database allocation has been added to Redbeams. This includes some
slight reworking of the allocation request from the Redbeams API, as
well as adding in Azure specific API requests.

Note: I ended up keeping with the Azure pattern of using defaultValues/interpolation to push values into the freemarker template. Feel free to push back if you'd rather I did this a different way, but I have a feeling it's done this way for a reason. Happy to be wrong about that, though.

### DIFF
--- a/redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/stacks/azure/AzureDatabaseServerV4Parameters.java
+++ b/redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/stacks/azure/AzureDatabaseServerV4Parameters.java
@@ -1,0 +1,142 @@
+package com.sequenceiq.redbeams.api.endpoint.v4.stacks.azure;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
+import com.sequenceiq.cloudbreak.common.mappable.MappableBase;
+import com.sequenceiq.redbeams.doc.ModelDescriptions.AzureDatabaseServerModelDescriptions;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.Map;
+
+@ApiModel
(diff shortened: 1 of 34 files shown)


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/stacks/azure/AzureDatabaseServerV4Parameters.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    @JsonIgnore
+    @ApiModelProperty(hidden = true)
+    public CloudPlatform getCloudPlatform() {
+        return CloudPlatform.AZURE;
+    }
+
+    @Override
+    public void parse(Map<String, Object> parameters) {
+        backupRetentionDays = getInt(parameters, "backupRetentionDays");
```
bhavanki: The string keys here should be constants

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/stacks/azure/AzureDatabaseServerV4Parameters.java
```diff
@@ -0,0 +1,135 @@
...
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.Map;
+
+@ApiModel
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
+public class AzureDatabaseServerV4Parameters extends MappableBase {
+
+    @ApiModelProperty
```
bhavanki: Please add descriptions for all of these properties to ModelDescriptions

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/stacks/azure/AzureNetworkV4Parameters.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public void setVirtualNetwork(String virtualNetwork) {
+        this.virtualNetwork = virtualNetwork;
+    }
+
+    @Override
+    public Map<String, Object> asMap() {
+        Map<String, Object> map = super.asMap();
+        putIfValueNotNull(map, "virtualNetwork", virtualNetwork);
```
bhavanki: Same as above, "virtualNetwork" should be a constant.

PR #5788, opened 2019-07-30
CB-2707 add Nifi to exposed services in Knox

keyki:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
```diff
@@ -31,7 +31,8 @@
     OOZIE_UI("Oozie Server", "OOZIE_SERVER", "OOZIE", "/oozie/", true, 11000, 11443, false, true),
     SOLR("Solr Server", "SOLR_SERVER", "SOLR", "/solr/", true, 8983, 8985, false, true),
     HBASE_UI("HBase UI", "MASTER", "HBASEUI", "/hbase/webui/master", true, 16010, 16010, false, false),
-    HBASE_REST("HBase Rest", "HBASERESTSERVER", "WEBHBASE", "/hbase/", true, 20550, 20550, true, true);
+    HBASE_REST("HBase Rest", "HBASERESTSERVER", "WEBHBASE", "/hbase/", true, 20550, 20550, true, true),
+    NIFI("Nifi", "NIFI_NODE", "NIFI", "/nifi-app/nifi", true, 8080, 8443, false, false);
```
risdenk: Port 8443 is potentially very problematic here - Knox gateway runs on port 8443. Need to make sure if we enable TLS in NiFi we don't run into issues.

Don't think that this blocks merging this just making sure we are aware.

keyki: Right, it's not going to work as long as 8443 is the default port for Nifi. Should we upgrade the config for Nifi or just leave it as it is for now?

risdenk: I think leave it as is and cross that bridge when we get there. There will be other changes required to make NiFi work with autotls.

PR #5795, opened 2019-07-31
Propagate cloud storage configuration to DistroX Cluster creation

biharitomi:
Changes in DistroX:
 - propagate Aws S3 Guard configs
 - propagate telemetry related cloud storage configs
 - allow only cloud storage storage configs on DistroX cluster's request and propagate it

Changes in Cloudbreak:
 - fix FilesystemValidator to allow cloud storages with only storage locations or identities - fix validation


common/src/main/java/com/sequenceiq/cloudbreak/util/NullUtil.java
```diff
@@ -28,4 +29,11 @@ private NullUtil() {
         }
         return null;
     }
+
+    public static <T, R, U> U getIfNotNull(T firstArgument, R secondArgument, BiFunction<T, R, U> consumer) {
```
bergerdenes: just cosmetics: the template parameter R is used for the result type, the first 2 parameters' types are T and U respectively

common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/StorageLocationBase.java
```diff
@@ -1,24 +1,27 @@
...
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.sequenceiq.common.model.CloudStorageCdpService;
 
 import io.swagger.annotations.ApiModel;
 
 @ApiModel
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class StorageLocationBase {
+public class StorageLocationBase implements Serializable {
```
bergerdenes: why do you need Serializable?

core/src/test/java/com/sequenceiq/distrox/v1/distrox/converter/CloudStorageLocationsToCloudStorageConverterTest.java
```diff
@@ -0,0 +1,127 @@
+package com.sequenceiq.distrox.v1.distrox.converter;
+
+import static org.junit.Assert.assertEquals;
```
bergerdenes: please use Jupiter's `Assertions` (JUnit5 instead of JUnit4)

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/CloudStorageLocationsToCloudStorageConverter.java
```diff
@@ -0,0 +1,57 @@
...
+            identity.setType(CloudIdentityType.LOG);
+            identity.setS3(logging.getS3());
+            identity.setWasb(logging.getWasb());
+            storageRequest.setIdentities(Collections.singletonList(identity));
+        }
+        if (CollectionUtils.isNotEmpty(cloudStorageLocations)) {
+            storageRequest.setLocations(new ArrayList<>(cloudStorageLocations));
+            fieldUpdated = true;
+        }
+        if (environment.getAws() != null
```
bergerdenes: Alternative:
```
String dynamoDbTableName = Optional.ofNullable(environment.getAws())
  .map(AwsParams::getS3guard)
  .map(S3Guard::getDynamoDbTableName)
  .orElse(null);
if (StringUtils.isNotEmpty(dynamoDbTableName)) {
...
```

common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/StorageLocationBase.java
biharitomi: It's required by Spotbugs, there is a rule that every REST response should implement it.

PR #5798, opened 2019-08-01
CDPD-1220  Add initial working DataMart template

drorke:
Making several updates to DataMart blueprint and template:
* Removing unnecessary services including Hive, YARN, and Oozie
* Tuning of specific Impala configs for performance
* Separation of the Impala deamon role config group into separate
  executor and coordinator types with distinct host templates
* Updating the DataMart cluster template to specify preferred defaults
  for instance type and disk configuration
* Removing the HA versions of the DataMart blueprint and cluster template


drorke:
> Please update the blueprint table in Cloudbreak database if you delete default (shipped) blueprints, like the HA ones. For reference check Cloudbreak source `core/src/main/resources/schema/app/20190520161304_CB-1334_remove_old_templates_without_hue.sql` or any similar changes how it was solved.

Added the SQL file to update the blueprint table and also removed a couple other references to the the HA blueprint and cluster template.

PR #5799, opened 2019-08-01
CB-2207 added storage config for logs

doktoric:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudStorageManifester.java
```diff
@@ -60,23 +65,63 @@ private void setStorageLocations(FileSystemParameterV4Responses fileSystemRecomm
...
+                S3Guard s3Guard = new S3Guard();
+                s3Guard.setDynamoTableName(environment.getAws().getS3guard().getDynamoDbTableName());
+                awsStorageParameters.setS3Guard(s3Guard);
+                cloudStorageRequest.setAws(awsStorageParameters);
+            }
+        }
+    }
+
+    private boolean isS3GuardConfigured(DetailedEnvironmentResponse environment) {
+        return environment.getAws() != null && environment.getAws().getS3guard() != null;
```
attilapalfi92: I think one more check could be useful here: `&& StringUtils.isNotEmpty(environment.getAws().getS3guard(). getDynamoDbTableName ())`

In case for some reason there is an S3Guard object with a null or empty dynamo table name. But I'm not sure.

doktoric: Oh I thought if s3guard is presented then dynametable is required

doktoric: anyway I think this what you searching https://github.com/hortonworks/cloudbreak/pull/5799/files#diff-932cffad5e6885b23e7012f1ca3a3085R70

PR #5800, opened 2019-08-01
CB-1588. Remove core API exceptions from Redbeams.

meowfaceman:
Redbeams API exceptions are now using internal Redbeams exceptions as
opposed to depending on core-api versions of the same exceptions.

Note: Manual testing showed the exception mapping to work as expected.


redbeams/src/main/java/com/sequenceiq/redbeams/exception/BadRequestException.java
```diff
@@ -0,0 +1,30 @@
...
+    public BadRequestException(String message) {
+        super(message);
+    }
+
+    public BadRequestException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
bhavanki: I wonder if this class needs equals and hashCode methods. Now is a good time to check if you are inclined to.

PR #5802, opened 2019-08-01
OPSAPS-51269 Hive Needs to be in HTTP Mode for Knox to Proxy J/ODBC Access

miklosgergely:
Added some parameters to the cdp-data-engineering.bp blueprint in order to integrate with knox.

Tested by using this modified blueprint in MOW, the parameters were present.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
@@ -148,7 +148,13 @@
           {
             "refName": "hive_on_tez-HIVESERVER2-BASE",
             "roleType": "HIVESERVER2",
-            "base": true
+            "base": true,
+            "configs": [
+              {
+                "name": "hive_hs2_config_safety_valve",
+                "value": "<property><name>hive.server2.thrift.http.port</name><value>10001</value></property><property><name>hive.server2.thrift.http.path</name><value>cliservice</value></property><property><name>hive.server2.transport.mode</name><value>http</value></property><property><name>hive.server2.allow.user.substitution</name><value>true</value></property><property><name>hive.server2.authentication.spnego.keytab</name><value>hive.keytab</value></property><property><name>hive.server2.authentication.spnego.principal</name><value>HTTP/_HOST@GCE.CLOUDERA.COM</value></property>"
+              }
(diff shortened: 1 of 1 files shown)


core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
```diff
@@ -148,7 +148,13 @@
           {
             "refName": "hive_on_tez-HIVESERVER2-BASE",
             "roleType": "HIVESERVER2",
-            "base": true
+            "base": true,
+            "configs": [
+              {
+                "name": "hive_hs2_config_safety_valve",
+                "value": "<property><name>hive.server2.thrift.http.port</name><value>10001</value></property><property><name>hive.server2.thrift.http.path</name><value>cliservice</value></property><property><name>hive.server2.transport.mode</name><value>http</value></property><property><name>hive.server2.allow.user.substitution</name><value>true</value></property><property><name>hive.server2.authentication.spnego.keytab</name><value>hive.keytab</value></property><property><name>hive.server2.authentication.spnego.principal</name><value>HTTP/_HOST@GCE.CLOUDERA.COM</value></property>"
```
adoroszlai: 1. `@GCE.CLOUDERA.COM` seems to be a hard-coded value, not sure if it's fit for production use.  Probably should be generated based on Kerberos config.

2. Most of the same values are being added to the config for Hive service by https://github.com/hortonworks/cloudbreak/blob/c6211e71b022336e07a861e5dcbdfc9931def30c/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java#L22-L25 I think a similar config provider is needed for Hive-on-Tez service.  Please try to reuse existing code.

keyki: I'm not sure this port would work for TLS enabled clusters.

keyki: Do we need to specify https port when TLS is enabled?

doktoric: I dont like spengo I preffer soska.

PR #5803, opened 2019-08-01
CB-2801: extend subnet range

tiborpopovics:


environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EnvironmentCreationValidator.java
```diff
@@ -55,6 +60,10 @@ private void validateNetwork(EnvironmentCreationDto request, String cloudPlatfor
         }
     }
 
+    private boolean isInvalidNetworkMask(String networkCidr) {
+        return !networkCidr.split("/")[1].equals(EXPECTED_NETWORK_MASK);
```
keyki: I think we should check if the cidr actually contains the subnet mask and it's not just an IP address.

tiborpopovics: Good idea. Please check now.

bergerdenes: there is a validator for this, take a look on that code ;)

tiborpopovics: I`ve updated wrong validation. Check it again please.

keyki: Can you please use the SubnetValidator?

PR #5804, opened 2019-08-01
 CB-2733,  CB-2688 Knox frontend URL to CM + trailing /

keyki:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
```diff
@@ -15,9 +15,9 @@
     ALL("Every Service", "ALL", "", "", true, null, null, false, true),
 
     CLOUDERA_MANAGER("CM-API", "CM-API", "CM-API", "/cm-api/", true, 7180, 7183, true, true),
-    CLOUDERA_MANAGER_UI("CM-UI", "CM-UI", "CM-UI", "/cmf/home", true, 7180, 7183, false, false),
+    CLOUDERA_MANAGER_UI("CM-UI", "CM-UI", "CM-UI", "/cmf/home/", true, 7180, 7183, false, false),
     HUE("HUE", "HUE_LOAD_BALANCER", "HUE", "/hue/", true, 8889, 8889, false, false),
-    WEBHDFS("WebHDFS", "NAMENODE", "WEBHDFS", "/webhdfs/v1", true, 9870, 9871, true, true),
+    WEBHDFS("WebHDFS", "NAMENODE", "WEBHDFS", "/webhdfs/v1/", true, 9870, 9871, true, true),
```
risdenk: if we are making a change to WEBHDFS: might want it to be `/webhdfs/v1/?OP=LISTSTATUS` which makes it a valid url in the UI. Otherwise if you click on it there is a failure.

I don't know what the long term way to display API endpoints, but sharing if trying to make a valid endpoint. There are other API calls in here that aren't valid if you click on them. So this recommendation can totally be ignored just wanted to share :)

core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
```diff
@@ -253,14 +253,14 @@ private String getHdfsUIUrlWithHostParameterFromGatewayTopology(String managerIp
...
-        String url = String.format("https://%s:%s/%s/%s%s?host=%s://%s:%s", managerIp, knoxPort, gateway.getPath(), gt.getTopologyName(),
+        String url = String.format("https://%s:%s/%s/%s%s?host=%s://%s:%s/", managerIp, knoxPort, gateway.getPath(), gt.getTopologyName(),
                 ExposedService.NAMENODE.getKnoxUrl(), protocol, nameNodePrivateIp, port);
         return url;
     }
 
     private String getHBaseUIUrlWithHostParameterFromGatewayTopology(String managerIp, GatewayTopology gt, String nameNodePrivateIp) {
         Gateway gateway = gt.getGateway();
-        String url = String.format("https://%s:%s/%s/%s%s?host=%s&port=%s", managerIp, knoxPort, gateway.getPath(), gt.getTopologyName(),
+        String url = String.format("https://%s:%s/%s/%s%s?host=%s&port=%s/", managerIp, knoxPort, gateway.getPath(), gt.getTopologyName(),
```
risdenk: This trailing slash isn't correct. Should be before the `?` 

I think the whole line should be:

`String url = String.format("https://%s:%s/%s/%s%s/?host=%s&port=%s", managerIp, knoxPort, gateway.getPath(), gt.getTopologyName(),`

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knox.settings.j2
```diff
@@ -2,4 +2,5 @@
 setsettings PROXYUSER_KNOX_GROUPS *
 setsettings PROXYUSER_KNOX_HOSTS *
 setsettings PROXYUSER_KNOX_PRINCIPAL knox
-setsettings PROXYUSER_KNOX_USERS *
\ No newline at end of file
+setsettings PROXYUSER_KNOX_USERS *
+setsettings FRONTEND_URL https://{{ knox_address }}:8443/
```
risdenk: I have NOT tested this with a deployment (not sure how), but would be good to check that the following two URLs work with CM and Knox:

* `https://KNOX_HOST:8443/CLUSTER_NAME/cdp-proxy/cmf/home`
* `https://KNOX_HOST:8443/cmf/home`

The first one in the list is what is generated in the CB gateway tab so want to make sure that still works. The second one is where CM is really rooted once any CM links are clicked on.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
```diff
@@ -27,12 +27,12 @@
     SPARK_HISTORY_SERVER("Spark History Server", "SPARK_YARN_HISTORY_SERVER", "SPARKHISTORYUI", "/sparkhistory/", true, 18088, 18488, false, false),
     ZEPPELIN("Zeppelin", "ZEPPELIN_SERVER", "ZEPPELIN", "/zeppelin/", false, 8885, 8886, false, false),
     RANGER("Ranger", "RANGER_ADMIN", "RANGER", "/ranger/", true, 6080, 6182, false, true),
-    LIVY_SERVER("Livy Server", "LIVY_SERVER", "LIVYSERVER1", "/livy/ui", true, 8998, 8998, false, true),
+    LIVY_SERVER("Livy Server", "LIVY_SERVER", "LIVYSERVER1", "/livy/ui/", true, 8998, 8998, false, true),
     OOZIE_UI("Oozie Server", "OOZIE_SERVER", "OOZIE", "/oozie/", true, 11000, 11443, false, true),
     SOLR("Solr Server", "SOLR_SERVER", "SOLR", "/solr/", true, 8983, 8985, false, true),
     HBASE_UI("HBase UI", "MASTER", "HBASEUI", "/hbase/webui/master", true, 16010, 16010, false, false),
```
risdenk: Not sure if this should be `/hbase/webui/master/`

core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
keyki: Right, silly me :) 👍

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
keyki: I could add that, but I guess the main purpose for the API endpoint is for programatic consumption and might add extra overhead to people if they have to remove the part from the URL.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
keyki: It worked for me, but it worked also without the trailing /. I just adjusted it as the JIRa described.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knox.settings.j2
keyki: As far as I know this property is not used in CM yet, but yes we can test if the second works also.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
keyki: Oh, I commented on the wrong line. Yes, so I can change it here and then I don't need to change in the String.format section above.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knox.settings.j2
risdenk: Let me see if I can track that change down in the next few min.

keyki:
@risdenk I updated the jira, can you please review again?

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knox.settings.j2
risdenk: Hmmm so yea the CM change isn't in yet. Sorry jumped the gun worrying about it.

PR #5813, opened 2019-08-02
CB-2695, CB-2766 : Initialize Impala Volume Configurations

smaniraju:
CB-2695 :  Expose volume configuration templates to role config providers.
CB-2766 :  Initialize Impala Volume Configurations.

Closes #CB-2695
Closes #CB-2766

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/impala/ImpalaVolumeConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/impala/ImpalaVolumeConfigProvider.java
@@ -1,0 +1,71 @@
+package com.sequenceiq.cloudbreak.cmtemplate.configproviders.impala;
+
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.config;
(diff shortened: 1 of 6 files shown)


smaniraju:
> We should handle the case where there are no attached volumes (only the root volume) as a special case. In this case we should still configure scratch_dirs to use the root volume but we should configure no data cache (datacache_enabled=false, 0 capacity, empty dirs string). Using half of the root volume for cache will consume too much space in the case of small root volumes - better to just disable the cache.
> 
> Sorry I didn't mention this in the original JIRA description.

@drorke  shall we enable cache only if min volume size is greater than 10gb ?

drorke:
> > We should handle the case where there are no attached volumes (only the root volume) as a special case. In this case we should still configure scratch_dirs to use the root volume but we should configure no data cache (datacache_enabled=false, 0 capacity, empty dirs string). Using half of the root volume for cache will consume too much space in the case of small root volumes - better to just disable the cache.
> > Sorry I didn't mention this in the original JIRA description.
> 
> @drorke I will disable cache if no volumes are attached. Also if volumes are attached should we enable cache only if attached volume size is greater than 10gb ?

For attached (non root) volumes the primary uses will be data cache and Impala scratch dirs (and potentially a bit of HDFS usage if someone were to configure Hive in addition to Impala).   I think it's OK to give up to half the attached non-root volume to the data cache even if the volume is small.   The scenario I have in mind for this is the demo scenario where someone might configure a very small volume when demoing to a customer but still want to show the cache feature.

PR #5814, opened 2019-08-02
DISTX-294. Support databus output for service logs and metering by server + DistroX API

oleewere:
## Changes:
- add back telemetry request to stack api
- for distrox api provide new boolean flag (for workload analytics - it's a requirement to be changeable ..but true by default, now it's a boolean flag, probably enum would be better but keep as this right now)
- for env api, new report deployment logs option (also global one on stack level ... so do not use it if not supported)
- add metering enable option -> global one, added to telemetry if it's set
- install new metering app through salt if metering available (currently a manual rpm install, location will change + probably we will have a repo option ... also won't cause issues if the install fails), also add metering config service
- extend fluent config service because of cluster deployment report + metering (now have 3 different logging), aslo add AltusIAMService which can generate + cleanup machine users for fluentd (move some CMDatabus code to UmsClient, as some of the parts now common)
- add instanceId and instanceType for node data (as those required for metering)
- user mock changes: TP and fluent databus credential requires different algorithm so for local testing, you can provide 2 different credentials instead of 1 (based on the algo).

note: these changes won't have any impact on cluster installs if the global flags are not set

## Testing
FT: distroX worked properly
In progress: re-test Env and SDX cluster creation


oleewere:
@attilapalfi92 
please check these files:
core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
environment/src/main/java/com/sequenceiq/environment/environment/v1/TelemetryApiConverter.java
core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/DistroXV1Base.java
core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/request/StackV4Request.java
core/src/main/java/com/sequenceiq/cloudbreak/service/altus/AltusIAMService.java
environment/src/main/java/com/sequenceiq/environment/environment/dto/telemetry/EnvironmentTelemetry.java

note: metering & reportLogs will be disabled

core/src/main/java/com/sequenceiq/cloudbreak/service/altus/AltusIAMService.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public AltusCredential generateDatabusMachineUserForFluent(Stack stack, Telemetry telemetry) {
+        if (isMeteringOrDeploymentReportingSupported(stack, telemetry)) {
+            return umsClient.createMachineUserAndGenerateKeys(
+                    getFluentDatabusMachineUserName(stack),
+                    stack.getCreator().getUserCrn(),
+                    umsClient.getBuiltInDatabusRoleCrn(),
+                    UserManagementProto.AccessKeyType.Value.ED25519);
+        }
+        return null;
```
attilapalfi92: Can this be an Optional.empty() instead?

orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
```diff
@@ -5,6 +5,10 @@
 
     private final String publicIp;
 
+    private final String instanceId;
```
oleewere: @keyki check this please

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/PillarSave.java
```diff
@@ -84,14 +85,17 @@ public PillarSave(SaltConnector sc, Set<String> targets, SaltPillarProperties pi
         originalTargets = targets;
     }
 
-    private Map<String, Object> discovery(String hostname, String publicAddress) {
+    private Map<String, Object> discovery(String hostname, String publicAddress,
```
oleewere: @keyki check this please

orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
keyki: LGTM.

orchestrator-salt/src/main/resources/salt/salt/nodes/settings.sls
```diff
@@ -1,9 +1,14 @@
 # https://tedops.github.io/how-to-find-default-active-ethernet-interface.html
-{% set has_public_address = salt['pillar.get']('hosts')[salt['network.interface_ip'](salt['network.default_route']('inet')[0]['interface'])]['public_address'] %}
 {% set private_address = salt['network.interface_ip'](salt['network.default_route']('inet')[0]['interface']) %}
+{% set host_details = salt['pillar.get']('hosts')[private_address] %}
```
oleewere: @keyki check this please

PR #5822, opened 2019-08-05
DISTX-296. Hue Load Balancer missing from DE blueprint

adoroszlai:
## What changes were proposed in this pull request?

Add Hue Load Balancer to the non-HA Data Engineering blueprint (and its LLAP version).  While there is only one Hue Server, Load Balancer is still required because Knox topology generator does not handle standalone Server (adb229cd86e8cad05bece04cf3ee36398df3703c).

## How was this patch tested?

Deployed data hub cluster with this blueprint.


adoroszlai:
Thanks @doktoric, @risdenk for reviewing it.

PR #5823, opened 2019-08-05
CB-2844. HA deployment fails due to invalid node counts

adoroszlai:
## What changes were proposed in this pull request?

Fix problems with HA cluster definitions:

 * correct instance count for `master` and `quorum`
 * make `gateway` the `GATEWAY` (the one with CM) instead of `master`, and change exposed ports accordingly


adoroszlai:
Moved to `rc-2.12` in #5832

PR #5835, opened 2019-08-05
CB-2863 Wire IDBMMS into Datalake Svc

lajosrodek:
Relies on the client established in #5608.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.inject.Inject;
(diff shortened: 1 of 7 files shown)


lajosrodek:
Merging blocked until IDBMMS can be brought up with cbd, otherwise all datalake creation flows would fail at the time of IDBMMS call due to the missing local service. On the other hand, the UI flow would be also completely blocked since currently it is not possible to create IDB mappings in the env creation UI flow.
The only path not affected is the CLI `sdx-internal create` that can specify a manual set of mappings, so that no IDBMMS call is required.

lajosrodek:
@attilapalfi92 Could you please re-review this? The following files have been updated:
* `GrpcIdbmmsClient.java`
* `IdBrokerMappingSource.java`
* `StackRequestManifester.java`
* `StackRequestManifesterTest.java`

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -219,4 +234,28 @@ private void setupClusterRequest(StackV4Request stackRequest) {
...
     }
+
+    void setupCloudStorageAccountMapping(StackV4Request stackRequest, String environmentCrn, IdBrokerMappingSource mappingSource, String cloudPlatform) {
+        CloudStorageRequest cloudStorage = stackRequest.getCluster().getCloudStorage();
+        if (cloudStorage != null && cloudStorage.getAccountMapping() == null) {
+            // In case of SdxClusterRequest with cloud storage, or SdxInternalClusterRequest with cloud storage but missing "accountMapping" property,
+            // getAccountMapping() == null means we need to fetch mappings from IDBMMS.
+            if (mappingSource == IdBrokerMappingSource.IDBMMS) {
+                LOGGER.info("Fetching account mappings from IDBMMS for environment {}", environmentCrn);
+                MappingsConfig mappingsConfig = idbmmsClient.getMappingsConfig(IAM_INTERNAL_ACTOR_CRN, environmentCrn, Optional.empty());
```
lnardai: I'm not sure if the use of Internal actor CRN usage here is correct. 
This setup is happenning in the context of some user. Internal CRN usage should be reserved for API calls that are initiated by services without user interaction. For e.g. Periscope periodically polling a Cloudbreak API.

lajosrodek: IDBMMS only serves requests to that specific method only for the internal actor, as it is meant to be an internal-only operation. (Request will always fail when using some other actor CRN.) It will not be made available to end users via public APIs (Environments2 REST, CDP CLI). Note that the public APIs do provide a `getMappings` operation that is indeed the user-facing variants, but whose response is inappropriate for us here.
In short, we must use `getMappingsConfig` and that requires the internal actor. 😃

lajosrodek: Comment has been added to justify the usage of the internal actor.

lajosrodek:
CLI change: https://github.com/hortonworks/cb-cli/pull/602

lajosrodek:
Log message improvements: #5924

PR #5839, opened 2019-08-06
Fix telemetry cleanup

oleewere:
Nullpointer can happen as during cluster cleanup, machine user deletion happens outside of telemetry context (as it was moved out in a more common module)


oleewere:
closing as it should go to rc-2.12

PR #5840, opened 2019-08-06
Fix telemetry cleanup

oleewere:
Nullpointer can happen as during cluster cleanup, machine user deletion happens outside of telemetry context (as it was moved out in a more common module)


oleewere:
@attilapalfi92 the return type is not related with the issue, error happens before that, i can do a follow up refactor around there

attilapalfi92:
I know it's not related but it is also error prone and you promised you will fix it. This is the time to keep your promise.

attilapalfi92:
Thank you @oleewere

PR #5844, opened 2019-08-06
CB-2858 Redbeams: make provisioning cancellable

gergopapi2:


doktoric:
@gergopapi2  I think this should be opened for 2.12

gergopapi2:
> @gergopapi2 I think this should be opened for 2.12

Thanks, I changed the base.

PR #5846, opened 2019-08-06
CB-2884 - Fixed NPE in FileSystemConfigurationsViewProvider

attilapalfi92:
Also fixing another issue in the same class: identity is no longer set in `BaseFileSystemConfigurationsView`


attilapalfi92:
@doktoric @akanto 
Is this legit that the instance profile is not necessary in the blueprints?

attilapalfi92:
It is legit according to @doktoric

PR #5848, opened 2019-08-06
CB-2878 - Hive Needs to be in HTTP Mode for Knox to Proxy J/ODBC Access

pdarvasi:


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java
```diff
@@ -36,14 +46,12 @@ public String getServiceType() {
 
     @Override
     public List<String> getRoleTypes() {
-        return List.of(HiveRoles.HIVE, HiveRoles.HIVEMETASTORE, HiveRoles.HIVESERVER2);
```
keyki: So we don't need this for metastore?

akanto:
To be honest I don't see where do you switch on the http mode. Anyway, if you enable http mode by default, then this will break the majority of QE tests (if not all).

pdarvasi:
Closing as it will be implemented in CM side: https://jira.cloudera.com/browse/OPSAPS-52015

PR #5851, opened 2019-08-06
CB-2452. Validate redbeams CRNs.

meowfaceman:
This commit adds in the ValidCrn annotation to all CRN instances in
redbeams.

Note: This has now been verified -- I was missing the @Validated annotation -- and CRNs are validated as expected.

### DIFF
--- a/redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
+++ b/redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
@@ -14,6 +14,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
+import com.sequenceiq.cloudbreak.validation.ValidCrn;
 import com.sequenceiq.redbeams.api.endpoint.v4.database.request.CreateDatabaseV4Request;
 import com.sequenceiq.redbeams.api.endpoint.v4.database.responses.CreateDatabaseV4Response;
 import com.sequenceiq.redbeams.api.endpoint.v4.databaseserver.requests.AllocateDatabaseServerV4Request;
@@ -32,10 +33,13 @@ import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 
+import org.springframework.validation.annotation.Validated;
+
 @Path("/v4/databaseservers")
 @Produces(MediaType.APPLICATION_JSON)
 @Api(tags = {"database servers"},
         protocols = "http,https")
+@Validated
 public interface DatabaseServerV4Endpoint {
 
     @GET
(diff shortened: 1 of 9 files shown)


redbeams/src/main/java/com/sequenceiq/redbeams/controller/mapper/CrnParseExceptionMapper.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.ws.rs.core.Response.Status;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class CrnParseExceptionMapper extends BaseExceptionMapper<CrnParseException> {
+
+    @Override
+    Status getResponseStatus() {
+        return Status.UNAUTHORIZED;
```
bhavanki: I feel like this should be a 400 instead of a 401.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -134,7 +142,7 @@ DatabaseServerV4Response deleteByName(
     @ApiOperation(value = DatabaseServerOpDescription.DELETE_MULTIPLE_BY_CRN, notes = DatabaseServerNotes.DELETE_MULTIPLE_BY_CRN,
             nickname = "deleteMultipleDatabaseServersByCrn")
     DatabaseServerV4Responses deleteMultiple(
-        @ApiParam(DatabaseServerParamDescriptions.CRNS) Set<String> crns
+        @ValidCrn @ApiParam(DatabaseServerParamDescriptions.CRNS) Set<String> crns
```
gergopapi2: A question: does this validator also work for Set<String>?

redbeams/src/main/java/com/sequenceiq/redbeams/controller/mapper/CrnParseExceptionMapper.java
gergopapi2: I also agree a bad request is better here.

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -134,7 +142,7 @@ DatabaseServerV4Response deleteByName(
...
     );
 
     @POST
```
gergopapi2: Following requests also contain crns:
- DatabaseServerTestV4Request: existingDatabaseServerCrn,
- CreateDatabaseV4Request: existingDatabaseServerCrn
- DatabaseV4Base: environmentCrn
-

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
meowfaceman: Thanks for bringing this up, I need to check this.

meowfaceman: Looks like if I move the annotation inside the Set it works, so:

```
Set<@ValidCrn String>
```

So that's what I've done.

PR #5853, opened 2019-08-07
CB-2781 Enforce the selection of 2 subnets at least at environment cr…

bbihari:
…eation, backend validation

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
@@ -1,14 +1,23 @@
 package com.sequenceiq.environment.environment.service;
 
+import static com.sequenceiq.environment.CloudPlatform.AWS;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 import javax.ws.rs.BadRequestException;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.cloud.model.CloudRegions;
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
 import com.sequenceiq.cloudbreak.util.ValidationResult;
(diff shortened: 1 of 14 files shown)


environment/src/main/java/com/sequenceiq/environment/network/NetworkService.java
```diff
@@ -60,12 +68,37 @@ public boolean hasExistingNetwork(BaseNetwork baseNetwork, CloudPlatform cloudPl
...
+            PlatformResourceRequest prr = new PlatformResourceRequest();
+            prr.setCredential(environment.getCredential());
+            prr.setCloudPlatform(environment.getCloudPlatform());
+            prr.setRegion(regionName);
+            getAwsVpcId(network)
+                    .ifPresent(vpcId -> prr.setFilters(Map.of("vpcId", vpcId)));
+
+            CloudNetworks cnsr = platformParameterService.getCloudNetworks(prr);
+            Set<CloudNetwork> cns = cnsr.getCloudNetworkResponses().get(regionName);
+            return cns.stream()
```
attilapalfi92: It's not easy to keep in mind what is `cns`, `cnsr`, `prr`. Pleaase expand the name of these variables so it becomes human-readable.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -65,14 +82,49 @@ public EnvironmentDto create(EnvironmentCreationDto creationDto, String accountI
...
         return environmentDtoConverter.environmentToDto(environment);
     }
 
+    private void validateNetworkRequest(Environment environment, NetworkDto network, Map<String, CloudSubnet> subnetMetas) {
+        String message;
+        if (AWS.name().equalsIgnoreCase(environment.getCloudPlatform()) && network != null && network.getNetworkCidr() == null) {
+            if (network.getSubnetIds().size() < 2) {
+                message = "Cannot create environment, there should be at least two subnets in the network";
+                LOGGER.info(message);
+                throw new BadRequestException(message);
```
attilapalfi92: These messages are thrown in their own separate exception respectfully, however these errors are not exclusive. More of these validation errors could occur in the same request but the user will only by notified about the first error. 

Please use the `com.sequenceiq.cloudbreak.util.ValidationResult` class to store multiple validation errors and notify the user about all of the errors.

bbihari: ok

PR #5854, opened 2019-08-07
CB-2903. Do not use booleans for Telemetry objects

oleewere:
use enum instead of true/false boolean input for setting workload Telemetry objects (both api and db as it was asked)

only `ENABLED` and `DISABLED` fields are acceptable in that case (no json aliases)

added db update scripts as well

### DIFF
--- a/common-model/src/main/java/com/sequenceiq/common/api/telemetry/model/Telemetry.java
+++ b/common-model/src/main/java/com/sequenceiq/common/api/telemetry/model/Telemetry.java
@@ -8,6 +8,7 @@ import java.io.Serializable;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.sequenceiq.common.api.telemetry.common.TelemetrySetting;
 
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 @JsonIgnoreProperties(ignoreUnknown = true)
@@ -22,11 +23,11 @@ public class Telemetry implements Serializable {
     @JsonProperty("databusEndpoint")
     private String databusEndpoint;
 
-    @JsonProperty("meteringEnabled")
-    private boolean meteringEnabled;
+    @JsonProperty("metering")
+    private TelemetrySetting metering;
 
     @JsonProperty("reportDeploymentLogs")
-    private boolean reportDeploymentLogs;
+    private TelemetrySetting reportDeploymentLogs;
 
     public Logging getLogging() {
         return logging;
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/altus/AltusIAMService.java
```diff
@@ -59,7 +60,9 @@ public void clearFluentMachineUser(Stack stack, Telemetry telemetry) {
 
     // for datalake metering is not supported/required right now
     private boolean isMeteringOrDeploymentReportingSupported(Stack stack, Telemetry telemetry) {
-        return telemetry != null && (telemetry.isReportDeploymentLogs() || (telemetry.isMeteringEnabled() && !StackType.DATALAKE.equals(stack.getType())));
+        return TelemetrySetting.ENABLED.equals(telemetry.getReportDeploymentLogs())
```
doktoric: telemetry != null

common-model/src/main/java/com/sequenceiq/common/api/telemetry/model/Telemetry.java
```diff
@@ -23,10 +24,10 @@
     private String databusEndpoint;
 
     @JsonProperty("meteringEnabled")
-    private boolean meteringEnabled;
+    private TelemetrySetting meteringEnabled;
 
     @JsonProperty("reportDeploymentLogs")
-    private boolean reportDeploymentLogs;
+    private TelemetrySetting reportDeploymentLogs;
```
doktoric: reportLogsEnable

core/src/main/java/com/sequenceiq/cloudbreak/service/altus/AltusIAMService.java
oleewere: it was checked before the method call

common-model/src/main/java/com/sequenceiq/common/api/telemetry/model/Telemetry.java
oleewere: i do not like the format then: `reportDeploymentLogsEnable: "enabled"`, then it should be reportDeploymentLogsEnable: "true", so `reportDeploymentLogs : "enabled"` more meaningful as we want to use enable(d) as a value

doktoric: daemonLogs?

oleewere: that feature will look like that: report service logs after cluster install/failure after deployment (for like 20 mins, but nothing after that), so just reportLogs or daemonLogs would not be correct

oleewere:
tomorrow, i will do a full cluster install with the rebased changes, please wait that, if it's successful it can be merged

daszabo:
@oleewere, was it successful?

oleewere:
@daszabo , technically none of my full cluster install finished successfully, but where the enums were used - that part went fine, i would wait with that change a bit anyway, i will have a better usage of these enums, which can be discussed tomorrow (got a bad feeling to change a boolean to ENABLED/DISABLED enum (as only those values are acceptable, and we need a null check anyway), but maybe would make more sense to use enums to wrap all the different configurations together and rename TelemetrySettings to TelemetryFeature which would contain the feature names...and we could pass the whole things in a list)
so i could imagine more like this (where enum makes more sense):
```json
"telemetry": 
  { "features":
      "metering",
      "workloadAnalytics",
      "reportDeploymentLogs"
  }
```
or (as an object instead of enum): 
```json
 { "features":
      "metering": false,
      "workloadAnalytics": true,
      "reportDeploymentLogs": false
  }
```

keyki:
I don't think this should go to rc-2.12 at this point

oleewere:
@keyki ok its not adding any functionality anyway. Fine for only master

oleewere:
dropping this PR - see: CB-3210

PR #5859, opened 2019-08-07
CB-2905 Redbeams api swagger conflicts with cli

gergopapi2:


bhavanki:
The purpose of the header is to allow the Swagger UI console to be usable. Without it, the console doesn't let you provide the actor CRN, so you can't use it to make calls. I would prefer that we find another solution.

PR #5864, opened 2019-08-07
DISTX-212. Cleanup inconsistencies in FMS user sync

handavid:
This commit makes the various user sync apis behave more similarly.

The environment crn filter was pulled out into a base class for all
the requests (set password, sync all, single user sync). The User
and Group models were removed from freeipa-api since they are now
only used internally to the FMS. They have also been renamed to
FmsUser and FmsGroup to make it easier to distinguish from UMS
users/groups and IPA users/groups.

All APIs can handle machine users as well as users. Sync code paths
were consolidated internally to use a single method for entry to the
UserService. Parameters are assumed (and validated) to be non-null
to simplify the code and the user controller was updated to adhere
to this change.

Retrieving stacks matching an environment crn filter has been moved
into the StackRepository instead of filtering a list of all stacks.

Logging is improved by adding toString to API models and MDC is
set.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
@@ -6,12 +6,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.logger.MDCBuilder;
 import com.sequenceiq.flow.event.EventSelectorUtil;
 import com.sequenceiq.flow.reactor.api.handler.EventHandler;
 import com.sequenceiq.freeipa.client.FreeIpaClient;
+import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.flow.freeipa.user.event.SetPasswordRequest;
 import com.sequenceiq.freeipa.flow.freeipa.user.event.SetPasswordResult;
 import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
+import com.sequenceiq.freeipa.service.stack.StackService;
 
 import reactor.bus.Event;
 
@@ -19,6 +22,9 @@ import reactor.bus.Event;
 public class SetPasswordHandler implements EventHandler<SetPasswordRequest> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SetPasswordHandler.class);
+
+    @Inject
+    private StackService stackService;
 
     @Inject
     private FreeIpaClientFactory freeIpaClientFactory;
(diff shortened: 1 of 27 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -31,14 +39,21 @@ public String selector() {
...
+
+            FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientForStack(stack);
             freeIpaClient.userSetPassword(request.getUsername(), request.getPassword());
             SetPasswordResult result = new SetPasswordResult(request);
             request.getResult().onNext(result);
         } catch (Exception e) {
             request.getResult().onError(e);
+        } finally {
+            MDCBuilder.cleanupMdc();
+            MDCBuilder.buildMdcContextFromMap(mdcContext);
```
lacikaaa: I don't think restoring the MdContext is necessary

handavid:
> Please be aware of due to the API changes the UI and CLI needs modification too, and this change can make it into the rc branch only when those are ready
> cc @zoltanzseli @keyki

The API changes are purely additive and are backwards compatible. I've tested with the current cb-cli and the commands work. I have an update for the cb-cli waiting for this change to go in that also clarifies that crns should be used instead of names.

SetPasswordRequest -- environments field moved to base class. No actual change to request.
SynchronizeUserRequest -- optional environments field added (from base class). empty requests still accepted
SynchronizeAllUserRequest -- environments field moved to base class. optional machineUsers field added. All existing requests still accepted.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
handavid: I see that the AsyncTaskExecutor is decorated to clean up the MDC. This one is invoked through the EventBus and I haven't tracked down whether the MDC is cleaned up automatically.
I'll investigate some more, but please let me know if you know the answer.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -31,14 +39,21 @@ public String selector() {
     @Override
     public void accept(Event<SetPasswordRequest> setPasswordRequestEvent) {
         SetPasswordRequest request = setPasswordRequestEvent.getData();
-        LOGGER.info("SetPasswordHandler accepting request {}", request);
+        Map<String, String> mdcContext = MDCBuilder.getMdcContextMap();
```
lacikaaa: Saving this context map is not relevant as `MDCCleanerThreadPoolExecutor` would cleanup it anyway after run, and on every accept the `FlowLogContextAspects` would set the context, but you can try it out

lacikaaa: also a log here would be nice as if stackService call fails, we won't have the accept log :)

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
handavid: done. I've also used the "flow" field to track the sync operation id.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
handavid: done

handavid:
I updated the PR for Mihaly's comments. I've also added the sync operation id to the MDC for better tracking. I also made some changes to the machine user handling that I had overlooked in the first commit. That change is called by UserService:113. I fixed the machine users returned by MockUserManagementService to have the correct crn.

handavid:
@keyki @lacikaaa Anything else for me to do here or is this change good to go?

lacikaaa:
you can merge it if the CLI part is ok

PR #5867, opened 2019-08-08
CB-2915 Refresh parcels during base image recovery

daszabo:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -117,6 +128,24 @@ public void upscaleCluster(HostGroup hostGroup, Collection<HostMetadata> hostMet
...
+                clouderaManagerPollingServiceProvider.deployClientConfigPollingService(stack, client, refreshParcelRepos.getId());
+        if (isExited(activateParcelsPollingResult)) {
+            throw new CancellationException("Cluster was terminated while waiting for parcel activation");
+        } else if (isTimeout(activateParcelsPollingResult)) {
+            throw new CloudbreakException("Timeout while Cloudera Manager was activating parcels.");
+        }
+        LOGGER.debug("Refreshed parcel repos");
+    }
+
+    private Boolean isPrewarmed(Long clusterId) {
```
bergerdenes: if the result is null, your check in L#118 will crash with NPE

PR #5871, opened 2019-08-08
CB-2937. Storage location parameter is missing for S3 for SDX

oleewere:
Storage location missing from SDX telemetry request


attilapalfi92:
I'm validating the fix. If it's working it's working I'm merging.

PR #5885, opened 2019-08-08
CB-2970. Telemetry publisher install fails with: account already exists error message

oleewere:
if cm cluster install flow step re-runs for some reason, it will try to re-create altus external account with the same name as it was created, so that api call can fail

can be solved easily just create an account with an additional uuid string

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerExternalAccountService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerExternalAccountService.java
@@ -11,6 +11,7 @@ import com.cloudera.api.swagger.client.ApiClient;
 import com.cloudera.api.swagger.client.ApiException;
 import com.cloudera.api.swagger.model.ApiConfigList;
 import com.cloudera.api.swagger.model.ApiExternalAccount;
+import com.cloudera.api.swagger.model.ApiExternalAccountList;
 
 @Service
 public class ClouderaManagerExternalAccountService {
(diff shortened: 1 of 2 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerMgmtTelemetryService.java
```diff
@@ -82,15 +82,15 @@ public void setupTelemetryRole(final Stack stack, final ApiClient client, final
         if (isWorkflowAnalyticsEnabled(stack, telemetry)) {
             WorkloadAnalytics workloadAnalytics = telemetry.getWorkloadAnalytics();
             ClouderaManagerResourceApi cmResourceApi = new ClouderaManagerResourceApi(client);
-            ApiConfigList apiConfigList = buildTelemetryCMConfigList(workloadAnalytics);
+            final String credentialName = String.format(ALTUS_CREDENTIAL_NAME_PATTERN, UUID.randomUUID().toString());
```
akanto: If the user is in CM, then would be it is easier to check that if the user exists then do not create it again?

(If the user would have been added to Altus, then shouldn't we store the generated user in the database to make sure that if the flow is restarted then we create the same user and if it needs to be deleted then we know what needs to be deleted)

oleewere: it's only CM cb, im checking only updateAccount call maybe that does not require user should exist

oleewere: unfortunately update throws exception if account does not exist, so i need to query accounts first, PR updated for this

oleewere:
seems it's failed on an unrelated test, will re-trigger UT

oleewere:
retest this please

PR #5886, opened 2019-08-08
CDPD-2339, CDPSDX-716, CDPSDX-715 - updated configs with warnings sup…

jasonwmcswain:
git commit -m "CDPD-2339, CDPSDX-716, CDPSDX-715 - updated configs with warnings suppressions and required service configs.  This commit includes changes to the HA blueprints to avoid technical debt in the post-GA/1.1 timeframe."

Closes #CDPD-2339
Closes #CDPSDX-715
Closes #CDPSDX-716

PR #5887, opened 2019-08-09
Revert "CDPD-2339, CDPSDX-716, CDPSDX-715 - updated configs with warnings sup…"

sodre90:
Reverts hortonworks/cloudbreak#5886

Does not work with our default CM version. @jasonwmcswain


keyki:
Unfortunately, it's a long ongoing issue that we cannot commit changes with the default version of CM in CB, because CM does not ignore unknown properties and we cannot upgrade and newer CM builds fail with different errors. We should have a jira to ignore every unknown property. Otherwise, it will be always a battle to get in new properties.

PR #5889, opened 2019-08-09
CB-2979. Instance profile is not set for Logging (DistroX)

oleewere:
- put back logging converter with extra checks


oleewere:
retest this please

oleewere:
retest this please

PR #5902, opened 2019-08-09
CB-2912 sdx flow fixes

sodre90:


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/model/common/Status.java
```diff
@@ -40,6 +48,10 @@ public boolean isAvailable() {
         return IS_AVAILABLE_LIST.contains(valueOf(name()));
     }
 
+    public boolean isTerminatedOrTerminationInProgress() {
```
gergopapi2: There is a list of termination states already in this file, you can get them via `getDeletingStatuses`.
There is one difference between the two: DELETE_FAILED is included in the `getDeletingStatuses`. I think it should be part of termination states as well. 
On the other hand I would keep only one of `TERMINATION_STATE` and `getDeletingStatuses`

PR #5909, opened 2019-08-10
CB-2958 Redbeams: Implement HaApplication

gergopapi2:


attilapalfi92:
Let's wait for another approval to be sure

PR #5910, opened 2019-08-10
CB-3000 - cloudStorage request validation in SDX and Cloudbreak

attilapalfi92:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -171,6 +176,29 @@ public SdxCluster createSdx(final String userCrn, final String name, final SdxCl
         return sdxCluster;
     }
 
+    private void validateCloudStorageRequest(SdxCloudStorageRequest cloudStorage) {
+        if (cloudStorage != null) {
+            ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
```
doktoric: can we also check that the baselocation contains the protocol ?  @attilapalfi92

attilapalfi92: For the telemetry (LOG identity) it's not necessary. I'm checking if it's necessary for SDX (IDBroker identity)

doktoric: cool thanks

attilapalfi92: added the validation for SDX

PR #5913, opened 2019-08-10
CB-2576 - Set admin group when creating environment

pdarvasi:
To be completed with Zeppelin and other services..


environment/src/main/resources/application.yml
```diff
@@ -77,6 +77,7 @@ environment:
   enabledplatforms: AWS, AZURE, YARN
   tunnel:
     default: DIRECT
+  admin.group.default.prefix: cpd_env_admin_
```
bergerdenes: typo? cdp_

pdarvasi: Thanks!

PR #5916, opened 2019-08-10
Distrox cs

doktoric:


core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/CloudStorageDecorator.java
```diff
@@ -46,19 +65,75 @@ public CloudStorageRequest decorate(CloudStorageRequest request, DetailedEnviron
...
+                recommendations = getFileSystemRecommendations(blueprintName, clusterName, baseLocation, fileSystemType.name());
+            }
+            if (request.getLocations() == null) {
+                request.setLocations(List.of());
+            }
+            for (ConfigQueryEntry recommendation : recommendations) {
+                StorageLocationBase storageLocationBase = new StorageLocationBase();
+                storageLocationBase.setType(recommendation.getType());
+                storageLocationBase.setValue(recommendation.getDefaultPath());
+                request.getLocations().add(storageLocationBase);
```
topolyai5: if the line#100 is called this will throw an exception because this is an immutable list:
`static <E> List<E> of() {
        return ImmutableCollections.emptyList();
    }
`

please write a unit test for it

PR #5918, opened 2019-08-11
CB-3005 Zeppelin Notebook For S3

smaniraju:
Zeppelin does not require any S3 Specific initialization now  and uses the default zeppelin.notebook.dir since CDPD-2933 is resolved

Closes #CB-3005


doktoric:
thanks @smaniraju

PR #5933, opened 2019-08-12
CB-3001. Use sdxcluster crn instead of stack crn of SDX for telemetry publisher

oleewere:
calculate sdx id from sdxcluster crn (not from stack crn of the sdx cluster),
as datalake api was introduced on distrox side, that can be done during request processing.

Workload analytics has attributes that can override CM service component settings, so using that if SDX response is available.
(if not available, will work as is it worked before - will try to use stack crn, if that also not available, will generate a random uuid)

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
@@ -1,9 +1,13 @@
 package com.sequenceiq.cloudbreak.converter.v4.stacks;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
```diff
@@ -13,10 +17,15 @@
 import com.sequenceiq.common.api.telemetry.response.LoggingResponse;
 import com.sequenceiq.common.api.telemetry.response.TelemetryResponse;
 import com.sequenceiq.common.api.telemetry.response.WorkloadAnalyticsResponse;
+import com.sequenceiq.sdx.api.model.SdxClusterResponse;
 
 @Component
 public class TelemetryConverter {
 
+    private static final String DATABUS_HEADER_SDX_ID = "databus.header.sdx.id";
```
doktoric: should we use configuration properties here? will this change?

oleewere: attributes are overriding the cm configs (dynamic way) for TP, this configs should go into safety valve unfortunately so no specific configs for these in CM. therefore if it will change on CM side, we will need to change that, although as i have checked, these are databus header names also dynamic in CM, but not changed for long time

PR #5934, opened 2019-08-12
CB-3021. SDX: cloud storage logging should only depend on telemetry o…

oleewere:
…bject (not on cloud storage data access)

If logging is set, but cloud storage (data access) is not, then logging is not set as well.

Expected behavior would be to setup log identity for SDX if it is enabled on environment side (that's how it is working for DistroX)

UI flows looked ok. I have tested with this cli change as well: https://github.com/hortonworks/cb-cli/pull/612

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudStorageManifester.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudStorageManifester.java
@@ -39,20 +39,29 @@ public class CloudStorageManifester {
 
     public CloudStorageRequest initCloudStorageRequest(DetailedEnvironmentResponse environment, String blueprint,
             SdxCluster sdxCluster, SdxClusterRequest clusterRequest) {
-        SdxCloudStorageRequest cloudStorage = clusterRequest.getCloudStorage();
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/converter/util/CloudStorageValidationUtil.java
```diff
@@ -17,8 +21,10 @@ public boolean isCloudStorageConfigured(CloudStorageBase cloudStorageRequest) {
             if (CollectionUtils.isEmpty(cloudStorageRequest.getLocations()) && CollectionUtils.isEmpty(cloudStorageRequest.getIdentities())) {
                 return false;
             }
-            validationBuilder.ifError(() -> CollectionUtils.isEmpty(cloudStorageRequest.getLocations()),
-                    "'locations' in 'cloudStorage' must not be empty!");
+            if (!containsOnlyLogIdentity(cloudStorageRequest.getIdentities())) {
+                validationBuilder.ifError(() -> CollectionUtils.isEmpty(cloudStorageRequest.getLocations()),
+                        "'locations' in 'cloudStorage' must not be empty!");
+            }
```
doktoric: no need for this

PR #5935, opened 2019-08-12
Cdpcp 496 env usersync refactoring on rc2.12 apurohit

abhishekaypurohit:
@handavid  - internal PR for adding machine user to grp map.

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -94,6 +94,28 @@ public class GrpcUmsClient {
     }
 
     /**
+     * Retrieves group Map from UMS for given set of machine users.
+     *
+     * @param accountId    the account Id
+     * @param requestId    an optional request Id
+     * @param machineUsers the machine users list for which all groups needs to be populated. if null or empty then returns empty map
+     * @return the map of machine user to list of groups
+     */
+    public Map<MachineUser, List<Group>> getMachineUsersToGroupsMap(String actorCrn, String accountId, List<MachineUser> machineUsers, Optional<String> requestId) {
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            Map<MachineUser, List<Group>> usersToGroupMap = new HashMap<>();
+            UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
+            LOGGER.debug("Listing group information for account {} using request ID {}", accountId, requestId);
+            for (MachineUser u : machineUsers) {
+                List<Group> groups = client.listGroupsForMembers(requestId.orElse(UUID.randomUUID().toString()), accountId, u.getCrn());
+                usersToGroupMap.put(u, groups);
+                LOGGER.debug("{} Groups found for account {}", groups.size(), accountId);
+            }
+            return usersToGroupMap;
+        }
+    }
+
+    /**
      * Retrieves user details from UMS.
      *
      * @param actorCrn  the CRN of the actor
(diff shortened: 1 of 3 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -93,6 +93,28 @@
...
 
+    /**
+     * Retrieves group Map from UMS for given set of machine users.
+     *
+     * @param accountId    the account Id
+     * @param requestId    an optional request Id
+     * @param machineUsers the machine users list for which all groups needs to be populated. if null or empty then returns empty map
+     * @return the map of machine user to list of groups
+     */
+    public Map<MachineUser, List<Group>> getMachineUsersToGroupsMap(String actorCrn, String accountId, List<MachineUser> machineUsers, Optional<String> requestId) {
```
handavid: seems strange to take MachineUsers rather than just their CRNs.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -45,89 +42,72 @@
...
 
-    private void processForEnvironmentRights(
-        Builder umsStateBuilder, String actorCrn, String envCRN, List<User> allUsers, List<MachineUser> allMachineUsers) {
+    private List<User> getUsersWithEnvironmentRights(
+        String actorCrn, String envCRN, List<User> allUsers) {
+        List<User> rightfulUsers = new ArrayList<>();
         // for all users, check right for the passed envCRN
         for (User u : allUsers) {
 
             if (umsClient.checkRight(actorCrn, u.getCrn(), environmentWrite, envCRN, Optional.empty())) {
```
handavid: does this work? environments/write is resourceIndependent just like setPassword is and should hit the same illegal argument error

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsState.java
```diff
@@ -41,10 +43,12 @@ public UmsState(Map<User, List<Group>> userToGroupsMap, Map<String, User> adminU
 
     public UsersState getUsersState(String environmentCrn) {
 
-        //TODO: parse env and get name of env
+        // TODO: Might not be needed if customer provided groups membership is done. Then we sync all members group.
         String adminGrpName = "cpd_env_admin_ap-test-env";
+
         UsersState.Builder builder = new UsersState.Builder();
 
+        // TODO: Question: Why is this used for.
```
handavid: this was used when groups were retrieved separately. The GetRightsResponse lists the group CRNs that a user is a member of. We needed to match the CRN to the Group.
I think we should go back to listing all the groups once. As it stands now, we are doing a lot of extra work to listGroups for every user.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsState.java
```diff
@@ -77,14 +81,24 @@ public UsersState getUsersState(String environmentCrn) {
...
+                    crnToGroup.put(g.getCrn(), group);
+                    builder.addMemberToGroup(group.getName(), user.getName());
+                    builder.addGroup(group);
+                });
+
+            });
         machineUserMap.entrySet()
                 .forEach(e -> {
                     FmsUser fmsUser = umsMachineUserToUser(e.getValue());
                     builder.addUser(fmsUser);
```
handavid: can put this up in the machineUserToGroupsMap.entrySet().forEach above. If we have Users and MachineUsers as the keys to the map then we don't need the extra machineUserMap.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: This is an Invalid Argument I just commented it since this is already there and this is not for master branch as of now.

PR #5939, opened 2019-08-13
CB-2862 SDX does not validate the name in post request time

horadla23:


core-api/src/main/java/com/sequenceiq/cloudbreak/validation/StackNameLengthValidator.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class StackNameLengthValidator implements ConstraintValidator<ValidStackNameLength, String> {
+
+    private static final Integer MIN_LENGTH = 5;
+
+    private static final Integer MAX_LENGTH = 40;
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (Range.between(MIN_LENGTH, MAX_LENGTH).contains(value.length())) {
```
topolyai5: you can write this: `return Range.between(MIN_LENGTH, MAX_LENGTH).contains(value.length()`

core-api/src/main/java/com/sequenceiq/cloudbreak/validation/StackNameFormatValidator.java
```diff
@@ -0,0 +1,20 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+public class StackNameFormatValidator implements ConstraintValidator<ValidStackNameFormat, String> {
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        Pattern pattern = Pattern.compile("(^[a-z][-a-z0-9]*[a-z0-9]$)");
+        Matcher matcher = pattern.matcher(value);
+        if (matcher.matches()) {
```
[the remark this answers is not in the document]
horadla23: right :D I'll fix it

core-api/src/main/java/com/sequenceiq/cloudbreak/validation/StackNameLengthValidator.java
horadla23: true :D

PR #5945, opened 2019-08-13
Cdpcp 496 env usersync refactoring on rc2.12 apurohit

abhishekaypurohit:
This PR is for a common map user/machineuser's crn -> groups 
@handavid


abhishekaypurohit:
I followed command-line instructions and pushed code to the parent branch

PR #5948, opened 2019-08-13
CB-2687 Disable redbeams dbserver test endpoint

bhavanki:
The database and database server test API endpoints for redbeams are
disabled, to eliminate the possibility of an outside, compromised
database server gaining access to redbeams or its hosting compute
resource. The endpoints may come back later once it's certain that they
can be implemented safely.

While testing, I found that CrnValidator was too strict and rejecting
null CRN values, which are valid in some cases. The validator now
permits nulls, and I added the @NotNull validation annotation in places
where the CRN really shouldn't be null.


bhavanki:
I forgot to disable the test endpoint for databases as well. I'll add that shortly.

PR #5949, opened 2019-08-13
CB-3054. Invalid Machine user Crn during Machine user creation.

oleewere:
actor is not a machine user so CRN, so it cannot get machine user details by itself. (copied that from a few lines below, but that is invalid for dbus user)


oleewere:
retest this please

oleewere:
see https://github.com/hortonworks/cloudbreak-deployer/pull/603 for it failures

PR #5950, opened 2019-08-13
CB-3041. Consider SDX internal request for cloud storage request manifester

oleewere:
creating cloud storage request behavior for SDX with internal SDX stack request:
- if cloud storage data is filled in sdx request use that
- if cloud storage data is not filled in sdx request but sdx internal request has, use that
- if logging is enabled, enrich cloud request with that, if none of the above 2 is enabled, create a new cloud request with logging
- if any from the above 3 is enabled, enrich with s3 guard dynamo table (if it has not set already in sdx internal request and it is enabled)

added a lot of logging as well
added new UTs


oleewere:
see hortonworks/cloudbreak-deployer#603 for it failures

oleewere:
retest this please

PR #5951, opened 2019-08-14
CDPCP-496. CDPCP-502. Sync users per environment based on roles.

handavid:
User rights are checked to determine which users are synced to each environment
rather than syncing all users in an account to every environment.

Functionality from UmsState was moved into the UmsUsersStateProvider which now
returns a map from environment to UsersState instead.

CDPCP-502.
Automatic user sync on environment creation is removed in this commit because
it is expected that there are no users enabled for the environment at environment
creation time.

Other minor improvements:
Logging is improved by propagating the request id from the MDC to the GrpcUmsClient
calls.


keyki:
Once you assign the proper privileges to users like PowerUser auto sync upon env creation is useful isn't it?

handavid:
> Once you assign the proper privileges to users like PowerUser auto sync upon env creation is useful isn't it?

My understanding is that the "PowerUser" check is a stopgap while we figure out why "checkRights" isn't working. I think that the end goal is to use an environment-dependent right to determine whether a user should be synced. 
@abhishekaypurohit should be able to explain more.

keyki:
> > Once you assign the proper privileges to users like PowerUser auto sync upon env creation is useful isn't it?
> 
> My understanding is that the "PowerUser" check is a stopgap while we figure out why "checkRights" isn't working. I think that the end goal is to use an environment-dependent right to determine whether a user should be synced.
> @abhishekaypurohit should be able to explain more.

Alright, that could be.

keyki:
Can you check the IT please?

abhishekaypurohit:
> > Once you assign the proper privileges to users like PowerUser auto sync upon env creation is useful isn't it?
> 
> My understanding is that the "PowerUser" check is a stopgap while we figure out why "checkRights" isn't working. I think that the end goal is to use an environment-dependent right to determine whether a user should be synced.
> @abhishekaypurohit should be able to explain more.

"checkRights" for resource level is not there yet, hence we need to check resource role and roles. Once checkRight for resource level is implemented, we can change this to "environments/write" and "environments/SetPassword" checks for right. 

However, we can still do user sync for Power Users during env creation. And once the stopgap group is there, during another user sync, membership can be updated.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -248,7 +266,7 @@ public boolean checkRight(String actorCrn, String userCrn, String right, String
         try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
             AuthorizationClient client = new AuthorizationClient(channelWrapper.getChannel(), actorCrn);
             LOGGER.info("Checking right {} for user {}!", right, userCrn);
-            client.checkRight(requestId.orElse(UUID.randomUUID().toString()), userCrn, right, null);
+            client.checkRight(requestId.orElse(UUID.randomUUID().toString()), userCrn, right, resource);
```
abhishekaypurohit: This change is not required, its merge issue, I removed this change. This method should be as is.

handavid:
> Can you check the IT please?
@keyki 
IT passed

PR #5954, opened 2019-08-14
OPSAPS-51232 Home Directories need to be setup in HDFS

lacikaaa:


orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -0,0 +1,74 @@
...
+  mapfile -t existingusers < <(cat $EXISTING_USERS_FILE)
+fi
+# hashtable for lookup later
+declare -A hasharr
+for existing in "${existingusers[@]}";
+do
+  hasharr[$existing]="exist"
+done
+number_of_dirs_created=0
+for user in "${users[@]}";
```
QwertyManiac: Just an aside: Since we have a pre-determined list of usernames (i.e. this array can be a list of paths here), it'd be far more efficient to pass them all at once to `mkdir` and `chown`, rather than do it one by one in a loop (spawns a new JVM each run).

lacikaaa: thanks for noticing, please contact with Devaraj under OPSAPS-51232 as the script belongs to him

ddraj: This is a nice suggestion; I totally forgot about the list of paths thing that I could make use of in mkdir. I tried and it indeed improves the perf significantly. The issue is that the chown cli doesn't provide a similar mechanism. We need to be able to assign individual owners to the respective home dirs, but chown doesn't support that kind of a cli. The only thing chown supports is it takes a single owner and a list of paths. That's where things break down. I guess Java is the best option to do this script functionality in...

acmurthy: A note of caution... there is a system-specific limit on the maximum number of args which can be passed to a command. YMMV.

PR #5956, opened 2019-08-14
CB-3063 Change partition in CRNs from 'altus' to 'cdp'

bergerdenes:
Make Crn class checkstyle-conform.

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/security/authentication/DisabledAuthCbUserProvider.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/security/authentication/DisabledAuthCbUserProvider.java
@@ -6,7 +6,7 @@ public class DisabledAuthCbUserProvider {
 
     public CloudbreakUser getCloudbreakUser() {
         return new CloudbreakUser("disabledAuth",
-                "crn:altus:iam:us-west-1:cloudera:user:disabledAuth",
+                "crn:cdp:iam:us-west-1:cloudera:user:disabledAuth",
                 "disabledAuth",
                 "disabledAuth@cloudera.com",
                 "cloudera");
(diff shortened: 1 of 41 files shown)


oleewere:
@bergerdenes , can you check this against real UMS for an install? i'm specifically curious about distroX install with telemetry publisher as it assign a built-in role to the generated machine user.
see: https://github.com/hortonworks/cloudbreak/blob/rc-2.12/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java#L459

to enable telemetry publisher + UMS you need the following env vars:
- ALTUS_UMS_HOST: ums.thunderhead-dev.cloudera.com
- ALTUS_UMS_PORT: 8982
- ALTUS_DATABUS_ENDPOINT: https://dbusapi.sigma-dev.cloudera.com
- CB_CM_TELEMETRYPUBLISHER_ENABLED: true
also the first 2 is needed for the cbd Profile as well (but maybe there you won't need ALTUS_ prefix...actually i have both, with or without that ... then CB will use UMS instead of use mock user service)
if you will use dp cli (i recommend it as you need to hack some cookies on the ui to make that work with UMS)

also if you are already there you can rename AltusCredential class to CDPCredential (not required, but it will be more proper from this point)

lnardai:
Also there is a special string used for InternalCrn-s, where the account id is special, and it's altus.

https://github.com/hortonworks/cloudbreak/blob/470ca39d747896151de3bbda8f1ca840fe04633f/common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/InternalCrnBuilder.java#L21

I guess that should be changed as well to CDP, but might work a check for UMS developers.

bergerdenes:
InternalCrnBuilder will be updated later (if ever). Chris Leroy approved UMS side.

PR #5957, opened 2019-08-14
CB-3001. Fix broken converter because of not rebased branch

oleewere:
one of the PRs (CB-3001) was not rebased before merging a code changed (without having any conflict), that causes an error

sdx cluster response needs to be passed to the converter, so including that


core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -179,7 +180,15 @@ private YarnStackV4Parameters getYarnProperties(DistroXV1Request source, Detaile
     public DistroXV1Request convert(StackV4Request source) {
         DistroXV1Request request = new DistroXV1Request();
         request.setName(source.getName());
-        request.setEnvironmentName(getIfNotNull(source.getEnvironmentCrn(), crn -> environmentClientService.getByCrn(crn).getName()));
+        DetailedEnvironmentResponse environment = null;
+        if (source.getEnvironmentCrn() != null) {
+            environment = environmentClientService.getByCrn(source.getEnvironmentCrn());
```
bergerdenes: possible NPE

PR #5962, opened 2019-08-14
CB-2417 Return datalake resource crn

bergerdenes:


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/view/ClusterApiViewToClusterViewV4ResponseConverter.java
```diff
@@ -49,6 +49,7 @@ protected ClusterViewV4Response createTarget() {
     private void addSharedServiceResponse(ClusterApiView cluster, ClusterViewV4Response clusterResponse) {
         SharedServiceV4Response sharedServiceResponse = new SharedServiceV4Response();
         if (cluster.getStack().getDatalakeId() != null) {
+            sharedServiceResponse.setResourceCrn(cluster.getStack().getResourceCrn());
```
oleewere: just be aware we are sending back the stack crn and no relation with the sdx cluster crn, can be confusing as you can use sdx crn for querying

doktoric: We need the sdx crn

bergerdenes: reworked

doktoric:
we still need some improvement

core/src/main/java/com/sequenceiq/cloudbreak/service/datalake/SdxClientService.java
```diff
@@ -29,4 +29,12 @@
             return new ArrayList<>();
         }
     }
+
+    public List<SdxClusterResponse> list() {
+        try {
+            return sdxEndpoint.list(null);
+        } catch (WebApplicationException | ProcessingException e) {
```
topolyai5: We should write some logs.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/SdxServiceDecorator.java
```diff
@@ -0,0 +1,63 @@
...
+        List<SdxClusterResponse> sdxClusters = sdxClientService.getByEnvironmentCrn(stack.getEnvironmentCrn());
+        if (sdxClusters.size() > 1) {
+            LOGGER.warn("More than 1 SDX cluster found for stack {}.", stack.getCrn());
+            throw new BadRequestException("Environment should have only one datalake.");
+        }
+        if (sdxClusters.size() == 1) {
+            return sdxClusters.get(0);
+        }
+        LOGGER.info("No SDX cluster found for stack {}.", stack.getCrn());
+        return new SdxClusterResponse();
```
topolyai5: why do we send back an empty response instead of null?

core/src/main/java/com/sequenceiq/cloudbreak/service/datalake/SdxClientService.java
bergerdenes: good idea

core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/SdxServiceDecorator.java
bergerdenes: reworked

PR #5965, opened 2019-08-14
CB-3052 add EKS required tags to the created subnets at env creation

biharitomi:


doktoric:
@biharitomi the unit test failed

PR #5971, opened 2019-08-14
CDPSDX-715 - suppress kafka warnings

jasonwmcswain:
This change will suppress the 3 kafka warnings that were showing up in the datalake, based on the light duty blueprint.

Notes:

including all 3 blueprint changes, but only tested the lite-duty version. (HA is not supported in GA, but adding the configs now to avoid technical debt that must be done after GA for 1.1.
Tests performed: (only "cdp-sdx.bp")

deployed cluster in mow-dev, verified via CM API that these 3 configuration values are available in the current supported CM version.
deploy via CLI with configuration changes to verify that there were no provision errors when including the 3 values in the blueprint.
Closes #CDPSDX-715


lnardai:
LGTM

PR #5973, opened 2019-08-15
OPSAPS-51232. Optimized the mkdirs (not chowns)

ddraj:
Made the mkdirs optimal (one single command).

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
@@ -38,6 +38,8 @@ EXISTING_USERS_FILE=/tmp/existing_users_file
 HDFS_KEYTAB=/run/cloudera-scm-agent/process/$(ls -t /run/cloudera-scm-agent/process/ | grep hdfs-NAMENODE$ | head -n 1)/hdfs.keytab
 echo "Keytab dir:$HDFS_KEYTAB"
 
+PATH_PREFIX="/user7"
+
 kinit -kt "$HDFS_KEYTAB" hdfs/$(hostname -f)
 if [[ $? -ne 0 ]]; then
   echo "Couldn't kinit as hdfs"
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -38,6 +38,8 @@ EXISTING_USERS_FILE=/tmp/existing_users_file
 HDFS_KEYTAB=/run/cloudera-scm-agent/process/$(ls -t /run/cloudera-scm-agent/process/ | grep hdfs-NAMENODE$ | head -n 1)/hdfs.keytab
 echo "Keytab dir:$HDFS_KEYTAB"
 
+PATH_PREFIX="/user7"
```
keyki: Do we need this prefix?

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -56,19 +58,26 @@ for existing in "${existingusers[@]}";
...
-     fi
+     echo "Will attempt to create HDFS home dir for $user"
+     newpaths+="$PATH_PREFIX/$user "
   fi
 done
-echo "End time of script $0, PID $$: `date`. Number of HDFS home dirs created $number_of_dirs_created."
+if [ -z "$newpaths" ]; then
+  echo "No home dirs to create in HDFS"
+else
+  hdfs dfs -mkdir -p $newpaths
```
lacikaaa: will it handle hundreds of arguments? this should be tested on mow env before merged

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
ddraj: Oh, the prefix should be set to /user .. I had set to user7 for testing and forgot to update.

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
ddraj: Yes tested on mow-dev

keyki:
I guess this can be closed, right?

PR #5977, opened 2019-08-15
CB-3016 The same instances will be deleted on the provider side as from Ambari

schfeca75:
@keyki @biharitomi pls review


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/FlowChainHandler.java
```diff
@@ -39,7 +39,7 @@ public void accept(Event<? extends Payload> event) {
         FlowEventChainFactory<Payload> flowEventChainFactory = flowChainConfigMap.get(key);
         String flowChainId = UUID.randomUUID().toString();
         flowChains.putFlowChain(flowChainId, parentFlowChainId, flowEventChainFactory.createFlowTriggerEventQueue(event.getData()));
-        flowChains.triggerNextFlow(flowChainId);
+        flowChains.triggerNextFlow(flowChainId, Map.of());
```
pdarvasi: Is this always an empty map?

PR #5985, opened 2019-08-15
CB-3095. NOT_FOUND: Role 'crn:cdp:iam:us-west-1:altus:role:DbusUpload…

oleewere:
…er' does not exist in the account 'altus

- change is related with https://github.com/hortonworks/cloudbreak/pull/5956

- there is a hidden built-in role in UMS, which we need to use for machine users, hopefully for the new cdp partition, there is a cdp account as well, then this change should be enough.

testing is in progress agains real UMS, i have just opening the PR to let Ut and It finish in time


keyki:
I can merge regardless. Can't harm anything :D

oleewere:
@keyki , right, to be honest we will require testing on mow dev anyway, and if the role still does not exist we need to ask the UMS guys about what is a valid one here (or the role is even needed for cdp machine users)

PR #5986, opened 2019-08-15
CB-3038 CRNs for Recipe, Blueprint, Image Catalog, Cluster definition should use official names

foldik:
Changed CRN for Blueprint, Image Catalog, Recipe, Cluster resources. @horadla23 @doktoric please review.

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
@@ -37,19 +37,24 @@ public class Crn {
     private static final boolean NON_ADMIN_SERVICE = false;
 
     /**
-     * The Altus partition in which the resource resides. An Altus partition is a
-     * namespace for sets of Altus services.
+     * The CDP partition in which the resource resides. A CDP partition is a
+     * namespace for sets of CDP services.
      */
     public enum Partition {
+        /**
+         * @deprecated since CDP project
+         */
+        @Deprecated
         ALTUS("altus", "ccs"),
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
```diff
@@ -144,7 +144,8 @@ public static Partition safeFromString(String input) {
         DSADMIN("dsadmin", ADMIN_SERVICE),
         SAMPLE("sample", NON_ADMIN_SERVICE),
         WORKSPACES("workspaces", NON_ADMIN_SERVICE),
-        FREEIPA("freeipa", NON_ADMIN_SERVICE);
+        FREEIPA("freeipa", NON_ADMIN_SERVICE),
```
doktoric: where we using this ?

foldik: In [CrnService.java](https://github.com/hortonworks/cloudbreak/blob/681e20a6f15ac4b5a648892df9ff0f73bc1870a0/freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java#L66)

horadla23: It seems we are using it here: https://github.com/hortonworks/cloudbreak/blob/681e20a6f15ac4b5a648892df9ff0f73bc1870a0/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaCreationService.java#L102

PR #5991, opened 2019-08-15
CB-2878 - Hive Needs to be in HTTP Mode for Knox to Proxy J/ODBC Access

pdarvasi:
Awaiting further input on parameters


lajosrodek:
Follow-up to #5848

PR #5996, opened 2019-08-15
CB-2954 Register db servers before launch

bhavanki:
The flow for redbeams database server creation is rearranged.
Registration of the new server - that is, recording of it in the
redbeams database as a database server config entity - now occurs
synchronously before the flow begins, instead of after the database
server (and other resources in its stack) is created. This enabled
retrieval of the server before launch completes.

The entity does not start out with a host and port, however; these are
null in the initial state. The former registration state / action of the
flow has been repurposed into a registration update state. The only work
done here is updating the server entity with the host and port, which
are available (as usual) after launch completes. This change caused the
renaming of events, states, and the corresponding handler, which does
much less work than before.

It is now permissible for a DatabaseServerConfig entity to have a null
host and port. Such a condition represents a server that has not yet
been fully provisioned. Until a host and port are entered for such a
server, redbeams refuses to create a database on it.

Other changes:

* A new unit test for RedbeamsCreationService is in place.
* The unit test for the former RegisterDatabaseServerHandler, now
UpdateDatabaseServerRegistrationHandler, has been chopped down
appropriately.
* The resource CRN for a new DB stack is determined in
RedbeamsCreationService instead of in DatabaseServerV4Controller. Code
in DatabaseServerConfigService responsible for creating a new server row
in the database only generates a resource CRN for the server if it isn't
already in place in the entity.
* The unused DB stack statuses PROVISION_SETUP and METADATA_COLLECTION
are eliminated.
* A new update method in DatabaseServerConfigService allows registration
update to work through that service instead of directly through the
corresponding JPA repository.


redbeams/src/main/java/com/sequenceiq/redbeams/service/stack/RedbeamsCreationService.java
```diff
@@ -65,8 +83,32 @@ public DBStack launchDatabaseServer(DBStack dbStack) {
...
+        dbServerConfig.setName(dbStack.getName());
+        dbServerConfig.setEnvironmentId(dbStack.getEnvironmentId());
+        dbServerConfig.setConnectionDriver(dbStack.getDatabaseServer().getConnectionDriver());
+        // username and password are set during conversion to DBStack
+        dbServerConfig.setConnectionUserName(dbStack.getDatabaseServer().getRootUserName());
+        dbServerConfig.setConnectionPassword(dbStack.getDatabaseServer().getRootPassword());
+        dbServerConfig.setDatabaseVendor(dbStack.getDatabaseServer().getDatabaseVendor());
+        dbServerConfig.setDbStack(dbStack);
+        // host and port are set after allocation is complete
+        dbServerConfig.setHost(null);
```
jadair-cloudera: Is this necessary? Does it not default to null?

bhavanki: Probably not. I have further changes to make, so I'll drop these at that point.

PR #6004, opened 2019-08-16
CB-2986 - Azure regions response is incorrect

pdarvasi:


cloud-azure/src/main/resources/definitions/azure-zone-coordinates.json
```diff
@@ -174,6 +174,71 @@
...
+      "latitude": "-25.731340",
+      "longitude": "28.218370"
+    },
+    {
+      "name": "South Africa West",
+      "latitude": "-34.075691",
+      "longitude": "18.843266"
+    },
+    {
+      "name": "China North",
```
lacikaaa: china north is already added. isn't it a problem?

lacikaaa: china north is already added. isn't it a problem?

pdarvasi: Corrected, thx

PR #6007, opened 2019-08-16
CB-3080 Set CapacityScheduler as default scheduler for YARN in the Data Engineering blueprint

miklosgergely:
Use CapacityScheduler instead of the default FairScheduler.

Tested by creating clusters with it on MOW.


keyki:
Spark, zeppelin and ozzie are also part of the template. My understanding is that those services won't work as they won't have the ability to submit jobs to YARN.

keyki:
According to jira, this needs to be worked on.

PR #6008, opened 2019-08-16
CDPD-3127 Update OpDB blueprints and add first cluster template

joshelser:
Two blueprints now: one for the cluster template, optimized for
the instance types in that template. A second with no configuration
optimizations.

### DIFF
--- a/cloud-common/src/main/resources/application.yml
+++ b/cloud-common/src/main/resources/application.yml
@@ -160,6 +160,7 @@ cb:
                 CDP 1.0 - Data Engineering HA: Apache Spark, Apache Livy, Apache Zeppelin=cdp-data-engineering-ha;
                 CDP 1.0 - Data Mart: Apache Impala, Hue=cdp-data-mart;
                 CDP 1.0 - Operational Database: Apache HBase=cdp-opdb;
+                CDP 1.0 - Operational Database: Apache HBase (Custom)=cdp-opdb-custom;
                 CDP 1.0 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx;
                 CDP 1.0 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha;
 
--- a/core/src/main/resources/defaults/blueprints/cdp-opdb-custom.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-opdb-custom.bp
@@ -1,8 +1,8 @@
 {
-  "description": "CDP 1.0 Operational Database with Apache HBase",
+  "description": "CDP 1.0 Operational Database with Apache HBase (custom)",
   "blueprint": {
     "cdhVersion": "7.0.0",
-    "displayName": "opdb",
+    "displayName": "opdb-custom",
     "services": [
       {
         "refName": "zookeeper",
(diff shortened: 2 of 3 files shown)


joshelser:
@keyki some updates to the opdb blueprint for HBase.

A pair of cluster template and blueprint, as well as a "basic" blueprint with the instance-size-dependent configuration properties omitted.

This is what I've been testing with. Should be OK.

keyki:
Do you want the new blueprint to be offered as default? In that case you need to include it here: https://github.com/hortonworks/cloudbreak/blob/rc-2.12/cloud-common/src/main/resources/application.yml#L155

joshelser:
> Do you want the new blueprint to be offered as default? In that case you need to include it here

Yeah. Let me add that (forgot about it).

core/src/main/resources/defaults/blueprints/cdp-opdb-custom.json
```diff
@@ -0,0 +1,158 @@
...
+            "refName": "knox-KNOX-GATEWAY-BASE",
+            "roleType": "KNOX_GATEWAY",
+            "configs": [
+              {
+                "name": "gateway_dispatch_whitelist",
+                "value": "^https?:\\/\\/.+\\.cloudera\\.site:[0-9]+.*$"
+              },
+              {
+                "name": "gateway_knox_admin_groups",
+                "value": "admins"
```
keyki: I don't think we need this property as it will change per env to be cdp_$env_name

core/src/main/resources/defaults/blueprints/cdp-opdb-custom.json
```diff
@@ -0,0 +1,158 @@
...
+        "serviceType": "KNOX",
+        "roleConfigGroups": [
+          {
+            "base": true,
+            "refName": "knox-KNOX-GATEWAY-BASE",
+            "roleType": "KNOX_GATEWAY",
+            "configs": [
+              {
+                "name": "gateway_dispatch_whitelist",
+                "value": "^https?:\\/\\/.+\\.cloudera\\.site:[0-9]+.*$"
```
keyki: Is this property needed?

core/src/main/resources/defaults/blueprints/cdp-opdb-custom.json
joshelser: Yeah, I can drop this one now, thanks.

core/src/main/resources/defaults/blueprints/cdp-opdb-custom.json
joshelser: It was a while back. I wasn't aware of anything that changed which would preclude the need for it, but I can double check.

joshelser: Yep, not sure when this was fixed, but I don't have to set this by hand anymore. Thanks for asking.

joshelser:
I tested fresh deploys on both blueprint updates.

joshelser:
Thanks sir!!

lnardai:
We need to revert this, as the cluster templates causes issue on mow-dev.
That a bug on our side, we don't have proper validation on the clustertemplate endpoint.
We also need to have a PR time validation check on that endpoint.

joshelser:
Bummer. Can you let me know what the issue is so I can fix it? I should have cycles later today to fix it if necessary.

PR #6016, opened 2019-08-17
CB-2966: remove hue hive service from blacklist

jdesjean:
In CB-2966 Hue team requested to add Hive to the blacklist. This was due to lack of resource for testing Hive in Hue in DE cluster. We managed to pull it off and can confirm it's working so we'd like to remove Hive from blacklist. Sorry of the back and forth.
Pig still needs to be there as Pig will not be supported by DE in 7.0


jdesjean:
@keyki @tiborpopovics

PR #6018, opened 2019-08-18
CB-3080 Set CapacityScheduler as default scheduler for YARN in the Data Engineering blueprint

miklosgergely:


sunilgovind:
6021 is latest. I think this need to be cancelled.
cc @ashutoshc @miklosgergely

PR #6030, opened 2019-08-19
CDPD-3127 Update OpDB blueprints and add first cluster template

joshelser:
This reverts commit bd8c3685f6bd419bdddcd00435cf3b705cbc93b2.

Revert the first revert (re-apply the original) with some modifications.

@lnardai @daszabo trying a re-apply of the same template from over the weekend which broke things. Best as I can tell, the `distroXTemplate` attribute was the only thing missing, but, for the life of me, I can't seem to figure out a way to validate before asking you all to review it. Sorry :(

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/aws/aws-opdb-700.json
+++ b/core/src/main/resources/defaults/clustertemplates/aws/aws-opdb-700.json
@@ -1,0 +1,159 @@
+{
+  "name": "CDP 1.0 - Operational Database template",
+  "description": "",
+  "type": "OTHER",
+  "cloudPlatform": "AWS",
+  "distroXTemplate": {
+    "cluster": {
+      "blueprintName": "CDP 1.0 - Operational Database: Apache HBase"
+    },
+    "instanceGroups": [
+      {
+        "name": "master",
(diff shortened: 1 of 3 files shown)


core/src/main/resources/defaults/clustertemplates/aws/aws-opdb-700.json
```diff
@@ -0,0 +1,159 @@
+{
+  "name": "CDP 1.0 - Operational Database template",
+  "description": "",
+  "type": "OPERATIONALDATABASE",
```
daszabo: This is an invalid value. What 'type' currently does, is it tells the UI, what icon it should present for this template. Right now there is no mapping for 'OPERATIONALDATABASE' and the enum 'ClusterTemplateV4Type' has no such value right now. Please change this to an existing value, like 'OTHER'.

joshelser: Ah ok. Thanks! I didn't realize it was an enum.

joshelser:
Ok! Pushed a change for that `type:OTHER`

PR #6035, opened 2019-08-20
CDPB-3228 Hue needs full blacklist 

lnardai:
- because our config overrides the original values


jdesjean:
+1

keyki:
Basically we're disabling everything?

jdesjean:
Basically only file browser and a couple oozie functions left

PR #6038, opened 2019-08-20
CB-3166 Add multiAZ and storageType RDS options

bhavanki:
Redbeams now supports parameterized choices for whether to deploy a
multi-AZ RDS instance (default is true) and for the storage type of the
instance. Because of a bug in the Go implementation of Swagger, the
multiAZ property must be a string instead of a Boolean, or else Go
clients will never sent a false value for the property.


keyki:
I don't think this should go to rc-2.12 at this point.

bhavanki:
> I don't think this should go to rc-2.12 at this point.

No problem, I'll retarget to master.

PR #6039, opened 2019-08-21
CDPSDX-606 Cleanup the CM keytabs secrets

jamisonbennett:
CDPSDX-606 Cleanup the CM keytabs secrets
    
    Delete the CM keytab secrets when deleting a cluster with CM. Add
    kerberosmgmt delete APIs that use query parameters rather than the
    entity body so that Jersey can validate the args. This fixes the
    error "Entity must be null for http method DELETE" when using Jersey
    clients.
    
    Tested manually using a local cloudbreak deployment.

This pull request is for master so that it makes it into 2.13 rather than 2.12.

Closes #CDPSDX-606

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
@@ -35,12 +35,12 @@ public class KerberosMgmtV1Controller implements KerberosMgmtV1Endpoint {
         return kerberosMgmtV1Service.getExistingServiceKeytab(request, accountId);
     }
 
-    public void deleteServicePrincipal(@Valid ServicePrincipalRequest request) throws FreeIpaClientException {
+    public void deleteServicePrincipal(@Valid ServicePrincipalRequest request) throws FreeIpaClientException, DeleteException {
         String accountId = crnService.getCurrentAccountId();
         kerberosMgmtV1Service.deleteServicePrincipal(request, accountId);
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/termination/CleanupFreeIpaTask.java
```diff
@@ -59,5 +63,14 @@ public void run() {
         } catch (Exception e) {
             LOGGER.error("FreeIPA cleanup failed", e);
         }
+        // Cleanup vault entries
+        fqdns.forEach((String fqdn) -> {
```
lacikaaa: this logic should be moved to freeipa CleanupService, as the request has all the necessary info, and invoking FreeIPA mgmt service for every host is not ideal

jamisonbennett: I moved it to the `CleanupService` which allows some of the changes to be simplified.
But I am not exactly sure how to know which hosts are the "CM" hosts. I tried to get the gateway like in `KeytabConfigurationHandler` and `KeytabProvider` (these generate the CM keytab/secret), but when this is run on the freeipa `CleanupService`, it tries to delete the gateway which ends up being `ipaserver0.cloudera.site` and that doesn't work since it is the IPA server. I am not sure if there is a better way to filter the contents or not. If not, maybe we just run the delete for nodes which are no-ops.

lacikaaa: calling delete for all hosts is not a problem

PR #6043, opened 2019-08-21
CB-3170 - NPE fixed during termination in CloudFileSystemViewBuilder

attilapalfi92:


core/src/main/java/com/sequenceiq/cloudbreak/converter/spi/StackToCloudStackConverter.java
```diff
@@ -202,7 +202,7 @@ InstanceTemplate buildInstanceTemplate(Template template, String name, Long priv
         List<Group> groups = new ArrayList<>();
         Cluster cluster = stack.getCluster();
         if (cluster != null && cluster.getBlueprint() != null) {
```
lnardai: If we are using the BlueprintText, probably we won't need to getBlueprint nullcheck

core/src/main/java/com/sequenceiq/cloudbreak/converter/spi/CloudFileSystemViewBuilder.java
```diff
@@ -25,7 +25,8 @@
         Optional<CloudFileSystemView> fileSystemView;
         if (fileSystem != null) {
             SpiFileSystem spiFileSystem = fileSystemConverter.fileSystemToSpi(fileSystem);
-            if (componentsByHostGroup.get(instanceGroup.getGroupName()).contains(KnoxRoles.IDBROKER)) {
+            Set<String> components = componentsByHostGroup.get(instanceGroup.getGroupName());
+            if (components != null && components.contains(KnoxRoles.IDBROKER)) {
                 fileSystemView = spiFileSystem.getCloudFileSystems().stream()
                         .filter(cloudFileSystemView -> cloudFileSystemView.getCloudIdentityType().equals(CloudIdentityType.ID_BROKER))
```
lnardai: Can you swap the equals to have constants in front?

keyki:
@attilapalfi92 Talked with @lnardai, can you please change the base version to rc-2.12?

attilapalfi92:
@keyki done

PR #6046, opened 2019-08-21
OPSAPS-52161. Avoid overwriting root krb5 ccache, speed up chown

toddlipcon:
This fixes the createuserhome.sh script (which is invoked once a minute
by cron) so that it no longer uses the default krb5 ticket cache
location. Instead it uses a cloudbreak-specific one. That means that it
won't interfere with any end user who might be trying to kinit on that
same node. This should fix some of our test cases which were doing that
same thing.

Along the way I noted that the existing 'chown' path was quite slow --
it took about 17 minutes to perform the chowns for 623 users (each one
takes about 2 seconds, sometimes slower). In contrast, using curl to
invoke webhdfs takes about 10ms per operation. The full sync for 623
users took 10 seconds.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
@@ -35,6 +35,26 @@ kinit_as_hdfs()     {
       echo "Couldn't kinit as hdfs"
       exit 1
     fi
+    # If we got an authentication error, maybe our cookie was somehow bad as well,
+    # so make sure we get a new one on the next attempt.
+    rm -f $WEBHDFS_COOKIE_JAR
+}
+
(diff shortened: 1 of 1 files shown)


toddlipcon:
I tested this by scping the script to a distrox master node, removing /tmp/existing_users, and running it. I also tested modifying the script temporarily to put an 'x' in the middle of the path and saw that the error handling was working for FileNotFoundException (the error got output and the script exited). Also verified that it properly kinits into the new path and doesn't impact the default credential cache of the root user anymore.

keyki:
It's a huge improvement, I think we should put this to rc-2.12 to release it with GA. @RamVenkatesh @akanto any objections?

keyki:
Thanks @ddraj, @toddlipcon can you please change the base branch to rc-2.12?

toddlipcon:
ok, base changed to 2.12.

PR #6052, opened 2019-08-21
DWX-945. Allow anonymous LDAP access to group membership

toddlipcon:
This changes the IPA setup to allow anonoymous (non-bound) LDAP clients
to get group membership. This is slightly more open, security-wise, but
we already expose the list of users, so exposing the list of groups
isn't much worse.

Longer term the solution is provide a read-only LDAP "bind" credential
(CB-2488) but this should unblock DWX in the short term.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
@@ -20,4 +20,21 @@ ipa-server-install \
           --auto-forwarders \
           --unattended
 
+export KRB5CCNAME=$(mktemp krb5cc.XXXXXX)
+echo $FPW | kinit admin
+
+# Allow anonymous (non-bound) LDAP users to search for
+# group membership. This is a workaround for DWX-945 until
+# we have a properly set up bind user/password.
+basedn=$(ipa env basedn | awk '{print $2}')
+ipa permission-add \
+  --right=search \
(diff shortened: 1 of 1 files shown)


toddlipcon:
I tested this command manually by copy-pasting it into a running environment, but not sure how to test that it works in the salt freeIPA setup context.

Also not sure if there is a better way to run this via salt so that it would get applied to already-launched IPAs (is that possible?)

If someone from the CB team could make sure this is safe before merging that would be great. This is critical to fix DWX-945 which is on the hot list.

jamisonbennett:
There isn't currently a way to update existing FreeIPA servers. Although salt supports it, cloudbreak doesn't do this. It is my understanding that this may be possible in the future when we have HA support, but as of a few weeks ago when I made a similar change to the salt config there was no way to perform and update.

toddlipcon:
I think given DWX pulled in https://github.infra.cloudera.com/CDH/dwx/pull/483/ this one isn't needed as urgently. They switched to using admin credentials to bind, which while not preferable, means this shouldn't be required yet.

We should still merge it, but perhaps into master instead of 2.12? Or alternatively we can set up a proper bind account with the appropriate privileges.

acmurthy:
Ok, let's track this or a better approach for master i.e. September drop. Thanks @toddlipcon!

lacikaaa:
also this kind of change should be applied through rest API in `FreeIpaPostInstallService`

keyki:
Closing the PR until we come to a conclusion to prevent accidental merge.

PR #6059, opened 2019-08-22
CB-3023. Improve create machine user retry logic

oleewere:
- make retry around the whole create machine user operation - with that we will generate a new request id for the call
- add retry annotation for generate keys as well (as create machine user is called through that)
- double the retries


oleewere:
retest this please

keyki:
Okay. I'm retesting it.

PR #6060, opened 2019-08-22
CDPSDX-800 Fix cleanup of kerberos secrets when deleting an env

jamisonbennett:
CDPSDX-800 Fix cleanup of kerberos secrets when deleting an env
    
    Delete all vault secrets that are associated with an environment when
    that environment is deleted. This prevents the secrets from remaining
    in the control plane after the environment has been deleted.
    
    Manually tested using a local development version of cloudbreak.

    
Closes #CDPSDX-800

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
@@ -1,6 +1,5 @@
 package com.sequenceiq.freeipa.kerberosmgmt.v1;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -64,8 +63,6 @@ public class KerberosMgmtV1Service {
 
     private static final String IPA_STACK_NOT_FOUND = "Stack for IPA server not found.";
 
-    private static final String VAULT_SECRET_TYPE = "ServiceKeytab";
-
     private static final String KEYTAB_SUB_TYPE = "keytab";
 
     private static final String PRINCIPAL_SUB_TYPE = "serviceprincipal";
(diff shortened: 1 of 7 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -170,12 +170,20 @@ public void cleanupByCluster(VaultCleanupRequest request, String accountId) thro
...
-                    .withEnvironmentCrn(request.getEnvironmentCrn())
-                    .withClusterCrn(request.getClusterCrn());
-            recursivelyCleanupVault(b.withSubType(PRINCIPAL_SUB_TYPE).build());
-            recursivelyCleanupVault(b.withSubType(KEYTAB_SUB_TYPE).build());
+            cleanupSecrets(request.getEnvironmentCrn(), request.getClusterCrn(), accountId);
+        } catch (DeleteException e) {
+            throw e;
+        } catch (Exception e) {
+            LOGGER.error("Cleanup cluster failed " + e.getLocalizedMessage(), e);
+            throw new DeleteException("Failed to cleanup");
```
lacikaaa: I think it would make sense to add at least the message of the exception here

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -394,7 +411,7 @@ private static String constructCanonicalPrincipal(String serviceName, String hos
 
     private static class VaultPathBuilder {
```
lacikaaa: this inner class could be moved out and then separately covered with unit tests, so we could ensure it's backward compatible

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -184,6 +192,15 @@ public void cleanupByCluster(VaultCleanupRequest request, String accountId) thro
         }
     }
 
+    private void cleanupSecrets(String environmentCrn, String clusterCrn, String accountId) throws DeleteException {
+        VaultPathBuilder b = new VaultPathBuilder()
```
lacikaaa: variable should be renamed to something more informative, like `vaultPathBuilder`

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
jamisonbennett: Done

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
jamisonbennett: Done

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
jamisonbennett: Done

PR #6061, opened 2019-08-22
CB-2054 - CB-2053 - Integrated Jaeger OpenTracing with Environment service and CB

attilapalfi92:


auth-connector/src/main/java/com/sequenceiq/cloudbreak/client/ThreadLocalUserCrnWebTargetBuilder.java
```diff
@@ -22,10 +22,17 @@
 
     private ClientRequestFilter clientRequestFilter;
 
+    private Class<?> tracer;
```
bbihari: Class<? extends Feature>  ?

PR #6064, opened 2019-08-22
CB-2878 remove http mode config from hive without tez execution engine

keyki:


keyki:
@lnardai @akanto this is the PR to remove the http mode when hiverserver is not in tez mode.

PR #6065, opened 2019-08-22
CB-3211 Skip RDS termination if CF template is missing

bhavanki:
When AwsRdsTerminateService fails to locate the CF template for a
database server to terminate, rather than fail, it now assumes that the
template has been properly deleted, and returns successfully (without
itself issuing a termination request). This prevents this likely common
occurrence from causing termination failures.

If a redbeams server is already in a DELETE_FAILED status because of
some earlier termination failure, it turns out you can issue another
termination and redbeams will try again. This provides an escape hatch
for failed terminations, by manually getting rid of the CF template and
then rerunning termination.

---

This is currently targeted at rc-2.12, but won't be merged without approval.


daszabo:
We are waiting for the green light from Ram, whether this may still go rc-2.12 or master.

lnardai:
Got the approve, merging.

PR #6058, opened 2019-08-22
CDPD-3348  Increase DE Workers Count to 2.

smaniraju:
Increase DE Workers Count to 2.

Closes #CDPD-3348

PR #6078, opened 2019-08-23
CB-3200 Increase DE Blueprint Worker Node Count to 3

smaniraju:
CDPD-3348 Increase DE Blueprint workers cardinality


lajosrodek:
Follow-up to #6058.

PR #6089, opened 2019-08-26
CDPCP-606. Metering consumer failed to convert payload to MeteringEvent entity

oleewere:
applying fix to rc-2.12 (in order to test metering on mow dev on this week)

also will apply https://github.com/hortonworks/cloudbreak/pull/6082 as well but only after CDPCP-628 is fixed


keyki:
Is this intended to go to rc-2.12? The PR was against master.

PR #6093, opened 2019-08-26
CB-3243 Cleanup CRN handling from CB-2907

jamisonbennett:
CRN handling is cleaned up to use the Crn class to get the resource
rather than just using everything after the last ":".

NOTE: cluster-ambari/src/main/java/com/sequenceiq/cloudbreak/ambari/AmbariBlueprintPortConfigCollector.java still uses the last ":" rather than the using the CRN class. I didn't update this because I don't know how to test it and I think it may be deprecated. If this assumption is wrong, please let me know and I will change it there too.

Closes #CB-3243

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/VaultPathBuilder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/VaultPathBuilder.java
@@ -4,9 +4,10 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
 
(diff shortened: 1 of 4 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/VaultPathBuilder.java
```diff
@@ -44,14 +45,16 @@ public VaultPathBuilder withSubType(String subType) {
...
+            return Crn.safeFromString(crn).getResource();
+        });
         return this;
     }
 
     public VaultPathBuilder withClusterCrn(String clusterCrn) {
-        String crn = StringUtils.substringAfterLast(clusterCrn, ":");
-        this.clusterCrn = Optional.ofNullable(crn);
+        this.clusterId = Optional.ofNullable(clusterCrn).map((String crn) -> {
+            return Crn.safeFromString(crn).getResource();
```
lacikaaa: while `StringUtils.substringAfterLast` won't throw exception, `Crn.safeFromString` can throw. Shouldn't we handle this?

PR #6106, opened 2019-08-27
CB-3230 Fix deleting vault kv version 2 secrets

jamisonbennett:
Deleting the vault kv version 2 secrets was fixed so that the key and
all values are fully deleted. Version 2 allows the secrets to be
versioned and keys can be undeleted. Version 2 also supports a
destroy which cleans up the data for a specific version so undelete
is not possible for that version. But if all versions are destroyed,
the key still remains stored without any versions of the value. The
key and all versions can be fully destroyed by deleting the metadata
for the secret. The delete operation was updated to delete the
metadata.

This was tested manually on a local build of cloudbreak. The vault
UI was checked to ensure the secret was actually deleted.

Closes #CB-3230


keyki:
Yes, unfortunately it's not implemented yet: https://github.com/spring-projects/spring-vault/issues/432

PR #6108, opened 2019-08-28
CB-2971 Increase DB max_connections from 100 to 500

foldik:
Configure embedded postgres via Salt to allow 500 connections.
@akanto please review.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
@@ -44,11 +44,26 @@ configure-listen-address:
       - service: start-postgresql
     - unless: test -f /var/log/pgsql_listen_address_configured
 
+/opt/salt/scripts/conf_pgsql_max_connections.sh:
+  file.managed:
+    - makedirs: True
+    - mode: 755
+    - source: salt://postgresql/scripts/conf_pgsql_max_connections.sh
+
+configure-max-connections:
+  cmd.run:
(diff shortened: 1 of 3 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
```diff
@@ -44,11 +44,26 @@ configure-listen-address:
...
+  file.managed:
+    - makedirs: True
+    - mode: 755
+    - source: salt://postgresql/scripts/conf_pgsql_max_connections.sh
+
+configure-max-connections:
+  cmd.run:
+    - name: runuser -l postgres -c '/opt/salt/scripts/conf_pgsql_max_connections.sh' && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/pgsql_max_connections_configured
+    - require:
+      - file: /opt/salt/scripts/conf_pgsql_max_connections.sh
```
akanto: Salt has a replace state called file.replace: https://docs.saltstack.com/en/develop/ref/states/all/salt.states.file.html

and it should have an append_if_not_found state.

foldik: Or even we could have just use the `psql -c 'ALTER SYSTEM SET max_connections TO 500'`. It would write this config into the postgres.auto.conf file which overwrites the value comming from postgres.conf. What do you think?

akanto: do we need a running postgres for this command? If not then I am ok with it.

foldik: Okay, so I checked it. We use Postgres 9.2 and the `ALTER SYSTEM` feature is only allowed from 9.4, so we can't change it with the official psql command. Should I rewrite to use salt replace states or sed based replacement is fine for now?

PR #6116, opened 2019-08-28
CB-3298. Query machine user for crn only if user already exist during user creation.

oleewere:
query against machine user only if it exists or the response does not contain the machine user crn. if user already exists that means that is in the UMS index so we can safely send a get query. therefore we can avoid eventual consistency issues for machine user creation

edit: validated


oleewere:
retest this please

PR #6117, opened 2019-08-28
CDPCP-591 - While environment creation, as a post installation proces…

abhishekaypurohit:
This change will enable user sync during the post-installation process of the newly launched environment. With this change, all PUs will be synced into FreeIPA. 

Closes CDPCP-591

@keyki @handavid @lacikaaa

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
@@ -13,7 +13,7 @@ import com.sequenceiq.freeipa.controller.exception.SyncOperationAlreadyRunningEx
 import com.sequenceiq.freeipa.service.freeipa.user.PasswordService;
 import com.sequenceiq.freeipa.service.freeipa.user.SyncOperationStatusService;
 import com.sequenceiq.freeipa.service.freeipa.user.UserService;
-import com.sequenceiq.freeipa.util.FMSUtil;
+import com.sequenceiq.freeipa.util.CrnService;
 
 import java.util.Set;
 
@@ -43,7 +43,7 @@ public class UserV1Controller implements UserV1Endpoint {
 
     @Override
     public SyncOperationStatus synchronizeUser(SynchronizeUserRequest request) {
-        String userCrn = FMSUtil.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
+        String userCrn = CrnService.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
         String accountId = threadBaseUserCrnProvider.getAccountId();
         LOGGER.debug("synchronizeUser() requested for user {} in account {}", userCrn, accountId);
         Set<String> environmentCrnFilter = request == null ? Set.of() : nullToEmpty(request.getEnvironments());
@@ -66,7 +66,7 @@ public class UserV1Controller implements UserV1Endpoint {
 
     @Override
     public SyncOperationStatus synchronizeAllUsers(SynchronizeAllUsersRequest request) {
-        String userCrn = FMSUtil.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
+        String userCrn = CrnService.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
         String accountId = threadBaseUserCrnProvider.getAccountId();
         LOGGER.debug("synchronizeAllUsers() requested for account {}", accountId);
 
@@ -76,7 +76,7 @@ public class UserV1Controller implements UserV1Endpoint {
 
     @Override
     public SyncOperationStatus setPassword(SetPasswordRequest request) {
-        String userCrn = FMSUtil.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
+        String userCrn = CrnService.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
         String accountId = threadBaseUserCrnProvider.getAccountId();
         LOGGER.debug("setPassword() requested for user {} in account {}", userCrn, accountId);
 
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
@@ -1,14 +1,5 @@
 package com.sequenceiq.freeipa.controller;
 
-import java.util.Set;
-
-import javax.inject.Inject;
-import javax.validation.constraints.NotNull;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.stereotype.Controller;
-
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.freeipa.api.v1.freeipa.user.UserV1Endpoint;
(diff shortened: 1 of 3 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -42,7 +43,7 @@
 
     @Override
     public SyncOperationStatus synchronizeUser(SynchronizeUserRequest request) {
-        String userCrn = checkUserCrn();
+        String userCrn = FMSUtil.checkUserCrn(threadBaseUserCrnProvider.getUserCrn());
```
lacikaaa: doesn't compile, most probably because it's missing FMSUtil class
but anyway, this method should be in CrnService as it seems like it belongs there

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -57,5 +59,12 @@ public void postInstallFreeIpa(Long stackId) throws Exception {
             freeIpaClient.setUsernameLength(MAX_USERNAME_LENGTH);
         }
         passwordPolicyService.updatePasswordPolicy(freeIpaClient);
+
+        // CDPCP-591 - Enable User Sync after environment created
```
lacikaaa: please remove comments

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -57,5 +59,12 @@ public void postInstallFreeIpa(Long stackId) throws Exception {
...
+        Set<String> environmentCrnFilter = new HashSet<>();
+        environmentCrnFilter.add(stack.getEnvironmentCrn());
+        userService.synchronizeUsers(
```
lacikaaa: env filter could be simply `Set.of(stack.getEnvironmentCrn())` and getting the usercrn would make more sense to be moved out

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -22,6 +13,16 @@
 import com.sequenceiq.freeipa.service.freeipa.user.PasswordService;
 import com.sequenceiq.freeipa.service.freeipa.user.SyncOperationStatusService;
 import com.sequenceiq.freeipa.service.freeipa.user.UserService;
+import com.sequenceiq.freeipa.util.CrnService;
```
handavid: unnecessary re-ordering of imports.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -1,5 +1,14 @@
 package com.sequenceiq.freeipa.service.freeipa.flow;
 
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
```
handavid: unnecessary re-ordering of imports

freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java
```diff
@@ -1,5 +1,11 @@
 package com.sequenceiq.freeipa.util;
 
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
```
handavid: unnecessary re-ordering of imports

freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java
```diff
@@ -69,4 +70,12 @@ public String createCrn(String accountId, Crn.ResourceType resourceType) {
                 .setResource(UUID.randomUUID().toString())
                 .build().toString();
     }
+
+    public static String checkUserCrn(String userCrn) {
+        if (userCrn == null) {
+            throw new BadRequestException("User CRN must be provided");
```
handavid: BadRequestException doesn't really make sense here. Other CrnService methods throw CrnParseExceptions. This method should throw a CrnParseException to match the behavior of the other methods.
UserV1Controller should catch CrnParseException and convert those to BadRequestExceptions otherwise those exceptions will go out at 500 Internal Server Errors.

freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java
abhishekaypurohit: why un-necessary? And moreover its due to settings from CDP and CB. I cant do much there.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
abhishekaypurohit: why un-necessary? And moreover its due to settings from CDP and CB. I cant do much there.

freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java
abhishekaypurohit: Make sense. I can change that.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
abhishekaypurohit: why un-necessary? And moreover its due to settings from CDP and CB. I cant do much there.

freeipa/src/main/java/com/sequenceiq/freeipa/util/CrnService.java
```diff
@@ -69,4 +69,12 @@ public String createCrn(String accountId, Crn.ResourceType resourceType) {
                 .setResource(UUID.randomUUID().toString())
                 .build().toString();
     }
+
+    public static String checkUserCrn(String userCrn) {
+        if (userCrn == null) {
```
handavid: This method doesn't really make much sense to me. It's not doing anything except a null-check.Also, why is it static?

I think it would make more sense to make this
`public Crn safeFromString(String userCrn)`
Then you can also call this from `getCurrentAccountId` and `getCurrentUserId` to ensure that they follow the same exception-handling behavior

handavid:
Also, can you squash your commits?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
handavid: I think that @lacikaaa was suggesting that you are attempting to add stack.getEnvironmentCrn() twice to environmentCrnFilter. You can simplify this block by
```
userService.synchronizeUsers(
    threadBasedUserCrnProvider.getAccountId(), 
    CrnService.checkUserCrn(threadBasedUserCrnProvider.getUserCrn()),
    Set.of(stack.getEnvironmentCrn(), Set.of(), Set.of());
```
It's strange to have to get the crn from the threadBasedUserCrnProvider and then call checkUserCrn. I think that we can change CrnService.checkUserCrn to CrnService.getUserCrn and not inject the ThreadBasedUserCrnProvider into this class at all.

PR #6123, opened 2019-08-29
CB-2933 - fixed NAMENODE port and JOBTRACKER url

attilapalfi92:


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology_api.xml.j2
```diff
@@ -196,7 +196,7 @@
     <service>
         <role>JOBTRACKER</role>
         {% for hostloc in salt['pillar.get']('gateway:location')['RESOURCEMANAGER'] -%}
-        <url>rpc://{{ hostloc }}:{{ ports['JOBTRACKER'] }}/ws</url>
```
keyki: Why is this removed?

attilapalfi92: I'm not sure what's needed and what is not here anymore. @ckoncz-hwx can you elaborate? Shall I leave this here?

ckoncz-hwx: The Knox documentation for Oozie service configuration does not say anything about a path in the rpc:// url:
https://knox.apache.org/books/knox-1-3-0/user-guide.html#Oozie
I did run Oozie tests through Knox and they worked fine with a JOBTRACKER url without a `/ws` path.

keyki: Right, this is RPC. Thanks.

PR #6129, opened 2019-08-29
CB-3077 GUI E2E teardown solution for Jenkins builds

aszegedi:
Please review and let me know where can I improve or refactor this if it is needed.

### DIFF
--- a/integration-test/src/main/resources/application.yml
+++ b/integration-test/src/main/resources/application.yml
@@ -8,6 +8,9 @@ logging:
     path: ${integrationtest.outputdir}
 
 integrationtest:
+  user:
+    accessKey:
+    secretKey:
   database:
     connectionUrl: jdbc:postgresql://localhost:5432/
     username: postgres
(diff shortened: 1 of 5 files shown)


integration-test/src/main/resources/application.yml
```diff
@@ -7,6 +7,14 @@ spring:
 logging:
     path: ${integrationtest.outputdir}
 
+cleanupapp:
```
lnardai: Do we need this? I mean we need to have the same account as the test run in, there is no real point defining it twice. It can only lead to errors.
Or at least it should default to the integrationtest variables

lnardai:
Will this be used as a best effort job or as a crtiteria to start the tests?

integration-test/src/main/resources/application.yml
aszegedi: `cleanupapp.user.accessKey` and `cleanupapp.user.secretKey` not that keys as we have in the `integrationtest` for credentials and not same as the integration test user's have. The environment can be same. So I'm checking this out how can it work if the `server` parameters are coming from the `integrationtest`.

aszegedi:
I can set the GUI E2E job to ignore the result of this.

PR #6130, opened 2019-08-29
CB-3306. Add HDFS user to Yarn admin ACL

ajyoung:
The Telemetry Publisher component in Cloudera Manager runs as the HDFS user in this context and needs to have permission to view all jobs in order to collect the YARN logs from S3 when cloud storage is enabled.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
@@ -47,7 +47,7 @@
         "serviceConfigs": [
           {
             "name": "yarn_admin_acl",
-            "value": "yarn,hive"
+            "value": "yarn,hive,hdfs"
           }
         ],
         "roleConfigGroups": [


core/src/main/resources/defaults/blueprints/cdp-data-engineering.bp
```diff
@@ -47,7 +47,7 @@
         "serviceConfigs": [
           {
             "name": "yarn_admin_acl",
-            "value": "yarn,hive"
+            "value": "yarn,hive,hdfs"
```
akanto: We have quite a few default blueprints, shouldn't it be fixed everywhere?

ajyoung: Hmm... This is a good question Attila :)
I looked at the other nine default blueprints. Of these, the following configure the YARN service:

1. **cdp-data-engineering**
2. cdp-data-engineering-ha
3. cdp-data-science
4. cdp-data-mart
5. cdp-de-llap
6. cdh6-shared-services
7. cdh6-data-science

But only the cdp-data-engineering blueprint has the yarn_admin_acl configured. If I add the yarn_admin_acl configuration to the other blueprints, I'm not sure what the impact/side effects may be.

Should I go ahead and add it anyway?

oleewere: @ajyoung we should focus on cdp ones. impact/side effects shoyld be figured out on CM side. i think it depends on what are the defaults. (like if the original default(s) would be different from hive/hdfs/yarn...that can have some bad affect)

ajyoung: @oleewere @akanto I would like to modify the other four blueprints on a separate commit. That way, if there are unforeseen problems, the other commit can be reverted without reverting this change which is needed to resolve a Blocker Bug (CB-3306). Accordingly, I've opened a separate pull request here: https://github.com/hortonworks/cloudbreak/pull/6160

oleewere: we already have the new one, so ok then

PR #6137, opened 2019-09-03
CB-3323 Add Retry to Ambari calls

bergerdenes:


bergerdenes:
I need to bump ambari-client version after that build is successful

bergerdenes:
https://github.com/hortonworks/ambari-rest-client/pull/162

PR #6141, opened 2019-09-04
CB-3175

gergopapi2:


redbeams/src/main/java/com/sequenceiq/redbeams/converter/stack/AllocateDatabaseServerV4RequestToDBStackConverter.java
```diff
@@ -87,12 +90,18 @@
     @Inject
     private NetworkParameterAdder networkParameterAdder;
 
+    @Inject
+    private GrpcUmsClient umsClient;
```
bhavanki: There is CrnUserDetailsService which seems to be a better, higher-level choice for retrieving the user details. It uses the umsClient under the hood, but also handles internal users and machine users.

gergopapi2: I checked that. However, it returns interface `org.springframework.security.core.userdetails.UserDetails`, that is more about grantedAuthorities and credentials, and is not really used. True I do need `CloudbreakUser` or `CrnUser`, subclasses of `UserDetails`, but then I would need to do a cast. 
Or did you have something else in mind that would be a better place to get the user email?

bhavanki: I see your point. Looking at CrnUserDetailsService, I think it would actually be good if the return type of CrnUserDetailsService::loadUserByUsername were changed to CrnUser. That would make it a "covariant return type", which is perfectly valid in Java.

https://stackoverflow.com/questions/1882584/what-is-a-covariant-return-type

With that change, a downcast would not be required.

gergopapi2: Ok, got it. I did do the change.

PR #6142, opened 2019-09-04
CB-3275 domain name provider based on environment and account names

biharitomi:


common/src/main/java/com/sequenceiq/cloudbreak/dns/EnvironmentBasedDomainNameProvider.java
```diff
@@ -0,0 +1,63 @@
...
+public class EnvironmentBasedDomainNameProvider {
+    static final String FAILED_TO_GENERATE_DOMAIN_MSG = "Failed to generate domain: The environment and/or account name contains invalid characters!";
+
+    static final String NAMES_SHOULD_BE_SPECIFIED_MSG = "Environment and account names should be specified!";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnvironmentBasedDomainNameProvider.class);
+
+    private static final String DOMAIN_PART_DELIMITER = ".";
+
+    private static final String DOMAIN_PATTERN = "^((?=[a-z0-9-]{1,63}.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*.)+[a-z]{2,63}$";
```
lajosrodek: I suppose `.` should be escaped as it was probably meant literally (in both occasions), not as an "any character" wildcard. The part `(xn--)?` is not clear to me, but I can live with it. 😄

biharitomi: Thanks for catching it! :D
[It is based on an OReilly pattern for internationalized domain names](https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s15.html).

lajosrodek: Ah, that convention is totally new to me. 😮 Then it makes perfectly sense! 👍

PR #6143, opened 2019-09-04
CB-3270: create private subnets

tiborpopovics:


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/base/EnvironmentNetworkBase.java
```diff
@@ -27,6 +27,9 @@
     @Size(max = 255)
     private String networkCidr;
 
+    @ApiModelProperty(EnvironmentModelDescription.PRIVATE_SUBNET_CREATION)
+    private PrivateSubnetCreation privateSubnetCreation;
```
doktoric: DEfault value ?

tiborpopovics: The default value is set in the request converter: https://github.com/hortonworks/cloudbreak/pull/6143/files#diff-dad87862505d41ef3739cace72eadf15R165

PR #6146, opened 2019-09-04
CDPCP-602 progress report sync operation

abhishekaypurohit:
Currently, we update the sync operation at the very end rather than for each env. This change wil l make that for each environment. 

- Closes CDPCP-602

@handavid @lacikaaa @keyki

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
@@ -91,21 +91,20 @@ public class PasswordService {
                 requests.add(triggerSetPassword(stack, stack.getEnvironmentCrn(), userId, password));
             }
 
-            List<SuccessDetails> success = new ArrayList<>();
-            List<FailureDetails> failure = new ArrayList<>();
             for (SetPasswordRequest request : requests) {
                 try {
                     waitSetPassword(request);
-                    success.add(new SuccessDetails(request.getEnvironment()));
+                    syncOperationStatusService.updateOperation(operationId, new SuccessDetails(request.getEnvironment()), null);
                 } catch (InterruptedException e) {
                     LOGGER.error("Interrupted while setting passwords for user {} in account {}", userCrn, accountId);
                     throw e;
                 } catch (Exception e) {
                     LOGGER.debug("Failed to set password for user {} in environment {}", userCrn, request.getEnvironment(), e);
-                    failure.add(new FailureDetails(request.getEnvironment(), e.getLocalizedMessage()));
+                    syncOperationStatusService.updateOperation(
+                        operationId, null, new FailureDetails(request.getEnvironment(), e.getLocalizedMessage()));
                 }
             }
-            syncOperationStatusService.completeOperation(operationId, success, failure);
+            syncOperationStatusService.completeOperation(operationId);
         } catch (InterruptedException e) {
             syncOperationStatusService.failOperation(operationId, e.getLocalizedMessage());
             Thread.currentThread().interrupt();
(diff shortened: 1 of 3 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -91,21 +91,20 @@ private void asyncSetPasswords(String operationId, String accountId, String acto
                 requests.add(triggerSetPassword(stack, stack.getEnvironmentCrn(), userId, password));
             }
 
-            List<SuccessDetails> success = new ArrayList<>();
-            List<FailureDetails> failure = new ArrayList<>();
             for (SetPasswordRequest request : requests) {
                 try {
                     waitSetPassword(request);
-                    success.add(new SuccessDetails(request.getEnvironment()));
+                    syncOperationStatusService.updateOperation(operationId, new SuccessDetails(request.getEnvironment()), null);
```
handavid: I would prefer two separate methods to addSuccess and addFailure rather than passing in null. I think that would make behavior of the methods more clear.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/SyncOperationStatusService.java
```diff
@@ -67,15 +67,29 @@ public SyncOperation startOperation(String accountId, SyncOperationType syncOper
...
+            List<SuccessDetails> successList = syncOperation.getSuccessList();
+            successList.add(success);
+            syncOperation.setSuccessList(successList);
+        }
+        if (failure != null) {
+            List<FailureDetails> failureList = syncOperation.getFailureList();
+            failureList.add(failure);
+            syncOperation.setFailureList(failureList);
+        }
+        return syncOperationRepository.save(syncOperation);
```
handavid: you need to catch and retry on OptimisticLockingExceptions. you also need to handle if the update operation method fails (i.e., out of retries).

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -118,29 +115,28 @@ private void asyncSynchronizeUsers(
                                 return synchronizeStack(stack, envToUmsStateMap.get(stack.getEnvironmentCrn()), userIdFilter);
                             })));
 
-            List<SuccessDetails> success = new ArrayList<>();
-            List<FailureDetails> failure = new ArrayList<>();
-
             statusFutures.forEach((envCrn, statusFuture) -> {
```
handavid: this doesn't really update the sync status on completion of each environment because statusFuture.get() is called serially. Update for completed environments can be blocked on the statusFuture.get() of an environment that it still running.
I think the status check and should be pushed into the synchronizeStack method.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -91,21 +91,20 @@ private void asyncSetPasswords(String operationId, String accountId, String acto
                 requests.add(triggerSetPassword(stack, stack.getEnvironmentCrn(), userId, password));
             }
 
-            List<SuccessDetails> success = new ArrayList<>();
-            List<FailureDetails> failure = new ArrayList<>();
             for (SetPasswordRequest request : requests) {
                 try {
                     waitSetPassword(request);
```
handavid: this doesn't really update the sync status on completion of each environment because waitSetPassword is called serially. Update for completed environments can be blocked on the waitSetPassword of an environment that it still running.
Moving the status update to where the work is actually done will be a little tricky because the password is being set by the SetPasswordHandler and triggered through the EventBus.

lacikaaa:
I'm closing this as there hasn't been any update for more than a month. please reopen if necessary, maybe rebased onto the latest rc or master branch

PR #6151, opened 2019-09-05
CB-3185 - Increase size of root disk on Azure

pdarvasi:


orchestrator-salt/src/main/resources/salt/bootstrapnodes/mount-disks.sh
```diff
@@ -120,6 +120,26 @@ create_directories() {
     return 1
 }
 
+grow_mount_partition() {
+    # TODO: remove this line if cloud-utils-growpart is already on the image
+    yum -y install cloud-utils-growpart
+    if growpart -N $(get_root_disk) $(get_root_disk_partition_number) ; then
+        log $LOG_FILE Before growing: $(df -h)
+        growpart $(get_root_disk) $(get_root_disk_partition_number)
```
gergopapi2: we could analyze the output of growpart, or at least log it: it might not be the same as the result with dryrun.

PR #6154, opened 2019-09-05
CB-3314 Allow null AZ in EnvironmentBaseNetworkConverter

bhavanki:
The convertToLegacyNetwork method in EnvironmentBaseNetworkConverter
filters subnets in a network response based on their availability
zones. However, Azure subnets have no availability zones. Therefore,
this method now checks for that condition, and allows matching those
subnets against a null AZ passed to the method, which occurs during
validation in Azure stack creation.


bhavanki:
Already fixed in CB-3332.

PR #6157, opened 2019-09-05
envbased authz poc

horadla23:


horadla23:
This UMS change needed, because currently environment rights are defined only as resource independent and will not work during resource based permission check. https://review.infra.cloudera.com/r/105182/

PR #6160, opened 2019-09-05
CB-3306. Add Yarn admin ACL to CDP blueprints containing YARN services

ajyoung:
While reviewing the commit for [CB-3306](https://github.com/hortonworks/cloudbreak/pull/6130), @akanto requested that I make the same ACL change on the other CDP services with YARN. However, upon inspection, I noticed the other CDP services with YARN don't have YARN ACLs specified at all. This commit adds YARN admin ACLs to the other blueprints containing YARN.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-ha.bp
@@ -175,6 +175,12 @@
       {
         "refName": "yarn",
         "serviceType": "YARN",
+        "serviceConfigs": [
+          {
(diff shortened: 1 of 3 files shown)


oleewere:
the change looking good from CB perspective. But on CM side you should test these changes, if the value is set by default as `*`, maybe it can have affect on other services like hbase or oozie. the first commit is ok, this one should be validated more (as in these bps, the default was used)

oleewere:
@ajyoung , any update here? do we still need the changes or defaults are ok? if ok then we can close this PR

oleewere:
as we discussed offline im closing it for now

PR #6161, opened 2019-09-05
CB-2187. Add CCM tunnel registration to cloud-init.

jadair-cloudera:
With these changes, all the plumbing has been done for communicating
CCM parameters down through the freemarker templates to the cloud-init
user data script.

Specific changes include:
1. New model objects for CCM parameters, including server parameters
   that represent information required to contact CCM, instance
   parameters required to identify the instance to CCM, and tunnel
   parameters which represent the tunnels to be registered with CCM.
2. The ability to pass the CCM parameters from the stack service
   through the image service to the user data builder. Note that
   with this commit, the CCM parameters are always absent. They
   will be populated from communication with the mina management
   service in a subsequent commit.
3. Updates to the freemarker templates for the cloud-init user data
   script for both CB and FreeIPA, to do all the work required to
   configure and invoke the CCM client scripts.
3. Logic to add the CCM parameters to the template models for
   freemarker template processing. Note that with this commit,
   the absent CCM parameters means the ccmEnabled template model
   parameter will be absent, and none of the CCM part will be
   included in the generated user data. I have tested locally with
   populated template model parameters based on hard-coded CCM
   parameters, and observed not only correct user data generation,
   but actual successful tunneling.
4. Updates to the freemarker template for the Cloud Formation stack,
   and corresponding template model updates in the AWS stack request
   helper, to support user data up to the fully allowed 16k, passed
   in up to four chunks, each of size up to 4k. The previous mechanism
   used only a single parameter, which was limited to 4k in size.
   Although the current commit does not increase the user data size,
   manual testing revealed that the new keys for CCM pushed us past
   the 4k limit, but nowhere near the 16k limit.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
@@ -36,6 +36,14 @@ import com.sequenceiq.common.api.type.InstanceGroupType;
 @Component
 public class AwsStackRequestHelper {
 
+    // The number of chunks for large parameters
+    @VisibleForTesting
+    static final int CHUNK_COUNT = 4;
+
+    // The size of each chunk for large parameters
+    @VisibleForTesting
+    static final int CHUNK_SIZE = 4096;
+
     @Inject
     private AwsTagPreparationService awsTagPreparationService;
 
@@ -75,9 +83,10 @@ public class AwsStackRequestHelper {
             keyPairName = awsClient.getExistingKeyPairName(stack.getInstanceAuthentication());
         }
 
-        Collection<Parameter> parameters = new ArrayList<>(asList(
-                new Parameter().withParameterKey("CBUserData").withParameterValue(stack.getImage().getUserDataByType(InstanceGroupType.CORE)),
-                new Parameter().withParameterKey("CBGateWayUserData").withParameterValue(stack.getImage().getUserDataByType(InstanceGroupType.GATEWAY)),
+        Collection<Parameter> parameters = new ArrayList<>();
+        addParameterChunks(parameters, "CBUserData", stack.getImage().getUserDataByType(InstanceGroupType.CORE), CHUNK_COUNT);
+        addParameterChunks(parameters, "CBGateWayUserData", stack.getImage().getUserDataByType(InstanceGroupType.GATEWAY), CHUNK_COUNT);
+        parameters.addAll(asList(
                 new Parameter().withParameterKey("StackName").withParameterValue(stackName),
                 new Parameter().withParameterKey("StackOwner").withParameterValue(String.valueOf(ac.getCloudContext().getUserName())),
                 new Parameter().withParameterKey("KeyName").withParameterValue(keyPairName),
@@ -103,6 +112,34 @@ public class AwsStackRequestHelper {
             }
         }
         return parameters;
+    }
+
+    @VisibleForTesting
+    void addParameterChunks(Collection<Parameter> parameters, String baseParameterKey, String parameterValue, int chunkCount) {
+        int chunk = 0;
+        String parameterKey = baseParameterKey;
+        if (parameterValue == null) {
+            parameters.add(new Parameter().withParameterKey(parameterKey).withParameterValue(null));
+        } else {
+            int len = parameterValue.length();
+            int offset = 0;
+            int limit = CHUNK_SIZE;
+            // Add full chunks
+            while ((chunk < chunkCount) && (len > limit)) {
+                parameters.add(new Parameter().withParameterKey(parameterKey).withParameterValue(parameterValue.substring(offset, limit)));
+                chunk++;
+                parameterKey = baseParameterKey + chunk;
+                offset = limit;
+                limit += CHUNK_SIZE;
+            }
+            // Add the final partial chunk
+            parameters.add(new Parameter().withParameterKey(parameterKey).withParameterValue(parameterValue.substring(offset, len)));
+        }
+        // Pad with empty chunks if necessary
+        while (++chunk < chunkCount) {
+            parameterKey = baseParameterKey + chunk;
+            parameters.add(new Parameter().withParameterKey(parameterKey).withParameterValue(""));
+        }
     }
 
     private String getRootDeviceName(AuthenticatedContext ac, CloudStack cloudStack) {
--- a/common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
@@ -1,0 +1,58 @@
+package com.sequenceiq.cloudbreak.ccm.cloudinit;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Default {@link CcmParameters} implementation.
+ */
+public class DefaultCcmParameters implements CcmParameters, Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * The CCM server parameters, which specify how to connect to CCM.
+     */
+    private final ServerParameters serverParameters;
+
+    /**
+     * The instance parameters, which specify how the instance identifies itself to CCM.
+     */
+    private final InstanceParameters instanceParameters;
+
+    /**
+     * The list of CCM tunnel parameters, each of which specifies how to register a tunnel for a single service with CCM.
+     */
+    private final List<TunnelParameters> tunnelParameters;
+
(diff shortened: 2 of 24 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
```diff
@@ -75,9 +81,10 @@ public DeleteStackRequest createDeleteStackRequest(String cFStackName) {
             keyPairName = awsClient.getExistingKeyPairName(stack.getInstanceAuthentication());
         }
 
-        Collection<Parameter> parameters = new ArrayList<>(asList(
-                new Parameter().withParameterKey("CBUserData").withParameterValue(stack.getImage().getUserDataByType(InstanceGroupType.CORE)),
-                new Parameter().withParameterKey("CBGateWayUserData").withParameterValue(stack.getImage().getUserDataByType(InstanceGroupType.GATEWAY)),
+        Collection<Parameter> parameters = new ArrayList<>();
+        addParameterChunks(parameters, "CBUserData", stack.getImage().getUserDataByType(InstanceGroupType.CORE), CHUNK_COUNT);
+        addParameterChunks(parameters, "CBGateWayUserData", stack.getImage().getUserDataByType(InstanceGroupType.CORE), CHUNK_COUNT);
```
lajosrodek: Not sure this change is intentional but the original code used `InstanceGroupType.GATEWAY` here.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Default {@link CcmParameters} implementation.
+ */
+public class DefaultCcmParameters implements CcmParameters, Serializable {
+
+    private static final long serialVersionUID = 1L;
```
lajosrodek: Do we rely on the Java default serialization format? If so, having the same `serialVersionUID` for all classes below might not be the best practice. 😃

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/InstanceParameters.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Adds keys and values corresponding to the CCM parameters to the specified template model.
+     *
+     * @param model the template model map
+     */
+    default void addToTemplateModel(Map<String, Object> model) {
+        getTunnelInitiatorId().ifPresent(s -> model.put(CcmParameterConstants.TUNNEL_INITIATOR_ID_KEY, s));
+        model.put(CcmParameterConstants.ENCIPHERED_PRIVATE_KEY,
+                BaseEncoding.base64().encode(getEncipheredPrivateKey().getBytes(StandardCharsets.UTF_8)));
+//        model.put(CcmParameterConstants.ENCIPHERED_PRIVATE_KEY,
```
lajosrodek: Perhaps should be removed.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/ServerParameters.java
```diff
@@ -0,0 +1,54 @@
...
+        model.put(CcmParameterConstants.CCM_HOST_KEY, ccmHostAddressString);
+        int ccmSshPort = ccmSshServiceEndpoint.getPort().orElse(CcmParameterConstants.DEFAULT_CCM_SSH_PORT);
+        model.put(CcmParameterConstants.CCM_SSH_PORT_KEY, ccmSshPort);
+        // Put public key into known_hosts format
+        String formattedPublicKey = String.format(
+                CcmParameterConstants.CCM_PUBLIC_KEY_FORMAT,
+                ccmHostAddressString, ccmSshPort, getCcmPublicKey());
+        model.put(CcmParameterConstants.CCM_PUBLIC_KEY_KEY,
+                BaseEncoding.base64().encode(formattedPublicKey.getBytes(StandardCharsets.UTF_8)));
+//        model.put(CcmParameterConstants.CCM_PUBLIC_KEY_KEY,
```
lajosrodek: Probably this comment is not needed either.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
```diff
@@ -0,0 +1,56 @@
...
+     * Creates default CCM parameters with the specified parameters.
+     *
+     * @param serverParameters   the CCM server parameters, which specify how to connect to CCM
+     * @param instanceParameters the instance parameters, which specify how the instance identifies itself to CCM
+     * @param tunnelParameters   the list of CCM tunnel parameters, each of which specifies how to register a tunnel for a single service with CCM
+     */
+    public DefaultCcmParameters(ServerParameters serverParameters, InstanceParameters instanceParameters, List<TunnelParameters> tunnelParameters) {
+        this.serverParameters = Objects.requireNonNull(serverParameters, "serverParameters is null");
+        this.instanceParameters = Objects.requireNonNull(instanceParameters, "instanceParameters is null");
+        this.tunnelParameters = Objects.requireNonNull(tunnelParameters, "tunnelParameters is null");
```
bhavanki: For full immutability, make an immutable copy of the passed-in list.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
jadair-cloudera: Great catch!

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/InstanceParameters.java
jadair-cloudera: Oops. Meant to remove those.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/ServerParameters.java
jadair-cloudera: Oops. Meant to remove those.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
jadair-cloudera: The specific value of the serialVersionUID does not matter, and does not need to be globally unique. It only needs to be unique within a single class, and only needs to change values when the class is modified in a way that is not serialization compatible. Setting the initial value to 1L is one convention. It happens to be the one we use in Director, but it is also used elsewhere in the CB codebase, such as in SecretProxy.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
jadair-cloudera: Done

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
```diff
@@ -105,6 +112,33 @@ public DeleteStackRequest createDeleteStackRequest(String cFStackName) {
         return parameters;
     }
 
+    private void addParameterChunks(Collection<Parameter> parameters, String baseParameterKey, String parameterValue, int chunkCount) {
```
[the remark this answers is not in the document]
jadair-cloudera: Done.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/ServerParameters.java
jadair-cloudera: Done.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/InstanceParameters.java
jadair-cloudera: Done.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/cloudinit/DefaultCcmParameters.java
lajosrodek: Alright, let us keep following that convention then. 👍

lajosrodek:
@jadair-cloudera Could you please rebase to latest `master`? The integration-test failure you ran into has since been fixed in #6164 (a couple of commits after the parent yours is currently based on).

PR #6163, opened 2019-09-06
CDPSDX-839 Allow a role to be deleted when deleting a host or service

jamisonbennett:
An optional way to delete roles was added to the kerberos management
delete host and delete service principal APIs. An attempt is made to
ensure the role is not in use, although the verification has a race
condition, the check will handle most common use cases and it is
better than nothing.

This was tested manually using a local deployment of cloudbreak.

Closes #CDPSDX-839


jamisonbennett:
I got the comment on the last PR about refactoring KerberosMgmtV1Service due to the size of the class. I will get that next and I created CB-3420 for it.

lacikaaa:
int test is going to be fixed with https://github.com/hortonworks/cloudbreak/pull/6164, please rerun when it's merged

PR #6164, opened 2019-09-06
Fix integration test license.txt issue

horadla23:


integration-test/scripts/download-cbd.sh
```diff
@@ -33,4 +33,6 @@ fi
 
 chmod +x cbd
 cbd_version=$(./cbd --version)
+mkdir etc
+cat ../../mock-caas/src/test/resources/etc/license.txt > etc/license.txt
```
lacikaaa: why cat and not cp?

PR #6165, opened 2019-09-06
CB-3365 new "delete multiple instances" stack api endpoint

bergerdenes:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v3/StackV3Endpoint.java
```diff
@@ -162,6 +163,16 @@ Response deleteInstance(@PathParam("workspaceId") Long workspaceId, @PathParam("
...
             @QueryParam("forced") @DefaultValue("false") boolean forced);
 
+    // deleteInstancesStackV3
+    @DELETE
+    @Path("{name}/instances")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = StackOpDescription.DELETE_MULTIPLE_INSTANCES_BY_ID_IN_WORKSPACE, produces = ContentType.JSON, notes = Notes.STACK_NOTES,
+            nickname = "deleteMultipleInstancesStackV3")
+    Response deleteMultipleInstances(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
+            @QueryParam("id") List<String> instances,
```
lacikaaa: I would add a `@NotEmpty` here

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackService.java
```diff
@@ -548,7 +548,17 @@ public void removeInstances(Long stackId, Long workspaceId, Set<String> instance
             InstanceMetaData metaData = validateInstanceForDownscale(instanceId, stack, workspaceId, user);
             instanceIdsByHostgroupMap.computeIfAbsent(metaData.getInstanceGroupName(), s -> new LinkedHashSet<>()).add(metaData.getPrivateId());
         }
-        flowManager.triggerStackRemoveInstances(stackId, instanceIdsByHostgroupMap);
+        flowManager.triggerStackRemoveInstances(stackId, instanceIdsByHostgroupMap, false);
+    }
+
+    public void removeInstances(Stack stack, Long workspaceId, List<String> instanceIds, boolean forced, User user) {
```
lacikaaa: most part of this method is duplicating `public void removeInstances(Long stackId, Long workspaceId, Set<String> instanceIds, User user)`. the common part should be moved out to a separate method

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v3/StackV3Endpoint.java
bergerdenes: ok

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackService.java
bergerdenes: ok

PR #6181, opened 2019-09-10
CB-3447: Added cloud storage location for DataCatalog Profiler Services

subrat-mishra:
DataCatalog Profiler Services requires cloud storage location. 
All the 3 services ProfilerAdmin, ProfilerMetrics & ProfilerScheduler need a property: **file_system_uri** which holds the **default Bucket location**.


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/profileradmin/ProfilerAdminCloudStorageRoleConfigProvider.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Component
+public class ProfilerAdminCloudStorageRoleConfigProvider
+        implements CmTemplateComponentConfigProvider {
+    private static final String FILE_SYSTEM_URI = "file_system_uri";
+
+    @Override
+    public List<ApiClusterTemplateConfig> getServiceConfigs(CmTemplateProcessor templateProcessor,
+                                                            TemplatePreparationObject templatePreparationObject) {
+        return ConfigUtils.getStorageLocationForServiceProperty(templatePreparationObject, FILE_SYSTEM_URI)
```
lajosrodek: I wonder how this will work when setting different paths for the 3 Profiler services in the JSON passed to the CDP / DP CLI. `getStorageLocationForServiceProperty()` finds the first storage location entry corresponding to `FILE_SYSTEM_URI`. In this case, all 3 services use the same property key, so picking the right location entry seems impossible if the service type is not included among the filtering criteria. (As seen in `cm-cloud-storage-location-specification.json` below, the new location entries all have the same `"defaultPath"` setting, so the UI will populate all of them with the same path. This does not necessarily hold when using the CLI.)

@smaniraju Interested in your opinion as well. I know the `StorageLocationView` fetched from `TemplatePreparationObject.getFileSystemConfigurationView().getLocations()` still reflects the legacy Ambari conventions. Should not we eventually include the service type there as an additional field (and probably remove `configFile` as it is not really sensible in CM semantics)?

smaniraju: Yes StorageLocationView can be updated to include service name(s). For this Jira, I think only one entry should be added in _cm-cloud-storage-location-specification.json_ and a comma separated list of related services. Also i think we need another flag to indicate whether a  property needs to be shown in UI  or not since the locations list is getting big.

subrat-mishra: I did the testing in dp-cli by adding s3 bucket path of Profilers like below:
`{
        "type": "PROFILER_ADMIN_FS_URI",
          "value": "s3a://BUCKET/storage"
        },
        {
          "type": "PROFILER_METRICS_FS_URI",
          "value": "s3a://BUCKET/storage"
        }`
Similar to the cdp-smoke-tests by kevin 
https://github.infra.cloudera.com/krisden/cdp-smoke-tests/blob/master/sdx.json#L72-L92
@lajosrodek, can you please let me know whether that's the right way of testing?

subrat-mishra: @smaniraju, Can I change the cm-cloud-storage-location-specification.json to below:
    {
      "type": "PROFILER_FS_URI",
      "propertyName": "file_system_uri",
      "propertyFile": "dpprofiler.properties",
      "description": "This is the file system URI for Profiler Service",
      "defaultPath": "{{{ defaultPath }}}",
      "propertyDisplayName": "Profiler Service file system URI",
      "relatedServices": ["PROFILER_ADMIN_AGENT", "PROFILER_METRICS_AGENT", "PROFILER_SCHEDULER_AGENT"],
      "requiredForAttachedCluster": true,
      "supportedStorages": ["ADLS", "WASB", "S3", "GCS", "ADLS_GEN_2"]
    }
I hope the way I am doing local testing as mentioned above is fine. Please let me know.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/ConfigUtils.java
```diff
@@ -85,4 +89,19 @@ public static String getDbTypePostgres(RdsView rdsView, String serviceType) {
...
+                    .orElse(HdfsRoleConfigProvider.DEFAULT_NAME_SERVICE);
+            return "hdfs://" + nameService;
+        }
+
+        Set<String> namenodeHosts = HdfsRoleConfigProvider.nameNodeFQDNs(templatePreparationObject);
+        if (namenodeHosts.size() == 1) {
+            return "hdfs://" + namenodeHosts.iterator().next();
+        }
+        return "";
+    }
```
smaniraju: I dont think this is required since by default all the services are configured to use hdfs by default in CM and invoking HdfsRoleConfigProvider from ConfigUtil is not recommended. The order of invocation of these configproviders is also not guaranteed. Please test the default value when cluster is launched without cloud storage before overriding these to non-cloud storage values.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/profileradmin/ProfilerAdminCloudStorageRoleConfigProvider.java
smaniraju: Yes PROFILER_FS_URI  can be used as above but it is not recommended for any service to write data at the root path of cloud storage, please append {{{defaultPath}}}/profiler/data or similar.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/profilermetrics/ProfilerMetricsCloudStorageRoleConfigProvider.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final String FILE_SYSTEM_URI = "file_system_uri";
+
+    @Override
+    public List<ApiClusterTemplateConfig> getServiceConfigs(CmTemplateProcessor templateProcessor,
+                                                            TemplatePreparationObject templatePreparationObject) {
+        return ConfigUtils.getStorageLocationForServiceProperty(templatePreparationObject, FILE_SYSTEM_URI)
+                .map(location -> List.of(config(FILE_SYSTEM_URI, location.getValue())))
+                .orElseGet(() -> List.of(config(FILE_SYSTEM_URI, ConfigUtils.getDefaultFileSystemUri(templateProcessor, templatePreparationObject))));
+    }
+
```
smaniraju: As above I don't think default filesystem as hdfs needs to be overridden.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/profileradmin/ProfilerAdminCloudStorageRoleConfigProvider.java
subrat-mishra: We don't write data to the root path. The appending logic is inside the profiler application. I hope it should be fine. Please confirm.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/ConfigUtils.java
subrat-mishra: I thought except Ranger all other services write it to the local filesystem. I followed the same code as of Ranger. 
https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/ranger/RangerCloudStorageServiceConfigProvider.java#L45-L58
Basically we need the FileSystemURI as "hdfs:///" or "s3a://BUCKET". 
As the code was repeating across all the services, I moved it to ConfigUtil.

smaniraju: No, for Ranger this field is not initialized by CM, hence ranger has explicit initialization. The values in cloudbreak configproviders are only for overriding default CM values, what is the default value for this service as initialized by CM ? Default values which map to hdfs should be initialized in CM, if it is not being done can be raised as CM Jira.

PR #6175, opened 2019-09-10
CDPCP-679 - Remove users, groups, and group membership related changes

abhishekaypurohit:
changes related to users, groups, group memberships update and deletion. 

Closes CDPCP-674

@handavid @lacikaaa @keyki

PR #6182, opened 2019-09-10
CDPCP-679 - user and groups delete and update

abhishekaypurohit:
changes related to users, groups, group memberships update and deletion.

Closes CDPCP-674

@handavid @lacikaaa @keyki

Related to https://github.com/hortonworks/cloudbreak/pull/6175

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
@@ -173,6 +173,12 @@ public class FreeIpaClient {
         return (Group) invoke("group_add", flags, params, Group.class).getResult();
     }
 
+    public void deleteGroup(String group) throws FreeIpaClientException {
+        List<String> flags = List.of(group);
+        Map<String, Object> params = Map.of();
+        invoke("group_del", flags, params, Object.class);
+    }
+
     // TODO unpack response into something meaningful
     // NOTE: API may partially succeed/fail
     //ipa: INFO: Response: {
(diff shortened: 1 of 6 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -173,6 +173,12 @@ public Group groupAdd(String group) throws FreeIpaClientException {
         return (Group) invoke("group_add", flags, params, Group.class).getResult();
     }
 
+    public void deleteGroup(String group) throws FreeIpaClientException {
+        List<String> flags = List.of(group);
+        Map<String, Object> params = Map.of();
+        invoke("delete_group", flags, params, Object.class);
```
handavid: does this work? all the other ipa commands are in the form `noun_verb` like `user_del`. I would have expected this to be `group_del` instead of `delete_group`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -78,28 +81,60 @@ public static UsersStateDifference fromUmsAndIpaUsersStates(UsersState umsState,
...
+                groupsToBeRemoved.add(group);
+            }
+        });
+
+        return groupsToBeRemoved;
+
+
+    }
+
+    private static Set<FmsUser> getUsersToBeRemoved(Collection<String> users, Set<FmsUser> ipaStateUsers) {
```
handavid: I don't think this is right. I read this as:
Remove users that 1. exist in ipa, and 2. do not exist in ums.
This is because all UMS users will be in the internal group. 

I think we want logic that says:
Remove users that 1. exist in ipa, and 2. are part of the internal group in ipa, and 3. do not exist in ums

abhishekaypurohit: Good Catch, I have changed this.

PR #6183, opened 2019-09-10
CB-3463. Log identity is not filled if identity is set to null manually for distrox template

oleewere:
if you are setting `identities` field in cloud storage request (in distrox cluster template) to `null`, then log identity is not added to the identity list (not causing any issues with the UI or default flows with cb cli) -> main problem with this is if identities field is missing , generated go client will fill that as null (workaround is to set identities as empty [])


oleewere:
retest this please

PR #6186, opened 2019-09-11
Initial commit of data mart blueprint with support for Kudu

adembo:
This commit introduces a new blueprint which I've called the "active data
mart". The name is a reference to Active Data Warehouse, a term we often use
to describe a data warehouse that leverages Kudu for additional real-time
storage and query capabilities. The blueprint is based on the existing data
mart template, complete with custom Impala performance tuning. Besides the
obvious inclusion of Kudu, I've also added Spark (and a dependent YARN) for
eventual Kudu BDR support.

The topology itself consists of three masters (for HA) and four tablet
servers (for 3-4-3 replication). To minimize the effect of a failure, each
master and tablet server runs on its own node, which makes the minimal
cluster size seven nodes. The worker nodes are partitioned into a singleton
"coordinator" group running an Impala coordinator and an "executor" group
running the remaining workers; users who need larger clusters should scale
up the size of the executor group. The non-Kudu master roles are spread out
across the three master nodes to make more efficient use of those resources.

Effective use of this blueprint requires a few things:
1. Configuration of Impala and Kudu for hierarchical storage management
   (i.e. aging out of cold data to object storage). This isn't something we
   can automate due to its domain-specific nature.
2. Connectivity with another cluster responsible for ingest. In theory we
   could automate this if we had a better sense for other "cluster shapes"
   that handled ingest (via NiFi, Spark Streaming, etc).

To get all of this to work, I had to add a few temporary workarounds:
- Kudu's hybrid clock is disabled. The NTP configuration found on VMs
  deployed by Cloudbreak is really bad inasfar as fast synchronization. I'm
  working on an alternative configuration, but for the time being we need to
  sidestep the NTP dependency altogether. This also means that Kudu BDR
  doesn't work, because its snapshot scans require the hybrid clock.
- Kudu's on-disk directories are hardcoded to the first EBS volume attached
  to each VM by Cloudbreak. This should be removed when CB-3347 is fixed.

What has been tested?
- Deploying the template. OPSAPS-52447 is a blocking bug, but when deploying
  with a custom CM including the fix, it works.
- `kudu perf loadgen` initiated from a cluster host.
- Kudu backup initiated from a cluster host. This fails due to the
  aforementioned hybrid clock issues.

### DIFF
--- a/cloud-common/src/main/resources/application.yml
+++ b/cloud-common/src/main/resources/application.yml
@@ -169,6 +169,7 @@ cb:
                 CDP 1.1 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie=cdp-data-engineering-710;
                 CDP 1.1 - Data Engineering HA: Apache Spark, Apache Livy, Apache Zeppelin=cdp-data-engineering-ha-710;
                 CDP 1.1 - Data Mart: Apache Impala, Hue=cdp-data-mart-710;
+                CDP 1.1 - Operations Data Mart: Apache Impala, Hue, Apache Kudu, Apache Spark=cdp-op-data-mart-710;
                 CDP 1.1 - Operational Database: Apache HBase=cdp-opdb-710;
                 CDP 1.1 - Operational Database: Apache HBase (Custom)=cdp-opdb-custom-710;
                 CDP 1.1 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-710;
--- a/core/src/main/resources/defaults/blueprints/cdp-op-data-mart-710.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-op-data-mart-710.bp
@@ -1,0 +1,208 @@
+{
+  "description": "CDP 1.1 Operations Data Mart template with Apache Impala, Hue, Apache Kudu, and Apache Spark",
+  "blueprint": {
+    "cdhVersion": "7.1.0",
+    "displayName": "op-datamart-710",
+    "services": [
+      {
+        "refName": "hdfs",
+        "serviceType": "HDFS",
+        "roleConfigGroups": [
+          {
+            "refName": "hdfs-NAMENODE-BASE",
+            "roleType": "NAMENODE",
+            "base": true
+          },
+          {
+            "refName": "hdfs-SECONDARYNAMENODE-BASE",
+            "roleType": "SECONDARYNAMENODE",
+            "base": true
+          },
+          {
+            "refName": "hdfs-DATANODE-BASE",
+            "roleType": "DATANODE",
+            "configs" : [ {
+              "name" : "dfs_datanode_max_locked_memory",
+              "value" : "0",
+              "autoConfig" : false
+            } ],
+            "base": true
+          },
+          {
+            "refName": "hdfs-BALANCER-BASE",
+            "roleType": "BALANCER",
+            "base": true
+          }
+        ]
+      },
+      {
+        "refName": "kudu",
+        "serviceType": "KUDU",
+        "roleConfigGroups": [
+          {
+            "refName": "kudu-MASTER-BASE",
+            "roleType": "KUDU_MASTER",
+            "base": true
+          },
+          {
+            "refName": "kudu-TSERVER-BASE",
+            "roleType": "KUDU_TSERVER",
+            "base": true
+          }
+        ]
+      },
+      {
+        "refName": "yarn",
+        "serviceType": "YARN",
+        "roleConfigGroups": [
+          {
+            "refName": "yarn-RESOURCEMANAGER-BASE",
+            "roleType": "RESOURCEMANAGER",
+            "base": true
+          },
+          {
+            "refName": "yarn-NODEMANAGER-BASE",
+            "roleType": "NODEMANAGER",
+            "base": true
+          },
+          {
+            "refName": "yarn-JOBHISTORY-BASE",
+            "roleType": "JOBHISTORY",
+            "base": true
+          }
+        ]
+      },
+      {
+        "refName": "spark_on_yarn",
+        "serviceType": "SPARK_ON_YARN",
+        "roleConfigGroups": [
+          {
+            "refName": "spark_on_yarn-SPARK_YARN_HISTORY_SERVER-BASE",
+            "roleType": "SPARK_YARN_HISTORY_SERVER",
+            "base": true
+          },
+          {
+            "refName": "spark_on_yarn-GATEWAY-BASE",
+            "roleType": "GATEWAY",
+            "base": true
+          }
+        ]
+      },
+      {
+        "refName": "impala",
+        "serviceType": "IMPALA",
+        "serviceConfigs" : [ {
+          "name" : "impala_cmd_args_safety_valve",
+          "value" : "--cache_s3_file_handles=true"
+        } ],
+        "roleConfigGroups": [
+          {
+            "refName": "impala-IMPALAD-COORDINATOR",
+            "roleType": "IMPALAD",
+            "configs" : [ {
+              "name" : "impalad_specialization",
+              "value" : "COORDINATOR_ONLY"
+            }, {
+              "name" : "impala_hdfs_site_conf_safety_valve",
+              "value" : "<property><name>fs.s3a.block.size</name><value>268435456</value></property><property><name>fs.s3a.experimental.input.fadvise</name><value>RANDOM</value></property><property><name>fs.s3a.fast.upload</name><value>true</value></property>"
+            } ],
+            "base": false
+          },
+          {
+            "refName": "impala-IMPALAD-EXECUTOR",
+            "roleType": "IMPALAD",
+            "configs" : [ {
+              "name" : "impalad_specialization",
+              "value" : "EXECUTOR_ONLY"
+            }, {
+              "name" : "impala_hdfs_site_conf_safety_valve",
+              "value" : "<property><name>fs.s3a.block.size</name><value>268435456</value></property><property><name>fs.s3a.experimental.input.fadvise</name><value>RANDOM</value></property><property><name>fs.s3a.fast.upload</name><value>true</value></property>"
+            } ],
+            "base": false
+          },
+          {
+            "refName": "impala-STATESTORE-BASE",
+            "roleType": "STATESTORE",
+            "base": true
+          },
+          {
+            "refName": "impala-CATALOGSERVER-BASE",
+            "roleType": "CATALOGSERVER",
+            "base": true
+          }
+        ]
+      },
+      {
+        "refName": "hue",
+        "serviceType": "HUE",
+        "roleConfigGroups": [
+          {
+            "refName": "hue-HUE_SERVER-BASE",
+            "roleType": "HUE_SERVER",
+            "base": true
+          },
+          {
+            "refName": "hue-HUE_LOAD_BALANCER-BASE",
+            "roleType": "HUE_LOAD_BALANCER",
+            "base": true
+          }
+        ]
+      }
+    ],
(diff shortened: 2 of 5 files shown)


core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
```diff
@@ -0,0 +1,226 @@
...
+          },
+          {
+            "refName": "impala-IMPALAD-EXECUTOR",
+            "roleType": "IMPALAD",
+            "configs" : [ {
+              "name" : "impalad_specialization",
+              "value" : "EXECUTOR_ONLY"
+            }, {
+              "name" : "impala_hdfs_site_conf_safety_valve",
+              "value" : "<property><name>fs.s3a.block.size</name><value>268435456</value></property><property><name>fs.s3a.experimental.input.fadvise</name><value>RANDOM</value></property><property><name>fs.s3a.fast.upload</name><value>true</value></property>"
```
doktoric: with this property I think we are tightly coupled with AWS do we really need this ? I think the blueprint should not depend on cloud providers.
ccc @akanto @keyki

lajosrodek:
Once #6191 gets merged in, the BP should probably be renamed to `cdp-data-mart-active-700.bp` and added to `application.yml`. If this is to target CDH-7.1 instead, `"cdhVersion"` should be updated and the file name suffix changed to `-710` accordingly.

lajosrodek:
FYI #6187 added storage volume configs for Kudu.

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
```diff
@@ -0,0 +1,226 @@
+{
+  "description": "CDP 1.0 Active Data Mart template with Apache Impala, Hue, and Kudu",
```
granthenke: Should this mention Spark?

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
```diff
@@ -0,0 +1,226 @@
...
+          "value" : "--use_hybrid_clock=false"
+        } ],
+        "roleConfigGroups": [
+          {
+            "refName": "kudu-MASTER-BASE",
+            "roleType": "KUDU_MASTER",
+            "base": true,
+            "configs" : [ {
+              "name" : "fs_wal_dir",
+              "value" : "/hadoopfs/fs1/master"
```
granthenke: Should this be a different directory than the data dir?

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
```diff
@@ -0,0 +1,226 @@
...
+            } ]
+          },
+          {
+            "refName": "kudu-TSERVER-BASE",
+            "roleType": "KUDU_TSERVER",
+            "base": true,
+            "configs" : [ {
+              "name" : "fs_wal_dir",
+              "value" : "/hadoopfs/fs1/tserver"
```
granthenke: Should this be a different directory than the data dir?

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
granthenke: These configurations are already set in `cdp-data-mart.bp`. I think they were copied from there.

adembo: Yeah, I copied these straight from `cdp-data-mart.bp` without thinking much about it. Guess this is a question for @drorke?

adembo:
> Once #6191 gets merged in, the BP should probably be renamed to `cdp-data-mart-active-700.bp` and added to `application.yml`. If this is to target CDH-7.1 instead, `"cdhVersion"` should be updated and the file name suffix changed to `-710` accordingly.

Didn't know I had to change `application.yml`; thanks for that pointer.

What's the idea behind CB-3372? That with every CM/CDH release, a copy of _every_ blueprint needs to be made, named after the new release's version? Given the speed at which we're probably going to release, that'd quickly become a huge maintenance burden for making blueprint changes.

adembo:
> FYI #6187 added storage volume configs for Kudu.

Thanks; I'll remove the volume workarounds.

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
adembo: Yeah I guess so. Added it (but not YARN, which is just a dependency like HDFS).

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
adembo: Now irrelevant given that I've removed the hardcoded directories.

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
adembo: Now irrelevant given that I've removed the hardcoded directories.

cloud-common/src/main/resources/application.yml
```diff
@@ -169,6 +169,7 @@ cb:
                 CDP 1.1 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie=cdp-data-engineering-710;
                 CDP 1.1 - Data Engineering HA: Apache Spark, Apache Livy, Apache Zeppelin=cdp-data-engineering-ha-710;
                 CDP 1.1 - Data Mart: Apache Impala, Hue=cdp-data-mart-710;
+                CDP 1.1 - Active Data Mart: Apache Impala, Hue, Kudu, Spark=cdp-data-mart-active-710;
```
grishick: Why "Apache" in "Apache Impala", but no "Apache" in Kudu and Spark?

core/src/main/resources/defaults/blueprints/cdp-data-mart-active-710.bp
```diff
@@ -0,0 +1,200 @@
+{
+  "description": "CDP 1.1 Active Data Mart template with Apache Impala, Hue, Kudu, and Spark",
+  "blueprint": {
+    "cdhVersion": "7.1.0",
+    "displayName": "datamart-active",
+    "services": [
+      {
+        "refName": "hdfs",
```
grishick: Why do we need HDFS in this template?

adembo: For the same reason there's HDFS in the existing data mart template: CM apparently can't yet provision a cluster without an HDFS service. See OPSAPS-50807 for details.

cloud-common/src/main/resources/application.yml
adembo: My bad; I forgot Hue wasn't an Apache project and assumed the "Apache" modifier applied to all of the listed projects. I'll fix this.

grishick:
Why the comment about HSM? I think this blueprint can be effectively used for dimension tables, for time series data that can be dropped after it ages, and for time series data does not require querying across hot and cold partitions (e.g. a client application or a spark job can move cold partitions to S3).

adembo:
> Why the comment about HSM? I think this blueprint can be effectively used for dimension tables, for time series data that can be dropped after it ages, and for time series data does not require querying across hot and cold partitions (e.g. a client application or a spark job can move cold partitions to S3).

The comment about HSM restrictions was a reference to today's implementation (i.e. as per Grant's HSM blog post): the "boundaries" between hot and cold data are completely user-defined as they hinge on the table's partition schema. As such, there's no way for us to include "one size fits all" HSM automation as part of cluster deployment; users have to do all that work manually.

adembo:
In addition to addressing the feedback raised, I've updated the PR in several other ways:
1. There's now an additional commit to reconfigure NTP on provisioned nodes so that they synchronize time with FreeIPA faster. Without this I've seen it sometimes take hours to establish sync, which causes Kudu to die during startup later on.
2. Thanks to the fast time sync, we can use the hybrid clock once again.
3. At Krishna Maheshwari's request, I renamed "Active Data Mart" to "Operations Data Mart", which is apparently a well-known term in data warehousing circles.

granthenke:
+1 LGTM 👍

core/src/main/resources/defaults/blueprints/cdp-data-mart-active.bp
drorke: Some of the S3 related settings will be removed after some additional work to improve the defaults in the Impala code (likely to land in M8) but some S3 config tuning is likely to remain in the blueprint.  I think these are fairly benign because they'll essentially be ignored when running on Azure or in any other environment where the S3A file system type isn't used (I'm not aware of any scenario where having these in the safety valve would cause an error).

Another option would be to add them via code in a CB configprovider, assuming the configprovider had a way to know what cloud provider it was running on.

adembo:
Thanks for the review.

> * Are you planning to also create a cluster definition similar to this https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/defaults/clustertemplates/aws/aws-datamart-710.json but including the Kudu support?   Assuming the answer is yes, we should discuss preferred instance types, disk configurations, etc for the cluster definition to make sure we have something suitable for both Impala and Kudu.  Assuming the cluster definition comes in as a separate PR so we can discuss those details in the subsequent review.

I haven't looked into that yet (I was hoping to get this merged first to simplify my testing), but yeah, it's something I'm going to look into next.

> * Another comment slightly outside the scope of the current review:  We'll want to be careful that the logic in the Impala and Kudu volume config providers (as they continue to evolve) allows all the volume configs for both components to peacefully coexist.  I don't see any major issues at present, but in the future if both the configproviders were to focus critical I/O functions on a small NVMe volume (for example) we'd want to make sure to manage that carefully.

Makes sense. FWIW, we normally recommend a single SSD or NVMe device for Kudu's write-ahead log directory, but the storage needs of that device are modest (512 GB is usually sufficient). 

> * FYI now that https://jira.cloudera.com/browse/OPSAPS-50807 is fixed, I'm planning to remove HDFS from the (non Kudu) Data Mart blueprint for M8.   I don't know offhand if the additional services in this template have dependencies on HDFS (e.g. does Spark on YARN use the distributed cache?) but if not you might consider removing HDFS from this blueprint as well.

Good to know. We don't need HDFS either; we expect data to migrate out of Kudu and into object storage. I'll drop it in a subsequent PR once I've verified the functionality in testing.

core/src/main/resources/defaults/blueprints/cdp-op-data-mart-710.bp
```diff
@@ -0,0 +1,208 @@
+{
+  "description": "CDP 1.1 Operations Data Mart template with Apache Impala, Hue, Apache Kudu, and Apache Spark",
+  "blueprint": {
+    "cdhVersion": "7.1.0",
```
akanto: In the last minutes, CDP  bits had been renamed from 7.1.0 tp 7.0.1

akanto:
I had to revert this since cluster creation failed with 

'
The health test result for HOST_CLOCK_OFFSET has become bad: The host's NTP service is not synchronized to any remote server.
'

akanto:
Screenshot 2019-09-17 at 15 27 05

adembo:
Hmm, I didn't see this issue in all the clusters I provisioned. I just provisioned a new env and data lake (with a local CB running my changes) and I don't see any NTP sync issues on the two data lake hosts. I'll let them run and monitor that health test over time.

Is this bad cluster still running somewhere? I couldn't access its idbroker host's IP address so I figure it's already down. How many times did the health check fail? Was it flip-flopping back and forth or did it only fail a certain number of times? In what environment was this running? AWS? Somewhere else? Was this failure across the board or just observed in one cluster?

The more info you can provide the better; this is all still super opaque to me.

PR #6188, opened 2019-09-11
CDPD-3641 - Configure RDS instance for DAS.

harishjp:
Add DAS service config to configure RDS instance for DAS.

Closes CDPD-3641


harishjp:
Should I squash these changes and doing force push? Or is there an option to do the squash and merge?

PR #6189, opened 2019-09-11
CB-3364 Check activation status for all product parcels

bergerdenes:
Side change: mockito was not working with IDEA. Fixing.


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/ClouderaManagerDeployClientConfigListenerTask.java
```diff
@@ -66,6 +64,44 @@ public boolean checkStatus(ClouderaManagerCommandPollerObject clouderaManagerPol
...
+        ParcelsResourceApi parcelsResourceApi = clouderaManagerClientFactory.getParcelsResourceApi(apiClient);
+        return parcelsResourceApi.readParcels(stackName, "summary");
+    }
+
+    private List<ApiParcel> getNotActivatedOrMissingParcels(List<ClouderaManagerProduct> products, ApiParcelList parcels) {
+        List<ApiParcel> notActivated = new ArrayList<>(parcels.getItems().size());
+        products.forEach(product -> {
+            parcels.getItems()
+                    .stream()
+                    .filter(parcel -> product.getName().equals(parcel.getProduct())
```
gregito: just a note. I would extract this – and the `ifPresentOrElse` – logic to a separate method to make the whole stream chain more readable

bergerdenes: made more clean.

PR #6200, opened 2019-09-11
CB-3382 Support Azure environment subnets for redbeams

bhavanki:
The task of listing subnets in an environment is refactored out of
AllocateDatabaseServerV4RequestToDBStackConverter into the new
SubnetListerService class. For AWS, the class behaves as the previous
code, simply listing all the subnets by their IDs. For Azure, the class
modifies each subnet, expanding out its ID into the complete Azure
resource ID. Some of the necessary information is already available in
the environment response, but the subscription ID must be retrieved from
the credential associated with the environment.

SubnetChooserService is still in use, but it is now aware of the cloud
platform and supports Azure. Since there are no AZ restrictons in Azure,
it just chooses all of the subnets.

Other changes:

* The new CloudSubnet::withId method makes it easy to create a new
  subnet definition with an updated ID. SubnetListerService uses this.
* A new AzureNetworkView helper class formalizes how to retrieve subnets
  from an Azure network cloud model (done in AzureTemplateBuilder).
* Constants in AwsNetworkView are renamed to line up better with their
  use across the code base.
* Methods in the redbeams classes
  AllocateDatabaseServerV4RequestToDBStackConverter,
  SubnetChooserService, NetworkParameterAdder are renamed to be clearer
  on what they do, and less specific to AWS.
* The Credential DTO in redbeams now carries along the Azure
  subscription ID, placed there by CredentialService. Its constructors
  are updated for a more natural ordering of arguments.
* NetworkParameterAdder::addParameters (formerly addVpcParameters) no
  longer sets the "vpcId" parameter for Azure, as it turns out it was
  unused. Although the case for Azure is now a no-op, it must remain for
  Azure processing to work.

### DIFF
--- a/redbeams/src/main/java/com/sequenceiq/redbeams/service/network/SubnetListerService.java
+++ b/redbeams/src/main/java/com/sequenceiq/redbeams/service/network/SubnetListerService.java
@@ -1,0 +1,74 @@
+package com.sequenceiq.redbeams.service.network;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
+import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
+import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
+import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentNetworkResponse;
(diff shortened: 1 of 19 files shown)


redbeams/src/main/java/com/sequenceiq/redbeams/service/network/SubnetListerService.java
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            throw new RedbeamsException(String.format("Retrieved credential {} for Azure environment {} which lacks subscription ID",
+                credential.getName(), environmentCrn));
+        }
+    }
+
+    @VisibleForTesting
+    static CloudSubnet expandAzureResourceId(CloudSubnet meta, DetailedEnvironmentResponse environmentResponse, String subscriptionId) {
+
+        StringBuilder expandedId = new StringBuilder("/subscriptions/");
```
jadair-cloudera: Would probably be easier to read with String.format.

PR #6218, opened 2019-09-13
CB-3434 SDX should have a force delete options

foldik:
Datalake from now sends an optional _forced_ parameter from cloudbreak.


foldik:
Redbeams force delete needs to come into this PR as well.

PR #6219, opened 2019-09-13
CB-3438  Initialize Ranger Default Policy Groups

smaniraju:
Initialize Ranger Default Policy Groups for CM Version >= 7.0.1

Closes #CB-3438


smaniraju:
This PR needs CM 7.x.0 build of 1425569 or later, hence can be merged only after default CM Build Number Update. Have tested on CM 7.x.0 1425569 that the cluster creation is successful.

keyki:
So it looks like we need to go with CM 7.0.1 instead of 7.1.0. Probably we'll need to change the condition here as well.

smaniraju:
> So it looks like we need to go with CM 7.0.1 instead of 7.1.0. Probably we'll need to change the condition here as well.

Ok changed version check to 7.0.1 from 7.1.0

smaniraju:
Removed the "DONOTMERGE" label since default CM Build Number in cloudbreak has been updated to 1425569

smaniraju:
Changed base branch to rc-2.13 since CB-3438 is cdp1.1committed jira

keyki:
I think PR's will stuck because it requires latest changes from rc-2.13. Can you please rebase your branch to be on top of it?

PR #6223, opened 2019-09-13
CB-2842: add telemetry for sdx e2e tests

tiborpopovics:


oleewere:
it's valid if the telemetry request is null or empty, so would require a fix in the cloud storage validator as well. (so technically e2e tests should be ok), otherwise - good to have this as, i will probably need to extend this with other telemetry options.

PR #6226, opened 2019-09-13
 CB-3471 Add CSD to stack request and salt downloader

doktoric:


orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/init.sls
```diff
@@ -0,0 +1,14 @@
...
+    - template: jinja
+    - mode: 755
+
+download-csd:
+  cmd.run:
+    - name: /opt/salt/scripts/csd-downloader.sh
+    - require:
+      - file: /opt/salt/scripts/csd-downloader.sh
+    - shell: /bin/bash
+    - unless: test -f /var/csd_downloaded
```
keyki: This is checking if the file exists. In the script once we download 1 csd we create this file, but if the next one fails this state will not run again. We should only create the file at the end of the script once.

doktoric: done

PR #6239, opened 2019-09-16
CB-3330 CB-3657 Only query the fields needed for stack listing AND paged event history endpoint

daszabo:


core/src/main/java/com/sequenceiq/cloudbreak/converter/stack/StackListItemToStackViewResponseConverter.java
```diff
@@ -0,0 +1,38 @@
...
+        response.setStatus(item.getStackStatus());
+        ClusterViewResponse clusterResponse = new ClusterViewResponse();
+        BlueprintViewResponse blueprintResponse = new BlueprintViewResponse();
+        blueprintResponse.setName(item.getBlueprintName());
+        blueprintResponse.setStackType(item.getStackType());
+        blueprintResponse.setStackVersion(item.getStackVersion());
+        clusterResponse.setBlueprint(blueprintResponse);
+        clusterResponse.setStatus(item.getClusterStatus());
+        response.setCluster(clusterResponse);
+        response.setNodeCount(stackInstanceCounts.get(item.getId()));
```
gergopapi2: On 2.9 list view I think the unhealthy node count is also represented.

PR #6240, opened 2019-09-16
CB-3544 Port improved salt targeting from 2.9 to master

gergopapi2:


akanto:
could you rather add these vhanges to 2.13

PR #6244, opened 2019-09-16
CDPCP-743. Usersync adds workload administration groups

handavid:
The first commit updates the options and usermanagement proto files from thunderhead and regenerates the code. There are no manual changes in that commit.

The second commit modifies usersync to retrieve the workloadAdministrationGroupNames from the getRights response.

NOTE: I've only tested this with the mock ums so far. I intend to test this with the real ums tomorrow morning. I'll leave a comment here when I'm done with that.

### DIFF
--- a/auth-connector/src/generated/main/grpc/com/cloudera/thunderhead/service/usermanagement/UserManagementGrpc.java
+++ b/auth-connector/src/generated/main/grpc/com/cloudera/thunderhead/service/usermanagement/UserManagementGrpc.java
@@ -2695,6 +2695,191 @@ public final class UserManagementGrpc {
      }
      return getSetWorkloadSubdomainMethod;
   }
+  @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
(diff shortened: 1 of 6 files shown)


auth-connector/src/generated/main/grpc/com/cloudera/thunderhead/service/usermanagement/UserManagementGrpc.java
```diff
@@ -2695,6 +2695,191 @@ private UserManagementGrpc() {}
      }
      return getSetWorkloadSubdomainMethod;
   }
+  @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
```
giladwolff: just curious, why check in the generated file?

handavid: @giladwolff not sure. I'm just following the existing pattern there.

keyki: We usually publish the generated classes for the proto files as they don't really change. It's one less step to do gradle build to have that file.

PR #6254, opened 2019-09-17
Expose impala coordinator

tiborpopovics:


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -386,13 +386,20 @@ public void setServices(List<ApiClusterTemplateService> services) {
     Map<String, ServiceComponent> mapRoleRefsToServiceComponents() {
         return Optional.ofNullable(cmTemplate.getServices()).orElse(List.of()).stream()
                 .filter(service -> service.getRoleConfigGroups() != null)
-                .flatMap(service -> service.getRoleConfigGroups().stream().map(rcg -> Pair.of(service.getServiceType(), rcg)))
+                .flatMap(service -> service.getRoleConfigGroups().stream()
+                .filter(filterNonCoordinatorImpalaRole())
```
bergerdenes: please use lambda predicates, i.e. your function should return a boolean

keyki: I'm not exactly sure if this fix is in correct place. I'll need to check if there is another way to do this. This method is invoked at many places so it could affect other parts of the code besides Knox topology generation.

PR #6256, opened 2019-09-17
CB-3573 Remove the opdb custom blueprint from the UI

joshelser:
Per MattS, we only want one shape per cluster and we
expect users to modify the one we provide if they would
make some change to the instance type.

The expectation is that deleting the "OpDB custom" entry from application.yml, it will stop showing up in the "Custom" tab under the "Cluster Template" drop-down.


keyki:
Hi,  can you please send the PR against 2.13?

keyki:
Also if you want to get rid of the blueprint for existing users also then you need to create an SQL migration script also.

joshelser:
> you need to create an SQL migration script also

Ok. Is there a tool to create the migration (For rails, `rails generate migration`) or just do it by hand?

keyki:
> > you need to create an SQL migration script also
> 
> Ok. Is there a tool to create the migration (For rails, `rails generate migration`) or just do it by hand?

cbd migrate cbdb new "CB-3372 remove CDP 1.0 blueprints and templates" will generate the base template, but you can create it by hand also

joshelser:
Switched the base to rc-2.13 and added the migration.

I wasn't sure if the `name` column is being built from the description in the blueprint or the entry in application.yml -- that's why I made the `LIKE` statement match `%ustom`. In the application.yml, it was "Custom" but in the blueprint description it was "custom".

No associated cluster template for this one. Thanks for the review!

joshelser:
Thanks Krisztian.

PR #6270, opened 2019-09-18
CB-3490 Fix database creation on Azure PostgreSQL

bhavanki:
The following two changes enable database creation via JDBC on Azure
PostgreSQL servers.

First, the root username stored for service-managed Azure database
servers is updated after allocation completes to the required format
of <username>@<hostname>, the hostname being just the first portion of
the complete server hostname. Azure requires this format when logging
in, while RDS (AWS) does not. The implication is that user-managed /
registered database servers that happen to be Azure PostgreSQL servers
must be registered with root usernames in this same format.

Second, the additional PostgreSQL JDBC driver connection property
"sslmode" is added, set to the value "allow". This permits JDBC
connections to use TLS/SSL if the server requires it. By default, Azure
PostgreSQL servers do require TLS, while RDS instances do not.


redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/provision/handler/UpdateDatabaseServerRegistrationHandler.java
```diff
@@ -64,6 +65,12 @@ public void accept(Event<UpdateDatabaseServerRegistrationRequest> event) {
             dbServerConfig.setHost(dbHostname.getName());
             dbServerConfig.setPort(Integer.parseInt(dbPort.getName()));
 
+            if (CloudPlatform.valueOf(dbStack.getCloudPlatform()) == CloudPlatform.AZURE) {
```
gergopapi2: I would put this code to UserGeneratorService, because we already have classes there performing platform specific utilities, like password generation and so, and I think it is great to have these platform specific things collected to one location. Also, those generators I would at some point move into one package (3 files already).

bhavanki: Done!

PR #6271, opened 2019-09-18
CDPCP-732 - Added support for getting workload credentials for a user…

abhishekaypurohit:
… or machine user

FYI: This is internal PR to the feature branch. 

This PR is to fetch workload credentials for the CDP user. Once user sets his/her password for CDP, hashed password is stored. This change will get the hashed password and krb keys of the user and populate that in FMSUser object.  

TODO: Unit Test cases are pending. 

./gradlew :auth-connector:clean :auth-connector:build :freeipa:clean :freeipa:build
BUILD SUCCESSFUL in 1m 38s
100 actionable tasks: 36 executed, 64 up-to-date

Closes CDPCP-732

@toddlipcon @PhilippeLan @keyki @handavid @lacikaaa


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
```diff
@@ -451,6 +453,23 @@ public Account getAccount(String requestId, String userCrn) {
...
+     * @param requestId the request ID for the request
+     * @param userCrn   the user CRN
+     * @return the ActorWorkloadCredentialsResponse
+     */
+    public GetActorWorkloadCredentialsResponse getActorWorkloadCredentials(String requestId, String userCrn) {
+        checkNotNull(requestId);
+        checkNotNull(userCrn);
+        return newStub(requestId).getActorWorkloadCredentials(
+            GetActorWorkloadCredentialsRequest.newBuilder()
+                .setActorCrn(Crn.fromString(userCrn).getAccountId())
```
philipl: Should be possible to use the userCrn directly - assuming it's actually a crn.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -59,8 +61,9 @@ public SyncOperationStatus synchronizeUser(SynchronizeUserRequest request) {
             default:
                 throw new BadRequestException(String.format("UserCrn %s is not of resoure type USER or MACHINE_USER", userCrn));
         }
+        // TODO: Get the Request Id from header
```
handavid: requestId is already added to MDC by the `MDCContextFilter`.
In `UmsUsersStateProvider` we've pulled this out to a `Threadlocal`. 

The benefit of taking it out of the MDC is that the MDC is already propagated to all the relevant threads (i.e., flow, UserService async stuff). The drawback is that we are using the MDC for a non-logging purpose and it's behavior may change.

The right way to do this is to follow the pattern used by redbeams and add a `RequestIdFilter` and a `ThreadLocalRequestIdProvider` classes to the FMS or pull those classes out to the common module. We'll also need to look into the flow code to ensure that the request id is propagated to any threads that flow uses.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -22,40 +24,46 @@
...
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
-                                                            Set<String> userCrns, Set<String> machineUserCrns) {
+                                                            Set<String> userCrns, Set<String> machineUserCrns, Optional<String> requestId) {
         try {
-            Optional<String> requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
-            if (!requestIdOptional.isPresent()) {
-                requestIdOptional = Optional.of(UUID.randomUUID().toString());
-                LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestIdOptional);
+            final Optional<String> requestIdNewIfEmpty;
```
handavid: I suggest leaving the request id handling here alone until we can do this the right way (see my comment on UserV1Controller).

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -84,11 +92,38 @@
...
     }
 
+    private WorkloadCredential getCredentials(String actorCrn, String userCrn, Optional<String> requestId) {
+        GetActorWorkloadCredentialsResponse response = grpcUmsClient.getActorWorkloadCredentials(actorCrn, userCrn, requestId);
+        String hashedPassword = response.getPasswordHash();
+        List<ActorKerberosKey> keys = response.getKerberosKeysList();
+        return new WorkloadCredential(hashedPassword, keys);
+    }
+
+    private void populateWorkloadCredentialMap(String actorCrn, Map<String, WorkloadCredential> usersWorkloadCredentialMap, List<User> users, List<MachineUser> machineUsers, Optional<String> requestId) {
```
handavid: This method looks redundant with your changes to `handleUser`. You should remove it.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -22,40 +24,46 @@
...
-            List<MachineUser> machineUsers = machineUserCrns.isEmpty() ? umsClient.listAllMachineUsers(actorCrn, accountId, requestIdOptional)
-                    : umsClient.listMachineUsers(actorCrn, accountId, List.copyOf(machineUserCrns), requestIdOptional);
+            List<User> users = userCrns.isEmpty() ? grpcUmsClient.listAllUsers(actorCrn, accountId, requestIdNewIfEmpty)
+                : grpcUmsClient.listUsers(actorCrn, accountId, List.copyOf(userCrns), requestIdNewIfEmpty);
+
+            List<MachineUser> machineUsers = machineUserCrns.isEmpty() ? grpcUmsClient.listAllMachineUsers(actorCrn, accountId, requestIdNewIfEmpty)
+                : grpcUmsClient.listMachineUsers(actorCrn, accountId, List.copyOf(machineUserCrns), requestIdNewIfEmpty);
+
+
+//            // get users hashed password
```
handavid: This commented out code looks like it is irrelevant due to your changes to `handleUser`. You should remove it

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FmsUser.java
```diff
@@ -10,6 +10,8 @@
 
     private String lastName;
 
+    private WorkloadCredential creds;
```
handavid: we need to be mindful of how changes to the FmsUser will affect state difference calculations. I see that the creds are not added to the `equals` or `hashCode` methods. This will currently provide the correct behavior (since we use equality to compare against the ipa state and the ipa state won't include workload credentials) but this breaks equality between FmsUsers.
I think it would be better to update `equals` and `hashCode` here and perhaps move the `UsersState` models to include a map from name-->FmsUser rather than a set. We can update the difference code to use filters rather than `Sets.difference`

abhishekaypurohit: So, I looked into the equals method of FmsUser also, hashcode only consider username, and both names so that should not be an issue. Also methods like such:  getUsersToBeRemoved(Collection<String> umsStateUsers, Collection<String> ipaStateUsers) are String based collections. 

But, current implementation equals method and hashcode doesn't seems right. We are checking extra information like first name and last name. We should only check for workload username because that's the unique information in IPA and any two or more objects having the same workload username (username) must be considered as similar. Consider the scenario where same workloadusername obj csso_apurohit's first name is changed in IPA manually (however we don't recommend (allow) that but it is possible), in that case, the user is same (since same workload username) hashcode could be different.

To avoid above, I can add a new Map entry in UserState for user -> Wrokload Creds. Fow now that should be good.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
abhishekaypurohit: Done. Thanks.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: Removed. Thanks.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
abhishekaypurohit: Taking requestId from MDC, but I don't like unwanted global objects and specially ThreadLocal just for requestId. 
I have moved request Id in the controller and passing from there.

horadla23:
@abhishekaypurohit  It seems based on the error message you sent and the ums code (https://github.infra.cloudera.com/thunderhead/thunderhead/blob/9c5f825a1e58360c2584ee9fade995ca01ec0e6f/services/usermanagement/src/main/java/com/cloudera/thunderhead/service/usermanagement/server/UserManagementService.java#L4399), you should call this getActorWorkloadCredential ums method with internal actor CRN

handavid:
> @abhishekaypurohit It seems based on the error message you sent and the ums code (https://github.infra.cloudera.com/thunderhead/thunderhead/blob/9c5f825a1e58360c2584ee9fade995ca01ec0e6f/services/usermanagement/src/main/java/com/cloudera/thunderhead/service/usermanagement/server/UserManagementService.java#L4399), you should call this getActorWorkloadCredential ums method with internal actor CRN

You can get the internal actor CRN like this
`private static final String INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();`

handavid:
We need to update the MockUserManagementService so that it will respond to the new UMS calls when running locally.
Do you have any ideas what the appropriate mock behavior should be?

philipl:
> We need to update the MockUserManagementService so that it will respond to the new UMS calls when running locally.
> Do you have any ideas what the appropriate mock behavior should be?

I would imagine the right behaviour is to return a fixed set of values. You can use the ones from the unit test:

https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/usermanagement/src/test/java/com/cloudera/thunderhead/service/usermanagement/server/WorkloadCredentialsUtilsTest.java

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,64 @@
...
-            Optional<String> requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
-            if (!requestIdOptional.isPresent()) {
+            final Optional<String> requestIdOptional;
+            if (!Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString())).isPresent()) {
                 requestIdOptional = Optional.of(UUID.randomUUID().toString());
                 LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestIdOptional);
+            } else {
+                requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
             }
             requestIdThreadLocal.set(requestIdOptional.get());
```
handavid: you can remove the `requestIdThreadLocal` if you are passing in the `requestIdOptional` to each sub method.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -176,13 +182,42 @@ private SyncStatusDetail synchronizeStack(Stack stack, UsersState umsUsersState,
...
+        if (config.getIpauserobjectclasses() != null && config.getIpauserobjectclasses().contains(Config.CDP_USER_ATTRIBUTE)) {
+            // found the attribute, password sync can be performed
+            LOGGER.debug("Going for Credentials Sync, seems new image FreeIPA Server, having config attribute");
+
+            // Should sync for all users and not just diff. At present there is no way to identify that there is a change in password for a user
+            for (FmsUser u : umsUsersState.getUsers()) {
+                WorkloadCredential workloadCredential = umsUsersState.getUsersWorkloadCredentialMap().get(u.getName());
+                if (workloadCredential != null
+                    && !StringUtils.isEmpty(workloadCredential.getHashedPassword())
+                    && CollectionUtils.isEmpty(workloadCredential.getKeys())) {
```
handavid: should this be !isEmpty?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
```diff
@@ -15,10 +17,14 @@
 
     private Multimap<String, String> groupMembership;
 
-    public UsersState(Set<FmsGroup> groups, Set<FmsUser> users, Multimap<String, String> groupMembership) {
+    private Map<String, WorkloadCredential> usersWorkloadCredentialMap;
```
handavid: I think we should split this off from the UsersState somehow because it isn't something we would expect back from the ipa server. Maybe subclass UsersState to a UmsState class? There's other information from the UMS that we would want to hold on to, like the user crn --> workloadUsername mapping. We'll need that for fixing the single user sync. Creating a UmsState class would give us a place to hold any extra UMS information we need.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FmsUser.java
abhishekaypurohit: I have moved this to UserState and as we discussed, we should re-visit and go more as interfaces and child classes. Fix related to this can be independent of this PR.

abhishekaypurohit:
> It looks like you've got three things going on here:
> 
> * Updating the proto file and regenerating the java code
> * Some changes to improve logging
> * Adding handling for the new workload credentials
> 
> I would split those into three commits so that they are easier to review separately. (One PR is still fine as far as I'm concerned - I can still review each commit separately)



> > We need to update the MockUserManagementService so that it will respond to the new UMS calls when running locally.
> > Do you have any ideas what the appropriate mock behavior should be?
> 
> I would imagine the right behaviour is to return a fixed set of values. You can use the ones from the unit test:
> 
> https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/usermanagement/src/test/java/com/cloudera/thunderhead/service/usermanagement/server/WorkloadCredentialsUtilsTest.java

I have updated Mock UMS and related simple test also. The test can be improved further.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: We can address these refactoring issues in separate PRs.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
abhishekaypurohit: +1 
We need to re-visit and have more interfaces rather than concrete implementations also child-parent relationships. Can be a separate PR.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -167,6 +167,11 @@ public User userMod(String user, String key, Object value) throws FreeIpaClientE
         return (User) invoke("user_mod", flags, params, User.class).getResult();
     }
 
+    public User userMod(String user, Map<String, Object> params) throws FreeIpaClientException {
```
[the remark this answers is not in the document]
abhishekaypurohit: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -44,7 +48,13 @@
     public SyncOperationStatus synchronizeUser(SynchronizeUserRequest request) {
         String userCrn = checkUserCrn();
         String accountId = threadBaseUserCrnProvider.getAccountId();
-        LOGGER.debug("synchronizeUser() requested for user {} in account {}", userCrn, accountId);
+        Optional<String> requestId = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
+        if (!requestId.isPresent()) {
+            requestId = Optional.of(UUID.randomUUID().toString());
```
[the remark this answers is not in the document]
abhishekaypurohit: Yup, reverted.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
abhishekaypurohit: yup, fixed immediately after :)

abhishekaypurohit:
./gradlew :mock-caas:build  :freeipa:build
BUILD SUCCESSFUL in 41s
98 actionable tasks: 8 executed, 90 up-to-date

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,64 @@
...
-    private GrpcUmsClient umsClient;
+    private GrpcUmsClient grpcUmsClient;
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
                                                             Set<String> userCrns, Set<String> machineUserCrns) {
         try {
-            Optional<String> requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
-            if (!requestIdOptional.isPresent()) {
+            final Optional<String> requestIdOptional;
+            if (!Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString())).isPresent()) {
```
handavid: nit: I would prefer to create the `Optional` once instead of on both lines 52 and 56. This may be slightly streamlined by 
```
String requestId = MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString());
if (requestId == null) {
    requestId = UUID.randomUUID().toString();
    LOGGER..debug("No requestId found. Setting request id to new UUID [{}]", requestIdString);
}
final Optional<String> requestIdOptional = Optional.of(requestId);
```

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -506,6 +507,34 @@ public void listRoles(UserManagementProto.ListRolesRequest request, StreamObserv
         responseObserver.onCompleted();
     }
 
+    @Override
+    public void getActorWorkloadCredentials(com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetActorWorkloadCredentialsRequest request,
```
handavid: Does the response here actually allow users to log in? What's the password used to generate these values? We need to have some actual values here otherwise folks running locally won't be able to log into their clusters.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: I think that this one is small enough and relevant to the changes that you are making in this file. Just remove the declaration above and this line here. As far as I can see, those are the only two usages of the requestIdThreadLocal.

PR #6275, opened 2019-09-19
CB-3570 - added streaming parcels and csds for base image clusters

attilapalfi92:


doktoric:
I think we need ui change as well

PR #6284, opened 2019-09-19
Hide the DE HA blueprint/template from the CDP prod UI

doktoric:


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentNetworkResponse.java
```diff
@@ -69,8 +69,6 @@ public void setExistingNetwork(boolean existingNetwork) {
 
         private Set<String> subnetIds;
 
-        private String networkCidr;
```
keyki: I think DWX relies on this field.

PR #6301, opened 2019-09-23
CB-3617 validation for unique host template name

gregito:
CB-3617 validation for unique host template name


doktoric:
rebase pls

PR #6308, opened 2019-09-23
CB-3306. Add HDFS user to Yarn admin ACL

ajyoung:
The Telemetry Publisher component in Cloudera Manager runs as the HDFS user in this context and needs to have permission to view all jobs in order to collect the YARN logs from S3 when cloud storage is enabled. This change was already committed and tested on the cdp-data-engineering.bp blueprint but this change was lost when the file was renamed cdp-data-engineering-701.bp about a week ago.


keyki:
Thank you for the PR. I'm closing this as the other one for 2.13 will be merged back to master.

PR #6315, opened 2019-09-24
Refactor of Public Endpoint Management service related logic

biharitomi:
Changes:
 - SAN support in CSR generation, which is now disabled as PEM doesn't support SANs
 - Separate the PEM service related logic into an individual service in core and call the `cluster-dns-connector` module from there
 - Move the new service to be wired after host metadata setup to internal FQDNs be available

The only refactor that needs to be done is to create a new flow step for this logic.

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
@@ -28,13 +28,20 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import javax.security.auth.x500.X500Principal;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang3.StringUtils;
+import org.bouncycastle.asn1.pkcs.PKCSObjectIdentifiers;
 import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.asn1.x509.Extension;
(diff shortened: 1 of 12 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
```diff
@@ -149,11 +156,12 @@ public static X509Certificate cert(KeyPair identity, String publicAddress, KeyPa
...
-    public static PKCS10CertificationRequest csr(KeyPair identity, String publicAddress) {
+    public static PKCS10CertificationRequest csr(KeyPair identity, String externalAddress) {
         try {
-            String name = String.format("C=US, CN=%s, O=Cloudera", publicAddress);
-            LOGGER.info("Generate CSR with X.500 distinguished name: '{}'", name);
-            return generateCsrWithName(identity, name);
+            String name = String.format("C=US, CN=%s, O=Cloudera", externalAddress);
+            List<String> sanList = List.of();
+            LOGGER.info("Generate CSR with X.500 distinguished name: '{}' and list of SAN: '{}'", name, String.join(",", sanList));
+            return generateCsrWithName(identity, name, sanList);
```
gregito: I would accept some null check for the identity arg, since after passing it to the generateCsrWithName method that will use it directly. Possible NPE if for some reason the KeyPair is null (even if "it can't be null")

gregito: My philosophy is that the public-facing methods should validate for null since if someone changes the code in the upper layers and passes null instead of some value, that could break the current code in runtime with unnecessary NPEs.
Since it's "just" a philosophy, I didn't mark these messages as a change request just giving my opinion. It's your decision to make such a change.

PR #6316, opened 2019-09-24
CB-3699. Add cdp binary usage event to mdc context

oleewere:
cloudbreak_log_message is not used anymore for fluentd, we are producing json output on manowar, message field is used only for the full message.

usage collection requires only the message part of the log message in binary format, so im putting this to mdc context in a dynamic way (that is processed as a context field in EK index), so the new output would be like this for usage events:

logger: CDP_BINARY_USAGE_EVENT
context.requestId: \<request id\>
context.binaryUsageEvent: \<binary event\>


oleewere:
retest this please

PR #6317, opened 2019-09-24
CB-3661 Quick fix to avoid failure reports in case of every non-final status

horadla23:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/model/Status.java
```diff
@@ -37,4 +37,9 @@ public boolean isAvailable() {
     public boolean isStopPhaseActive() {
         return name().contains("STOP");
     }
+
+    public boolean isInProgress() {
```
akanto: As agreed this should be refactored

PR #6319, opened 2019-09-24
CDPSDX-935 Fix invalid cluster CRNs from deleting wrong vault secrets

jamisonbennett:
If a cluster CRN which was formatted in an invalid way was provided
to the cleanup cluster secrets API, previously all secrets for the
entire environment would be deleted. This fix throws exceptions when
and invalid CRN is provided.

Testing was performed manually using a local deployment of cloudbreak
and unit tests were added.

In https://github.com/hortonworks/cloudbreak/pull/6093, there was a discussion about whether or not we should throw exceptions for invalid CRNs. At the time, the thought was to require a valid CRN for the environment since it must match and to ignore invalid cluster CRNs since it was optional. The problem with ignoring the invalid CRN is that recursive deletes will now delete a vault path which does not have the CRN so it will delete the secrets from all CRNs. Additionally, I looked at the logs since we added the logging about a month ago, and the only times invalid CRNs were actually being used was by the test team. So this should not affect anyone that is using the API in the MOW-dev environment.

Closes #CDPSDX-935

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/VaultPathBuilder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/VaultPathBuilder.java
@@ -89,13 +89,8 @@ public class VaultPathBuilder {
     }
 
     public VaultPathBuilder withClusterCrn(String clusterCrn) {
-        this.clusterId = Optional.ofNullable(clusterCrn).map((String crnString) -> {
-            Crn crn = Crn.fromString(crnString);
-            if (crn == null) {
-                LOGGER.debug("An invalid cluster CRN was provided, it will be ignored.");
-                return null;
-            }
-            return crn.getResource();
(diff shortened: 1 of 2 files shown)


keyki:
Do you think this change is needed for CDP CP prod today? If so the branch would be rc-2.13

jamisonbennett:
> Do you think this change is needed for CDP CP prod today? If so the branch would be rc-2.13

While it is a bug that is in prod, there is currently no API consumers calling it. Also the caller of this API also has to have a bug (e.g. an invalidly formatted CRN) in order to trigger this bug. And if the bug is triggered, there is only a small window of time where the bug could be triggered and cause an issue (based on how the secrets are used for DWX).

So I think its probably not a high enough priority to push to prod without the full testing cycle. Right now the problem was only experienced by the test team which was validating the API. More consumers will start using this API more in CDP 1.2, so it just needs to be included with that.

keyki:
Thank you.

PR #6321, opened 2019-09-25
CB-3330 Add utc date time but in String format

sodre90:


structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/OperationDetails.java
```diff
@@ -101,6 +107,10 @@ public void setTimestamp(Long timestamp) {
         this.timestamp = timestamp;
     }
 
+    public String getUTCDateTime() {
+        return ZonedDateTime.ofInstant(Instant.ofEpochMilli(timestamp), ZoneOffset.UTC).format(DATE_TIME_FORMATTER);
```
akanto: Shouldn't we add an extra null check to it?

sodre90: We can add, there wasn't in the previous implementation, but it's better to check if it is null or not

akanto: Right, today morning we already run 3 different NPEs in different places.

PR #6325, opened 2019-09-25
CB-3553 ugly fix for SDX creation on brand new accounts, will have a …

lnardai:
- sdx was failing when trying to get cloudstorage recommendation from cloudbreak that required the blueprint
- will have a followup story for proper fix on master

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/FileSystemV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/FileSystemV4Controller.java
@@ -12,7 +12,9 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.filesystems.responses.FileSyste
 import com.sequenceiq.cloudbreak.api.util.ConverterUtil;
 import com.sequenceiq.cloudbreak.domain.FileSystem;
 import com.sequenceiq.cloudbreak.service.blueprint.BlueprintService;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/FileSystemV4Controller.java
```diff
@@ -35,6 +40,9 @@ public FileSystemParameterV4Responses getFileSystemParameters(
             String fileSystemType,
             Boolean attachedCluster,
             Boolean secure) {
+        //TODO: quick fix for SDX creation, followup for proper fix CB-3731
+        Workspace ws = workspaceService.getByIdForCurrentUser(0L);
+        blueprintService.getAllAvailableInWorkspace(ws);
```
sodre90: it's not ugly.. it's a spew.. are we sure there is no better place than a controller related to file system?!

lnardai: Validation message is correct both on the UI and CLI.
We've discussed this on slack, the sum is, Neither solution is better, this ensures that this endpoint is safe to use with default blueprints. Same logic is used during cluster creation, but basically we would need to do the same call before/after every call that is related to a blueprint.
The account bootstrap fix will take care of these problems implmeneted in the next ticket.

PR #6326, opened 2019-09-25
CB-3721 Made forced downscale possible even with unreachable Ambari

daszabo:


core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
```diff
@@ -67,13 +68,15 @@ public void accept(Event<DecommissionRequest> event) {
...
+            if (!skipAmbariDecomission(request)) {
+                LOGGER.debug("Downscale is not forced. Collecting nodes for decommission.");
+                Map<String, HostMetadata> hostsToRemove = ambariDecommissioner.collectHostsToRemove(stack, hostGroupName, hostNames);
+                if (!CollectionUtils.isEmpty(hostsToRemove)) {
+                    LOGGER.debug("Hosts about to be removed are present in Ambari. Proceeding with decommission.");
+                    executePreTerminationRecipes(stack, request.getHostGroupName(), hostsToRemove.keySet());
+                    decommissionedHostNames = ambariDecommissioner.decommissionAmbariNodes(stack, hostsToRemove);
+                }
             }
             result = new DecommissionResult(request, decommissionedHostNames);
```
sodre90: What I don't understand is... why are we going toward with decommissioned hosts? Wouldn't be better just return with the original hostnames? If I send 10 node to downscale and the decommissioner gives back only 8 nodes, then the other 2 nodes might not be removed

core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
```diff
@@ -67,13 +68,15 @@ public void accept(Event<DecommissionRequest> event) {
...
-            } else {
-                executePreTerminationRecipes(stack, request.getHostGroupName(), hostsToRemove.keySet());
-                decommissionedHostNames = ambariDecommissioner.decommissionAmbariNodes(stack, hostsToRemove);
+            Set<String> decommissionedHostNames = hostNames;
+            if (!skipAmbariDecomission(request)) {
+                LOGGER.debug("Downscale is not forced. Collecting nodes for decommission.");
+                Map<String, HostMetadata> hostsToRemove = ambariDecommissioner.collectHostsToRemove(stack, hostGroupName, hostNames);
+                if (!CollectionUtils.isEmpty(hostsToRemove)) {
+                    LOGGER.debug("Hosts about to be removed are present in Ambari. Proceeding with decommission.");
+                    executePreTerminationRecipes(stack, request.getHostGroupName(), hostsToRemove.keySet());
```
sodre90: Are we sure pretermination recipes are not necessary in case of forced downscale or in case of there is 0 hosts to remove on ambari side?

PR #6328, opened 2019-09-25
CB-3716 validation for env network and cloud platform relation

gregito:
CB-3716 validation for env network and cloud platform relation


doktoric:
why we have 7 commits here?

PR #6331, opened 2019-09-25
CDPCP-732 - Added support for getting workload credentials for a user…

abhishekaypurohit:
…… (#6271)

Squashed Commits: 

* CDPCP-732 - Added support for getting workload credentials for a user or machine user

* CDPCP-732 - updated getActorWorkloadCredentials since passing CRN itself

* CDPCP-732 - added support for internal actor since get workload credentials ums api requires internal crn

* CDPCP-732 - Added support for getting hased credentials from UMS, Internal Actor Support, Updated FreeIPA Client and Config

* CDPCP-732 - Added password sync to FreeIPA.
Added util class for Credentials and updated ASNUtil

* CDPCP-732 - Added ANS encoded support and calling user-mod for password, reverted bk freeipaclient factory for hard coded test purpose password

- https://github.com/hortonworks/cloudbreak/commit/888b2e2a8887def28a24e26efcd6d9ca5863733a

Closes CDPCP-732

@toddlipcon @PhilippeLan @handavid @keyki @lacikaaa @horadla23


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -112,6 +113,24 @@ public User getUserDetails(String actorCrn, String userCrn, Optional<String> req
...
+     *
+     * @param actorCrn  the CRN of the actor
+     * @param userCrn   the CRN of the user
+     * @param requestId an optional request Id
+     * @return the user associated with this user CRN
+     */
+    public GetActorWorkloadCredentialsResponse getActorWorkloadCredentials(String actorCrn, String userCrn, Optional<String> requestId) {
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
+            LOGGER.debug("Getting user's: {}, workload Credentials for using request ID {}", userCrn, requestId);
```
toddlipcon: this message formatting is a bit weird/non-grammatical

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -284,6 +289,13 @@ public int getUsernameLength() throws FreeIpaClientException {
         return config.getIpamaxusernamelength();
     }
 
+    public Config getConfig() throws FreeIpaClientException {
+        Map<String, Object> params = new HashMap<>();
+        params.put("all", Boolean.TRUE);
+        Config config = (Config) invoke("config_show", List.of(), params, Config.class).getResult();
```
toddlipcon: maybe just use ImmutableMap.of("all", Boolean.TRUE) here? also, does auto-boxing 'true' not work? (you may need to change the signature for 'invoke' to be Map<String, ? extends Object> instead of Map<String, Object>

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -284,6 +289,13 @@ public int getUsernameLength() throws FreeIpaClientException {
...
+        return config;
```
toddlipcon: no need for this extra line, just put return above

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,64 @@
...
-    private GrpcUmsClient umsClient;
+    private GrpcUmsClient grpcUmsClient;
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
                                                             Set<String> userCrns, Set<String> machineUserCrns) {
         try {
-            Optional<String> requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
-            if (!requestIdOptional.isPresent()) {
+            final Optional<String> requestIdOptional;
+            if (!Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString())).isPresent()) {
```
toddlipcon: this seems goofy to construct an optional and then check isPresent vs just checking getMdcContextMap.containsKey(...)

toddlipcon: or generally why not use:

Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()).orElseGet(()-> UUID.randomUUID().toString())

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -176,13 +182,44 @@ private SyncStatusDetail synchronizeStack(Stack stack, UsersState umsUsersState,
...
             return SyncStatusDetail.succeed(environmentCrn, "TODO- collect detail info");
         } catch (Exception e) {
             LOGGER.warn("Failed to synchronize environment {}", stack.getEnvironmentCrn(), e);
             return SyncStatusDetail.fail(environmentCrn, e.getLocalizedMessage());
         }
     }
 
+    private void processUsersWorkloadCredentials(String environmentCrn, UsersState umsUsersState, FreeIpaClient freeIpaClient) throws Exception {
+        Config config = freeIpaClient.getConfig();
+        if (config.getIpauserobjectclasses() != null && config.getIpauserobjectclasses().contains(Config.CDP_USER_ATTRIBUTE)) {
```
toddlipcon: style nit: invert this condition and return early if there is nothing to do. In general it's easier to read code with fewer levels of indentation

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -176,13 +182,44 @@ private SyncStatusDetail synchronizeStack(Stack stack, UsersState umsUsersState,
...
+                    && !CollectionUtils.isEmpty(workloadCredential.getKeys())) {
+
+                    // Call ASN_1 Encoder for encoding hashed password and then call user mod for password
+                    LOGGER.debug("Found Credentials for user {}", u.getName());
+                    String ansEncodedKrbPrincipalKey = ASNEncoder.getASNEncodedKrbPrincipalKey(workloadCredential.getKeys());
+                    Map<String, Object> params = new HashMap<>();
+                    List<String> attr = new ArrayList<>();
+                    attr.add("cdpHashedPassword=" + workloadCredential.getHashedPassword());
+                    attr.add("cdpUnencryptedKrbPrincipalKey=" + ansEncodedKrbPrincipalKey);
+                    params.put("setattr", attr);
```
toddlipcon: think above lines would be easier to read as just:

```
Map<String, Object> params = ImmutableMap.of("setattr", ImmutableList.of(
    "cdpHashedPassword=" + workloadCredential.getHashedPassword(),
    "cdpUnencryptedKrbPrincipalKey=" + ansEncodedKrbPrincipalKey));
```

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredential.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class WorkloadCredential {
+
+    private String hashedPassword;
+
+    private List<ActorKerberosKey> keys;
+
+    public WorkloadCredential(String hashedPassword, List<ActorKerberosKey> keys) {
+        this.hashedPassword = hashedPassword;
+        this.keys = keys;
```
toddlipcon: nit: can make both members final, and make an ImmutableList.copyOf(keys) here to be defensive and make sure the class is truly immutable

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Base64;
+import java.util.List;
+
+import org.bouncycastle.asn1.ASN1Encodable;
+import org.bouncycastle.asn1.DERInteger;
+import org.bouncycastle.asn1.DEROctetString;
+import org.bouncycastle.asn1.DERSequence;
+import org.bouncycastle.asn1.DERTaggedObject;
+
+public final class ASNEncoder {
```
toddlipcon: - add javadocs
- maybe a better name woudl be something like KrbKeySetEncoder

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final int TAG_0 = 0;
+
+    private static final int TAG_1 = 1;
+
+    private static final int TAG_2 = 2;
+
+    private static final int TAG_3 = 3;
+
+    private static final int TAG_4 = 4;
```
toddlipcon: do these constants really add much value? If you want the code to be more descriptive, perhaps a small utility method like:
```
static DERTaggedObject tag(int tag, ASN1Encodable obj) {
  return new DERTaggedObject(true, tag, obj);
}
```

would do the trick?

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static String getASNEncodedKrbPrincipalKey(List<ActorKerberosKey> keys) throws Exception {
+        if (keys == null || keys.size() < NUM_KEYS) {
+            throw new IllegalArgumentException("Invalid kerberos keys provided");
+        }
+
+        ASN1Encodable[] asn1Encodables = new ASN1Encodable[NUM_KEYS];
+
+        for (int i = 0; i < NUM_KEYS; i++) {
```
toddlipcon: why is this hard-coded to a particular number of keys? should be based on keys.size()

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+            ActorKerberosKey key = keys.get(i);
+            byte[] byteValue = Base64.getDecoder().decode(key.getKeyValue().getBytes("UTF-8"));
```
toddlipcon: use StandardCharsets.UTF_8 so that you don't have to worry about InvalidEncodingException or whatever

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private static DERSequence makeKrbKey(DERSequence salt, DERSequence key) {
+        return new DERSequence(new ASN1Encodable[] {
+            new DERTaggedObject(true, TAG_0, salt),
+            new DERTaggedObject(true, TAG_1, key)
+        });
+    }
+
+    public static String getASNEncodedKrbPrincipalKey(List<ActorKerberosKey> keys) throws Exception {
```
toddlipcon: can you make this exception list tighter? this is awfully broad

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
```diff
@@ -0,0 +1,77 @@
...
+        for (int i = 0; i < NUM_KEYS; i++) {
+            ActorKerberosKey key = keys.get(i);
+            byte[] byteValue = Base64.getDecoder().decode(key.getKeyValue().getBytes("UTF-8"));
+            asn1Encodables[i] = makeKrbKey(makeSalt(key.getSaltType(), key.getSaltValue()), makeEncryptionKey(key.getKeyType(), byteValue));
+        }
+
+        DERSequence krbKeys = new DERSequence(asn1Encodables);
+
+        DERSequence krbKeySet = new DERSequence(new ASN1Encodable[]{
+            new DERTaggedObject(true, TAG_0, new DERInteger(1)),
```
toddlipcon: think it's worth adding a comment on each line here for the name of the fields:

// attribute-major-vno
// attribute-minor-vno
// kvno
// mkvno

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -506,6 +507,34 @@ public void listRoles(UserManagementProto.ListRolesRequest request, StreamObserv
...
+
+        final int keyType17 = 17;
+        final int keyType18 = 18;
+        final int saltType = 4;
+        GetActorWorkloadCredentialsResponse.Builder respBuilder = GetActorWorkloadCredentialsResponse.getDefaultInstance().toBuilder();
+        respBuilder.addKerberosKeysBuilder(0)
+            .setSaltType(saltType)
+            .setKeyType(keyType17)
+            .setKeyValue("testKeyValue17")
+            .setSaltValue("NonIodizedGrainOfSalt")
```
toddlipcon: does this mock need to actually provide values that will work when pushed to a real IPA? if so, these clearly aren't going to work. You can use these values:
```
 'keys': [{'key': {'enctype': 18,
                   'keyvalue': b'\xbb?=i\xd8\x98\x89\x15\x98K#g\xc4\x98\xc8W'
                               b'\xfe\x9d\xb5T\xcck\xfeP\x89\t,\rYl\xe8\xc8'},
           'salt': {'salt': b'd1QmHyuRC8xW0jbf', 'type': 4}},
          {'key': {'enctype': 17,
                   'keyvalue': b"\x19\xdf\x8e\xc4W\x8fz0\xd9'\x13_"
                               b'\x9f\x93\t\xd1'},
           'salt': {'salt': b'VgUdOYXP4iuw1DRw', 'type': 4}}],
```

toddlipcon: and hashedpassword: {SSHA512}e1NTSEE1MTJ9ODk1SGNYSFFzZWIzS2dtTEdPYTZZRDcvSXJEM29Od0pqdGVlWjlhMmZvdTJsaTE5T240ejhFWDVyTTNiY0xCOHorRVY4Y05RckhyMApKbisrR1RXaG5YcFZZME5xVUcxYVRWRkpaelJ4UjFOMVZXaE9XRkJXVEdSTWIyTlZVVGR2Cg==

mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
```diff
@@ -61,4 +65,69 @@ public void testCreateWorkloadUsername() {
 
         Assert.assertEquals(expected, underTest.sanitizeWorkloadUsername(username));
     }
+
+    @Test
+    public void testgetActorWorkloadCredentials() {
```
toddlipcon: this test seems superfluous since it just duplicates exactly the code in the implementation

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
abhishekaypurohit: ImmutableMap related change I can do as separate PR since we need to change in every method call and scope for testing increases. For this PR, auto-boxing 'true' and return config can be changed. I kept xtra line for the breakpoint purpose to evaluate the values and forgot to remove that. Thanks.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: Thanks, Agreed, I should have checked only getMdcContextMap.containsKey(...). 

Also, the suggestion to use orElseGet is incompatible types because it requires Supplier<? extends String> however, orElseGet inferred to Optional<T>. 

Moreover, lambda calls in further down code require requestId. Lambda expects the final variable, hence this blank constant way.

freeipa/src/main/java/com/sequenceiq/freeipa/util/ASNEncoder.java
abhishekaypurohit: This way I cheat checkstyle.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
abhishekaypurohit: This is not going to be in real IPA, this is for unit tests and local uses. For real real IPA, it must me real ums.

mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
abhishekaypurohit: True, I have made it shorter.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
abhishekaypurohit: Done, this was for my breakpoint.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -284,6 +289,12 @@ public int getUsernameLength() throws FreeIpaClientException {
         return config.getIpamaxusernamelength();
     }
 
+    public Config getConfig() throws FreeIpaClientException {
+        Map<String, Object> params = new HashMap<>();
```
toddlipcon: didn't quite follow your comment earlier about this, but it seems to work fine if I change this code to:
```
    public Config getConfig() throws FreeIpaClientException {
        Map<String, Object> params = Map.of("all", true);
        return (Config) invoke("config_show", List.of(), params, Config.class).getResult();
    }
```
which is a bit more concise

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,65 @@
...
     @Inject
-    private GrpcUmsClient umsClient;
+    private GrpcUmsClient grpcUmsClient;
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
                                                             Set<String> userCrns, Set<String> machineUserCrns) {
         try {
-            Optional<String> requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
-            if (!requestIdOptional.isPresent()) {
+            final Optional<String> requestIdOptional;
```
toddlipcon: how about this concise assignment?

```
            final Optional<String> requestIdOptional = Optional.of(
                Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()))
                        .orElseGet(() -> UUID.randomUUID().toString()));
```

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeySetEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.bouncycastle.asn1.ASN1Encodable;
+import org.bouncycastle.asn1.DERInteger;
+import org.bouncycastle.asn1.DEROctetString;
+import org.bouncycastle.asn1.DERSequence;
+import org.bouncycastle.asn1.DERTaggedObject;
+
+public final class KrbKeySetEncoder {
+
+    private static final int TAG_0 = 0;
```
toddlipcon: if you need to have these to cheat checkstyle (ugh) then I think it would be better to at least name them something useful like TAG_MAJOR_VNO, TAG_MINOR_VNO, TAG_KVNO, etc corresponding to the field names.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeySetEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Base64;
+import java.util.List;
+
+import org.bouncycastle.asn1.ASN1Encodable;
+import org.bouncycastle.asn1.DERInteger;
+import org.bouncycastle.asn1.DEROctetString;
+import org.bouncycastle.asn1.DERSequence;
+import org.bouncycastle.asn1.DERTaggedObject;
+
+public final class KrbKeySetEncoder {
```
toddlipcon: still needs javadoc, per earlier comment

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeySetEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+            new DERTaggedObject(true, TAG_1, key)
+        });
+    }
+
+    public static String getASNEncodedKrbPrincipalKey(List<ActorKerberosKey> keys) throws IOException {
+        if (keys == null || keys.size() < NUM_KEYS) {
+            throw new IllegalArgumentException("Invalid kerberos keys provided");
+        }
+
+        ASN1Encodable[] asn1Encodables = new ASN1Encodable[NUM_KEYS];
```
toddlipcon: same question as before: why not use keys.size()? there's no reason this code has to care how many keys are encoded (it would be fine for us to have only one key, if we decide that AES256 is bad in the future or whatever)

mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
```diff
@@ -61,4 +65,44 @@ public void testCreateWorkloadUsername() {
...
     }
+
+    @Test
+    public void testGetActorWorkloadCredentials() {
+        GetActorWorkloadCredentialsRequest req = GetActorWorkloadCredentialsRequest.getDefaultInstance();
+        MockUMSServiceCallStreamObserverImpl<GetActorWorkloadCredentialsRequest, GetActorWorkloadCredentialsResponse> responseObserver
+            = new MockUMSServiceCallStreamObserverImpl();
+
+        underTest.getActorWorkloadCredentials(req, responseObserver);
+        Assert.assertNotNull("Response must not be null", responseObserver.getResponse());
```
toddlipcon: I think this mocking could be done more idiomatically using mockito like:

```
StreamObserver s = Mockito.mock(StreamObserver.class);
underTest.getActorWorkloadCredentials(req, s);
Mockito.verify(s).onNext(Matchers.notNull());
Mockito.verify(s).onCompleted();
Mockito.verify(s, Mockito.never()).onError(Mockito.any());
```

toddlipcon: (but I still think you could just remove this test entirely since it doesn't really do anything past duplicating the implementation)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: I think we want to keep the log statement to indicate that this request id isn't actually coming in from a request to the FMS.
```
.orElseGet(() -> {
    String requestId = UUID.randomUUID().toString();
    LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestIdOptional);
    return requestId;
}));
```

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,65 @@
...
 import com.sequenceiq.freeipa.service.freeipa.user.model.FmsUser;
 import com.sequenceiq.freeipa.service.freeipa.user.model.UsersState;
+import com.sequenceiq.freeipa.service.freeipa.user.model.UsersState.Builder;
+import com.sequenceiq.freeipa.service.freeipa.user.model.WorkloadCredential;
 
 @Service
 public class UmsUsersStateProvider {
     private static final Logger LOGGER = LoggerFactory.getLogger(UmsUsersStateProvider.class);
 
     private static ThreadLocal<String> requestIdThreadLocal = new ThreadLocal<>();
```
handavid: remove this line

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -14,52 +14,65 @@
...
-            if (!requestIdOptional.isPresent()) {
+            final Optional<String> requestIdOptional;
+            if (MDCBuilder.getMdcContextMap().containsKey(LoggerContextKey.REQUEST_ID.toString())) {
+                requestIdOptional = Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()));
+            } else {
                 requestIdOptional = Optional.of(UUID.randomUUID().toString());
                 LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestIdOptional);
             }
+
             requestIdThreadLocal.set(requestIdOptional.get());
```
handavid: remove this line

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -506,6 +507,16 @@ public void listRoles(UserManagementProto.ListRolesRequest request, StreamObserv
         responseObserver.onCompleted();
     }
 
+    @Override
+    public void getActorWorkloadCredentials(com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetActorWorkloadCredentialsRequest request,
+                                            io.grpc.stub.StreamObserver<com.cloudera.thunderhead.service.usermanagement
+                                                .UserManagementProto.GetActorWorkloadCredentialsResponse> responseObserver) {
+
+        GetActorWorkloadCredentialsResponse.Builder respBuilder = GetActorWorkloadCredentialsResponse.getDefaultInstance().toBuilder();
```
handavid: I think we need to have some value returned here. Otherwise  folks won't be able to log in to clusters created from a locally running FMS and CBD.
We'll also need to advertise the password that this response would correspond to to engineering.

mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
handavid: I agree that this test isn't necessary

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: Done.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
abhishekaypurohit: Added hardcoded impl back and added Docs to mention string value of hash.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -124,9 +124,9 @@ public User getUserDetails(String actorCrn, String userCrn, Optional<String> req
     public GetActorWorkloadCredentialsResponse getActorWorkloadCredentials(String actorCrn, String userCrn, Optional<String> requestId) {
         try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
             UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
-            LOGGER.debug("Getting user's: {}, workload Credentials for using request ID {}", userCrn, requestId);
+            LOGGER.debug("Getting user's: {}, workload credentials", userCrn);
             GetActorWorkloadCredentialsResponse response = client.getActorWorkloadCredentials(requestId.orElse(UUID.randomUUID().toString()), userCrn);
-            LOGGER.debug("User's: {}, workload Credentials retrieved for using request ID {}", userCrn, requestId);
+            LOGGER.debug("Got workload credentials for user:{}", userCrn);
```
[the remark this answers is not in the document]
abhishekaypurohit: I have have made changes for consistency as before getting log message

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
abhishekaypurohit: Done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
abhishekaypurohit: Done, both comments including log.

mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
abhishekaypurohit: Removed test.

abhishekaypurohit:
> just a couple of minor comments. The only serious issue I have with this is that the MockUserManagementService doesn't return a valid value. I think that this is needed to support local development (i.e., for folks running CBD)

Agreed and I have put back implementation.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeySetEncoder.java
abhishekaypurohit: Apologies, I missed out earlier. Added javadoc now.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeySetEncoder.java
abhishekaypurohit: Thanks. This is good idea. Added this.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
abhishekaypurohit: For ImmutableMap, we will revisit FreeIPAClient after this PR. For now, closing this conversation.

horadla23:
It seems I am late, but I am missing more unit tests from this change (for new code in UserService and UmsUsersStateProvider for example)

PR #6334, opened 2019-09-26
CB-3734 - Instantiating CM tasks instead of using them as services

attilapalfi92:
This change is necessary for two reasons:
- Task is a unit of work that can be passed to an executor to execute it. Therefore a task must be instantiable and should be able to store state. The executor in our case is the PollingService.
- For this task: https://jira.cloudera.com/browse/CB-3749


daszabo:
There is no need for the tasks to store state. For the ticket, a wrapper around the CM API and `@Retryable` on its method would be sufficient IMHO.

attilapalfi92:
@daszabo In this case each retry would be inside a single polling attempt of the PollingService. That would be kind of strange IMHO. Using statefulness of the tasks would make it possible to create a solution that better aligns with the PollingService and would result in a more simple execution of the code. Easier to debug too.

PR #6336, opened 2019-09-26
CB-3003 Determine S3Guard existence both in AWs and in local environments

bergerdenes:
A little surplus: I started tagging the DynamoDB tables belonging
to S3Guard


bergerdenes:
fixing NPE

PR #6343, opened 2019-09-26
EAR-11336 Add missing property to ambari.properties in case of external MySQL Ambari database

horadla23:


horadla23:
tested on latest rc-2.9, external mysql database, created hdp3.1, hdp2.6 and hdf3.3 clusters successfully

PR #6348, opened 2019-09-27
Configure Hue knox_proxyhosts with external domain if available

biharitomi:
### Please do not merge, it is not a working solution from Hue/CM side!


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hue/HueConfigProvider.java
```diff
@@ -72,4 +81,30 @@ protected DatabaseType dbType() {
     protected List<ApiClusterTemplateConfig> getRoleConfigs(String roleType, TemplatePreparationObject source) {
         return List.of();
     }
+
+    private void configureKnoxProxyHostsServiceConfig(TemplatePreparationObject source, List<ApiClusterTemplateConfig> result) {
+        GatewayView gateway = source.getGatewayView();
+        GeneralClusterConfigs generalClusterConfigs = source.getGeneralClusterConfigs();
```
bergerdenes: you could push down these to the .externalFQDNShouldConfigured()
otherwise LGTM

PR #6350, opened 2019-09-27
CM http 500 mock test added and CM pollers are http 500 tolerant

attilapalfi92:


attilapalfi92:
Integration tests seem to be good on my local machine so I'll have to do a deeper investigation why are failing on Jenkins. :'(

attilapalfi92:
Fix for the int tests is in the 3rd commit.

PR #6351, opened 2019-09-27
Send operation requests to GCP in batch tasks and wait for the instances as last step

keyki:


cloud-common/src/main/java/com/sequenceiq/cloudbreak/common/type/ResourceType.java
```diff
@@ -64,4 +66,8 @@
     public CommonResourceType getCommonResourceType() {
         return commonResourceType;
     }
+
+    public static boolean isInstanceResource(ResourceType resourceType) {
+        return List.of(GCP_INSTANCE, OPENSTACK_INSTANCE, MOCK_INSTANCE).contains(resourceType);
```
bergerdenes: make the list private static final

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpNetworkInterfaceProvider.java
```diff
@@ -76,7 +76,7 @@ private String getInstanceNamePrefix(List<CloudResource> instances) {
                 request.setPageToken(response.getNextPageToken());
             } while (response.getNextPageToken() != null);
         } catch (IOException e) {
-            LOGGER.debug("Error dunging metadata collection", e);
+            LOGGER.debug("Error dunging instance collection", e);
```
bergerdenes: during?

keyki: It's not mine, but I'll fix it.

cloud-common/src/main/java/com/sequenceiq/cloudbreak/common/type/ResourceType.java
keyki: This is an enum class. Moving the list to a new class is not ideal I think.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpNetworkInterfaceProvider.java
keyki: Fixed.

cloud-common/src/main/java/com/sequenceiq/cloudbreak/common/type/ResourceType.java
keyki: Fixed.

PR #6352, opened 2019-09-27
CB-3760 Add null check to IT SDX DTO getAll methods

bhavanki:


aszegedi:
@bhavanki thank you for your pointers on this! This was a great catch! @afarsang please review our changes and let me know if I can improve or need to fix this!

PR #6356, opened 2019-09-28
CB-3648: improve GCP instance metadata collection

tiborpopovics:


keyki:
We'll test and merge once we have the quota limits.

PR #6360, opened 2019-09-30
CB-3664 Cover masssive scaling and stop start with mock

afarsang:
Cover masssive scaling and stop start with mock 
currently added with low numbers

Closes 3664

### DIFF
--- a/cloud-mock/src/main/java/com/sequenceiq/cloudbreak/cloud/mock/MockInstanceConnector.java
+++ b/cloud-mock/src/main/java/com/sequenceiq/cloudbreak/cloud/mock/MockInstanceConnector.java
@@ -39,6 +39,14 @@ public class MockInstanceConnector implements InstanceConnector {
             CloudVmInstanceStatus instanceStatus = getVmInstanceStatus(authenticatedContext, instance, "start");
             cloudVmInstanceStatuses.add(instanceStatus);
         }
+
+        MockCredentialView mockCredentialView = mockCredentialViewFactory.createCredetialView(authenticatedContext.getCloudCredential());
+        LOGGER.info("start instance statuses to mock spi, server address: " + mockCredentialView.getMockEndpoint());
+        try {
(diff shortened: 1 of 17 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/mock/DefaultModel.java
```diff
@@ -122,4 +136,26 @@ public void terminateInstance(Map<String, CloudVmMetaDataStatus> instanceMap, St
         CloudVmMetaDataStatus cloudVmMetaDataStatus = new CloudVmMetaDataStatus(cloudVmInstanceStatus, vmMetaDataStatus.getMetaData());
         instanceMap.put(instanceId, cloudVmMetaDataStatus);
     }
+
+    public void stopAllInstances() {
```
gregito: maybe the stopAllInstances and startAllInstances methods could call a more generic third - private - method, since their bodies are almost the same

cloud-mock/src/main/java/com/sequenceiq/cloudbreak/cloud/mock/MockInstanceConnector.java
```diff
@@ -37,6 +37,14 @@
             CloudVmInstanceStatus instanceStatus = new CloudVmInstanceStatus(instance, InstanceStatus.CREATED);
             cloudVmInstanceStatuses.add(instanceStatus);
         }
+
+        MockCredentialView mockCredentialView = mockCredentialViewFactory.createCredetialView(authenticatedContext.getCloudCredential());
+        LOGGER.info("stop instance statuses to mock spi, server address: " + mockCredentialView.getMockEndpoint());
```
gregito: "stop instance statuses". maybe should be "start instance statuses"?

afarsang: yepp

PR #6369, opened 2019-10-01
CB-3734 fix polling issue during decommission & CB-1659 check connection issue during startup

keyki:


keyki:
Wait. There is 1 more issue with stop polling..

PR #6383, opened 2019-10-02
CB-3724 Implement statusreason for UNHEALTHY status on 2.9

bbihari:


daszabo:
Will the API part come in a different PR?

PR #6384, opened 2019-10-02
CB-3827 YARN E2E fail fix

bergerdenes:
S3 cloudstorage creation in TelemetryTestDto should depend on
cloud provider type

Remove mistakenly committed file (application.yml)


integration-test/application.yml
```diff
@@ -1,438 +0,0 @@
-spring:
-    freemarker:
-        checkTemplateLocation: false
-
-# The logfiles will be created in this directory, LOG_PATH system property will be set and can be used in logback.xml
-# http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-logging.html#boot-features-logging-file-output
-logging:
-    path: ${integrationtest.outputdir}
-
-integrationtest:
```
attilapalfi92: These default values are required for e2e tests to work

attilapalfi92: application.yml is part of the integration-test project

attilapalfi92: also necessary for mock tests

bergerdenes: the file itself was committed by mistake. the correct one resides under src/main/resources

PR #6389, opened 2019-10-02
Configure VMs with cloud provider NTP servers

adembo:
The time synchronization status quo as of commit 909127e is that
provisioned VMs will use their out-of-the-box time service and
configuration as-is. For CDP CentOS 7 VMs, this means using chronyd to sync
with public NTP pool servers on the Internet.

Not all deployments will have Internet access, however. In fact, we've seen
some sporadic instances[1] where the NTP port appears to be blocked. To fix
this, let's reconfigure the VMs to incorporate whatever time sync service is
prescribed by the relevant cloud provider. Given our exclusive use of
CentOS 7 VMs I only bothered to do so for chronyd.

Note: I've only tested AWS deployments. Azure is especially frustrating
since the recommendation is to forgo NTP-based sync in favor of their Linux
Integration Services, so I've held off on that.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/ntp/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/ntp/init.sls
@@ -1,0 +1,37 @@
+{% if salt['file.file_exists' ]('/etc/chrony.conf') %}
+
+# Restart chronyd if its configuration file changes, but only if it was actually
+# running from the get go.
+chronyd:
+  service.running:
+    - watch:
+      - file: /etc/chrony.conf
+    - onlyif:
+      - service chronyd status
+
+# Add the cloud provider's NTP server to chronyd's configuration. We'll still
+# consider whatever servers chronyd was configured with out of the box; this
+# just gives preference to the cloud native one.
+add_cloud_platform_time_server_to_chrony:
+  file.append:
+    - name: /etc/chrony.conf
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/ntp/init.sls
```diff
@@ -0,0 +1,37 @@
+{% if salt['file.file_exists' ]('/etc/chrony.conf') %}
+
+# Restart chronyd if its configuration file changes, but only if it was actually
```
keyki: We should include a check if we're not on YCloud (Yarn)

orchestrator-salt/src/main/resources/salt/salt/top.sls
```diff
@@ -8,6 +8,7 @@ base:
     - recipes.runner
     - fluent
     - metering
+    - ntp
```
keyki: We should include a check if we're not on YCloud (Yarn) and skip all of these

adembo: I considered doing that, but then I figured that when running on YCloud, either chrony won't be installed, or chronyd won't be running (because the containers shouldn't have CAP_SYS_TIME). Do you still think I should add an explicit check?

keyki: In that case it's fine. I re-triggered the IT

PR #6398, opened 2019-10-03
Merge 2.13 to 2.14

oleewere:


oleewere:
it it goes through, i will merge rc-2.13 to rc-2.14 (not by this PR directly, but with the same commits)

PR #6399, opened 2019-10-03
CB-3777 Set ScmParams.CDP_ENV to PUBLIC_CLOUD

gergopapi2:


keyki:
Can you send against 2.14 please?

PR #6400, opened 2019-10-03
CB-3822 fix chrony.conf on Azure

keyki:


orchestrator-salt/src/main/resources/salt/salt/ntp/init.sls
```diff
@@ -15,23 +17,7 @@ chronyd:
...
-{% if salt['pillar.get']('platform') == 'AWS' %}
-# https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/
-    - text: |
-
-        server 169.254.169.123 prefer iburst
-
-{% elif salt['pillar.get']('platform') == 'GCP' %}
-# https://cloud.google.com/compute/docs/instances/managing-instances#configure_ntp_for_your_instances
     - text: |
-
```
adembo: FWIW I left this empty line here on purpose, so that there's some separation between our addition to /etc/chrony.conf and the already existing last line of the file.

keyki: Added back the new line

PR #6403, opened 2019-10-03
DISTX-327 - Get all users from FreeIPA those are requested.

abhishekaypurohit:
This change is for left out users. In case, if the set of users are passed and the call is not for "user sync all" then, if one of the pre-existing user's access is being changed in CDP, (removed user from env), then handleUser method will filter-out that user and the user will not be removed in IPA.

Currently, we are adding a new map in UserState but later we should have something like "UserSyncRequestContext" and all needed state should be composite in that including user state, ipa state, and additional information. 

Closes DISTX-327

@handavid @keyki @lacikaaa

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
@@ -19,12 +19,16 @@ public class UsersState {
 
     private Map<String, WorkloadCredential> usersWorkloadCredentialMap;
 
+    private Set<FmsUser> requestedWorkloadUsers = new HashSet<>();
+
     public UsersState(
-        Set<FmsGroup> groups, Set<FmsUser> users, Multimap<String, String> groupMembership, Map<String, WorkloadCredential> usersWorkloadCredentialMap) {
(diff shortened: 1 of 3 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
```diff
@@ -78,7 +88,12 @@ public void addWorkloadCredentials(String userName, WorkloadCredential creds) {
         }
 
         public UsersState build() {
-            return new UsersState(fmsGroups, fmsUsers, groupMembership, workloadCredentialMap);
+            return new UsersState(fmsGroups, fmsUsers, groupMembership, workloadCredentialMap, requestedWorkloadUsers);
+        }
+
+        public void addWorkloadUsername(FmsUser user) {
```
handavid: nit: change the name to `addRequestedWorkloadUsername` or something to match the internal field name

abhishekaypurohit: Fixed.

keyki:
Closing in favour of: https://github.com/hortonworks/cloudbreak/pull/6405

PR #6414, opened 2019-10-04
CB-2682 Initialize SMM Config With CM FQDN

smaniraju:
Initialize SMM Config With CM FQDN instead of CM IP Address.

Closes #CB-2682

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/smm/StreamsMessagingManagerServiceConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/smm/StreamsMessagingManagerServiceConfigProvider.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.c
 
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 
 import com.cloudera.api.swagger.model.ApiClusterTemplateConfig;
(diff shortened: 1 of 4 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/smm/StreamsMessagingManagerServiceConfigProvider.java
```diff
@@ -20,7 +20,7 @@
     @Override
     public List<ApiClusterTemplateConfig> getServiceConfigs(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
         List<ApiClusterTemplateConfig> configs = Lists.newArrayList(
-                config("cm.metrics.host", source.getGeneralClusterConfigs().getClusterManagerIp()),
+                config("cm.metrics.host", source.getGeneralClusterConfigs().getExternalFQDN()),
```
biharitomi: It could be null if the certification generation fails due to any reason. I think that case should also be handled here.

smaniraju: If it is null, should an exception be thrown, since there isn't any default CM FQDN ? 
CM IP Address does not work for this property.

smaniraju: As discussed, If externalFQDN is null, initialized the "cm.metrics.host" to CM IP as per earlier implementation.

PR #6433, opened 2019-10-08
CB-3804 Allow multiple host repair.

foldik:
- Previously datalake only allowed one host to repair now, it can handle multiple hosts as well.
- Previously during multi host repair when a gateway and a core node were unhealthy the core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/ClusterRepairFlowEventChainFactory.java would have generate an incorrect flow where it adds a cluster downscale event to the flow for thee idbroker host. This is problematic because during this process the master might not be available so that the salt's tearDown wont work.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/ClusterRepairFlowEventChainFactory.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/ClusterRepairFlowEventChainFactory.java
@@ -5,12 +5,11 @@ import static com.sequenceiq.cloudbreak.core.flow2.stack.downscale.StackDownscal
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedDeque;
-import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/ClusterRepairFlowEventChainFactory.java
```diff
@@ -74,106 +72,165 @@ public String initEvent() {
...
                 hostNames.size(), ScalingType.UPSCALE_TOGETHER, Sets.newHashSet(hostNames), singlePrimaryGateway,
-                kerberosSecured, event.accepted(), singleNodeCluster, cmType));
+                kerberosSecured, event.accepted(), singleNodeCluster, cmType);
     }
 
     private boolean isKerberosSecured(Stack stack) {
         return kerberosConfigService.isKerberosConfigExistsForEnvironment(stack.getEnvironmentCrn(), stack.getName());
     }
+
+    private static class RepairConfig {
```
lnardai: Nice!

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -449,11 +449,20 @@ public void updateHostMetadata(Long clusterId, Map<String, List<String>> hostsPe
...
                                 .forEach(hostName -> {
-                                    HostMetadata hostMetadataEntry = new HostMetadata();
-                                    hostMetadataEntry.setHostName(hostName);
-                                    hostMetadataEntry.setHostGroup(hostGroup.get());
-                                    hostMetadataEntry.setHostMetadataState(hostMetadataState);
-                                    hostGroup.get().getHostMetadata().add(hostMetadataEntry);
+                                    Optional<HostMetadata> hostMetadata = hostGroup.get().getHostMetadata()
+                                            .stream()
+                                            .filter(hm -> hostName.equals(hm.getHostName()))
+                                            .findFirst();
```
lnardai: You could continue with ifPresent(...).orElse(...)

PR #6434, opened 2019-10-08
CB-3791 Begin enabling Azure datalakes (no external db servers yet)

bhavanki:
A datalake / SDX cluster may now be created in an Azure environment,
although not with an external database server, yet.

First, the subnet checks in RdsWaitHandler now only apply to AWS. They
are unnecessary for Azure. With this change alone, it is possible to
create an Azure datalake without having redbeams create an external
database server.

SDX templates are introduced for Azure, to go along with the existing
ones for AWS. Currently, the templates are identical, each calling for a
GP_Gen5_2 instance for the database server. The map of templates in
DatabaseService was keyed only by SDX cluster shape, but is now keyed by
both cluster shape and cloud platform.

TLS is disabled for new Azure database servers. This is done because the
Cloudera Manager code which is responsible for creating databases on a
database server does not support connecting over TLS. (RDS instances in
AWS already do not use TLS.)

Usernames for Azure databases are now stored in the format
'username'@'short-hostname', as required to connect to them specifically
by Azure. A pattern in RedbeamsDbServerConfigurer is responsible for
detecting Azure database servers by their domain names.

Accordingly, queries issued by the salt script init_db_remote.sh are
sure to strip the @'short-hostname' suffix from the database username
and admin username before using them in query contents.

Also in init_db_remote.sh, an additional SQL command is inserted which
grants the role of the new database's user to the admin user. This is
necessary for the subsequent ALTER SCHEMA command to succeed, at least
in Azure PostgreSQL.

SdxService does not yet enable Azure database server use by default,
because at this time use of one causes failure during Cloudera Manager's
First Run command, while it attempts to create Hive metastore tables.

The following additional changes are not related to Azure specifically.

* The core cloudbreak service now correctly creates and configures a
  unique username and password for each remote database. Before this, it
  would reuse the administrative username and password for the entire
  database server.
* The GRANT ALL PRIVILEGES SQL command in the salt scripts init_db.sh
  and init_db_remote.sh had the database and username arguments
  reversed. (This wasn't a problem in practice, because the two are
  normally identical.) They have been switched to their correct
  positions.
* The salt pillar key "remote_db_url" has been renamed to
  "remote_db_host", since the value is in fact only a host name.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
@@ -5,7 +5,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -20,14 +19,14 @@ import com.google.common.collect.Lists;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.DatabaseVendor;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.ResourceStatus;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.database.base.DatabaseType;
-import com.sequenceiq.cloudbreak.common.database.DatabaseCommon;
-import com.sequenceiq.cloudbreak.common.database.DatabaseCommon.JdbcConnectionUrlFields;
 import com.sequenceiq.cloudbreak.domain.Blueprint;
 import com.sequenceiq.cloudbreak.domain.RDSConfig;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.Cluster;
 import com.sequenceiq.cloudbreak.service.cluster.ClusterService;
+import com.sequenceiq.cloudbreak.service.secret.service.SecretService;
 import com.sequenceiq.cloudbreak.util.PasswordUtil;
+import com.sequenceiq.redbeams.api.endpoint.v4.databaseserver.responses.DatabaseServerV4Response;
 
 public abstract class AbstractRdsConfigProvider {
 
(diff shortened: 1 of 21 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
```diff
@@ -42,66 +41,97 @@
...
             if (rdsConfig.getStatus() == ResourceStatus.DEFAULT && rdsConfig.getDatabaseEngine() != DatabaseVendor.EMBEDDED) {
                 Map<String, Object> postgres = new HashMap<>();
-                String db = getDb();
                 if (dbServerConfigurer.isRemoteDatabaseNeeded(cluster)) {
-                    JdbcConnectionUrlFields parsedUrl = dbCommon.parseJdbcConnectionUrl(rdsConfig.getConnectionURL());
-                    postgres.put("remote_db_url", parsedUrl.getHost());
-                    postgres.put("remote_db_port", parsedUrl.getPort());
-                    postgres.put("remote_admin", rdsConfig.getConnectionUserName());
-                    postgres.put("remote_admin_pw", rdsConfig.getConnectionPassword());
+                    DatabaseServerV4Response dbServerResponse = dbServerConfigurer.getDatabaseServer(cluster.getDatabaseServerCrn());
```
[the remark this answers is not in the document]
bhavanki: You're right, there is a prior call in the control flow. This call is in order to get the administrative username and password of the server, because that information is not saved in the RDSConfig object that is created from the call in `RedbeamsDbServerConfigurer`.

I'll see if there is some refactoring that can be done so that this second call can be eliminated.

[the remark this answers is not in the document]
bhavanki: Well, I did a bunch of refactoring, but I don't think it will help. I believe that, after the first call site, the data is only preserved in the core database, and then retrieved from there at the second call site. (I cannot tell which one comes first, but it does seem they are very far apart.) So, eliminating the second call requires adding the server information to RDSConfig. I would rather have that done as a separate task to improve performance, than try to do it along with all the other changes here.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/init_db_remote.sh
```diff
@@ -6,18 +6,23 @@
...
+    PGPASSWORD={{ values['remote_admin_pw'] }} createdb --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} {{ values['database'] }}
+    echo "CREATE USER $username WITH PASSWORD '{{ values['password'] }}';" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
+    echo "GRANT ALL PRIVILEGES ON DATABASE {{ values['database'] }} TO $username;" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
+    echo "GRANT $username TO $admin_username" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
```
gergopapi2: Should `GRANT <user> TO <admin>` not be the other way around? `GRANT <admin> TO <user>`?

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/init_db_remote.sh
```diff
@@ -6,18 +6,23 @@
...
-    echo "GRANT ALL PRIVILEGES ON DATABASE {{ values['user'] }} TO {{ values['database'] }};" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_url'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
-    echo "ALTER SCHEMA public OWNER TO {{ values['user'] }};" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_url'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
+    PGPASSWORD={{ values['remote_admin_pw'] }} createdb --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} {{ values['database'] }}
+    echo "CREATE USER $username WITH PASSWORD '{{ values['password'] }}';" | PGPASSWORD={{ values['remote_admin_pw'] }} psql --host={{ values['remote_db_host'] }} --port={{ values['remote_db_port'] }} --username={{ values['remote_admin'] }} -v "ON_ERROR_STOP=1" {{ values['database'] }}
```
gergopapi2: Don't we need login role for this user? Apparently not, as the old code had no login, either, I just do not understand why.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/init_db_remote.sh
bhavanki: This should be the correct ordering. See the answers to this SO question for more information.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/init_db_remote.sh
bhavanki: When you say "CREATE USER", the login attribute is automatically added.

https://www.postgresql.org/docs/9.1/sql-createuser.html

datalake/src/main/java/com/sequenceiq/datalake/flow/create/handler/RdsWaitHandler.java
```diff
@@ -101,18 +101,20 @@ private void setRdsCreatedStatus(SdxCluster cluster) {
...
-            throw new BadRequestException(message);
-        }
-        Map<String, Long> zones = env.getNetwork().getSubnetMetas().values().stream()
-                .collect(Collectors.groupingBy(CloudSubnet::getAvailabilityZone, Collectors.counting()));
-        if (zones.size() < 2) {
-            message = String.format("Cannot create external database for sdx: %s, the subnets in the vpc should be at least in two different availabilityzones",
-                    sdxId);
-            LOGGER.debug(message);
-            throw new BadRequestException(message);
+        if ("aws".equalsIgnoreCase(env.getCloudPlatform())) {
```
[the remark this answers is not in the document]
bhavanki: Right, I think it is redundant. I decided to leave it in place for now, because it doesn't hurt, and the goal of the jira is to focus on Azure, not AWS.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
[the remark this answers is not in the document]
lnardai: I though it would work the same as it worked for AWS postgres and every information is present in the rdsConfig.
What's in the RdsConfig itself or what's the difference?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/DistroXClusterCreationTest.java
```diff
@@ -181,7 +181,7 @@ public void testDistroxCloudStorage(MockedTestContext testContext) {
                 .withImageSettings(DIX_IMG_KEY)
                 .withNetwork(DIX_NET_KEY)
                 .when(distroXClient.create(), key(DISTRO_X_STACK))
-                .await(STACK_AVAILABLE)
+                .await(STACK_AVAILABLE, key(DISTRO_X_STACK))
```
lnardai: what required this change?

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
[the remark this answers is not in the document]
bhavanki: The RDSConfig contains the connection information for the database, like the Hive database. This now properly uses / stores a specific user for the database, and not the administrative username / password. Therefore, the administrative username and password have to be retrieved from the lookup, in redbeams, of the database server itself.

You are correct that this change is not specific to Azure. I changed what is saved in RDSConfig, in general, so that it is now the database-specific user, and not the administrative user.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/DistroXClusterCreationTest.java
bhavanki: I am not sure it is necessary. I had trouble getting the integration tests to run, and I made this change because it seemed like a good idea at the time. I know that the tests are passing now, so at worst, it doesn't hurt anything.

I will revert this change and see if the tests still pass.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
```diff
@@ -42,66 +41,97 @@
...
+     * Creates an RDSConfig object for a specific database.
+     *
+     * @param  stack   stack for naming purposes
+     * @param  cluster cluster to associate database with
+     * @param  dbName  database name
+     * @param  dbUser  database user
+     * @param  dbPort  port for database connections (through gateway)
+     * @return         RDSConfig object for database
+     */
+    private RDSConfig createNewRdsConfig(Stack stack, Cluster cluster, String dbName, String dbUserName, String dbPort) {
```
jadair-cloudera: NIT: createNew... seems redundant.

datalake/src/main/java/com/sequenceiq/datalake/flow/create/handler/RdsWaitHandler.java
jadair-cloudera: NIT: I guess we don't have constants for the various cloud providers?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/DistroXClusterCreationTest.java
bhavanki: Clarification - I'll try removing the change after merging this :)

datalake/src/main/java/com/sequenceiq/datalake/flow/create/handler/RdsWaitHandler.java
bhavanki: We do. I'll take another look at this after I merge.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
bhavanki: That was on purpose, remembering how confused I was whether an RDSConfig was being reused or not. :D

PR #6437, opened 2019-10-09
OPSAPS-52977 Ensure that certificates from JDK default trustore are i…

akanto:
I have just extended autotls with and extended functionality, I have already done some cleanup and refactor of this class on the master which I have not ported back.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/setup-autotls.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/setup-autotls.sh
@@ -8,6 +8,7 @@ CM_KEYTAB_FILE={{ cm_keytab.path }}
 CM_PRINCIPAL={{ cm_keytab.principal }}
 CERTMANAGER_DIR="/etc/cloudera-scm-server/certs"
 CERTMANAGER_ARGS=
+CACERTS_DIR=/opt/cacerts
 OUT_FILE=`mktemp -t signed_ca_chain.XXXXXX.pem`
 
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/setup-autotls.sh
```diff
@@ -16,9 +17,17 @@ source /bin/activate_salt_env
 
 kinit -kt $CM_KEYTAB_FILE $CM_PRINCIPAL
 
-/opt/cloudera/cm-agent/bin/certmanager --location $CERTMANAGER_DIR setup --configure-services $CERTMANAGER_ARGS --override ca_dn="CN=${HOSTNAME}" --stop-at-csr --altname DNS:${HOSTNAME}
+mkdir -p ${CACERTS_DIR}
+
+rm -f ${CACERTS_DIR}/cacerts.p12
+rm -f ${CACERTS_DIR}/cacerts.pem
+
+keytool -importkeystore -srckeystore /usr/lib/jvm/java/jre/lib/security/cacerts -srcstorepass changeit -destkeystore ${CACERTS_DIR}/cacerts.p12 -deststorepass changeit -deststoretype PKCS12
```
lacikaaa: are you sure that this java path is always valid? or we don't bother with custom images/other OS-s now?

akanto: Thanks, fixed to JAVA_HOME

PR #6441, opened 2019-10-09
Service alias + relocate vault path validation

akanto:


core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabProvider.java
```diff
@@ -26,7 +26,12 @@ public ServiceKeytabResponse getServiceKeytabResponse(Stack stack, GatewayConfig
             throws Exception {
         ServiceKeytabRequest request = new ServiceKeytabRequest();
         request.setEnvironmentCrn(stack.getEnvironmentCrn());
-        request.setServerHostName(primaryGatewayConfig.getHostname());
+        String hostname = primaryGatewayConfig.getHostname();
+        request.setServerHostName(hostname);
+        int dotIndex = hostname.indexOf('.');
+        if (dotIndex != -1) {
+            request.setServiceAlias(hostname.substring(0, dotIndex));
+        }
```
lacikaaa: StringUtils#substringBefore would be simpler: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringBefore-java.lang.String-java.lang.String-

akanto: Makes sense, and I will add some unit test

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Service.java
```diff
@@ -349,7 +360,7 @@ private String getExistingKeytab(String canonicalPrincipal, FreeIpaClient ipaCli
         }
     }
 
-    private static String constructCanonicalPrincipal(String serviceName, String hostName, String realm) {
+    private static String constructPrincipal(String serviceName, String hostName, String realm) {
```
lacikaaa: `hostName` is a bit misleading as we pass on alias too

akanto: Unfortunately, we use hostname for fqdn and also we use hostname for hostname in many places. I have renamed the to serverHostNameAlias, but consistent naming everywhere in the code is a massive refactor

PR #6444, opened 2019-10-09
Valid certification phase #2

biharitomi:
Proxy Knox through NGiNX with cookie based rewrite rules.
CM UI is available directly on `https://CLUSTER_ADDRESS/clouderamanager/` path but be aware of https://github.com/hortonworks/cloudbreak/commit/6f919ec3aa9bf0eeb379cf8e4110061a4a195bb5
This also means that port 8443 is totally eliminated and everything goes through port 433.


orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
```diff
@@ -3,34 +3,47 @@ map $http_upgrade $connection_upgrade {
...
+        proxy_set_header   Referer https://$host/$1;
+    }
+    location ~ .*/ {
+        proxy_pass         https://knox;
+        proxy_redirect     off;
+        proxy_set_header   Host $host;
+        proxy_set_header   Referer https://$host/;
+        proxy_set_header   X-Forwarded-Host $server_name;
+        proxy_set_header   X-Forwarded-Proto $scheme;
     }
```
lacikaaa: do we really want to redirect everything to knox in every case? like what happens if we choose not to install knox?

cloud-common/src/main/resources/application.yml
```diff
@@ -240,7 +240,7 @@ cb:
   nginx:
     port: 9443
   knox:
```
akanto: shouldn't we eliminate this property, it isa bit confusing, isn't it?

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/clusterproxy/ClusterProxyService.java
```diff
@@ -168,7 +168,7 @@ private ClusterServiceConfig serviceConfig(Stack stack) {
     private String knoxUrl(Stack stack) {
         String gatewayIp = stack.getPrimaryGatewayInstance().getPublicIpWrapper();
         Cluster cluster = stack.getCluster();
-        return String.format("https://%s:8443/%s", gatewayIp, cluster.getGateway().getPath());
+        return String.format("https://%s:443/%s", gatewayIp, cluster.getGateway().getPath());
```
akanto: Can we just delete, :443?

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
biharitomi: I mark this one as resolved because CB does redirect to the CM UI's direct route if gateway/Knox is not enabled.
https://github.com/hortonworks/cloudbreak/pull/6444/files#diff-6e00cc8735c353460aa0fd4845162ea4R78

akanto:
LGTM

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
akanto: Right, it had caused some trouble if both are configured.  So the only thing what we can do is to redirect everything to knox, and if something fails and you would like to debug then you need to touch the nginx conf.

@biharitomi maybe some comment or guidance in the file would be nice, which says that if knox has failed then please uncomment these few lines and we will bypass knox... This would be super useful for debugging scenarios.

PR #6451, opened 2019-10-10
CB-3931 Knox - Fix Manager/Admin API topology authorization

risdenk:
This adjusts the manager/admin api topologies to ensure that authorization takes place.

* Uses `HadoopGroupProvider` for group lookup
* Adjusts `AclsAuthz` to fall back to the Knox CM gateway-site.xml config for admin group
* Whitespace adjustment to match current Knox repo version of these files


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/manager.xml.j2
```diff
@@ -16,73 +16,69 @@
...
+            <role>authorization</role>
+            <name>AclsAuthz</name>
+            <enabled>true</enabled>
+            <param>
+                <name>knox.acl.mode</name>
+                <value>OR</value>
+            </param>
+            <param>
+                <name>knox.acl</name>
+                <value>KNOX_ADMIN_USERS;KNOX_ADMIN_GROUPS;*</value>
```
risdenk: @keyki this is the only change I don't understand. Why is `ldap.adminGroup` needed for CB? 

Since these files are under the `/cm/` folder I wasn't sure if this affects any Ambari deployments. I would assume not and we should rely on the CM Knox gateway-site config for group which is `KNOX_ADMIN_GROUPS`.

keyki: we don't need it.

keyki:
retest please

risdenk:
Rebased this on latest master in case that is why the tests were failing.

risdenk:
@keyki FYI looks like tests pass now and should be good to go.

PR #6452, opened 2019-10-10
CB-3842. Add get DistroX api endpoint for internal CB user

oleewere:
Requirement:
- have an ability to get some stack details (env crn/name, stack crn/name, stack create/terminate timestamp) based on a crn for an internal user 

Solution:
- added **internal** distrox endpoint (similar as sdx internal) - that is for internal usage inside Control Plane, so that is the name right now, that can be changed in another commit if that needs to be changed (like support by a query param like tenantUnawere or something . else)
- use INTERNAL role for the operation (that can be re-used) - map to AUTOSCALE role only for autoscale (as that is internal from CB perspective)

edit: use only default iam internal actor

testing is in progress ...
manual tests looked ok

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXInternalV1Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXInternalV1Endpoint.java
@@ -1,0 +1,29 @@
+package com.sequenceiq.distrox.api.v1.distrox.endpoint;
+
+import static com.sequenceiq.distrox.api.v1.distrox.doc.DistroXOpDescription.GET_BY_CRN_INTERNAL;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackViewV4Response;
+import com.sequenceiq.cloudbreak.doc.Notes;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+@Path("/v1/internal/distrox")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+@Api(value = "/v1/internal/distrox", protocols = "http,https")
+public interface DistroXInternalV1Endpoint {
+
+    @GET
+    @Path("crn/{crn}")
+    @ApiOperation(value = GET_BY_CRN_INTERNAL, produces = MediaType.APPLICATION_JSON, notes = Notes.STACK_NOTES,
+            nickname = "getDistroXInternalV1ByCrn")
+    StackViewV4Response getByCrn(@PathParam("crn") String crn);
+}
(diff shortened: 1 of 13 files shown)


sodre90:
What is the use case for this?

doktoric:
I think we already mentioned in the ticket this is not really a good solution for the problem

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/InternalStackViewResponse.java
```diff
@@ -0,0 +1,77 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.JsonEntity;
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModelProperty;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class InternalStackViewResponse implements JsonEntity {
```
doktoric: is this a stack or a distrox ?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperation.java
```diff
@@ -143,6 +150,22 @@ public StackV4Response get(@NotNull StackAccessDto stackAccessDto, Long workspac
...
+        CloudbreakUser user = restRequestThreadLocalService.getCloudbreakUser();
+        if (Crn.isCrn(user.getUserCrn()) && InternalCrnBuilder.isInternalCrn(user.getUserCrn())) {
+            Stack stack = stackService.findByCrnAndStackType(stackAccessDto.getCrn(), stackType);
+            LOGGER.info("Query Stack successfully finished with crn {}", stackAccessDto.getCrn());
+            InternalStackViewResponse stackResponse = internalStackConverter.convert(stack);
+            LOGGER.info("Adding environment name to the response.");
+            environmentServiceDecorator.prepareEnvironment(stackResponse);
+            return stackResponse;
+        } else {
+            throw new BadRequestException(String.format("Only internal users can use internal stack API (crn: %s)", user.getUserCrn()));
```
doktoric: intrnal distrox

oleewere:
@doktoric , i'm recommending to do the integration tests in another PR as it requires a bit more work (CB client set with API keys, where you get a CRN, but for the internal api example it needs to work with the internal actor crn so it requires to intercept the client which is used for testing)

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/InternalStackViewResponse.java
oleewere: updated to use stackviewv4response

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXInternalV1Endpoint.java
```diff
@@ -0,0 +1,29 @@
...
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackViewV4Response;
+import com.sequenceiq.cloudbreak.doc.Notes;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+@Path("/v1/internal/distrox")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+@Api(value = "/v1/internal/distrox", protocols = "http,https")
```
lnardai: This will cause a confusion in concept. We also have an internal Endpoint in SDX but it's used for a totally different purpose, to supply custom properties.
Can we come up with a different name?

oleewere: idea is to be internal in CP (not internal as internal user), also we did not find any good name (as I discussed with Ricsi as well)...that can be changed later if someone came up with a better name. But also ... i think SDX internal should be extended with other internal calls if there will be newer ones

doktoric:
after the merge please implement  https://jira.cloudera.com/browse/CB-3982

PR #6458, opened 2019-10-11
CB-3956 preview template support

doktoric:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/ResourceRelease.java
```diff
@@ -0,0 +1,6 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.common;
+
+public enum ResourceRelease {
```
attilapalfi92: I think FeatureState or FeatureReleaseState would be a better name.

core/src/main/resources/defaults/clustertemplates/aws/aws-opdb-702.json
```diff
@@ -2,6 +2,7 @@
   "name": "CDP 1.2 - Operational Database template",
   "description": "",
   "type": "OTHER",
+  "release": "RELEASED",
```
attilapalfi92: featureState would be more self-explanatory in my humble opinion if you ask me.

doktoric: cool I will do that

PR #6462, opened 2019-10-14
CB-3370 add Impala Debug UI to exposed services

keyki:


core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
```diff
@@ -360,4 +388,12 @@ private String getHBaseUIUrlWithHostParameterFromGatewayTopology(String managerI
         return String.format("https://%s/%s/%s%s?host=%s&port=%s", managerIp, gateway.getPath(), gt.getTopologyName(),
                 ExposedService.HBASE_UI.getKnoxUrl(), nameNodePrivateIp, ExposedService.HBASE_UI.getPort());
     }
+
+    private String getImpalaCoordinatorUrlWithHostFromGatewayTopology(String managerIp, GatewayTopology gt, String impalaPrivateIp, boolean autoTlsEnabled) {
+        Gateway gateway = gt.getGateway();
+        String protocol = autoTlsEnabled ? "https" : "http";
+        Integer port = autoTlsEnabled ? ExposedService.IMPALA_DEBUG_UI.getTlsPort() : ExposedService.IMPALA_DEBUG_UI.getPort();
+        return String.format("https://%s/%s/%s%s?scheme=http&host=%s://%s&port=%s", managerIp, gateway.getPath(), gt.getTopologyName(),
+                ExposedService.IMPALA_DEBUG_UI.getKnoxUrl(), protocol, impalaPrivateIp, port);
```
lacikaaa: this logic looks the same as `getHdfsUIUrlWithHostParameterFromGatewayTopology`. could you generalize it? at least `protocol` extaction can be moved out.

keyki: I moved the protocol generation to a private method, but I'm not sure if it's worth generalising more

PR #6465, opened 2019-10-14
CB-3963 - experimentalFeatures object in Environment entity

attilapalfi92:


environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDtoConverter.java
```diff
@@ -48,14 +48,12 @@ public EnvironmentDto environmentToDto(Environment environment) {
                 .withTelemetry(environment.getTelemetry())
                 .withEnvironmentStatus(environment.getStatus())
                 .withCreator(environment.getCreator())
-                .withCreateFreeIpa(environment.isCreateFreeIpa())
```
doktoric: do we rally want to remove this?

attilapalfi92:
@doktoric there is no need for a migration script with @Embedded

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDtoConverter.java
attilapalfi92: It was a duplicate, see line 53

PR #6467, opened 2019-10-14
CB-3932 enable ccm

doktoric:


doktoric:
do not merge yet I waiting for a decision

PR #6479, opened 2019-10-15
CB-3791 Configure CM with ordinary usernames for Azure databases

bhavanki:
Although Cloudbreak itself stores the connection username for Azure
databases in the username@short-hostname format, it now configures
Cloudera Manager with the bare usernames for the databases. This is
implemented by having the RdsView class automatically strip any "@xxx"
suffix encountered for a database username.

Meanwhile, AbstractRdsConfigProvider is fixed so that it is sure to
write the full username@short-hostname format, as saved in RDSConfig
objects, into the postgres.sls salt pillar file referenced when running
CM's scm_prepare_database.sh script to set up new databases.

Another salt script, cloudera/manager/start.sls, also uses the
scm_prepare_database.sh script at another point in the bootstrap
process. It specifically retrieves the full username@short-hostname
format of the username that it needs. This is available to it because
when RdsView objects are written out to the pillar, they now include a
new "fullConnectionUserName" field containing it. (The field will match
the existing "connectionUserName" field for all other databases.)

Finally, a couple of items are fixed based on review feedback from the
prior changeset for CB-3791.


core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
```diff
@@ -72,9 +72,9 @@
                 }
                 String dbName = getDb();
                 postgres.put("database", dbName);
-                postgres.put("user", getDbUser());
+                postgres.put("user", rdsConfig.getConnectionUserName());
```
lnardai: Are we sure we want to use the same username for all databases?
Previously this was based on RdsProvider type.

bhavanki: This doesn't use the same username for all databases. Above this line, the method fetches the RDSConfig object for the provider type (based on `getRdsType()`). That object, in turn, is saved earlier in the datalake creation, by the `createPostgresRdsConfigIfNeeded` method which is below this line in the file. The user for the RDS provider implementation is used there, so each RDSConfig object already receives a username specific to the provider type.

```
newRdsConfig = dbServerConfigurer.createNewRdsConfig(stack, cluster, getDb(), getDbUser(), getRdsType());
```

This change is here now so that the value of "user" written to the salt pillar is in *username@short-hostname* format, which is necessary for scm_prepare_database.sh to work with it.

bhavanki: To avoid further rebase problems, I'm going to close out this question / discussion, but do hit me up if you need more information. I will have one more small PR after this as well.

PR #6480, opened 2019-10-15
CDPCP-721. Trigger user sync through polling event generation ids

handavid:
This commit supports the automatic user sync feature by polling the UMS
for changes in each account. The UMS now provides an API for retrieving
event generation ids.

A UserSyncStatus model is added to the database to hold the event generation
ids for the last successful full user sync for each environment. A poller
class is added that runs on a configurable interval. The poller retrieves
the event generation ids from the UMS, compares them to the stored event
generation ids, and triggers the user sync if the ids do not match.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/repository/StackRepository.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/repository/StackRepository.java
@@ -23,6 +23,10 @@ import com.sequenceiq.freeipa.entity.Stack;
 public interface StackRepository extends BaseJpaRepository<Stack, Long> {
 
     @CheckPermission(action = ResourceAction.READ)
+    @Query("SELECT s FROM Stack s WHERE s.terminated = -1")
+    List<Stack> findAllRunning();
+
+    @CheckPermission(action = ResourceAction.READ)
     @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig LEFT JOIN FETCH ig.instanceMetaData WHERE s.id= :id ")
     Optional<Stack> findOneWithLists(@Param("id") Long id);
 
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
@@ -5,7 +5,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -26,8 +25,6 @@ import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
 import com.sequenceiq.cloudbreak.auth.altus.exception.UmsOperationException;
 import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
-import com.sequenceiq.cloudbreak.logger.LoggerContextKey;
-import com.sequenceiq.cloudbreak.logger.MDCBuilder;
 import com.sequenceiq.freeipa.service.freeipa.user.model.FmsGroup;
 import com.sequenceiq.freeipa.service.freeipa.user.model.FmsUser;
 import com.sequenceiq.freeipa.service.freeipa.user.model.UsersState;
@@ -44,16 +41,8 @@ public class UmsUsersStateProvider {
     private GrpcUmsClient grpcUmsClient;
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
-                                                            Set<String> userCrns, Set<String> machineUserCrns) {
+                                                            Set<String> userCrns, Set<String> machineUserCrns, Optional<String> requestIdOptional) {
         try {
-            final Optional<String> requestIdOptional = Optional.of(
-                Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()))
-                    .orElseGet(() -> {
-                String requestId = UUID.randomUUID().toString();
-                LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestId);
-                return requestId;
-            }));
-
             LOGGER.debug("Getting UMS state for environments {} with requestId {}", environmentCrns, requestIdOptional);
 
             Map<String, UsersState> envUsersStateMap = new HashMap<>();
@@ -63,7 +52,6 @@ public class UmsUsersStateProvider {
 
             List<MachineUser> machineUsers = machineUserCrns.isEmpty() ? grpcUmsClient.listAllMachineUsers(actorCrn, accountId, requestIdOptional)
                 : grpcUmsClient.listMachineUsers(actorCrn, accountId, List.copyOf(machineUserCrns), requestIdOptional);
-
 
             Map<String, FmsGroup> crnToFmsGroup = grpcUmsClient.listGroups(actorCrn, accountId, List.of(), requestIdOptional).stream()
                     .collect(Collectors.toMap(Group::getCrn, this::umsGroupToGroup));
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
@@ -3,11 +3,13 @@ package com.sequenceiq.freeipa.service.freeipa.user;
 import static java.util.Objects.requireNonNull;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
@@ -28,6 +30,8 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.CrnParseException;
+import com.sequenceiq.cloudbreak.common.json.Json;
+import com.sequenceiq.cloudbreak.logger.LoggerContextKey;
 import com.sequenceiq.cloudbreak.logger.MDCBuilder;
 import com.sequenceiq.freeipa.api.v1.freeipa.user.model.FailureDetails;
 import com.sequenceiq.freeipa.api.v1.freeipa.user.model.SuccessDetails;
@@ -43,6 +47,7 @@ import com.sequenceiq.freeipa.controller.exception.NotFoundException;
 import com.sequenceiq.freeipa.converter.freeipa.user.SyncOperationToSyncOperationStatus;
 import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.entity.SyncOperation;
+import com.sequenceiq.freeipa.entity.UserSyncStatus;
 import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
 import com.sequenceiq.freeipa.service.freeipa.user.model.FmsGroup;
 import com.sequenceiq.freeipa.service.freeipa.user.model.FmsUser;
@@ -79,6 +84,12 @@ public class UserService {
     @Inject
     private SyncOperationToSyncOperationStatus syncOperationToSyncOperationStatus;
 
+    @Inject
+    private UmsEventGenerationIdsProvider umsEventGenerationIdsProvider;
+
+    @Inject
+    private UserSyncStatusService userSyncStatusService;
+
     public SyncOperationStatus synchronizeUsers(String accountId, String actorCrn, Set<String> environmentCrnFilter,
                                                 Set<String> userCrnFilter, Set<String> machineUserCrnFilter) {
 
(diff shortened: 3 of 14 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -112,39 +123,60 @@ private void asyncSynchronizeUsers(
         Set<String> userCrnFilter, Set<String> machineUserCrnFilter) {
         try {
             Set<String> environmentCrns = stacks.stream().map(Stack::getEnvironmentCrn).collect(Collectors.toSet());
+
+            Optional<String> requestIdOptional = Optional.of(
```
lacikaaa: why is it optional if it always has a value?

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPollerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        when(stack.getEnvironmentCrn()).thenReturn(ENVIRONMENT_CRN);
+        setupUserSyncStatus(stack, mock(UmsEventGenerationIds.class));
+        when(stackService.findAllRunning()).thenReturn(List.of(stack));
+
+        underTest.syncFreeIpaStacks();
+
+        verify(userService).synchronizeUsers(stack.getAccountId(), UsersyncPoller.INTERNAL_ACTOR_CRN,
+                Set.of(stack.getEnvironmentCrn()), Set.of(), Set.of());
+//        verify(userService).synchronizeUsers(eq(stack.getAccountId()), eq(UsersyncPoller.INTERNAL_ACTOR_CRN),
+//                eq(Set.of(stack.getEnvironmentCrn())), eq(Set.of()), eq(Set.of()));
```
lacikaaa: please delete if not needed

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -112,39 +123,60 @@ private void asyncSynchronizeUsers(
...
             Map<String, Future<SyncStatusDetail>> statusFutures = stacks.stream()
                     .collect(Collectors.toMap(Stack::getEnvironmentCrn,
                             stack -> asyncTaskExecutor.submit(() -> {
-                                        MDCBuilder.buildMdcContext(stack);
-                                        return synchronizeStack(stack, envToUmsStateMap.get(stack.getEnvironmentCrn()), userCrnFilter, machineUserCrnFilter);
-                                    })));
-
-            List<SuccessDetails> success = new ArrayList<>();
-            List<FailureDetails> failure = new ArrayList<>();
+                                MDCBuilder.buildMdcContext(stack);
```
lacikaaa: maybe we should save the MdcContext prior this and restore when finished, otherwise it would contain the last stack's properties which doesn't seem right at first glance

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
handavid: This is eventually used in the calls to the GrpcUmsClient so that the UMS logs will contain the same requestId. The GrpcUmsClient method signatures take Optional<String>, so this is an optimization to avoid repetitively creating Optionals.
If you prefer, I could keep it as a String in this class and create an Optional in the UmsUsersStateProvider and the UmsEventGenerationIdsUtil instead. That would keep the UserService a little cleaner.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
handavid: The AsyncTaskExecutor is decorated with the MDCCleanerTaskDecorator. That decorator should clean the MDC after each task completes.

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPollerTest.java
handavid: done

handavid:
@lacikaaa I've answered your questions. let me know if you want any changes made

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -44,16 +41,8 @@
     private GrpcUmsClient grpcUmsClient;
 
     public Map<String, UsersState> getEnvToUmsUsersStateMap(String accountId, String actorCrn, Set<String> environmentCrns,
-                                                            Set<String> userCrns, Set<String> machineUserCrns) {
+                                                            Set<String> userCrns, Set<String> machineUserCrns, Optional<String> requestIdOptional) {
```
giladwolff: So again, I don't understand the calling convention, but it seems to me that passing the request id everywhere is prone to create problems. What you want is that every grpc call to have a request id. It is better if you can ensure this in the request builder. The request builder can look at a thread local to see if the thread is part of an ongoing request. If not, generate a new request ID. I'm not sure if this is tracked but it is definitely worth a jira if there is no jira for that.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -0,0 +1,110 @@
...
+    @Inject
+    private UserSyncStatusService userSyncStatusService;
+
+    @Inject
+    private UserService userService;
+
+    @Inject
+    private UmsEventGenerationIdsUtil umsEventGenerationIdsUtil;
+
+    @Value("${freeipa.syncoperation.poller.enabled:true}")
```
giladwolff: I am not sure how easy it is to do, but instead of having the notion of enabled and disabled in the poller you can instead just instantiate the poller if the sync poller is enabled.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -0,0 +1,110 @@
...
+    private boolean enabled;
+
+    @Scheduled(fixedDelayString = "${freeipa.syncoperation.poller.fixed-delay-millis:60000}",
+            initialDelayString = "${freeipa.syncoperation.poller.initial-delay-millis:300000}")
+    public void pollUms() {
+        try {
+            if (enabled) {
+                syncFreeIpaStacks();
```
giladwolff: So how many instances of this service are running at the same time? 3? Every one of them is running this loop? or are you choose one instance to run the sync? Should you coordinate between them?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -0,0 +1,110 @@
...
+                    .collect(Collectors.groupingBy(Stack::getAccountId))
+                    .entrySet().stream()
+                    .forEach(stringListEntry -> {
+                        UmsEventGenerationIds currentGeneration =
+                                umsEventGenerationIdsUtil.getEventGenerationIds(INTERNAL_ACTOR_CRN, stringListEntry.getKey(), requestId);
+                        stringListEntry.getValue().stream()
+                                .forEach(stack -> {
+                                    if (isStale(stack, currentGeneration)) {
+                                        LOGGER.debug("Environment {} in Account {} is stale.", stack.getEnvironmentCrn(), stack.getAccountId());
+                                        SyncOperationStatus status = userService.synchronizeUsers(stack.getAccountId(), INTERNAL_ACTOR_CRN,
```
giladwolff: I want to make sure that I understand the expectations here. This will trigger user sync from every poller right? so if you have 3 instances the expectation is that 2 will fail (as there will be one ongoing sync)? Should you coordinate who runs this code?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -112,39 +123,60 @@ private void asyncSynchronizeUsers(
...
+            Optional<String> requestIdOptional = Optional.of(
+                    Optional.ofNullable(MDCBuilder.getMdcContextMap().get(LoggerContextKey.REQUEST_ID.toString()))
+                            .orElseGet(() -> {
+                                String requestId = UUID.randomUUID().toString();
+                                LOGGER.debug("No requestId found. Setting request id to new UUID [{}]", requestId);
+                                MDCBuilder.addRequestId(requestId);
+                                return requestId;
+                            }));
+
+            Json umsEventGenerationIdsJson = userCrnFilter.isEmpty() && machineUserCrnFilter.isEmpty() ?
```
giladwolff: I have a preference to have a variable isFullSync so you can say something like isFullSync ? ... : ....

freeipa/src/main/java/com/sequenceiq/freeipa/repository/StackRepository.java
```diff
@@ -22,6 +22,10 @@
 @AuthorizationResourceType(resource = AuthorizationResource.ENVIRONMENT)
 public interface StackRepository extends BaseJpaRepository<Stack, Long> {
 
+    @CheckPermission(action = ResourceAction.READ)
+    @Query("SELECT s FROM Stack s WHERE s.terminated = -1")
```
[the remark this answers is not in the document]
handavid: opened CDPCP-907

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsEventGenerationIdsUtil.java
```diff
@@ -0,0 +1,65 @@
...
+    private GrpcUmsClient grpcUmsClient;
+
+    public UmsEventGenerationIds getEventGenerationIds(String actor, String accountId, Optional<String> requestId) {
+        GetEventGenerationIdsResponse response = grpcUmsClient.getEventGenerationIds(actor, accountId, requestId);
+
+        UmsEventGenerationIds umsEventGenerationIds = new UmsEventGenerationIds();
+        Map<String, String> eventGenerationIdsMap = new HashMap<>();
+        for (EventMapping eventMapping : EventMapping.values()) {
+            String eventId = eventMapping.converter.apply(response);
+            if (eventId != null) {
```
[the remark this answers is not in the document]
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsEventGenerationIdsUtil.java
```diff
@@ -0,0 +1,65 @@
...
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetEventGenerationIdsResponse;
+import com.google.common.annotations.VisibleForTesting;
+import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
+import com.sequenceiq.freeipa.service.freeipa.user.model.UmsEventGenerationIds;
+
+@Component
+public class UmsEventGenerationIdsUtil {
+
+    @VisibleForTesting
+    enum EventMapping {
```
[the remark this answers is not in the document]
handavid: changed to UmsEventGenerationIdsProvider. Extractor doesn't work for me because the class retrieves the values as well as converts it. hmm... maybe that suggests that I should split the class or have the caller invoke the GrpcUmsClient directly. I'll consider that as a future improvement.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: I agree. requestId handling could be pushed to the GrpcUmsClient class. I opened CB-4021

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: I can investigate that for the future. It should be doable in one of the Configuration classes to either explicitly create/not create the bean

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: I don't know of any way for the FMS services to communicate/coordinate amongst themselves other than through the db. I'm relying on the db to reject simultaneous syncs.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: you are correct.
it would be better to coordinate but I don't think it's possible at the moment.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: How would Quartz help if we don't have any communication between the FMS instances? If we need to differentiate the instances in some way, then can't we just "disable" the poller in the configuration for 2 of the FMS instances?

I'm not too concerned about this redundancy because we will eventually replace the polling with Cadence. In the meantime, the work performed here is not very heavy compared to the usersync process itself. The first thing user sync does is to accept or reject the operation if there is already a sync running. Also, I do not see any reason for the timing of the three pollers to converge, so the polls will be arbitrarily spread across the polling interval.

If you are really concerned with separating the polling, we can add a table to the db where the pollers can store a lock with and expiration time to guarantee a minimum time between polls. Let me know if you want that. I would like to add that in future work rather than in this commit.

giladwolff: So using the DB for synchronizing this is I think pretty easy to do. Chris mentioned to me that you can use k8s to "elect" a master. Basically there is a way to tell k8s to mark one of the containers as master. You can then run the poller from there and only from there. That may be the easiest solution for now.

keyki: For Kubernetes based leader election you either include a new side container to your POD or you need to invoke Kube endpoints. In autoscaling we implemented leader election based on DB only simply try to lock a certain row and whoever is able to do that is leader. If that one dies the next one who can will be leader. I would avoid leveraging Kubernetes for this, because FMS is not tied to any platform and can run simply on mac in Docker. This is how we use it and i wouldn't tie it to K8S if we don't have to.

PR #6486, opened 2019-10-16
CB-3981 Add Kafka LDAP configuration

bergerdenes:


doktoric:
waiting for @in-park

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/kafka/KafkaLdapConfigProviderTest.java
```diff
@@ -0,0 +1,130 @@
...
+        S3FileSystemConfigurationsView fileSystemConfigurationsView =
+                new S3FileSystemConfigurationsView(new S3FileSystem(), storageLocations, false);
+
+        return TemplatePreparationObject.Builder.builder().withFileSystemConfigurationView(fileSystemConfigurationsView)
+                .withHostgroupViews(Set.of(master, worker)).build();
+    }
+
+    private String getBlueprintText(String path) {
+        return FileReaderUtils.readFileFromClasspathQuietly(path);
+    }
```
gregito: these methods are not in use, kindly asking to remove them

bergerdenes: removed

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/kafka/KafkaLdapConfigProvider.java
```diff
@@ -0,0 +1,44 @@
...
+    public List<ApiClusterTemplateConfig> getServiceConfigs(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
+        LdapView ldapView = source.getLdapConfig().get();
+        return List.of(
+                config(LDAP_AUTH_ENABLE, "true"),
+                config(LDAP_AUTH_URL, ldapView.getConnectionURL()),
+                config(LDAP_AUTH_USER_DN_TEMPLATE, ldapView.getUserDnPattern()));
+    }
+
+    @Override
+    public boolean isConfigurationNeeded(CmTemplateProcessor cmTemplateProcessor, TemplatePreparationObject source) {
```
benyoka: to be more flexible in the future, I'd suggest the following in the logic:

1. Don't set LDAP_AUTH_ENABLE
2. Set the other two properties only is LDAP_AUTH_ENABLE is set to true.

This way the behaviour will be controllable from the blueprint. When Kafka will use better authentication (PAM?) then source code will not have to change, only the blueprint.

PR #6487, opened 2019-10-16
CB-3982 Implement ITs for DistroX internal API.

oleewere:


integration-test/docker-compose_template.yml
```diff
@@ -32,6 +32,7 @@ services:
       - INTEGRATIONTEST_SDX_SERVER=http://dev-gateway
       - INTEGRATIONTEST_REDBEAMS_SERVER=http://dev-gateway
       - CLOUDBREAK_URL=cloudbreak:8080
+      - ENVIRONMENT_URL=environment:8088
```
doktoric: isn't this one duplicated by INTEGRATIONTEST_ENVIRONMENT_SERVER

oleewere: a bit different, as i need ionternal urls not the gateway

PR #6497, opened 2019-10-17
Supporting virtualgroups for env/freeipa

akanto:


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
```diff
@@ -654,4 +655,34 @@ public String getIdentityProviderMetadataXml(String requestId, String accountId)
                 .build());
     }
 
+    public UserManagementProto.SetWorkloadAdministrationGroupNameResponse setWorkloadAdministrationGroupName(String requestId, String accountId,
+            String right, String resource) {
+        checkNotNull(accountId);
```
handavid: Are right and resource required for these requests? If so then we should null-check them also.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -210,28 +209,27 @@ public void listUsers(ListUsersRequest request, StreamObserver<ListUsersResponse
...
+                .setCrn(mockCrnService.createCrn(ACCOUNT_ID_ALTUS, Crn.Service.IAM, Crn.ResourceType.ROLE, "PowerUser").toString())
                 .setCreationDateMs(CREATION_DATE_MS)
                 .addPolicy(powerUserPolicy)
                 .build();
         RoleAssignment roleAssignment = RoleAssignment.newBuilder().setRole(powerUserRole).build();
         GetRightsResponse.Builder responseBuilder = GetRightsResponse.newBuilder()
                 .addGroupCrn(group.getCrn())
-                .addRoleAssignment(roleAssignment)
-                .addWorkloadAdministrationGroupName("mockworkloadadministrationgroup0");
+                .addRoleAssignment(roleAssignment);
```
handavid: we need to have some workload administration group returned in the GetRights call because that is how user sync determines group membership.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -510,6 +510,29 @@ public void notifyResourceDeleted(String userCrn, String resourceCrn, Optional<S
         }
     }
 
+    public String setVirtualGroupName(String actorCrn, String accountId, Optional<String> requestId, String right, String resource) {
```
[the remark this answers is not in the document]
akanto: Thanks, it was an accidental rename from IDEA

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
handavid: right now anything is fine from the mock. your suggested virtual group is fine. I just want to make sure there's something there for user sync to read and push to the ipa server.

akanto: Okay, thanks, will be fixed immediately.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockGroupManagementService.java
```diff
@@ -0,0 +1,118 @@
...
+        UserManagementProto.GetWorkloadAdministrationGroupNameResponse.Builder respBuilder =
+                UserManagementProto.GetWorkloadAdministrationGroupNameResponse.getDefaultInstance().toBuilder();
+        respBuilder.setWorkloadAdministrationGroupName(generateVirtualGroupName(request.getRightName()));
+        responseObserver.onNext(respBuilder.build());
+        responseObserver.onCompleted();
+    }
+
+    void setWorkloadAdministrationGroupName(UserManagementProto.SetWorkloadAdministrationGroupNameRequest request,
+            StreamObserver<UserManagementProto.SetWorkloadAdministrationGroupNameResponse> responseObserver) {
+        responseObserver.onCompleted();
```
lajosrodek: The code above in `GrpcUmsClient.setWorkloadAdministrationGroupName()` seems to depend on the workloadAdministrationGroupName being returned in the response. Thus, I suspect, lines 59--62 of this class (from `getWorkloadAdministrationGroupName()`) should be also copied here to populate the response appropriately. (`EnvironmentCreationService.create()` calls `GrpcUmsClient.setWorkloadAdministrationGroupName()`, so I wonder how the current logic can work right with the mock-caas. 😃 )

akanto: Right, it doesn't work, if the GET is not invoked.

akanto: Fixed, even the set method will return the proper group value

PR #6498, opened 2019-10-17
CB-3988 – create default cluster definition(s) for azure

gregito:
CB-3988 – create default cluster definition(s) for azure


doktoric:
when you bring this back to master you need to create the 1.2 definitions as well

PR #6500, opened 2019-10-17
CB-3942 Start sync when primary gateway detected as unhealthy

horadla23:
Wait for this PR #6492


horadla23:
I have written the unit test for this, but the logic is implemented by @foldik, I opened this PR so @foldik please give an approve for this and if jobs are succeeded, then I can merge this.

PR #6502, opened 2019-10-17
CB-4039 KnoxSSO - JWT Tokens Are Valid for Extremely Long Duration

risdenk:
The KnoxSSO topology TTL should be limited to a duration instead of `-1`. This sets the duration to 24hrs in milliseconds. After this ttl expires, the user will need to login again.


risdenk:
Ugh picked the wrong base. opening a new PR

PR #6509, opened 2019-10-18
CB-4019. Add ResourceCrn annotation for DistroX get by CRN Api endpoint

oleewere:
- adding resource crn to get By crn endpoint (only !!! that - following calls will be needed to be validated)

validating is in progress with real cluster ...

not deleting internal endpoint yet, in order to deploy this first to mow-dev, until internal usage is not deleted from metering service


oleewere:
retest this please

PR #6510, opened 2019-10-18
Disable Scheduler Configuration Mutation Api

PrabhuJoseph:
Disable Scheduler Configuration Mutation Api. Have tested with launching a CM cluster using CloudBreak and the config is added properly in Safety Valve.


sunilgovind:
CB-3958

keyki:
As I read through the jira comments this should be enabled as TP feature as not many people is confident making this change to 1.2. If it's still the case then we should not merge this change.

keyki:
It seems like there was an issue when we removed CDP 1.1 templates so that is why IT is failing on this PR. Can you please rebase the branch to be on top of master?

sunilgovind:
recheck

PR #6517, opened 2019-10-18
CB-4040 KnoxSSO - Improve knoxsso.redirect.whitelist.regex

risdenk:
If a user facing fqdn is defined, Knox will use that to set the domain for the redirect whitelist regex.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knoxsso.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knoxsso.xml.j2
@@ -123,13 +123,17 @@
             <value>true</value>
         </param>
         <param>
-	    <!-- 24 hours in milliseconds = 86400000 -->
+            <!-- 24 hours in milliseconds = 86400000 -->
             <name>knoxsso.token.ttl</name>
             <value>86400000</value>
         </param>
         <param>
            <name>knoxsso.redirect.whitelist.regex</name>
-           <value>.*</value>
+           {% if salt['pillar.get']('gateway:userfacingdomain') is defined and salt['pillar.get']('gateway:userfacingdomain')|length > 1 %}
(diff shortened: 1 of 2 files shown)


risdenk:
@keyki I don't know if this should be merged as is. I opened it more to get some thoughts on it. I think it would be better to have the domain populated from CB somehow via Salt. I just don't know enough about that part.

keyki:
In local environments we don't have the public domain management service so we can only access Knox via private IP.

risdenk:
Yea I figured it was more nuanced. So I'll need some help figuring out what variables to keep this on to make the regex reasonable. For local dev `.*` should be fine, but should be limited to domains when deployed.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/knoxsso.xml.j2
```diff
@@ -127,9 +127,14 @@
...
         </param>
         <param>
+	   <!-- 
+	        This assumes that valid DNS is setup and access via a
+	        subdomain that ends in cloudera.com or cloudera.site.
+		If access is over IP the redirect will not match.
+           -->
            <name>knoxsso.redirect.whitelist.regex</name>
-           <value>.*</value>
+           <value>^\/.*$;^https?:\/\/(.+.cloudera.(?:com|site)):[0-9]+\/?.*$</value>
```
akanto: In production it might be ok, it would work also on Mow *, once we register the endpoints to PEM service, bu we need a different approach in local:
- either use PEM locally
- put the pattern into java code and push down to Salt

risdenk:
I'm still looking at the best way to do this.

risdenk:
I still need to test this somehow.

keyki:
I think @biharitomi  can help you configure this value from salt.

PR #6526, opened 2019-10-21
CB-3873 SDX to configure HMS with the required changes for password b…

topolyai5:


core/src/main/resources/defaults/blueprints/cdp-sdx-ha.bp
```diff
@@ -219,6 +219,12 @@
                         "roleType": "HIVEMETASTORE"
                     }
                 ],
+                "serviceConfigs": [
+                    {
+                        "name": "hive_metastore_enable_ldap_auth",
```
doktoric: should it be in the configprovider?

topolyai5: no, it should not, it is fixed for sdx and could not be set by user

doktoric: we are setting hive_metastore_ldap_uri in the configprovider which means every cluster will use this property

topolyai5: Yes, but we don't enable the feature.

topolyai5: we don't know is it a DL or not, here: https://github.com/hortonworks/cloudbreak/pull/6526/files#diff-3ba4d7ee120408de64e69ad4097ed9da

PR #6527, opened 2019-10-21
Disable Scheduler Configuration Mutation Api

PrabhuJoseph:
Disable Scheduler Configuration Mutation Api. YARN Queue Manager is Technical Preview in CDP-7.0.2. This config is by default disabled and will be enabled when Queue Manager is installed by CM.
Have tested with launching a CM cluster using CloudBreak and the config is added properly in Safety Valve.


keyki:
We've create rc-2.15 to deploy to dev, can you please change the base version to that one.

PR #6529, opened 2019-10-21
CB-1755 - Attach user-assigned Azure MSI to IDBroker host

pdarvasi:
This commit contains the following extras:
- ADLS as log location
- lazy init exception fix


keyki:
verified with mock idbroker msi and I was able to access my store account + logs are written there

PR #6530, opened 2019-10-21
CB-2321 Cannot determine the ARN associated with a role credential

topolyai5:


environment-api/src/main/java/com/sequenceiq/environment/api/doc/credential/CredentialModelDescription.java
```diff
@@ -25,6 +25,7 @@
     public static final String ATTRIBUTES = "provider specific attributes of the credential";
     public static final String VERIFICATION_STATUS_TEXT = "verification status text for credential, if empty then there is no verification issue";
     public static final String CREATOR = "crn of the creator";
+    public static final String AWS_ROLE_ARN = "the role ARN of the credential ";
```
lajosrodek: Minor: superfluous space at the end.

PR #6543, opened 2019-10-22
CB-3585 FreeIPA cleanup process must be finished before you can launch new cluster with the same name

lacikaaa:
- move cleanup to flow on FreeIPA side
- generalize SyncOperation -> new Operation endpoint, but left old api for compatibility
- CB will poll FreeIPA operation to see if it's finished/failed/etc during downscale and terminate. Force terminate/downscale will ignore the result of cleanup


lacikaaa:
@handavid thanks for reviewing. I have fixed the issues you mentioned, please check again

PR #6544, opened 2019-10-22
Use internal actors to call virtual group UMS API

akanto:
CB-3542 use internal actor to call virtual group api.

It seems that we are instantiating the internal actor everywhere, that is the reson why I implemented in this way.


akanto:
~/prj/hortonworks/cloudbreak-demos/cdp/local(master ✗) $ dp env create from-file --file env-aws-local.json
ERROR: status code: 500, message: Internal server error: NOT_FOUND: right 'environments/write' not found

PR #6547, opened 2019-10-22
Backward compatibility for service URLs when Knox is on port 8443

biharitomi:


core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
```diff
@@ -49,6 +50,9 @@
 
     private static final String API_TOPOLOGY_POSTFIX = "-api";
 
+    @Value("${cb.https.port}")
+    private String httpsPort;
```
doktoric: Do we need this?

biharitomi: I think yes, nginx will be listening on that port for public facing.

PR #6548, opened 2019-10-22
Cdpd 3643

harishjp:
* Adding DAS in DE template.
* Fix DAS service name in ExposedServices.
* Fix Tez proto base directory in DE template.

Closes #CDPD-3643
Closes #CB-3432

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
@@ -167,6 +167,10 @@
           {
             "name": "tez_container_size",
             "value": "4096"
+          },
+          {
+            "name": "hive_service_config_safety_valve",
+            "value": "<property><name>hive.server2.authentication.spnego.principal</name><value>HTTP/_HOST@CLOUDERA.SITE</value></property><property><name>hive.server2.authentication.spnego.keytab</name><value>hive.keytab</value></property><property><name>hive.hook.proto.file.per.event</name><value>true</value></property>"
           }
         ],
         "roleConfigGroups": [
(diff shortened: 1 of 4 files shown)


harishjp:
I do not have access to http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-pull-request-builder-integration-test/3578/ and do not know how to look for test failures. Please help.

core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
```diff
@@ -167,6 +161,10 @@
           {
             "name": "tez_container_size",
             "value": "4096"
+          },
+          {
+            "name": "hive_service_config_safety_valve",
+            "value": "<property><name>hive.server2.authentication.spnego.principal</name><value>HTTP/_HOST@CLOUDERA.SITE</value></property><property><name>hive.server2.authentication.spnego.keytab</name><value>hive.keytab</value></property><property><name>hive.hook.proto.file.per.event</name><value>true</value></property>"
```
keyki: This hard coded value is not going to work since are not using cloudera.site anymore but it's generated on the fly. But I believe this is already handled so this config can be removed from here.

harishjp: I've to add property: <property><name>hive.hook.proto.file.per.event</name><value>true</value></property>, how do I add to the safety value without overriding the previously set one.

harishjp: Shall I change this instead? template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java

keyki: You can add them here: 
HIVE_ON_TEZ: https://github.com/hortonworks/cloudbreak/blob/rc-2.15/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java#L25
HIVE_METASTORE: https://github.com/hortonworks/cloudbreak/blob/rc-2.15/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveMetastoreConfigProvider.java#L44

I guess you'll need it for Hive on Tez

harishjp: yup, its required for hive on tez. But since the safety_value can only be set in one config, I'll just add to the HiveKnoxConfigProvider for now, is there a way to query if its already set and append to it?

harishjp: Thanks @keyki. I've updated the code.

keyki:
I checked and it still does not expose the DAS link to the UI

PR #6549, opened 2019-10-23
CDPCP-951 added hashed password for Password123

abhishekaypurohit:
Added code in mock service to generate hash password. 

FYI: This is not tested with FreeIPA. 

@keyki @lacikaaa 

Closes: CDPCP-951

### DIFF
--- a/mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
+++ b/mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
@@ -9,6 +9,7 @@ import static org.springframework.security.jwt.JwtHelper.decodeAndVerify;
 
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -63,6 +64,7 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.io.Resources;
+import com.google.protobuf.util.JsonFormat;
 import com.sequenceiq.caas.grpc.GrpcActorContext;
 import com.sequenceiq.caas.model.AltusToken;
 import com.sequenceiq.caas.util.CrnHelper;
@@ -143,6 +145,8 @@ public class MockUserManagementService extends UserManagementGrpc.UserManagement
 
     private LoadingCache<String, GetEventGenerationIdsResponse> eventGenerationIdsCache;
 
+    private GetActorWorkloadCredentialsResponse actorWorkloadCredentialsResponse;
+
     @PostConstruct
     public void init() {
         this.cbLicense = getLicense();
@@ -162,6 +166,21 @@ public class MockUserManagementService extends UserManagementGrpc.UserManagement
                         return respBuilder.build();
                     }
                 });
+
+        initializeActorWorkloadCredentials();
+    }
+
+    @VisibleForTesting
+    void initializeActorWorkloadCredentials() {
+
+        GetActorWorkloadCredentialsResponse.Builder builder = GetActorWorkloadCredentialsResponse.newBuilder();
+        try {
+            JsonFormat.parser().merge(Resources.toString(
+                Resources.getResource("mock-responses/ums/get-workload-credentials.json"), StandardCharsets.UTF_8), builder);
+            actorWorkloadCredentialsResponse = builder.build();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


mock-caas/src/test/java/com/sequenceiq/caas/grpc/service/MockUserManagementServiceTest.java
```diff
@@ -61,4 +65,13 @@ public void testCreateWorkloadUsername() {
 
         Assert.assertEquals(expected, underTest.sanitizeWorkloadUsername(username));
     }
+
+    @Test
+    public void testGwtWorkloadCredentials() {
```
handavid: typo: testGetWorkloadCredentials

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -491,25 +499,52 @@ public void getActorWorkloadCredentials(com.cloudera.thunderhead.service.userman
...
-
-        respBuilder.addKerberosKeysBuilder(1)
-                .setSaltType(saltType)
-                .setKeyType(keyType18)
-                .setKeyValue("testKeyValue18")
-                .setSaltValue("IodizedGrainOfSalt")
-                .build();
+        final int hashsaltlength = 32;
+        final int keysaltlength = 8;
+        final String password = "Password123!";
```
handavid: it looks like the response is the same regardless of the request. can we build it once rather than for each request?

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -162,6 +172,60 @@ public GetEventGenerationIdsResponse load(String key) throws Exception {
...
+            baos.write(salt);
+            String hash = saltprefix + Base64.encodeBase64String(baos.toByteArray());
+
+            String userName = RandomStringUtils.randomAlphanumeric(keysaltlength);
+            String realm = RandomStringUtils.randomAlphanumeric(keysaltlength);
+            String principalName = userName + '@' + realm;
+            String keySalt = realm + userName;
+            KerberosPrincipal principal = new KerberosPrincipal(principalName);
+            KerberosKey keyAES128 = new KerberosKey(principal, password.toCharArray(), "AES128");
+            KerberosKey keyAES256 = new KerberosKey(principal, password.toCharArray(), "AES256");
```
toddlipcon: If we only need to care about there being a single password here (Password123) why don't we just precompute the value and stick it in as a constant? That seems easier than duplicating all the code from UMS.

abhishekaypurohit: yeah, let me explore and get the actual coming from UMS and use that.

abhishekaypurohit: @toddlipcon Thanks. I have changed for JSON and parsing it directly into the response. 
Unit test is only for checking if the values are getting loaded or not.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -7,11 +7,15 @@
...
 import static org.springframework.security.jwt.JwtHelper.decodeAndVerify;
 
+import java.io.ByteArrayOutputStream;
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.MessageDigest;
```
toddlipcon: a lot of these new imports are now unnecessary, right?

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
```diff
@@ -162,6 +171,21 @@ public GetEventGenerationIdsResponse load(String key) throws Exception {
...
+        initializeActorWorkloadCredentials();
+    }
+
+    @VisibleForTesting
+    void initializeActorWorkloadCredentials() {
+
+        GetActorWorkloadCredentialsResponse.Builder builder = GetActorWorkloadCredentialsResponse.newBuilder();
+        try {
+            JsonFormat.parser().merge(Resources.toString(
+                Resources.getResource("mock-responses/ums/get-workload-credentials.json"), StandardCharsets.UTF_8), builder);
```
toddlipcon: also looks like maybe you forgot to check this in?

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
abhishekaypurohit: yup cleaning.

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockUserManagementService.java
abhishekaypurohit: added.

PR #6552, opened 2019-10-23
CB-3791 Revert to passing full db usernames to CM

bhavanki:
One change in the prior commit for CB-3791 was having Cloudbreak
configure Cloudera Manager with only the bare usernames for database
connections, instead of the "full" username@short-hostname format needed
for connections to Azure-managed databases. This change undoes that
work, so that once again the entire username is configured in Cloudera
Manager.

This still does not allow Cloudera Manager to support clusters backed
by Azure databases, because of OPSAPS-53099. However, once that CM work
is done, Hive metastore table creation should succeed, and perhaps
overall cluster creation will as well.

Other previous changes in CB-3791 to username handling are preserved.
This change only affects how Cloudera Manager is configured.


lajosrodek:
Follow-up to #6479.

PR #6553, opened 2019-10-23
several fix

doktoric:


lajosrodek:
The CB-3873 part is a follow-up to #6526.

PR #6564, opened 2019-10-24
CB-3713 reverted previous solution, removed freeipa fetch from structured event

lnardai:
- Removed freeipa call because it was invoked with every transition during flow and with missing freeipa it was preventing termination for certain clusters


core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/FlowStructuredEventHandler.java
```diff
@@ -91,6 +91,7 @@ public void transition(Transition<S, E> transition) {
                 structuredEvent = structuredFlowEventFactory.createStucturedFlowEvent(stackId, flowDetails, true, exception);
                 exception = null;
             }
+//            throw new CloudbreakServiceException("Yolo faiuure");
```
lacikaaa: you might want to delete this line

PR #6567, opened 2019-10-24
CDPCP-944. Force sync on environment creation.

handavid:
This commit ensures that a user sync happens for newly created environments by
1. adding a 'force' flag to skip the acceptor when starting operations
2. track actual environments syncing rather than using the empty list to represent all
3. only get available stacks rather than all non-terminated stacks

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/AvailableStackProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/AvailableStackProvider.java
@@ -1,0 +1,45 @@
+package com.sequenceiq.freeipa.service.freeipa.user;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.stack.StackService;
+
+@Component
+public class AvailableStackProvider {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AvailableStackProvider.class);
+
+    @Inject
+    private StackService stackService;
+
+    public List<Stack> getAvailableStacks() {
+        LOGGER.debug("Retrieving all stacks");
+        return filterAvailableStacks(stackService.findAllRunning());
+    }
+
+    public List<Stack> getAvailableStacksByAccountId(String accountId) {
+        LOGGER.debug("Retrieving stacks for account {}", accountId);
+        return filterAvailableStacks(stackService.getAllByAccountId(accountId));
+    }
+
+    public List<Stack> getAvailableStacksByAccountIdAndEnvironmentCrns(String accountId, Collection<String> environmentCrnFilter) {
+        if (environmentCrnFilter.isEmpty()) {
+            return getAvailableStacksByAccountId(accountId);
+        } else {
+            LOGGER.debug("Retrieving stacks for account {} that match environment crns {}", accountId, environmentCrnFilter);
+            return filterAvailableStacks(stackService.getMultipleByEnvironmentCrnAndAccountId(environmentCrnFilter, accountId));
+        }
+    }
+
+    private List<Stack> filterAvailableStacks(List<Stack> stacks) {
+        return stacks.stream().filter(Stack::isAvailable).collect(Collectors.toList());
+    }
+}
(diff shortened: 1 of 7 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -91,36 +90,52 @@
     private UserSyncStatusService userSyncStatusService;
 
     public SyncOperationStatus synchronizeUsers(String accountId, String actorCrn, Set<String> environmentCrnFilter,
-                                                Set<String> userCrnFilter, Set<String> machineUserCrnFilter) {
+            Set<String> userCrnFilter, Set<String> machineUserCrnFilter) {
+        return synchronizeUsers(accountId, actorCrn, environmentCrnFilter, userCrnFilter, machineUserCrnFilter, false);
```
giladwolff: Can we have readability constants for these? DO_NOT_FORCE_SYNC and FORCE_SYNC?

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPollerTest.java
```diff
@@ -23,7 +23,6 @@
 import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.entity.UserSyncStatus;
```
giladwolff: Can we have tests that make sure that force works for both actor sync and full syncs?

handavid:
> I don't mind this going in, though I am not sure why we need this. Before the environment is created nobody is assigned to it, right? So this first sync is basically "mostly empty", no? Except for group names and WAG names?

we do have groups, but won't have WAGs until we add the listWorkloadAdministrationGroups call to the sync.

Right now we still have the incorrect role checking instead of rights checking so we currently have a bunch of users being synced. We already invoke this sync but it can get rejected. This commit fixes it so that the sync can't get rejected.

giladwolff:
> > I don't mind this going in, though I am not sure why we need this. Before the environment is created nobody is assigned to it, right? So this first sync is basically "mostly empty", no? Except for group names and WAG names?
> 
> we do have groups, but won't have WAGs until we add the listWorkloadAdministrationGroups call to the sync.
> 
> Right now we still have the incorrect role checking instead of rights checking so we currently have a bunch of users being synced. We already invoke this sync but it can get rejected. This commit fixes it so that the sync can't get rejected.

Bear with me, as I am still not very familiar with how things work. So apologies if you explained it to me and I forgot/misunderstood. This commit will force sync on environment creation, right? Are you saying that because the authorization checks are wrong this is useful because we will get the power users in?

handavid:
> > > I don't mind this going in, though I am not sure why we need this. Before the environment is created nobody is assigned to it, right? So this first sync is basically "mostly empty", no? Except for group names and WAG names?
> > 
> > 
> > we do have groups, but won't have WAGs until we add the listWorkloadAdministrationGroups call to the sync.
> > Right now we still have the incorrect role checking instead of rights checking so we currently have a bunch of users being synced. We already invoke this sync but it can get rejected. This commit fixes it so that the sync can't get rejected.
> 
> Bear with me, as I am still not very familiar with how things work. So apologies if you explained it to me and I forgot/misunderstood. This commit will force sync on environment creation, right? Are you saying that because the authorization checks are wrong this is useful because we will get the power users in?

sort of. Current behavior is wrong and broken and this fixes the broken part of it but not the wrong part of it. I don't mind delaying this commit until the next release because in the future 1) less stuff  will get synced because users will have to assign rights after environment creation and 2) automatic sync will create the users after some delay anyways.

So... force may not be necessary in the future. I think that points 2 and 3 in the commit message are still useful, but are not tied to force and can be delayed.

giladwolff:
> > > > I don't mind this going in, though I am not sure why we need this. Before the environment is created nobody is assigned to it, right? So this first sync is basically "mostly empty", no? Except for group names and WAG names?
> > > 
> > > 
> > > we do have groups, but won't have WAGs until we add the listWorkloadAdministrationGroups call to the sync.
> > > Right now we still have the incorrect role checking instead of rights checking so we currently have a bunch of users being synced. We already invoke this sync but it can get rejected. This commit fixes it so that the sync can't get rejected.
> > 
> > 
> > Bear with me, as I am still not very familiar with how things work. So apologies if you explained it to me and I forgot/misunderstood. This commit will force sync on environment creation, right? Are you saying that because the authorization checks are wrong this is useful because we will get the power users in?
> 
> sort of. Current behavior is wrong and broken and this fixes the broken part of it but not the wrong part of it. I don't mind delaying this commit until the next release because in the future 1) less stuff will get synced because users will have to assign rights after environment creation and 2) automatic sync will create the users after some delay anyways.
> 
> So... force may not be necessary in the future. I think that points 2 and 3 in the commit message are still useful, but are not tied to force and can be delayed.

Yes, I agree that (2) and (3) are useful. I would take out the triggering of the sync on environment creation, but I'll leave it up to you.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/AvailableStackProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.stack.StackService;
+
+@Component
+public class AvailableStackProvider {
```
lacikaaa: I would do the filtering on db side with a query so we won't move the unnecessary amount of data to the app and build object from them just to drop it right away

handavid: good suggestion!

PR #6570, opened 2019-10-24
CB-4166 Knox - gateway.knox.admin.groups should be configured to the admin group

risdenk:
Uses the configured LDAP admin group to set the `gateway.knox.admin.groups` property for both Knox and IDBroker.

Follows a similar pattern to what was done for Zeppelin in CB-3094


risdenk:
Hmmm looks like I have checkstyle issues. let me take a look.

risdenk:
This will conflict with PR #6571 most likely since they change the same area of code. Which ever PR gets merged first I can fix the conflicts.

PR #6571, opened 2019-10-24
CB-4164 Knox - default.app.topology.name is an exposed config instead of safety valve

risdenk:
Configure Knox `default.app.topology.name` instead of through safety valve with other configs


risdenk:
This will conflict with PR #6570 most likely since they change the same area of code. Which ever PR gets merged first I can fix the conflicts.

pdarvasi:
I will merge once the conflicts are resolved

risdenk:
Pushed update to resolve merge conflicts from PR #6570

PR #6572, opened 2019-10-24
CB-4167 Enable Nginx websockets proxying for Knox/Zeppelin

risdenk:
Add config for Nginx to enable websockets when proxying Knox/Zeppelin


biharitomi:
Thanks @risdenk !

PR #6577, opened 2019-10-25
CB-4073 Introduce v3 rc default catalog

horadla23:


lnardai:
We need to find out a solution to the case where we have the following scenario:
Version2.16-rc.21 on mow-dev
Version2.16-rc.10 on mow-int

mow-int would lose stability since we would upgrade images not verified by QA.

PR #6579, opened 2019-10-25
CB-4157 Add CDP Azure entitlement check to env creation

lajosrodek:


akanto:
Did you raise a jira to make sure that this entitlement is enabled on Cloudera tenant.

lajosrodek:
@akanto Good point, almost forgot. Just raised CDPCP-1002 to track this.

PR #6581, opened 2019-10-25
CB-4189 Datalake on azure should not create external db by default

gergopapi2:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -221,12 +222,14 @@ public Long getResourceIdByResourceName(String resourceName) {
         return getSdxByNameInAccount(userCrn, resourceName).getId();
     }
 
-    private void createDatabaseByDefaultForAWS(SdxClusterRequest sdxClusterRequest, SdxCluster sdxCluster, DetailedEnvironmentResponse environment) {
-        //if (AWS.equals(environment.getCloudPlatform()) &&
-        if (dbServiceSupportedPlatforms.contains(environment.getCloudPlatform()) &&
+    @VisibleForTesting
```
lnardai: Is this a new trend? Need to read on it.

lnardai:
@bhavanki 
Yes, but I think that would not allow embedded db as well.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
gergopapi2: Normally if I have the need to test a private method I would extract it to another class and write tests for it there. However, I saw @VisibleForTesting in the code at several places and seemed a better alternative:
1) code that needs to work together is still in one class => better readability
2) less work

If, however, there are too many methods tagged with @VisibleForTesting then it would be a sign for me to refactor that class into several smaller ones.

PR #6587, opened 2019-10-25
dh custom instance types

doktoric:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -147,16 +146,28 @@
...
+    @Value("#{'${cb.aws.distrox.disabled.instance.types:}'.split(',')}")
+    private List<String> disabledDistroxInstanceTypes;
+
     private final Predicate<VmType> enabledInstanceTypeFilter = vmt -> disabledInstanceTypes.stream()
             .filter(it -> !it.isEmpty())
             .noneMatch(di -> vmt.value().startsWith(di));
 
+    private final Predicate<VmType> enabledDistroxInstanceTypeFilter = vmt -> disabledDistroxInstanceTypes.stream()
+            .filter(it -> !it.isEmpty())
+            .map(it -> it.toLowerCase().trim().replaceAll("\\s+", ""))
```
gregito: could you please extract this replacement to a method with an example what's going to be replaced?

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/GetPlatformVmTypesHandler.java
```diff
@@ -38,9 +39,16 @@ public void accept(Event<GetPlatformVmTypesRequest> getPlatformVmTypesRequest) {
             CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(
                     Platform.platform(request.getExtendedCloudCredential().getCloudPlatform()),
                     Variant.variant(request.getVariant()));
-            CloudVmTypes platformVirtualMachinesJson = cloudPlatformConnectors.get(cloudPlatformVariant)
-                    .platformResources()
-                    .virtualMachines(request.getCloudCredential(), Region.region(request.getRegion()), request.getFilters());
+            CloudVmTypes platformVirtualMachinesJson;
+            if (CdpResourceType.DATAHUB.equals(request.getCdpResourceType())) {
+                platformVirtualMachinesJson = cloudPlatformConnectors.get(cloudPlatformVariant)
+                        .platformResources()
```
gregito: this can be extracted to line 42, since this call is the same in both sides of the if-else block

common-model/src/main/java/com/sequenceiq/common/api/type/CdpResourceType.java
```diff
@@ -0,0 +1,17 @@
...
+
+    DEFAULT;
+
+    public static CdpResourceType fromStackType(String type) {
+        if (type.toLowerCase().contains("ditrox") || type.toLowerCase().contains("workload")) {
+            return DATAHUB;
+        } else if (type.toLowerCase().contains("sdx") || type.toLowerCase().contains("datalake")) {
+            return DATALAKE;
+        }
+        return DEFAULT;
```
gregito: what is a default resource type? just askin'. if there is no such type, maybe we should handle it as an unidentified type with some exception

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3GuardConfigProvider.java
```diff
@@ -0,0 +1,83 @@
...
+            hdfsCoreSiteSafetyValveValue.append(ConfigUtils
+                    .getSafetyValveProperty(S3GUARD_TABLE_TAG_PARAM + entry.getKey(), entry.getValue()));
+        }
+        hdfsCoreSiteSafetyValveValue.append(ConfigUtils
+                .getSafetyValveProperty(S3GUARD_TABLE_TAG_PARAM + CDP_TABLE_ROLE, S3GUARD_TABLE_TAG_VALUE));
+    }
+
+    private boolean isS3GuardTableConfigured(S3FileSystemConfigurationsView s3FileSystemConfigurationsView) {
+        return StringUtils.isNotBlank(s3FileSystemConfigurationsView.getS3GuardDynamoTableName());
+    }
```
gregito: please cover it with unit tests

common-model/src/main/java/com/sequenceiq/common/api/type/CdpResourceType.java
doktoric: like freeipa etc..

gregito: Ah I see!

PR #6589, opened 2019-10-25
CDPD-4978 - updated blueprints with new configuration

jasonwmcswain:
add new config to hdfs name node to support long hostnames.

`
{
"name": "namenode_config_safety_valve",
"value": "<property><name>dfs.namenode.fs-limits.max-component-length</name><value>2048</value></property>"
}
`

- Testing
in order to validate, i did a dp sdx-internal create with this blueprint.
verified provision correctly.
verified no errors in CM.
verified config change in CM > HDFS > configurations > "dfs.namenode.fs-limits.max-component-length"

- https://github.com/hortonworks/cloudbreak/commit/888b2e2a8887def28a24e26efcd6d9ca5863733a

If your PR resolves any open issue(s), please indicate them like this so they will be closed when your PR is merged:

Closes # CDPD-4978

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-sdx-702.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-sdx-702.bp
@@ -197,7 +197,13 @@
           {
             "base": true,
             "refName": "hdfs-NAMENODE-BASE",
-            "roleType": "NAMENODE"
+            "roleType": "NAMENODE",
+            "configs": [
+              {
+                "name": "namenode_config_safety_valve",
+                "value": "<property><name>dfs.namenode.fs-limits.max-component-length</name><value>2048</value></property>"
+              }
+            ]
           },
           {
             "base": true,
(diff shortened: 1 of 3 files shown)


jasonwmcswain:
I'm waiting for feedback in https://jira.cloudera.com/browse/CDPD-4978 to determine if this should be merged to master (CDPM9) or XYZ for (CDPM8)?

risdenk:
There is a mix of safety valves already in play. Need to make sure these aren't overwritten. The ones in CB already relating to DynamoDB and S3 configs are super important that they still exist.

Probably makes more sense to set this in CB itself if it needs to be there? 

https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hdfs/HdfsConfigProvider.java#L54

lacikaaa:
closing this as there hasn't been any activity in the last month. please reopen if necessary

PR #6591, opened 2019-10-28
CDPSDX-1080. Register FreeIPA servers to cluster proxy with tunnelling

aarman-cloudera:
When CCM is enabled we now register FreeIPA server with the necessary
tunnelling configs to cluster proxy.

### DIFF
--- a/config/checkstyle/checkstyle-suppressions.xml
+++ b/config/checkstyle/checkstyle-suppressions.xml
@@ -23,7 +23,7 @@
     <suppress checks="ParameterNumber" files="UserDataBuilder.java|ServiceEndpointCollector.java"/>
     <suppress checks="EmptyLineSeparator" files=".*Description.java|.*Descriptions.java|Notes.java"/>
     <suppress checks=".*" files="PrometheusResponse.java"/>
-    <suppress checks="CyclomaticComplexity" files="PrometheusEvaluator.java|MetadataSetupService.java|AzureRoleManager.java|DBStackToDatabaseStackConverter.java|ClusterServiceConfig.java"/>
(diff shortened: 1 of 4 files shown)


config/checkstyle/checkstyle-suppressions.xml
```diff
@@ -23,7 +23,7 @@
     <suppress checks="ParameterNumber" files="UserDataBuilder.java|ServiceEndpointCollector.java"/>
     <suppress checks="EmptyLineSeparator" files=".*Description.java|.*Descriptions.java|Notes.java"/>
     <suppress checks=".*" files="PrometheusResponse.java"/>
-    <suppress checks="CyclomaticComplexity" files="PrometheusEvaluator.java|MetadataSetupService.java|AzureRoleManager.java|DBStackToDatabaseStackConverter.java"/>
+    <suppress checks="CyclomaticComplexity" files="PrometheusEvaluator.java|MetadataSetupService.java|AzureRoleManager.java|DBStackToDatabaseStackConverter.java|ClusterServiceConfig.java"/>
```
lacikaaa: please don't add exception if not really necessary. usually CyclomaticComplexity can be workaround with some refactor. Or if it's only one method, you can use `@SuppressWarnings` annotation, so the rest of the class would be still validated

aarman-cloudera: Good point. This warning is coming from the `equals()` method which I don't think should be refactored. 

Just tried using `@SuppressWarnings` but I couldn't get it working with this checkstyle config.

Using `//CHECKSTYLE:OFF: CyclomaticComplexity` however worked.

PR #6600, opened 2019-10-29
Cherry pick CB-4166 and CB-4164 back to rc-2.15

risdenk:
Cherry pick PR #6570 and PR #6571 from master back to rc-2.15

This is as one PR since they conflicted going into master. Easier to cherry-pick together.


akanto:
Does it work with a bit older CM/CDP, say 7.0.1?

risdenk:
@akanto yes should work for almost any old CDP CM version. `gateway.knox.admin.groups` was added from the very beginning of the Knox CSD around April 2019. `default.app.topology.name` was added around July 4th so should be there. These are both in CM so no concern with CDH versions.

PR #6602, opened 2019-10-30
CDPCP-1017. Entitle automatic usersync poller

handavid:
The UmsClient and EntitlementService were modified to accept an accountId
in addition to an actorCrn. This change is needed to support calling
with an internal actor crn.

The UsersyncPoller was modified to check the CDP_AUTOMATIC_USERSYNC_POLLER
entitlement for each account.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
@@ -17,6 +17,7 @@ import org.springframework.stereotype.Service;
 import com.google.common.annotations.VisibleForTesting;
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
 import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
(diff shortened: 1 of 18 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -77,19 +81,25 @@ void syncFreeIpaStacks() {
...
-                                        LOGGER.debug("Environment {} in Account {} is stale.", stack.getEnvironmentCrn(), stack.getAccountId());
-                                        SyncOperationStatus status = userService.synchronizeUsers(stack.getAccountId(), INTERNAL_ACTOR_CRN,
-                                                Set.of(stack.getEnvironmentCrn()), Set.of(), Set.of());
-                                        LOGGER.debug("Sync request resulted in operation {}", status);
-                                    } else {
-                                        LOGGER.debug("Environment {} in Account {} is up-to-date.", stack.getEnvironmentCrn(), stack.getAccountId());
-                                    }
-                                });
+                        String accountId = stringListEntry.getKey();
+                        if (entitlementService.automaticUsersyncPollerEnabled(INTERNAL_ACTOR_CRN, accountId)) {
```
giladwolff: super nit: I would reverse the check and return if it's not entitled and gain one level of indentation.

handavid: opened https://github.com/hortonworks/cloudbreak/pull/6610 to respond to this comment

PR #6606, opened 2019-10-30
CB-3939 Vault read/writes should be retryable and timeout configurable

bbihari:


secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/VaultRetryService.java
```diff
@@ -0,0 +1,59 @@
...
+                    multiplierExpression = "${vault.retry.multiplier:2}",
+                    maxDelayExpression = "${vault.retry.maxdelay:10000}")
+    )
+    public <T> T testVaultRead(Supplier<T> action) throws Retry.ActionWentFailException {
+        try {
+            return action.get();
+        } catch (RuntimeException e) {
+            LOGGER.error("Exception during reading vault", e);
+            metricService.incrementMetricCounter(MetricType.VAULT_READ_FAILED);
+            throw new Retry.ActionWentFailException(e.getMessage());
```
attilapalfi92: Instead of WentFail which is improper in english, please rename it to ActionFailedException

secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/VaultRetryService.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Retryable(
+            value = Retry.ActionWentFailException.class,
+            maxAttemptsExpression = "${vault.retry.maxattempt:5}",
+            backoff = @Backoff(delayExpression = "${vault.retry.delay:2000}",
+                    multiplierExpression = "${vault.retry.multiplier:2}",
+                    maxDelayExpression = "${vault.retry.maxdelay:10000}")
+    )
+    public <T> T testVaultRead(Supplier<T> action) throws Retry.ActionWentFailException {
```
attilapalfi92: Why `test`? Simply calling it `read` and other method `write` would be more straightforward in my opinion. The service name already contains the information that it retries every operation. I don't understand why `test` is applicable here.

secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/VaultRetryService.java
bbihari: ok

secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/VaultRetryService.java
bbihari: ok

PR #6608, opened 2019-10-30
CDPD-3643 - Configure DAS in DE template.

szlta:
- Adding DAS in DE template.
- Fix DAS service name in ExposedServices.
- Fix Tez proto base directory in DE template.

Closes #6548 
Closes #CDPD-3643

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
@@ -116,13 +116,7 @@
           {
             "refName": "tez-GATEWAY-BASE",
             "roleType": "GATEWAY",
-            "base": true,
-            "configs": [
-              {
-                "name": "tez-conf/tez-site.xml_client_config_safety_valve",
-                "value": "<property><name>tez.am.history.logging.enabled</name><value>false</value></property>"
-              }
-            ]
+            "base": true
           }
         ]
       },
@@ -252,6 +246,32 @@
         ]
       },
       {
+        "refName": "das",
+        "serviceType": "DAS",
+        "roleConfigGroups": [
+          {
+            "refName": "das-DAS_WEBAPP",
+            "roleType": "DAS_WEBAPP",
+            "base": true,
+            "configs": [
+              {
+                "name": "data_analytics_studio_user_authentication",
+                "value": "KNOX_PROXY"
+              },
+              {
+                "name": "data_analytics_studio_admin_users",
+                "value": "*"
+              }
+            ]
+          },
+          {
+            "refName": "das-DAS_EVENT_PROCESSOR",
+            "roleType": "DAS_EVENT_PROCESSOR",
+            "base": true
+          }
+        ]
+      },
+      {
         "refName": "oozie",
         "serviceType": "OOZIE",
         "roleConfigGroups": [
@@ -285,7 +305,9 @@
           "oozie-OOZIE_SERVER-BASE",
           "yarn-JOBHISTORY-BASE",
           "yarn-RESOURCEMANAGER-BASE",
-          "zookeeper-SERVER-BASE"
+          "zookeeper-SERVER-BASE",
+          "das-DAS_WEBAPP",
+          "das-DAS_EVENT_PROCESSOR"
         ]
       },
       {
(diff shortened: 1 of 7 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java
```diff
@@ -23,9 +23,8 @@
         String realm = kerberosConfigOpt.getRealm();
         String keytab = ConfigUtils.getSafetyValveProperty("hive.server2.authentication.spnego.keytab", "hive.keytab");
         String principal = ConfigUtils.getSafetyValveProperty("hive.server2.authentication.spnego.principal", "HTTP/_HOST@" + realm);
-        return List.of(config(HIVE_SERVICE_CONFIG_SAFETY_VALVE,
-                principal
-                        + keytab));
+        String filePerEvent = ConfigUtils.getSafetyValveProperty("hive.hook.proto.file.per.event", "true");
```
risdenk: Is `filePerEvent` being used anywhere? I see `principal` and `keytab` being used on line 27. I would assume `filePerEvent` needs to be returned in the list somehow?

risdenk: Looks like it is being set differently for the Hive metastore below so not sure if it needs to be set here.

core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
```diff
@@ -116,13 +116,7 @@
           {
             "refName": "tez-GATEWAY-BASE",
             "roleType": "GATEWAY",
-            "base": true,
-            "configs": [
-              {
-                "name": "tez-conf/tez-site.xml_client_config_safety_valve",
-                "value": "<property><name>tez.am.history.logging.enabled</name><value>false</value></property>"
```
risdenk: Is this being set by CM now? Was this supposed to be removed?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hive/HiveKnoxConfigProvider.java
horadla23: I copied the change from old PR incorrectly, will fix, will add filePerEvent

core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
```diff
@@ -251,6 +245,32 @@
...
+        "refName": "das",
+        "serviceType": "DAS",
+        "roleConfigGroups": [
+          {
+            "refName": "das-DAS_WEBAPP",
+            "roleType": "DAS_WEBAPP",
+            "base": true,
+            "configs": [
+              {
+                "name": "data_analytics_studio_user_authentication",
```
doktoric: I think we should add this to the configprovider as well because probably every cluster which using DAS should include this property. cc @horadla23

core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
harishjp: This was set to disable logging last time because of some bug. This should not be set for DAS to work.

horadla23: Ok, so the change here is to remove this property

core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
harishjp: Moved.

horadla23:
@harishjp checkstyle-main CI job failed, please fix it and then this should be ok

horadla23:
I have started a distrox cluster with these changes and with the modified bp, so LGTM

doktoric:
LGTM

PR #6609, opened 2019-10-31
CDPCP-1023. 

handavid:
The FreeIpaClient returns a null response object when the cluster-proxy request times out. The first commit null-checks the response and throws an exception. The second commit partitions the known potentially long-running requests into smaller chunks of work.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
@@ -485,6 +485,11 @@ public class FreeIpaClient {
         try {
             RPCResponse<T> response = (RPCResponse<T>) jsonRpcHttpClient.invoke(method, List.of(flags, parameterMap), type);
             LOGGER.debug("Response object: {}", response);
+            if (response == null) {
+                // TODO CDPCP-1028 investigate why invoke returns null instead of throwing an exception
+                // when the cluster-proxy request times out.
+                throw new NullPointerException("JSON-RPC response is null");
+            }
             return response;
         } catch (Throwable throwable) {
             String message = String.format("Invoke FreeIpa failed: %s", throwable.getLocalizedMessage());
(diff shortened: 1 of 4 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -485,6 +486,11 @@ public Host showHost(String fqdn) throws FreeIpaClientException {
         try {
             RPCResponse<T> response = (RPCResponse<T>) jsonRpcHttpClient.invoke(method, List.of(flags, parameterMap), type);
             LOGGER.debug("Response object: {}", response);
+            if (response == null) {
+                // TODO CDPCP-1028 investigate why invoke returns null instead of throwing an exception
+                // when the cluster-proxy request times out.
+                throw new NullPointerException("JSON-RPC response is null");
```
keyki: Shouldn't we throw a freeipa client exception?

handavid: it gets immediately wrapped and thrown as a FreeIpaClientException on line 498.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -485,6 +485,11 @@ public Host showHost(String fqdn) throws FreeIpaClientException {
         try {
             RPCResponse<T> response = (RPCResponse<T>) jsonRpcHttpClient.invoke(method, List.of(flags, parameterMap), type);
             LOGGER.debug("Response object: {}", response);
+            if (response == null) {
```
giladwolff: as a suggestion, you can start using checkState from guava to do this kind of thing in the future. It's more concise.

handavid: noted. thanks!

PR #6616, opened 2019-10-31
CB-3133. Resource role assignment is not cleaned up after environment deletion

oleewere:
- added ums resource delete step for the environment (non critical error - happens right before finish state)


environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/EnvDeleteActions.java
```diff
@@ -211,6 +212,23 @@ protected void doExecute(CommonContext context, EnvDeleteFailedEvent payload, Ma
...
+    @Bean(name = "UMS_RESOURCE_DELETE_STARTED_STATE")
+    public Action<?, ?> umsResourceDeleteAction() {
+        return new AbstractEnvDeleteAction<>(EnvDeleteEvent.class) {
+
+            @Override
+            protected void doExecute(CommonContext context, EnvDeleteEvent payload, Map<Object, Object> variables) {
+                EnvironmentStatus environmentStatus = EnvironmentStatus.UMS_RESOURCE_DELETE_IN_PROGRESS;
+                ResourceEvent resourceEvent = ResourceEvent.ENVIRONMENT_UMS_RESOURCE_DELETION_STARTED;
+                EnvDeleteState envDeleteState = EnvDeleteState.UMS_RESOURCE_DELETE_STARTED_STATE;
+                String logDeleteState = "Assigned UMS resources";
```
doktoric: @oleewere are you sure that this message is correct?

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/EnvironmentUMSResourceDeleteHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.environment.flow.deletion.event.EnvDeleteEvent;
+import com.sequenceiq.flow.reactor.api.event.EventSender;
+import com.sequenceiq.flow.reactor.api.handler.EventSenderAwareHandler;
+
+import reactor.bus.Event;
+
+@Component
+public class EnvironmentUMSResourceDeleteHandler extends EventSenderAwareHandler<EnvironmentDto> {
```
doktoric: Can you please cover it with unit tests?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
```diff
@@ -19,6 +19,8 @@
 
     CLUSTER_DEFINITION_CLEANUP_PROGRESS,
 
+    UMS_RESOURCE_DELETE_IN_PROGRESS,
```
doktoric: This requires an API update in the UI. Please mention here the UI pull request as well.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/EnvironmentUMSResourceDeleteHandler.java
oleewere: ok, in progress..

oleewere: done

oleewere:
few notes:
- only internal iam user can call notify resource deletion ums api call (turned out during testing against real UMS)
- resource crn is not passed through environment steps in the event objects, so i calling environment service find by id (which is also happening in another flow steps, like ClusterDefinitionDeleteHandler)...maybe would make sense fill resource crn in the events in the future, then these findbyid calls would be not required. (cc @attilapalfi92 )

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/EnvDeleteActions.java
lajosrodek: I suggest using lowercase except for proper names and abbreviations; so "assigned" instead of  "Assigned". This string will be inserted into the log message in line 243 below.

PR #6624, opened 2019-11-04
CB-4290 Enable auto restart for Hive on Tez

miklosgergely:
Auto restart should be on for Hive on Tez by default.


doktoric:
@miklosgergely can you please this on top of master?

PR #6630, opened 2019-11-04
CB-4085 Move all the E2E tests to the new AWS account

aszegedi:
@lnardai or @sodre90 please review then merge if it is OK for you.

### DIFF
--- a/integration-test/src/main/resources/testsuites/clusterawstestset.yaml
+++ b/integration-test/src/main/resources/testsuites/clusterawstestset.yaml
@@ -2,8 +2,8 @@ name: "AWS cluster tests"
 parallel: tests
 threadCount: 10
 parameters:
-  awsRegion: eu-west-1
-  awsAvailabilityZone: eu-west-1a
+  awsRegion: eu-central-1
+  awsAvailabilityZone: eu-central-1a
   awsCredentialName: autotesting-clusters-aws
 listeners:
   - com.sequenceiq.it.cloudbreak.newway.listener.FirstLastTestExecutionBehaviour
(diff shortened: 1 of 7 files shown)


integration-test/src/main/resources/testsuites/clusterawstestset.yaml
```diff
@@ -2,8 +2,8 @@ name: "AWS cluster tests"
 parallel: tests
 threadCount: 10
 parameters:
-  awsRegion: eu-west-1
```
attilapalfi92: I think this file is not used anymore. Instead edit the application.yml. Please correct me if I'm wrong.

integration-test/src/main/resources/testsuites/clusterostestset.yaml
```diff
@@ -2,8 +2,8 @@ name: "AWS cluster tests on different oses"
 parallel: tests
 threadCount: 4
 parameters:
-  awsRegion: eu-west-1
```
attilapalfi92: I think this file is not used anymore. Instead edit the application.yml. Please correct me if I'm wrong.

integration-test/src/main/resources/testsuites/clusterrecoverytestset.yaml
```diff
@@ -2,8 +2,8 @@ name: "AWS cluster recovery tests"
 parallel: tests
 threadCount: 2
 parameters:
-  awsRegion: eu-west-1
```
attilapalfi92: I think this file is not used anymore. Instead edit the application.yml. Please correct me if I'm wrong.

integration-test/src/main/resources/testsuites/clustertestset.yaml
```diff
@@ -2,8 +2,8 @@ name: Cluster_tests
 parallel: tests
 threadCount: 4
 parameters:
-  awsRegion: eu-west-1
```
attilapalfi92: I think this file is not used anymore. Instead edit the application.yml. Please correct me if I'm wrong.

PR #6633, opened 2019-11-05
CLOUD-4180 Using different virtual admin groups for cm and cdp services

schfeca75:
@keyki or @akanto pls review

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
@@ -1,0 +1,76 @@
+package com.sequenceiq.cloudbreak.auth.altus;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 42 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public Map<UmsRight, String> createVirtualGroups(String accountId, String environmentCrn) {
+        Map<UmsRight, String> virtualGroups = new HashMap<>();
+        for (UmsRight right : UmsRight.values()) {
+            virtualGroups.put(right, createOrGetVirtualGroup(accountId, environmentCrn, right.getRight()));
+        }
+        return virtualGroups;
+    }
+
+    public String getVirtualGroup(String adminGroup, String accountId, String environmentCrn, String right) {
```
keyki: We could create a new method which takes the VirtualGroupView as parameter and we can reduce the code duplications for each invocation with the same parameters
```
virtualGroupService.getVirtualGroup(virtualGroupView.getAdminGroup(), virtualGroupView.getAccountId(),
                            virtualGroupView.getEnvironmentCrn(),...
```

PR #6636, opened 2019-11-05
CB-4085 Move all the E2E tests to the new AWS account

aszegedi:
Please review. If it is OK for you, please merge.


integration-test/application.yml
```diff
@@ -47,14 +47,14 @@ integrationtest:
...
   aws:
     defaultBlueprintName: "CDP 1.1 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie"
-    region: eu-west-1
-    location: eu-west-1
-    availabilityZone: eu-west-1a
-    vpcId: vpc-071202b5656f08a83
+    region: eu-central-1
+    location: eu-central-1
+    availabilityZone: eu-central-1a
+    vpcId: vpc-0fc0a422f82ea8eec
```
attilapalfi92: This file is a leftover redundancy and not in use. In rc-2.15 it's already removed. Maybe can you remove here as well?

PR #6639, opened 2019-11-05
CB-3513 IDBroker Mapping Validation

risdenk:
Adds IDBroker mapping validation for the following scenarios:
* User provided mappings with IDBMMS (does not validate mock mappings)
* Checks IDBroker instance profile exists
* Checks IDBroker instance profile has EC2 trust - ability to be attached to EC2 instance
* Checks IDBroker instance profile role has ability to assume all provided mappings
* Checks log instance profile exists
* Checks log instance profile has EC2 trust - ability to be attached to EC2 instance
* Checks data access user mapped roles for proper permissions (compared against policy with simulate)
* Checks baseline / ranger audit user mapped roles for proper AWS permissions (compared against policy with simulate)

The validation happens before the RDS is created with redbeams and returns an error message that explains what is missing.

Potential follow up items:
* Validate for distrox
* Consolidate other storage validation into one place (prefix and region validation)

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsDataAccessRolePermissionValidator.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsDataAccessRolePermissionValidator.java
@@ -1,0 +1,41 @@
+package com.sequenceiq.cloudbreak.cloud.aws.validator;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.service.identitymapping.AccountMappingSubject;
+import com.sequenceiq.common.api.cloudstorage.StorageLocationBase;
+import com.sequenceiq.common.model.FileSystemType;
+
+@Component
+public class AwsDataAccessRolePermissionValidator extends AwsIDBrokerMappedRolePermissionValidator {
+    @Override
+    Set<String> getUsers() {
+        return AccountMappingSubject.DATA_ACCESS_USERS;
+    }
+
+    @Override
+    List<String> getPolicyFileNames(boolean s3guardEnabled) {
+        List<String> policyFileNames = new ArrayList<>();
+        policyFileNames.add("aws-cdp-bucket-access-policy.json");
+        policyFileNames.add("aws-cdp-datalake-admin-s3-policy.json");
+        if (s3guardEnabled) {
+            policyFileNames.add("aws-cdp-dynamodb-policy.json");
+        }
+
+        return policyFileNames;
+    }
+
+    @Override
+    String getStorageLocationBase(StorageLocationBase location) {
+        return location.getValue().replace(FileSystemType.S3.getProtocol() + "://", "");
+    }
+
+    @Override
+    boolean checkLocation(StorageLocationBase location) {
+        return true;
+    }
+}
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerMappedRolePermissionValidator.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerMappedRolePermissionValidator.java
@@ -1,0 +1,155 @@
+package com.sequenceiq.cloudbreak.cloud.aws.validator;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+
+import com.amazonaws.auth.policy.Policy;
+import com.amazonaws.services.identitymanagement.AmazonIdentityManagement;
+import com.amazonaws.services.identitymanagement.model.EvaluationResult;
+import com.amazonaws.services.identitymanagement.model.Role;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.cloudbreak.cloud.aws.util.AwsIamService;
+import com.sequenceiq.cloudbreak.cloud.model.filesystem.CloudS3View;
+import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
+import com.sequenceiq.common.api.cloudstorage.AccountMappingBase;
+import com.sequenceiq.common.api.cloudstorage.StorageLocationBase;
+
+public abstract class AwsIDBrokerMappedRolePermissionValidator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AwsIDBrokerMappedRolePermissionValidator.class);
+
+    @Inject
+    private AwsIamService awsIamService;
+
+    /**
+     * Returns list of users to filter roles
+     * @return list of user strings
+     */
+    abstract Set<String> getUsers();
+
+    /**
+     * Returns policy file names
+     * @param s3guardEnabled true if s3guard is enabled false otherwise
+     * @return list of policy file names
+     */
+    abstract List<String> getPolicyFileNames(boolean s3guardEnabled);
+
+    /**
+     * Returns the storage location base string
+     * @param location StorageLocationBase to evaluate
+     * @return storage location base string
+     */
+    abstract String getStorageLocationBase(StorageLocationBase location);
+
+    /**
+     * Checks if the location should be checked or not for role permissions
+     * @param location StorageLocationBase to evaluate
+     * @return true if location is valid to check false otherwise
+     */
+    abstract boolean checkLocation(StorageLocationBase location);
+
+    /**
+     * Validates the cloudFileSystem
+     * @param cloudFileSystem cloud file system to evaluate
+     * @param validationResultBuilder builder for any errors encountered
+     */
+    public void validate(AmazonIdentityManagement iam, CloudS3View cloudFileSystem,
+                            ValidationResultBuilder validationResultBuilder) {
+        AccountMappingBase accountMappings = cloudFileSystem.getAccountMapping();
+        if (accountMappings != null) {
+            SortedSet<String> roleArns = getRoleArnsForUsers(getUsers(), accountMappings.getUserMappings());
+            Set<Role> roles = awsIamService.getValidRoles(iam, roleArns, validationResultBuilder);
+
+            boolean s3guardEnabled = cloudFileSystem.getS3GuardDynamoTableName() != null;
+            List<String> policyFileNames = getPolicyFileNames(s3guardEnabled);
+
+            SortedSet<String> failedActions = new TreeSet<>();
+            for (StorageLocationBase location : cloudFileSystem.getLocations()) {
+                if (checkLocation(location)) {
+                    Map<String, String> replacements = getPolicyJsonReplacements(location, cloudFileSystem);
+                    List<Policy> policies = getPolicies(policyFileNames, replacements);
+                    for (Role role : roles) {
+                        List<EvaluationResult> evaluationResults = awsIamService.validateRolePolicies(iam,
+                            role, policies);
+                        failedActions.addAll(getFailedActions(role, evaluationResults));
+                    }
+                }
+            }
+            if (!failedActions.isEmpty()) {
+                validationResultBuilder.error(String.format("The role(s) (%s) don't have the required permissions:%n%s",
+                    String.join(", ", roles.stream().map(Role::getArn).collect(Collectors.toCollection(TreeSet::new))),
+                    String.join("\n", failedActions)));
+            }
+        }
+    }
+
+    /**
+     * Get the replacements necessary for the policy json
+     * @param location StorageLocationBase to get storage paths from
+     * @param cloudFileSystem CloudFileSystem to get dynamodb table name
+     * @return map of simple replacements for policy json
+     */
+    protected Map<String, String> getPolicyJsonReplacements(StorageLocationBase location,
+                                                            CloudS3View cloudFileSystem) {
+        String storageLocationBase = getStorageLocationBase(location);
+        String datalakeBucket = storageLocationBase.split("/", 2)[0];
+        return Map.ofEntries(
+            Map.entry("${STORAGE_LOCATION_BASE}", storageLocationBase),
+            Map.entry("${DATALAKE_BUCKET}", datalakeBucket),
+            Map.entry("${DYNAMODB_TABLE_NAME}", cloudFileSystem.getS3GuardDynamoTableName())
+        );
+    }
+
+    /**
+     * Retrieve the unique set of role arns for the list of provided users
+     * @param users list of users to filter by
+     * @param userMappings user mappings to filter on
+     * @return set of AWS roles
+     */
+    SortedSet<String> getRoleArnsForUsers(Collection<String> users,
+                                            Map<String, String> userMappings) {
+        return userMappings.entrySet().stream()
+                    .filter(m -> users.contains(m.getKey()))
+                    .map(Entry::getValue).collect(Collectors.toCollection(TreeSet::new));
+    }
+
+    /**
+     * Returns Policy objects after the policy json has had replacements
+     * @param policyFileNames policy file names to read before replacement
+     * @param replacements simple replacements to be made to the policy json
+     * @return list of Policy objects after replacements made to policy json
+     */
+    List<Policy> getPolicies(List<String> policyFileNames,
+                                Map<String, String> replacements) {
+        List<Policy> policies = new ArrayList<>(policyFileNames.size());
+        for (String policyFileName : policyFileNames) {
+            Policy policy = awsIamService.getPolicy(policyFileName, replacements);
+            if (policy != null) {
+                policies.add(policy);
+            }
+        }
+        return policies;
+    }
+
+    /**
+     * Finds all the denied results and generates a set of failed actions
+     * @param role Role that was being evaluated
+     * @param evaluationResults result of simulating the policy
+     */
+    SortedSet<String> getFailedActions(Role role, List<EvaluationResult> evaluationResults) {
+        return evaluationResults.stream()
+                    .filter(evaluationResult -> evaluationResult.getEvalDecision().toLowerCase().contains("deny"))
+                    .map(evaluationResult -> String.format("%s:%s:%s", role.getArn(),
+                        evaluationResult.getEvalActionName(), evaluationResult.getEvalResourceName()))
+                    .collect(Collectors.toCollection(TreeSet::new));
+    }
+}
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerObjectStorageValidator.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerObjectStorageValidator.java
@@ -1,0 +1,101 @@
+package com.sequenceiq.cloudbreak.cloud.aws.validator;
+
+import javax.inject.Inject;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import com.amazonaws.services.identitymanagement.AmazonIdentityManagement;
+import com.amazonaws.services.identitymanagement.model.InstanceProfile;
+import com.amazonaws.services.identitymanagement.model.Role;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.aws.util.AwsIamService;
+import com.sequenceiq.cloudbreak.cloud.model.SpiFileSystem;
+import com.sequenceiq.cloudbreak.cloud.model.filesystem.CloudFileSystemView;
+import com.sequenceiq.cloudbreak.cloud.model.filesystem.CloudS3View;
+import com.sequenceiq.cloudbreak.validation.ValidationResult;
+import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
+import com.sequenceiq.common.api.cloudstorage.AccountMappingBase;
+import com.sequenceiq.common.model.CloudIdentityType;
+
+@Component
+public class AwsIDBrokerObjectStorageValidator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AwsIDBrokerObjectStorageValidator.class);
+
+    @Inject
+    private AwsInstanceProfileEC2TrustValidator awsInstanceProfileEC2TrustValidator;
+
+    @Inject
+    private AwsIDBrokerAssumeRoleValidator awsIDBrokerAssumeRoleValidator;
+
+    @Inject
+    private AwsDataAccessRolePermissionValidator awsDataAccessRolePermissionValidator;
+
+    @Inject
+    private AwsRangerAuditRolePermissionValidator awsRangerAuditRolePermissionValidator;
+
+    @Inject
+    private AwsLogRolePermissionValidator awsLogRolePermissionValidator;
+
+    @Inject
+    private AwsIamService awsIamService;
+
+    public ValidationResult validateObjectStorage(AmazonIdentityManagement iam,
+                                                    SpiFileSystem spiFileSystem,
+                                                    ValidationResultBuilder resultBuilder) {
+        List<CloudFileSystemView> cloudFileSystems = spiFileSystem.getCloudFileSystems();
+        for (CloudFileSystemView cloudFileSystemView : cloudFileSystems) {
+            CloudS3View cloudFileSystem = (CloudS3View) cloudFileSystemView;
+            String instanceProfileArn = cloudFileSystem.getInstanceProfile();
+            InstanceProfile instanceProfile = awsIamService.getInstanceProfile(iam, instanceProfileArn,
+                resultBuilder);
+            if (instanceProfile != null) {
+                CloudIdentityType cloudIdentityType = cloudFileSystem.getCloudIdentityType();
+                if (CloudIdentityType.ID_BROKER.equals(cloudIdentityType)) {
+                    validateIDBroker(iam, instanceProfile, cloudFileSystem, resultBuilder);
+                } else if (CloudIdentityType.LOG.equals(cloudIdentityType)) {
+                    validateLog(instanceProfile, cloudFileSystem, resultBuilder);
+                }
+            }
+        }
+
+        return resultBuilder.build();
+    }
+
+    private void validateIDBroker(AmazonIdentityManagement iam, InstanceProfile instanceProfile,
+                                    CloudS3View cloudFileSystem, ValidationResultBuilder resultBuilder) {
+        awsInstanceProfileEC2TrustValidator.isTrusted(instanceProfile, resultBuilder);
+
+        Set<Role> allMappedRoles = getAllMappedRoles(iam, cloudFileSystem, resultBuilder);
+        awsIDBrokerAssumeRoleValidator.canAssumeRoles(iam, instanceProfile, allMappedRoles, resultBuilder);
+
+        awsDataAccessRolePermissionValidator.validate(iam, cloudFileSystem, resultBuilder);
+        awsRangerAuditRolePermissionValidator.validate(iam, cloudFileSystem, resultBuilder);
+    }
+
+    private void validateLog(InstanceProfile instanceProfile, CloudS3View cloudFileSystem,
+                                ValidationResultBuilder resultBuilder) {
+        awsInstanceProfileEC2TrustValidator.isTrusted(instanceProfile, resultBuilder);
+
+        //TODO - need to figure out how to get LOGS_LOCATION_BASE value
+        //awsLogRolePermissionValidator.validate(iam, instanceProfile, cloudFileSystem, resultBuilder);
+    }
+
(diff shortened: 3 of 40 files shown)


risdenk:
I'd appreciate any feedback as far as structure or integration goes. This is the first time trying to do cross microservice integration in cloudbreak. I tried to follow other examples but there might be better ways to do this.

risdenk:
My one concern is that the policies copied from the documentation [1] are too broad. This change will prevent clusters from coming up if the minimal set of permissions are not met.

[1] https://docs.cloudera.com/management-console/cloud/environments/topics/mc-idbroker-minimum-setup.html

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
...
+
+    private void checkRolePermissions(Collection<String> roleArns, List<String> policyFileNames,
+                                        String datalakeBucket, String storageLocationBase,
+                                        String s3GuardDynamoDBTableName) {
+        Set<String> failedActions = new HashSet<>();
+        for (String policyFileName : policyFileNames) {
+            try {
+                String policyJsonTemplate = getResourceFileAsString("definitions/" + policyFileName);
+                if (policyJsonTemplate != null) {
+                    String policyJson = policyJsonTemplate
```
doktoric: We already using handlebar in the project which is a templating engine. Can we use that here ?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
...
+            if (s3GuardDynamoDBTableName != null) {
+                policyFileNames.add("aws-cdp-dynamodb-policy.json");
+            } else {
+                s3GuardDynamoDBTableName = "";
+            }
+
+            for (StorageLocationBase location : cloudFileSystem.getLocations()) {
+                if (CloudStorageCdpService.RANGER_AUDIT.equals(location.getType())) {
+                    String storageLocationBase = location.getValue()
+                                                    .replace("s3a://", "")
```
doktoric: FileSystemType class contains the protocol maybe we should use that here

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+        return instanceProfile;
+    }
+
+    private void checkDataAccessRolePermissions(CloudS3View cloudFileSystem) {
+        // This list needs to match the IDBMMS list of data access users
+        // https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/
+        //  services/idbrokermappingmanagement/src/main/java/com/cloudera/thunderhead/
+        //  service/idbrokermappingmanagement/server/IdBrokerMappingManagementService.java#L109
+        List<String> dataAccessUsers = Arrays.asList("hbase", "hdfs", "hive", "impala", "yarn",
```
doktoric: maybe this list should be the same what we are using in AwsMockAccountMappingService?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
+package com.sequenceiq.cloudbreak.cloud.aws;
```
doktoric: Can we add some unit test for this class please?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
...
+
+    public void validateObjectStorage(List<CloudFileSystemView> cloudFileSystems) {
+        for (CloudFileSystemView cloudFileSystemView : cloudFileSystems) {
+            CloudS3View cloudFileSystem = (CloudS3View) cloudFileSystemView;
+            InstanceProfile instanceProfile = getInstanceProfile(cloudFileSystem.getInstanceProfile());
+            if (instanceProfile != null) {
+                CloudIdentityType cloudIdentityType = cloudFileSystem.getCloudIdentityType();
+                if (CloudIdentityType.ID_BROKER.equals(cloudIdentityType)) {
+                    checkInstanceProfileTrust(instanceProfile);
+                    checkIDBrokerAssumeRolePermissions(instanceProfile, cloudFileSystem);
```
attilapalfi92: A lot of permission checks are done here and if any of them fails an exception is thrown. However there might be other cases of insufficient permissions, and the user won't be notified about those, because an exception was thrown at the first failure. To mitigate this I suggest to use the com.sequenceiq.cloudbreak.validation.ValidationResult and it's builder inner class to collect all the errors, and at the end of this method this contains anything, then an exception can be thrown - now containing all the errors that this validator can find.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,258 @@
...
+                instanceProfile = iam.getInstanceProfile(instanceProfileRequest).getInstanceProfile();
+            } catch (NoSuchEntityException | ServiceFailureException e) {
+                throw new CloudConnectorException(String.format("Unable to find instance profile: %s",
+                    instanceProfileArn), e);
+            }
+        }
+        return instanceProfile;
+    }
+
+    private void checkDataAccessRolePermissions(CloudS3View cloudFileSystem) {
```
attilapalfi92: Each of these private methods are pretty large and do relatively complex checks. Extracting them to separate classes of their own, and then unit testing each of them would be great. Suggestion with referring to my previous comment: The input and output of the methods of these validator classes could be the `ValidationResultBuilder` in which you could collect the validation results.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Yea I had seen this used elsewhere. I will use that to collect all the errors up front.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: The list in `AwsMockAccountMappingService` is ALL the users which we don't want. We actually want to separate by data access and baseline/audit. The only place this logic exists today is in the IDBMMS service. The comment above tries to explain the difference.

The `MOCK_IDBROKER_USER_MAPPINGS` list in `AwsMockAccountMappingService` should be a superset of the `dataAccessUsers` list and the `rangerAuditUsers` list. Basically union the two lists together.

PS - I honestly wish the mock stuff was removed since it causes a disconnect with IDBMMS. @lajosrodek are there plans to remove the mock stuff? 

PPS - The more I deal w/ IDBMMS and Cloudbreak - it would be nice if IDBMMS wasn't a completely separate project in a separate repo :(

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: I looked into this. `handlebars` is only currently used in the `template-manager-core` module. It seems to be pulled in for building template objects like blueprints.

I don't really like the string replacement, but I wanted to keep the policy json files human readable. This should allow @dbialek2 to reference/embed them in the documentation (https://docs.cloudera.com/management-console/cloud/environments/topics/mc-idbroker-minimum-setup.html) instead of docs having to keep a separate copy of the policy json files.

ie: https://raw.githubusercontent.com/risdenk/cloudbreak/CB-3513/cloud-aws/src/main/resources/definitions/aws-cdp-bucket-access-policy.json (eventually will be in CB repo instead of in my fork but same idea)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Another thought is to set the data access and ranger audit users in `AwsMockAccountMappingService`. Then literally union them together to create the "mock" list. This would match a lot closer to what IDBMMS is doing.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Latest pushed change removes hardcoded `s3a`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Latest push collects validation results all at once.

risdenk:
Latest push fixed some of the review comments. I am still working on adding more tests to this. Wanted to push to make sure not to lose local work.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
lajosrodek: @risdenk I wish the mock impl was removed some day too. 😄 I think it is still useful (actually essential) for integration testing CB when IDBMMS is likely not running to have a complete E2E setup. @doktoric may confirm this.

lajosrodek: @risdenk Regarding the (overly broad) user list in `AwsMockAccountMappingService`: Agreed, it should be indeed simplified to match the current data access & ranger audit lists in IDBMMS. We chose to include all possible daemon users for laziness, and also because the authoritative list of IDBMMS had not yet been finalized by that time. Shall I raise a separate PR for that, or would you prefer covering that change here?
Fun fact: there is also `AzureMockAccountMappingService` containing yet another duplicate of the above jumbo user list. 😄 

CC @bergerdenes @pdarvasi

risdenk: Lets do a separate change since it affects AWS and Azure.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,302 @@
...
+                resultBuilder);
+            if (instanceProfile != null) {
+                CloudIdentityType cloudIdentityType = cloudFileSystem.getCloudIdentityType();
+                if (CloudIdentityType.ID_BROKER.equals(cloudIdentityType)) {
+                    checkInstanceProfileTrust(instanceProfile, resultBuilder);
+                    checkIDBrokerAssumeRolePermissions(instanceProfile, cloudFileSystem, resultBuilder);
+                    checkDataAccessRolePermissions(spiFileSystem, cloudFileSystem, resultBuilder);
+                    checkRangerAuditRolePermissions(spiFileSystem, cloudFileSystem, resultBuilder);
+                } else if (CloudIdentityType.LOG.equals(cloudIdentityType)) {
+                    checkInstanceProfileTrust(instanceProfile, resultBuilder);
```
lajosrodek: Though not related to IDBroker, it might be beneficial to check log bucket permissions as well as per `aws-cdp-log-policy.json`.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
doktoric: 👍

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,302 @@
...
+                    }
+                }
+            } catch (IOException e) {
+                LOGGER.debug("Unable to load policy json", e);
+            }
+        }
+        return failedActions;
+    }
+
+    private void checkIDBrokerAssumeRolePermissions(InstanceProfile instanceProfile,
```
attilapalfi92: Can you please create separate classes for each of these methods, and unit tests them separately? These are pretty complex and this class is getting too large with it's 300 lines.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: yea that was supposed to happen (hence why I copied the json file). I'll add that too. Looks like I missed it when going through the first time.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Yea I have this mostly refactored locally. Didn't push the latest yet. Will comment when finished refactoring into separate classes.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
lajosrodek: Agreed. I just raised CB-4438 to track this.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
attilapalfi92: Alrighty, thanks for the update!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: So this is harder than it looks. Not sure where the log location is stored. I put a placeholder in with a TODO to find how to pass in the log location.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: Just pushed with all the testing changes.

risdenk:
Pushed changes with refactoring into smaller classes with unit tests for each validator.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerObjectStorageValidator.java
```diff
@@ -0,0 +1,83 @@
...
+        //TODO - need to figure out how to get LOGS_LOCATION_BASE value
+        //AwsLogRolePermissionValidator.validate(iam, instanceProfile, cloudFileSystem, resultBuilder);
+    }
+
+    private void validateIDBroker(InstanceProfile instanceProfile, SpiFileSystem spiFileSystem,
+                                    CloudS3View cloudFileSystem, ValidationResultBuilder resultBuilder) {
+        AwsInstanceProfileEC2TrustValidator.isTrusted(instanceProfile, resultBuilder);
+
+        Set<Role> allMappedRoles = getAllMappedRoles(cloudFileSystem, resultBuilder);
+        AwsIDBrokerAssumeRoleValidator.canAssumeRoles(iam, instanceProfile, allMappedRoles, resultBuilder);
```
attilapalfi92: Some of the validators have static methods, some of them are instantiated. Neither of these are the preferred solution. I would suggest another approach:

In `AwsIDBrokerObjectStorageValidator` please make `AmazonIdentityManagement` a parameter of the `validateObjectStorage` method like here with the `AwsIDBrokerAssumeRoleValidator.canAssumeRoles` method. 
In all the validator classes make `iam, spiFileSystem` parameters of the `validate` method.

This way you won't have to instantiate any of these, because they will be stateless. And since we are writing stateless components, services, and not a collection on static methods, please make these methods non-static, and annotate them with @Component, and inject them wherever they are used.

risdenk: I think I addressed what you meant with the latest push.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsIDBrokerMappedRolePermissionValidator.java
```diff
@@ -0,0 +1,151 @@
...
+     * Get the replacements necessary for the policy json
+     * @param location StorageLocationBase to get storage paths from
+     * @param cloudFileSystem CloudFileSystem to get dynamodb table name
+     * @return map of simple replacements for policy json
+     */
+    protected Map<String, String> getPolicyJsonReplacements(StorageLocationBase location,
+                                                            CloudS3View cloudFileSystem) {
+        String storageLocationBase = getStorageLocationBase(location);
+        String datalakeBucket = storageLocationBase.split("/", 2)[0];
+        Map<String, String> replacements = new HashMap<>();
```
lajosrodek: Minor: A builder construct like `Map.of(Map.entry(key, value), ...)` would be more elegant / modern, provided the return value of `Map.put()` is irrelevant and an immutable map is acceptable.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsLogRolePermissionValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+@Component
+public class AwsLogRolePermissionValidator {
+    public void validate(AmazonIdentityManagement iam, InstanceProfile instanceProfile,
+                            CloudS3View cloudFileSystem, ValidationResultBuilder validationResultBuilder) {
+        SortedSet<String> failedActions = new TreeSet<>();
+
+        Map<String, String> replacements = new HashMap<>();
```
lajosrodek: Minor: Potentially replace with `Map.of(Map.entry(key, value))` like above.

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsIamUtilTest.java
```diff
@@ -0,0 +1,286 @@
...
+import com.amazonaws.services.identitymanagement.AmazonIdentityManagement;
+import com.amazonaws.services.identitymanagement.model.GetInstanceProfileRequest;
+import com.amazonaws.services.identitymanagement.model.GetInstanceProfileResult;
+import com.amazonaws.services.identitymanagement.model.GetRoleRequest;
+import com.amazonaws.services.identitymanagement.model.GetRoleResult;
+import com.amazonaws.services.identitymanagement.model.InstanceProfile;
+import com.amazonaws.services.identitymanagement.model.NoSuchEntityException;
+import com.amazonaws.services.identitymanagement.model.Role;
+import com.amazonaws.services.identitymanagement.model.ServiceFailureException;
+import org.junit.Assert;
```
lajosrodek: Could you please use JUnit 5 and assertJ (`Assertions.assertThat()`) for all new tests if possible? As I see, module `cloud-aws` lacks JUnit 5 dependencies at the moment, so feel free to add them (leaving JUnit 4 ones as well for any legacy tests). assertJ is already included in `spring-boot-starter-test`. Cloudbreak is moving away from JUnit 4, and assertJ assertions are more readable thanks to its fluent API.

Not an authoritative example, but my recent changes to `cloud-common/build.gradle` in #6700 can be simply copy-pasted. 😄

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsRangerAuditRolePermissionValidator.java
```diff
@@ -0,0 +1,50 @@
...
+import com.sequenceiq.common.model.CloudStorageCdpService;
+import com.sequenceiq.common.model.FileSystemType;
+
+@Component
+public class AwsRangerAuditRolePermissionValidator extends AwsIDBrokerMappedRolePermissionValidator {
+    // The data access and ranger audit lists need to match the IDBMMS lists
+    // https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/
+    //  services/idbrokermappingmanagement/src/main/java/com/cloudera/thunderhead/
+    //  service/idbrokermappingmanagement/server/IdBrokerMappingManagementService.java#L109
+    private static final List<String> RANGER_AUDIT_USERS = Arrays.asList("kafka", "solr", "knox",
```
lajosrodek: #6700 has been merged, so please free to use `AccountMappingSubject.RANGER_AUDIT_USERS` instead.

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsRangerAuditRolePermissionValidatorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import com.sequenceiq.common.model.CloudStorageCdpService;
+import com.sequenceiq.common.model.FileSystemType;
+
+public class AwsRangerAuditRolePermissionValidatorTest extends AwsIDBrokerMappedRolePermissionValidatorTest {
+    private final AwsRangerAuditRolePermissionValidator awsRangerAuditRolePermissionValidator =
+        new AwsRangerAuditRolePermissionValidator();
+
+    @Override
+    public void testGetUsers() {
+        List<String> expectedUsers = Arrays.asList("kafka", "solr", "knox", "atlas");
```
lajosrodek: #6700 has been merged, so please free to use `AccountMappingSubject.RANGER_AUDIT_USERS` instead.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsDataAccessRolePermissionValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import com.sequenceiq.common.api.cloudstorage.StorageLocationBase;
+import com.sequenceiq.common.model.FileSystemType;
+
+@Component
+public class AwsDataAccessRolePermissionValidator extends AwsIDBrokerMappedRolePermissionValidator {
+    // The data access and ranger audit lists need to match the IDBMMS lists
+    // https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/
+    //  services/idbrokermappingmanagement/src/main/java/com/cloudera/thunderhead/
+    //  service/idbrokermappingmanagement/server/IdBrokerMappingManagementService.java#L109
+    private static final List<String> DATA_ACCESS_USERS = Arrays.asList("hbase", "hdfs", "hive",
```
lajosrodek: #6700 has been merged, so please free to use `AccountMappingSubject.DATA_ACCESS_USERS` instead.

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/validator/AwsDataAccessRolePermissionValidatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.sequenceiq.common.model.CloudIdentityType;
+import com.sequenceiq.common.model.FileSystemType;
+
+public class AwsDataAccessRolePermissionValidatorTest extends AwsIDBrokerMappedRolePermissionValidatorTest {
+    private final AwsDataAccessRolePermissionValidator awsDataAccessRolePermissionValidator =
+        new AwsDataAccessRolePermissionValidator();
+
+    @Override
+    public void testGetUsers() {
+        List<String> expectedUsers = Arrays.asList("hbase", "hdfs", "hive",
```
lajosrodek: #6700 has been merged, so please free to use `AccountMappingSubject.DATA_ACCESS_USERS` instead.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
oleewere: it is stored on env side (globally, in db: Environment -> EnvironmentTelemetry -> EnvironmentLogging, from api, it's similar), like s3://cloudbreak-group/logs , for every specific cluster (DH/DL), a path is generated from the environment setting (like s3://cloudbreak-group/logs/datalake/... ) - the specific one is immutable

if the stack is already persisted (for SDX/DistroX) and you can access that, then it's inside the components table with the generated full path (as i mentioned above), for that, see: https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/service/ComponentConfigProviderService.java#L58

risdenk: yea @attilapalfi92 shared some of this detail with me as well. The issue is that its not already persisted. Trying to do this validation before hand (before waiting for RDS to be spun up). I see where the telemetry is being setup. Let me create a separate jira for this. It might not make sense to validate this in the same place.

risdenk:
Latest push addresses @lajosrodek comments about Junit5, using `Map.of...`, and using the updated mock mappings list which matches IDBMMS.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsIamUtil.java
```diff
@@ -0,0 +1,263 @@
...
+import com.amazonaws.services.identitymanagement.model.ServiceFailureException;
+import com.amazonaws.services.identitymanagement.model.SimulatePrincipalPolicyRequest;
+import com.amazonaws.services.identitymanagement.model.SimulatePrincipalPolicyResult;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
+
+public class AwsIamUtil {
```
attilapalfi92: Why is this a util, why not AwsIamService?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsIDBrokerObjectStorageValidator.java
risdenk: https://jira.cloudera.com/browse/CB-4476

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsIamUtil.java
risdenk: Because I definitely missed that one when I went through in a different package than I looked. Fixing it now.

risdenk: I definitely broke something with the latest push. The class aren't fully injected during the tests somehow. I end up with NPE when trying to use the `AwsIamService`

risdenk: Attila helped me out explaining how `@Spy` was needed since the nested injection doesn't work in the tests.

risdenk:
Thanks for the reviews. I'm taking a look at the remaining integration test failures.

risdenk:
Think that latest push should fix the `MockSdxTests` tests. Can't easily test this locally since requires credentials it looks like to do the integration tests. This might fix the `DistroX` test as well since the fix was to make sure there was cloud storage before validating it.

risdenk:
Woot all green :)

doktoric:
@risdenk please rebase and I will merge it

risdenk:
Rebased changes on latest master

risdenk:
@doktoric / @attilapalfi92 I merged this with reviews and latest tests passing using "Rebase and merge". Thanks for the reviews!

PR #6642, opened 2019-11-05
CB-4289 Knox files should not be world writable

risdenk:
Remove world writable privileges for the Knox files


pdarvasi:
We might consider including this in the RC branch, too.

risdenk:
@pdarvasi moved the commit to rc-2.15 and changed the PR base.

PR #6643, opened 2019-11-05
CB-4336 Knox - Misc topologies cleanup

risdenk:
* Remove `;maxRetryAttempts=300;retrySleep=1000` since it has no effect after KNOX-1093
* Remove explicit `:443` from https url
* Remove unnecessary `<uri>` section


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/cdp-token.xml.j2
```diff
@@ -1,16 +1,14 @@
...
    <gateway>
-
       <provider>
          <role>federation</role>
          <name>SSOCookieProvider</name>
          <enabled>true</enabled>
          <param>
             <name>sso.authentication.provider.url</name>
-            <value>https://{{ salt['pillar.get']('gateway:address') }}:443/{{ salt['pillar.get']('gateway:path') }}/knoxsso/api/v1/websso</value>
+            <value>/{{ salt['pillar.get']('gateway:path') }}/knoxsso/api/v1/websso</value>
```
risdenk: Matches other `sso.authentication.provider.url` in topologies

PR #6646, opened 2019-11-06
CDPD-5555 increasing the nginx timeout to 600sec

akanto:


akanto:
I have added 600 first, and after that I have seen that in CDPD-5555 they are asking for 300s only, and I fixed the code, but forget to fix the message

PR #6650, opened 2019-11-06
CB-4299 Stack listing and posting optimisations

daszabo:


structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/rest/RestRequestDetails.java
```diff
@@ -19,7 +19,7 @@
 
     private Map<String, String> cookies;
 
-    @JsonSerialize(using = AnonymizingBase64Serializer.class)
+    @JsonSerialize(using = Base64Serializer.class)
```
akanto: Rest request can have sensitive data in it

akanto: @bbihari is taking a look a the performance impact of AnonymizingBase64Serializer

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -180,15 +182,8 @@ private void updateCustomDomainOrKerberos(StackV4Request source, Stack stack) {
...
-            kerberosConfig.ifPresent(kb -> {
-                if (kb.getType() == KerberosType.ACTIVE_DIRECTORY) {
-                    if (isEmpty(kb.getRealm())) {
-                        throw new BadRequestException("Realm cannot be null in case of ACTIVE_DIRECTORY");
-                    }
-                    stack.setCustomDomain(kb.getRealm().toLowerCase());
-                }
-            });
+            String domain = freeIPAAdapter.getDomain(stack.getEnvironmentCrn());
+            stack.setCustomDomain(domain);
```
akanto: I think the original code set the customDomain only in case of ACTIVE_DIRECTORY, since AD is not supported I think the entire else if part can be deleted

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -323,14 +318,15 @@ private void validateStackAuthentication(StackV4Request source) {
     private void updateCluster(StackV4Request source, Stack stack, Workspace workspace) {
         if (source.getCluster() != null) {
             source.getCluster().setName(stack.getName());
-            Cluster cluster = getConversionService().convert(source.getCluster(), Cluster.class);
-            Set<HostGroup> hostGroups = source.getInstanceGroups().stream()
+            Cluster cluster = measure(() -> getConversionService().convert(source.getCluster(), Cluster.class),
```
akanto: measure() leads to unreadable code, I am not sure why this is preferred instead of annotating a method

core/src/main/java/com/sequenceiq/cloudbreak/freeipa/FreeIPAAdapter.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.FreeIpaV1Endpoint;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.describe.DescribeFreeIpaResponse;
+
+@Service
+public class FreeIPAAdapter {
```
akanto: FreeIpa is called FreeIpa everywhere in the code,  so I think it is better to change FreeIPA to FreeIpa

PR #6654, opened 2019-11-06
CB-4366 Nginx should allow underscores in header names

risdenk:
Nginx by default prevents underscores in header names.
The HTTP spec allows underscores in headers. This change
configures Nginx to allow underscores in headers. It needs
to be configured at the top level since there are multiple
Nginx server config blocks.

http://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers

Some additional background on underscores in headers can
be found here:

https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/nginx/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/nginx/init.sls
@@ -34,6 +34,13 @@
 
 {% endif %}
 
+ensure_underscores_in_headers_nginx_conf:
+  file.line:
+    - name: /etc/nginx/nginx.conf
+    - mode: ensure
+    - content:     underscores_in_headers on;
+    - after: ^http {$
+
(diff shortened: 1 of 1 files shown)


risdenk:
So I'm not 100% sure how to test this. I can spin up an env with a local CB build. Is there a better way to test salt script changes?

lacikaaa:
I would prefer this change to go into cloudbreak-images repo

keyki:
https://github.com/hortonworks/cloudbreak-images/blob/master/saltstack/base/salt/nginx/etc/nginx/nginx.conf

risdenk:
Ah thanks I assumed the nginx.conf was default OS installed. I'll open a PR against the base image. thanks!

risdenk:
Opened https://github.com/hortonworks/cloudbreak-images/pull/353

PR #6655, opened 2019-11-06
CB-4367. Add CCM tunnel for Knox.

jadair-cloudera:
With these changes, gateway nodes with Knox are configured to register
a CCM tunnel to the nginx fronting Knox on port 443.


keyki:
There were failing tests due to a recent cbd change which was fixed. Can you please rebase to master?

PR #6663, opened 2019-11-07
CB-4088 Ensure DELETE_FAILED status if termination finalization fails

horadla23:


freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
```diff
@@ -70,6 +70,7 @@ public void finalizeTermination(Long stackId) {
             });
         } catch (TransactionService.TransactionExecutionException ex) {
             LOGGER.info("Failed to terminate cluster infrastructure. Stack id {}", stack.getId());
+            stackUpdater.updateStackStatus(stackId, DetailedStackStatus.DELETE_FAILED, "Finalization of stack termination failed.");
```
lnardai: Maybe we could also take into account the forced flag.
I guess we want the forced delete functionality to finalize the deletion at any cost.
`if (forced) -> DELETE_COMPLETED`

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/TerminationService.java
```diff
@@ -99,27 +99,40 @@ public void finalizeTermination(Long stackId, boolean force) {
...
                 return null;
             });
         } catch (TransactionExecutionException ex) {
             LOGGER.info("Failed to terminate cluster infrastructure. Stack id {}", stack.getId());
+            setOnlyStackDeleted(stack, force, terminatedName);
             throw new TerminationFailedException(ex);
         }
     }
 
+    private void setOnlyStackDeleted(Stack stack, boolean force, String terminatedName) {
```
lnardai: probably would be better to call it something like `deleteOnlyIfForced`

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterTerminationService.java
```diff
@@ -192,8 +192,14 @@ private void deleteFileSystemResources(Long stackId, FileSystem fileSystem) {
                 fsConfiguration.setStorageContainer("cloudbreak" + stackId);
                 fsConfigurator.deleteResources(fsConfiguration);
             }
-        } catch (IOException e) {
-            throw new TerminationFailedException("File system resources could not be deleted: ", e);
+        } catch (IOException ioe) {
```
lnardai: Do we want to handle the IOException as a special case where force delete should be skipped?

PR #6664, opened 2019-11-07
CB-4370 - yarn and basic stack tests now create freeipa

attilapalfi92:


aszegedi:
Maybe we should consider to retire our old E2E test setup:
```
protected void setupTest(TestContext testContext) {
        createDefaultUser(testContext);
        createDefaultCredential(testContext);
        createDefaultEnvironment(testContext);
        initializeDefaultBlueprints(testContext);
    }
```
and apply the complete one:
```
protected void setupTest(TestContext testContext) {
        createDefaultUser(testContext);
        createDefaultCredential(testContext);
        createEnvironmentWithNetworkAndFreeIPA(testContext);
        initializeDefaultBlueprints(testContext);
    }
```

attilapalfi92:
@aszegedi there are still many cases when the default setup is used. This was my idea originally as well, but I checked and it's not worth it.

PR #6666, opened 2019-11-07
Remove Cumulus

doktoric:


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -428,6 +428,14 @@ public Cluster updateAmbariClientConfig(Long clusterId, HttpClientConfig ambariC
         return cluster;
     }
 
+    public Cluster updateExtendedBlueprint(Long clusterId, String blueprint) {
```
topolyai5: unused method

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/DistroXClusterCreationTest.java
```diff
@@ -287,7 +287,7 @@ private static DistroXTestDto distroxClusterTemplateContainsMockHostname(TestCon
                 MockVerification.verify(HttpMethod.POST,
                         ClouderaManagerMock.IMPORT_CLUSTERTEMPLATE)
                         .bodyContains(MOCK_HOSTNAME)
-                        .exactTimes(2),
+                        .atLeast(1),
```
topolyai5: what was the reason to change to at least once?

PR #6668, opened 2019-11-07
Cb 4329 Strict validation for OS upgrade for SDX

foldik:
Upgrade looks for new images with the same parcel and package versions

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
@@ -400,8 +400,7 @@ public class StackCreatorService {
             try {
                 boolean base = blueprintService.isAmbariBlueprint(blueprint) ? shouldUseBaseAmbariImage(clusterRequest) : shouldUseBaseCMImage(clusterRequest);
                 LOGGER.info("The stack with name {} will use base image: {}", stackName, base);
-                return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user,
-                        Optional.empty());
+                return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user, image -> true);
             } catch (CloudbreakImageNotFoundException | CloudbreakImageCatalogException e) {
                 throw new BadRequestException(e.getMessage(), e);
             }
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
@@ -169,16 +169,16 @@ public class UpgradeService {
                 .map(ClouderaManagerProduct::getParcel)
                 .collect(toSet());
         originalClusterComponentUrls.add(clusterComponentConfigProvider.getClouderaManagerRepoDetails(stack.getCluster().getId()).getBaseUrl());
-        return imageFromCatalaog -> {
-            if (imageFromCatalaog.getPreWarmParcels() == null) {
+        return imageFromCatalog -> {
+            if (imageFromCatalog.getPreWarmParcels() == null) {
                 return false;
             } else {
-                Set<String> newClusterComponentUrls = imageFromCatalaog.getPreWarmParcels()
+                Set<String> newClusterComponentUrls = imageFromCatalog.getPreWarmParcels()
                         .stream()
                         .map(parts -> parts.get(1))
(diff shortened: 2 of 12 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -400,8 +400,7 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
             try {
                 boolean base = blueprintService.isAmbariBlueprint(blueprint) ? shouldUseBaseAmbariImage(clusterRequest) : shouldUseBaseCMImage(clusterRequest);
                 LOGGER.info("The stack with name {} will use base image: {}", stackName, base);
-                return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user,
-                        Optional.empty());
+                return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user, image -> true);
```
horadla23: just a minor thing: `Predicates.alwaysTrue()` could be better for reading the code

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
```diff
@@ -129,6 +140,47 @@ private UpgradeOptionV4Response getUpgradeOption(Stack stack, Long workspaceId,
...
+                    .and(parcelFilter(stack));
+        }
+    }
+
+    private Predicate<com.sequenceiq.cloudbreak.cloud.model.catalog.Image> parcelFilter(Stack stack) {
+        Set<String> parcels = clusterComponentConfigProvider.getClouderaManagerProductDetails(stack.getCluster().getId())
+                .stream()
+                .map(ClouderaManagerProduct::getParcel)
+                .collect(toSet());
+        return imageFromCatalaog -> {
```
horadla23: please fix the typo

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
```diff
@@ -129,6 +140,47 @@ private UpgradeOptionV4Response getUpgradeOption(Stack stack, Long workspaceId,
...
+        };
+    }
+
+    private Predicate<com.sequenceiq.cloudbreak.cloud.model.catalog.Image> packageVersionFilter(Map<String, String> packageVersions) {
+        return image -> {
+            Map<String, String> catalogPackageVersions = new HashMap<>(image.getPackageVersions());
+            catalogPackageVersions.remove(SALT_BOOTSTRAP);
+            Map<String, String> originalPackageVersions = new HashMap<>(packageVersions);
+            originalPackageVersions.remove(SALT_BOOTSTRAP);
+            return originalPackageVersions.equals(catalogPackageVersions);
```
horadla23: afaik in case of upgrading to a baseimage, this will return false, because running stack's package versions contains cm and cdh version, but base images in image catalog contains only salt and salt-bootstrap in image catalog

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
```diff
@@ -129,6 +140,47 @@ private UpgradeOptionV4Response getUpgradeOption(Stack stack, Long workspaceId,
...
+    private Predicate<com.sequenceiq.cloudbreak.cloud.model.catalog.Image> parcelFilter(Stack stack) {
+        Set<String> parcels = clusterComponentConfigProvider.getClouderaManagerProductDetails(stack.getCluster().getId())
+                .stream()
+                .map(ClouderaManagerProduct::getParcel)
+                .collect(toSet());
+        return imageFromCatalaog -> {
+            if (imageFromCatalaog.getPreWarmParcels() == null) {
+                return false;
+            } else {
+                Set<String> imageParcels = imageFromCatalaog.getPreWarmParcels()
```
horadla23: you can use `CollectionUtils.emptyIfNull()` here and then you do not need the `if-else` statement above

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
foldik: Yes, originally I used it but IntelliJ suggested that it can be replaced with such a simple expression which might makes sense to not depend on external libraries, but since this is already the case I can put it back.

horadla23:
Repo URL of CM should be compared too as we discussed.

PR #6676, opened 2019-11-08
CB-4107 - Fix to always fill security group id in ARM template

pdarvasi:


environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EnvironmentNetworkProviderValidator.java
```diff
@@ -109,7 +109,7 @@ private boolean isInvalidNetworkMask(String networkCidr) {
     private void validateSecurityGroup(EnvironmentDto request, String cloudPlatform, ValidationResultBuilder resultBuilder) {
         SecurityAccessDto securityAccess = request.getSecurityAccess();
         NetworkDto networkDto = request.getNetwork();
-        if (securityAccess != null && networkDto != null && Strings.isNullOrEmpty(securityAccess.getCidr())) {
+        if (securityAccess != null && networkDto != null) {
```
doktoric: I think we should check here the network.networkCidr

pdarvasi: Honestly, I don't know why will speak with @doktoric if he is available

PR #6681, opened 2019-11-08
CDPCP-1028. Throw exception on cluster proxy error in FMS

aarman-cloudera:
The JSON RPC client we use for FMS returns a null when there
is an HTTP error and the error response body doesn't conform
to json rpc standards. This is not ideal behavior because the
cluster proxy may return an HTTP error that has valid json
body but not valid json rpc format.

To surface cluster proxy errors, we now throw an error if the
json in the RPC response matches what we expect the cluster
proxy errors to look like.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
@@ -1,0 +1,43 @@
+package com.sequenceiq.freeipa.client;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.googlecode.jsonrpc4j.JsonRpcClient;
+import com.sequenceiq.cloudbreak.clusterproxy.ClusterProxyException;
+
+import java.util.Optional;
+
+public class ClusterProxyErrorRpcListener implements JsonRpcClient.RequestListener {
+
+    private Optional<ClusterProxyError> deserializeAsClusteProxyError(ObjectNode objectNode) {
+        ObjectMapper mapper = new ObjectMapper();
+        ClusterProxyError clusterProxyError;
+        try {
+            clusterProxyError = mapper.treeToValue(objectNode, ClusterProxyError.class);
+            if (!clusterProxyError.getCode().contains("cluster-proxy")) {
+                return Optional.empty();
+            }
+        } catch (Exception ex) {
+            return Optional.empty();
+        }
+        return Optional.of(clusterProxyError);
+    }
+
+    private void throwIfClusterProxyError(ObjectNode node) {
+        Optional<ClusterProxyError> clusterProxyError = deserializeAsClusteProxyError(node);
+        if (clusterProxyError.isPresent()) {
+            throw new ClusterProxyException(String.format("Cluster proxy service returned error: %s", clusterProxyError.get()));
+        }
+    }
+
+    @Override
+    public void onBeforeRequestSent(JsonRpcClient client, ObjectNode request) {
+        // no op
+    }
+
+    @Override
+    public void onBeforeResponseProcessed(JsonRpcClient client, ObjectNode response) {
+        throwIfClusterProxyError(response);
+    }
+
+}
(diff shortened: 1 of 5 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
```diff
@@ -0,0 +1,71 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.annotations.VisibleForTesting;
+import com.googlecode.jsonrpc4j.JsonRpcClient;
+import com.sequenceiq.cloudbreak.clusterproxy.ClusterProxyException;
+
+import java.util.Optional;
+
+public class ClusterProxyErrorRpcListener implements JsonRpcClient.RequestListener {
+
+    private static Optional<ClusterProxyError> deserializeAsClusteProxyError(ObjectNode objectNode) {
```
lacikaaa: why is this method static?

freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
```diff
@@ -0,0 +1,71 @@
...
+                return Optional.empty();
+            }
+        } catch (Exception ex) {
+            return Optional.empty();
+        }
+        return Optional.of(clusterProxyError);
+    }
+
+    @VisibleForTesting
+    static void throwIfClusterProxyError(ObjectNode node) {
```
lacikaaa: why is this method static?
can we make it private?

freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
```diff
@@ -0,0 +1,71 @@
...
+            if (!clusterProxyError.code.contains("cluster-proxy")) {
+                return Optional.empty();
+            }
+        } catch (Exception ex) {
+            return Optional.empty();
+        }
+        return Optional.of(clusterProxyError);
+    }
+
+    @VisibleForTesting
```
lacikaaa: the logic should be tested using public methods if possible. I think in this case it's easy to test via `onBeforeResponseProcessed`

freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
```diff
@@ -0,0 +1,71 @@
...
+    public void onBeforeRequestSent(JsonRpcClient client, ObjectNode request) {
+        // no op
+    }
+
+    @Override
+    public void onBeforeResponseProcessed(JsonRpcClient client, ObjectNode response) {
+        throwIfClusterProxyError(response);
+    }
+
+    public static class ClusterProxyError {
```
lacikaaa: can we move this out?

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -226,4 +232,16 @@ private CloseableHttpResponse execute(HttpPost post, CookieStore cookieStore) th
     private static HostnameVerifier hostnameVerifier() {
         return (s, sslSession) -> true;
     }
+
+    private static JsonRpcClient.RequestListener noOpRequestListener() {
```
lacikaaa: we could avoid this noop implementation if we don't set anything if it's not necessary. Also the JsonRpcClient can handle null as RequestListener

freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
aarman-cloudera: To indicate that it doesn't change the internal state of object.  I'll remove the static since I don't have a strong preference.

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
aarman-cloudera: ah good point didn't realize null can be handled

freeipa/src/main/java/com/sequenceiq/freeipa/client/ClusterProxyErrorRpcListener.java
aarman-cloudera: true, I'll move it to the public method.

PR #6682, opened 2019-11-11
CB-2551, CB-4422 Env start stop

topolyai5:


datalake/src/main/java/com/sequenceiq/datalake/flow/stop/SdxStopActions.java
```diff
@@ -68,6 +69,34 @@ protected Object getFailurePayload(SdxStartStopEvent payload, Optional<SdxContex
         };
     }
 
+    @Bean(name = "SDX_STOP_ALL_DATAHUB_STATE")
+    public Action<?, ?> sdxStackStartAllDatahub() {
```
lajosrodek: Rename to `sdxStackStopAllDatahub`.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/DistroxService.java
```diff
@@ -31,4 +51,66 @@
             throw e;
         }
     }
+
+    public void stopAttachedDistrox(String envName) {
+        Collection<StackViewV4Response> attachedDistroXClusters = getAttachedDistroXClusters(envName, null);
+        List<String> pollingCrn = new ArrayList<>();
+        attachedDistroXClusters.forEach(d -> {
+            if (!d.getStatus().isStopState()) {
+                distroXV1Endpoint.putStopByCrn(d.getCrn());
```
attilapalfi92: If any of these rest calls fail for any reason, the whole stop operation will fail. There could be temporary network issues or anything.

There are 2 ways to mitigate this:
- better: introduce multi-stop and multi-start endpoints
- worse: handle various exceptions here

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
```diff
@@ -0,0 +1,82 @@
...
+            throw e;
+        }
+    }
+
+    public void stopAttachedDistrox(String envName) {
+        Collection<StackViewV4Response> attachedDistroXClusters = getAttachedDistroXClusters(envName, null);
+        ArrayList<String> pollingCrn = new ArrayList<>();
+        attachedDistroXClusters.forEach(d -> {
+            if (!d.getStatus().isStopState()) {
+                distroXV1Endpoint.putStopByCrn(d.getCrn());
```
attilapalfi92: If any of these rest calls fail for any reason, the whole stop operation will fail. There could be temporary network issues or anything.

There are 2 ways to mitigate this:
- better: introduce multi-stop and multi-start endpoints
- worse: handle various exceptions here

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
```diff
@@ -0,0 +1,82 @@
...
+                .waitPeriodly(sleeptime, TimeUnit.SECONDS)
+                .run(distroxPollerCollection.stopPoller(pollingCrn));
+    }
+
+    public void startAttachedDistrox(String envName) {
+        Collection<StackViewV4Response> attachedDistroXClusters = getAttachedDistroXClusters(envName, null);
+        ArrayList<String> pollingCrn = new ArrayList<>();
+        attachedDistroXClusters.forEach(d -> {
+            if (!d.getStatus().isStartState()) {
+                distroXV1Endpoint.putStartByName(d.getName());
```
attilapalfi92: If any of these rest calls fail for any reason, the whole stop operation will fail. There could be temporary network issues or anything.

There are 2 ways to mitigate this:
- better: introduce multi-stop and multi-start endpoints
- worse: handle various exceptions here

environment/src/main/java/com/sequenceiq/environment/environment/service/DatalakeService.java
```diff
@@ -0,0 +1,176 @@
...
+        attachedDatalake.forEach(d -> {
+            if (!isStopState(d.getStatus())) {
+                sdxEndpoint.stopByCrn(d.getCrn());
+            }
+            pollingCrn.add(d.getCrn());
+        });
+
+        List<String> remaining = new ArrayList<>();
+
+        Polling.stopAfterAttempt(attempt)
```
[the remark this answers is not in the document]
topolyai5: It's running on the current thread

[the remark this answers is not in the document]
attilapalfi92: thanks for the info

datalake/src/test/java/com/sequenceiq/datalake/service/sdx/start/SdxStartServiceTest.java
```diff
@@ -80,15 +91,46 @@ public void testTriggerStartWhenSdxStopped() {
...
+    public void testTriggerStartWhenFreeipaNull() {
         SdxCluster sdxCluster = sdxCluster();
         SdxStatusEntity sdxStatusEntity = new SdxStatusEntity();
         sdxStatusEntity.setStatus(DatalakeStatusEnum.RUNNING);
 
         when(sdxStatusService.getActualStatusForSdx(sdxCluster)).thenReturn(sdxStatusEntity);
 
+        underTest.triggerStartIfClusterNotRunning(sdxCluster);
+
+        verify(sdxReactorFlowManager, times(0)).triggerSdxStartFlow(CLUSTER_ID);
```
bergerdenes: times(0) == never()

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
```diff
@@ -0,0 +1,64 @@
...
+        private String selector;
+
+        private Long resourceId;
+
+        private Promise<Boolean> accepted;
+
+        private EnvStopEventBuilder() {
+        }
+
+        public static EnvStopEventBuilder anEnvStopEvent() {
```
bergerdenes: move this to the built class with name builder()

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
```diff
@@ -0,0 +1,64 @@
+package com.sequenceiq.environment.environment.flow.stop.event;
+
+import com.sequenceiq.flow.reactor.api.event.BaseNamedFlowEvent;
+
+import reactor.rx.Promise;
+
+public class EnvStopEvent extends BaseNamedFlowEvent {
+
+    public EnvStopEvent(String selector, Long resourceId, String resourceName, String resourceCrn) {
```
bergerdenes: create a constructor that takes a builder

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
```diff
@@ -0,0 +1,64 @@
...
+            return this;
+        }
+
+        public EnvStopEventBuilder withResourceCrn(String resourceCrn) {
+            this.resourceCrn = resourceCrn;
+            return this;
+        }
+
+        public EnvStopEvent build() {
+            return new EnvStopEvent(selector, resourceId, accepted, resourceName, resourceCrn);
```
bergerdenes: use new constructor that takes Builder

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopAllDatahubHandler.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Inject
+    private EventBus eventBus;
+
+    @Inject
+    private SdxStopService sdxStopService;
+
+    @Override
+    public String selector() {
+        return "SdxStopAllDatahubRequest";
```
bergerdenes: this is fishy
shouldn't it come from an Enum somewhere?

environment/src/main/java/com/sequenceiq/environment/environment/service/DatalakeService.java
```diff
@@ -0,0 +1,176 @@
...
+        try {
+            return sdxEndpoint.list(environmentName);
+        } catch (WebApplicationException | ProcessingException e) {
+            String message = String.format("Failed to get Datalake clusters from service due to: '%s' ", e.getMessage());
+            LOGGER.error(message, e);
+            throw e;
+        }
+    }
+
+    public void stopAttachedDatalake(String envName) {
```
bergerdenes: please refactor this method to smaller ones as this is too complex

environment/src/main/java/com/sequenceiq/environment/environment/service/DatalakeService.java
```diff
@@ -0,0 +1,176 @@
...
+                    if (count > 0) {
+                        return AttemptResults.justContinue();
+                    }
+                    pollingCrn.retainAll(remaining);
+                    return AttemptResults.finishWith(null);
+                });
+
+    }
+
+    public void startAttachedDatalake(String envName) {
```
bergerdenes: please refactor this method to smaller ones as this is too complex

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
bergerdenes: or do it in retry ;)

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStartService.java
```diff
@@ -0,0 +1,39 @@
...
+        EnvironmentDto environment = environmentService.getByCrnAndAccountId(crn, accountId);
+        reactorFlowManager.triggerStartFlow(environment.getId(), environment.getName(), userCrn);
+    }
+
+    public void startByName(String name) {
+        String accountId = threadBasedUserCrnProvider.getAccountId();
+        String userCrn = threadBasedUserCrnProvider.getUserCrn();
+        EnvironmentDto environment = environmentService.getByNameAndAccountId(name, accountId);
+        reactorFlowManager.triggerStartFlow(environment.getId(), environment.getName(), userCrn);
+    }
```
bergerdenes: there are some duplicated code lines, please simplify

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStatusUpdateService.java
```diff
@@ -0,0 +1,55 @@
...
+import com.sequenceiq.environment.api.v1.environment.model.response.SimpleEnvironmentResponse;
+import com.sequenceiq.environment.environment.EnvironmentStatus;
+import com.sequenceiq.environment.environment.domain.Environment;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.environment.v1.EnvironmentApiConverter;
+import com.sequenceiq.flow.core.CommonContext;
+import com.sequenceiq.notification.NotificationService;
+
+@Service
+public class EnvironmentStatusUpdateService {
```
bergerdenes: I miss the unit test for this class

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStopService.java
```diff
@@ -0,0 +1,43 @@
...
+        String userCrn = threadBasedUserCrnProvider.getUserCrn();
+        EnvironmentDto environment = environmentService.getByCrnAndAccountId(crn, accountId);
+        reactorFlowManager.triggerStopFlow(environment.getId(), environment.getName(), userCrn);
+    }
+
+    public void stopByName(String name) {
+        String accountId = threadBasedUserCrnProvider.getAccountId();
+        String userCrn = threadBasedUserCrnProvider.getUserCrn();
+        EnvironmentDto environment = environmentService.getByNameAndAccountId(name, accountId);
+        reactorFlowManager.triggerStopFlow(environment.getId(), environment.getName(), userCrn);
```
bergerdenes: simplify please, duplicated code

environment/src/main/java/com/sequenceiq/environment/environment/service/FreeipaService.java
```diff
@@ -0,0 +1,55 @@
...
+
+@Service
+public class FreeipaService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FreeipaService.class);
+
+    @Value("${env.stop.polling.attempt:90}")
+    private Integer attempt;
+
+    @Value("${env.stop.polling.sleep.time:5}")
```
bergerdenes: 5s is too frequent IMHO

environment/src/main/java/com/sequenceiq/environment/environment/service/DatalakeService.java
```diff
@@ -0,0 +1,176 @@
...
+
+@Service
+public class DatalakeService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DatalakeService.class);
+
+    @Value("${env.stop.polling.attempt:90}")
+    private Integer attempt;
+
+    @Value("${env.stop.polling.sleep.time:5}")
```
bergerdenes: 5s is too frequent IMHO

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxPollerCollection.java
```diff
@@ -0,0 +1,121 @@
...
+    public DistroxPollerCollection(DistroXV1Endpoint distroXV1Endpoint) {
+        this.distroXV1Endpoint = distroXV1Endpoint;
+    }
+
+    public AttemptMaker<Void> stopPoller(ArrayList<String> pollingCrn) {
+        return () -> {
+            List<String> remaining = new ArrayList<>();
+            List<AttemptResult<Void>> results = pollingCrn.stream()
+                    .map(crn -> {
+                        StackV4Response stack = distroXV1Endpoint.getByCrn(crn, Collections.emptySet());
```
bergerdenes: move this to a separate method pls

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxPollerCollection.java
```diff
@@ -0,0 +1,121 @@
...
+            return evaluateResult(pollingCrn, remaining, results);
+        };
+    }
+
+    public AttemptMaker<Void> startPoller(ArrayList<String> pollingCrn) {
+        return () -> {
+            List<String> remaining = new ArrayList<>();
+            List<AttemptResult<Void>> results = pollingCrn.stream()
+                    .map(crn -> {
+                        StackV4Response stack = distroXV1Endpoint.getByCrn(crn, Collections.emptySet());
```
bergerdenes: move this to a separate method pl

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
```diff
@@ -0,0 +1,82 @@
...
+
+@Service
+public class DistroxService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DistroxService.class);
+
+    @Value("${env.stop.polling.attempt:90}")
+    private Integer attempt;
+
+    @Value("${env.stop.polling.sleep.time:5}")
```
bergerdenes: 5s sleep time is too frequent IMHO

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopAllDatahubHandler.java
topolyai5: Unfortunately, the selectors are the weak point of the flow2. If I use another enum It will make more unreadable the flow config and events

environment/src/main/java/com/sequenceiq/environment/environment/service/FreeipaService.java
topolyai5: FreeIpa stop finishes quickly generally. I think it is reasonable.

environment/src/main/java/com/sequenceiq/environment/environment/service/DatalakeService.java
topolyai5: FreeIpa stop finishes quickly generally. I think it is reasonable.

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
topolyai5: DistroX stop finishes quickly generally. I think it is reasonable.

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
topolyai5: this is the convention in this service

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
topolyai5: this is the convention in this service

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/event/EnvStopEvent.java
topolyai5: this is the convention in this service

datalake/src/test/java/com/sequenceiq/datalake/service/sdx/start/SdxStartServiceTest.java
topolyai5: It's true, but I prefer times(0) form

attilapalfi92: I agree with @bergerdenes here

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStartService.java
topolyai5: If I try to simplify this method, I will make it unreadable only. The first 2 lines the same only.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStatusUpdateService.java
topolyai5: If I try to simplify this method, I will make it unreadable only. The first 2 lines the same only.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStartService.java
topolyai5: If I try to simplify this method, I will make it unreadable only. The first 2 lines the same only.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStopService.java
topolyai5: If I try to simplify this method, I will make it unreadable only. The first 2 lines the same only.

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopAllDatahubHandler.java
attilapalfi92: In the Environment service for example for the env creation flow we use an enum for this, and it does not make it more unreadable at all. On the contrary, it is more robust. It is used in both the handlers and in the actions class.
Can you please adopt this pattern?
See `EnvCreationHandlerSelectors` and `EnvDeleteHandlerSelectors`

You also did the same in this pull request with `EnvStopHandlerSelectors`.

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testStopAttachedDistroxWhenNoAttachedDistroX() {
+        when(distroXV1Endpoint.list("envName", null)).thenReturn(new StackViewV4Responses(Collections.emptySet()));
+        when(pollerCollection.stopDistroXPoller(new ArrayList<>(), 1L)).thenReturn(() -> AttemptResults.finishWith(null));
+
+        underTest.stopAttachedDistrox(1L, "envName");
+
+        verify(distroXV1Endpoint, times(0)).putStopByCrns(List.of("crn"));
```
bergerdenes: for correct checking it should be `verify(distroXV1Endpoint, times(0)).putStopByCrns(anyListOf(String.class)));` or even better, any()

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testStartAttachedDistroxWhenNoAttachedDistroX() {
+        when(distroXV1Endpoint.list("envName", null)).thenReturn(new StackViewV4Responses(Collections.emptySet()));
+        when(pollerCollection.startDistroXPoller(new ArrayList<>(), 1L)).thenReturn(() -> AttemptResults.finishWith(null));
+
+        underTest.startAttachedDistrox(1L, "envName");
+
+        verify(distroXV1Endpoint, times(0)).putStartByCrns(List.of("crn"));
```
bergerdenes: similar `verify(distroXV1Endpoint, times(0)).putStopByCrns(any());`

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import javax.ws.rs.ProcessingException;
+import javax.ws.rs.WebApplicationException;
+
+import org.jetbrains.annotations.NotNull;
```
bergerdenes: remove this please

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
```diff
@@ -0,0 +1,77 @@
...
+        Polling.stopAfterAttempt(attempt)
+                .stopIfException(true)
+                .waitPeriodly(sleeptime, TimeUnit.SECONDS)
+                .run(pollerCollection.startDistroXPoller(pollingCrn, envId));
+
+    }
+
+    @NotNull
+    private ArrayList<String> mapStackToCrn(Collection<StackViewV4Response> attachedDistroXClusters) {
+        return attachedDistroXClusters.stream().map(StackViewV4Response::getCrn).collect(Collectors.toCollection(ArrayList::new));
```
bergerdenes: Collectors.toList();

environment/src/main/java/com/sequenceiq/environment/environment/service/PollerCollection.java
```diff
@@ -0,0 +1,218 @@
+package com.sequenceiq.environment.environment.service;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import org.jetbrains.annotations.NotNull;
```
bergerdenes: remove this please

environment/src/test/java/com/sequenceiq/environment/environment/service/FreeipaPollerCollectionTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.AVAILABLE;
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.START_FAILED;
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.STOPPED;
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.STOP_FAILED;
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.STOP_IN_PROGRESS;
+import static com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status.UPDATE_IN_PROGRESS;
+
+import java.util.stream.Stream;
+
+import org.jetbrains.annotations.NotNull;
```
bergerdenes: remove this please

environment/src/test/java/com/sequenceiq/environment/environment/service/PollerCollectionTest.java
```diff
@@ -0,0 +1,202 @@
...
+import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.STOP_FAILED;
+import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.STOP_IN_PROGRESS;
+import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.UPDATE_IN_PROGRESS;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.jetbrains.annotations.NotNull;
```
bergerdenes: remove please

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
topolyai5: what is the problem with the exact string? I want to run with this string only.

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
topolyai5: what is the problem with the exact string? I want to run with this string only.

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
topolyai5: why did you add this to project than if it is not usable?

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
topolyai5: I need the exact ArrayList implementation, but the `Collectors.toList();` has another return type

environment/src/main/java/com/sequenceiq/environment/environment/service/PollerCollection.java
```diff
@@ -0,0 +1,218 @@
...
+        if (sdxStarted(sdx)) {
+            return AttemptResults.finishWith(null);
+        } else {
+            remaining.add(crn);
+            return checkStartStatus(sdx);
+        }
+    }
+
+    @NotNull
+    private AttemptResult<Void> evaluateResult(ArrayList<String> pollingCrn, List<String> remaining, List<AttemptResult<Void>> results) {
```
attilapalfi92: Why not using the Set interface instead of ArrayList? A set has the uniqueness attribute you need, and you don't need ordering what a list provides. I think the logic here should be rewritten to use the proper collection type (set).

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
attilapalfi92: You should use set instead of this list implementation. You need uniqueness, but don't need ordering. See my other comment.

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
attilapalfi92: exact evaluations are always better, if the author has the energy to write the test like that

environment/src/test/java/com/sequenceiq/environment/environment/service/DistroxServiceTest.java
attilapalfi92: exact evaluations are always better, if the author has the energy to write the test like that

environment/src/main/java/com/sequenceiq/environment/environment/service/PollerCollection.java
attilapalfi92: javax.validation.constraints.NotNull should be used

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
attilapalfi92: javax.validation.constraints.NotNull should be used

topolyai5: ok, but Denes added jetbrains dependency to project and Idea added this import automatically. So it does not disturb me, It is ok for me

environment/src/main/java/com/sequenceiq/environment/environment/service/DistroxService.java
topolyai5: The main problem is modifiable or unmodifiable implementations. I remove values from this list. The list or set declaration do not solve my problem. The most of factory methods are providing an unmodifiable list (like Collectors.toList(), Arrays.asList(), List.of()) and it will break my code in runtime. This is true for the Set as well. So I would use `ArraytList`.

attilapalfi92: You are propagating ArrayList dependency unnecessarily. You should use the Set interface, and for modifications, wrap them into modifiable Sets at the last step.

topolyai5: I don't want to use an unnecessary wrap, it is fully internal in this class

environment/src/main/java/com/sequenceiq/environment/environment/service/PollerCollection.java
topolyai5: it is not an internal method but I don't want to use an unnecessary wrap here

attilapalfi92: I cannot approve the PR. Code quality is more important than avoiding a wrapping.

environment/src/main/java/com/sequenceiq/environment/environment/service/PollerCollection.java
```diff
@@ -0,0 +1,216 @@
...
+        return () -> {
+            if (PollGroup.CANCELLED.equals(EnvironmentInMemoryStateStore.get(envId))) {
+                LOGGER.info("Datalake polling cancelled in inmemory store, id: " + envId);
+                return AttemptResults.breakFor("Datalake polling cancelled in inmemory store, id: " + envId);
+            }
+            List<String> remaining = new ArrayList<>();
+            List<AttemptResult<Void>> results = pollingCrn.stream()
+                    .map(crn -> fetchStartDatalakeResult(crn, remaining))
+                    .collect(Collectors.toList());
+            return evaluateResult(new ArrayList<>(pollingCrn), remaining, results);
```
attilapalfi92: Wrapping is repeated many times, and it's delegated to the callers of `evaluateResult`. It's just an internal behaviour but it would still be better if `evaluateResult` would be responsible for wrapping the things that needs to be wrapped for itself. Then you don't have to wrap them 4 times. It's error prone an requires attention unnecessarily.

topolyai5: I have to revert all of ArrayList change because does not work

PR #6684, opened 2019-11-11
CB-4277 Add Kudu web UI to Knox topology

doktoric:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
```diff
@@ -47,7 +47,9 @@
     PROFILER_ADMIN("Profiler Admin", "PROFILER_ADMIN_AGENT", "PROFILER-ADMIN-API", "/profiler-admin/", false, 21700, 21700, true, true),
     PROFILER_METRICS("Profiler Metrics", "PROFILER_METRICS_AGENT", "PROFILER-METRICS-API", "/profiler-metrics/", false, 21800, 21800, true, true),
     PROFILER_SCHEDULER("Profiler Scheduler", "PROFILER_SCHEDULER_AGENT", "PROFILER-SCHEDULER-API", "/profiler-scheduler/", false, 21900, 21900, true, true),
-    DAS("Data Analytics Studio", "DAS_WEBAPP", "DAS", "/das/", true, 30800, 30800, false, true);
+    DAS("Data Analytics Studio", "DAS_WEBAPP", "DAS", "/das/", true, 30800, 30800, false, true),
+    KUDU("Kudu Master", "KUDU_MASTER", "KUDUUI", "/kuduui/", true, 8051, 8051, false, true),
+    KUDU_TABLET_SERVER("Kudu Tablet Server", "KUDU_TSERVER", "KUDUUIT", "/kuduui/", true, 8050, 8050, false, true);
```
adembo: Is `KUDUUIT` a typo? Shouldn't it also be `KUDUUI`?

Reading more in the review, I see that at some level you're distinguishing between the master and tserver web UIs. Why? They behave identically apart from the difference in port number.

core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
```diff
@@ -312,6 +318,28 @@ private void addImplaDebugUrl(String managerIp, Gateway gateway, Map<String, Lis
...
+            urls.add(coordinatorUrl.get());
+        }
+    }
+
+    private void addKuduTUrl(String managerIp, Gateway gateway, Map<String, List<String>> privateIps, boolean autoTlsEnabled, List<String> urls) {
+        Optional<String> coordinatorUrl = privateIps.get(ExposedService.KUDU_TABLET_SERVER.getServiceName())
+                .stream()
+                .map(coordinator -> getKuduTUrl(gateway, managerIp, coordinator, autoTlsEnabled))
+                .flatMap(Optional::stream)
+                .findFirst();
```
adembo: This suggests the ability to publish a link to the tserver web UIs, and an arbitrary selection of the first tserver. I would recommend against that. Let's only publish links to the "first" master's web UI; all tserver web UIs are reachable from any master web UI.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
doktoric: @adembo I did this because Kudu Tablet Server should be available when KUDU_TSERVER is in the blueprint and Kudu Master should be visible when KUDU_MASTER is presented

core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
doktoric: So we dont want to publish the tserver url at all ?

adembo: I don’t think so. A typical Kudu cluster may have 3 or 5 masters and dozens (or more) tservers. Individual tserver web UIs aren’t that interesting with a cluster of that size, but the master is always compelling.

Moreover, there are links to the tserver web UIs (and to the other masters too) from inside each master’s web UI.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/ExposedService.java
adembo: I see. Perhaps you could change the strings then so `KUDUUIT` doesn't look like a typo? Maybe something like `KUDUUI_MASTER` and `KUDUUI_TSERVER`?

doktoric: So  I would like to separate these two component KUDUUIT is the **t**ablet server

core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
doktoric: removed that so the tablet server is not in the list

PR #6700, opened 2019-11-13
CB-4438 Realistic mock IDBroker mapping service user list

lajosrodek:
Contents:
* Extract Data Access and Ranger Audit service user sets into the new utility class `AccountMappingSubject`
  * Located in module `cloud-common` so that it becomes available to `AwsIDBrokerObjectStorageValidator` of #6639
  * Matches the corresponding collections of IDBMMS
* Replace field injection with constructor injection in `AwsMockAccountMappingService`
* Eliminate superfluous injected dependencies in `AzureMockAccountMappingService`
* Construct mappings in `AwsMockAccountMappingService` and `AzureMockAccountMappingService` using the user set of `AccountMappingSubject`
* Add JUnit 5 support to module `cloud-common`, while keeping JUnit 4 dependencies as well for legacy tests
* Add JUnit 5 unit tests for all classes involved

### DIFF
--- a/cloud-common/build.gradle
+++ b/cloud-common/build.gradle
@@ -66,10 +66,21 @@ dependencies {
     compile group: 'net.sf.json-lib',                       name: 'json-lib',                       version: '2.4',  classifier: 'jdk15'
 
     testCompile group: 'junit',                             name: 'junit',                          version: junitVersion
+    testCompile group: 'org.junit.jupiter',                 name: 'junit-jupiter-api',              version: junitJupiterVersion
+    testCompile group: 'org.junit.jupiter',                 name: 'junit-jupiter-params',           version: junitJupiterVersion
     testCompile group: 'org.mockito',                       name: 'mockito-core',                   version: mockitoVersion
+    testImplementation group: 'org.mockito',                name: 'mockito-junit-jupiter',          version: mockitoVersion
     testCompile group: 'org.springframework.boot',          name: 'spring-boot-starter-test',       version: springBootVersion
     testCompile group: 'org.awaitility',                    name: 'awaitility',                     version: '3.1.6'
 
+    testRuntime group: 'org.junit.jupiter',                 name: 'junit-jupiter-engine',           version: junitJupiterVersion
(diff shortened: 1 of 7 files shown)


cloud-common/build.gradle
```diff
@@ -66,10 +66,29 @@ dependencies {
...
+    testCompile group: 'org.junit.jupiter',                 name: 'junit-jupiter-params',           version: junitJupiterVersion
     testCompile group: 'org.mockito',                       name: 'mockito-core',                   version: mockitoVersion
+    testImplementation group: 'org.mockito',                name: 'mockito-junit-jupiter'
     testCompile group: 'org.springframework.boot',          name: 'spring-boot-starter-test',       version: springBootVersion
     testCompile group: 'org.awaitility',                    name: 'awaitility',                     version: '3.1.6'
 
+    testRuntime group: 'org.junit.jupiter',                 name: 'junit-jupiter-engine'
+    testRuntime group: 'org.junit.vintage',                 name: 'junit-vintage-engine'
+
+    constraints {
```
bergerdenes: why?
I think version selectors are in a good place in the `dependency` closure, the `constraints` closure is just redundant

cloud-common/src/test/java/com/sequenceiq/cloudbreak/service/identitymapping/AccountMappingSubjectTest.java
```diff
@@ -0,0 +1,22 @@
+package com.sequenceiq.cloudbreak.service.identitymapping;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class AccountMappingSubjectTest {
+
+    @Test
+    void testUserSetsDisjointness() {
```
bergerdenes: testDisjunctiveUserSets

cloud-common/build.gradle
lajosrodek: Not sure, copied that part from elsewhere. Will remove it then.

PR #6701, opened 2019-11-14
CB-4467 | Removing /clouderamanager from the internal url that gets r…

dvguruprasad:


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/clusterproxy/ClusterProxyService.java
```diff
@@ -159,7 +159,7 @@ private String clusterManagerUrl(Stack stack) {
 
     private String clusterManagerUrl(Stack stack, int port) {
         String gatewayIp = stack.getPrimaryGatewayInstance().getPublicIpWrapper();
-        return String.format("https://%s:%d/clouderamanager", gatewayIp, port);
```
lacikaaa: I think the `/` should be removed also

zabhi: Removed.

PR #6702, opened 2019-11-14
CB-4088 decrease pre-terminate recipe retry count

horadla23:


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
```diff
@@ -136,22 +136,22 @@ public void postClusterInstall(Stack stack) throws CloudbreakException {
...
-    public void preTerminationRecipesOnNodes(Stack stack, Set<Node> nodes) throws CloudbreakException {
+    public void preTerminationRecipesOnNodes(Stack stack, Set<Node> nodes, boolean forced) throws CloudbreakException {
         if (stack.getCluster() == null) {
             throw new NotFoundException("Cluster does not found, pre-termination will not be run.");
         }
         HostOrchestrator hostOrchestrator = hostOrchestratorResolver.get(stack.getOrchestrator().getType());
         GatewayConfig gatewayConfig = gatewayConfigService.getPrimaryGatewayConfig(stack);
         try {
-            hostOrchestrator.preTerminationRecipes(gatewayConfig, nodes, ClusterDeletionBasedExitCriteriaModel.nonCancellableModel());
+            hostOrchestrator.preTerminationRecipes(gatewayConfig, nodes, ClusterDeletionBasedExitCriteriaModel.nonCancellableModel(), forced);
```
lnardai: Maybe here we could extend the DeletionBasedExitCriteria with the forced flag, so it would be available in the SaltOrchestrator for the retry count configuration.

horadla23: It turned out that ClusterDeletionBasedExitCriteria is in the core module and we cannot access it from SaltOrchestrator where we need to use the forced flag, so it would require bigger refactor, anyway forced flag is not used to define the exit criteria, so I think it could confuse someone who reads the code.

PR #6705, opened 2019-11-14
CB-4455. Support shipping Freeipa logs to cloud storage.

oleewere:
Support fluentd log shipping for freeipa:
- as different logs needs to be monitored, different input.conf.j2 is used for freeipa and cm salt definitions (those are merged on salt-common scripts, see: https://github.com/hortonworks/cloudbreak/pull/6582)
- altus iam service is moved to auth-connector, and that is used directly from core & freeipa
- fill identity in cloud stack object for freeipa (that should work with both s3 and abdfs, tested only s3 for now)
- make freeipa telemetry support disabled by default (maybe it can be changed later)
- machine user creation is happening during salt config creation (we can access credentials once, we do not want to pass that really anywhere), the machine user deletion happening in a flow step (not too strict so i did not add a failed event there)


freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/cleanup/handler/MachineUserRemoveHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import com.sequenceiq.flow.event.EventSelectorUtil;
+import com.sequenceiq.flow.reactor.api.handler.EventHandler;
+import com.sequenceiq.freeipa.flow.freeipa.cleanup.event.ums.RemoveMachineUserRequest;
+import com.sequenceiq.freeipa.flow.freeipa.cleanup.event.ums.RemoveMachineUserResponse;
+import com.sequenceiq.freeipa.service.freeipa.cleanup.CleanupService;
+
+import reactor.bus.Event;
+import reactor.bus.EventBus;
+
+public class MachineUserRemoveHandler implements EventHandler<RemoveMachineUserRequest> {
```
attilapalfi92: Shouldn't this be a @Component?

oleewere:
update: i will movethe machine user deletion stuff to stack termination flow

freeipa/src/main/java/com/sequenceiq/freeipa/converter/cloud/StackToCloudStackConverter.java
```diff
@@ -239,6 +247,28 @@ private Security buildSecurity(InstanceGroup ig) {
...
+        if (telemetry != null && telemetry.getLogging() != null) {
+            Logging logging = telemetry.getLogging();
+            if (logging.getStorageLocation() != null) {
+                if (logging.getS3() != null) {
+                    CloudS3View s3View = new CloudS3View(CloudIdentityType.LOG);
+                    s3View.setInstanceProfile(logging.getS3().getInstanceProfile());
+                    return Optional.of(s3View);
+                } else if (logging.getAdlsGen2() != null) {
+                    CloudAdlsGen2View adlsGen2View = new CloudAdlsGen2View(CloudIdentityType.LOG);
+                    adlsGen2View.setAccountKey(logging.getAdlsGen2().getAccountKey());
```
lajosrodek: Minor: extract `logging.getAdlsGen2()` as a local variable and replace all below occurrences.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+        this.reportDeplymentLogs = reportDeplymentLogs;
+        this.databusEndpoint = databusEndpoint;
+    }
+
+    public Telemetry convert(TelemetryRequest request) {
+        Telemetry telemetry = null;
+        if (freeIpaTelemetryEnabled && request != null) {
+            telemetry = new Telemetry();
+            LoggingRequest loggingRequest = request.getLogging();
+            Logging logging = createLoggingFromRequest(loggingRequest);
```
lajosrodek: Variable not necessary, can be inlined in setter invocation.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+            telemetry.setLogging(logging);
+            Features features = null;
+            features = createFeaturesFromRequest(request, features);
+            telemetry.setFeatures(features);
```
lajosrodek: Variable not necessary, can be inlined in setter invocation. Does it make sense to pass in `null` to `createFeaturesFromRequest()`? Why not initialize the default return value inside that method?

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return telemetry;
+    }
+
+    public TelemetryResponse convert(Telemetry telemetry) {
+        TelemetryResponse response = null;
+        if (freeIpaTelemetryEnabled && telemetry != null) {
+            response = new TelemetryResponse();
+            response.setFluentAttributes(telemetry.getFluentAttributes());
+            Logging logging = telemetry.getLogging();
```
lajosrodek: Variable not necessary, can be inlined in the below invocation.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+            LoggingResponse loggingResponse = createLoggingResponseFromSource(logging);
```
lajosrodek: This one can be also inlined.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+            response.setLogging(loggingResponse);
+            Features features = telemetry.getFeatures();
+            FeaturesResponse featuresResponse = createFeaturesResponseFromSource(features);
+            response.setFeatures(featuresResponse);
```
lajosrodek: Similarly, I suggest refactoring this into a single setter invocation with variables / expressions inlined.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+        if (loggingRequest != null) {
+            logging = new Logging();
+            logging.setStorageLocation(loggingRequest.getStorageLocation());
+            if (loggingRequest.getS3() != null) {
+                S3CloudStorageV1Parameters s3Params = new S3CloudStorageV1Parameters();
+                s3Params.setInstanceProfile(loggingRequest.getS3().getInstanceProfile());
+                logging.setS3(s3Params);
+            } else if (loggingRequest.getAdlsGen2() != null) {
+                AdlsGen2CloudStorageV1Parameters adlsGen2Params = new AdlsGen2CloudStorageV1Parameters();
+                adlsGen2Params.setAccountKey(loggingRequest.getAdlsGen2().getAccountKey());
```
lajosrodek: Extract `loggingRequest.getAdlsGen2()` as local variable and replace all below occurrences.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+        if (logging != null) {
+            loggingResponse = new LoggingResponse();
+            loggingResponse.setStorageLocation(logging.getStorageLocation());
+            if (logging.getS3() != null) {
+                S3CloudStorageV1Parameters s3Params = new S3CloudStorageV1Parameters();
+                s3Params.setInstanceProfile(logging.getS3().getInstanceProfile());
+                loggingResponse.setS3(s3Params);
+            } else if (logging.getAdlsGen2() != null) {
+                AdlsGen2CloudStorageV1Parameters adlsGen2Params = new AdlsGen2CloudStorageV1Parameters();
+                adlsGen2Params.setAccountKey(logging.getAdlsGen2().getAccountKey());
```
lajosrodek: Extract `logging.getAdlsGen2()` as local variable and replace all below occurrences.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -0,0 +1,131 @@
...
+                adlsGen2Params.setAccountName(logging.getAdlsGen2().getAccountName());
+                adlsGen2Params.setSecure(logging.getAdlsGen2().isSecure());
+                adlsGen2Params.setManagedIdentity(logging.getAdlsGen2().getManagedIdentity());
+                loggingResponse.setAdlsGen2(adlsGen2Params);
+            }
+        }
+        return loggingResponse;
+    }
+
+    private Features createFeaturesFromRequest(TelemetryRequest request, Features features) {
```
lajosrodek: As mentioned above, this method seems to be always called with `features == null`. Why not remove the second parameter, and rather initialize the return value as the first statement before the `if` block?

freeipa/src/test/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverterTest.java
```diff
@@ -0,0 +1,66 @@
+package com.sequenceiq.freeipa.converter.telemetry;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+
+import org.junit.Before;
+import org.junit.Test;
```
lajosrodek: Could you please use JUnit 5 and assertJ (`Assertions.assertThat()`) if possible?

oleewere:
manual tests:
+ delete was done successfully
+ s3 output looks like this:

freeipa-logs

freeipa/src/main/java/com/sequenceiq/freeipa/service/AltusMachineUserService.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public AltusMachineUserService(AltusIAMService altusIAMService, CrnService crnService) {
+        this.altusIAMService = altusIAMService;
+        this.crnService = crnService;
+    }
+
+    public Optional<AltusCredential> createMachineUserWithAccessKeys(Stack stack) {
+        String userCrn = crnService.getUserCrn();
+        String machineUserName = getFluentMachineUser(stack);
+        return altusIAMService.generateMachineUserWithAccessKey(machineUserName, userCrn);
```
lacikaaa: How this `generateMachineUserWithAccessKey` will behave if we restart the service during provision and it is called again?

oleewere: it will generate a new access key (let's say the machine user already exists because of the re-run)..reason is for that we can only get the private key secret during creation. but on termination it deletes all the access keys

PR #6706, opened 2019-11-14
CB-4041 extend cluster's security rules to nginx with control plane cidrs

schfeca75:
@keyki pls review

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
@@ -46,6 +46,8 @@ import com.sequenceiq.cloudbreak.converter.AbstractConversionServiceAwareConvert
 import com.sequenceiq.cloudbreak.converter.v4.environment.network.EnvironmentNetworkConverter;
 import com.sequenceiq.cloudbreak.domain.Network;
 import com.sequenceiq.cloudbreak.domain.Orchestrator;
+import com.sequenceiq.cloudbreak.domain.SecurityGroup;
+import com.sequenceiq.cloudbreak.domain.SecurityRule;
 import com.sequenceiq.cloudbreak.domain.StackAuthentication;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.domain.stack.StackStatus;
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -152,9 +163,35 @@ public Stack convert(StackV4Request source) {
...
+    private void extendGatewaySecurityGroupWithDefaultGatewayCidrs(Stack stack) {
+        Set<InstanceGroup> gateways =
+                stack.getInstanceGroups().stream().filter(ig -> InstanceGroupType.isGateway(ig.getInstanceGroupType())).collect(Collectors.toSet());
+        Set<String> defaultGatewayCidrs = defaultGatewayCidr
+                .stream()
+                .filter(StringUtils::isNotBlank)
+                .collect(Collectors.toSet());
+        for (InstanceGroup gateway : gateways) {
+            Set<SecurityRule> rules = gateway.getSecurityGroup().getSecurityRules();
+            if (!defaultGatewayCidrs.isEmpty()) {
```
lajosrodek: I suggest pulling the `if` outside the `for` loop as the condition is constant throughout all iterations.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -107,6 +112,12 @@
     @Value("${cb.platform.default.regions:}")
     private String defaultRegions;
 
+    @Value("#{'${cb.default.gateway.cidr:}'.split(',')}")
+    private Set<String> defaultGatewayCidr;
+
+    @Value("${cb.nginx.port:9443}")
```
keyki: We shouldn't have default value here as it's part of the app.yml

cloud-common/src/main/resources/application.yml
```diff
@@ -83,6 +83,7 @@ cb:
     AZURE: 50
     GCP: 50
   enabled.linux.types: redhat6,redhat7,centos6,centos7
+  default.gateway.cidr:
```
keyki: This should have a value for 0.0.0.0/0

freeipa/src/main/resources/application.yml
```diff
@@ -24,6 +24,7 @@ management:
 freeipa:
   cert.dir: /certs/
   client.id: freeipa
+  default.gateway.cidr:
```
keyki: This should have a value for 0.0.0.0/0

keyki:
In FreeIPA application.yml there is a part
```
  security.default:
    ports: 22,443,9443
    cidr: 0.0.0.0/0
```
I think this should be removed.

schfeca75:
> In FreeIPA application.yml there is a part
> 
> ```
>   security.default:
>     ports: 22,443,9443
>     cidr: 0.0.0.0/0
> ```
> 
> I think this should be removed.

Done

schfeca75:
> No loglines were added

Logs are added

PR #6722, opened 2019-11-15
CB-4312: set SSD for default disc type on azure

tiborpopovics:


keyki:
SSD instead of SSH, but it's not a big deal

PR #6726, opened 2019-11-15
CB-4466. Configureble Datalake RDB instance types (in static templates)

oleewere:
that is only a draft
- replace properties (dynamic way - so it can be named anything) in static definitions (use the actual default one, maybe would make more sense to use cheaper instances by default...that would be cheaper fore development) -> these can be set as env vars in helm charts -> use different values for different envs
- I have provided only one example, probably we should fill all the templates + volume size can be configurable as well


oleewere:
im closing it as probably it would be better with providing something to the UI

PR #6728, opened 2019-11-15
CB-1358 - Rename every Ambari related DB column to generic one

gregito:
CB-1358 - Rename every Ambari related DB column to generic one

### DIFF
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
@@ -14,6 +14,8 @@ import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.STOP_FAILE
 import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.STOP_IN_PROGRESS;
 import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.STOP_REQUESTED;
 import static com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status.UPDATE_IN_PROGRESS;
+import static com.sequenceiq.cloudbreak.util.NullUtil.getIfNotNull;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
 import java.util.HashSet;
 import java.util.Objects;
@@ -114,7 +116,15 @@ public class Cluster implements ProvisionEntity, WorkspaceAwareResource {
 
     @Convert(converter = SecretToString.class)
     @SecretValue
+    private Secret cloudbreakClusterManagerUser = Secret.EMPTY;
+
+    @Convert(converter = SecretToString.class)
+    @SecretValue
     private Secret cloudbreakAmbariPassword = Secret.EMPTY;
+
+    @Convert(converter = SecretToString.class)
+    @SecretValue
+    private Secret cloudbreakClusterManagerPassword = Secret.EMPTY;
 
     @Convert(converter = SecretToString.class)
     @SecretValue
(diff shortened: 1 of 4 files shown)


core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/DatalakeResources.java
```diff
@@ -96,27 +104,54 @@ public void setDatalakeStackId(Long datalakeStackId) {
...
+        setDatalakeClusterManagerIp(datalakeAmbariIp);
+    }
+
+    public void setDatalakeClusterManagerIp(String datalakeClusterManagerIp) {
+        this.datalakeClusterManagerIp = datalakeClusterManagerIp;
     }
 
     public String getDatalakeAmbariFqdn() {
-        return datalakeAmbariFqdn;
+        return isNotEmpty(getDatalakeClusterManagerFqdn()) ? getDatalakeClusterManagerFqdn() : datalakeAmbariFqdn;
```
akanto: Do we need DataLake...Fqdn? I have teh feeling that this might be removed completely

akanto:
I am afraid we need to set the clusterManager* values everywhere in the code where we set ambari*

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/DatalakeResources.java
gregito: Probably you're right, but I think removing/changing anything which is not - closely - related to the/this renaming task has no place here.
Would you like to have a ticket for it? If so, I'll create one and suggest that to put it into our future sprint.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
```diff
@@ -425,51 +446,91 @@ public void setConfigStrategy(ConfigStrategy configStrategy) {
...
+        return cloudbreakClusterManagerUser.getSecret();
     }
 
     public void setCloudbreakAmbariUser(String cloudbreakAmbariUser) {
         this.cloudbreakAmbariUser = new Secret(cloudbreakAmbariUser);
+        this.cloudbreakClusterManagerUser = new Secret(cloudbreakAmbariUser);
+    }
+
+    public void setCloudbreakClusterManagerUser(String cloudbreakAmbariUser) {
+        this.cloudbreakClusterManagerUser = new Secret(cloudbreakAmbariUser);
```
bergerdenes: should write the old fields as well. same for the other setters.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
```diff
@@ -425,51 +446,91 @@ public void setConfigStrategy(ConfigStrategy configStrategy) {
...
+        return isNotEmpty(getCloudbreakClusterManagerUser()) ? getCloudbreakClusterManagerUser() : cloudbreakAmbariUser.getRaw();
+    }
+
+    public String getCloudbreakClusterManagerUser() {
+        return cloudbreakClusterManagerUser.getRaw();
     }
 
     public String getCloudbreakAmbariUserSecret() {
-        return cloudbreakAmbariUser.getSecret();
+        return isNotEmpty(getCloudbreakClusterManagerUserSecret()) ? getCloudbreakClusterManagerUserSecret() : cloudbreakAmbariUser.getSecret();
```
bergerdenes: this should go to the `getCloudbreakClusterManagerUserSecret`
eventually, this (and other old) getters will cease
you can even annotate it with @Depracated

core/src/main/resources/schema/app/20191114152446_CB-1358_Rename_every_Ambari_related_DB_column_to_generic_one.sql
```diff
@@ -0,0 +1,138 @@
+-- // CB-1358 Rename every Ambari related DB column to generic one
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE cluster
+ADD COLUMN IF NOT EXISTS cloudbreakclustermanageruser varchar(255);
+UPDATE cluster
```
bergerdenes: it is not useful in this deployment step.
please remove all updates from all tables.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/DatalakeResources.java
akanto: I mean, it does not make sense to rename if we are not using it at all. Then it can stay Ambari and in a next commit you can delete it.

PR #6729, opened 2019-11-15
CDPCP-1015. Usersync uses separate threadpool

handavid:
A separate threadpool is configured for use by the UserService and PasswordService.
Minor related configuration improvements are included:
- UsersyncPoller bean is conditionally created instead of checking the enabled flag
- application.yml is updated to move relevant configuration properties under freeipa.usersync

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
@@ -20,6 +20,7 @@ import javax.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.task.AsyncTaskExecutor;
 import org.springframework.stereotype.Service;
@@ -44,6 +45,7 @@ import com.sequenceiq.freeipa.client.FreeIpaClient;
 import com.sequenceiq.freeipa.client.FreeIpaClientException;
 import com.sequenceiq.freeipa.client.model.Config;
 import com.sequenceiq.freeipa.client.model.RPCResponse;
+import com.sequenceiq.freeipa.configuration.UsersyncConfig;
 import com.sequenceiq.freeipa.controller.exception.BadRequestException;
 import com.sequenceiq.freeipa.controller.exception.NotFoundException;
 import com.sequenceiq.freeipa.converter.freeipa.user.OperationToSyncOperationStatus;
@@ -69,7 +71,7 @@ public class UserService {
     private static final int DEFAULT_MAX_SUBJECTS_PER_REQUEST = 10;
 
     @VisibleForTesting
-    @Value("${freeipa.syncoperation.max-subjects-per-request:10}")
+    @Value("${freeipa.usersync.max-subjects-per-request}")
     int maxSubjectsPerRequest = DEFAULT_MAX_SUBJECTS_PER_REQUEST;
 
     @Inject
@@ -85,6 +87,7 @@ public class UserService {
     private FreeIpaUsersStateProvider freeIpaUsersStateProvider;
 
     @Inject
+    @Qualifier(UsersyncConfig.USERSYNC_TASK_EXECUTOR)
     private AsyncTaskExecutor asyncTaskExecutor;
 
     @Inject
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
@@ -10,7 +10,7 @@ import javax.inject.Inject;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 
@@ -20,6 +20,7 @@ import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
 import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
 import com.sequenceiq.cloudbreak.logger.MDCBuilder;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status;
 import com.sequenceiq.freeipa.api.v1.freeipa.user.model.SyncOperationStatus;
 import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.entity.UserSyncStatus;
@@ -27,6 +28,11 @@ import com.sequenceiq.freeipa.service.freeipa.user.model.UmsEventGenerationIds;
 import com.sequenceiq.freeipa.service.stack.StackService;
 
 @Service
+@ConditionalOnProperty(
+        value = "freeipa.usersync.poller.enabled",
+        havingValue = "true",
+        matchIfMissing = false
+)
 public class UsersyncPoller {
     @VisibleForTesting
     static final String INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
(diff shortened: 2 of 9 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -69,7 +71,7 @@
     private static final int DEFAULT_MAX_SUBJECTS_PER_REQUEST = 10;
 
     @VisibleForTesting
-    @Value("${freeipa.syncoperation.max-subjects-per-request:10}")
+    @Value("${freeipa.usersync.max-subjects-per-request:10}")
```
akanto: It has 10 as default value and also assigned the same value as a final static int

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -48,31 +53,28 @@
...
 
-    @Value("${freeipa.syncoperation.poller.enabled:true}")
-    private boolean enabled;
-
     @Inject
     private EntitlementService entitlementService;
 
-    @Scheduled(fixedDelayString = "${freeipa.syncoperation.poller.fixed-delay-millis:60000}",
-            initialDelayString = "${freeipa.syncoperation.poller.initial-delay-millis:300000}")
+    @Scheduled(fixedDelayString = "${freeipa.usersync.poller.fixed-delay-millis:60000}",
```
akanto: I am not sure whether it is intended or not, but if we have 3 pods, then this is expected to run 3 times, exactly in the same time, on the same data, doing the same thing

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -48,31 +53,28 @@
...
     @VisibleForTesting
     void syncFreeIpaStacks() {
-        threadBasedUserCrnProvider.setUserCrn(INTERNAL_ACTOR_CRN);
-        Optional<String> requestId = Optional.of(UUID.randomUUID().toString());
-        LOGGER.debug("Setting request id = {} for this poll", requestId);
-        MDCBuilder.addRequestId(requestId.get());
         try {
+            threadBasedUserCrnProvider.setUserCrn(INTERNAL_ACTOR_CRN);
+            Optional<String> requestId = Optional.of(UUID.randomUUID().toString());
+            LOGGER.debug("Setting request id = {} for this poll", requestId);
```
akanto: If you log first and then you add the requestid to MDC Context then I am afraid the log context will not have the requestid

handavid:
@giladwolff 
> So the change itself looks good. I have a few questions though:
> 
> 1. Why do we think this is going to solve the problem? Is there a test we can write that would fail before this change and succeed with it?
> 2. Asked differently, what was the root cause for the original problem?
> 
> What is the MDCCleanerTaskDecorator btw?

The problem was that user sync was sharing a threadpool with the rest of the FMS. This threadpool was also used to run some tasks during the create requests. User sync used up the threadpool, which caused freeipa creation requests to fail. Unfortunately, I don't think there's a reasonable way to test this change.

The MDCCleanerTaskDecorator removes fields (e.g., requestId, flowId, environment, etc) from the MDC upon task completion so that erroneous values are not logged when the thread is re-used.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
handavid: I removed the default from the @Value annotation. It's useful to have the `DEFAULT_MAX_SUBJECTS_PER_REQUEST` so that a value is always assigned, even if the class is not instantiated through spring (i.e., for the test)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: This is a known issue. The operation will be rejected if another full user sync is currently running for the same environment so that the bulk of the work will not be duplicated. We have some other JIRAs for further improving the behavior.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: re-ordered. thanks!

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -45,31 +50,28 @@
...
-        LOGGER.debug("Setting request id = {} for this poll", requestId);
-        MDCBuilder.addRequestId(requestId.get());
         try {
+            ThreadBasedUserCrnProvider.setUserCrn(INTERNAL_ACTOR_CRN);
+            Optional<String> requestId = Optional.of(UUID.randomUUID().toString());
+            MDCBuilder.addRequestId(requestId.get());
+            LOGGER.debug("Setting request id = {} for this poll", requestId);
+
             LOGGER.debug("Attempting to sync users to FreeIPA stacks");
             List<Stack> stackList = stackService.findAllRunning();
```
lacikaaa: this will list all not terminated stack instead of running. I think you should add a new method which filters for state `AVAILABLE`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -45,31 +50,28 @@
...
     }
 
     @VisibleForTesting
     void syncFreeIpaStacks() {
-        ThreadBasedUserCrnProvider.setUserCrn(INTERNAL_ACTOR_CRN);
-        Optional<String> requestId = Optional.of(UUID.randomUUID().toString());
-        LOGGER.debug("Setting request id = {} for this poll", requestId);
-        MDCBuilder.addRequestId(requestId.get());
         try {
+            ThreadBasedUserCrnProvider.setUserCrn(INTERNAL_ACTOR_CRN);
```
lacikaaa: this won't work anymore when @sodre90 change is in, see please rebase your PR after his change is in

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -45,31 +50,28 @@
...
 
-    @Value("${freeipa.syncoperation.poller.enabled:true}")
-    private boolean enabled;
-
     @Inject
     private EntitlementService entitlementService;
 
-    @Scheduled(fixedDelayString = "${freeipa.syncoperation.poller.fixed-delay-millis:60000}",
-            initialDelayString = "${freeipa.syncoperation.poller.initial-delay-millis:300000}")
+    @Scheduled(fixedDelayString = "${freeipa.usersync.poller.fixed-delay-millis:60000}",
```
lacikaaa: can we remove the default value from here if it's defined in the `application.yml`?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
```diff
@@ -45,31 +50,28 @@
...
+            initialDelayString = "${freeipa.usersync.poller.initial-delay-millis:300000}")
```
lacikaaa: can we remove the default value from here if it's defined in the `application.yml`?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: added.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: rebased

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UsersyncPoller.java
handavid: done

PR #6730, opened 2019-11-15
CB-4481. Deregister CCM key on FreeIPA or cluster termination.

jadair-cloudera:
With these changes, the flows for terminating a FreeIPA instance or cluster
have been updated to deregister the corresponding CCM key.

Specifically:
1. The CCM gRPC helper classes have been updated to support calling
   unregisterSshTunnelingKey on the minasshd management service.
2. The CcmResourceTerminationListener interface has been added. Implementations
   of this interfact are responsible for deregistering the CCM key associated
   with a resource.
3. A default implementation of this interface which delegates to the CCM
   gRPC helper classes has been provided as a Spring Component.
4. The logic to generate CCM key IDs from resource CRNs has been refactored
   into a helper class.
5. Flow events, actions, states, etc. have been added on both the FreeIPA
   and stack side to check whether CCM is enabled, and if so, to trigger
   the CCM resource termination listener after deregistering with the
   cluster proxy.


ccm-connector/src/main/java/com/sequenceiq/cloudbreak/ccmimpl/altus/MinaSshdManagementClient.java
```diff
@@ -200,6 +202,49 @@ public GenerateAndRegisterSshTunnelingKeyPairResponse generateAndRegisterSshTunn
...
+        checkNotNull(minaSshdServiceId);
+        checkNotNull(keyId);
+
+        MinaSshdManagementBlockingStub blockingStub = newStub(requestId);
+        UnregisterSshTunnelingKeyRequest.Builder requestBuilder = UnregisterSshTunnelingKeyRequest.newBuilder()
+                .setMinaSshdServiceId(minaSshdServiceId)
+                .setKeyId(keyId);
+
+        try {
+            if (LOGGER.isDebugEnabled()) {
```
handavid: it doesn't look like you are doing any work in this block so the `LOGGER.isDebugEnabled` check is unnecessary

ccm-connector/src/main/java/com/sequenceiq/cloudbreak/ccmimpl/termination/DefaultCcmResourceTerminationListener.java
```diff
@@ -0,0 +1,40 @@
...
+public class DefaultCcmResourceTerminationListener implements CcmResourceTerminationListener {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCcmResourceTerminationListener.class);
+
+    @Inject
+    private GrpcMinaSshdManagementClient grpcMinaSshdManagementClient;
+
+    @Override
+    public void deregisterCcmSshTunnelingKey(@Nonnull String actorCrn, @Nonnull String accountId, @Nonnull String keyId) {
+        if (grpcMinaSshdManagementClient != null) {
```
handavid: I don't think we need to null-check the `grpdMinaSshdManagementClient`. The service will fail to start if the bean can't be wired in.

jadair-cloudera: Fixed.

ccm-connector/src/main/java/com/sequenceiq/cloudbreak/ccmimpl/altus/MinaSshdManagementClient.java
jadair-cloudera: Done.

keyki:
Is this ready to be merged or needs more review?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/handler/CcmKeyDeregistrationHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+        CcmKeyDeregistrationRequest request = requestEvent.getData();
+        if (Boolean.TRUE.equals(request.getUseCcm())) {
+            LOGGER.debug("De-registering CCM key for freeipa stack {}", request.getResourceId());
+            try {
+                ccmResourceTerminationListener.deregisterCcmSshTunnelingKey(request.getActorCrn(), request.getAccountId(), request.getKeyId());
+            } catch (Exception ex) {
+                LOGGER.error("CCM key de-registration failed", ex);
+            }
+        }
```
bhavanki: The other handler logs the fact that CCM is disabled; that would be useful here too.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/recipe/CcmKeyDeregisterSuccess.java
```diff
@@ -0,0 +1,10 @@
+package com.sequenceiq.cloudbreak.reactor.api.event.recipe;
+
+import com.sequenceiq.cloudbreak.reactor.api.event.StackEvent;
+
+public class CcmKeyDeregisterSuccess extends StackEvent {
```
bhavanki: This should probably not be called "Success" because the handler triggers the event even if deregistration fails. "Finished" makes sense, as used in the FreeIPA service.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/recipe/CcmKeyDeregisterHandler.java
```diff
@@ -0,0 +1,66 @@
...
+                    LOGGER.warn("CCM key deregister failed", ex);
+                }
+            } else {
+                LOGGER.info("CCM is DISABLED, skipping de-registering of key from CCM. Cluster CRN: {}",
+                        stack.getResourceCrn());
+            }
+            result = new CcmKeyDeregisterSuccess(stack.getId());
+        } catch (Exception ex) {
+            LOGGER.warn("CCM key deregister failed", ex);
+            result = new StackFailureEvent(request.getResourceId(), ex);
```
bhavanki: This block is weird because any exception thrown from the deregistration attempt is caught in a prior catch block. I guess this one only triggers if stackService.getByIdWithListsInTransaction fails, in which case the log message is misleading - it didn't really even try to deregister.

If this block stays, the logging should be at error level, since it leads to a StackFailureEvent.

common/src/main/java/com/sequenceiq/cloudbreak/ccm/key/CcmResourceUtil.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * Returns the CCM SSH tunneling key ID for the specified resource CRN.
+     *
+     * @param resourceCrn the resource CRN
+     * @return the CCM SSH tunneling key ID for the specified resource CRN
+     */
+    public static String getKeyId(String resourceCrn) {
+        return StringUtils.right(Objects.requireNonNull(resourceCrn, "resourceCrn is null"), CCM_KEY_ID_LENGTH);
```
bhavanki: This is just extracting the resource ID from the CRN, right? Consider just doing standard parsing.

```
return Crn.safeFromString(resourceCrn).getResource();
```

If not, it'd be helpful to add a comment explaining why 36 makes sense.

ccm-connector/src/main/java/com/sequenceiq/cloudbreak/ccmimpl/termination/DefaultCcmResourceTerminationListener.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCcmResourceTerminationListener.class);
+
+    @Inject
+    private GrpcMinaSshdManagementClient grpcMinaSshdManagementClient;
+
+    @Override
+    public void deregisterCcmSshTunnelingKey(@Nonnull String actorCrn, @Nonnull String accountId, @Nonnull String keyId) {
+        // JSA TODO get request ID from somewhere?
+        String requestId = UUID.randomUUID().toString();
```
bhavanki: Redbeams and datalake have a `ThreadBasedRequestIdProvider` which gets the incoming request's request ID via a request filter. Maybe a similar mechanism is called for here.

jadair-cloudera:
@keyki It is do not merge pending more testing. Right now, we know the attempt to deregister the key will fail, because the backend service requires the minasshd service ID, which we do not have (that's why we are passing a dummy ID). Until Harrison finishes the backend changes, I need to confirm that the failure of the deregistration does not cause a failure of the entire flow.

I would still appreciate your review, primarily around the flow changes.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/event/ccm/CcmKeyDeregistrationRequest.java
```diff
@@ -0,0 +1,41 @@
...
+    private String keyId;
+
+    private Boolean useCcm;
+
+    public CcmKeyDeregistrationRequest(Long stackId, Boolean forced, String actorCrn, String accountId, String keyId, Boolean useCcm) {
+        super(EventSelectorUtil.selector(CcmKeyDeregistrationRequest.class), stackId, forced);
+        this.actorCrn = Objects.requireNonNull(actorCrn, "actorCrn is null");
+        this.accountId = Objects.requireNonNull(accountId, "accountId is null");
+        this.keyId = Objects.requireNonNull(keyId, "keyId is null");
+        this.useCcm = useCcm;
```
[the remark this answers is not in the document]
jadair-cloudera: No, I want to preserve the distinction between unspecified and specifically off as long as possible.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/recipe/CcmKeyDeregisterRequest.java
```diff
@@ -0,0 +1,40 @@
...
+    private String keyId;
+
+    private Boolean useCcm;
+
+    public CcmKeyDeregisterRequest(Long stackId, String actorCrn, String accountId, String keyId, Boolean useCcm) {
+        super(stackId);
+        this.actorCrn = Objects.requireNonNull(actorCrn, "actorCrn is null");
+        this.accountId = Objects.requireNonNull(accountId, "accountId is null");
+        this.keyId = Objects.requireNonNull(keyId, "keyId is null");
+        this.useCcm = useCcm;
```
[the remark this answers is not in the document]
jadair-cloudera: No, I want to preserve the distinction between unspecified and specifically off as long as possible.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/recipe/CcmKeyDeregisterSuccess.java
jadair-cloudera: I struggled with the diamond problem in this code, and had to choose between making it consistent with the other classes in the same package (e.g. ClusterProxyDeregisterSuccess), or with the parallel classes in FreeIPA. Given that the code structure matches the other classes in the package, I will leave this as is.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/handler/CcmKeyDeregistrationHandler.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void accept(Event<CcmKeyDeregistrationRequest> requestEvent) {
+
+        CcmKeyDeregistrationRequest request = requestEvent.getData();
+        if (Boolean.TRUE.equals(request.getUseCcm())) {
+            LOGGER.debug("De-registering CCM key for freeipa stack {}", request.getResourceId());
+            try {
+                ccmResourceTerminationListener.deregisterCcmSshTunnelingKey(request.getActorCrn(), request.getAccountId(), request.getKeyId());
+            } catch (Exception ex) {
+                LOGGER.error("CCM key de-registration failed", ex);
```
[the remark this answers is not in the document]
jadair-cloudera: I don't think there's anything they can do. I'll change the severity, though.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/recipe/CcmKeyDeregisterHandler.java
jadair-cloudera: Will change the error message and log level.

PR #6734, opened 2019-11-18
CB-4491 Always created database in case of AWS

lnardai:
now it honors if the flag is set to false. Problem was only for AWS.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
@@ -21,7 +21,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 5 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -217,11 +218,15 @@ public Long getResourceIdByResourceName(String resourceName) {
 
     @VisibleForTesting
```
akanto: It is visible for testing, then can we have a proper test for it, please?

lnardai: Yeah sure.

lnardai: Refactored the ExternalDatabase configuration logic to another class and wrote some unit tests for it.

PR #6745, opened 2019-11-18
DRAFT Duplicate rest calls

akanto:


core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/DistroXCreationService.java
```diff
@@ -0,0 +1,48 @@
...
+import com.sequenceiq.distrox.v1.distrox.StackOperations;
+import com.sequenceiq.distrox.v1.distrox.converter.DistroXV1RequestToStackV4RequestConverter;
+import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
+
+@Service
+public class DistroXCreationService {
+
+    // TODO I am not sure whether we need lazy
+    @Inject
+    @Lazy
```
attilapalfi92: Without it, sometimes the beanpostprocessor that creates the proxy that wraps StackOperations for it's @Async method is not executed, but this bean is already injected here. It causes spring context to be in invalid state and the app does not start. With the @Lazy the beanpostprocessor is able to run.

akanto: But I think the lazy just hides the real problem

attilapalfi92: https://stackoverflow.com/questions/28985144/spring-autowired-aop-circular-dependency/56234376#56234376

This is the only thing we found to mitigate this

PR #6749, opened 2019-11-18
CB-4525 ValidationResult should not have duplicate errors

risdenk:
Using the `ValidationResultBuilder` should not have duplicate errors. This changes the internal representation from an `ArrayList` to `TreeSet`


risdenk:
The failure is real:
```
org.opentest4j.AssertionFailedError: expected: <1. error> but was: <1. error
 >
```

I'll take a look tomorrow.

PR #6761, opened 2019-11-19
CB-4543 Enable controller logging

lacikaaa:


redbeams/src/main/java/com/sequenceiq/redbeams/aspect/ControllerLogContextAspects.java
```diff
@@ -48,6 +50,9 @@ public void buildLogContextForControllerCalls(JoinPoint joinPoint) {
             logContextService.buildMDCParams(joinPoint.getTarget(), paramNames, args);
 
             LOGGER.debug("Added selected controller method parameters to MDC; param names {}", Arrays.toString(paramNames));
+            String anonymizedArgs = Arrays.stream(args).map(Object::toString).map(AnonymizerUtil::anonymize).collect(Collectors.joining());
+            LOGGER.debug("A controller method has been intercepted: {} with params {}, {}, MDC logger context is built.", joinPoint.toShortString(),
+                    sig.getParameterNames(), anonymizedArgs);
```
bhavanki: Probably need to say `Arrays.toString(paramNames)` here for correct output. (Same for other changes.) Also, the names are logged already on line 52, so maybe just skip them here.

lacikaaa: good point, I rewrote to `Arrays.toString`

PR #6771, opened 2019-11-20
CB-4444 Fix token expiration issue (version 1)

oleewere:
Use credential provider with aws client builders instead of deprecated client creation -> clients always accessing credentials through credential providers (with depricated client -> it creates a static credential provider without refresh function, and that is used there)

Solution 1

Pros:
- we can refresh the token more often (based on our custom cache), even if expiration time is close (can be configurable)
- log expiration time for role arn

Cons:
- custom credential provider

(note, not tested too well yet + if it will be acceptable, i will change the base branch to rc-2.15 if it's needed)

solution 2:  https://github.com/hortonworks/cloudbreak/pull/6772

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSessionCredentialProvider.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSessionCredentialProvider.java
@@ -1,0 +1,66 @@
+package com.sequenceiq.cloudbreak.cloud.aws;
+
+import java.util.Date;
+
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
+
+public class AwsSessionCredentialProvider implements AWSCredentialsProvider {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AwsSessionCredentialProvider.class);
+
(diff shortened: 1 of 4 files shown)


oleewere:
retest this please

oleewere:
retest this please

oleewere:
@sodre90 
i have reverted the usage of Amazon ClientBuilders -> reason is that, because of as we have used the deprecated form for creating a client, we often try to change its state later in our code -> the support of deprecated clients are solved in a way that the aws code checks was the client created with builder or not (with a volatile isImmutable variable), if it was it throws an exception anywhere where we want to change the state of the client (we are doing that in multiple places in the code)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSessionCredentialProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        this.awsCredentialView = awsCredentialView;
+        this.awsSessionCredentialClient = awsSessionCredentialClient;
+        this.expirationTimeCheck = expirationTimeCheck;
+        this.forceTokenRefreshTimeInMin = forceTokenRefreshTimeInMin;
+    }
+
+    @Override
+    public AWSCredentials getCredentials() {
+        AwsSessionCredentials sessionCredentials = awsSessionCredentialClient.retrieveCachedSessionCredentials(awsCredentialView);
+        sessionCredentials = checkExpirationTime(sessionCredentials, sessionCredentials.getExpiration());
```
sodre90: Why pass expiration if it is in sessionCredentials?

oleewere: right, make no sense to have that in parameter, probably i left that there as the implementation changed

oleewere:
retest this please

PR #6784, opened 2019-11-21
CB-4532 StructuredEvent ResourceType is not filled in case of REST calls

lacikaaa:


lacikaaa:
please wait with the merge for the UI

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/StructuredEventFilter.java
```diff
@@ -86,11 +88,17 @@
...
+    private static final String CRN = "crn";
+
     private final List<String> skippedHeadersList = Lists.newArrayList("authorization");
 
     private final Map<String, WorkspaceResourceRepository<?, ?>> pathRepositoryMap = new HashMap<>();
 
-    private final Pattern extendRestParamsFromResponsePattern = Pattern.compile("\"id\":([0-9]*)");
+    private final Pattern extendIdRestParamFromResponsePattern = Pattern.compile("\"" + ID + "\":([0-9]*)");
+
+    private final Pattern extendCrnRestParamFromResponsePattern = Pattern.compile("\"" + CRN + "\":\"([0-9a-zA-Z:-]*)\"");
```
attilapalfi92: What is extending crn rest param from response exactly? I don't understand what this pattern does.

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/StructuredEventFilter.java
```diff
@@ -236,41 +244,72 @@ private void putResourceIdFromRepository(ContainerRequestContext requestContext,
...
-                    resourceId = matcher.group(1);
-                }
-            }
+        boolean resourceIdIsAbsentOrNull = isResourceIdIsAbsentOrNull(params);
+        boolean resourceCrnIsAbsentOrNull = isResourceCrnIsAbsentOrNull(params);
+        if (responseBody != null && (resourceIdIsAbsentOrNull || resourceCrnIsAbsentOrNull)) {
+            Set<String> resourcesParamsToCollect = new HashSet<>();
+            decorateSetWithValueIfNecessary(resourceIdIsAbsentOrNull, ID, resourcesParamsToCollect);
+            decorateSetWithValueIfNecessary(resourceCrnIsAbsentOrNull, CRN, resourcesParamsToCollect);
+            Map<String, String> resourceParams = extractResourceIdFromJson(responseBody, resourcesParamsToCollect);
```
attilapalfi92: `extractResourceIdFromJson`: is it only extracting the resourceId, or other params as well?

lacikaaa: thanks, I forgot to rename

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/StructuredEventFilter.java
lacikaaa: check the tests then :) when the response is not json. I don't know when it can happen

attilapalfi92: then call it extracting instead of extending please, because with this name it does not make sense

PR #6786, opened 2019-11-22
CB-1141 Add the ability of the FMS to provision multiple FreeIPA nodes

jamisonbennett:
This adds support to the FreeIPA management service to support
provisioning multiple instances, but this does not change the
environment service which is hardcoded to 1 instance.

The primary gateway is chosen as the CA master, the CRL master, the
node for configuring the FreeIPA servers, and the first FreeIPA node
to be installed.

All the other FreeIPA nodes are installed next as a part of the
FreeIPA install step. During the FreeIPA post install, the network
topology is updated to be a more complete mesh with a maximum of 4
replication agreements.

This was tested manually by setting the MASTER_NODE_COUNT in
FreeIpaCreationHandler.java (which is not included in this change).

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaInstallService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaInstallService.java
@@ -9,11 +9,14 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
-import org.springframework.beans.factory.annotation.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
+import org.springframework.beans.factory.annotation.Value;
 
 import com.sequenceiq.cloudbreak.auth.altus.model.AltusCredential;
 import com.sequenceiq.cloudbreak.orchestrator.exception.CloudbreakOrchestratorException;
+import com.sequenceiq.cloudbreak.orchestrator.exception.CloudbreakOrchestratorFailedException;
 import com.sequenceiq.cloudbreak.orchestrator.host.HostOrchestrator;
 import com.sequenceiq.cloudbreak.orchestrator.model.GatewayConfig;
 import com.sequenceiq.cloudbreak.orchestrator.model.Node;
@@ -32,6 +35,7 @@ import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.orchestrator.StackBasedExitCriteriaModel;
 import com.sequenceiq.freeipa.service.AltusMachineUserService;
 import com.sequenceiq.freeipa.service.GatewayConfigService;
+import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
 import com.sequenceiq.freeipa.service.freeipa.FreeIpaService;
 import com.sequenceiq.freeipa.service.freeipa.dns.ReverseDnsZoneCalculator;
 import com.sequenceiq.freeipa.service.stack.NetworkService;
@@ -40,6 +44,8 @@ import com.sequenceiq.freeipa.service.stack.instance.InstanceMetaDataService;
 
 @Service
 public class FreeIpaInstallService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FreeIpaInstallService.class);
 
     @Value("${info.app.version:}")
     private String version;
@@ -74,6 +80,9 @@ public class FreeIpaInstallService {
     @Inject
     private ReverseDnsZoneCalculator reverseDnsZoneCalculator;
 
+    @Inject
+    private FreeIpaClientFactory freeIpaClientFactory;
+
     public void installFreeIpa(Long stackId) throws CloudbreakOrchestratorException {
         Stack stack = stackService.getByIdWithListsInTransaction(stackId);
         Set<InstanceMetaData> instanceMetaDatas = stack.getNotDeletedInstanceMetaDataSet();
@@ -84,18 +93,32 @@ public class FreeIpaInstallService {
                 .collect(Collectors.toSet());
         FreeIpa freeIpa = freeIpaService.findByStack(stack);
 
+        if (allNodes.isEmpty()) {
+            String errorMessage = "There are no nodes to install with FreeIPA.";
+            LOGGER.error(errorMessage);
+            throw new CloudbreakOrchestratorFailedException(errorMessage);
+        }
+        GatewayConfig primaryGatewayConfig = gatewayConfigService.getPrimaryGatewayConfig(stack);
+
         SaltConfig saltConfig = new SaltConfig();
         Map<String, SaltPillarProperties> servicePillarConfig = saltConfig.getServicePillarConfig();
+        Set<Object> hosts = allNodes.stream().map(n ->
+                Map.of("ip", n.getPrivateIp(),
+                        "fqdn", n.getHostname()))
+                .collect(Collectors.toSet());
         Map<String, String> subnetWithCidr = networkService.getFilteredSubnetWithCidr(stack);
         String reverseZones = reverseDnsZoneCalculator.reverseDnsZoneForCidrs(subnetWithCidr.values());
-        Map<String, String> freeipaPillar = Map.of("realm", freeIpa.getDomain().toUpperCase(),
+        Map<String, Object> freeipaPillar = Map.of("realm", freeIpa.getDomain().toUpperCase(),
                 "domain", freeIpa.getDomain(),
                 "password", freeIpa.getAdminPassword(),
-                "reverseZones", reverseZones);
+                "reverseZones", reverseZones,
+                "hosts", hosts,
+                "admin_user", freeIpaClientFactory.getAdminUser(),
+                "freeipa_to_replicate", primaryGatewayConfig.getHostname());
         servicePillarConfig.put("freeipa", new SaltPillarProperties("/freeipa/init.sls", Collections.singletonMap("freeipa", freeipaPillar)));
         decoratePillarsWithTelemetryConfigs(stack, servicePillarConfig);
         hostOrchestrator.initSaltConfig(gatewayConfigs, allNodes, saltConfig, new StackBasedExitCriteriaModel(stackId));
-        hostOrchestrator.runService(gatewayConfigs, allNodes, saltConfig, new StackBasedExitCriteriaModel(stackId));
+        hostOrchestrator.installFreeIPA(primaryGatewayConfig, gatewayConfigs, allNodes, new StackBasedExitCriteriaModel(stackId));
     }
 
     private void decoratePillarsWithTelemetryConfigs(Stack stack, Map<String, SaltPillarProperties> servicePillarConfig) {
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
@@ -1,0 +1,164 @@
+package com.sequenceiq.freeipa.service.freeipa.flow;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.sequenceiq.freeipa.client.FreeIpaClient;
+import com.sequenceiq.freeipa.client.FreeIpaClientException;
+import com.sequenceiq.freeipa.client.model.TopologySegment;
+import com.sequenceiq.freeipa.client.model.TopologySuffix;
+import com.sequenceiq.freeipa.entity.InstanceMetaData;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
+import com.sequenceiq.freeipa.service.stack.StackService;
+import org.springframework.stereotype.Service;
+
+import javax.inject.Inject;
+import java.util.Iterator;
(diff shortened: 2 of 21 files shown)


keyki:
Thank you for the PR. We will review it shortly.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/init.sls
```diff
@@ -17,17 +26,36 @@ net.ipv6.conf.lo.disable_ipv6:
     - mode: 700
     - source: salt://freeipa/scripts/freeipa_install.sh
 
+{% if salt['pillar.get']('freeipa:mode') == 'FIRST_INSTALL' %}
 install-freeipa:
   cmd.run:
-    - name: /opt/salt/scripts/freeipa_install.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/freeipa_install-executed
+    - name: /opt/salt/scripts/freeipa_install.sh "FIRST_INSTALL" && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/freeipa_install-executed
```
akanto: Could you evaluate solving this problem using grains instead of pillar?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaInstallService.java
```diff
@@ -71,16 +81,36 @@ public void installFreeIpa(Long stackId) throws CloudbreakOrchestratorException
...
-                "domain", freeIpa.getDomain(),
-                "password", freeIpa.getAdminPassword());
+        Set<Object> hosts = allNodes.stream().map(n ->
+                Map.of("ip", n.getPrivateIp(),
+                        "fqdn", n.getHostname()))
+                .collect(Collectors.toSet());
+        Map<String, Object> freeipaPillar = new HashMap<>();
+        freeipaPillar.put("realm", freeIpa.getDomain().toUpperCase());
+        freeipaPillar.put("hosts", hosts);
+        freeipaPillar.put("mode", "FIRST_INSTALL");
```
akanto: Usually, we use grains, better to say the role field of grains, to target certain states to a particular node.  E.g.  ambari_server_standby: https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/top.sls#L106

I think we can use the following roles:
freeipa_primary
freeipa_replica

freeipa/src/main/resources/freeipa-salt/salt/freeipa/init.sls
jamisonbennett: I updated it to use grains for the role assignment. I still have to see why the integration tests are failing (they don't run locally for me so I will see if I can get help on running it tomorrow).

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaInstallService.java
jamisonbennett: I updated it to use grains for the role assignment. I still have to see why the integration tests are failing (they don't run locally for me so I will see if I can get help on running it tomorrow).

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
```diff
@@ -0,0 +1,32 @@
+/opt/salt/scripts/fix_next_update_grace_period.sh:
```
lacikaaa: instead of this script, could we use a salt state `file.replace`, as I checked the script it replaces a file content

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/fix_next_update_grace_period.sh
```diff
@@ -0,0 +1,20 @@
...
+
+function setCaConfiguration() {
+    PROPERTY=$1
+    VALUE=$2
+    ESCAPED_PROPERTY=${PROPERTY//./[.]}
+    TMP_FILE=`mktemp`
+    cat /var/lib/pki/pki-tomcat/ca/conf/CS.cfg | sed "s/.*$ESCAPED_PROPERTY=.*/$PROPERTY=$VALUE/g" > $TMP_FILE
+    cat $TMP_FILE > /var/lib/pki/pki-tomcat/ca/conf/CS.cfg
+    rm -f $TMP_FILE
+    systemctl restart pki-tomcatd@pki-tomcat
```
lacikaaa: this should be definitely be in a salt state

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/fix_next_update_grace_period.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+
+function setCaConfiguration() {
+    PROPERTY=$1
+    VALUE=$2
+    ESCAPED_PROPERTY=${PROPERTY//./[.]}
+    TMP_FILE=`mktemp`
+    cat /var/lib/pki/pki-tomcat/ca/conf/CS.cfg | sed "s/.*$ESCAPED_PROPERTY=.*/$PROPERTY=$VALUE/g" > $TMP_FILE
```
lacikaaa: this could be replaced with `sed -i`, so you don't need a temp file

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -492,6 +492,40 @@ public void changePrimaryGateway(GatewayConfig formerGateway, GatewayConfig newP
...
+
+            // add freeipa_replica role to the others
+            Set<String> replicaServersHostnames = allGatewayConfigs.stream()
+                    .filter(gwc -> !gwc.getHostname().equals(primaryGateway.getHostname()))
+                    .map(GatewayConfig::getHostname).collect(Collectors.toSet());
+            LOGGER.debug("Set replica FreeIPA: {}", replicaServersHostnames);
+            runSaltCommand(sc, new GrainAddRunner(replicaServersHostnames, allNodes, "freeipa_replica"), exitCriteriaModel);
+
+            // salt '*' state.highstate
+            runNewService(sc, new HighStateRunner(replicaServersHostnames, allNodes), exitCriteriaModel);
```
lacikaaa: it would be nice to see the comments as method names :)

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -492,6 +492,40 @@ public void changePrimaryGateway(GatewayConfig formerGateway, GatewayConfig newP
...
+    @Override
+    public void installFreeIPA(GatewayConfig primaryGateway, List<GatewayConfig> allGatewayConfigs, Set<Node> allNodes,
+            ExitCriteriaModel exitCriteriaModel) throws CloudbreakOrchestratorException {
+        LOGGER.debug("Set primary FreeIPA: {}", primaryGateway);
+        Set<String> serverHostname = Collections.singleton(primaryGateway.getHostname());
+        try (SaltConnector sc = createSaltConnector(primaryGateway)) {
+            // set freeipa_primary role
+            runSaltCommand(sc, new GrainAddRunner(serverHostname, allNodes, "freeipa_primary"), exitCriteriaModel);
+
+            // salt '*' state.highstate
```
lacikaaa: I think `HighStateRunner` makes this comment unnecessary

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
```diff
@@ -0,0 +1,163 @@
...
+        for (TopologySegment segment : segmentsToRemove) {
+            freeIpaClient.deleteTopologySegment(topologySuffixCn, segment);
+        }
+    }
+
+    @VisibleForTesting
+    public static class UnorderedPair {
+        private final String l;
+
+        private final String r;
```
lacikaaa: could you rename these to `left` and `right` as their get method states also. thanks

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
```diff
@@ -0,0 +1,163 @@
...
+                    return s;
+                }).collect(Collectors.toSet());
+        List<TopologySuffix> topologySuffixes = freeIpaClient.findAllTopologySuffixes();
+        for (TopologySuffix topologySuffix : topologySuffixes) {
+            createMissingSegments(freeIpaClient, topologySuffix.getCn(), topology);
+            removeExtraSegements(freeIpaClient, topologySuffix.getCn(), topology);
+        }
+    }
+
+    @VisibleForTesting
```
lacikaaa: if it's public, do we need this annotation?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
```diff
@@ -0,0 +1,163 @@
...
+                .collect(Collectors.toSet());
+        Set<TopologySegment> segmentsToRemove = freeIpaClient.findTopologySegments(topologySuffixCn).stream()
+                .filter(segment -> !topologyToKeep.contains(new UnorderedPair(segment.getLeftNode(), segment.getRightNode())))
+                .collect(Collectors.toSet());
+        for (TopologySegment segment : segmentsToRemove) {
+            freeIpaClient.deleteTopologySegment(topologySuffixCn, segment);
+        }
+    }
+
+    @VisibleForTesting
```
lacikaaa: if it's public, do we need this annotation?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/primary-install.sls
```diff
@@ -0,0 +1,12 @@
+install-freeipa:
+  cmd.run:
+    - name: /opt/salt/scripts/freeipa_install.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/freeipa_install-executed
+    - env:
+        - FPW: {{salt['pillar.get']('freeipa:password')}}
+        - DOMAIN: {{salt['pillar.get']('freeipa:domain')}}
+        - REALM: {{salt['pillar.get']('freeipa:realm')}}
+        - ADMIN_USER: {{salt['pillar.get']('freeipa:admin_user')}}
+        - FREEIPA_TO_REPLICATE: {{salt['pillar.get']('freeipa:freeipa_to_replicate')}}
```
lacikaaa: this isn't used in the script I think, could you remove? thanks

lacikaaa:
`This was tested manually by setting the MASTER_NODE_COUNT in
FreeIpaCreationHandler.java (which is not included in this change).`
I'm wondering if it would make sense to not increase the master nodes count, and instead add slaves (https://github.com/hortonworks/cloudbreak/blob/master/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/instance/InstanceGroupType.java#L5) so it would be clear that there is one primary GW node and some other GW. This won't make a real difference in the end as it would result in the same output if you check this part of the code: https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/MetadataSetupService.java#L59-L71

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
jamisonbennett: Thanks that works much better. I didn't know about `file.replace`.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/fix_next_update_grace_period.sh
jamisonbennett: This entire script was replaced with the salt `file.replace`.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
jamisonbennett: Thanks for catching this, I meant to make it package private, so I made that correction.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaTopologyService.java
jamisonbennett: Thanks for catching this, I meant to make it package private, so I made that correction.

lacikaaa:
let me rephrase it. if you send every instance in with `MASTER` type, you let the code select on of them to be `PRIMARY_GATEWAY`, other will be simply `GATEWAY`. If you send in 1 `MASTER` and the others as `SLAVE`s, then the latter ones become `GATEWAY`. That's why I wrote it won't make a big difference, at least for now.

holleyism:
If I understand this correctly, if we make 1 Master and the others Slaves, and the master fails, we won't have any new CRLs until the master node is recovered because the Slaves won't have salt master to be able to receive configuration changes.  Isn't this a risk that if for some reason the master takes a much longer time to recover, we're left without CRL functionality?

PR #6791, opened 2019-11-22
CB-4575 Polling aws statuses after the start or stop requests fired

topolyai5:


keyki:
We should cover the polling with some unit tests

PR #6792, opened 2019-11-22
CB-4501. Enable freeipa cloud storage logging by default.

oleewere:


oleewere:
dummy change so im merging it

PR #6800, opened 2019-11-25
CB-4609 Update openstack4j-connectors dependency

daszabo:
Basic cluster creation, upscale, downscale and termination scenarios tested.


pdarvasi:
@daszabo Could you pls check the integration tests?

PR #6804, opened 2019-11-25
CDPCP-1140 Set FreeIPA admin password expiration

lacikaaa:
- set admin password expiration on creation
- to update existing we run a modification for admin password expiration on application startup which can be dropped later when all env is updated

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
@@ -1,0 +1,61 @@
+package com.sequenceiq.freeipa.init;
+
+import java.util.List;
+import java.util.UUID;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.core.task.AsyncTaskExecutor;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.logger.MDCBuilder;
+import com.sequenceiq.freeipa.client.FreeIpaClient;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
+import com.sequenceiq.freeipa.service.freeipa.flow.FreeIpaPostInstallService;
+import com.sequenceiq.freeipa.service.stack.StackService;
+
+@Component
+public class ModifyAdminPassordExpiration implements ApplicationListener<ContextRefreshedEvent> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ModifyAdminPassordExpiration.class);
+
+    @Inject
+    private StackService stackService;
+
+    @Inject
+    private FreeIpaClientFactory freeIpaClientFactory;
+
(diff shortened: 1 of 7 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Inject
+    private UserService userService;
+
+    @Override
+    public void onApplicationEvent(ContextRefreshedEvent event) {
+        asyncTaskExecutor.submit(this::modifyPasswordExpiratiomForAllStack);
+    }
+
+    private void modifyPasswordExpiratiomForAllStack() {
```
handavid: typo in method name: Expiratiom -> Expiration

freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.sequenceiq.cloudbreak.logger.MDCBuilder;
+import com.sequenceiq.freeipa.client.FreeIpaClient;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
+import com.sequenceiq.freeipa.service.freeipa.user.UserService;
+import com.sequenceiq.freeipa.service.stack.StackService;
+
+@Component
+public class ModifyAdminPassordExpiration implements ApplicationListener<ContextRefreshedEvent> {
```
handavid: Are we concerned about Spring triggering multiple ContextRefreshEvents?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
```diff
@@ -399,4 +400,18 @@ void validateCrnFilter(Set<String> crnFilter, Crn.ResourceType resourceType) {
     private Set<String> union(Collection<String> collection1, Collection<String> collection2) {
         return Stream.concat(collection1.stream(), collection2.stream()).collect(Collectors.toSet());
     }
+
+    public void modifyAdminPasswordExpirationIfNeeded(FreeIpaClient client) throws FreeIpaClientException {
```
handavid: The UserService currently only performs the  user sync. I don't think this method really fits in well in the UserService. Consider moving it to another (perhaps new) class.

freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
lacikaaa: thanks, fixed it

freeipa/src/main/java/com/sequenceiq/freeipa/init/ModifyAdminPassordExpiration.java
lacikaaa: not really, code is idempotent, and we will drop it after it has been rolled out to prod

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserService.java
lacikaaa: moved it, but I think then `UserService` name is not correct for this class, it should be `UserSyncService`

PR #6812, opened 2019-11-25
CB-4194 Remove HiveServer2 kerberos authentication safety valve setting

risdenk:
OPSAPS-53430 starting in CM 7.1.0 gbn://1643975 or
higher removed the need to set the HS2 kerberos
authentication safety valve settings. This change
checks if the CM version is >=7.1.0 and if so
doesn't set the HS2 kerberos authentication configs.

Adds a test to verify behavior.

This follows the pattern from CB-4250 to detect CM 
version.


risdenk:
This is targetting master. I don't see any reason to rush this into CB 2.16.

risdenk:
Looks like results from failed integration test weren't available anymore by the time I checked. Rebased on master to get latest test results.

risdenk:
Well looks like integration test failure previously wasn't an issue :)

PR #6817, opened 2019-11-26
CB-4625. Fix AWS throttling errors during credential verification

oleewere:
- creating 5 second cache for verification (in order to not use aws client too often, especially problematic in parallel scenarios - like creating multiple credentials for e2e tests )
- also adding retry around request exceeded calls -> that can avoid throttling errors if an another application uses the same aws account for iam operations -> added 1 sec / 5 times retry as well as the issue happened during create credential actions (2 sec delay with 15 times will be too large there)

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsClient.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsClient.java
@@ -36,6 +36,7 @@ import com.amazonaws.services.securitytoken.AWSSecurityTokenService;
 import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
 import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonAutoScalingRetryClient;
 import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonCloudFormationRetryClient;
+import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonIdentityManagementRetryClient;
 import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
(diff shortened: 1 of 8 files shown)


environment/src/main/resources/schema/app/20191126150347_CB-4625_credential_skip_verify.sql
```diff
@@ -0,0 +1,9 @@
+-- // CB 4625 add skip verify credential field
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE credential ADD COLUMN  IF NOT EXISTS skip_verify NOT NULL DEFAULT false;
```
doktoric: update for existing?

environment/src/main/java/com/sequenceiq/environment/credential/domain/Credential.java
```diff
@@ -45,6 +45,9 @@
     @Column
     private Boolean govCloud = Boolean.FALSE;
 
+    @Column
+    private Boolean skipVerify = Boolean.FALSE;
```
doktoric: boolean ?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsCredentialView.java
```diff
@@ -69,6 +70,16 @@ public Boolean isGovernmentCloudEnabled() {
         return false;
     }
 
+    public Boolean isSkipVerify() {
```
doktoric: boolean?

oleewere:
checking cache works well with credential id or not (probably not, if that's the case a different object key generation  will be needed for the cache)

oleewere:
retest this please

PR #6820, opened 2019-11-26
CB-3857 configure Knox to allow access to only those who have environ…

keyki:


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology_api.xml.j2
```diff
@@ -25,6 +25,88 @@
...
+                <name>solr.acl</name>
+                <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
+            </param>
+            <param>
+                <name>webhbase.acl</name>
+                <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
+            </param>
+            <param>
+                <name>webhdfs.acl</name>
+                <value>*;users;*</value>
```
risdenk: This needs salt variable

keyki: Good catch.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/cdp-token.xml.j2
```diff
@@ -23,6 +23,33 @@
...
+         <enabled>true</enabled>
+         <param>
+            <name>CENTRAL_GROUP_CONFIG_PREFIX</name>
+            <value>gateway.group.config.</value>
+         </param>
+      </provider>
+
+      <provider>
+         <role>authorization</role>
+         <name>AclsAuthz</name>
```
lmccay: The CompositeAuthz provider was intended to be used here which will allow us to enable Ranger as well at some point. This shouldn't be considered a blocker but it needs to be understood that enabling Ranger in the gateway will just require some changes to this config - which would be needed anyway to add Ranger but there will be a little more as well.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -25,6 +25,148 @@
...
+        <provider>
+            <role>authorization</role>
+            <name>AclsAuthz</name>
+            <enabled>true</enabled>
+            <param>
+                <name>knox.acl.mode</name>
+                <value>AND</value>
+            </param>
+            <param>
+                <name>atlas.acl</name>
```
lmccay: I will look at the possibility of simplifying this with an "all.acl" or the like - so that we don't have to duplicate it for all services. Also, not a blocker for the current needs.

keyki:
Tested with real UMS also, it generated this new virtual group `_c_env_assignees_f3b60c2`.

keyki:
We can merge this now and deploy to dev.

PR #6821, opened 2019-11-27
CB-4642 Create more specific Event history messages

bbihari:


common/src/main/resources/messages/messages.properties
```diff
@@ -107,8 +107,8 @@ cluster.ambari.cluster.services.stopping=Stopping Cloudera Manager services.
 cluster.ambari.cluster.services.stopped=Cloudera Manager services have been stopped.
```
lnardai: Since this is a production issue, and not a risky change, it should go to 2.15.

common/src/main/resources/messages/messages.properties
```diff
@@ -107,8 +107,8 @@ cluster.ambari.cluster.services.stopping=Stopping Cloudera Manager services.
...
 cluster.autorecovery.requested=Cluster autorecovery requested, failed nodes: {0}
 cluster.manualrecovery.requested=Cluster manual recovery requested on entities: {0}
-cluster.failednodes.reported=Manual recovery is needed for the following failed nodes: {0}
-cluster.recoverednodes.reported=Recovered nodes detected: {0}
+cluster.failednodes.reported=Node(s) {0} status became UNHEALTHY, manual recovery is needed.
```
lnardai: This is still ambiguous. First question will be why is it unhealthy. It should state the exact thing that happens in the background. The goal is to guide the user to the source of the problem, like we did with the VPC settings.

```
The CM agent is not responding on Node(s) {0}. Recovery might be needed.
Please check CM UI for further information. It's also possible that the CM server is unavailable, in that case please check why CM server is not responding on the instance. 
```
CM agent/server became available on Node(s) {0}.

It would be even better if we could write a message based on wether the node is a primary gateway or not, but if it's complicated we can leave it as written above.

PR #6827, opened 2019-11-27
CB-4605 Freeipa password generator

tiborpopovics:


.cbd_history
```diff
@@ -0,0 +1 @@
+2019-11-27 16:55:08 tiborpopovics kill
```
keyki: this file is not needed

.cbd_output_history
```diff
@@ -0,0 +1,7 @@
+2019-11-27 16:55:07 tiborpopovics kill
```
keyki: same

.starttime
```diff
@@ -0,0 +1 @@
+1574870107
```
keyki: same

Profile
```diff
@@ -0,0 +1,4 @@
+export CB_INSTANCE_UUID="24b26578-442c-4f4b-acfe-2a37da4ce769"
+export CB_INSTANCE_NODE_ID="c1cc1655-e6b9-4f00-a0b0-0876a9782492"
+export CB_INSTANCE_PROVIDER="unknown"
+export CB_INSTANCE_REGION="unknown"
```
keyki: same

common/src/main/java/com/sequenceiq/cloudbreak/util/FreeIpaPasswordUtil.java
```diff
@@ -0,0 +1,31 @@
+package com.sequenceiq.cloudbreak.util;
+
+import org.apache.commons.lang3.RandomStringUtils;
+
+public class FreeIpaPasswordUtil {
+
+    private static final int PWD_PART_LENGTH = 8;
+
+    private static final String SPECIAL_CHARS = "?!./-*_+";
```
keyki: Can you please remove `/` so we can use the password safely in console

.cbd_history
tiborpopovics: deleted

.cbd_output_history
tiborpopovics: deleted

.starttime
tiborpopovics: deleted

Profile
tiborpopovics: deleted

PR #6830, opened 2019-11-28
CB-3591 Generate swagger from tests

foldik:
It generates the following files during test:
- `build/swagger/cb.json`
- `build/swagger/autoscale.json`
- `build/swagger/datalake.json`
- `build/swagger/redbeams.json`
- `build/swagger/freeipa.json`
- `build/swagger/environment.json`

Will be useful to speed up other pipelines.


autoscale-api/src/main/java/com/sequenceiq/periscope/api/endpoint/v1/AlertEndpoint.java
```diff
@@ -37,7 +37,7 @@
 
 @Path("/v1/clusters/{clusterId}/alerts")
 @Consumes(MediaType.APPLICATION_JSON)
-@Api(value = "/v1/alerts", description = ALERT_DESCRIPTION, protocols = "http,https")
+@Api(value = "/v1/alerts", description = ALERT_DESCRIPTION, protocols = "http,https", consumes = JSON)
```
foldik: This doesn't change the API. It only needed to generate the same json that Spring would generate (with the help of the above `@Consumes` annotation).

doktoric: this should be consumes = MediaType.APPLICATION_JSON just to be consistent

foldik: Yes I agree. I saw other examples as well where we had other constants like this which is already defined in the `MediaType`. Shall I remove them?

PR #6831, opened 2019-11-28
CB-4664 security group validator should check registration type

doktoric:


environment/src/main/java/com/sequenceiq/environment/environment/validation/securitygroup/AwsEnvironmentSecurityGroupValidator.java
```diff
@@ -31,7 +32,7 @@ public void validate(EnvironmentDto environmentDto, ValidationResult.ValidationR
             if (onlyOneSecurityGroupIdDefined(securityAccessDto)) {
                 resultBuilder.error(securityGroupIdsMustBePresent());
             } else if (isSecurityGroupIdDefined(securityAccessDto)) {
-                if (!Strings.isNullOrEmpty(environmentDto.getNetwork().getNetworkCidr())) {
+                if (RegistrationType.CREATE_NEW.equals(environmentDto.getNetwork().getRegistrationType())) {
```
akanto: enums can be compared with ==, it is easier to read

akanto: is it possible that the getNetwork returns with null, because in that case, NPE would occur

akanto: we need logs here, since the entire validation does not contain even a single line of log, and I was not able to find the error based on logs just with repro the case

doktoric: the network can not be null because we have converter for every platform

keyki:
Comments will be addressed later

PR #6835, opened 2019-11-28
CB-4670 extra logs and disabling the faulty part of the validation

akanto:
Added extra logs to figure out why log storage validation fails on mow dev. It cannot be reproduced in local.


lajosrodek:
Follow-up to #6639

lajosrodek:
Seen this change too late. As discussed in the associated Jira ticket, I believe the original validation logic was right as per the recent changes in IDBMMS rule generation (namely, the Baseline / Ranger Audit role being inferred from the Log instance profile).

PR #6838, opened 2019-11-29
Fixing unit test failures on jenkins

attilapalfi92:


environment/src/test/java/com/sequenceiq/environment/environment/service/DatahubServiceTest.java
```diff
@@ -24,9 +24,15 @@
...
 import com.sequenceiq.environment.environment.poller.ClusterPollerResultEvaluator;
 import com.sequenceiq.environment.environment.poller.DatahubPollerProvider;
+import com.sequenceiq.environment.store.EnvironmentInMemoryStateStore;
 
+import net.jcip.annotations.NotThreadSafe;
+
+@NotThreadSafe
 class DatahubServiceTest {
 
+    private static final long ENV_ID = 1L;
```
topolyai5: it will break sometimes the test

attilapalfi92: Why is that?

PR #6839, opened 2019-11-29
CB-4574: improve env creation error handling

tiborpopovics:


environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -83,24 +84,32 @@ public EnvironmentCreationService(
     }
 
     public EnvironmentDto create(EnvironmentCreationDto creationDto) {
-        if (environmentService.isNameOccupied(creationDto.getName(), creationDto.getAccountId())) {
-            throw new BadRequestException(String.format("Environment with name '%s' already exists in account '%s'.",
-                    creationDto.getName(), creationDto.getAccountId()));
+        Environment environment = new Environment();
+        try {
```
attilapalfi92: Would be nicer if you moved the beginning of the try { after the validateCreation call and then creating this empty env variable would be unnecessary.

attilapalfi92: Also, in this case the environment is being saved even when some validation error happened and we should just reject the request with a http 400.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentModificationService.java
```diff
@@ -140,9 +141,16 @@ private void editNetworkIfChanged(Environment environment, EnvironmentEditDto ed
...
-            BaseNetwork network = createAndSetNetwork(environment, editDto.getNetworkDto(), editDto.getAccountId());
-            if (network != null) {
-                environment.setNetwork(network);
+            try {
+                BaseNetwork network = createAndSetNetwork(environment, editDto.getNetworkDto(), editDto.getAccountId());
+                if (network != null) {
+                    environment.setNetwork(network);
+                }
+            } catch (Exception e) {
+                environment.setStatus(EnvironmentStatus.CREATE_FAILED);
```
attilapalfi92: Use the UPDATE_FAILED status

PR #6840, opened 2019-11-29
CB-4675 fix NPE when no security group is defined

keyki:


keyki:
@schfeca75 we need to cover this with unit tests in a follow up PR

PR #6842, opened 2019-11-29
CB-4683 Fix gradle Eclipse IDE project generation

stoty:
Adds an eclipse specific fixup function to the root gradle.build file that removes the erronous  eclipse classpath entries for protobuf generated code, and adds the correct classpath entries.

The function is only ever run by the eclipse project setup gradle plugin, so it should have minimal risk.

testing done:

run './gradlew eclipse', import the project into eclipse. 
Eclipse does not have dependency problems with the fix.

run './gradlew build'
check that the build completes.

Closes #CB-4683

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,4 +180,20 @@ subprojects {
       url "https://repository.cloudera.com/cloudera/list/cm-private/"
     }
   }
+
+  eclipse {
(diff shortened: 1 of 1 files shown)


doktoric:
thanks @stoty

PR #6844, opened 2019-11-29
CB-4678 If error happens during Image selection from img catalog

lnardai:
-should return 400 BadRequest with detailed information

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
@@ -397,22 +397,30 @@ public class StackCreatorService {
             return null;
         }
         return executorService.submit(() -> {
-            try {
-                boolean base = blueprintService.isAmbariBlueprint(blueprint) ? shouldUseBaseAmbariImage(clusterRequest) : shouldUseBaseCMImage(clusterRequest);
-                LOGGER.info("The stack with name {} will use base image: {}", stackName, base);
-                return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user, image -> true);
-            } catch (CloudbreakImageNotFoundException | CloudbreakImageCatalogException e) {
-                throw new BadRequestException(e.getMessage(), e);
-            }
+            boolean base = blueprintService.isAmbariBlueprint(blueprint) ? shouldUseBaseAmbariImage(clusterRequest) : shouldUseBaseCMImage(clusterRequest);
+            LOGGER.info("The stack with name {} will use base image: {}", stackName, base);
+            return imageService.determineImageFromCatalog(workspace.getId(), stackRequest.getImage(), platformString, blueprint, base, user, image -> true);
         });
     }
 
     private StatedImage getImageCatalog(Future<StatedImage> imgFromCatalogFuture) {
+        int time = 1;
+        TimeUnit unit = TimeUnit.MINUTES;
         return Optional.ofNullable(imgFromCatalogFuture).map(f -> {
             try {
-                return f.get(1, TimeUnit.HOURS);
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                throw new RuntimeException("Image catalog determination failed", e);
+                return f.get(time, unit);
+            } catch (ExecutionException e) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -397,21 +397,25 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
...
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof CloudbreakImageNotFoundException) {
+                    throw new BadRequestException("Image id settings are incorrect.", e);
+                }
+                if (e.getCause() instanceof CloudbreakImageCatalogException) {
+                    throw new BadRequestException("Image catalog settings are incorrect.", e);
+                }
+                throw new RuntimeException("Unkown error happened when determining image catalog", e);
+            } catch (InterruptedException | TimeoutException e) {
```
topolyai5: I think we can add more specific message in these cases

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -397,21 +397,25 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
...
                 return f.get(1, TimeUnit.HOURS);
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof CloudbreakImageNotFoundException) {
+                    throw new BadRequestException("Image id settings are incorrect.", e);
+                }
+                if (e.getCause() instanceof CloudbreakImageCatalogException) {
+                    throw new BadRequestException("Image catalog settings are incorrect.", e);
+                }
+                throw new RuntimeException("Unkown error happened when determining image catalog", e);
```
topolyai5: Are we send back to the ui this message?

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -397,21 +397,25 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
...
         return Optional.ofNullable(imgFromCatalogFuture).map(f -> {
             try {
                 return f.get(1, TimeUnit.HOURS);
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof CloudbreakImageNotFoundException) {
+                    throw new BadRequestException("Image id settings are incorrect.", e);
+                }
+                if (e.getCause() instanceof CloudbreakImageCatalogException) {
+                    throw new BadRequestException("Image catalog settings are incorrect.", e);
```
topolyai5: maybe better if we use e.getCause() instead of e

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -397,21 +397,25 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
...
     }
 
     private StatedImage getImageCatalog(Future<StatedImage> imgFromCatalogFuture) {
         return Optional.ofNullable(imgFromCatalogFuture).map(f -> {
             try {
                 return f.get(1, TimeUnit.HOURS);
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof CloudbreakImageNotFoundException) {
+                    throw new BadRequestException("Image id settings are incorrect.", e);
```
topolyai5: maybe better if we use e.getCause() instead of e

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
lnardai: Yes

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
lnardai: Split the to cases, also fin tuned the messages.

PR #6851, opened 2019-12-02
DISTX-316. Support cloudwatch / fluentd plugin upgrade & small refactor (Preview / POC)

oleewere:
- upgrade fluentd plugins at startup (check versions where needed, also uses custom repo if required for the plugin)
- add cloudwatch support as log service (use the cloud storage identity there) -> main reason, could be useful for e2e tests for node service logs (using cloud storage instance profile + storageLocation as cloudwatch log group) - hack here to handle cloudwatch output as s3 from cloud api perspective as instance profile stuff is built around cloud storage (would require larger change to support a different kind of log identity) - of course objectstorage validations are skipped if cloudwatch is used
- add stream key for cloudwatch (default host) - that determines which key should be used for creating the stream, in the example i have used component, but host is the default (for larger cluster it's better to handle the throughput by host, not by component)
- add new filters -> detect-exceptipons , for log service streams, in the end we want to send data to ES, but we do not want to send lines one-by-one especially for stack traces (for java apps), so we will try to cacth those multiline messages, aslo append the actual dates for the outputs (as we do not know the patterns) - this feature can be useful for the databus output as well (in order to append specific headers)
- small refactor -> FluentClusterDetails object, as in FluentConfigService the number of inputs stated to grow, would make sense to use an object there with a builder (regarding the cluster related inputs)
- upgrade databus plugin version, i have found an encoding bug, that i have fixed

cloudwatch example for sdx cluster: 
cloudwatch_poc

one record looks like this:
```json
{
    "component": "hdfs-datanode",
    "cluster": "oszabo-local-dl-3",
    "cluster_type": "datalake",
    "cluster_crn": "crn:cdp:datalake:us-west-1:cloudera:datalake:c4cac7f4-ade2-444c-8e51-83923d484fb3",
    "cluster_owner": "crn:cdp:iam:us-west-1:cloudera:user:oszabo@cloudera.com",
    "cluster_version": "2.17.0-dev.38-1-g014e082",
    "host": "oszabo-local-dl-3-master1",
    "@message": "2019-12-02 14:13:10,283 INFO org.apache.hadoop.hdfs.server.datanode.DataNode: Receiving BP-490250763-10.65.65.116-1575295912401:blk_1073741825_1001 src: /10.65.65.116:51001 dest: /10.65.65.116:1004",
    "@timestamp": "2019-12-02T14:13:10.284170550Z"
}
```

Tested against freeipa/datalake/datahub

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigView.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigView.java
@@ -17,8 +17,6 @@ public class FluentConfigView implements TelemetryConfigView {
 
     private static final String PROVIDER_PREFIX_DEFAULT = "stdout";
 
-    private static final String DBUS_APP_NAME_DEFAULT = "datahub";
-
     private static final String EMPTY_CONFIG_DEFAULT = "";
 
     private static final Integer PARTITION_INTERVAL_DEFAULT = 5;
@@ -27,11 +25,13 @@ public class FluentConfigView implements TelemetryConfigView {
 
     private final boolean cloudStorageLoggingEnabled;
 
+    private final boolean cloudLoggingServiceEnabled;
+
     private final boolean reportClusterDeploymentLogs;
 
     private final boolean meteringEnabled;
 
-    private final String databusAppName;
+    private final FluentClusterDetails clusterDetails;
 
     private final String user;
 
@@ -43,7 +43,7 @@ public class FluentConfigView implements TelemetryConfigView {
 
     private final String serviceLogFolderPrefix;
 
-    private final String platform;
+    private final String region;
 
     private final String providerPrefix;
 
@@ -59,6 +59,8 @@ public class FluentConfigView implements TelemetryConfigView {
 
     private final String s3LogArchiveBucketName;
 
+    private final String cloudwatchStreamKey;
+
     private final String logFolderName;
 
     private final Map<String, Object> overrideAttributes;
@@ -66,19 +68,21 @@ public class FluentConfigView implements TelemetryConfigView {
     private FluentConfigView(Builder builder) {
         this.enabled = builder.enabled;
         this.cloudStorageLoggingEnabled = builder.cloudStorageLoggingEnabled;
+        this.cloudLoggingServiceEnabled = builder.cloudLoggingServiceEnabled;
         this.reportClusterDeploymentLogs = builder.reportClusterDeploymentLogs;
         this.meteringEnabled = builder.meteringEnabled;
-        this.databusAppName = builder.databusAppName;
+        this.clusterDetails = builder.clusterDetails;
         this.user = builder.user;
         this.group = builder.group;
         this.serverLogFolderPrefix = builder.serverLogFolderPrefix;
         this.agentLogFolderPrefix = builder.agentLogFolderPrefix;
         this.serviceLogFolderPrefix = builder.serviceLogFolderPrefix;
-        this.platform = builder.platform;
+        this.region = builder.region;
         this.providerPrefix = builder.providerPrefix;
         this.partitionIntervalMin = builder.partitionIntervalMin;
         this.logFolderName = builder.logFolderName;
         this.s3LogArchiveBucketName = builder.s3LogArchiveBucketName;
+        this.cloudwatchStreamKey = builder.cloudwatchStreamKey;
         this.azureContainer = builder.azureContainer;
         this.azureStorageAccount = builder.azureStorageAccount;
         this.azureInstanceMsi = builder.azureInstanceMsi;
@@ -106,8 +110,12 @@ public class FluentConfigView implements TelemetryConfigView {
         return serviceLogFolderPrefix;
     }
 
-    public String getPlatform() {
-        return platform;
+    public String getRegion() {
+        return region;
+    }
+
+    public FluentClusterDetails getClusterDetails() {
+        return clusterDetails;
     }
 
     public String getProviderPrefix() {
@@ -124,6 +132,10 @@ public class FluentConfigView implements TelemetryConfigView {
 
     public String getS3LogArchiveBucketName() {
         return s3LogArchiveBucketName;
+    }
+
+    public String getCloudwatchStreamKey() {
+        return cloudwatchStreamKey;
     }
 
     public String getAzureStorageAccount() {
@@ -150,16 +162,16 @@ public class FluentConfigView implements TelemetryConfigView {
         return cloudStorageLoggingEnabled;
     }
 
+    public boolean isCloudLoggingServiceEnabled() {
+        return cloudLoggingServiceEnabled;
+    }
+
     public boolean isReportClusterDeploymentLogs() {
         return reportClusterDeploymentLogs;
     }
 
     public boolean isMeteringEnabled() {
         return meteringEnabled;
-    }
-
-    public String getDatabusAppName() {
-        return databusAppName;
     }
 
     public Map<String, Object> getOverrideAttributes() {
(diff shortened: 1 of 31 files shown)


oleewere:
retest this please

oleewere:
retest this please

oleewere:
for testing (env json) -both cloudwatch + upgraded databus:
```json
 "telemetry": {
    "features": {
      "reportDeploymentLogs" : {
        "enabled": true
      },
      "logging": {
        "cloudwatch": {
          "instanceProfile": "<instanceProfileArn>"
        },
        "storageLocation": "/test/oszabo"
      }
    }
  }
```

oleewere:
retest this please

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -73,6 +74,12 @@ private Logging createLoggingFromRequest(LoggingRequest loggingRequest) {
                 adlsGen2Params.setSecure(adlsGen2FromRequest.isSecure());
                 adlsGen2Params.setManagedIdentity(adlsGen2FromRequest.getManagedIdentity());
                 logging.setAdlsGen2(adlsGen2Params);
+            } else if (loggingRequest.getCloudwatch() != null) {
+                CloudwatchParams cloudwatchRequest = loggingRequest.getCloudwatch();
+                CloudwatchParams cloudwatchParams = new CloudwatchParams();
```
lajosrodek: Like in `TelemetryApiConverter`, this could be refactored using a copy constructor or static factory method.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -94,6 +101,12 @@ private LoggingResponse createLoggingResponseFromSource(Logging logging) {
                 adlsGen2Params.setSecure(logging.getAdlsGen2().isSecure());
                 adlsGen2Params.setManagedIdentity(logging.getAdlsGen2().getManagedIdentity());
                 loggingResponse.setAdlsGen2(adlsGen2Params);
+            } else if (logging.getCloudwatch() != null) {
+                CloudwatchParams cloudwatchParams = logging.getCloudwatch();
+                CloudwatchParams cloudwatchResponse = new CloudwatchParams();
```
lajosrodek: Like above.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/telemetry/TelemetryConverter.java
```diff
@@ -107,7 +120,7 @@ private Features createFeaturesFromRequest(FeaturesRequest featuresRequest) {
                 features.setReportDeploymentLogs(featuresRequest.getReportDeploymentLogs());
             } else {
                 FeatureSetting reportDeploymentLogsFeature = new FeatureSetting();
```
lajosrodek: Adding a log message like in the core `TelemetryConverter` could not hurt.

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigView.java
```diff
@@ -277,8 +296,8 @@ public Builder withServiceLogFolderPrefix(String serviceLogFolderPrefix) {
             return this;
         }
 
-        public Builder withPlatform(String platform) {
-            this.platform = platform;
+        public Builder withRegion(String region) {
```
lajosrodek: Where is this method supposed to be used? Not called anywhere at the moment.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
```diff
@@ -244,7 +280,7 @@ fluentd_delalyed_restart:
    cmd.run:
     - names:
        - "nohup sh /etc/td-agent/delayed_restart.sh > /etc/td-agent/delayed_restart.out 2>&1 &"
-       - "cp /etc/td-agent/td-agent_simple_profile.conf /etc/td-agent/td-agent.conf"
+       - "nohup sleep 30; cp /etc/td-agent/td-agent_simple_profile.conf /etc/td-agent/td-agent.conf &"
```
lajosrodek: Is 30 sec an empirical value? 😃

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
```diff
@@ -0,0 +1,51 @@
...
+    echo "Checking plugin ${plugin} with the right version ${version}"
+  fi
+  echo "Run check command: $check_command"
+  local result=$($check_command)
+  echo "Check ${plugin} command output: $result"
+  if [[ "$result" == "false" ]]; then
+    echo "Plugin ${plugin} does not exist, installing it..."
+    install_plugin "${plugin}" "${custom_repo}" "${version}"
+  else
+    echo "Pluging ${plugin} exists."
```
lajosrodek: Typo: Pluging -> Plugin.

orchestrator-salt/src/main/resources/salt-common/pillar/fluent/init.sls
```diff
@@ -6,16 +6,23 @@ fluent:
   agentLogFolderPrefix: /var/log
   serviceLogFolderPrefix: /var/log
   platform:
+  region:
```
lajosrodek: I think this will remain uninitialized due to `FluentConfigView.Builder.withRegion()` not being called anywhere.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -263,7 +277,6 @@ void setupCloudStorageAccountMapping(StackV4Request stackRequest, String environ
                 accountMapping.setGroupMappings(mappingsConfig.getGroupMappings());
                 accountMapping.setUserMappings(mappingsConfig.getActorMappings());
                 cloudStorage.setAccountMapping(accountMapping);
-                LOGGER.info("Initial account mappings fetched from IDBMMS: {}", JsonUtil.writeValueAsStringSilent(accountMapping));
```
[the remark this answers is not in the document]
oleewere: probably that happened because of conflicts, thanks for pointing it out

orchestrator-salt/src/main/resources/salt-common/pillar/fluent/init.sls
oleewere: that's true, what we are doing in fluentConfigView, all the properties can be overriden by fluentAttributes, the withRegion should be used if we would try to support cloudwatch from non-aws envs..i can remove the with method but that field will be a bit hidden

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
oleewere: it was added because at startup td-agent prints out the configuration then right after that re-reading it (so log did not tell something else is used),rearly it happens... probably few seconds wait would be enough before copying

environment/src/main/java/com/sequenceiq/environment/environment/v1/TelemetryApiConverter.java
```diff
@@ -188,4 +186,14 @@ private AdlsGen2CloudStorageV1Parameters convertAdlsV2(AdlsGen2CloudStorageV1Par
         }
         return adlsGen2CloudStorageV1Parameters;
     }
+
+    private CloudwatchParams convertCloudwatch(CloudwatchParams cloudwatch) {
```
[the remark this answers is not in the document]
oleewere: good idea, will do that here, i will create a jira to do the same with other params (s3,abfs etc.)

orchestrator-salt/src/main/resources/salt-common/pillar/fluent/init.sls
oleewere: i think it's just better to have that as an option in cloudwatch params (by default if it's not there, that is calculated but only on aws platform)

oleewere:
retest this please

PR #6852, opened 2019-12-02
CB-4643 Create module for status syncer

bbihari:


datalake/src/main/java/com/sequenceiq/datalake/flow/start/SdxStartActions.java
```diff
@@ -48,7 +48,7 @@
     private SdxStartService startService;
 
     @Inject
-    private SdxJobService sdxJobService;
+    private JobService jobService;
```
horadla23: please delete

bbihari: ok

PR #6857, opened 2019-12-03
CB-4701 External Rds on azure: handle username@hostname on local datahub

gergopapi2:


core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
```diff
@@ -0,0 +1,47 @@
...
+     * Converts a connection username to a db username.
+     * On azure, connection username follows the 'username@short-hostname' pattern.
+     * On postgres, however, users have to be in 'username' format, without the '@' sign => code has to take username.
+     *
+     * @param username a connection username
+     * @return returns the username part before the '@' sign
+     */
+    public String toDatabaseUsername(String username) {
+        int atIndex = username.indexOf("@");
+        return atIndex > -1 ? username.substring(0, atIndex) : username;
```
keyki: This could be simplified by using StringUtils.substringBefore(username, "@")

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    private static class DbHostnameMatcher {
```
lacikaaa: why is this separate inner class necessary?

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Service
+public class DbUsernameConverterService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DbUsernameConverterService.class);
+
+    public String toConnectionUsername(String host, String dbUser) {
+        if (DbHostnameMatcher.isAzure(host)) {
+            LOGGER.debug("Detected Azure database server {}, appending short hostname to connection username", host);
+            return dbUser + "@" + host.substring(0, host.indexOf('.'));
```
lacikaaa: this could also be simplified with `StringUtils.substringBefore`

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
gergopapi2: The goal was to express the higher cohesion between the isAzure method and the DbHostnameMatcher that is used from one place (hence embedded class). If you think this decreases readability, I can revert it back to a private method and a constant within DbUsernameConverterService.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
gergopapi2: Thanks, I added it.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
gergopapi2: Thanks, I added it.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Converts a connection username to a db username.
+     * On azure, connection username follows the 'username@short-hostname' pattern.
+     * On postgres, however, users have to be in 'username' format, without the '@' sign => code has to take username.
+     *
+     * @param username a connection username
+     * @return returns the username part before the '@' sign
+     */
+    public String toDatabaseUsername(String username) {
+        return username.contains("@") ? StringUtils.substringBefore(username, "@") : username;
```
keyki: I think substringBefore will return the original string if it does not contain the separator so it's unnecessary to check before.

gergopapi2: You are right ... Thanks!

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/DbUsernameConverterService.java
gergopapi2: As a result of our private discussion I got rid of the inner class.

PR #6867, opened 2019-12-03
CB-4709 Fix for Sync, when CM is not available

daszabo:
1. Check CM accessibility, before checking hosts, during sync flow. If CM is not available, set its state to unhealthy and send notification
2. Aggregated host saves and notification sending for multiple hosts for optimization

### DIFF
--- a/cluster-ambari/src/main/java/com/sequenceiq/cloudbreak/ambari/status/AmbariClusterStatusService.java
+++ b/cluster-ambari/src/main/java/com/sequenceiq/cloudbreak/ambari/status/AmbariClusterStatusService.java
@@ -11,6 +11,7 @@ import javax.inject.Inject;
 import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
 
 import com.sequenceiq.ambari.client.AmbariClient;
 import com.sequenceiq.cloudbreak.ambari.AmbariClientFactory;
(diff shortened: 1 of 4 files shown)


cluster-ambari/src/main/java/com/sequenceiq/cloudbreak/ambari/status/AmbariClusterStatusService.java
```diff
@@ -75,4 +75,9 @@ public ClusterStatusResult getStatus(boolean blueprintPresent) {
     public Map<String, String> getHostStatusesRaw() {
         return ambariClient.getHostStatuses();
     }
+
+    @Override
+    public boolean isCMRunning() {
```
lacikaaa: `isCMRunning` in ambari class?

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -948,28 +950,41 @@ public Cluster updateCreationDateOnCluster(Cluster cluster) {
...
+                    String hostNotificationMessage = updatedInstanceMetaData.stream()
+                            .map(instanceMetaData -> cloudbreakMessagesService.getMessage(Msg.CLUSTER_HOST_STATUS_UPDATED.code(),
+                                    Arrays.asList(instanceMetaData.getDiscoveryFQDN(), instanceMetaData.getInstanceStatus().name())))
+                            .collect(joining("; "));
+                    eventService.fireCloudbreakEvent(stack.getId(), AVAILABLE.name(), hostNotificationMessage);
+                    return stack.getCluster();
+                });
+            } catch (TransactionExecutionException e) {
+                throw new TransactionRuntimeExecutionException(e);
+            }
```
lacikaaa: this method got a little bit big and hard to overview, please refactor it into some smaller separate method

cluster-ambari/src/main/java/com/sequenceiq/cloudbreak/ambari/status/AmbariClusterStatusService.java
daszabo: Its CM as in Cluster Manager :trollface:

PR #6870, opened 2019-12-03
CB-4670 Reenable IDBroker role mapping validation

risdenk:
Reenables IDBroker role mapping validation since it is correct. The issue is IDBMMS changing from instance profile to role behind the scenes.


risdenk:
Ping @akanto for review? Didn't add anyone else to review since you reverted this originally.

lajosrodek:
Follow-up to #6835.

akanto:
I think the formatting of the error message could be a bit more structured, and a bit more user friendly

PR #6871, opened 2019-12-03
CB-4198 Remove HBase hbase.rest.support.proxyuser safety valve config

risdenk:
OPSAPS-53433 starting in CM 7.1.0 gbn://1639677 or
higher removed the need to set the HBase
hbase.rest.support.proxyuser safety valve config.
This change checks if the CM version is >=7.1.0 and
if so doesn't set the HBase REST proxyuser config.

Adds a test to verify behavior.

This follows the pattern from CB-4250 to detect CM
version.


risdenk:
Targetting CB master since no rush to get this in.

PR #6872, opened 2019-12-03
CB-4705 remove CRN from ThreadBasedUserCrnProvider in MDCCleanerThrea…

sodre90:


common/src/main/java/com/sequenceiq/cloudbreak/auth/ThreadBasedUserCrnProvider.java
```diff
@@ -33,27 +29,18 @@ public String getAccountId() {
...
-            stack = new Stack<>();
-            USER_CRN.set(stack);
-        }
-        if (!stack.isEmpty() && !stack.peek().equals(userCrn)) {
-            LOGGER.error("Trying to push crn to stack {} when it already contains {}", userCrn, stack.get(0));
-            // REMOVE IT UNTIL WE FIX
-            // throw new IllegalStateException(String.format("Trying to push crn to stack %s when it already contains %s", userCrn, stack.get(0)));
+    public static void setUserCrn(String userCrn) {
+        if (USER_CRN.get() != null) {
+            String errorMessage = String.format("Trying to set crn %s when it already contains %s, please check where we didn't remove it",
```
keyki: We can print out the calltrace here so would be easy to detect

sodre90: good idea, thanks

sodre90:
> ThreadBasedUserCrnProvider is now not a bean anymore. In this case unit testing/mocking is harder. CCan we use a singleton bean please?

Bean is not necessary. It is a static class, so I can set user CRN in a normal way without mocking like this: ThreadBasedUserCrnProvider.setUserCrn(USER_CRN);

PR #6873, opened 2019-12-04
Merge rc-2.16 to master

foldik:
@horadla23 please check if I missed something from the requestId related change.


foldik:
I need to add the requestId to MDCContext...

PR #6876, opened 2019-12-05
CB-4728 Deleting a broken Env that failed at FreeIPA create leaves an…

topolyai5:


environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentResourceDeletionService.java
```diff
@@ -54,25 +54,25 @@ public void deleteClusterDefinitionsOnCloudbreak(String environmentCrn) {
...
             propagateException("Failed to get SDX clusters from SDX service due to:", e);
         } catch (ProcessingException e) {
             propagateException("Failed to get SDX clusters from SDX service due to:", e);
         }
-        return clusterNames;
+        return clusterCrns;
     }
 
-    Set<String> getDatalakeClusterNames(Environment environment) {
+    public Set<String> getDatalakeClusterName(Environment environment) {
```
bergerdenes: could remain plural as it returns a `Set`

topolyai5: fixed

PR #6884, opened 2019-12-05
CB-4749. Add global databus secret/access keys to skip machine user generation

oleewere:
- add global config to enable to use a common altus credential for all cb cluster (that should be enabled for e2e environments)
- the shared altus credential is only used if that is set in telemetry request (as a feature setting) - if the global config is disabled, you cannot enable that feature
- access / secret key can be set in the configuration (use char array for the secret key)
- as many configurations passed to the converters, create config objects instead of extending the parameters (Telemetry + Altus config object)

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
@@ -2,11 +2,13 @@ package com.sequenceiq.cloudbreak.auth.altus.service;
 
 import java.util.Optional;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto;
+import com.sequenceiq.cloudbreak.altus.AltusDatabusConfiguration;
 import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
 import com.sequenceiq.cloudbreak.auth.altus.model.AltusCredential;
 
@@ -15,16 +17,36 @@ public class AltusIAMService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AltusIAMService.class);
 
+    private final boolean useSharedAltusCredential;
+
+    private final String sharedAltusAccessKey;
+
+    private final char[] sharedAltusSecretKey;
+
     private final GrpcUmsClient umsClient;
 
-    public AltusIAMService(GrpcUmsClient umsClient) {
+    public AltusIAMService(GrpcUmsClient umsClient,
+            AltusDatabusConfiguration altusDatabusConfiguration) {
         this.umsClient = umsClient;
+        this.useSharedAltusCredential = altusDatabusConfiguration.isUseSharedAltusCredential();
+        this.sharedAltusAccessKey = altusDatabusConfiguration.getSharedAccessKey();
+        this.sharedAltusSecretKey = altusDatabusConfiguration.getSharedSecretKey();
     }
 
     /**
      * Generate machine user with access keys
      */
-    public Optional<AltusCredential> generateMachineUserWithAccessKey(String machineUserName, String actorCrn) {
+    public Optional<AltusCredential> generateMachineUserWithAccessKey(String machineUserName, String actorCrn, boolean useSharedCredential) {
+        if (useSharedAltusCredential) {
+            LOGGER.debug("Use shared altus credential is turned on for generating altus credential and access keys.");
+            if (areDatabusCredentialsFilled(useSharedCredential)) {
+                LOGGER.debug("Access and secret keys are set manually application wide for Databus, skip machine user and access key generation");
+                return Optional.of(new AltusCredential(sharedAltusAccessKey, sharedAltusSecretKey));
+            } else {
+                LOGGER.debug("Use shared credential global config credential is set, but no shared access/secret keypair is used.");
+            }
+        }
+
         return Optional.of(umsClient.createMachineUserAndGenerateKeys(
                 machineUserName,
                 actorCrn,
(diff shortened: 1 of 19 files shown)


oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

oleewere:
retest this please

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
```diff
@@ -15,33 +17,67 @@
...
      * Generate machine user with access keys
      */
-    public Optional<AltusCredential> generateMachineUserWithAccessKey(String machineUserName, String actorCrn) {
+    public Optional<AltusCredential> generateMachineUserWithAccessKey(String machineUserName, String actorCrn, boolean useSharedCredential) {
+        if (useSharedAltusCredential) {
+            LOGGER.debug("Use shared altus credential is turned on for generating altus credential and access keys.");
+            if (areDatabusCredentialsFilled(useSharedCredential)) {
+                LOGGER.debug("Access and secret keys are set manually application wide for Databus, skip machine user and access key generation");
+                return Optional.of(new AltusCredential(accessKey, secretKey));
+            }
```
lajosrodek: Maybe adding an `else` clause with another log msg would be helpful for debugging. Alternatively, the 2 `if`s could be collapsed into a single one that checks `useSharedAltusCredential && areDatabusCredentialsFilled(useSharedCredential)`, similarly to `clearMachineUser()` below.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
```diff
@@ -15,33 +17,67 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AltusIAMService.class);
 
+    private final boolean useSharedAltusCredential;
+
+    private final String accessKey;
+
+    private final char[] secretKey;
```
lajosrodek: Not a big deal, but I suggest renaming these by adding the prefix `sharedAltus` (or just `shared` to match the getters of `AltusDatabusConfiguration`).

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
```diff
@@ -15,33 +17,67 @@
...
                     + "but note that you could have remaining UMS resources for your account", e);
         }
     }
 
+    public void clearMachineUser(String machineUserName, String actorCrn) {
+        clearMachineUser(machineUserName, actorCrn, false);
+    }
+
+    private boolean areDatabusCredentialsFilled(boolean useSharedCredential) {
+        return secretKey != null && StringUtils.isNoneBlank(accessKey, new String(secretKey));
```
lajosrodek: This will execute `new String(secretKey)` every time this method is called, creating a new object unnecessarily. Why not change the type of `secretKey` (the field in this class) to `String` and evaluate `new String(secretKey)` only once in the constructor?

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
```diff
@@ -15,33 +17,67 @@
...
             LOGGER.warn("Cluster Databus resource cleanup failed (fluent - databus user). It is not a fatal issue, "
                     + "but note that you could have remaining UMS resources for your account", e);
         }
     }
 
+    public void clearMachineUser(String machineUserName, String actorCrn) {
+        clearMachineUser(machineUserName, actorCrn, false);
+    }
+
+    private boolean areDatabusCredentialsFilled(boolean useSharedCredential) {
```
lajosrodek: `useSharedCredential` is not used anywhere in the logic below! Did you possibly intend to evaluate `useSharedCredential && secretKey != null && ...` instead?

common-model/src/main/java/com/sequenceiq/common/api/telemetry/doc/TelemetryModelDescription.java
```diff
@@ -15,6 +15,7 @@
     public static final String TELEMETRY_REPORT_DEPLOYMENT_LOGS_ENABLED = "enable cluster deployment log reporting.";
     public static final String TELEMETRY_CLOUDWATCH_PARAMS = "telemetry - cloudwatch releated parameters";
     public static final String TELEMETRY_CLOUDWATCH_PARAMS_REGION = "telemetry - cloudwatch related AWS region (should be used only outside of AWS platform)";
```
lajosrodek: Minor: Although this belongs to the earlier #6851, `cloudwatch` should be spelled with the right casing: `CloudWatch`.

common-model/src/main/java/com/sequenceiq/common/api/telemetry/doc/TelemetryModelDescription.java
```diff
@@ -15,6 +15,7 @@
...
+    public static final String TELEMETRY_USE_SHARED_ALTUS_CREDENTIAL_ENABLED = "enable shared altus credential usage";
```
lajosrodek: Minor: `altus` should be spelled as `Altus`.

common/src/main/java/com/sequenceiq/cloudbreak/altus/AltusDatabusConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public boolean isUseSharedAltusCredential() {
+        return useSharedAltusCredential;
+    }
+
+    public String getSharedAccessKey() {
+        return sharedAccessKey;
+    }
+
+    public char[] getSharedSecretKey() {
```
lajosrodek: Minor: If we want complete immutability, a clone of `sharedSecretKey` shall be returned.

oleewere: for now it's stay as is

oleewere:
retest this please

PR #6900, opened 2019-12-08
Dr fixes 1

doktoric:


core/src/main/java/com/sequenceiq/cloudbreak/service/ServiceEndpointCollector.java
```diff
@@ -249,9 +249,6 @@ private SSOType getSSOType(ExposedService exposedService, Gateway gateway) {
                 case KUDU:
                     addKuduUrl(managerIp, gateway, privateIps, autoTlsEnabled, urls);
                     break;
-                case SCHEMA_REGISTRY:
-                    addSchemaRegistry(managerIp, gateway, topologyName, exposedService, urls);
```
keyki: addSchemaRegistry function can be removed as well

PR #6908, opened 2019-12-09
CB-4787 fix duplicate identity-assertion field in cdp-token topology

keyki:


keyki:
> FWIW - don't technically need the else statement in there. A default `identity-assertion` provider is added even if it doesn't exist.

Got it. I'll remove it then.

keyki:
Done.

PR #6909, opened 2019-12-09
CB-4769 validateRolePolicies hits AmazonIdentityManagement rate limit

risdenk:
This removes the newly added `AmazonIdentityManagementRetryClient` from CB-4625 and increases the retries for the IAM client with the builder. I think this could be done for all the AWS SDK clients, but requires moving to the builder model. 

This also adds try/catch around the simulate principal policy in the IDBroker mapping validation so that it will not cause cluster creation to fail incorrectly if there is an IAM issue.


risdenk:
Test failure looks like a TLS certificate error not related to this change?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/client/AmazonIdentityManagementRetryClient.java
```diff
@@ -1,23 +0,0 @@
...
-
-    private final Retry retry;
-
-    public AmazonIdentityManagementRetryClient(AmazonIdentityManagement client, Retry retry) {
-        this.client = client;
-        this.retry = retry;
-    }
-
-    public SimulatePrincipalPolicyResult simulatePrincipalPolicy(SimulatePrincipalPolicyRequest request) {
-        return retry.testWith1SecDelayMax5Times(() -> mapThrottlingError(() -> client.simulatePrincipalPolicy(request)));
```
oleewere: you can also remove testWith1SecDelayMax5Times from RetryService then. Probably it also used in a test somewhere (HeartbeatServiceTest or something similar)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsClient.java
```diff
@@ -50,6 +48,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsClient.class);
 
+    // Default retries is 3. This allows for more time for backoff during throttling
+    private static final int MAX_CLIENT_RETRIES = 30;
```
oleewere: is it retries every operation where there is an error code in the response, im wondering is this valid then (like with the retry clients we are catching specifically the rate limit issues)

risdenk: This uses https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/retry/PredefinedRetryPolicies.html which uses the default backoff and retry conditions.

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/retry/PredefinedRetryPolicies.html#DEFAULT_RETRY_CONDITION

This highlights that for throttling and other retriable errors are retried. Its not for any error, but for ones that are AWS service side and we want to retry for. The ones that we may hit more often would be the throttling unless AWS has some big issue.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/client/AmazonIdentityManagementRetryClient.java
risdenk: yea there is probably more cleanup that can be done. There seems to be a lot of pressure to get this fixed so went with minimal changes.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsIamService.java
```diff
@@ -235,14 +235,14 @@ public Policy getPolicy(String policyFileName, Map<String, String> replacements)
...
-            Collection<Policy> policies) {
+            Collection<Policy> policies) throws AmazonIdentityManagementException {
         List<EvaluationResult> evaluationResults = new ArrayList<>();
         for (Policy policy : policies) {
             for (Statement statement : policy.getStatements()) {
                 SortedSet<String> actions = getStatementActions(statement);
                 SortedSet<String> resources = getStatementResources(statement);
-                evaluationResults.addAll(simulatePrincipalPolicy(iam, role.getArn(),
-                        actions, resources));
+                List<EvaluationResult> results = simulatePrincipalPolicy(iam, role.getArn(), actions, resources);
```
bergerdenes: why?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsIamService.java
```diff
@@ -235,14 +235,14 @@ public Policy getPolicy(String policyFileName, Map<String, String> replacements)
      * @return list of evaluation results
      */
     public List<EvaluationResult> validateRolePolicies(AmazonIdentityManagement iam, Role role,
-            Collection<Policy> policies) {
+            Collection<Policy> policies) throws AmazonIdentityManagementException {
         List<EvaluationResult> evaluationResults = new ArrayList<>();
         for (Policy policy : policies) {
             for (Statement statement : policy.getStatements()) {
```
sodre90: I merged it, but there are 4 embedded "for cycles" still:
```
        for (Policy policy : policies) {	        for (Policy policy : policies) {
            for (Statement statement : policy.getStatements()) {
...
                         for (Policy policy : policies) {
            for (Statement statement : policy.getStatements()) {
                SortedSet<String> actions = getStatementActions(statement);
                SortedSet<String> resources = getStatementResources(statement);
                evaluationResults.addAll(simulatePrincipalPolicy(iam, role.getArn(),
                        actions, resources));
            }
        }
```
Can we optimize this like we did here?

https://github.com/hortonworks/cloudbreak/blob/e98a3317bfba3731ba23605f4ecef1869acf89fa/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialVerifier.java#L75
https://github.com/hortonworks/cloudbreak/blob/e98a3317bfba3731ba23605f4ecef1869acf89fa/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialVerifier.java#L60

lajosrodek:
Follow-up to #6639, #6835, #6870, #6817.

PR #6915, opened 2019-12-10
CB-4802 Update displayed component versions for OpDB components

stoty:
also add Phoenix to compinent version list for CDH 7.1

Updates the descriptors that specify the component versions to display.
Adds Phoenix to 7.1 descriptor

I have used the versions from the POM, even though they may not be accurate. I left the HBase version in 7.1 alone, because while it does not match the POM version, it matches reality.

Test:
Start modified CB

Check that the UI displays the new versions on DataHub creation screen
Check that the UI display Phoenix with logo and correct version on DataHub creation screen

Closes #CB-4802

### DIFF
--- a/template-manager-cmtemplate/src/main/resources/cloudera-manager-template/cdh/7.0.2.json
+++ b/template-manager-cmtemplate/src/main/resources/cloudera-manager-template/cdh/7.0.2.json
@@ -14,11 +14,11 @@
     },
     {
       "name": "ZOOKEEPER",
-      "version": "3.4.6"
+      "version": "3.5.5"
     },
     {
       "name": "HBASE",
-      "version": "2.2.2"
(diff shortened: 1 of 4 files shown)


template-manager-cmtemplate/src/main/resources/cloudera-manager-template/cdh/7.0.2.json
```diff
@@ -14,11 +14,11 @@
     },
     {
       "name": "ZOOKEEPER",
-      "version": "3.4.6"
+      "version": "3.5.5"
```
tamaashu: here we still had 3.4.6

stoty:
I have updated the PR.
I fixed the Zookeeper version for 7.0.0
I updated the HBase versions to match the real included upstream released version number.

PR #6925, opened 2019-12-10
CB-4702 Azure e2e test for repair

gergopapi2:
Contains a refactor enabling the porting of existing aws tests to other cloud platforms. Currently the SdxRepairTest was done only.

This would be the first step in a series of refactors.

The goal is to separate the actual tests from cloud vendor specific logic to directly manipulate cloud resources (e.g. stop host instances). 
Tests would be refactored as SdxRepairTest looks like now:
- test case moved into a domain specific (here sdx) package from a cloud-vendor specific one (like aws)
- test class would be renamed into domain specific name (SdxRepairTest)
- code to directly manipulate cloud resources in test cases (e.q. delete instances in host groups) would be hidden behind the CloudFunctionality interface.
- cloud-vendor specific parts of requests (e.g. cloud storage in env / sdx requests) will be hidden in dtos as is done already.

In this way actual test cases have to be written once only (with possible minor alterations) and runnable on all cloud platforms (given the needed cloud functionality is written already).

Follow-up PRs would contain refactoring of  the other e2e tests as well.


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/environment/EnvironmentTestDto.java
```diff
@@ -185,8 +186,11 @@ public EnvironmentTestDto withS3Guard(String tableName) {
     }
 
     public EnvironmentTestDto withS3Guard() {
-        String tableName = dynamoTableName + '-' + UUID.randomUUID().toString();
-        return withS3Guard(tableName);
+        if (CloudPlatform.AWS.equals(getTestContext().getCloudProvider().getCloudPlatform())) {
+            String tableName = dynamoTableName + '-' + UUID.randomUUID().toString();
+            return withS3Guard(tableName);
+        }
+        return this;
```
afarsang: would be a good idea to drop a warn message here into the logs, don't you think?

gergopapi2: In personal communication we agreed to place a log line with severity INFO. Thanks!

PR #6934, opened 2019-12-12
CDPCP-1212 Update password expiration for hashed password codepath

handavid:
The UMS now provides a password expiration time along with the
workload credentials. This value is epoch millis. User sync has
been updated to set this expiration time when setting the hashed
password. The userMod request was refactored, moving the param
map creation into the FreeIpaClient to match the pattern used by
other FreeIpaClient methods.

The UsersState was slightly refactored to split out the UMS-specific
information into a UmsUsersState because the UsersState model is
also used to represent the ipa state.

The Mock user management service has been modified to return an
expiration time that is different from the max expiration time
that the FMS uses.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -3,8 +3,10 @@
 
 <module name="Checker">
     <property name="severity" value="error" />
+    <module name="SuppressWarningsFilter" />
     <module name="TreeWalker">
         <property name="tabWidth" value="4" />
+        <module name="SuppressWarningsHolder" />
 
         <!-- Format -->
         <module name="EmptyForIteratorPad" />
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
@@ -2,6 +2,10 @@ package com.sequenceiq.freeipa.client;
 
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -34,7 +38,9 @@ import com.sequenceiq.freeipa.client.model.User;
 
 public class FreeIpaClient {
 
-    public static final String PASSWORD_EXPIRATION_DATETIME = "20380101000000Z";
+    public static final String MAX_PASSWORD_EXPIRATION_DATETIME = "20380101000000Z";
+
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMddHHmmssVV");
 
     private static final Logger LOGGER = LoggerFactory.getLogger(FreeIpaClient.class);
 
@@ -150,7 +156,7 @@ public class FreeIpaClient {
                 "sn", lastName,
                 "loginshell", "/bin/bash",
                 "random", true,
-                "setattr", "krbPasswordExpiration=" + PASSWORD_EXPIRATION_DATETIME
+                "setattr", "krbPasswordExpiration=" + MAX_PASSWORD_EXPIRATION_DATETIME
         );
         return (User) invoke("user_add", flags, params, User.class).getResult();
     }
@@ -159,11 +165,35 @@ public class FreeIpaClient {
         // FreeIPA expires any password that is set by another user. Work around this by
         // performing a separate API call to set the password expiration into the future
         userMod(user, "userpassword", password);
-        updateUserPasswordExpiration(user);
+        updateUserPasswordMaxExpiration(user);
     }
 
-    public void updateUserPasswordExpiration(String user) throws FreeIpaClientException {
-        userMod(user, "setattr", "krbPasswordExpiration=" + PASSWORD_EXPIRATION_DATETIME);
+    public void updateUserPasswordExpiration(String user, Optional<Instant> expiration) throws FreeIpaClientException {
+        String passwordExpirationDate = formatDate(expiration);
+        userMod(user, "setattr", "krbPasswordExpiration=" + passwordExpirationDate);
+    }
+
+    public void updateUserPasswordMaxExpiration(String user) throws FreeIpaClientException {
+        updateUserPasswordExpiration(user, Optional.empty());
+    }
+
+    public User userSetPasswordHash(String user, String hashedPassword,
+            String unencryptedKrbPrincipalKey, Optional<Instant> expiration) throws FreeIpaClientException {
+        String passwordExpirationDate = formatDate(expiration);
+        Map<String, Object> params =
+                Map.of("setattr", List.of(
+                        "cdpHashedPassword=" + hashedPassword,
+                        "cdpUnencryptedKrbPrincipalKey=" + unencryptedKrbPrincipalKey,
+                        "krbPasswordExpiration=" + passwordExpirationDate));
+        return userMod(user, params);
+    }
+
+    String formatDate(Optional<Instant> instant) {
+        if (instant.isPresent()) {
+            return DATE_TIME_FORMATTER.format(ZonedDateTime.ofInstant(instant.get(), ZoneOffset.UTC));
+        } else {
+            return MAX_PASSWORD_EXPIRATION_DATETIME;
+        }
     }
 
     public User userMod(String user, String key, Object value) throws FreeIpaClientException {
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
@@ -80,11 +80,11 @@ public class FreeIpaPostInstallService {
 
     private void modifyAdminPasswordExpirationIfNeeded(FreeIpaClient client) throws FreeIpaClientException {
         Optional<User> user = client.userFind(freeIpaClientFactory.getAdminUser());
-        if (user.isPresent() && !FreeIpaClient.PASSWORD_EXPIRATION_DATETIME.equals(user.get().getKrbPasswordExpiration())) {
+        if (user.isPresent() && !FreeIpaClient.MAX_PASSWORD_EXPIRATION_DATETIME.equals(user.get().getKrbPasswordExpiration())) {
             User actualUser = user.get();
             LOGGER.debug(String.format("Modifying user [%s] current password expiration time [%s] to [%s]",
-                    actualUser.getUid(), actualUser.getKrbPasswordExpiration(), FreeIpaClient.PASSWORD_EXPIRATION_DATETIME));
-            client.updateUserPasswordExpiration(actualUser.getUid());
+                    actualUser.getUid(), actualUser.getKrbPasswordExpiration(), FreeIpaClient.MAX_PASSWORD_EXPIRATION_DATETIME));
+            client.updateUserPasswordMaxExpiration(actualUser.getUid());
         } else if (user.isEmpty()) {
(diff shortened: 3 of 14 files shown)


config/checkstyle/checkstyle.xml
```diff
@@ -3,8 +3,10 @@
 
 <module name="Checker">
     <property name="severity" value="error" />
+    <module name="SuppressWarningsFilter" />
     <module name="TreeWalker">
         <property name="tabWidth" value="4" />
+        <module name="SuppressWarningsHolder" />
```
giladwolff: For my edification, why these changes?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -150,7 +156,7 @@ public User userAdd(String user, String firstName, String lastName) throws FreeI
                 "sn", lastName,
                 "loginshell", "/bin/bash",
                 "random", true,
-                "setattr", "krbPasswordExpiration=" + PASSWORD_EXPIRATION_DATETIME
+                "setattr", "krbPasswordExpiration=" + MAX_PASSWORD_EXPIRATION_DATETIME
```
giladwolff: again, for my edification, why are we setting the password expiration when we add the user?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -159,11 +165,34 @@ public void userSetPassword(String user, String password) throws FreeIpaClientEx
         // FreeIPA expires any password that is set by another user. Work around this by
         // performing a separate API call to set the password expiration into the future
         userMod(user, "userpassword", password);
-        updateUserPasswordExpiration(user);
+        updateUserPasswordMaxExpiration(user);
+    }
+
+    public void updateUserPasswordExpiration(String user, long expiration) throws FreeIpaClientException {
+        String passwordExpirationDate = formatDate(expiration);
+        userMod(user, "setattr", "krbPasswordExpiration=" + passwordExpirationDate);
```
giladwolff: So above you set the krbPasswordExpiration with a date time string (with a certain pattern). Here it's just a long. Are both correct? If they are, let's settle on one way to identity the expiration time. Personally I like to use java Instant and Duration to hold and calculate time - it is more robust than passing along longs.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -159,11 +165,34 @@ public void userSetPassword(String user, String password) throws FreeIpaClientEx
...
+    }
+
+    public void updateUserPasswordMaxExpiration(String user) throws FreeIpaClientException {
+        updateUserPasswordExpiration(user, 0);
```
giladwolff: same question. Here you pass 0, does that mean that the password never expires? Also, maybe have a readability constant for this?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -159,11 +165,34 @@ public void userSetPassword(String user, String password) throws FreeIpaClientEx
...
+        Map<String, Object> params =
+                Map.of("setattr", List.of(
+                        "cdpHashedPassword=" + hashedPassword,
+                        "cdpUnencryptedKrbPrincipalKey=" + unencryptedKrbPrincipalKey,
+                        "krbPasswordExpiration=" + passwordExpirationDate));
+        return userMod(user, params);
+    }
+
+    String formatDate(long timestampMillis) {
+        if (timestampMillis <= 0) {
```
giladwolff: I think it's safer if you assert that timestampMillis < 0. It's a bug.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -80,11 +80,11 @@ public void postInstallFreeIpa(Long stackId) throws Exception {
 
     private void modifyAdminPasswordExpirationIfNeeded(FreeIpaClient client) throws FreeIpaClientException {
         Optional<User> user = client.userFind(freeIpaClientFactory.getAdminUser());
-        if (user.isPresent() && !FreeIpaClient.PASSWORD_EXPIRATION_DATETIME.equals(user.get().getKrbPasswordExpiration())) {
+        if (user.isPresent() && !FreeIpaClient.MAX_PASSWORD_EXPIRATION_DATETIME.equals(user.get().getKrbPasswordExpiration())) {
```
giladwolff: So the admin password is our own freeipa admin user? and this is to ensure that the password never expires?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -120,7 +122,8 @@ private WorkloadCredential getCredentials(String userCrn, Optional<String> reque
         GetActorWorkloadCredentialsResponse response = grpcUmsClient.getActorWorkloadCredentials(IAM_INTERNAL_ACTOR_CRN, userCrn, requestId);
         String hashedPassword = response.getPasswordHash();
         List<ActorKerberosKey> keys = response.getKerberosKeysList();
-        return new WorkloadCredential(hashedPassword, keys);
+        long expirationDate = response.getPasswordHashExpirationDate();
```
giladwolff: Suggestion: convert it to an Instant here.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
+package com.sequenceiq.freeipa.service.freeipa.user.model;
```
giladwolff: Feel free to ignore if this is not the coding convention in this repo, but copywrong statement is missing.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class UmsUsersState {
```
giladwolff: javadoc?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final UsersState usersState;
+
+    private final Map<String, WorkloadCredential> usersWorkloadCredentialMap;
```
giladwolff: Can this be immutable?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final Set<FmsUser> requestedWorkloadUsers;
```
giladwolff: Can this be immutable?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public UmsUsersState(UsersState usersState, Map<String, WorkloadCredential> usersWorkloadCredentialMap, Set<FmsUser> requestedWorkloadUsers) {
+        this.usersState = requireNonNull(usersState, "UsersState is null");
+        this.usersWorkloadCredentialMap = requireNonNull(usersWorkloadCredentialMap, "workload credential map is null");
+        this.requestedWorkloadUsers = requireNonNull(requestedWorkloadUsers, "requested workload users is null");
```
giladwolff: I would make a copy of the map and set, even if you don't make the members immutable.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
```diff
@@ -0,0 +1,59 @@
...
+            this.usersState = usersState;
+        }
+
+        public void addWorkloadCredentials(String userName, WorkloadCredential creds) {
+            workloadCredentialMap.put(userName, creds);
+        }
+
+        public void addRequestedWorkloadUsers(FmsUser user) {
+            requestedWorkloadUsers.add(user);
+        }
```
lacikaaa: builders' methods usually returns with `this` so you can chain methods

config/checkstyle/checkstyle.xml
handavid: these changes allow me to use the @SuppressWarnings annotation on one method that has 8 parameters. `UmsUsersStateProvider.handleUser()`

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: We set a random password so that we can generate tgts. I think that we had to set this so that the random password isn't expired.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: all expirations are currently passed through the method parameters as longs. They are converted to the map {"setattr":"krbPasswordExpiration=1234567890Z"} when constructing the ipa JSON-RPC request. The userMod() method doesn't know anything about password expiration.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: I'll add a constant

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: will do

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
handavid: yes. otherwise the FMS won't be able to sync users or set passwords after some time in the future.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: converted it to an Optional<Instant> so we can do the 0-checking here where the 0 is meaningful.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UmsUsersState.java
handavid: done

handavid:
> super nit: you can declare them as ImmutableMap/ImmutableSet.

I'll add that to my next PR

PR #6935, opened 2019-12-12
CB-1672 - restricting ssh access to hosts running idbroker roles

attilapalfi92:


orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
```diff
@@ -9,6 +11,8 @@
 
     private final String instanceType;
 
+    private final CloudIdentityType cloudIdentityType;
```
lacikaaa: please check if we could solve this another way without adding this field to the node

PR #6936, opened 2019-12-12
CB-4815 Instance profiles or Roles could be filtered by API via query…

topolyai5:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -657,23 +660,14 @@ public CloudAccessConfigs accessConfigs(CloudCredential cloudCredential, Region
...
-                    properties.put("roleArn", Strings.isNullOrEmpty(roleName) ? instanceProfile.getArn() : roleName);
-                }
-                cloudAccessConfigs.getCloudAccessConfigs().add(
-                        new CloudAccessConfig(
-                                instanceProfile.getInstanceProfileName(),
-                                instanceProfile.getInstanceProfileId(),
-                                properties));
+            String accessConfigType = filters.get(ACCESS_CONFIG_TYPE);
+            Set<CloudAccessConfig> cloudAccessConfigSet;
+            if ("ROLE".equals(accessConfigType)) {
```
doktoric: Can we use an Enum for this ?

doktoric: also can we add some logs here ?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/platformresource/PlatformResourceEndpoint.java
```diff
@@ -150,7 +151,8 @@ PlatformAccessConfigsResponse getAccessConfigs(
             @QueryParam("credentialCrn") String credentialCrn,
             @QueryParam("region") String region,
             @QueryParam("platformVariant") String platformVariant,
-            @QueryParam("availabilityZone") String availabilityZone);
+            @QueryParam("availabilityZone") String availabilityZone,
+            @QueryParam("accessConfigType") AccessConfigTypeQueryParam accessConfigType);
```
doktoric: can we set default here ?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -657,23 +660,14 @@ public CloudAccessConfigs accessConfigs(CloudCredential cloudCredential, Region
...
+                cloudAccessConfigSet = getAccessConfigByRole(client);
+            } else {
+                cloudAccessConfigSet = getAccessConfigByInstanceProfile(client);
             }
+            cloudAccessConfigs.getCloudAccessConfigs().addAll(cloudAccessConfigSet);
         } catch (AmazonServiceException ase) {
             if (ase.getStatusCode() == UNAUTHORIZED) {
                 String policyMessage = "Could not get instance profile roles because the user does not have enough permission.";
```
doktoric: I think the message here is not correct

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
keyki: We should also add a log if the type is not ROLE or null then we will return all instance profiles

PR #6939, opened 2019-12-12
CB-4812 Sdx on azure should now create externaldb by default

gergopapi2:
Prerequisite: CM 7.1.0 be the default CM to be installed, as only CM 7.1.0 supports external RDS on azure.
Do not merge until then.


gergopapi2:
The refactor part extracted to https://github.com/hortonworks/cloudbreak/pull/6958.

PR #6944, opened 2019-12-12
CDPCP-1212. Password expiration set in set password api workflow

handavid:
The password expiration time is not included as part of the set password
request. The PasswordService retrieves the password policy from the UMS and
calculates the expiration Instant. The SetPasswordHandler calls the FreeIpaClient
with the expiration Instant when setting the password.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
@@ -1,5 +1,8 @@
 package com.sequenceiq.freeipa.flow.freeipa.user.event;
 
+import java.time.Instant;
+import java.util.Optional;
+
 public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> {
 
     private final String environment;
@@ -8,11 +11,14 @@ public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> 
 
     private final String password;
 
-    public SetPasswordRequest(Long stackId, String environment, String username, String password) {
+    private final Optional<Instant> expirationInstant;
+
+    public SetPasswordRequest(Long stackId, String environment, String username, String password, Optional<Instant> expirationInstant) {
         super(stackId);
         this.environment = environment;
         this.username = username;
         this.password = password;
+        this.expirationInstant = expirationInstant;
     }
 
     public String getEnvironment() {
@@ -27,12 +33,17 @@ public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> 
         return password;
     }
 
+    public Optional<Instant> getExpirationInstant() {
+        return expirationInstant;
+    }
+
     @Override
     public String toString() {
         return "SetPasswordRequest{"
                 + "stackId='" + getResourceId() + '\''
                 + "environment='" + environment + '\''
                 + "username='" + username + '\''
+                + "expirationInstant=" + expirationInstant
                 + '}';
     }
 }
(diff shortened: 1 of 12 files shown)


mdzz:
This is needed to complete Dec 16 deliverable CDPCP-1212 & CDPCP-1041.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
```diff
@@ -27,12 +33,17 @@ public String getPassword() {
         return password;
     }
 
+    public Optional<Instant> getExpirationInstant() {
```
giladwolff: So if I understand it correctly, you allow people to add the expiration date on the set password call directly? I think we shouldn't do that. It means that anybody can "override" the policy. Let's remove it, if my understanding is correct.

handavid: no. this is part of the internal workflow. this `SetPasswordRequest` is put on the event bus for a handler to pick up. there have been no api changes.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -121,6 +130,25 @@ private void asyncSetPasswords(Optional<String> requestId, String operationId,
...
 
+    @VisibleForTesting
+    Optional<Instant> calculateExpirationTime(String actorCrn, String accountId) {
+        LOGGER.debug("calculating expiration time for password in account {}", accountId);
+        UserManagementProto.Account account = umsClient.getAccountDetails(actorCrn, accountId, MDCUtils.getRequestId());
+        if (account.hasPasswordPolicy()) {
+            long maxLifetime = account.getPasswordPolicy().getWorkloadPasswordMaxLifetime();
+            if (maxLifetime != 0L) {
+                LOGGER.debug("Calculating password expiration by adding lifetime '{}' to current time", maxLifetime);
+                return Optional.of(Instant.ofEpochMilli(currentTimeSupplier.get() + maxLifetime));
```
topolyai5: What is the reason for the supplier? you can use the _Clock_ class if you want to mock the current millis.

handavid: yes it was just for testing.

handavid: yes it was just for testing.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -121,6 +130,25 @@ private void asyncSetPasswords(Optional<String> requestId, String operationId,
         }
     }
 
+    @VisibleForTesting
+    Optional<Instant> calculateExpirationTime(String actorCrn, String accountId) {
+        LOGGER.debug("calculating expiration time for password in account {}", accountId);
+        UserManagementProto.Account account = umsClient.getAccountDetails(actorCrn, accountId, MDCUtils.getRequestId());
```
handavid: i had one nagging thought about whether we should use the internal actor here rather than the caller. i don't know whether the getAccount call would need rights that users may not have.

giladwolff: So there is no authorization for this in the backend, but I think that calling this using the internal actor is better as it is future proofing this in case we change the authorization on the get-account call in the future. That said, feel free to do it in a follow-up commit.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
aarman-cloudera: changed to Clock

PR #6964, opened 2019-12-16
CB-4873 Fix subnet choosing logic for sdx and distrox

lacikaaa:


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
```diff
@@ -106,7 +107,8 @@ public PlatformRecommendation createForBlueprint(Long workspaceId, String bluepr
         }
         VmRecommendations recommendations = cloudParameterService.getRecommendation(cloudPlatform);
 
-        Set<VmType> availableVmTypes = vmTypes.getCloudVmResponses().get(availabilityZone);
+        Set<VmType> availableVmTypes = StringUtils.isNotBlank(availabilityZone) ? vmTypes.getCloudVmResponses().get(availabilityZone)
+                : vmTypes.getCloudVmResponses().values().iterator().next();
```
akanto: Instead of null you might get a NoSuchElementException from iterator().next() method call

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
```diff
@@ -189,7 +191,8 @@ private boolean isThereMasterComponents(ClusterManagerType clusterManagerType, S
     }
 
     private VmType getDefaultVmType(String availabilityZone, CloudVmTypes vmtypes) {
-        VmType vmType = vmtypes.getDefaultCloudVmResponses().get(availabilityZone);
+        VmType vmType = StringUtils.isNotBlank(availabilityZone) ? vmtypes.getDefaultCloudVmResponses().get(availabilityZone)
+                : vmtypes.getDefaultCloudVmResponses().values().iterator().next();
```
akanto: Instead of null you might get a NoSuchElementException from iterator().next() method call

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -168,7 +168,7 @@ private void validateSubnetIds(NetworkV4Request network, DetailedEnvironmentResp
     }
 
     private void validateSubnet(NetworkV4Request network, DetailedEnvironmentResponse environment, String subnetId) {
-        if (!environment.getNetwork().getSubnetIds().contains(subnetId)) {
+        if (subnetId != null && !environment.getNetwork().getSubnetIds().contains(subnetId)) {
```
akanto: Minor one: I know that this is old code, but since there are 3 dots in environment.getNetwork().getSubnetIds().contains(...), this means that we can get NPE from 3 places. I prefer a bit more defensive programming approach

PR #6966, opened 2019-12-17
CB-4773 Freeipa fails silently if sending in request without subnet info

gergopapi2:


environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
```diff
@@ -32,39 +34,80 @@ public NetworkCreationValidator(NetworkService networkService) {
...
-            if (subnetMetas.size() != network.getSubnetIds().size()) {
-                message = String.format("Subnets of the environment (%s) are not found in the vpc (%s). ",
-                        String.join(",", getSubnetDiff(network.getSubnetIds(), subnetMetas.keySet())),
-                        networkService.getAwsVpcId(network).orElse(""));
+        }
+        validateNetworkIdAndCidr(environment, network, resultBuilder);
+        return resultBuilder;
+    }
+
+    private void validateAwsExistingNetwork(NetworkDto network, Map<String, CloudSubnet> subnetMetas, ValidationResultBuilder resultBuilder) {
```
attilapalfi92: This AWS specific validation should be in AwsEnvironmentNetworkValidator

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
```diff
@@ -32,39 +34,80 @@ public NetworkCreationValidator(NetworkService networkService) {
...
+        Map<String, Long> zones = subnetMetas.values().stream()
+                .collect(Collectors.groupingBy(CloudSubnet::getAvailabilityZone, Collectors.counting()));
+        if (zones.size() < 2) {
+            message = "Cannot create environment, the subnets in the vpc should be present at least in two different availability zones";
+            LOGGER.info(message);
+            resultBuilder.error(message);
+        }
+    }
+
+    private void validateAzureExistingNetwork(NetworkDto network, ValidationResultBuilder resultBuilder) {
```
attilapalfi92: This Azure specific validation should be in AzureEnvironmentNetworkValidator

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
```diff
@@ -32,39 +34,80 @@ public NetworkCreationValidator(NetworkService networkService) {
...
     public ValidationResultBuilder validateNetworkCreation(Environment environment, NetworkDto network, Map<String, CloudSubnet> subnetMetas) {
         ValidationResultBuilder resultBuilder = new ValidationResultBuilder();
-        if (AWS.name().equalsIgnoreCase(environment.getCloudPlatform()) && network != null && network.getNetworkCidr() == null) {
-            String message;
-            if (network.getSubnetIds().size() < 2) {
-                message = "Cannot create environment, there should be at least two subnets in the network";
-                LOGGER.info(message);
-                resultBuilder.error(message);
+        if (network != null && network.getNetworkCidr() == null) {
+            if (AWS.name().equalsIgnoreCase(environment.getCloudPlatform())) {
```
attilapalfi92: Instead of ifing here, EnvironmentNetworkProviderValidator should decide which provider specific validator should be used

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
gergopapi2: That validator currently is called from the flow only. The task was to validate during request time.
I can of course use this validator from the EnvironmentCreationService as well, if this is what you meant.

attilapalfi92: Yes, that's what I meant

PR #6971, opened 2019-12-17
CB-4778 handle every flow exception, termination retry allowed after 60 mins

sodre90:


flow/src/main/java/com/sequenceiq/flow/service/flowlog/FlowLogDBService.java
```diff
@@ -124,13 +124,29 @@ public void updateLastFlowLogStatus(FlowLog lastFlowLog, boolean failureEvent) {
         flowLogRepository.updateLastLogStatusInFlow(lastFlowLog.getId(), stateStatus);
     }
 
+    public void cancelTooOldTerminationFlowForResource(Long resourceId, long olderThan) {
```
akanto: This method looks quite complex with, filters, map, anymatch, findfirst etc. This definitely requires a junit test.

flow/src/main/java/com/sequenceiq/flow/reactor/config/EventBusConfig.java
```diff
@@ -49,6 +64,22 @@ public EventBus reactor(MDCCleanerThreadPoolExecutor threadPoolExecutor, Environ
...
+                                LOGGER.error("Unhandled exception happened in flow {}, lets cancel it", flowId, throwable);
+                                flowLogDBService.getLastFlowLog(flowId).ifPresent(flowLog -> {
+                                    flowLogDBService.updateLastFlowLogStatus(flowLog, true);
+                                    applicationFlowInformation.handleFlowFail(flowLog);
+                                });
+                            }
+                        }
+                    }
+                    LOGGER.error("Exception happened in dispatcher", throwable);
+                })
```
lacikaaa: could you move it into a method, so it would be a little bit more readable

flow/src/main/java/com/sequenceiq/flow/reactor/config/EventBusConfig.java
```diff
@@ -49,6 +64,22 @@ public EventBus reactor(MDCCleanerThreadPoolExecutor threadPoolExecutor, Environ
                 .env(env)
                 .dispatcher(new ThreadPoolExecutorDispatcher(eventBusThreadPoolBacklogSize, eventBusThreadPoolCoreSize, threadPoolExecutor))
                 .traceEventPath()
+                .dispatchErrorHandler(throwable -> {
+                    if (throwable.getCause() instanceof Exceptions.ValueCause) {
+                        if (((Exceptions.ValueCause) throwable.getCause()).getValue() instanceof Event) {
+                            Event event = (Event) ((Exceptions.ValueCause) throwable.getCause()).getValue();
+                            if (event.getHeaders() != null && event.getHeaders().get("FLOW_ID") != null) {
+                                String flowId = event.getHeaders().get("FLOW_ID").toString();
+                                LOGGER.error("Unhandled exception happened in flow {}, lets cancel it", flowId, throwable);
```
lacikaaa: would it make sense to add all the headers to the log line?

flow/src/main/java/com/sequenceiq/flow/reactor/config/EventBusConfig.java
lacikaaa: also there are some call here, maybe we should wrap it with try-catch if something goes wrong during error handling, shouldn't we?

flow/src/main/java/com/sequenceiq/flow/service/flowlog/FlowLogDBService.java
sodre90: Makes sense, I created 2 unit test for this code

flow/src/main/java/com/sequenceiq/flow/reactor/config/EventBusConfig.java
sodre90: Makes sense, I added some log line which can be helpful.

flow/src/main/java/com/sequenceiq/flow/reactor/config/EventBusConfig.java
sodre90: Moved to a method and put into try-catch, thanks for the suggestion, makes sense

PR #6972, opened 2019-12-17
CB-4904 test report improvements

afarsang:
test report improvements
Closes 4904


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/log/Log.java
```diff
@@ -4,28 +4,30 @@
...
+    public static final String TEST_CONTEXT_REPORTER = "testContextReporter";
+
+    public static final String ALTERNATE_LOG = "alternateLog";
 
-public class Log {
     private Log() {
     }
 
-    public static void logJSON(Logger logger, String message, Object jsonObject) throws IOException {
+    public static void logWhenJson(Logger logger, String message, Object jsonObject) throws IOException {
```
topolyai5: I think whenJson is enough like the others methods (when, then..)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/log/Log.java
```diff
@@ -40,7 +42,67 @@ public static void log(Logger logger, String message, Object... args) {
...
+        return reporter;
+    }
+
+    private static void publishReport(TestContextReporter testContextReporter) {
+        ITestResult res = Reporter.getCurrentTestResult();
+        if (res == null) {
+            return;
+        }
+        res.setAttribute(ALTERNATE_LOG, testContextReporter.print());
+    }
```
topolyai5: You can use negation and add to the if block the operation

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/log/TestContextReporter.java
```diff
@@ -0,0 +1,80 @@
...
+    public void addStep(String step, String description) {
+        steps.add(new Step(step, description));
+    }
+
+    public void addStep(String step, String description, String json) {
+        steps.add(new Step(step, description, json));
+    }
+
+    public String print() {
+        if (steps.size() == 0) {
```
topolyai5: Please use .isEmpty()

afarsang: Ok

PR #6976, opened 2019-12-17
CB-4839 - FreeIPA: Configure all services for auto restart

lacikaaa:


freeipa/src/main/resources/freeipa-salt/pillar/freeipa/services.sls
```diff
@@ -0,0 +1,9 @@
+freeipa:
+  services:
```
jamisonbennett: I was thinking about this a little bit more and I was thinking we might want to also have nginx and sssd.

lacikaaa: added

PR #6977, opened 2019-12-17
CB-4908 Ensure strict HTTPS whitelist if TLS enabled

risdenk:
If TLS is enabled, then whitelist should be strict HTTPS only.


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/knox/KnoxGatewayConfigProvider.java
```diff
@@ -90,7 +90,11 @@
                     config.add(config(SIGNING_KEY_ALIAS, SIGNING_IDENTITY));
                     if (kerberosConfig.isPresent()) {
                         String domain = kerberosConfig.get().getDomain();
-                        config.add(config(GATEWAY_WHITELIST, "^/.*$;^https?://(.+." + domain + "):[0-9]+/?.*$"));
+                        if (generalClusterConfigs.getAutoTlsEnabled()) {
```
akanto: This looks good, although I think we could refactor the getRoleConfigs method and add a few more junit tests

risdenk: Refactored and added tests.

PR #6985, opened 2019-12-18
CB-1672 - restricting ssh access to hosts running idbroker roles

attilapalfi92:


core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
```diff
@@ -164,6 +171,22 @@ public void setAttributes(Json attributes) {
         this.attributes = attributes;
     }
 
+    public Optional<CloudIdentityType> getCloudIdentityType() {
+        if (attributes != null && StringUtils.isNotEmpty(attributes.getValue())) {
+            Map<String, Object> attributeMap = attributes.getMap();
+            if (attributeMap.containsKey(IDENTITY_TYPE_ATTRIBUTE)) {
+                return Optional.of(CloudIdentityType.valueOf(attributeMap.get(IDENTITY_TYPE_ATTRIBUTE).toString()));
```
lacikaaa: this is not nullpointer proof, which should be for backward compatibility

PR #6989, opened 2019-12-19
CB-4942 Knox - Increase connection and read timeout values for Impala

twmarshall:
Increase the connection/read timeout from 20s to 5m for Impala Hiveserver2


akanto:
Since we have 5mins already on nginx side, this looks good to me

```
    location ~ .*/ {
        proxy_pass         https://knox;
        proxy_connect_timeout       300;
        proxy_send_timeout          300;
        proxy_read_timeout          300;
        send_timeout                300;
        proxy_redirect     off;
```

PR #6996, opened 2019-12-31
CB-4446 fixing schemaregistry version

lnardai:
- will do some base image test locally


lnardai:
SCHEMAREGISTRY still fails in local

PR #7015, opened 2020-01-06
CB-3762 Error when passing an extra '/' at end of cloud storage path

topolyai5:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -418,4 +420,11 @@ private String getAccountIdFromCrn(String userCrn) {
     private DetailedEnvironmentResponse getEnvironment(SdxClusterRequest sdxClusterRequest) {
         return environmentClientService.getByName(sdxClusterRequest.getEnvironment());
     }
+
+    private String trimBaseLocation(String baseLocation) {
+        if (baseLocation.endsWith("/")) {
```
gergopapi2: method content could be changed to Stringutils.substringBeforeLast().

core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
```diff
@@ -438,4 +438,10 @@ private String createCRN(String accountId) {
                 .toString();
     }
 
+    private String trimBaseLocation(String baseLocation) {
```
gergopapi2: method content could be changed to Stringutils.substringBeforeLast().

gergopapi2: Sorry, that's bullshit I wrote :D

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
gergopapi2: Again, bullshit! :D

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -418,4 +420,11 @@ private String getAccountIdFromCrn(String userCrn) {
     private DetailedEnvironmentResponse getEnvironment(SdxClusterRequest sdxClusterRequest) {
         return environmentClientService.getByName(sdxClusterRequest.getEnvironment());
     }
+
+    private String trimBaseLocation(String baseLocation) {
```
gergopapi2: StringUtils.stripEnd() could be used.

core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
gergopapi2: StringUtils.stripEnd() could be used.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
topolyai5: The following base locations will produces wrong results: 
* s3a://some/dir ->s3a://some
* s3a://some -> s3a:/

This function only trims the last / if exist otherwise return the unmodified string.

topolyai5:
> Maybe it is worth to create one `trimBaseLocation` method (somewhere, in an util class, or sthing like that) and using that everywhere.
> Also please add unit test for this logic to avoid any other issue in the future.

@gergopapi2 pointed to a better solution, so I will change this function to that function

PR #7025, opened 2020-01-08
Wrong status in log

topolyai5:


doktoric:
Please modify the log message

topolyai5:
> Please modify the log message

Ok, thanks your feedback. It is very valuable for me

PR #7027, opened 2020-01-08
CB-4927 add iam:ListRoles to cross account role

topolyai5:


dbialek2:
@topolyai5 Will this change be merged to master eventually? I am linking to this IAM policy definition in master from docs.

PR #7029, opened 2020-01-08
CB-4748 In case of environments with CCM, the 9443 port should not be opened for control plane

lacikaaa:


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/GatewaySecurityGroupWithDefaultGatewayCidrsDecorator.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.sequenceiq.cloudbreak.domain.SecurityGroup;
+import com.sequenceiq.cloudbreak.domain.SecurityRule;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
+import com.sequenceiq.common.api.type.InstanceGroupType;
+import com.sequenceiq.common.api.type.Tunnel;
+
+@Component
+public class GatewaySecurityGroupWithDefaultGatewayCidrsDecorator {
```
keyki: Can we use a simpler name? Like GatewaySecurityGroupDecorator?

lacikaaa: renamed

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/GatewaySecurityGroupDecorator.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Value("${cb.https.port}")
+    private Integer httpsPort;
+
+    public void extendGatewaySecurityGroupWithDefaultGatewayCidrs(Stack stack, Tunnel tunnel) {
+        Set<InstanceGroup> gateways = filterForGateways(stack);
+        Set<String> defaultGatewayCidrs = filterEmptyCidrs();
+        LOGGER.debug("Extending security group for gateways: [{}] DefaultGatewayCidrs: [{}] Tunnel: [{}]",
+                gateways.stream().map(InstanceGroup::getGroupName).collect(Collectors.joining(", ")), defaultGatewayCidrs, tunnel);
+        if (!defaultGatewayCidrs.isEmpty() && !tunnel.useCcm()) {
```
keyki: I think we should use stack.getUseCcm it handles if the tunnel is not set.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/stack/CreateFreeIpaRequestToStackConverter.java
```diff
@@ -121,7 +121,7 @@ private void extendGatewaySecurityGroupWithDefaultGatewayCidrs(Stack stack) {
         Set<InstanceGroup> gateways =
                 stack.getInstanceGroups().stream().filter(ig -> InstanceGroupType.MASTER == ig.getInstanceGroupType()).collect(Collectors.toSet());
         Set<String> defaultGatewayCidrs = defaultGatewayCidr.stream().filter(StringUtils::isNotBlank).collect(Collectors.toSet());
-        if (!defaultGatewayCidrs.isEmpty()) {
+        if (!defaultGatewayCidrs.isEmpty() && !stack.getTunnel().useCcm()) {
```
keyki: Same stack.getUseCcm()

PR #7034, opened 2020-01-09
DISTX-367-Introduce optional gateway HostGroup in DE Templates

smaniraju:
Introduce optional Gateway HostGroup in DE Templates.
Default Gateway Node Count will be 0.

Closes #DISTX-367


smaniraju:
Followed up in https://github.com/hortonworks/cloudbreak/pull/7046 due to failing integration tests.

PR #7064, opened 2020-01-13
CDPCP-1113. User sync checks rights for access to environment

handavid:
This commit changes user sync to use rights instead of roles for
determining access to the environment. A hasRights method was
added to the GrpcUmsClient and AuthorizationClient to support this
use case since hasRights is a better match than checkRight.

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/AuthorizationClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/AuthorizationClient.java
@@ -48,6 +48,24 @@ public class AuthorizationClient {
         );
     }
 
+    public boolean hasRight(String requestId, String userCrn, String right, String resource) {
+        checkNotNull(requestId);
+        checkNotNull(userCrn);
+        checkNotNull(right);
+
+        AuthorizationProto.HasRightsRequest.Builder requestBuilder = AuthorizationProto.HasRightsRequest.newBuilder()
+                .setActorCrn(userCrn);
+
+        AuthorizationProto.RightCheck.Builder rightCheckBuilder = AuthorizationProto.RightCheck.newBuilder().setRight(right);
+        if (!StringUtils.isEmpty(resource)) {
+            rightCheckBuilder.setResource(resource);
+        }
+        requestBuilder.addCheck(rightCheckBuilder.build());
+
+        AuthorizationProto.HasRightsResponse response = newStub(requestId).hasRights(requestBuilder.build());
+        return response.getResultList().get(0);
+    }
+
     /**
      * Creates a new stub with the appropriate metadata injecting interceptors.
      *
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -300,14 +300,21 @@ public class GrpcUmsClient {
         }
     }
 
-    @Cacheable(cacheNames = "umsUserRightsCache", key = "{ #actorCrn, #userCrn, #right, #resource }")
-    public boolean checkRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
+    public boolean precheckRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
         if (InternalCrnBuilder.isInternalCrn(actorCrn)) {
             LOGGER.info("InternalCrn, allow right {} for user {}!", right, userCrn);
             return true;
         }
         if (isReadRight(right)) {
             LOGGER.info("Letting read operation through for right {} for user {}!", right, userCrn);
+            return true;
+        }
+        return false;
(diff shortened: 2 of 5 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/AuthorizationClient.java
```diff
@@ -48,6 +48,24 @@ public void checkRight(String requestId, String userCrn, String right, String re
         );
     }
 
+    public boolean hasRight(String requestId, String userCrn, String right, String resource) {
```
giladwolff: So the interface is hasRights, right? maybe mirror the actual interface?

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/AuthorizationClient.java
```diff
@@ -48,6 +48,24 @@ public void checkRight(String requestId, String userCrn, String right, String re
...
+                .setActorCrn(userCrn);
+
+        AuthorizationProto.RightCheck.Builder rightCheckBuilder = AuthorizationProto.RightCheck.newBuilder().setRight(right);
+        if (!StringUtils.isEmpty(resource)) {
+            rightCheckBuilder.setResource(resource);
+        }
+        requestBuilder.addCheck(rightCheckBuilder.build());
+
+        AuthorizationProto.HasRightsResponse response = newStub(requestId).hasRights(requestBuilder.build());
+        return response.getResultList().get(0);
```
giladwolff: So while this is fine it means that we are doing some extra work. Because I believe you are also bringing the actor's rights with the call to GetRights somewhere (you need it for the group membership and workload-administration-group membership information). Moreover, there is a potential for discrepancy between the two calls - the actor rights may have change in between. So I think it is better to use the actor rights you get from the GetRights response to check for the rights the actor has. Let's talk about this offline I am not sure how much of the scrutinizer was ported to CB code base.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -310,6 +309,14 @@ public boolean checkRight(String actorCrn, String userCrn, String right, String
             LOGGER.info("Letting read operation through for right {} for user {}!", right, userCrn);
             return true;
         }
+        return false;
+    }
+
+    @Cacheable(cacheNames = "umsUserRightsCache", key = "{ #actorCrn, #userCrn, #right, #resource }")
```
giladwolff: let's not cache anything on the client. It complicates reasoning about the state of the world and if caching is needed is should be done in the UMS. The actual network hoop is not that expensive I don't think.

handavid: This is cached in the existing code and I didn't want to change it. I did dig through the cache initialization and it is set with an expiration of 1s. We can open a JIRA to remove the cache separately from this JIRA.

horadla23: We definitely need caching, we had serious performance issues if we do not used caching, it has small expiration time so I do not see what could be the problem with that.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -300,8 +300,7 @@ public Account getAccountDetails(String actorCrn, String accountId, Optional<Str
         }
     }
 
-    @Cacheable(cacheNames = "umsUserRightsCache", key = "{ #actorCrn, #userCrn, #right, #resource }")
-    public boolean checkRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
+    public boolean precheckRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
```
horadla23: this method has no call to UMS so here we do not need caching

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -327,6 +334,19 @@ public boolean checkRight(String actorCrn, String userCrn, String right, Optiona
         return checkRight(actorCrn, userCrn, right, null, requestId);
     }
 
+    public boolean hasRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
```
horadla23: But this method has logic which calls UMS, so we need cache here

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/AuthorizationClient.java
horadla23: I agree with Gilad, it is worth to use list here in the same way as interface does

handavid:
discarding this pr for now

PR #7066, opened 2020-01-14
CB-5062 Azure AppId required permissions on the UI

foldik:
Add azure role definition template.


foldik:
@gergopapi2 will add additional items to support external databases on Azure.

PR #7068, opened 2020-01-14
CB-5068 Fix index created on wrong environment_network instead of environment_parameters

bergerdenes:


bergerdenes:
because on environments above mow-dev that field might be null and a unique index creation would fail.

PR #7088, opened 2020-01-16
CB-5117 Update existing security groups of environments

topolyai5:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -483,6 +484,18 @@ public CloudSecurityGroups securityGroups(CloudCredential cloudCredential, Regio
         return new CloudSecurityGroups(result);
     }
 
+    private List<SecurityGroup> fetchSecurityGroups(AmazonEC2Client ec2Client, DescribeSecurityGroupsRequest describeSecurityGroupsRequest) {
+        try {
+            return ec2Client.describeSecurityGroups(describeSecurityGroupsRequest).getSecurityGroups();
+        } catch (AmazonEC2Exception e) {
+            if (e.getStatusCode() == HttpStatus.BAD_REQUEST.value() || e.getStatusCode() == HttpStatus.NOT_FOUND.value()) {
+                throw new IllegalArgumentException(e.getErrorMessage(), e);
```
gregito: I think this should be IllegalStateException or some other one but definitely not IllegalArgumentException since we don't know whether this method has got some illegal input argument or just something happened.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzurePlatformResources.java
```diff
@@ -121,15 +122,17 @@ public CloudSshKeys sshKeys(CloudCredential cloudCredential, Region region, Map<
...
-                properties.put("resourceGroupName", securityGroup.resourceGroupName());
-                properties.put("networkInterfaceIds", securityGroup.networkInterfaceIds());
-                CloudSecurityGroup cloudSecurityGroup = new CloudSecurityGroup(securityGroup.name(), securityGroup.id(), properties);
-                result.computeIfAbsent(actualRegion, s -> new HashSet<>()).add(cloudSecurityGroup);
+        PlatformResourceSecurityGroupFilterView filter = new PlatformResourceSecurityGroupFilterView(filters);
+        String groupId = filter.getGroupId();
+        if (groupId != null) {
+            NetworkSecurityGroup networkSecurityGroup = client.getSecurityGroups().getById(groupId);
+            if (networkSecurityGroup == null) {
+                throw new IllegalArgumentException("Nothing found on Azure with id: " + groupId);
```
gregito: same here. we don't know why the networkSecurityGroup is null, hence this is a state where we don't know what to do. -> IllegalStateException

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
```diff
@@ -207,9 +208,25 @@ public CreateAWSEnvironmentRequest getCreateAWSEnvironmentForCli(EnvironmentDto
...
+            environment.setDefaultSecurityGroupId(securityAccess.getDefaultSecurityGroupId());
+        }
+        if (StringUtils.isNotBlank(securityAccess.getSecurityGroupIdForKnox())) {
+            environment.setSecurityGroupIdForKnox(securityAccess.getSecurityGroupIdForKnox());
+        }
+    }
+
     void setSecurityAccess(Environment environment, SecurityAccessDto securityAccess) {
         if (securityAccess != null) {
-            environment.setCidr(securityAccess.getCidr());
```
bergerdenes: this is not good as EnvironmentCreationService is also using this method and CIDR setting is impossible with this change

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
```diff
@@ -207,9 +208,25 @@ public CreateAWSEnvironmentRequest getCreateAWSEnvironmentForCli(EnvironmentDto
         return environmentRepository.findByResourceCrnAndAccountIdAndArchivedIsFalse(resourceCrn, accountId);
     }
 
+    /**
+     * The CIDR could not be edited so the first step we clear the the existed CIDR every time.
```
bergerdenes: is it a hard requirement?

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
```diff
@@ -207,9 +208,25 @@ public CreateAWSEnvironmentRequest getCreateAWSEnvironmentForCli(EnvironmentDto
...
+     * We are assuming the security access has the proper security group ids and we don't check it again.
```
bergerdenes: why?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
topolyai5: It is just an `internal` exception. I'm fine with a dedicated exception, but the CloudConnectorEception is not the best solution. @bergerdenes if you search in the pr, I handle the exception in `GetPlatformSecurityGroupsHandler`. I need to mark as a Permanent exception and I need to exit from the retry. I think the CloudConnector is a retryable exception in this code

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/GetPlatformSecurityGroupsHandler.java
```diff
@@ -44,6 +45,8 @@ public void accept(Event<GetPlatformSecurityGroupsRequest> getPlatformSecurityGr
             GetPlatformSecurityGroupsResult getPlatformSecurityGroupsResult = new GetPlatformSecurityGroupsResult(request.getResourceId(), securityGroups);
             request.getResult().onNext(getPlatformSecurityGroupsResult);
             LOGGER.debug("Query platform networks types finished.");
+        } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
topolyai5: I need to mark as a Permanent exception and I need to exit from the retry. I think the CloudConnector is a retryable exception in this code

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
topolyai5: what do you mean `hard`? Yes, if we edit a security group we always clear the cidr because we replace by a security group id

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
topolyai5: Because I check the security access one step before it. I don't want to check here again, so if anyone wants to use this function, calculate with that, I clear the cidr and set anything that they set to SecurityAccess without validation.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
topolyai5: hm, I revert this

PR #7106, opened 2020-01-20
CB-5189 disable SDX upgrade tests on AWS

biharitomi:


integration-test/src/main/resources/testsuites/e2e/aws-e2e-tests.yaml
```diff
@@ -11,7 +11,6 @@ tests:
       - name: com.sequenceiq.it.cloudbreak.testcase.e2e.sdx.SdxBaseImageTests
       - name: com.sequenceiq.it.cloudbreak.testcase.e2e.sdx.SdxCloudStorageTests
       - name: com.sequenceiq.it.cloudbreak.testcase.e2e.sdx.SdxRepairTests
-      - name: com.sequenceiq.it.cloudbreak.testcase.e2e.sdx.SdxUpgradeTests
```
gergopapi2: The aws-e2e-tests.yaml file is now the collection of tests that can be run on aws (i.e. someone has written all the code needed). If we delete the SdxUpgradeTests entry we will forget that the SdxUpgradeTests was working for aws. I therefore suggest to create an aws-e2e-tests-inactive.yaml file and move the deleted test-file there with some comments why the test is not working.

lnardai:
Go Go Go!

integration-test/src/main/resources/testsuites/e2e/aws-e2e-tests.yaml
gergopapi2: But it is ok for me as well to create a follow-up jira with this content to have the tests fixed asap. As you like :)

biharitomi: I don't think that we should mark this in our code base, on the other hand a ticket has been created to cover the fix of the mentioned test. It is hard to be forgotten this way:
https://jira.cloudera.com/browse/CB-5189

gergopapi2: I am not sure if that is the best practice as we might not know about all the jiras that are available. For someone who just looks at the code it might not be obvious what is to be done. Nevertheless I approved the PR as a jira is way better than nothing.

PR #7107, opened 2020-01-20
CB-4031 - Failed to delete Data Lake master host :: message fix

gregito:
CB-4031 - Failed to delete Data Lake master host :: message fix

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackDownscaleValidatorService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackDownscaleValidatorService.java
@@ -5,6 +5,7 @@ import static com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.InstanceMeta
 
 import java.util.Objects;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.security.access.AccessDeniedException;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackDownscaleValidatorService.java
```diff
@@ -18,8 +19,13 @@
 
     public void checkInstanceIsTheClusterManagerServerOrNot(String instancePublicIp, InstanceMetadataType metadataType) {
         if (GATEWAY.equals(metadataType) || GATEWAY_PRIMARY.equals(metadataType)) {
-            throw new BadRequestException(String.format("Downscale for node [public IP: %s] is prohibited because it " +
-                            "serves as a host the Cluster Manager server", instancePublicIp));
+            String additionalIpPartForTemplate = "";
+            String messageTemplate = "Downscale for the given node%s is prohibited because it " +
```
doktoric: node %s

gregito: if I make it that way that could cause a double whitespace between the words 'node' and 'is' if there is no public ip for the given node.

PR #7114, opened 2020-01-21
CB-5145 - metrics for inter-service REST calls

attilapalfi92:


common/src/main/java/com/sequenceiq/cloudbreak/jerseyclient/AugmentedRestClient.java
```diff
@@ -7,5 +7,5 @@
 
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface RetryingRestClient {
+public @interface AugmentedRestClient {
```
topolyai5: What was the problem with RetryingRestClient? This is retrying only like previously, right?

bbihari: 'augmented' is a bit too general

attilapalfi92: @topolyai5, no, please check the aspect's code.

attilapalfi92: @bbihari any suggestions?

attilapalfi92: MetricCollectingAndRetryingRestClient would be better? I think the current name is fine.

PR #7122, opened 2020-01-22
CB-4613 Apply syncer to CB for cluster syncs, fix SERVICES_RUNNING st…

bbihari:
…atus


lnardai:
This should go to 2.17 since we plan to enable the syncer there.

PR #7124, opened 2020-01-22
CB-5192 Eliminate Exposed Service enum

bergerdenes:


template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hdfs/HdfsConfigProviderTest.java
```diff
@@ -94,13 +95,13 @@ private TemplatePreparationObject getTemplatePreparationObject(boolean useS3File
                     new AdlsFileSystemConfigurationsView(new AdlsFileSystem(), locations, false);
         }
 
-        Gateway gateway = TestUtil.gatewayEnabledWithExposedKnoxServices(NAMENODE.getKnoxService());
+        Gateway gateway = TestUtil.gatewayEnabledWithExposedKnoxServices(ExposedServiceUtil.exposedService("NAMENODE").getKnoxService());
```
doktoric: getNameNodeService?

core-api/src/test/resources/definitions/exposed-services.json
```diff
@@ -0,0 +1,472 @@
+{
```
topolyai5: I think you should test from the main/resources because these files can be different.

PR #7128, opened 2020-01-23
CB-5235 Refactor ResourceAccessDto

gergopapi2:
This is the first step of refactorings.
Current PR contains Step 1 described under solution.

**Problem**
Many operations work as byName or byCrn. There are separate implementations for byName and byCrn, although the difference is usually in 1 line (and +1 log line). This is a duplication and should be removed.

**Solution**
Step 1: refactor `String name` and `String Crn` into one class: `NameOrCrn`. The class should be instantiable only from either crn or name, never both and never none (thus a NameXorCrn would be a better name). 

Step 2: methods that exist in byName and byCrn flavor should get NameOrCrn as parameter and pass it on to a getEntity method in an underlying service (but not repository).


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/dto/NameOrCrn.java
```diff
@@ -0,0 +1,87 @@
...
+        private NameOrCrnReader(NameOrCrn nameOrCrn) {
+            this.nameOrCrn = nameOrCrn;
+        }
+
+        public static NameOrCrnReader create(NameOrCrn nameOrCrn) {
+            throwIfNull(nameOrCrn, () -> new IllegalArgumentException(NULL_DTO_EXCEPTION_MESSAGE));
+            return new NameOrCrnReader(nameOrCrn);
+        }
+
+        public boolean hasName() {
```
attilapalfi92: Why not moving the methods if this inner class to the original class, and remove `NameOrCrnReader`? Why would we bother instantiating this to be able to read the `NameOrCrn` class?

gergopapi2: The sole purpose of this inner-class thing is an enforced null check. The original code did check for nulls. I did think a while whether to do just delete the null check (and so risk an NPE), or keep it. The (small) added value I see in enforcing null check is the opportunity to better log the error and so help diagnostics.

attilapalfi92: I see. I still have doubts regarding whether it will provide us any real life diagnostic assistance. I think in most cases without this utility we would be able to identify the source of the NPE right away.

gergopapi2: Yeah, I have doubts as well, we do not win much. Ok, so if @gregito agrees, I would then remove the null checks. Please confirm.

gregito: IMHO NPE should never occur. if the only purpose of this NameOrCrnReader class is to ensure that nobody can create an invalid NameOrCrn instance then the null check should be there. on the other hand, null and emptiness checks should be performed at the same time (StringUtils.isEmpty also checks for null), hence one of the checks can be abandoned.

gergopapi2: @gregito You cannot create an empty NameOrCrn object, it either contains the name or the crn. If name is null when you use NameOrCrn.ofName(name), then you get an exception right away (and the same for crn). The only thing NameOrCrn cannot guard for is that instead of creating it you pass null (could happen accidentally).
Another guard the class does not contain but could: if NameOrCrn was create with name, then reading the crn should throw. I think that would make sense.

gergopapi2: Change done.

PR #7132, opened 2020-01-23
CB-5176 Create new test case(s) for validate S3 FreeIPA, DataLake and DataHub logs are created

aszegedi:
@lnardai please review and merge if it is OK for you.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/SdxCloudStorageTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/SdxCloudStorageTests.java
@@ -30,10 +30,10 @@ public class SdxCloudStorageTests extends BasicSdxTests {
     @Test(dataProvider = TEST_CONTEXT)
     @Description(
             given = "there is a running Cloudbreak",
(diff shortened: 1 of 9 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/SdxCloudStorageTests.java
```diff
@@ -55,6 +55,34 @@ public void testSDXWithCloudStorageCanBeCreatedThenDeletedSuccessfully(TestConte
                 .validate();
     }
 
+    @Test(dataProvider = TEST_CONTEXT)
+    @Description(
+            given = "there is a running Cloudbreak",
+            when = "a basic SDX create request with FreeIPA and DataLake Cloud Storage has been sent",
+            then = "SDX should be available along with the created Cloud storage objects"
+    )
+    public void testSDXWithDataLakeAndFreeIPAStorageCanBeCreatedSuccessfully(TestContext testContext) {
```
horadla23: The only difference between this and the first test of class is that this test check freeipa logs, but the first test of the class also creates freeipa (within parent class), i think it would be better if we only add the freeipa log check check into the first test and remove this second one.

PR #7134, opened 2020-01-23
CB-5241 - fixing incorrect MDCContext

attilapalfi92:


common/src/main/java/com/sequenceiq/cloudbreak/logger/MDCBuilder.java
```diff
@@ -79,26 +79,14 @@ public static void buildMdcContext(Object object) {
...
         }
     }
 
-    public static void buildMdcContext(Long resourceId, String resourceName, String type) {
-        buildMdcContext(resourceId == null ? "" : String.valueOf(resourceId), resourceName, type);
-    }
-
-    public static void buildMdcContext(String resourceId, String resourceName, String type) {
-        MdcContext.builder()
-                .resourceCrn(resourceId)
```
attilapalfi92: This was the problem

PR #7135, opened 2020-01-23
Fix of Datahub cluster terminations when FreeIPA clean failed previously

biharitomi:


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -71,25 +72,43 @@ private String toClusterProxyBasepath(String freeIpaClusterCrn) {
...
     public FreeIpaClient getFreeIpaClientForStack(Stack stack) throws FreeIpaClientException {
         LOGGER.debug("Creating FreeIpaClient for stack {}", stack.getResourceCrn());
-
-        try {
-            if (clusterProxyService.isCreateConfigForClusterProxy(stack)) {
-                return getFreeIpaClientBuilderForClusterProxy(stack).build();
-            } else {
-                return getFreeIpaClientBuilder(stack).build();
+        Status stackStatus = stack.getStackStatus().getStatus();
+        if (!Status.FREEIPA_UNREACHABLE_STATUSES.contains(stackStatus)) {
```
lacikaaa: please move this condition into the status as a method, so you don't have the implementation twice

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -71,25 +72,43 @@ private String toClusterProxyBasepath(String freeIpaClusterCrn) {
...
+            } catch (Exception e) {
+                throw new FreeIpaClientException("Couldn't build FreeIPA client. "
+                        + "Check if the FreeIPA security rules have not changed and the instance is in running state. " + e.getLocalizedMessage(), e);
             }
-        } catch (Exception e) {
-            throw new FreeIpaClientException("Couldn't build FreeIPA client. "
-                    + "Check if the FreeIPA security rules have not changed and the instance is in running state. " + e.getLocalizedMessage(), e);
+        } else {
+            String message = String.format("Couldn't build FreeIPA client. Because FreeIPA is in invalid state: '%s'", stackStatus);
+            LOGGER.warn(message);
```
lacikaaa: this else branch can be moved into a separate method

PR #7150, opened 2020-01-27
knox changes

doktoric:


risdenk:
@doktoric I would go more conservative on what is exposed in datalake. My understanding is almost nothing should be exposed in datalake. There are a bunch like Zeppelin that are visible in datalake, but aren't installed today. Probably makes sense to default to false?

doktoric:
@risdenk zeppelin is not really exposed because that is not part of the DL cluster. This change contains that even if we have SOLR on DL cluster and configured then we should not show on the UI.

risdenk:
Yea I just used Zeppelin as an example. I'm saying that for anything we don't want exposed: `visibleForDatalake`  should be false. Hue, Zeppelin, Hive, Spark History, Livy, etc. Even if they aren't installed in the datalake today, I think it would be more clear to say they won't be exposed in the future by default.

doktoric:
Are you sure ? we can do this I just did not know

risdenk:
> Are you sure ? we can do this I just did not know

Nope I don't know for sure. My understanding is that everything in datalake is internal details only and not exposed by default. Might want to check w/ Gunnar?

PR #7152, opened 2020-01-27
CDPCP-624. Prevent cross-account queries for operation status

handavid:
A check has been added to ensure that only an internal actor or an
actor in the same account can see the operation status of an operation
in that account. One difference in behavior is that the UserV1Controller
gets the account id for the request from the requesting user (i.e.,
the x-cdp-actor header). Since the request does not include an account
parameter, there is no way for an internal actor to request status of
an operation in a different account.

This commit splits the existing OperationStatusService into two services
to clarify the division of responsibilities. 

OperationService - provides methods for working with the operation table,
i.e., requesting, accepting, completing, or failing an operation.

OperationStatusService - retrieves operation entities.. The retrieval 
method now requires both an actor crn and account id in addition to the
operation id. A NotFoundException is thrown if the actor crn does not 
match the account id. Internal actors are allowed to query for operations
in any account.

Entity conversion into API models has been moved to the controllers so
that all internal code works with the Operation entity instead.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/OperationV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/OperationV1Controller.java
@@ -5,23 +5,25 @@ import javax.validation.constraints.NotNull;
 
 import org.springframework.stereotype.Controller;
 
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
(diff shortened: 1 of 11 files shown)


handavid:
hmm. I have some second thoughts about splitting the OperationStatusService into two. Initially, I did it to separate out the converters from the rest of the methods. I then moved the converters out to the Controllers. This seems better to me, but I think this also removes the need to split the class. I'll merge the two classes back together and update this PR.

PR #7153, opened 2020-01-28
CB 5278 Ycloud environment creation is broken because of missing validation branch

gergopapi2:


doktoric:
I think it would be great to handle this on a more generic level. This way is the easy and dirty solution for solving the problem.

gergopapi2:
> I think it would be great to handle this on a more generic level. This way is the easy and dirty solution for solving the problem.

Ok, I did a more generic solution.

PR #7159, opened 2020-01-28
CB-5150 added parcel manifest checks

lnardai:
- e2e test was failing because the parcels got removed due to CM build artifact policy
- we should statically check for the presented URL's so Base images will select the correct urls.

### DIFF
--- a/core/src/test/java/com/sequenceiq/cloudbreak/util/ParcelAccessibilityTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/util/ParcelAccessibilityTest.java
@@ -18,6 +18,8 @@ import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpHead;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/sequenceiq/cloudbreak/util/ParcelAccessibilityTest.java
```diff
@@ -26,29 +28,47 @@
...
+    }
+
     @Test
-    public void testParcelAccessibility() throws URISyntaxException, IOException {
+    void validateParcelUrls() {
+        List<DefaultCDHInfo> defaultCDHInfo = getCsds(appYaml);
+        defaultCDHInfo.stream()
+                .filter(s -> !s.getRepo().getStack().isEmpty())
+                .flatMap(s -> s.getRepo().getStack().entrySet().stream())
+                .map(s -> s.getValue() + "manifest.json")
```
topolyai5: Is the parcel url should be ended with `/`?

lnardai: So far all parcel URL's ended with a slash and from now it will be enforced.

PR #7160, opened 2020-01-28
CDPCP-624. Prevent cross-account queries for operation status

handavid:
This commit renames the existing OperationStatusService to OperationService.

Entity conversion into API models has been moved to the controllers so
that all internal code works with the Operation entity instead.

An account id must always be specified when retrieving an operation from
the database. The controller for the getSyncOperationStatus api gets the
account id from the caller (i.e., from the x-cdp-actor header). This ensures
that no actor can retrieve operation status for any account other than its
own account. Since the request does not include an account
parameter, there is no way for an internal actor to request status of
an operation in a different account.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/operation/OperationService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/operation/OperationService.java
@@ -17,29 +17,19 @@ import org.springframework.stereotype.Service;
 
 import com.sequenceiq.freeipa.api.v1.freeipa.user.model.FailureDetails;
 import com.sequenceiq.freeipa.api.v1.freeipa.user.model.SuccessDetails;
-import com.sequenceiq.freeipa.api.v1.freeipa.user.model.SyncOperationStatus;
 import com.sequenceiq.freeipa.api.v1.operation.model.OperationState;
-import com.sequenceiq.freeipa.api.v1.operation.model.OperationStatus;
 import com.sequenceiq.freeipa.api.v1.operation.model.OperationType;
 import com.sequenceiq.freeipa.controller.exception.NotFoundException;
-import com.sequenceiq.freeipa.converter.freeipa.user.OperationToSyncOperationStatus;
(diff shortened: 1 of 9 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/operation/OperationService.java
```diff
@@ -97,12 +87,13 @@ public Operation failOperation(String operationId, String failureMessage) {
...
     }
 
-    public SyncOperationStatus getSyncOperationStatus(String operationId) {
-        return operationToSyncOperationStatus.convert(getOperationForOperationId(operationId));
-    }
-
-    public OperationStatus getOperationStatus(String operationId, String accountId) {
-        return operationToOperationStatusConverter.convert(getOperationForOperationId(operationId));
+    public Operation getOperationForAccountIdAndOperationId(String accountId, String operationId) {
+        Optional<Operation> operationOptional = operationRepository.findByOperationIdAndAccountId(operationId, accountId);
```
giladwolff: So was this functionality was always there but you just didn't use it? What are the use cases for using getOperationForOperationID? should that be removed?

handavid: yeah. this code was there but not used.

The getter by operation id only is used internally for the "complete" and "fail" methods that transition the operation to either the COMPLETED or FAILED state. These are used internally to the FMS. I didn't want to change the behavior of the callers in this PR. We would have to update the flow CleanupEvent to either hold the account id or retrieve the account id from the Stack prior to completing or failing an operation.

I can do that in a follow-on commit.

handavid:
> the one thing I would change/add is to have the account ID explicitly in the call to retrieve the operation, that is, not to derive it from the actor CRN. This can be done in a later commit. This is not only to support internal actor calls but also supporters for example. Supporters will be from a different account than the account they are trying to support. In general our best practice is to not derive any parameters to the call from the actor who is making the call.

I understand your point and can address it in a follow-on commit.

handavid:
@lacikaaa I've spun up and terminated an environment, dl, and dh successfully.

PR #7161, opened 2020-01-28
CB-5306. Global options to customize default values of telemetry logging features

oleewere:
add global options to set features (if those are not provided at all on the UI), could be useful for e2e testing


oleewere:
closing for now

PR #7167, opened 2020-01-29
CB-5320 Add cluster template for OpDb with Phoenix for Azure

stoty:
This is extending CB-4551 to Azure.

In CB-4551 we have added a new OpDB blueprint that adds Phoenix, but we have only added the cluster blueprint for AWS.

This patch adds the cluster blueprint for Azure as well.

The blueprint is mostly identical to the existing opdb 7.0.2 Azure blueprint, save that it refers to the newer blueprint, and it includes the approved, but not yet merged CB-5275 fix.

Testing done:

Since I have been unable to get a local Cloudbreak instance working with Azure, I used the mow-dev environment, where created an environment on azure, uploaded the blueprint, and built an OpDB DH cluster with the latest CM and CDPD.

I have run a simple Phoenix smoke test on the cluster, and it passed. 

Since the existing Azure OpDB cluster template, and this one only differs in the referred blueprint, I expect that this was sufficient testing.

Closes #CB-5320

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/azure/azure-opdb-710.json
+++ b/core/src/main/resources/defaults/clustertemplates/azure/azure-opdb-710.json
@@ -1,9 +1,9 @@
 {
-  "name": "Operational Database with SQL for AWS",
+  "name": "Operational Database with SQL for Azure",
   "description": "",
   "type": "OPERATIONALDATABASE",
   "featureState": "PREVIEW",
-  "cloudPlatform": "AWS",
+  "cloudPlatform": "AZURE",
   "distroXTemplate": {
     "cluster": {
       "blueprintName": "CDP 1.2 - Operational Database: Apache HBase, Phoenix"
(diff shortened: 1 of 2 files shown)


core/src/main/resources/defaults/clustertemplates/azure/azure-opdb-710.json
```diff
@@ -0,0 +1,122 @@
+{
+  "name": "Operational Database with Phoenix for AWS",
```
joshelser: nit: with Apache Phoenix

_maybe_ pedantic, but I'm worried where cloudbreak might display this name and we then get into a trademarks issue with the ASF.

stoty: This is displayed very prominently, but your suggestion definitely won't fit into the list.

How about replacing "with Phoenix" with "with SQL" in the descriptions ? That way we fit into a UI, and sidestep the trademark problem. The other DataHbus do similarly, they do not name the specific software packages.

stoty:
The test failures seem to be independent of the patch.

core/src/main/resources/defaults/clustertemplates/azure/azure-opdb-710.json
joshelser: I like that suggestion!

I think this is all temporary, too, and the normal will be to have SQL in all opdb :)

stoty: Thanks. It's done. Could you resolve/remove the request for change to indicate that this is resolved ?

stoty:
See CB-5752 instead

PR #7178, opened 2020-01-30
CB-5164 Improve polling logic in E2E tests - flow api for every service

horadla23:


horadla23:
- FlowAPI added to every service
- polling based on flow api added to WaitUtil in case of SDX and Distrox
- polling based on flow added to one E2E test

PR #7183, opened 2020-01-30
CB-5371. Feature refactor & (E2E) remove use shared credential from the API (request)

oleewere:
- remove use shared credential from the API

#### WHY?

Right now, it's not enough to set use altus credential globally in order to use a shared altus credential instead of communicate with UMS, you also need to provide this "ask" in the request. 
That works fine for e2e tests but not for the GUI test, as that option should be always hidden.
So make sense to remove from the API - at least the request part. (respone is kept so we can see on the API if that is used)

also refactor features object to set wrapper objects more easily

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
@@ -88,7 +88,7 @@ public class TelemetryConverter {
             telemetry.setWorkloadAnalytics(workloadAnalytics);
             setWorkloadAnalyticsFeature(telemetry, features);
             setReportDeploymentLogs(request, features);
-            setUseSharedAltusCredential(request, features);
+            setUseSharedAltusCredential(features);
             telemetry.setFluentAttributes(request.getFluentAttributes());
         }
         setMeteringFeature(type, features);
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/TelemetryConverter.java
```diff
@@ -367,10 +364,12 @@ private void setReportDeploymentLogs(TelemetryRequest request, Features features
...
-    private void setUseSharedAltusCredential(TelemetryRequest request, Features features) {
-        if (useSharedAltusCredential && request.getFeatures() != null && request.getFeatures().getUseSharedAltusCredential() != null) {
-            LOGGER.debug("Fill shared altus credential setting from telemetry feature request");
-            features.setUseSharedAltusCredential(request.getFeatures().getUseSharedAltusCredential());
+    private void setUseSharedAltusCredential(Features features) {
+        if (useSharedAltusCredential) {
+            LOGGER.debug("Fill shared altus credential setting as that is enabled globally");
+            FeatureSetting useSharedAltusCredentialFeature = new FeatureSetting();
+            useSharedAltusCredentialFeature.setEnabled(true);
+            features.setUseSharedAltusCredential(useSharedAltusCredentialFeature);
```
horadla23: this logic occurs three times in this PR, maybe it is worth to extract it into some static class and call it everywhere

oleewere: it's just a few steps, maybe in features object there should be a method to set this wrapper objects based on a boolean param and add a `@JsonIgnore` there. but that should be done to every occurrences of these feature settings, expect that this PR will be probably larger

PR #7194, opened 2020-02-03
CB-5240 Refactor byName and byCrn methods in StackService, StackCommo…

gergopapi2:
For a description please see problem description at: https://github.com/hortonworks/cloudbreak/pull/7128

This refactor contains changes only where resourceAccessDto was already used. 
Methods that still use String name or String crn need to be changed => a later refactorign step.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
@@ -207,6 +207,10 @@ public interface StackRepository extends WorkspaceResourceRepository<Stack, Long
     @Query("SELECT s.id FROM Stack s WHERE s.name= :name AND s.workspace.id= :workspaceId AND s.terminated = null")
     Optional<Long> findIdByNameAndWorkspaceId(@Param("name") String name, @Param("workspaceId") Long workspaceId);
 
+    @DisableCheckPermissions
+    @Query("SELECT s.id FROM Stack s WHERE s.resourceCrn= :crn AND s.workspace.id= :workspaceId AND s.terminated = null")
+    Optional<Long> findIdByCrnAndWorkspaceId(@Param("crn") String crn, @Param("workspaceId") Long workspaceId);
+
     @CheckPermissionsByWorkspaceId
     @Query("SELECT s.id as id, "
             + "s.resourceCrn as resourceCrn, "
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
@@ -16,6 +16,7 @@ import org.springframework.stereotype.Component;
 
 import com.google.common.collect.Sets;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status;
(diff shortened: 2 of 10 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -207,6 +207,10 @@
     @Query("SELECT s.id FROM Stack s WHERE s.name= :name AND s.workspace.id= :workspaceId AND s.terminated = null")
     Optional<Long> findIdByNameAndWorkspaceId(@Param("name") String name, @Param("workspaceId") Long workspaceId);
 
+    @DisableCheckPermissions
+    @Query("SELECT s.id FROM Stack s WHERE s.resourceCrn= :crn AND s.workspace.id= :workspaceId AND s.terminated = null")
+    Optional<Long> findIdByCrnAndWorkspaceId(@Param("crn") String name, @Param("workspaceId") Long workspaceId);
```
foldik: Please rename the param `name` to `crn`.

gergopapi2: Thanks, a nice catch! Done.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradeService.java
```diff
@@ -74,18 +75,18 @@
...
+                Optional<Stack> stack = stackService.findStackByNameOrCrnAndWorkspaceId(nameOrCrn, workspaceId);
                 if (stack.isPresent()) {
                     try {
                         return getUpgradeOption(stack.get(), workspaceId, user);
                     } catch (CloudbreakImageNotFoundException | CloudbreakImageCatalogException e) {
                         throw new BadRequestException(e.getMessage());
                     }
                 } else {
-                    throw notFoundException("Stack", stackName);
+                    throw notFoundException("Stack", nameOrCrn.toString());
```
horadla23: small thing: why are you using toString here instead of name?

gergopapi2: Yes, that does not make sense any more. In the first implementation if you called getName() but it had the crn instead, there was no exception thrown. So I just wanted to make sure that it is logged to see the class in full.
In the present implementation if you call the wrong get method (name instead of crn) you get an exception and will be notified of the mismatch => so toString here does not make much sense any more.

gergopapi2: Long story short, I will change it.

gergopapi2: On this very place I need to use toString as nameOrCrn can hold either name or crn, it is not known statically. NameOrCrn object is passed to the method findStackByNameOrCrnAndWorkspaceId accepting both name and crn.

gergopapi2: @horadla23 Tests: I ran e2e aws tests locally:
testCreateScaleStopStartCluster
testSDXRepairIDBRokerWithStoppedEC2Instance

doktoric:
@gergopapi2 please rebase and merge

gergopapi2:
> @gergopapi2 please rebase and merge

Yep, it is rebased and merging it right now.

PR #7195, opened 2020-02-03
CB-5385 Flow API check in SDX might return with false result

horadla23:


horadla23:
tested with latest 2.16 build locally using AwsSdxTests.testSDXMultiRepairIDBRokerAndMasterWithTerminatedECInstances, will test with latest 2.17 build soon

horadla23:
tested with latest 2.17 build locally using SdxTests.disabledTestSDXRepairIDBRokerWithStoppedEC2Instance (which was enabled locally)

lnardai:
We should also reenable the tests right? This should fix that issue?

PR #7196, opened 2020-02-03
CB-5394 West US region is not enabled in environment. Enabled regions westus

aszegedi:
@lnardai or @doktoric please review and merge if it is OK for you


aszegedi:
Need to be investigate further in case of other providers

PR #7205, opened 2020-02-04
CB-4966 reduce db queries, decrease thread pool size

topolyai5:


freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -40,18 +43,18 @@
     @Inject
     private InstanceMetaDataService instanceMetaDataService;
 
-    private RPCResponse<Boolean> checkStatus(Stack stack) throws Exception {
+    private RPCResponse<Boolean> checkStatus(Stack stack, Set<InstanceMetaData> checkableInstances) throws Exception {
```
keyki: checkableInstances sounds weird, can we just call it instances?

PR #7212, opened 2020-02-04
CB-5423 ensure that UserDataSetup flow is restartable

akanto:


doktoric:
pushed the unit test fix nd wiating for the build

PR #7214, opened 2020-02-04
CB-5308 Manage autorestart_on_stop property in Knox

daszabo:


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/termination/ClusterTerminationActions.java
```diff
@@ -34,8 +32,8 @@
     @Inject
     private JobService jobService;
 
-    @Bean(name = "PREPARE_CLUSTER_STATE")
-    public Action<?, ?> prepareCluster() {
+    @Bean(name = "DISABLE_KERBEROS_STATE")
```
keyki: I'm not sure what happens with clusters if we remove a flow step. We should test it and if it causes problem we should keep it, but just send the success event maybe.

PR #7215, opened 2020-02-04
CB-5385 check flow logs and queues of related flow chains + new flow api method

horadla23:


horadla23:
tested successfully locally on AWS with e2e test SdxRepairTests.testSDXRepairIDBRokerWithStoppedEC2Instance (which has failed earlier caused by corner case of flow check)

PR #7221, opened 2020-02-05
CB-5434 fix sdx-internal request validation - to 2.17

gergopapi2:
The same commit as https://github.com/hortonworks/cloudbreak/pull/7216/commits/0cc1911608106c8acf1366808309e4da3a7a0a83
Difference: base is 2.17, to avoid duplicate commits after cherry-pick.


keyki:
We don't need this commit on 2.17

PR #7224, opened 2020-02-05
CB-5303. Add stop telemetry agents action before termination.

oleewere:
Shutdown in some cases are too agressive or too fast (like for AWS) and the shutdown scripts are not finishing for td-agent on the nodes.
Solution for that is to run fluent.agent-stop.sls salt script on the nodes before termination
- Add as an action for freeipa
- Add as an another task for core code (to pre stack termination)

If command fails, that should not be considered as an error, so put the salt execution in try catch block and just log the errors 

tested manually

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/StackTerminationFlowConfig.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/StackTerminationFlowConfig.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEven
 import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.CLUSTER_PROXY_DEREGISTRATION_FINISHED_EVENT;
 import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.REMOVE_MACHINE_USER_FINISHED_EVENT;
 import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.STACK_TERMINATION_FAIL_HANDLED_EVENT;
+import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.STOP_TELEMETRY_AGENT_FINISHED_EVENT;
 import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.TERMINATION_EVENT;
 import static com.sequenceiq.freeipa.flow.stack.termination.StackTerminationEvent.TERMINATION_FAILED_EVENT;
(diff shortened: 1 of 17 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/StackTerminationFlowConfig.java
```diff
@@ -32,7 +34,8 @@
                     .defaultFailureEvent(TERMINATION_FAILED_EVENT)
                     .from(INIT_STATE).to(DEREGISTER_CLUSTERPROXY_STATE).event(TERMINATION_EVENT).noFailureEvent()
                     .from(DEREGISTER_CLUSTERPROXY_STATE).to(DEREGISTER_CCMKEY_STATE).event(CLUSTER_PROXY_DEREGISTRATION_FINISHED_EVENT).noFailureEvent()
-                    .from(DEREGISTER_CCMKEY_STATE).to(REMOVE_MACHINE_USER_STATE).event(CCM_KEY_DEREGISTRATION_FINISHED_EVENT).noFailureEvent()
+                    .from(DEREGISTER_CCMKEY_STATE).to(STOP_TELEMETRY_AGENT_STATE).event(CCM_KEY_DEREGISTRATION_FINISHED_EVENT).noFailureEvent()
+                    .from(STOP_TELEMETRY_AGENT_STATE).to(REMOVE_MACHINE_USER_STATE).event(STOP_TELEMETRY_AGENT_FINISHED_EVENT).noFailureEvent()
```
lacikaaa: this doesn't seem right in case of CCM. if we deregister freeipa from CCM and CP we won't be able to communicate with it anymore.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -195,6 +195,27 @@ public String submit(SaltConnector saltConnector) {
         }
     }
 
+    @Override
+    public void stopTelemetryAgent(List<GatewayConfig> allGateway, Set<Node> nodes, ExitCriteriaModel exitModel)
+            throws CloudbreakOrchestratorFailedException {
+        GatewayConfig primaryGateway = getPrimaryGatewayConfig(allGateway);
+        Set<String> gatewayTargets = getGatewayPrivateIps(allGateway);
+        try (SaltConnector sc = createSaltConnector(primaryGateway)) {
+            BaseSaltJobRunner baseSaltJobRunner = new BaseSaltJobRunner(gatewayTargets, nodes) {
```
lacikaaa: could you move this out?

PR #7231, opened 2020-02-06
CB-5271: get images from cm for stack upgrade

tiborpopovics:


core-api/src/main/java/com/sequenceiq/cloudbreak/cloud/MaintenanceVersionComparator.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * This comparator supports only the three-part version number format like 1.2.3
+ */
+@Component
+public class MaintenanceVersionComparator {
```
daszabo: According to the name of this class, when we eventually will support minor and major upgrades, new comparators will be needed for those, which smells like code duplication. I would advise to extend the already existing Version class with segment-specific comparison reusing parts of the code there.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/StackUpgradeImageFilter.java
```diff
@@ -0,0 +1,82 @@
...
+    Images filter(Images availableImages, Image currentImage, String cloudPlatform) {
+        return new Images(null, null, null, availableImages.getCdhImages().stream()
+                .filter(validateCmVersion(currentImage).or(validateStackVersion(currentImage)))
+                .filter(validateCloudPlatform(cloudPlatform))
+                .filter(validateOsVersion(currentImage))
+                .filter(validateCfmVersion(currentImage))
+                .filter(validateCspVersion(currentImage))
+                .filter(validateSaltVersion(currentImage))
+                .filter(filterCurrentImage(currentImage))
+                .collect(Collectors.toList()), null);
```
daszabo: Shouldn't this filter images also based on supported cb versions?

core/src/main/java/com/sequenceiq/cloudbreak/service/image/StackUpgradeImagesService.java
```diff
@@ -0,0 +1,63 @@
...
+    private Image getImage(Stack stack) throws CloudbreakImageNotFoundException {
+        return imageService.getImage(stack.getId());
+    }
+
+    private Images getImagesFromCatalog(String imageCatalogUrl) throws CloudbreakImageCatalogException {
+        return imageCatalogProvider.getImageCatalogV2(imageCatalogUrl).getImages();
+    }
+
+    private Images filterImages(Images images, Image currentImage, String cloudPlatform) {
+        return stackUpgradeImageFilter.filter(images, currentImage, cloudPlatform);
```
daszabo: This only filters CDH images. Might as-well only pass that list to the filter method.

PR #7238, opened 2020-02-06
CB-5465 RDS subnet selection should use limited number of subnets in an environment

lacikaaa:


redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public void validate(Collection<CloudSubnet> cloudSubnets, int minSubnetNeededInDifferentAz) {
+        validateEnoughSubnetIsPresent(cloudSubnets, minSubnetNeededInDifferentAz);
+        validateAzNumber(cloudSubnets, minSubnetNeededInDifferentAz);
+    }
+
+    private void validateAzNumber(Collection<CloudSubnet> cloudSubnets, int minSubnetNeededInDifferentAz) {
+        long numAZs = cloudSubnets.stream().map(CloudSubnet::getAvailabilityZone).distinct().count();
+        if (numAZs < minSubnetNeededInDifferentAz) {
+            throw new BadRequestException("All subnets are in the same availability zone: at least two subnets in two different availability zones needed");
```
gergopapi2: I think if the solution is generalized then this message is not good anymore: "Subnets are in %d different AZ, but subnets in %d different AZs required."

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidator.java
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+
+    private void validateEnoughSubnetIsPresent(Collection<CloudSubnet> cloudSubnets, int minSubnetNeededInDifferentAz) {
+        if (cloudSubnets.size() < minSubnetNeededInDifferentAz) {
+            throw new BadRequestException("Insufficient number of subnets: at least two subnets in two different availability zones needed");
```
gergopapi2: In a generalized solution this message is not valid any more. I would suggest "Insufficient number of subnets: at least %d subnets required"

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private void validateAzNumber(Collection<CloudSubnet> cloudSubnets, int minSubnetNeededInDifferentAz) {
+        long numAZs = cloudSubnets.stream().map(CloudSubnet::getAvailabilityZone).distinct().count();
+        if (numAZs < minSubnetNeededInDifferentAz) {
+            throw new BadRequestException("All subnets are in the same availability zone: at least two subnets in two different availability zones needed");
+        }
+    }
+
+    private void validateEnoughSubnetIsPresent(Collection<CloudSubnet> cloudSubnets, int minSubnetNeededInDifferentAz) {
+        if (cloudSubnets.size() < minSubnetNeededInDifferentAz) {
```
doktoric: I think this should be based on the AWS db deploy config. As I know currently we are using non multi AZ deployments in local mode so probably we dont need 2 subnet.

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooserTest.java
```diff
@@ -0,0 +1,165 @@
...
+                new CloudSubnet("subnet3", "", "us-west-2b", "", false, true, true),
+                new CloudSubnet("subnet4", "", "us-west-2a", "", false, true, true),
+                new CloudSubnet("subnet5", "", "us-west-2a", "", false, true, true),
+                new CloudSubnet("subnet6", "", "us-west-2b", "", false, true, true)
+        );
+
+        List<CloudSubnet> chosenSubnets = underTest.chooseSubnets(subnets);
+
+        assertThat(chosenSubnets, hasSize(2));
+        verify(awsSubnetValidator).validate(subnets, 2);
```
doktoric: Can you please validate that we choose 2 subnet in different az ?

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidatorTest.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void testNOkWithTooFewSubnet() {
+        underTest.validate(List.of(cloudSubnet1), 2);
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void testNOkWithTooFewAz() {
+        underTest.validate(List.of(cloudSubnet1, cloudSubnet2), 2);
```
doktoric: I think it shoulf be great to check the error message as well.

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooser.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final int MIN_SUBNET_IN_DIFFERENT_AZ = 2;
+
+    private static final int MAX_SUBNET_IN_DEFFERENT_AZ = 3;
+
+    @Inject
+    private AwsSubnetValidator awsSubnetValidator;
+
+    public List<CloudSubnet> chooseSubnets(List<CloudSubnet> subnetMetas) {
+        LOGGER.debug("Subnets to choose from: [{}]", subnetMetas);
+        awsSubnetValidator.validate(subnetMetas, MIN_SUBNET_IN_DIFFERENT_AZ);
```
doktoric: this should be depend on the deployment config

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooser.java
```diff
@@ -0,0 +1,88 @@
...
+        } else {
+            LOGGER.debug("No public subnets needed");
+            return Set.of();
+        }
+    }
+
+    private Set<CloudSubnet> collectLimitedPublicSubnetsInDifferentAz(int chosenSubnetSize, Map<String, Collection<CloudSubnet>> publicSubnetPerAz) {
+        Set<CloudSubnet> chosenPublicSubnets = publicSubnetPerAz.values().stream()
+                .map(cloudSubnets -> cloudSubnets.stream().findFirst().get())
+                .limit(MIN_SUBNET_IN_DIFFERENT_AZ - chosenSubnetSize)
```
gergopapi2: I would limit to MAX_SUBNET_IN_DIFFERENT_AZ

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooserTest.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String SUBNET_4 = "subnet-4";
+
+    @Mock
+    private AwsSubnetValidator awsSubnetValidator;
+
+    @InjectMocks
+    private AwsSubnetChooser underTest;
+
+    @Test
+    public void testMultiplePrivateSubnetDifferentAz() {
```
gergopapi2: I would rename it to testMultiplePrivateSubnet**Two**DifferentAz

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidator.java
lacikaaa: we won't address it in this pr

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooser.java
lacikaaa: we won't address it in this pr

redbeams/src/main/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooser.java
lacikaaa: we don't want to add public subnet if we can avoid it

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/AwsSubnetChooserTest.java
lacikaaa: fixed

redbeams/src/test/java/com/sequenceiq/redbeams/service/network/AwsSubnetValidatorTest.java
lacikaaa: fixed

PR #7243, opened 2020-02-07
CB-5475 Don't check permission while migrating secrets to vault

lacikaaa:


akanto:
This seems to OK to me as a temporary solution, since we are redesigning the AuthZ to UMS. Thanks, Misi.

PR #7244, opened 2020-02-07
CB-4067 – service to ensure no experience attached to env on delete

gregito:
CB-4067 – service to ensure no experience attached to env on delete

### DIFF
--- a/environment/src/main/resources/application.yml
+++ b/environment/src/main/resources/application.yml
@@ -58,7 +58,12 @@ vault:
       login.role: "cloudbreak.default"
 
 environment:
-  cert.dir: /certs/
+  rest:
+    debug: false
+  cert:
+    dir: /certs/
+    ignorePreValidation: false
+    validation: true
   structuredevent:
     rest:
       enabled: false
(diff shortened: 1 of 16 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/util/ConditionBasedEvaluatorUtil.java
```diff
@@ -41,4 +41,6 @@ private ConditionBasedEvaluatorUtil() {
         }
     }
 
+    public static void throwIfTrue(boolean empty) {
+    }
```
lajosrodek: Apart from the empty implementation (no `throw`), I wonder if this new overloaded variant is used anywhere.

common/src/main/java/com/sequenceiq/cloudbreak/jerseyclient/client/WebTargetClientProvider.java
```diff
@@ -0,0 +1,18 @@
...
+import com.sequenceiq.cloudbreak.client.RestClientUtil;
+import org.springframework.stereotype.Component;
+
+import javax.ws.rs.client.Client;
+
+/**
+ * Simple component to make testing classes/methods easier which uses the javax.ws.rs.client.Client
+ */
+@Component
+public class WebTargetClientProvider {
```
topolyai5: we have an injectable version with ssl and more shiny things, please talk with @keyki how you can use that stuff

environment/src/main/resources/application.yml
```diff
@@ -156,4 +156,17 @@ spring:
...
 notification:
-  urls: http://localhost:3000/notifications
\ No newline at end of file
+  urls: http://localhost:3000/notifications
+
+xp:
+  path:
+    componentToReplace: "{crn}"
+    postfix: "/environment/{crn}"
+  protocol:
```
topolyai5: why dont you set the https here?

environment/src/main/resources/application.yml
```diff
@@ -156,4 +156,17 @@ spring:
...
+xpservices:
+  experiences[Machine Learning Experience].pathPrefix: ${MLX_PATH_PREFIX}
```
topolyai5: where should be the prefix set?

environment/src/main/resources/application.yml
gregito: because it's https by default but I left it to be configurable

https://github.com/hortonworks/cloudbreak/pull/7244/files#diff-38bb82f2416cf83514e295898ba18688R39

topolyai5: Yeh, I saw this constructor, but we try to handle the default values in applciation.yml instead of hardcoded values, as I know

PR #7245, opened 2020-02-07
CB-5483 - paging for role fetching from aws

gregito:
CB-5483 - paging for role fetching from aws


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -157,6 +159,12 @@
     @Value("#{'${cb.aws.distrox.enabled.instance.types:}'.split(',')}")
     private List<String> enabledDistroxInstanceTypes;
 
+    @Value("${cb.aws.fetch.max.items:500}")
```
topolyai5: please fetch all items

gregito: I'll keep the max item to protect the API from fetching a GB of data at once, but I'm going to remove the iteration count, so we'll fetch the access configs until all of them have arrived.

PR #7249, opened 2020-02-10
CB 5486: Add Instance ID to health check response.

holleyism:
Adding instance ID to the health check response.

Updated unit tests initialization.  Additionally ran manual tests.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
@@ -1,8 +1,10 @@
 package com.sequenceiq.freeipa.service.stack;
 
+import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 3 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -0,0 +1,152 @@
...
+            parseMessages(rpcResponse.get(), response);
+            if (isOverallHealthy(response)) {
+                response.setStatus(DetailedStackStatus.PROVISIONED.getStatus());
+            } else {
+                response.setStatus(DetailedStackStatus.UNHEALTHY.getStatus());
+            }
+        } else {
+            response.setStatus(DetailedStackStatus.UNREACHABLE.getStatus());
+            NodeHealthDetails nodeResponse = null;
+            nodeResponse = new NodeHealthDetails();
```
topolyai5: You can do it inline

lacikaaa:
please rebase

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: missed this in the refactoring.  Fixed.

PR #7255, opened 2020-02-10
CB-5408. Apply anonymization rules on td-agent configs (by Salt)

oleewere:
Anonymization rules skeleton if the anonymization rules is defined on UMS account side (that part gives back null)

Notes:
- UMS is not supporting this option yet, so the method always gives back null
- As anonymization rules field is empty it will be missing from the salt config, so technicallly this change won't do anything until UMS side is not implemented (but i have tried out this by manually set the objects and set it through salt)
- If UMS side is done, only change here will be to transform the UMS response to our object format (that can be done in UMS client, as it was done for Altus credential - if that happens, maybe the model can be moved to auth-connector...but for know it's good at its actual place as that is related with telemetry)
- maybe later makes sense to persist the rules for DataLake/DataHub/FreeIPA, just to show how the cluster was deployed

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -2,6 +2,7 @@ package com.sequenceiq.cloudbreak.auth.altus;
 
 import static io.grpc.internal.GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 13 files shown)


doktoric:
I think we should not persist this information on our side because then if the user modify the rule set in UMS then we need to synchronize.

oleewere:
@doktoric, we are not persisting that here in this PR

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -453,6 +454,18 @@ public void deleteMachineUserAccessKeys(String actorCrn, String machineUserCrn,
...
+    /**
+     * Gather anonymization rules for a specific account
+     * NOTE: not supported yet on UMS side
+     * @param accountId account that owns the anonymization rules
+     * @param actorCrn actor that requests to gather the anonymization rules
+     * @return a list of anonymization rules for an UMS account
+     */
+    public List<AnonymizationRule> getAnonymizationRules(String accountId, String actorCrn) {
+        LOGGER.debug("Try getting anonymization rules for {} by {}", accountId, actorCrn);
+        return null;
```
topolyai5: The method should be returned with an empty list, we should avoid the unnecessary NPE

oleewere: if you read the description, it's returning null "by design" as it does nothing, that's always the case, it cannot cause NPE .
I can do that you asked but does not make that much sense to change it or not

PR #7265, opened 2020-02-11
CB-5501 Increase timeout for ms clients in IT

lacikaaa:


akanto:
Beyond this, we shall also investigate why simple check status can not return a response with 20sec.

PR #7274, opened 2020-02-12
CB-5526 Subnet should be private if it has no IG

gergopapi2:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -389,7 +389,7 @@ private DescribeVpcsRequest getDescribeVpcsRequestWIthFilters(Map<String, String
                             subnetName.orElse(subnet.getSubnetId()),
                             subnet.getAvailabilityZone(),
                             subnet.getCidrBlock(),
-                            !subnet.isMapPublicIpOnLaunch(),
```
topolyai5: We use this method here `AwsNetworkService#80`. Is it same?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -389,7 +389,7 @@ private DescribeVpcsRequest getDescribeVpcsRequestWIthFilters(Map<String, String
...
+                            !hasInternetGateway,
```
keyki: We are passing hasInternetGateway's value 2 times here.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -1,27 +1,58 @@
...
 
-    private String getSubnetIdByPreferedSubnetType(NetworkDto network, boolean preferPrivate) {
-        Optional<CloudSubnet> subnet =
-                network.getSubnetMetas().values().stream().filter(cloudSubnet -> preferPrivate == cloudSubnet.isPrivateSubnet()).findAny();
-        return subnet.isPresent() ? subnet.get().getId() : network.getSubnetIds().iterator().next();
+    private String tryGetOnePrivateSubnet(List<CloudSubnet> subnetMetas) {
+        Optional<CloudSubnet> privateSubnet = subnetMetas.stream()
+                .filter(CloudSubnet::isPrivateSubnet)
+                .findFirst();
+        return privateSubnet.isPresent() ? privateSubnet.get().getId() : "";
```
lacikaaa: return with the optional here

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -1,27 +1,58 @@
...
+            chosenSubnetId = subnetMetas.get(0).getId();
+        }
+        return chosenSubnetId;
+    }
+
+    private String tryGetOnePublicSubnet(List<CloudSubnet> subnetMetas) {
+        Optional<CloudSubnet> publicSubnet = subnetMetas.stream()
+                .filter(sn -> !sn.isPrivateSubnet() && sn.isMapPublicIpOnLaunch())
+                .findAny();
+        return publicSubnet.isPresent() ? publicSubnet.get().getId() : "";
```
lacikaaa: return with optional here

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -1,27 +1,58 @@
...
+    }
+
+    private String getSubnetIdByPreferedSubnetType(NetworkDto network, boolean preferPrivate, CloudPlatform cloudPlatform) {
+        List<CloudSubnet> subnetMetas = new ArrayList<>(network.getSubnetMetas().values());
+        return AZURE.equals(cloudPlatform)
+                ? subnetMetas.get(0).getId()
+                : getSubnetIdByPreferedSubnetTypeAws(subnetMetas, preferPrivate);
+    }
+
+    private String getSubnetIdByPreferedSubnetTypeAws(List<CloudSubnet> subnetMetas, boolean preferPrivate) {
```
lacikaaa: I'm missing the scenario here, when we prefer private, but we don't have one, we should choose a public with mapPublicIpOnLaunch enabled if it's possible, and fall back to any subnet if it fails

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
lacikaaa: thats ok, as private subnet flag is driven by it

gergopapi2: Ok, I see the value in that and addressed it with a new ctor. However, it would be great to refactor CloudSubnet as it has too many ctors.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
gergopapi2: Good idea, did it, thanks.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
gergopapi2: Good idea, did it, thanks.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
gergopapi2: Yes, that is also useful, added it. Thanks.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -1,27 +1,66 @@
...
+        Optional<CloudSubnet> foundCloudSubnet = preferPrivate ? tryGetOnePrivateSubnet(subnetMetas) : tryGetOnePublicSubnet(subnetMetas);
+        if (foundCloudSubnet.isEmpty()) {
+            LOGGER.debug("Choosing subnet alternate strategy");
+            // This is just trying the reverse option, i.e. if we preferred private but there was no private then let's look for a suitable public subnet
+            foundCloudSubnet = preferPrivate ? tryGetOnePublicSubnet(subnetMetas) : tryGetOnePrivateSubnet(subnetMetas);
+        }
+        if (foundCloudSubnet.isEmpty()) {
+            LOGGER.debug("Choosing subnet, fallback to any subnet");
+            foundCloudSubnet = Optional.of(subnetMetas.get(0));
+        }
```
lacikaaa: it would reflect the logic more if you move this if into the upper one

gergopapi2: I see. Actually I like to avoid many indentations, they are hard to read. Also, this is yet another step to try to get a subnet, different from the previous one.

lacikaaa: also can be moved to a separate method called like `fallbackToAny`

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -1,27 +1,66 @@
...
+    }
+
+    private String getSubnetIdByPreferedSubnetTypeAws(List<CloudSubnet> subnetMetas, boolean preferPrivate) {
+        LOGGER.debug("Choosing aws subnet, preferring private: {}, available", preferPrivate);
+        Optional<CloudSubnet> foundCloudSubnet = preferPrivate ? tryGetOnePrivateSubnet(subnetMetas) : tryGetOnePublicSubnet(subnetMetas);
+        if (foundCloudSubnet.isEmpty()) {
+            LOGGER.debug("Choosing subnet alternate strategy");
+            // This is just trying the reverse option, i.e. if we preferred private but there was no private then let's look for a suitable public subnet
+            foundCloudSubnet = preferPrivate ? tryGetOnePublicSubnet(subnetMetas) : tryGetOnePrivateSubnet(subnetMetas);
+        }
```
lacikaaa: this could be moved out and givne a meaningful name so we don't need the comment

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudSubnet.java
```diff
@@ -39,6 +39,16 @@ public CloudSubnet(String id, String name, String availabilityZone, String cidr)
         this.cidr = cidr;
     }
 
+    public CloudSubnet(String id, String name, String availabilityZone, String cidr, boolean privateSubnet, boolean mapPublicIpOnLaunch) {
```
lacikaaa: I don't like it's called `privateSubnet` and we set the IGW from it. We request the IGW setting `boolean hasInternetGateway = awsSubnetIgwExplorer.hasInternetGatewayOfSubnet(describeRouteTablesResult, subnet.getSubnetId());` from AWS so it would make sense, that the parameter is called IGW, and we set the privateSubnet from it's value and not reversed.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
gergopapi2: Yes, that's what I did first, but then the tests were harder to read. The two terms on aws are interchangeable. But of course, I can change.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudSubnet.java
gergopapi2: Yes, that's what I did first, but then the tests were harder to read. The two terms on aws are interchangeable. But of course, I can change.

PR #7276, opened 2020-02-12
CB-5353 Add runtime to datalake

sodre90:


akanto:
The code looks ok, although what I have seen is:
- the template picker on DataHub is greyed out
- the DataHub templates are not filtered out

PR #7288, opened 2020-02-14
CB-5122 Remove eager fetching of child environments

Bajzathd:
CB-5122 Remove eager fetching of child environments


lacikaaa:
I meant do we need that relationship on stack at all? Can't we just fetch that info when we need it?

PR #7301, opened 2020-02-17
CB-5343 Resizing compute nodes does not work

topolyai5:


common/src/main/java/com/sequenceiq/cloudbreak/validation/FunctionChecked.java
```diff
@@ -0,0 +1,7 @@
+package com.sequenceiq.cloudbreak.validation;
+
+@FunctionalInterface
+public interface FunctionChecked<T, R, E extends Exception> {
```
attilapalfi92: Why not CheckedFunction? Or ValidatedFunction? Would sound more correct.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -187,32 +187,40 @@ private Boolean isPrewarmed(Long clusterId) {
...
+            pollRefresh(refreshServicesCmd);
+            LOGGER.debug("Config deployed and stale services are refreshed in Cloudera Manager.");
         } else {
             LOGGER.debug("No stale services found in Cloudera Manager.");
         }
     }
 
+    private ApiCommand getApiCommand(List<ApiCommand> commands, String commandString, String clusterName, FunctionChecked<String, ApiCommand, ApiException> fn)
+            throws ApiException {
+        Optional<ApiCommand> optionalDeployClientConfigCommand = commands.stream().filter(cmd -> commandString.equals(cmd.getName())).findFirst();
```
attilapalfi92: I think `optionalDeployClientConfigCommand` is a leftover name from refactor and should be `optionalApiCommand`, isn't it?

topolyai5: yes, you are right, i will fix it

common/src/main/java/com/sequenceiq/cloudbreak/validation/FunctionChecked.java
topolyai5: bad package, I would create in util. I prefer the function first because to find easily by another one

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -187,32 +187,40 @@ private Boolean isPrewarmed(Long clusterId) {
...
-                restartServicesCommand = optionalRestartCommand.get();
-                LOGGER.debug("Restart for stale services is already running with id: [{}]", restartServicesCommand.getId());
-            } else {
-                restartServicesCommand = clustersResourceApi.restartCommand(stack.getName(),
-                        new ApiRestartClusterArgs().redeployClientConfiguration(Boolean.TRUE).restartOnlyStaleServices(Boolean.TRUE));
-            }
-            pollRestart(restartServicesCommand);
-            LOGGER.debug("Restarted stale services in Cloudera Manager.");
+        boolean clientConfigStale = services.stream().anyMatch(service -> !ApiConfigStalenessStatus.FRESH.equals(service.getClientConfigStalenessStatus()));
+        if (configStale || clientConfigStale) {
```
gregito: what's behind the `configStale` and the `clientConfigStale`? the actual values come from a specific logic but the variable name does not reflect it. let's say something like `serviceContainsStatusWithNotFreshStaleness`.. or other.. it's much longer but a good variable name could make unnecessary to understand the stream-filter line(s) above.

also, do we have a unit-test for this logic somewhere? if not please make some.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -238,6 +246,26 @@ private void pollRestart(ApiCommand restartCommand) throws CloudbreakException {
         }
     }
 
+    private void pollDeployConfig(ApiCommand restartCommand) throws CloudbreakException {
```
gregito: please cover this method with unit test(s)

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -238,6 +246,26 @@ private void pollRestart(ApiCommand restartCommand) throws CloudbreakException {
...
+        PollingResult hostTemplatePollingResult = clouderaManagerPollingServiceProvider.startPollingCmClientConfigDeployment(
+                stack, apiClient, restartCommand.getId());
+        if (isExited(hostTemplatePollingResult)) {
+            throw new CancellationException("Cluster was terminated while waiting for config deploy");
+        } else if (isTimeout(hostTemplatePollingResult)) {
+            throw new CloudbreakException("Timeout while Cloudera Manager was config deploying services.");
+        }
+    }
+
+    private void pollRefresh(ApiCommand restartCommand) throws CloudbreakException {
```
gregito: please cover this method with unit test(s)

common/src/main/java/com/sequenceiq/cloudbreak/validation/FunctionChecked.java
attilapalfi92: But this name does not make sense. You can find it anyway.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
topolyai5: I think this variable name too long. The other hand it just a boolean flag, not a list of services with not fresh staleness.

gregito: surely it's too long, I didn't say that you have to use my suggestion, but if you read this condition for the first time you must read the assignation line to understand what the variable actually represents. that's why it's a bad choice of name.

PR #7308, opened 2020-02-17
CB-5387 Hive Kudu TServer link and Kudu master link from API/gateway …

keyki:


adembo:
I'm late to the party, but this looks good to me.

PR #7311, opened 2020-02-18
CB-5604 Update SDX internal E2E test to 7.1.0 template

horadla23:


horadla23:
tested with InternalSdxTests

horadla23:
and it choosed 9a3dd513-61fd-4999-6d3b-37531b42ff03 which is 7.1.0 prewarmed image from v3-dev, created yesterday, so it looks good

PR #7314, opened 2020-02-18
CB-5363 Fix MDC context in image selection

foldik:


core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -423,6 +422,8 @@ private Blueprint determineBlueprint(StackV4Request stackRequest, Workspace work
                     baseImageEnabled,
                     user,
                     image -> true);
+            MDCBuilder.cleanupMdc();
```
horadla23: please check if this cleanup is needed

foldik: Yes we need this because we use a thread pool and other totally unrelated tasks will run on the same thread in the future.

PR #7319, opened 2020-02-18
CB-5365 Cut mock integration jobs on PRs into two

nadamdb:


doktoric:
I think it would be beneficial to separate the test by package like .mock.* and .perf.*

integration-test/src/main/resources/testsuites/v4/mock/mock-performance-tests.yaml
```diff
@@ -0,0 +1,12 @@
+name: "mock-performance-tests"
+tests:
+  - name: "newway mock performance testcases"
+    classes:
+      - com.sequenceiq.it.cloudbreak.testcase.mock.DistroXClusterStopStartTest
+      - com.sequenceiq.it.cloudbreak.testcase.mock.EnvironmentStartStopTest
+      - com.sequenceiq.it.cloudbreak.testcase.mock.KerberosConfigTest
+      - com.sequenceiq.it.cloudbreak.testcase.mock.RecipeClusterTest
+      - com.sequenceiq.it.cloudbreak.testcase.mock.DistroXClusterCreationTest
+      - com.sequenceiq.it.cloudbreak.testcase.mock.MockSdxTests
```
foldik: What is the reason to put the `MockSdxTests` into the performance related bucket?

integration-test/src/main/resources/testsuites/v4/mock/mock-performance-tests.yaml
```diff
@@ -0,0 +1,12 @@
+name: "mock-performance-tests"
+tests:
+  - name: "newway mock performance testcases"
+    classes:
+      - com.sequenceiq.it.cloudbreak.testcase.mock.DistroXClusterStopStartTest
```
foldik: As far as I can tell the `DistroXClusterStopStartTest` is the longest test. Previously it took 17 minutes and caused 13GB traffic between CB and Postgres from the overall 14.3GB. The duration went to 9-10 minutes fr this test and the traffic was reduced to 4.1GB in this particular test and the overall traffic dropped to 5.2GB? What was the reason or a certain threshold to put these tests into a different bucket?

nadamdb: The goal was to cut the overall duration of the tests into two separate test groups that have roughly the same running time. The tests were grouped based on the average duration of the different tests during previous runs. The tests in the performance related bucket take 27-30 minutes and the other tests take 22-25 minutes on average.

foldik: Currently the overall test execution takes around 15-20 minutes. Other tasks before the tests take 15-20 minutes and the cleanup takes 10-15 minutes. Based on the previous data I think we wouldn't get too much performance improvement in PR time by splitting this job and it would take another Jenkins executor, more instability, more things to keep in our heads.

PR #7321, opened 2020-02-19
CB-5197 moved test blueprints under test folder so classpath will be …

lnardai:
Initially I wanted to create the validation in the @PostConstruct part but I was afraid that it could cause issues during prod deployment where you can have different configurations hence the unit test level validation.

Example test error message:
`java.lang.AssertionError: Template { 7.1.0 - Data Engineering for Azure } has invalid instance type { raspberrypi } in hostgroup { master } added to it `

Also changed the parsing in the Value annotation as it was causing unnecessary whitespaces and trimming.


lnardai:
Let's wait for a review from the Wolf team as well.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -153,10 +153,10 @@
     @Value("${cb.aws.vm.parameter.definition.path:}")
     private String awsVmParameterDefinitionPath;
 
-    @Value("#{'${cb.aws.disabled.instance.types:}'.split(',')}")
+    @Value("${${cb.aws.disabled.instance.types:}")
```
topolyai5: hm, is it valid?

lnardai: Yes, I think this is a leftover from the previous format of the instance type which was the following:

`      disabled.instance.types: >
                              a1,
                              cr1,cc2,c3,c5d,c5n,c5.x,c5.24x,c1,c4,c5.m,c5.l,c5.12,c5.18,
                              d2,
                              f1,
                              g2,g3,
                              hs1,h1.16,
                              i3en,i3.l,i3.m,i3.x,i3.16,i2,
                              m1,m2,m3,m5d,m5ad,m4,m5a,m5.x,m5.m,m5.l,m5.12,m5.16,m5.24,
                              p3dn,p2.x,p3.16,
                              r3,r4,r5d.l,r5d.m,r5d.x,r5d.2x,r5d.12,r5d.24,r5d.8,r5d.16,r5a,r5.l,r5.x,r5.m,r5.24,r5.12,r5.16,
                              t1,t2,t3,
                              x1,
                              z1d`

It was an array where every line contained multiple entries, that's why you needed the split. Now there is only one entry in each line. I only noticed the additional whitespaces in the unit tests and was annoyed by it.

lnardai: Thanks I did not notify the double $ mark at first. Corrected that as well.

PR #7323, opened 2020-02-19
CB-5548 ensure that structured events are proper;y written into logs,…

akanto:
… since entities are not always jsonseriasible therfore it is better to log the StructuredEvent instead StructuredEventEntity


doktoric:
can you please fix the commit message?

PR #7324, opened 2020-02-19
CB-5395 Fix FreeIpa delete/detach conditions

Bajzathd:
CB-5395 Fix FreeIpa delete/detach conditions


environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/freeipa/FreeIpaDeletionHandler.java
```diff
@@ -56,24 +55,12 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
         EnvironmentDto environmentDto = environmentDtoEvent.getData();
         Optional<Environment> envOptional = environmentService.findEnvironmentById(environmentDto.getId());
         try {
-            if (envOptional.isPresent() && envOptional.get().isCreateFreeIpa() && freeIpaExistsForEnvironment(envOptional.get())) {
+            if (envOptional.isPresent() && freeIpaExistsForEnvironment(envOptional.get())) {
```
biharitomi: I think we should use the deleted condition in these cases because `envOptional.get().isCreateFreeIpa()` is dedicated to reduce the number of calls on FreeIPA service as FreeIPA creation hasn't been requested then we have no business on FreeIPA service side

PR #7328, opened 2020-02-19
CB-5643 Periscope is unable to call FreeIPA endpoints with internal CRN

anmolnar:
As part of HBase Autoscaling initiative we came across a bug in Periscope: LDAP and Kerberos services haven't been prepared to accept Internal CRNs which is a must for Periscope to work properly.

This patch enables the required endpoint for Internal usage.

No tests added yet, but recommendations are welcome.

Closes CB-5643


anmolnar:
Moving to 2.18

PR #7329, opened 2020-02-19
CB-5293 CM integration is needed for Autoscaling

anmolnar:
Enables Periscope to retrieve `getTotalHosts` from a CM cluster. Previous it was Ambari-only, but this patch adds another level of abstraction which makes the call transparent for the caller and supports both cluster managers.

No tests added, but recommendations are welcome.


anmolnar:
Moving to 2.18

PR #7332, opened 2020-02-19
CB-5636. E2E tests: ABFS logs are missing from idbroker host.

oleewere:
Pass idbroker identity for fluentd configs (not the log identity one), technically multiple identities should be assigned, but as we are not doing that just use the idbroker one

- UTs added
- few permission fixes (databus credentials)

cluster testing is in progress

### DIFF
--- a/orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
+++ b/orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
@@ -36,6 +36,11 @@
 {% set s3_log_bucket = salt['pillar.get']('fluent:s3LogArchiveBucketName') %}
 {% set azure_container = salt['pillar.get']('fluent:azureContainer') %}
 {% set azure_storage_instance_msi = salt['pillar.get']('fluent:azureInstanceMsi') %}
+{% if salt['pillar.get']('fluent:azureIdBrokerInstanceMsi') %}
+    {% set azure_storage_idbroker_instance_msi = salt['pillar.get']('fluent:azureIdBrokerInstanceMsi') %}
+{% else %}
+    {% set azure_storage_idbroker_instance_msi = salt['pillar.get']('fluent:azureInstanceMsi') %}
+{% endif %}
 {% set azore_storage_account = salt['pillar.get']('fluent:azureStorageAccount') %}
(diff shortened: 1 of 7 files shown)


orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
```diff
@@ -36,6 +36,8 @@
 {% set s3_log_bucket = salt['pillar.get']('fluent:s3LogArchiveBucketName') %}
 {% set azure_container = salt['pillar.get']('fluent:azureContainer') %}
 {% set azure_storage_instance_msi = salt['pillar.get']('fluent:azureInstanceMsi') %}
+{% set azure_storage_instance_msi = salt['pillar.get']('fluent:azureIdBrokerInstanceMsi') %}
```
lajosrodek: Hmm, this is certainly wrong as it overwrites the value just retrieved in the previous line. 😄

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/output.conf.j2
```diff
@@ -56,7 +56,11 @@
     azure_storage_account    {{fluent.azureStorageAccount}}
     azure_container          {{fluent.azureContainer}}
 {% if fluent.azureInstanceMsi is defined and fluent.azureInstanceMsi is not none and fluent.azureInstanceMsi %}
+    {% if grains['hostgroup'] == "idbroker" %}
+    azure_instance_msi       {{fluent.azureIdBrokerInstanceMsi}}
```
lajosrodek: At first sight this did not make sense to me as log shipping normally does not work on the IDBroker node (see [CB-5107](https://jira.cloudera.com/browse/CB-5107)). I guess this setting only works in case of internal E2E testing when the IDBroker VM MSI (i.e. the assumer identity) also contains the necessary role assignment for accessing the log storage account location. And similarly for AWS, the IDBroker EC2 instance profile role should have a policy for accessing the log S3 bucket.

oleewere: just to answer that here as well, it's better now to fail during log shipping (because of the lack of proper contributor roles), then fail during obtaining the azure access token (because of wrong vm identity). some would happen for AWS if idbroker instance profile does not have the right policies

PR #7338, opened 2020-02-20
CB-5547 added notification for user on whether repair can start or not

lnardai:
- added new graceful state when we can't start the repair
- based on sodre's branch added notification so user is aware that no repair was started


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterRepairService.java
```diff
@@ -362,12 +364,14 @@ private void triggerRepairOrThrowBadRequest(Long stackId, Result<Map<HostGroupNa
             boolean removeOnly,
             Set<String> recoveryMessageArgument) {
         if (repairValidationResult.isError()) {
+            eventService.fireCloudbreakEvent(stackId, RECOVERY, CLUSTER_MANUALRECOVERY_COULD_NOT_START, repairValidationResult.getError().getValidationErrors());
```
foldik: Can you extend the unit test as well?

lnardai: On hold because Unit test are broken.

lnardai: Added unit tests.

core/src/test/java/com/sequenceiq/cloudbreak/service/cluster/ClusterRepairServiceTest.java
```diff
@@ -237,6 +241,7 @@ public void testCannotRepairGatewayWithoutExternalDatabase() throws CloudbreakIm
         boolean result = underTest.canRepairAll(stack);
 
         assertFalse(result);
+        //eventService.fireCloudbreakEvent(stackId, RECOVERY, CLUSTER_MANUALRECOVERY_COULD_NOT_START, repairValidationResult.getError().getValidationErrors());
```
foldik: Do we need that in the future?

PR #7345, opened 2020-02-20
CDPCP-1291. Add support for retrieving a keytab in FMS

aarman-cloudera:
This updates FreeIPA service to include a new endpoint that allows retrieving a keytab for a user or machine user under a provided environment.

### DIFF
--- a/freeipa/build.gradle
+++ b/freeipa/build.gradle
@@ -50,6 +50,9 @@ dependencies {
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jaxrs2',                       version: opentracingJaxrs2Version
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jdbc',                         version: opentracingJdbcVersion
 
+
+  implementation     group: 'org.apache.kerby',          name: 'kerb-util',                            version: '2.0.0'
+  
   implementation     group: 'com.github.briandilley.jsonrpc4j', name: 'jsonrpc4j',                     version: '1.5.3'
   testImplementation ('org.springframework.boot:spring-boot-starter-test') {
     exclude group: 'junit'
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/KrbKeySetEncoder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/KrbKeySetEncoder.java
@@ -1,4 +1,4 @@
-package com.sequenceiq.freeipa.util;
+package com.sequenceiq.freeipa.service.freeipa.user.kerberos;
 
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ActorKerberosKey;
 
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/UserKeytabGenerator.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/UserKeytabGenerator.java
@@ -1,0 +1,75 @@
+package com.sequenceiq.freeipa.service.freeipa.user.kerberos;
+
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ActorKerberosKey;
+import com.sequenceiq.cloudbreak.common.service.Clock;
+import org.apache.kerby.kerberos.kerb.keytab.Keytab;
+import org.apache.kerby.kerberos.kerb.keytab.KeytabEntry;
+import org.apache.kerby.kerberos.kerb.type.KerberosTime;
+import org.apache.kerby.kerberos.kerb.type.base.EncryptionKey;
+import org.apache.kerby.kerberos.kerb.type.base.EncryptionType;
+import org.apache.kerby.kerberos.kerb.type.base.PrincipalName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Base64;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Kerberos keytab generator - This class is repsonsible for generating a keytab using existing keys
+ * belonging to an actor in User Management Service.
+ */
+@Component
+public final class UserKeytabGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserKeytabGenerator.class);
+
+    // For the key version number (kvno) in the keytab entries, we use a value of 0. Based on
+    // expirementiation we've found that for user principles, keytab entries with kvno of zero
+    // works fine, regardless of what the actual key versions are set in the KDC / Directory Server.
+    private static final int KEY_VERSION_NUMBER_ZERO = 0;
+
+    @Inject
+    private Clock clock;
+
+    private KeytabEntry toKeytabEntry(String user, String realm, ActorKerberosKey actorKerberosKey) {
+        requireNonNull(user, "user must not be null");
+        requireNonNull(realm, "realm must not be null");
+        requireNonNull(actorKerberosKey, "actorKerberosKey must not be null");
+
+        PrincipalName principalName = new PrincipalName(user + '@' + realm);
+
+        EncryptionType encryptionType = EncryptionType.fromValue(actorKerberosKey.getKeyType());
+        byte[] key = Base64.getDecoder().decode(actorKerberosKey.getKeyValue());
+        EncryptionKey encryptionKey = new EncryptionKey(encryptionType, key);
+
+        KerberosTime time = new KerberosTime(clock.getCurrentTimeMillis());
+
+        return new KeytabEntry(principalName, time, KEY_VERSION_NUMBER_ZERO, encryptionKey);
+    }
+
+    public String generateKeytabBase64(String username, String realm, List<ActorKerberosKey> actorKerberosKeys) throws IOException {
+        LOGGER.info("Generating keytab for username = {} with realm = {}", username, realm);
+
+        if (actorKerberosKeys.isEmpty()) {
+            throw new IllegalArgumentException("Expected at least 1 actorKerberosKeys");
+        }
+
+        List<KeytabEntry> keytabEntries = actorKerberosKeys.stream()
+                .map(key -> toKeytabEntry(username, realm, key))
+                .collect(Collectors.toList());
+        Keytab keytab = new Keytab();
+        keytab.addKeytabEntries(keytabEntries);
+        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
+            keytab.store(outputStream);
+            byte[] keyBytes = outputStream.toByteArray();
+            return Base64.getEncoder().encodeToString(keyBytes);
+        }
+    }
+}
--- a/auth-connector/src/main/proto/usermanagement.proto
+++ b/auth-connector/src/main/proto/usermanagement.proto
@@ -2292,6 +2292,8 @@ message GetActorWorkloadCredentialsResponse {
   uint64 passwordHashExpirationDate = 3;
   // A list of kerberos keys for the actor.
   repeated ActorKerberosKey kerberosKeys = 2;
+  // The workload username for the actor.
+  string workloadUsername = 4;
 }
 
 message GetEventGenerationIdsRequest {
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/KerberosMgmtV1Endpoint.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/KerberosMgmtV1Endpoint.java
@@ -60,6 +60,13 @@ public interface KerberosMgmtV1Endpoint {
             notes = KeytabModelNotes.GET_HOST_KEYTAB_NOTES, nickname = "getHostKeytabV1")
     HostKeytabResponse getHostKeytab(@Valid HostKeytabRequest request) throws Exception;
 
+    @GET
+    @Path("userkeytab")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = KeytabOperationsDescription.DESCRIBE_USER_KEYTAB, produces = MediaType.APPLICATION_JSON,
+            notes = KeytabModelNotes.GET_USER_KEYTAB_NOTES, nickname = "getUserKeytabV1")
+    String getUserKeytab(@NotEmpty @QueryParam("environmentCrn") String environmentCrn, @NotEmpty @QueryParam("userCrn") String userCrn) throws Exception;
+
     @DELETE
     @Path("serviceprincipal")
     @Produces(MediaType.APPLICATION_JSON)
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/doc/KeytabModelNotes.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/doc/KeytabModelNotes.java
@@ -12,6 +12,8 @@ public class KeytabModelNotes {
             + "the keytab already generated for that principal invalid. The keytab in the response is base64 encoded.";
     public static final String GET_HOST_KEYTAB_NOTES = "Retrieves the existing keytab for the host provided. Gets the existing keytab without modification "
             + "and not effecting the prior keytab. The keytab in the response is base64 encoded.";
+    public static final String GET_USER_KEYTAB_NOTES = "Retrieves the existing keytab for the user provided. Gets the existing keytab without modification "
+            + "and not effecting the prior keytab. The keytab in the response is base64 encoded.";
     public static final String DELETE_SERVICE_PRINCIPAL_NOTES = "Deletes the principal from the FreeIPA. It also deletes vault secrets associated with the "
             + "principal. If a role is specified, it will also be deleted. However, if the role is still in use, it will not be deleted in a normal scenerio.";
     public static final String DELETE_HOST_NOTES = "Deletes the host and all the principals associated with the host in the FreeIPA. It also deletes vault "
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/doc/KeytabOperationsDescription.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/kerberosmgmt/doc/KeytabOperationsDescription.java
@@ -6,11 +6,11 @@ public class KeytabOperationsDescription {
     public static final String DESCRIBE_SERVICE_KEYTAB = "Get the keytab for the provided service on a specific host";
     public static final String DESCRIBE_GENERATE_HOST_KEYTAB = "Create the host and then get the keytab for the provided host";
     public static final String DESCRIBE_HOST_KEYTAB = "Get the keytab for the provided host";
+    public static final String DESCRIBE_USER_KEYTAB = "Get the keytab for the provided user";
     public static final String DESCRIBE_DELETE_SERVICE_PRINCIPAL = "Delete the service principal";
     public static final String DESCRIBE_DELETE_HOST = "Delete the host";
     public static final String DESCRIBE_CLUSTER_CLEANUP = "Cleanup the secrets associated with the cluster";
     public static final String DESCRIBE_ENVIRONMENT_CLEANUP = "Cleanup all the secrets associated with the environment";
-
 
     private KeytabOperationsDescription() {
     }
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
@@ -3,7 +3,12 @@ package com.sequenceiq.freeipa.kerberosmgmt.v1;
 import javax.inject.Inject;
 import javax.transaction.Transactional;
 import javax.validation.Valid;
+import javax.validation.constraints.NotEmpty;
 
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
+import com.sequenceiq.freeipa.controller.exception.BadRequestException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Controller;
 
 import com.sequenceiq.authorization.annotation.CheckPermissionByAccount;
@@ -26,11 +31,17 @@ import com.sequenceiq.freeipa.util.CrnService;
 @Transactional(Transactional.TxType.NEVER)
 @AuthorizationResource(type = AuthorizationResourceType.ENVIRONMENT)
 public class KerberosMgmtV1Controller implements KerberosMgmtV1Endpoint {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KerberosMgmtV1Controller.class);
+
     @Inject
     private KerberosMgmtV1Service kerberosMgmtV1Service;
 
     @Inject
     private CrnService crnService;
+
+    @Inject
+    private UserKeytabService userKeytabService;
 
     @CheckPermissionByAccount(action = AuthorizationResourceAction.WRITE)
     public ServiceKeytabResponse generateServiceKeytab(@Valid ServiceKeytabRequest request) throws FreeIpaClientException {
@@ -79,4 +90,26 @@ public class KerberosMgmtV1Controller implements KerberosMgmtV1Endpoint {
         String accountId = crnService.getCurrentAccountId();
         kerberosMgmtV1Service.cleanupByEnvironment(environmentCrn, accountId);
     }
+
+    @Override
+    public String getUserKeytab(@NotEmpty String environmentCrn, @NotEmpty String targetUserCrn) {
+        String actorCrn = checkActorCrn();
+        LOGGER.debug("getUserKeytab() request for environmentCrn={} for targetUserCrn={} as actorCrn={}",
+                environmentCrn, actorCrn, targetUserCrn);
+        // TODO: For now we only allow retrieving keytab for the calling user. When we enable calling this against
+        //       other users we need to ensure the caller has the neccessary authorization / rights.
+        if (!actorCrn.equals(targetUserCrn)) {
+            throw new BadRequestException("Retrieving a keytab for another user is unsupported");
+        }
+        return userKeytabService.getKeytabBase64(targetUserCrn, environmentCrn);
+    }
+
+    private String checkActorCrn() {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        if (userCrn == null) {
+            throw new BadRequestException("An actor CRN must be provided");
+        }
+        return userCrn;
+    }
+
 }
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabGenerator.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabGenerator.java
@@ -1,4 +1,4 @@
-package com.sequenceiq.freeipa.service.freeipa.user.kerberos;
+package com.sequenceiq.freeipa.kerberosmgmt.v1;
 
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ActorKerberosKey;
 import com.sequenceiq.cloudbreak.common.service.Clock;
@@ -26,13 +26,14 @@ import static java.util.Objects.requireNonNull;
  * belonging to an actor in User Management Service.
  */
 @Component
-public final class UserKeytabGenerator {
+public class UserKeytabGenerator {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UserKeytabGenerator.class);
 
     // For the key version number (kvno) in the keytab entries, we use a value of 0. Based on
     // expirementiation we've found that for user principles, keytab entries with kvno of zero
-    // works fine, regardless of what the actual key versions are set in the KDC / Directory Server.
+    // works fine (in fact, any arbitrary number should work), regardless of what the actual key
+    // versions are set in the KDC / Directory Server.
     private static final int KEY_VERSION_NUMBER_ZERO = 0;
 
     @Inject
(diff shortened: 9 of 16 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
+package com.sequenceiq.freeipa.util;
```
handavid: I'm not too fond of the location of this file. I'd prefer it to reside closer to the service that will actually use it.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static final byte[] KEYTAB_VERSION_0X502 = { (byte) 0x05, (byte) 0x02 };
+
+    private static final byte KEY_VERSION_NUMBER_ZERO = 0;
+
+    @Inject
+    private Clock clock;
+
+    private KeytabEntry toKeytabEntry(String user, String realm, ActorKerberosKey actorKerberosKey) {
+        String principalName = user + '@' + realm;
```
giladwolff: verify/assert parameters?

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+@Component
+public final class KrbKeytabGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KrbKeytabGenerator.class);
+
+    private static final int PRINCIPAL_TYPE_1 = 1;
+
+    private static final byte[] KEYTAB_VERSION_0X502 = { (byte) 0x05, (byte) 0x02 };
+
+    private static final byte KEY_VERSION_NUMBER_ZERO = 0;
```
giladwolff: This is the kvno? Add a comment?

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Component
+public final class KrbKeytabGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KrbKeytabGenerator.class);
+
+    private static final int PRINCIPAL_TYPE_1 = 1;
```
giladwolff: What is this principal type?

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+        byte[] key = Base64.getDecoder().decode(actorKerberosKey.getKeyValue());
+        EncryptionKey encryptionKey = new EncryptionKey(encryptionType, key);
+
+        KerberosTime time = new KerberosTime(clock.getCurrentTimeMillis());
+
+        return new KeytabEntry(principalName, PRINCIPAL_TYPE_1, time, KEY_VERSION_NUMBER_ZERO, encryptionKey);
+    }
+
+    public String generateKeytabBase64(String workloadUserName, String realm, List<ActorKerberosKey> actorKerberosKeys) {
+        LOGGER.info("Generating keytab for workloadUserName = {} with realm = {}", workloadUserName, realm);
```
giladwolff: verify/assert parameters?

freeipa/build.gradle
```diff
@@ -51,6 +51,9 @@ dependencies {
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jaxrs2',                       version: opentracingJaxrs2Version
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jdbc',                         version: opentracingJdbcVersion
 
+  implementation     group: 'org.apache.directory.api',  name: 'api-asn1-api',                         version: '2.0.0'
+  implementation     group: 'org.apache.directory.api',  name: 'api-asn1-ber',                         version: '2.0.0'
+
```
giladwolff: So what is this needed for? You copied the code, no?

freeipa/src/main/java/org/apache/directory/README.md
```diff
@@ -0,0 +1,12 @@
+This packge contains classes to aid in the generation of Kerberos Keytabs. The classes here are taken
```
giladwolff: I would just drop the org/apache/directory from the package and move it to somewhere where the rest of you code lives (KeytabUtils?) You can put the readme and the classes there.

freeipa/src/test/java/com/sequenceiq/freeipa/util/KrbKeyTabGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
```
giladwolff: Are there any meaningful tests you want to bring from apache.directory?

freeipa/src/main/java/org/apache/directory/shared/kerberos/components/EncryptionKey.java
```diff
@@ -0,0 +1,330 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
```
giladwolff: So what is the apache license? Can we copy the code and use it/modify it as long as we have the copyright statement?

giladwolff: I think I answered my own question by doing a git grep on thunderhead git repo. I think it's fine.

freeipa/src/test/java/com/sequenceiq/freeipa/util/KrbKeyTabGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final long MOCK_TIME = 1576171731141L;
+
+    @InjectMocks
+    KrbKeytabGenerator underTest;
+
+    @Mock
+    private Clock clock;
+
+    @Test
+    void testGenerateKeytabBase64() {
```
lacikaaa: it looks like to me there is only 1 happy path tested. Are you sure there isn't anything which can go wrong here? Like a parameter is wrong

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
aarman-cloudera: Ah sure let me find a better place (I've placed it here because KrbKeySetEncoder is here which plays an analogous role). I think we want to expose this as part of UserV1Endpoint and maybe add a UserKeytabGenService.

lacikaaa: that class is in a wrong package also IMHO

freeipa/src/main/java/org/apache/directory/README.md
aarman-cloudera: Yes, the entire Apache DS library is a fairly heavy dependency. I've just looked into Apache Kerby which Kevin Risden mentioned to me earlier. There's a 'kerb-util' artifact that we can use which allows creating keytab. That may be more reasonable to pull in so we don't have to copy over source code. Let me try that.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
handavid: I gree that KrbKeySetEncoder is also in the wrong spot. I suggest com.sequenceiq.freeipa.service.freeipa.user.krb or com.sequenceiq.freeipa.service.freeipa.user.util

aarman-cloudera: Thanks, I'll move both of them.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
aarman-cloudera: This is goes away in my upcoming change that uses kerby-util package. Internally it will still set a name type of 1 meaning NT-PRINCIPAL, which is what we want I believe.

I'm not sure how important this field is either ways, as per https://tools.ietf.org/html/rfc4120

"The name-type SHOULD be treated only as a hint to interpreting the meaning of a name.  It is not significant when checking for equivalence. Principal names that differ only in the name-type identify the same principal."

freeipa/build.gradle
aarman-cloudera: This is no longer needed with my upcoming change.

freeipa/src/test/java/com/sequenceiq/freeipa/util/KrbKeyTabGeneratorTest.java
aarman-cloudera: No longer needed.

freeipa/src/main/java/org/apache/directory/shared/kerberos/components/EncryptionKey.java
aarman-cloudera: Cool. I think all we need to do is just include the original license, and if we've modified the code we just put some Cloudera modification header.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+import java.util.Base64;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Component
+public final class KrbKeytabGenerator {
```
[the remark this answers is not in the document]
aarman-cloudera: The functionality will be exposed as a public endpoint in a follow-on commit.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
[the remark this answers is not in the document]
aarman-cloudera: Next commit gets rid of some of these variables (internally they still get these values). I've added a comment about key version number.

freeipa/src/test/java/com/sequenceiq/freeipa/util/KrbKeyTabGeneratorTest.java
aarman-cloudera: I think there's some more validation (i.e. has to be a valid workload user) we can add, but it might be at a higher level.

freeipa/src/main/java/com/sequenceiq/freeipa/util/KrbKeytabGenerator.java
aarman-cloudera: Added

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Kerberos keytab generator - This class is repsonsible for generating a keytab using existing keys
+ * belonging to an actor in User Management Service.
+ */
+@Component
+public final class KrbKeytabGenerator {
```
lacikaaa: Not a big deal, but could you drop `krb` in favor of `kerberos`? I think the package should be `user.kerberos` and the class name `UserKeytabGenerator`. This would reflect more of it's purpose in my opinion.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeytabGenerator.java
```diff
@@ -0,0 +1,70 @@
...
+        EncryptionType encryptionType = EncryptionType.fromValue(actorKerberosKey.getKeyType());
+        byte[] key = Base64.getDecoder().decode(actorKerberosKey.getKeyValue());
+        EncryptionKey encryptionKey = new EncryptionKey(encryptionType, key);
+
+        KerberosTime time = new KerberosTime(clock.getCurrentTimeMillis());
+
+        return new KeytabEntry(principalName, time, KEY_VERSION_NUMBER_ZERO, encryptionKey);
+    }
+
+    public String generateKeytabBase64(String username, String realm, List<ActorKerberosKey> actorKerberosKeys) throws IOException {
```
lacikaaa: I see one `try` without catch, so my concern is if something goes wrong in this code, maybe we would like to convert it to a custom exception a propagate it upwards in that format. What do you think about it?

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeyTabGeneratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.io.IOException;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class KrbKeyTabGeneratorTest {
```
lacikaaa: I think we need test for those non null checks at least.
Could you come up with any scenario where other part of the code breaks? I mean a lot static calls are used so this is not easy to test, but what if `ActorKerberosKey` holds wrong value, or the username holds the whole principal. I think we should cover some unhappy path here.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeytabGenerator.java
aarman-cloudera: That catch for this would be an IOException. I highly doubt we hit the exception here since it's not actually doing file / network IO. I have added a check for this in the service that calls this function, see updated PR.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeytabGenerator.java
aarman-cloudera: Done.

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/user/krb/KrbKeyTabGeneratorTest.java
aarman-cloudera: Added some null tests and a test for empty key list. As for the other unhappy paths, I don't think it's worth testing at this layer. A user won't provide any of these inputs directly. We pull the keys and username off of UMS and the realm will be pulled from kerberos config. I've updated the PR with the actual service that uses this component. Please see updated PR.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -98,6 +102,19 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    @Override
+    public String getKeytab(@NotNull String environmentCrn, String targetUserCrn) {
+        String callerUserCrn = checkUserCrn();
+        LOGGER.debug("getKeytab() request for environmentCrn={}, callerUserCrn={}, targetUserCrn={}", environmentCrn, callerUserCrn, targetUserCrn);
+        // TODO For now, we ignore the optional targetUserCrn and only allow retrieving keytab for the calling user. When
+        //      we later enable this, we need to make sure that the calling has the necessary authorization / right to get
```
aarman-cloudera: 'the calling has' -> 'the calling user has'

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/UserV1Endpoint.java
```diff
@@ -52,4 +52,11 @@
     @ApiOperation(value = UserOperationDescriptions.SYNC_OPERATION_STATUS, notes = UserNotes.USER_NOTES, produces = MediaType.APPLICATION_JSON,
             nickname = "getSyncOperationStatusV1")
     SyncOperationStatus getSyncOperationStatus(@NotNull @QueryParam("operationId") String operationId);
+
+    @GET
+    @Path("getKeytab")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = UserOperationDescriptions.GET_KEYTAB, notes = UserNotes.USER_NOTES, produces = MediaType.APPLICATION_JSON,
+            nickname = "getKeytabV1")
+    String getKeytab(@NotNull @QueryParam("environment") String environment, @QueryParam("user") String user);
```
lacikaaa: I think this should be in `KerberosMgmtV1Endpoint`. What do you think @handavid ?

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -98,6 +102,19 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
...
+        //      keytab for another user.
+        if (targetUserCrn != null) {
+            throw new BadRequestException("Retrieving a keytab for another user is not yet supported");
+        }
```
lacikaaa: We might take of that `targetUserCrn` from the API and add it when we support it, as it won't be a breaking change. If we keep this way I think we should not throw an exception when target and caller crn are the same.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/UserV1Endpoint.java
lacikaaa: @jamisonbennett

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
aarman-cloudera: Good point, I've updated to not throw when target is same.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/UserV1Endpoint.java
aarman-cloudera: Since this endpoint is tied to user in ums, I think this place is fine too. Happy to change if I'm outvoted though :)

handavid: I'll defer to Jamison here regarding keytab generation. Based on his comments, I think that this API probably belongs next to the other keytab APIs

handavid: if the query params are CRNs, then let's include CRN in the names. e.g., environmentCrn

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -107,6 +111,19 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    @Override
+    public String getKeytab(@NotNull String environmentCrn, String targetUserCrn) {
```
handavid: based on the expected future use-case, we should require the targetUserCrn field. i.e., mark it @NotNull.
As a design principle, we do not want to pull request arguments out of the actor or other metadata. All required information should be in the request itself. (yes, a bunch of the other APIs violate this but I'll fix this in the v2 API)

can we rename callerUserCrn to actorCrn?
we can keep the actorCrn.equals(targetUserCrn) check as a faux authorization check for now.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserKeytabService.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Objects;
+
+import static com.sequenceiq.freeipa.controller.exception.NotFoundException.notFound;
+
+@Service
+public class UserKeytabService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserKeytabService.class);
+
+    private static final String IAM_INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
```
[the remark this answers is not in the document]
handavid: we had a talk ;)

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -107,6 +111,19 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
...
+        String callerUserCrn = checkUserCrn();
+        LOGGER.debug("getKeytab() request for environmentCrn={}, callerUserCrn={}, targetUserCrn={}", environmentCrn, callerUserCrn, targetUserCrn);
+        // TODO For now, we ignore the optional targetUserCrn and only allow retrieving keytab for the calling user. When
+        //      we later enable this, we need to make sure that the calling user has the necessary authorization / right to get
+        //      keytab for another user.
+        if (targetUserCrn != null && !callerUserCrn.equals(targetUserCrn)) {
+            throw new BadRequestException("Retrieving a keytab for another user is not yet supported");
+        }
+        return userKeytabService.getKeytabBase64(callerUserCrn, environmentCrn);
```
handavid: call this with the targetUserCrn rather than the actorCrn. The actorCrn will be available to other code in through the ThreadBasedUserCrnProvider (I think I remembered that correctly)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserKeytabService.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private String getKerberosRealm(String accountId, String environmentCrn) {
+        KerberosConfig krbConfig =  kerberosConfigRepository
+                .findByAccountIdAndEnvironmentCrnAndClusterNameIsNull(accountId, environmentCrn)
+                .orElseThrow(notFound("KerberosConfig for environment", environmentCrn));
+        return krbConfig.getRealm();
+    }
+
+    private void validateSameAccount(String userAccountId, String environmentCrn) {
+        String environmentCrnAccountId = Objects.requireNonNull(Crn.fromString(environmentCrn)).getAccountId();
```
handavid: Crn.safeFromString instead of Objects.requireNonNull

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserKeytabService.java
```diff
@@ -0,0 +1,72 @@
...
+        if (!environmentCrnAccountId.equals(userAccountId)) {
+            throw new BadRequestException("User and Environment must be in the same account");
+        }
+    }
+
+    public String getKeytabBase64(String userCrn, String environmentCrn) {
+        String userAccountId = Objects.requireNonNull(Crn.fromString(userCrn)).getAccountId();
```
handavid: Crn.safeFromString instead of Objects.requireNonNull

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserKeytabService.java
```diff
@@ -0,0 +1,72 @@
...
+
+        GetActorWorkloadCredentialsResponse getActorWorkloadCredentialsResponse =
+                grpcUmsClient.getActorWorkloadCredentials(IAM_INTERNAL_ACTOR_CRN, userCrn, MDCUtils.getRequestId());
+        String workloadUsername = getActorWorkloadCredentialsResponse.getWorkloadUsername();
+        List<ActorKerberosKey> actorKerberosKeys = getActorWorkloadCredentialsResponse.getKerberosKeysList();
+
+        try {
+            return userKeytabGenerator.generateKeytabBase64(workloadUsername, realm, actorKerberosKeys);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to generate keytab", e);
```
handavid: can you throw a more meaningful exception? perhaps make a custom exception so that we can map it to a response code in the api layer

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/KrbKeySetEncoder.java
```diff
@@ -1,4 +1,4 @@
-package com.sequenceiq.freeipa.util;
+package com.sequenceiq.freeipa.service.freeipa.user.kerberos;
```
handavid: thank you!

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/UserKeytabGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+ * belonging to an actor in User Management Service.
+ */
+@Component
+public class UserKeytabGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserKeytabGenerator.class);
+
+    // For the key version number (kvno) in the keytab entries, we use a value of 0. Based on
+    // expirementiation we've found that for user principles, keytab entries with kvno of zero
+    // works fine, regardless of what the actual key versions are set in the KDC / Directory Server.
```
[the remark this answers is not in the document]
toddlipcon: I think you might have already gotten the answer elsewhere, but yea, this builds on work that Gilad (I think) did in the UMS which can produce keytabs for users. That work, in turn, builds on the stuff we worked on a few months back where the UMS stores the hashed password and kerberos principal keys for users locally for password sync purposes. Given that the UMS has the kerberos principal keys, it's just a file formatting exercise to embed those keys into a keytab file, with no FreeIPA instance in the loop at all. (in fact this API should work even if there is no environment created)

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/UserV1Endpoint.java
giladwolff: We do not use the public FreeIPA APIs. We have the password hashes in the control plane and we just package them in a keytab format. We use kvno 0 which works with an arbitrary kvno in the kdc. Asif did a POC before sending this out.

aarman-cloudera: Thanks for looking Jamison. I've moved it to KerberosMgmtV1Endpoint As mentioned this work is heavily tied with 'syncing password hash' feature that others have worked on in the past. Some more details about this approach can be found on the design doc here:

https://docs.google.com/document/d/1-M9-WD0SGBNMsnJbZtXTMmrh0EHRG8tGVYUlIR64byg

Can we use a string and not a vault path, for now at least? This will be exposed as a public endpoint in environments2 api service in thunderhead and we don't have vault integration there yet. I can file jira to look into it, but I'd rather do it as a follow up.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/doc/UserOperationDescriptions.java
```diff
@@ -5,6 +5,7 @@
     public static final String SYNC_ALL = "Synchronizes Groups and Users to the FreeIPA servers";
     public static final String SET_PASSWORD = "Sets the user's password in the FreeIPA servers";
     public static final String SYNC_OPERATION_STATUS = "Gets the status of a sync operation";
+    public static final String GET_KEYTAB = "Gets a keytab as a base64 encoded string for a User under a specified Environment";
```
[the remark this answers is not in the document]
aarman-cloudera: As mentioned, I'd like to not go with vault path for now and just return the base64 string directly.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/UserKeytabGenerator.java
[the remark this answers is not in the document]
aarman-cloudera: Just a bit more info, my original comment on kvno = 0 here maybe a bit misleading. As per my testing, the key version number didn't matter and I don't think 0 is special. Setting this to an arbitrary number like 42 would probably still work (though a bulk of my testing was using kvno = 0).

Using FreeIPA API is get keytab for user principles, generates new keys and updates the KDC with the new keys and increments the key version number (even if you use the same password; new random salts get used, resulting in different keys). So the older keytab getting invalidated is due to keys not matching and nothing to do with key version numbers. As Todd mentioned since we have keys already in UMS we use those as the basis for the keytab.

Another thing to note, based on above; any keytab a user gets using this api will still get invalidated if the user were to later call ipa-getkeytab directly (but would start working again after user sync occurs). So we should discourage any user from using this ipa-getkeytab or trying to get the keytab by other means outside the api we provide.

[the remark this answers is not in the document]
aarman-cloudera: Also, this API still requires an environment, because we need the realm inside the keytab. In order to generate the keytab, we perform a lookup of the realm based on environment.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
aarman-cloudera: Done. I've also renamed to actorCrn (I wonder if we should rename ThreadBasedUserCrnProvider -> ThreadBasedActorrCrnProvider).

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserKeytabService.java
aarman-cloudera: I've moved the exception handling into UserKeytabGenerator. I don't see how the IOException will ever trigger since we're passing our own ByteArrayOutputStream rather than doing file i/o. So I don't think a custom exception here is warranted.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/kerberos/KrbKeySetEncoder.java
aarman-cloudera: np! :)

PR #7346, opened 2020-02-21
CB-5642 Put flow into runningFlows before save flow to the DB

sodre90:


flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
```diff
@@ -109,10 +109,16 @@ private void handle(String key, Payload payload, FlowParameters flowParameters,
...
                         flow.initialize(contextParams);
-                        flowLogService.save(flowParameters, flowChainId, key, payload, null, flowConfig.getClass(), flow.getCurrentState());
+                        runningFlows.put(flow, flowChainId);
+                        try {
+                            flowLogService.save(flowParameters, flowChainId, key, payload, null, flowConfig.getClass(), flow.getCurrentState());
+                        } catch (Exception e) {
+                            LOGGER.error("Can't save flow: {}", flowId);
+                            runningFlows.remove(flowId);
+                            throw e;
+                        }
```
topolyai5: Could you please write few unit tests?

flow/src/main/java/com/sequenceiq/cloudbreak/service/ha/HeartbeatService.java
```diff
@@ -119,12 +119,14 @@ public void heartbeat() {
     @Scheduled(initialDelay = 35000L, fixedDelay = 30000L)
     public void scheduledFlowDistribution() {
         if (shouldRun()) {
```
foldik: This method says nothing about when it should run. Can you please rename it or just inline the underlying implementation which is so simple:

```java
nodeConfig.isNodeIdSpecified()
```

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
sodre90: Done! Thanks

PR #7353, opened 2020-02-21
CB-5616 [CB+UI] Distrox in running state when all the nodes are stopped

bbihari:


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/StackSyncService.java
```diff
@@ -217,6 +222,7 @@ private void handleSyncResult(Stack stack, Map<InstanceSyncState, Integer> insta
             eventService.fireCloudbreakEvent(stack.getId(), AVAILABLE.name(), STACK_SYNC_INSTANCE_STATE_SYNCED);
         } else if (isAllStopped(instanceStateCounts, instances.size()) && stack.getStatus() != STOPPED) {
             updateStackStatusIfEnabled(stack.getId(), DetailedStackStatus.STOPPED, SYNC_STATUS_REASON, stackStatusUpdateEnabled);
+            updateClusterStatusIfEnabled(stack.getId(), STOPPED, stackStatusUpdateEnabled);
```
topolyai5: you should update the cluster status in the other case as well

bbihari: I don't think so

topolyai5: We talked in slack. The cluster status will be set at the other part of the sync. I accept the change

PR #7356, opened 2020-02-21
CB-5604 Distrox E2E creation fails in case of long stackname and 7.1.0 template

horadla23:


horadla23:
Cluster with longer name fails with Hue error:
Screenshot 2020-02-21 at 16 18 28

PR #7359, opened 2020-02-21
Fix a type error in the README.md

cegganesh84:
1. manager is misspelled in the example URL.
2. Added a note about significance of trailing slash in the URL.


doktoric:
thanks @cegganesh84

PR #7361, opened 2020-02-22
CB-5688 - Cannot stop environment if there is no attached data lake

gregito:
it's merely a workaround for an edge case when the deletion timestamp has not set at the end of SDX cluster deletion. we weren't able to reproduce the root cause but with this, the designated environment could be stopped even if the timestamp is missing, and makes the handling of this deleted state more flexible.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/repository/SdxStatusRepository.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/repository/SdxStatusRepository.java
@@ -2,7 +2,9 @@ package com.sequenceiq.datalake.repository;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.stereotype.Repository;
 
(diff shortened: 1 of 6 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -413,4 +416,14 @@ private String getAccountIdFromCrn(String userCrn) {
     private DetailedEnvironmentResponse getEnvironment(String environmentName) {
         return environmentClientService.getByName(environmentName);
     }
+
+    private List<SdxCluster> filterOutClustersWithDeletedStatus(List<SdxCluster> sdxClusters) {
```
lnardai: This will just hide the original problem, I think we should rather fix the source of the problem.
@sodre90 has on his branch an implementation that will make the Status change and Deleted timestamp addition so this won't happen again.

For fixing the already stuck clusters we should create an SQL script.
I see there are also some additional unit tests, those can be useful.

Talk with @sodre90 please.

datalake/src/main/java/com/sequenceiq/datalake/repository/SdxStatusRepository.java
```diff
@@ -25,4 +27,8 @@
     List<SdxStatusEntity> findDistinctFirstByStatusInAndDatalakeIdInOrderByIdDesc(Collection<DatalakeStatusEnum> datalakeStatusEnums,
             Collection<Long> datalakeId);
 
+    @CheckPermission(action = ResourceAction.READ)
+    @Query("SELECT sdxstatus.datalake.id FROM SdxStatusEntity sdxstatus WHERE sdxstatus.status IN ('DELETED') GROUP BY sdxstatus.datalake.id")
+    Set<Long> findAllSdxClusterIdWhichHasDeletedState();
```
sodre90: It is just a workaround and we need to fix the original problem. When we set datalake status to deleted, then we update the deleted timestamp, but they are not in one transaction.

PR #7362, opened 2020-02-24
CB-1141 & CB-5159: Add high availability options to FreeIPA

jamisonbennett:
Simplified AWS FreeIPA cloud formation template. Some properties were
made optional in the template builder so that unused FreeIPA
properties could be removed. All FreeIPA nodes are gateway nodes, so
the FreeIPA CORE node option was removed.

Fixed the HA logic for the number of processes to use for krb5kdc.

Added a salt hardfail rule when installing FreeIPA.

Added an optional environment property called instancesPerGroup to
specify the number of HA instances of FreeIPA. This property was also
added to the database table.

Rename master group to master0 so that it will accommodate upgrades to
multi AZ in the future which will be named master0, master1, etc.

Added validation checks to allow for up to 4 instances.

Tested manually with a local deployment of cloudbreak by deploying an
HA cluster in Azure and AWS.

Closes #CB-1141
Closes #CB-5159


cloud-aws/src/main/resources/templates/aws-cf-stack-freeipa.ftl
```diff
@@ -0,0 +1,236 @@
...
+    <#list instanceGroups as group>
+    "${group.autoScalingGroupName}" : {
+      "Type" : "AWS::AutoScaling::AutoScalingGroup",
+      "Properties" : {
+        <#if group.subnetId??>
+        "VPCZoneIdentifier" : [ "${group.subnetId}" ],
+        <#else>
+        "VPCZoneIdentifier" : [{ "Ref" : "SubnetId" }],
+        </#if>
+        "LaunchConfigurationName" : { "Ref" : "AmbariNodeLaunchConfig${group.groupName?replace('_', '')}" },
```
lacikaaa: Could you rename `AmbariNodeLaunchConfig`? Or is there a reason to keep it?

environment/src/main/resources/schema/app/20200323193238_CB-5159_freeipa_instances_per_group_is_optional.sql
```diff
@@ -0,0 +1,10 @@
+-- // CB-1617 freeipa creation is optional
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE environment ADD COLUMN IF NOT EXISTS freeIpaInstancesPerGroup integer NOT NULL DEFAULT 1;
```
lacikaaa: please don't use `DEFAULT`, because it can lock the table. instead you can use an update

freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
```diff
@@ -27,9 +27,27 @@
 
     @Override
     public ValidationResult validate(CreateFreeIpaRequest subject) {
+        // FreeIPA's recommends no more than 4 replication agreements per node. To simply things, only allow a fully connected mesh.
+        // Also don't allow a special case with 5 nodes for 1 group when the there is a maximum of 4 instances when using multiple groups.
+        final int maxInstances = 4;
+
+        final int maxInstanceGroups = 1;
```
lacikaaa: I think these should be class level constants, or variables configurable via env parameters, so we can override them

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSetup.java
```diff
@@ -128,7 +128,8 @@ public void prerequisites(AuthenticatedContext ac, CloudStack stack, Persistence
     private void validateRegionAndZone(CloudCredential cloudCredential, Location location) {
         CloudRegions regions = awsPlatformResources.regions(cloudCredential, location.getRegion(), Collections.emptyMap());
         List<AvailabilityZone> availabilityZones = regions.getCloudRegions().get(location.getRegion());
-        if (!availabilityZones.contains(location.getAvailabilityZone())) {
+        if (location.getAvailabilityZone() != null
```
doktoric: is it possible that the az is null here?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
```diff
@@ -98,7 +100,8 @@ public DeleteStackRequest createDeleteStackRequest(String cFStackName) {
         if (awsInstanceProfileView.isInstanceProfileAvailable()) {
             parameters.add(new Parameter().withParameterKey("InstanceProfile").withParameterValue(awsInstanceProfileView.getInstanceProfile()));
         }
-        if (ac.getCloudContext().getLocation().getAvailabilityZone().value() != null) {
+        if (ac.getCloudContext().getLocation().getAvailabilityZone() != null
```
doktoric: is this possible that the ac.getCloudContext().getLocation().getAvailabilityZone()  null ?

cloud-aws/src/main/resources/templates/aws-cf-stack-freeipa.ftl
doktoric: unfortunately if we change this here then we should take care with the backward compatibility

cloud-aws/src/main/resources/templates/aws-cf-stack-freeipa.ftl
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+
+    <#if group.cloudSecurityIds?size == 0>,
+    "ClusterNodeSecurityGroup${group.groupName?replace('_', '')}" : {
+      "Type" : "AWS::EC2::SecurityGroup",
+      "Properties" : {
+        "GroupDescription" : "Allow access from web and bastion as well as outbound HTTP and HTTPS traffic",
+        "Tags" : [
+            { "Key" : "cb-resource-type", "Value" : "${securitygroup_resource}" },
```
doktoric: cb-resource-type owner and Owner should be removed these are not required anymore

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/AttachedFreeIpaRequest.java
```diff
@@ -14,6 +14,9 @@
     @ApiModelProperty(value = EnvironmentModelDescription.CREATE_FREEIPA, required = true)
     private Boolean create;
 
+    @ApiModelProperty(value = EnvironmentModelDescription.FREEIPA_INSTANCES_PER_GROUP)
+    private Integer instancesPerGroup;
```
doktoric: instanceCountByGroup or instanceCountPerGroup ?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentBaseResponse.java
```diff
@@ -32,6 +32,9 @@
     @ApiModelProperty(EnvironmentModelDescription.CREATE_FREEIPA)
     private Boolean createFreeIpa = Boolean.TRUE;
 
+    @ApiModelProperty(EnvironmentModelDescription.FREEIPA_INSTANCES_PER_GROUP)
+    private Integer freeIpaInstancesPerGroup = 1;
```
doktoric: I think we should create a FreeipaResponse Object here because probably we will see more and more property here

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/SimpleEnvironmentResponse.java
```diff
@@ -37,6 +37,8 @@ public static Builder builder() {
 
         private boolean createFreeIpa;
 
+        private int freeIpaInstancesPerGroup;
```
doktoric: I think we should create a FreeipaResponse Object here because probably we will see more and more property here

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentCreationDto.java
```diff
@@ -148,8 +148,8 @@ public EnvironmentTelemetry getTelemetry() {
         return telemetry;
     }
 
-    public boolean isCreateFreeIpa() {
-        return createFreeIpa;
+    public FreeIpaCreationDto getFreeIpaCreation() {
```
doktoric: thanks for created this object here 👍

freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
```diff
@@ -27,9 +27,27 @@
...
+
         ValidationResultBuilder validationBuilder = ValidationResult.builder();
-        if (CollectionUtils.isEmpty(subject.getInstanceGroups()) || subject.getInstanceGroups().size() != 1) {
-            validationBuilder.error("Stack request must contain a single instance group.");
+        if (CollectionUtils.isEmpty(subject.getInstanceGroups())) {
+            validationBuilder.error("Stack request must contain at least one instance group.");
+        } else {
+            int nodesPerInstanceGroup = subject.getInstanceGroups().get(0).getNodeCount();
+            if (subject.getInstanceGroups().stream().filter(ig -> ig.getNodeCount() != nodesPerInstanceGroup || ig.getNodeCount() < 1).count() > 0) {
+                validationBuilder.error("All instance groups in the stack request must contain the same number of nodes per instance group " +
```
doktoric: in the freeipa request

freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
```diff
@@ -27,9 +27,27 @@
...
+                        "and there must be at least 1 instance per instance group.");
+            }
+            if (nodesPerInstanceGroup * subject.getInstanceGroups().size() > maxInstances) {
+                validationBuilder.error(String.format("Stack request must contain at most %d instances.", maxInstances));
```
doktoric: I think it shoud be good not to call stack request but freeipa request here

freeipa/src/main/java/com/sequenceiq/freeipa/converter/image/ImageConverter.java
```diff
@@ -16,8 +16,7 @@
     public com.sequenceiq.cloudbreak.cloud.model.Image convert(Image source) {
         com.sequenceiq.cloudbreak.cloud.model.Image image =
                 new com.sequenceiq.cloudbreak.cloud.model.Image(source.getImageName(),
-                        Map.of(InstanceGroupType.GATEWAY, Optional.ofNullable(source.getUserdata()).orElse(""),
-                                InstanceGroupType.CORE, Optional.ofNullable(source.getUserdata()).orElse("")),
+                        Map.of(InstanceGroupType.GATEWAY, Optional.ofNullable(source.getUserdata()).orElse("")),
```
doktoric: do we really want to remove the CORE part here ?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/provision/action/AbstractStackProvisionAction.java
```diff
@@ -49,7 +48,7 @@ protected AbstractStackProvisionAction(Class<P> payloadClass) {
     protected StackContext createFlowContext(FlowParameters flowParameters, StateContext<StackProvisionState, StackProvisionEvent> stateContext, P payload) {
         Stack stack = stackService.getByIdWithListsInTransaction(payload.getResourceId());
         MDCBuilder.buildMdcContext(stack);
-        Location location = location(region(stack.getRegion()), availabilityZone(stack.getAvailabilityZone()));
```
doktoric: why we removed this part ?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSetup.java
jamisonbennett: The FreeIPA cloud formation template was simplified. The AZs were needed when creating subnets which only cloudbreak core does. As a result it is no longer needed in the cloud formation template for FreeIPA. In order for the template to process I had to avoid sending the AZ parameter into the template processor. But I needed to only do this for FreeIPA and not core. As a result, I changed FreeIPA to not set the AZ since it is not used that way the common code could check if it was null and optionally use the AZ parameter.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackRequestHelper.java
jamisonbennett: Yes, for the same reason as above.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/SimpleEnvironmentResponse.java
jamisonbennett: Done, but I am keeping the createFreeIpa in the parent class for API compatibility.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/image/ImageConverter.java
jamisonbennett: This was done to simplify the FreeIPA templates. I discussed this with @lacikaaa and @keyki

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/provision/action/AbstractStackProvisionAction.java
jamisonbennett: It was replaced with `Location location = location(region(stack.getRegion()));` which does not set the AZ.

The FreeIPA cloud formation template was simplified. The AZs were needed when creating subnets which only cloudbreak core does. As a result it is no longer needed in the cloud formation template for FreeIPA. In order for the template to process I had to avoid sending the AZ parameter into the template processor. But I needed to only do this for FreeIPA and not core. As a result, I changed FreeIPA to not set the AZ since it is not used that way the common code could check if it was null and optionally use the AZ parameter.

jamisonbennett:
I have addressed all comments except for the following which I will hopefully get to tomorrow:
1. adding an e2e test that does a provision using an old version, upgrades cbd, then performs a stop/start.
2. adding an e2e test that provisions an HA cluster

lacikaaa:
we might skip that e2e test, as we don't have anything we use the template for. we just have to be careful when adding scaling features that we don't support those with old templates

freeipa/src/main/java/com/sequenceiq/freeipa/converter/image/ImageConverter.java
doktoric: 👍

environment/src/main/java/com/sequenceiq/environment/environment/v1/FreeIpaConverter.java
```diff
@@ -0,0 +1,19 @@
...
+@Component
+public class FreeIpaConverter {
+
+    public FreeIpaResponse convert(FreeIpaCreationDto freeIpaCreation) {
+        if (freeIpaCreation == null) {
+            return null;
+        }
+        FreeIpaResponse response = new FreeIpaResponse();
+        response.setInstanceCountByGroup(freeIpaCreation.getInstanceCountByGroup());
+        return response;
```
lacikaaa: put this part into an `else` please

jamisonbennett:
Should the integration test be added to this repo or to the integration test repo?

lacikaaa:
int tests are under the `integration-test` module

PR #7366, opened 2020-02-24
CB-5708 Remove cloud storage object listing from suite logs

aszegedi:
@lnardai please review.


lnardai:
Every log line was trashing the log output?

aszegedi:
> I'm fine with it, although it seems like a lot of logging got removed.
> Are all of these useless?

I'm implementing a similar solution for Azure as we have for AWS. So on this way we can see a summary about the Blob content and do not need to flip through all of this object in the test suite logs.

PR #7376, opened 2020-02-25
CB-5254: Add ability to reboot a single FreeIPA instance.

holleyism:
This patch adds the APIs and flows to support requesting
a reboot of a single instance of a FreeIPA cluster.

https://github.com/hortonworks/cloudbreak/commit/7bedab08331049d704829df0b1a6bfc9c9e6f5cf?diff=split#diff-bb2a2c4ffdd122b890ccd439056e72c4

https://github.com/hortonworks/cloudbreak/commit/7bedab08331049d704829df0b1a6bfc9c9e6f5cf?diff=split#diff-a17be58a8a607f24bfc72b427ea2b4ec

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpInstanceConnector.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpInstanceConnector.java
@@ -33,6 +33,11 @@ public class GcpInstanceConnector extends AbstractInstanceConnector {
     private boolean verifyHostKey;
 
     @Override
+    public List<CloudVmInstanceStatus> reboot(AuthenticatedContext authenticatedContext, List<CloudInstance> vms) {
+        throw new CloudOperationNotSupportedException("Reboot instances operation is not supported on GCP");
+    }
+
+    @Override
     public List<CloudVmInstanceStatus> check(AuthenticatedContext ac, List<CloudInstance> vms) {
         List<CloudVmInstanceStatus> statuses = new ArrayList<>();
         CloudCredential credential = ac.getCloudCredential();
--- a/cloud-openstack/src/main/java/com/sequenceiq/cloudbreak/cloud/openstack/nativ/OpenStackNativeInstanceConnector.java
+++ b/cloud-openstack/src/main/java/com/sequenceiq/cloudbreak/cloud/openstack/nativ/OpenStackNativeInstanceConnector.java
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
+import com.sequenceiq.cloudbreak.cloud.exception.CloudOperationNotSupportedException;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
 import com.sequenceiq.cloudbreak.cloud.model.CloudVmInstanceStatus;
 import com.sequenceiq.cloudbreak.cloud.model.InstanceStatus;
@@ -27,6 +28,11 @@ public class OpenStackNativeInstanceConnector extends AbstractInstanceConnector 
 
     @Inject
     private OpenStackClient openStackClient;
+
+    @Override
+    public List<CloudVmInstanceStatus> reboot(AuthenticatedContext authenticatedContext, List<CloudInstance> vms) {
+        throw new CloudOperationNotSupportedException("Reboot instances operation is not supported on OpenStack");
+    }
 
     @Override
     public List<CloudVmInstanceStatus> check(AuthenticatedContext ac, List<CloudInstance> vms) {
--- a/cloud-reactor-api/src/main/java/com/sequenceiq/cloudbreak/cloud/event/instance/RebootInstancesRequest.java
+++ b/cloud-reactor-api/src/main/java/com/sequenceiq/cloudbreak/cloud/event/instance/RebootInstancesRequest.java
@@ -1,0 +1,29 @@
+package com.sequenceiq.cloudbreak.cloud.event.instance;
+
+import java.util.List;
+
+import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
+import com.sequenceiq.cloudbreak.cloud.event.CloudPlatformRequest;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
+import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
+
+public class RebootInstancesRequest<T> extends CloudPlatformRequest<T> {
+
+    private final List<CloudInstance> cloudInstances;
+
+    public RebootInstancesRequest(CloudContext cloudContext, CloudCredential cloudCredential,
+            List<CloudInstance> cloudInstances) {
+        super(cloudContext, cloudCredential);
+        this.cloudInstances = cloudInstances;
+    }
+
+    public List<CloudInstance> getCloudInstances() {
+        return cloudInstances;
+    }
+
+    @Override
+    public String toString() {
+        super.toString();
+        return "RebootInstancesRequest{cloudInstances=" + cloudInstances + "}";
+    }
+}
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaOperationDescriptions.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaOperationDescriptions.java
@@ -14,6 +14,7 @@ public final class FreeIpaOperationDescriptions {
     public static final String REGISTER_WITH_CLUSTER_PROXY = "Registers FreeIPA stack with given environment CRN with cluster proxy";
     public static final String DEREGISTER_WITH_CLUSTER_PROXY = "Deregisters FreeIPA stack with given environment CRN with cluster proxy";
     public static final String HEALTH = "Provides a detailed health of the FreeIPA stack";
+    public static final String REBOOT = "Reboot one or more instances";
 
     private FreeIpaOperationDescriptions() {
     }
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
@@ -1,0 +1,51 @@
+package com.sequenceiq.freeipa.api.v1.freeipa.stack.model.reboot;
+
+import java.util.List;
+
+import javax.validation.constraints.NotEmpty;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.sequenceiq.service.api.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel("RebootInstancesV1Request")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RebootInstancesRequest {
+    @NotEmpty
+    @ApiModelProperty(value = ModelDescriptions.ENVIRONMENT_CRN, required = true)
+    private String environmentCrn;
+
+    @ApiModelProperty(value = ModelDescriptions.FORCE_REBOOT, required = false)
+    private boolean forceReboot;
+
+    @ApiModelProperty(value = ModelDescriptions.INSTANCE_ID, required = false)
+    private List<String> instanceIds;
+
+    public String getEnvironmentCrn() {
+        return environmentCrn;
+    }
+
+    public void setEnvironmentCrn(String environmentCrn) {
+        this.environmentCrn = environmentCrn;
+    }
+
+    public boolean isForceReboot() {
+        return forceReboot;
+    }
+
+    public void setForceReboot(boolean forceReboot) {
+        this.forceReboot = forceReboot;
+    }
+
+    public List<String> getInstanceIds() {
+        return instanceIds;
+    }
+
+    public void setInstanceIds(List<String> instanceIds) {
+        this.instanceIds = instanceIds;
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/RebootService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/RebootService.java
@@ -1,0 +1,37 @@
+package com.sequenceiq.freeipa.flow.instance.reboot;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.event.instance.RebootInstancesResult;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
+import com.sequenceiq.freeipa.service.stack.RebootInstancesService;
+import com.sequenceiq.freeipa.service.stack.instance.InstanceMetaDataService;
+
+@Component
+public class RebootService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RebootService.class);
+
+    @Inject
+    private InstanceMetaDataService instanceMetaDataService;
+
+    @Inject
+    private RebootInstancesService rebootInstancesService;
+
+    public void startInstanceReboot(RebootContext context) {
+        instanceMetaDataService.updateStatus(context.getStack(), context.getInstanceIdList(), InstanceStatus.REBOOTING);
+    }
+
+    public void handleInstanceRebootError(RebootContext context) {
+        instanceMetaDataService.updateStatus(context.getStack(), context.getInstanceIdList(), InstanceStatus.FAILED);
+        LOGGER.error("Reboot failed for instances: {}", context.getInstanceIds());
+    }
+
+    public void finishInstanceReboot(RebootContext context, RebootInstancesResult rebootInstancesResult) {
+        instanceMetaDataService.updateStatus(context.getStack(), context.getInstanceIdList(), InstanceStatus.CREATED);
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
@@ -1,0 +1,20 @@
+package com.sequenceiq.freeipa.flow.instance.reboot.action;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.cloudbreak.common.event.Payload;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootContext;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootEvent;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootState;
+import com.sequenceiq.freeipa.flow.stack.AbstractStackAction;
+
+public abstract class AbstractRebootAction<P extends Payload>
+        extends AbstractStackAction<RebootState, RebootEvent, RebootContext, P>  {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractRebootAction.class);
+
+    protected AbstractRebootAction(Class<P> payloadClass) {
+        super(payloadClass);
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
@@ -1,0 +1,190 @@
+package com.sequenceiq.freeipa.flow.instance.reboot.action;
+
+import static com.sequenceiq.cloudbreak.cloud.model.AvailabilityZone.availabilityZone;
+import static com.sequenceiq.cloudbreak.cloud.model.Location.location;
+import static com.sequenceiq.cloudbreak.cloud.model.Region.region;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.convert.ConversionService;
+import org.springframework.statemachine.StateContext;
+import org.springframework.statemachine.action.Action;
+
+import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
+import com.sequenceiq.cloudbreak.cloud.event.instance.RebootInstancesRequest;
+import com.sequenceiq.cloudbreak.cloud.event.instance.RebootInstancesResult;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
+import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
+import com.sequenceiq.cloudbreak.cloud.model.Location;
+import com.sequenceiq.cloudbreak.common.event.Selectable;
+import com.sequenceiq.cloudbreak.logger.MDCBuilder;
+import com.sequenceiq.flow.core.FlowParameters;
+import com.sequenceiq.flow.core.PayloadConverter;
+import com.sequenceiq.freeipa.converter.cloud.CredentialToCloudCredentialConverter;
+import com.sequenceiq.freeipa.converter.cloud.InstanceMetaDataToCloudInstanceConverter;
+import com.sequenceiq.freeipa.dto.Credential;
+import com.sequenceiq.freeipa.entity.InstanceMetaData;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.flow.freeipa.cleanup.event.failure.RebootInstancesResultToCleanupFailureEventConverter;
+import com.sequenceiq.freeipa.flow.instance.InstanceEvent;
+import com.sequenceiq.freeipa.flow.instance.InstanceFailureEvent;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootContext;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootEvent;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootService;
+import com.sequenceiq.freeipa.flow.instance.reboot.RebootState;
+import com.sequenceiq.freeipa.repository.InstanceMetaDataRepository;
+import com.sequenceiq.freeipa.service.CredentialService;
+import com.sequenceiq.freeipa.service.stack.StackService;
+import com.sequenceiq.freeipa.service.stack.instance.InstanceMetaDataService;
+
+@Configuration
+public class RebootActions {
+    private static final Logger LOGGER = LoggerFactory.getLogger(RebootActions.class);
+
+    @Inject
+    private RebootService rebootService;
+
+    @Inject
+    private InstanceMetaDataToCloudInstanceConverter metadataConverter;
+
+    @Inject
+    private CredentialService credentialService;
+
+    @Inject
+    private StackService stackService;
+
+    @Inject
+    private InstanceMetaDataService instanceMetaDataService;
+
+    @Inject
+    private InstanceMetaDataRepository instanceMetaDataRepository;
+
+    @Inject
+    private CredentialToCloudCredentialConverter credentialConverter;
+
+    @Inject
+    @Qualifier("conversionService")
+    private ConversionService conversionService;
+
+    @Bean(name = "REBOOT_STATE")
+    public Action<?, ?> rebootAction() {
+        return new AbstractRebootAction<>(InstanceEvent.class) {
+            @Override
+            protected void doExecute(RebootContext context, InstanceEvent payload, Map<Object, Object> variables) {
+                rebootService.startInstanceReboot(context);
+                sendEvent(context);
+                LOGGER.info("Starting reboot for {}", context.getInstanceIds());
+            }
+
+            @Override
+            protected Object getFailurePayload(InstanceEvent payload, Optional<RebootContext> flowContext, Exception ex) {
+                return new InstanceFailureEvent(payload.getResourceId(), ex, payload.getInstanceIds());
+            }
+
+            @Override
+            protected Selectable createRequest(RebootContext context) {
+                List<CloudInstance> cloudInstances = context.getInstanceMetaDataList().stream().map(instanceMetaData ->
+                        conversionService.convert(instanceMetaData, CloudInstance.class)).collect(Collectors.toList());
+                return new RebootInstancesRequest<>(context.getCloudContext(), context.getCloudCredential(), cloudInstances);
+            }
+
+            @Override
+            protected RebootContext createFlowContext(FlowParameters flowParameters, StateContext<RebootState, RebootEvent> stateContext,
+                    InstanceEvent payload) {
+                Long stackId = payload.getResourceId();
+                Stack stack = stackService.getStackById(stackId);
+                List<InstanceMetaData> instances = instanceMetaDataRepository.findAllInStack(stackId).stream()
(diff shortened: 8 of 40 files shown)


cloud-reactor-api/src/main/java/com/sequenceiq/cloudbreak/cloud/event/instance/RebootInstancesRequest.java
```diff
@@ -0,0 +1,31 @@
...
+        this.cloudInstances = cloudInstances;
+    }
+
+    public List<CloudInstance> getCloudInstances() {
+        return cloudInstances;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder("RebootInstancesRequest{");
```
lacikaaa: Stringbuilder is overkill here I think. I know there is some class where it's implemented that way, but I don't think that's a good pattern. Also inherited classes should include `super.toString()` or we can loose useful data

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaOperationDescriptions.java
```diff
@@ -14,6 +14,8 @@
     public static final String REGISTER_WITH_CLUSTER_PROXY = "Registers FreeIPA stack with given environment CRN with cluster proxy";
     public static final String DEREGISTER_WITH_CLUSTER_PROXY = "Deregisters FreeIPA stack with given environment CRN with cluster proxy";
     public static final String HEALTH = "Provides a detailed health of the FreeIPA stack";
+    public static final String REBOOT = "Reboot one or more instances";
+
```
lacikaaa: extra empty line

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.sequenceiq.service.api.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel("RebootInstancesV1Request")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RebootInstancesRequest {
+    @NotNull
```
lacikaaa: maybe `@NotEmpty` is better here

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+    private Map<String, InstanceMetaData> getAllInstancesMap(Stack stack) {
+        return stack.getInstanceGroups().stream().flatMap(instanceGroup -> instanceGroup.getInstanceMetaData().stream())
+                .collect(Collectors.toMap(InstanceMetaData::getInstanceId, Function.identity()));
+    }
+
+    public void rebootInstances(String accountId, RebootInstancesRequest request) {
+        // If no instance passed in request, reboot all bad instances
+        // If instances passed in request, reboot all valid passed bad instances
+        // If force and instances passed in request, reboot all valid passed instances
+        // If force and no instances passed in request, reboot all instances
```
lacikaaa: if we need these comments please convert this into a javadoc for the method

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+        Stack stack = stackService.getByEnvironmentCrnAndAccountIdWithLists(request.getEnvironmentCrn(), accountId);
+        MDCBuilder.buildMdcContext(stack);
+        Map<String, InstanceMetaData> allInstances = getAllInstancesMap(stack);
+        Set<String> invalidInstanceIds = getInvalidInstances(allInstances.keySet(), request);
+        if (!invalidInstanceIds.isEmpty()) {
+            String msg = MessageFormat.format("Invalid instanceIds are present in request: {0}.", String.join(", ", invalidInstanceIds));
+            LOGGER.error(msg);
+            throw new BadRequestException(msg);
+        }
```
lacikaaa: please move into a `validate` method

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+
+        Map<String, InstanceStatus> health = getInstanceHealth(accountId, request);
+
+        Map<String, InstanceMetaData> instancesToReboot = getInstancesToReboot(allInstances, request.getInstanceIds(), health);
+        for (Map.Entry<String, InstanceMetaData> instance: instancesToReboot.entrySet()) {
+            instanceMetaDataService.updateStatus(stack, instance.getKey(), InstanceStatus.REBOOTING);
+            flowManager.notify(REBOOT_EVENT.event(), new InstanceEvent(REBOOT_EVENT.event(), instance.getValue().getId()));
+        }
```
lacikaaa: why do we start a new flow for every instance? I think we should start 1 flow with all the instance IDs which we would like to restart.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
```diff
@@ -0,0 +1,71 @@
...
+        super(payloadClass);
+    }
+
+    @Override
+    protected RebootContext createFlowContext(FlowParameters flowParameters, StateContext<RebootState, RebootEvent> stateContext, P payload) {
+        Long stackId = payload.getResourceId();
+        Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+        MDCBuilder.buildMdcContext(stack);
+
+        Optional<InstanceMetaData> instance = instanceMetaDataService.getById(payload.getResourceId());
```
lacikaaa: something definitely wrong here, you use the `payload.getResourceId()` for fetching the stack and the instance. It should be the stack, and I would rewrite this to fetch all instancemetadata you need. And if you need the instanceMetadata in only 1 flow step, maybe you should fetch it only in that step.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
```diff
@@ -0,0 +1,71 @@
...
+        Location location = location(region(stack.getRegion()), availabilityZone(stack.getAvailabilityZone()));
+        CloudContext cloudContext = new CloudContext(stack.getId(), stack.getName(), stack.getCloudPlatform(), stack.getCloudPlatform(),
+                location, stack.getOwner(), stack.getAccountId());
+        Credential credential = credentialService.getCredentialByEnvCrn(stack.getEnvironmentCrn());
+        CloudCredential cloudCredential = credentialConverter.convert(credential);
+
+        return new RebootContext(flowParameters, stack, instance.get(), cloudContext, cloudCredential);
```
lacikaaa: if the `instance.get()` failes here is pretty strange, I think we should fail earlier

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
```diff
@@ -0,0 +1,92 @@
...
+        };
+    }
+
+    @Bean(name = "REBOOT_FINISHED_STATE")
+    public Action<?, ?> rebootFinishedAction() {
+        return new AbstractRebootAction<>(RebootInstancesResult.class) {
+            @Override
+            protected void doExecute(RebootContext context, RebootInstancesResult payload, Map<Object, Object> variables) {
+                rebootService.finishInstanceReboot(context, payload);
+                LOGGER.info("Finished rebooting.");
```
lacikaaa: please add instanceids to the log, makes investigation easier

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/InstanceMetaDataService.java
```diff
@@ -42,6 +43,19 @@ public void saveInstanceRequests(Stack stack, List<Group> groups) {
...
+        Set<InstanceGroup> instanceGroups = stack.getInstanceGroups();
+        for (InstanceGroup group : instanceGroups) {
+            for (InstanceMetaData instanceMetaData : group.getInstanceMetaData()) {
+                if (instanceId.equals(instanceMetaData.getInstanceId())) {
+                    instanceMetaData.setInstanceStatus(status);
+                    instanceMetaDataRepository.save(instanceMetaData);
+                    return;
+                }
+            }
+        }
```
lacikaaa: stream usage seems reasonable here

lacikaaa: also a log for changing the status. and unit test please

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+                .peek(logit)
+                .filter(id -> !allInstances.contains(id))
+                .collect(Collectors.toSet());
+    }
+
+    private Map<String, InstanceMetaData> getInstancesToReboot(Map<String, InstanceMetaData> allInstances,
+            List<String> instanceIds, Map<String, InstanceStatus> health) {
+        Consumer<Map.Entry> logit = entry -> {
+            if (health.get(entry.getKey()) != null && health.get(entry.getKey()).isAvailable()) {
+                LOGGER.info("Not rebooting AVAILABLE instance {} because force reboot was not selected.", entry.getKey());
```
lacikaaa: please replace `AVAILABLE` string with the actual value

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+        };
+        return allInstances.entrySet().stream()
+                .peek(logit)
+                .filter(entry -> instanceIds == null || instanceIds.isEmpty() || instanceIds.contains(entry.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
lacikaaa: I think something is wrong here. You used the heath to filter what to log, but you don't use it to filter. In my opinion peek should just log and don't have any logic at all, this could be misleading

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+        if (!request.isForceReboot()) {
+            return healthDetailsService.getHealthDetails(request.getEnvironmentCrn(), accountId).getNodeHealthDetails().stream()
+                    .collect(Collectors.toMap(NodeHealthDetails::getInstanceId, NodeHealthDetails::getStatus));
+        }
+        return Collections.emptyMap();
+    }
+
+    private Set<String> getInvalidInstances(Collection<String> allInstances, RebootInstancesRequest request) {
+        Consumer<String> logit = id -> {
+            if (!allInstances.contains(id)) {
```
lacikaaa: similarly as in the other log, I don't think having logic in peek is a good idea, or at least move out the logic, so it's not duplicated. If somebody touches the filter and not the `logit` logic it will differ

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+    @Inject
+    private FreeIpaFlowManager flowManager;
+
+    @Inject
+    private FreeIpaHealthDetailsService healthDetailsService;
+
+    @Inject
+    private InstanceMetaDataService instanceMetaDataService;
+
+    private Map<String, InstanceStatus> getInstanceHealth(String accountId, RebootInstancesRequest request) {
```
lacikaaa: it's misleading you don't return health if it's forced. you should ignore health where you filter for instances to restart if it's forced, but this method doesn't exactly do what you expect by it's name

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public void rebootInstances(String accountId, RebootInstancesRequest request) {
+        // If no instance passed in request, reboot all bad instances
+        // If instances passed in request, reboot all valid passed bad instances
+        // If force and instances passed in request, reboot all valid passed instances
+        // If force and no instances passed in request, reboot all instances
+        Stack stack = stackService.getByEnvironmentCrnAndAccountIdWithLists(request.getEnvironmentCrn(), accountId);
+        MDCBuilder.buildMdcContext(stack);
+        Map<String, InstanceMetaData> allInstances = getAllInstancesMap(stack);
```
lacikaaa: could you rename the variable? like `allInstanceByInstanceId`, because I had to lookup the code to know what is in this map

freeipa/src/test/java/com/sequenceiq/freeipa/service/instance/RebootInstanceServiceTest.java
```diff
@@ -0,0 +1,149 @@
...
+import com.sequenceiq.freeipa.flow.instance.InstanceEvent;
+import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
+import com.sequenceiq.freeipa.service.freeipa.flow.FreeIpaFlowManager;
+import com.sequenceiq.freeipa.service.stack.FreeIpaHealthDetailsService;
+import com.sequenceiq.freeipa.service.stack.RebootInstancesService;
+import com.sequenceiq.freeipa.service.stack.StackService;
+import com.sequenceiq.freeipa.service.stack.instance.InstanceMetaDataService;
+
+@ExtendWith(MockitoExtension.class)
+public class RebootInstanceServiceTest {
```
lacikaaa: it looks like only 1 instance scenario is tested, I think more complex cases needs to be tested

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/operation/model/OperationType.java
```diff
@@ -5,6 +5,7 @@
 public enum OperationType {
     USER_SYNC,
     SET_PASSWORD,
+    REBOOT,
```
lacikaaa: where is this used?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstanceAcceptor.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.freeipa.api.v1.operation.model.OperationType;
+import com.sequenceiq.freeipa.entity.Operation;
+import com.sequenceiq.freeipa.repository.OperationRepository;
+import com.sequenceiq.freeipa.service.freeipa.user.AcceptResult;
+import com.sequenceiq.freeipa.service.operation.OperationAcceptor;
+
+@Component
+public class RebootInstanceAcceptor extends OperationAcceptor {
```
lacikaaa: I don't think this is used anywhere

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpInstanceConnector.java
```diff
@@ -32,6 +32,11 @@
     @Value("${cb.gcp.hostkey.verify:}")
     private boolean verifyHostKey;
 
+    @Override
+    public List<CloudVmInstanceStatus> reboot(AuthenticatedContext authenticatedContext, List<CloudInstance> vms) {
+        return null;
```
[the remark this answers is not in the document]
holleyism: fixed

cloud-openstack/src/main/java/com/sequenceiq/cloudbreak/cloud/openstack/nativ/OpenStackNativeInstanceConnector.java
```diff
@@ -28,6 +28,11 @@
     @Inject
     private OpenStackClient openStackClient;
 
+    @Override
+    public List<CloudVmInstanceStatus> reboot(AuthenticatedContext authenticatedContext, List<CloudInstance> vms) {
+        return null;
```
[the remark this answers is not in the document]
holleyism: fixed

cloud-reactor-api/src/main/java/com/sequenceiq/cloudbreak/cloud/event/instance/RebootInstancesRequest.java
holleyism: fixed

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaOperationDescriptions.java
holleyism: fixed

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: We eventually want the ability to do a rolling reboot of the servers instead of rebooting all at the same time.  My understanding is the way to accomplish this is by implementing a custom FlowEventChainFactory that can dynamically tie the events together so we'd need separate events to achieve that.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/operation/model/OperationType.java
holleyism: removed.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstanceAcceptor.java
holleyism: removed.

holleyism: removed.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/AbstractRebootAction.java
holleyism: Think I refactored this correctly.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: refactored to move logic out.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/InstanceMetaDataService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: refactored to for loops instead of streams because of the split nature of logging the items to be filtered out.

freeipa/src/test/java/com/sequenceiq/freeipa/service/instance/RebootInstanceServiceTest.java
holleyism: Added additional tests.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
lacikaaa: I think we should have the flow capable of restarting multiple instance at a time. If we have 2 unhealthy node from 5 we should restart them together. If we would like to restart all, we can still do it in 2 batch. 
If all node is in a bad state does it have any benefit to reboot them in a rolling mode?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            @Override
+            protected Selectable createRequest(RebootContext context) {
+                List<CloudInstance> cloudInstances = converterUtil.convertAll(List.of(context.getInstanceMetaData()), CloudInstance.class);
+                return new RebootInstancesRequest<>(context.getCloudContext(), context.getCloudCredential(), cloudInstances);
+            }
+
+            @Override
+            protected RebootContext createFlowContext(FlowParameters flowParameters, StateContext<RebootState, RebootEvent> stateContext,
```
lacikaaa: ok, so it's still a bit confusing here, so let me explain what I would do:
- InstanceEvent would contain stackId as resourceId as you need that to create the cloudcreadentials etc
- InstanceEvent would contain a set of instanceIds selected for reboot
- I would use `findAllInStack` for fetching instancemetadatas and filter in a stream, or have a separate repository query which contains stackId and instanceids too. Shouldn't make a big difference as freeipa won't have hundreds of instances
- You have to decide what happens if one/more/all instancemetadata is missing. I think this should not happen as it had been validated earlier, but maybe you should fail the flow with a message so it would be obvious that something is pretty wrong. Maybe you let in the force flag here and restart anyway the found instances and just log an error in that case
- if you really need these cloud stuff and real instancemetadata here then this override should be ok, so we wont make unnecessary db load for other steps.
- as this step is getting a big one, maybe you should move it a separate file
- if we send this event the restart will begin, but if we kill that service which is running this flow another will pick up from here and will send the reboot command again. Is it ok?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private Map<String, InstanceStatus> getInstanceHealthMap(String accountId, RebootInstancesRequest request) {
+        return healthDetailsService.getHealthDetails(request.getEnvironmentCrn(), accountId).getNodeHealthDetails().stream()
+                .collect(Collectors.toMap(NodeHealthDetails::getInstanceId, NodeHealthDetails::getStatus));
+    }
+
+    private Collection<String> getValidInstanceIds(Collection<String> allInstances, Collection<String> instanceIds) {
+        if (instanceIds == null || instanceIds.isEmpty()) {
+            return allInstances;
+        }
```
lacikaaa: please use an else here, so the indentation would reflect a logic too

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,116 @@
...
+            } else {
+                LOGGER.error("{} is an invalid instanceId.", instanceId);
+                badIds = true;
+            }
+        }
+        if (badIds) {
+            String msg = "Invalid instanceIds are present in request.";
+            LOGGER.error(msg);
+            throw new BadRequestException(msg);
+        }
```
lacikaaa: so we could say here `if(instanceIds.size() != validInstanceIds.size())` and just calculate the difference of the 2 collection and log at once and add it to the message.
this way the for could be a stream:
`Collection<String> validInstanceIds = instanceIds.stream().filter(allInstances.contains).collect()`
and you could drop the `badIds` flag

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
holleyism: refactored to have one event.  It is ok to reboot twice.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: changed.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: This had been refactored.  However, we will be adding rolling reboot later.  One of the uniqueness here is that we're running multiple services per instance.  If one service, say DNS, is down on one node, and PKI is down on another, both would be considered bad, but we'd want to rolling reboot these so we don't have total loss of service.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
```diff
@@ -0,0 +1,50 @@
...
+
+@ApiModel("RebootInstancesV1Request")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RebootInstancesRequest {
+    @NotEmpty
+    @ApiModelProperty(value = ModelDescriptions.ENVIRONMENT_CRN, required = true)
+    private String environmentCrn;
+
+    private boolean forceReboot;
```
lacikaaa: could you add a description here, so it would be documented on the API what's the difference. thanks

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/RebootService.java
```diff
@@ -0,0 +1,39 @@
...
+    private InstanceMetaDataService instanceMetaDataService;
+
+    @Inject
+    private RebootInstancesService rebootInstancesService;
+
+    public void startInstanceReboot(RebootContext context) {
+        instanceMetaDataService.updateStatus(context.getStack(), context.getInstanceIdList(), InstanceStatus.REBOOTING);
+    }
+
+    public void handleInstanceRebootError(String instanceIds) {
```
lacikaaa: so we would leave the instances in rebooting state in case of an error?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private Collection<String> getValidInstanceIds(Collection<String> allInstances, Collection<String> instanceIds) {
+        if (instanceIds == null || instanceIds.isEmpty()) {
+            return allInstances;
+        } else {
+            Collection<String> validInstanceIds = instanceIds.stream().filter(allInstances::contains).collect(Collectors.toSet());
+            if (validInstanceIds.size() != allInstances.size()) {
+                String badIds = instanceIds.stream()
+                        .filter(not(allInstances::contains)).collect(Collectors.joining(","));
+                if (badIds.length() > 0) {
```
lacikaaa: this `if` seems unnecessary. `badIds` can't be empty if we are in this part of the code

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
```diff
@@ -0,0 +1,166 @@
...
+            @Override
+            protected void doExecute(RebootContext context, InstanceFailureEvent payload, Map<Object, Object> variables) {
+                rebootService.handleInstanceRebootError(String.join(",", payload.getInstanceIds()));
+                LOGGER.error("Rebooting failed for {}.", context.getInstanceIds());
+                sendEvent(context);
+            }
+
+            @Override
+            protected Selectable createRequest(RebootContext context) {
+                return new InstanceEvent(RebootEvent.REBOOT_FAIL_HANDLED_EVENT.event(), context.getStack().getId(), context.getInstanceIdList());
```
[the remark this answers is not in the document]
holleyism: fixed.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/action/RebootActions.java
```diff
@@ -0,0 +1,166 @@
...
+            @Override
+            protected void doExecute(RebootContext context, InstanceEvent payload, Map<Object, Object> variables) {
+                rebootService.startInstanceReboot(context);
+                sendEvent(context);
+                LOGGER.info("Starting reboot for {}", context.getInstanceIds());
+            }
+
+            @Override
+            protected Selectable createRequest(RebootContext context) {
+                List<CloudInstance> cloudInstances = converterUtil.convertAll(context.getInstanceMetaDataList(), CloudInstance.class);
```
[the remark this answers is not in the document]
holleyism: changed.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RebootInstancesService.java
holleyism: removed

freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/RebootService.java
holleyism: fixed.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/reboot/RebootInstancesRequest.java
holleyism: added

holleyism:
The original reason for reboot was prompted by two failures that were experienced by customers.  The first one turned out to be the gssproxy daemon was not running and simply needed to be restarted.  The second one was because the underlying AWS hardware was unavailable and no cloudwatch triggers had been created to do an automatic reboot.  While these were addressed separately with their own fixes, the desire for reboot was for future unknown circumstances.  It's current usage is somewhat of a last resort to be able to try and recover.  Since FreeIPA has a lot of moving parts and is not perfect, it's difficult to predict what may go wrong in which piece and the effort to go through everything to ensure it's perfect is much greater than the simple reboot response.  There is an effort to better monitor individual services and in the future have finer grained control over restarting items, but having reboot is more about expediency and giving customers an immediate option instead of having to experience a new problem and have no option.

Additionally, reboot as part of an HA solution (future planned work) is necessary to enable rolling upgrades.  The current start/stop function is not sufficient as it stops the entire stack as opposed to a single node, and start could needlessly incur additional costs for the customer.

PR #7379, opened 2020-02-25
CB-5750. Drop oldest fluentd chunks - avoid buffer overflow

oleewere:
### 4 fluentd config changes

main change is to set the retry timeout (+ if it reaches the 2h limit, drop the backups as well from tmp) + set a total buffer limit (1-2 GB)

retry timeout: 
Default: 72h
The maximum seconds to retry to flush while failing, until plugin discards buffer chunks
- change it to 2h

disable_chunk_backup:
Default: false
Instead of storing unrecoverable chunks in the backup directory, just discard them
- enable this, will only impact /tmp folder

overflow_action
drop_oldest_chunk: drop/purge oldest chunk to accept newly incoming chunk
(default is throw exception)

total_limit_size
Once the total size of stored buffer reached this threshold, all append operations will fail with error (and data will be lost)
- set 1G for CM agent commands and 2G for service logs


lajosrodek:
LGTM

PR #7390, opened 2020-02-25
CB-5765 Skip declaring a variable for hue-hue_service_safety_valve fo…

cegganesh84:
…r CDPD versions below 7.1.0

1. As part of moving knox proxy hosts to dedicated CM setting in CB-4530, the Hue safety valve variable value setting was skipped by CB for CDPD versions 7.1.0 and above.
2. This was failing for 7.1.0 datamart clusters.
3. Here I attempted to fix this by setting the variable only for CDPD versions below 7.1.0

./gradlew build


cegganesh84:
The failures are due to this error, which at first glance looks unrelated to my change,

```
/grid/0/jenkins/workspace/workspace/cloudbreak-pull-request-builder-spotbugs-main/common/src/main/java/com/sequenceiq/cloudbreak/validation/SubnetType.java:8: error: unmappable character (0x80) for encoding US-ASCII
    // 16-bit block 192.168.0.0 ??? 192.168.255.255
```

cegganesh84:
retest this please

PR #7312, opened 2020-02-18
CB-5546 Mock integration tests for parent-child environment relation

pkedvessy:
CB-5546
Mock integration tests for parent-child environment relation

PR #7396, opened 2020-02-26
CB-5725 Child environment detach should remove reverse DNS zone from FreeIPA

Bajzathd:
CB-5725
Child environment detach should remove reverse DNS zone from FreeIPA

Depends on https://github.com/hortonworks/cloudbreak/pull/7312


doktoric:
retest this please

doktoric:
I think this should go to 2.19

PR #7400, opened 2020-02-26
CDPCP-1113 Determine environment access using rights

handavid:
This is based on two commits. One for updating the UMS proto and one for updating the FMS.

    CDPCP-1113. Sync users based on accessEnvironment right
    
    Users are now granted access to an environment based on the
    environments/accessEnvironment right instead of the EnvironmentUser,
    EnvironmentAdmin, or PowerUser role.
    
    Additionally, users are now added to the ipa "admins" group if they
    have the environments/adminFreeIPA right. This right is not granted
    to the EnvironmentAdmin role, so EnvironmentAdmins will be removed
    from the "admins" group. PowerUsers will still have this right.
    
    As part of this change, a hasRights method is added to the GrpcUmsClient.
    UmsUsersStateProvider has been modified to use the hasRights method and
    getWorkloadAdministrationGroups instead of getRights.

    CDPCP-1113. Update UMS for retrieving WAGs for a member
    
    The usermanagementservice.proto has been updated to include a new
    request ListWorkloadAdministrationGroupsForMember. This request
    returns the workload administration groups for the specified member
    (i.e., user or machine user) and must be called by the internal actor.
    
    The Grpc and MockUserManagementService have been updated to match.


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -129,95 +133,38 @@ private WorkloadCredential getCredentials(String userCrn, Optional<String> reque
...
+            });
 
-                List<String> workloadAdministrationGroupNames = rightsResponse.getWorkloadAdministrationGroupNameList();
-                LOGGER.debug("workloadAdministrationGroupNameList = {}", workloadAdministrationGroupNames);
-                workloadAdministrationGroupNames.forEach(groupName -> {
-                    usersStateBuilder.addGroup(nameToGroup(groupName));
-                    usersStateBuilder.addMemberToGroup(groupName, fmsUser.getName());
-                });
+            // Since this user is eligible, add this user to internal group
+            usersStateBuilder.addMemberToGroup(UserServiceConstants.CDP_USERSYNC_INTERNAL_GROUP, username);
```
lacikaaa: could you move this to a method called sth like `addEligibleUserToInnternalGroup` so we can drop  the comment

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -129,95 +133,38 @@ private WorkloadCredential getCredentials(String userCrn, Optional<String> reque
...
-            if (e.getStatus().getCode() == Status.Code.NOT_FOUND) {
-                LOGGER.warn("CRN {} not found in UMS. Treating as if CRN has no rights to environment {}: {}",
-                        memberCrn, environmentCrn, e.getLocalizedMessage());
-            } else {
-                throw e;
+            if (environmentAccessRights.hasAdminFreeIPARight()) {
+                usersStateBuilder.addMemberToGroup(ADMIN_FREEIPA_GROUP, username);
             }
+
+            // get credentials
```
lacikaaa: I think you can drop this command or move this to a method with a self explaining name

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
```diff
@@ -0,0 +1,82 @@
...
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.altus.CrnParseException;
+import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
+import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
+import com.sequenceiq.freeipa.service.freeipa.user.model.EnvironmentAccessRights;
+
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+
+public class EnvironmentAccessChecker {
```
horadla23: I would annotate this class with `@Service` and inject it into the proper class or make it static (maybe `@Service` is better because you can write mockito tests for it), instead of instantiate it every time

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
```diff
@@ -0,0 +1,82 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnvironmentAccessChecker.class);
+
+    private static final String IAM_INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
```
horadla23: Since this is Freeipa service, then I would create internal CRN for freeipa service (or environment because freeipa service is not a public service afaik), not for IAM service

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -329,6 +348,22 @@ public boolean checkRight(String actorCrn, String userCrn, String right, Optiona
...
+    /**
+     * Retrieves whether the member has the specified rights.
+     *
+     * @param actorCrn the CRN of the actor
+     * @param memberCrn the CRN of the member
+     * @param rightChecks the rights to check
+     * @param requestId an optional request id
+     * @return a list of booleans indicating whether the member has the specified rights
+     */
+    public List<Boolean> hasRights(String actorCrn, String memberCrn, List<AuthorizationProto.RightCheck> rightChecks, Optional<String> requestId) {
```
horadla23: please check checkRight method and add caching to this method similarly

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -329,6 +348,22 @@ public boolean checkRight(String actorCrn, String userCrn, String right, Optiona
...
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            AuthorizationClient client = new AuthorizationClient(channelWrapper.getChannel(), actorCrn);
+            return client.hasRights(requestId.orElse(UUID.randomUUID().toString()), memberCrn, rightChecks);
```
horadla23: we are currently not making call to UMS in case of internal CRN because right now everything allowed witn internalCRN afaik, there is an `if (InternalCrnBuilder.isInternalCrn(actorCrn))` in checkRight call for it, I think we should use that here also

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
handavid: I instantiate this each time I sync an environment so I can re-use the resource-dependent right check "environments/accessEnvironment".
If you prefer, I can try to swap this out for a Prototype bean and BeanFactory.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
handavid: The internal CRN needs to specify IAM.
See https://github.com/hortonworks/cloudbreak/blob/master/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/security/InternalCrnBuilder.java#L12-L15
See https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/libs/protocols/src/main/java/com/cloudera/thunderhead/service/common/crn/Crns.java#L712-L724

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
handavid: Caching is not a good idea for my use case. Here's a bad scenario that may occur with caching.
1. User access is removed
2. FMS checks event generation ids, sees that they have changed, and runs user sync
3. user sync gets the cached value for hasRights that still grants user access to environment
4. FMS saves the event generation ids for successful sync and thinks that all users are up-to-date
In this case, a user may retain access to an environment that they should not have access to.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
keyki: Also, many of the UMS calls fail with internal CRN.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: Our caches has small expiration time (1 sec afaik), so if permissions change, then it will turn out quickly and it prevents cb to overload network with ums calls, that is the reason why we have caches. If you think 1 sec expiration time still can cause issue, then we can skip cache here, but we had performance issues with UMS on mow-dev and/or mow-int.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
horadla23: If @keyki or @lacikaaa thinks it is ok, then you can leave this as-is, but you can reuse the result of `hasAccess` anywhere as a local variable or parameter

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: Not to mention that we are overloading network with internalCrn requests to UMS, while we know that everything is allowed with internalCrn

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
handavid: I'm not re-using the result because I'm calling hasAccess only once for each user and machineUser.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
handavid: I think the performance issues you had/have w/ checkRight is because the authorization layer is down at the database access level. From my reading, each database access, either read or write, requires a checkRight call.

The use-case here is different from protecting database access. Load here is determined by the application layer. I'm explicitly calling hasRights once per user/machineUser per user sync. The problem with caching isn't the expiration time, but that the rights check has been decoupled from other activities in the UMS or AuthorizationService. Decoupling can result in incorrect behavior and lead to a security hole. I think this is especially dangerous because the caching is invisible to the caller of hasRights and changes to the caching behavior may be made behind the scenes without understanding the callers.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
horadla23: this should be ok

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
handavid: I just looked at the checkRights code and it's wrong. You should check if the `InternalCrnBuilder.isInternalCrn(userCrn)` instead of actorCrn. I think that the only caller right now calls checkRight with actorCrn == userCrn, so we've been lucky so far.

@keyki Can you give an example of a UMS call that fails with the internal CRN? Do you mean internal CRN as the actor (grpc context) or as the member (part of request)? I'd like to understand the issue because it doesn't make sense to me. Could it be that you are creating internal CRNs with a service other than IAM?

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: I see your point, alright, then we should go with this.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
handavid: > Not to mention that we are overloading network with internalCrn requests to UMS, while we know that everything is allowed with internalCrn

I'm setting the actorCrn (grpc context) to internal Crn, but requesting whether the memberCrn has rights. I should never repeat a request within a user sync operation.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: moved

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
handavid: removed

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -347,6 +350,33 @@ public boolean checkRight(String actorCrn, String userCrn, String right, Optiona
...
+            return Arrays.asList(values);
+        }
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            AuthorizationClient client = new AuthorizationClient(channelWrapper.getChannel(), actorCrn);
+            List<Boolean> retVal = client.hasRights(requestId.orElse(UUID.randomUUID().toString()), memberCrn, rightChecks);
+            LOGGER.info("member {} has rights {}", memberCrn, retVal);
+            return retVal;
+        }
+    }
+
```
giladwolff: So I prefer not to do the right check here, in the grpc ums client. Let's not optimize this right now. I think that the right way to optimize this is to have distributed policies that have ttl on them. The policy for an internal actor can be "forever" so that the clients will never go to the authorization service. But I do not want to copy the authorization checks all over the place.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentAccessChecker.java
```diff
@@ -0,0 +1,82 @@
...
+        requireNonNull(requestId, "requestId is null");
+
+        try {
+            List<Boolean> hasRights = grpcUmsClient.hasRights(IAM_INTERNAL_ACTOR_CRN, memberCrn, rightChecks, requestId);
+            return new EnvironmentAccessRights(hasRights.get(0), hasRights.get(1));
+        } catch (StatusRuntimeException e) {
+            // NOT_FOUND errors indicate that a user/machineUser has been deleted after we have
+            // retrieved the list of users/machineUsers from the UMS. Treat these users as if
+            // they do not have the right to access this environment and belong to no groups.
+            if (e.getStatus().getCode() == Status.Code.NOT_FOUND) {
```
giladwolff: So this is fine for now, but maybe add the jira number tracking deletion of users from the FreeIPA here? Or is there an explicit check for the existence of users somewhere else?

PR #7402, opened 2020-02-27
CB-5791 Don't attempt to disable autorestart without KNOX service name

daszabo:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerConfigService.java
```diff
@@ -82,23 +85,35 @@ public void disableKnoxAutorestartIfCmVersionAtLeast(Versioned versionAtLeast, A
...
-                .findFirst().orElse(KNOX_SERVICE);
+    private Optional<String> getKnoxServiceName(String clusterName, ServicesResourceApi servicesResourceApi) {
+        try {
+            ApiServiceList serviceList = servicesResourceApi.readServices(clusterName, DataView.SUMMARY.name());
+            return serviceList.getItems().stream()
+                    .filter(service -> KNOX_SERVICE.equals(service.getType()))
+                    .map(ApiService::getName)
+                    .findFirst();
+        } catch (ApiException e) {
+            LOGGER.debug("Failed to get KNOX service name from Cloudera Manager.", e);
```
doktoric: Maybe the cluster name should be added to the log message

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerConfigService.java
```diff
@@ -82,23 +85,35 @@ public void disableKnoxAutorestartIfCmVersionAtLeast(Versioned versionAtLeast, A
...
             ApiConfig autorestartConfig = new ApiConfig().name(KNOX_AUTORESTART_ON_STOP).value(Boolean.FALSE.toString());
             ApiServiceConfig serviceConfig = new ApiServiceConfig().addItemsItem(autorestartConfig);
-            servicesResourceApi.updateServiceConfig(clusterName, knoxServiceName, "", serviceConfig);
-        } catch (ApiException e) {
-            LOGGER.debug("Failed to set autorestart_on_stop to KNOX in Cloudera Manager.", e);
-        }
+            try {
+                servicesResourceApi.updateServiceConfig(clusterName, knoxServiceName, "", serviceConfig);
+            } catch (ApiException e) {
+                LOGGER.debug("Failed to set autorestart_on_stop to KNOX in Cloudera Manager.", e);
```
doktoric: Maybe the cluster name should be added to the log message

doktoric:
I think unit test should be useful to cover these new lines in the code

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerConfigService.java
daszabo: Ideally the log entry should already contain it.

PR #7406, opened 2020-02-27
CB-5752 OpDB 7.1.0 cluster definitions out of sync for AWS and Azure

stoty:
for both the AWS and Azure definitions:
 * change  the name of the definition
 * set featureState to RELEASED
 * set bluePrintName to match the name of the 7.1.0 blueprint
* removed old duplicate 7.1.0 references from application.yml

Please apply to 2.18 as well !

testing done:

Started local cloudbreak
created environment
checked that the 7.1.0 AWS definition appears with the correct name
built the OpDB cluster
checked that uses the intended blueprint (with Phoenix)

Closes #CB-5752


stoty:
See https://github.com/hortonworks/cloudbreak/pull/7415 instead

PR #7415, opened 2020-02-28
CB-5752 OpDB 7.1.0 cluster definitions out of sync for AWS and Azure

stoty:
for both the AWS and Azure definitions:
* change the name of the definition
* set featureState to RELEASED
*set bluePrintName to match the name of the 7.1.0 blueprint

removed old duplicate 7.1.0 references from application.yml

testing done:

Started local cloudbreak
created environment
checked that the 7.1.0 AWS definition appears with the correct name
built the OpDB cluster
checked that uses the intended blueprint (with Phoenix)

Closes #CB-5752

### DIFF
--- a/core/src/main/resources/schema/app/20200228140101_CB-5752_rename_template.sql
+++ b/core/src/main/resources/schema/app/20200228140101_CB-5752_rename_template.sql
@@ -1,0 +1,15 @@
+-- // CB-5752 OpDB 7.1.0 cluster definitions out of sync for AWS and Azure
+-- Migration SQL that makes the change goes here.
+
+UPDATE clustertemplate SET name = '7.1.0 - Operational Database with SQL for AWS' WHERE status = 'DEFAULT' AND name = '7.1.0 - Operational Database with Phoenix for AWS';
+UPDATE clustertemplate SET name = '7.1.0 - Operational Database with SQL for Azure' WHERE status = 'DEFAULT' AND name = '7.1.0 - Operational Database for Azure';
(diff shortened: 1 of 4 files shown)


doktoric:
You need to take care with the old templates because you renamed on of the default one with the old name and one with the new name will exist.

stoty:
Is this SQL ok ?

core/src/main/resources/schema/app/20200228140101_CB-5752_rename_template.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-5752 OpDB 7.1.0 cluster definitions out of sync for AWS and Azure
+-- Migration SQL that makes the change goes here.
+
+UPDATE clustertemplate SET name = '7.1.0 - Operational Database with SQL for AWS' WHERE status = 'DEFAULT' AND name = '7.1.0 - Operational Database with Phoenix for AWS';
```
keyki: Should be extended to set the `7.1.0 - Operational Database: Apache HBase` blueprint to `DEFAULT_DELETED`

stoty: I have added the update statement for that now.

doktoric: thanks @stoty

doktoric:
@keyki I think with the new migration this change can be land in 2.18

PR #7417, opened 2020-02-28
CB-5813 fix license.txt permission

keyki:


doktoric:
I think we need this for 2.18 right ?

keyki:
I think 2.19 is fine.

PR #7428, opened 2020-02-28
CDPCP-1549. Fix issue where user sync failure is not recorded

handavid:
This PR contains two commits to address the behavior exhibited in CDPCP-1549.

CDPCP-1549. Run user sync internals as internal user

An authorization mismatch between the Operation db and the UserSyncStatus
db resulted in a bug where a user sync operation can be accepted, but
will fail when saving the user sync status. This commit changes the
authorization behavior so that authoriation is only performed on acceptance
of the sync operation. All the user sync internals are run as the internal
user.

CDPCP-1549. Catch and fail user sync operation

There was a gap where exceptions could cause user sync to fail but
not be recorded in the Operation database. This would prevent another
user sync from being accepted even though the current user sync was
not actually running. Eventually, the operation would be TIMEDOUT,
allowing another user sync to be accepted.

This commit closes the gap by catching all RuntimeExceptions and
recording the failure in the Operation database.

Testing: manual
1. created a machine user w/o write rights. verified that user sync was accepted but failed w/o updating the operation to FAILED.
2. Added change to catch and fail user sync. verified that user sync was accepted then failed
3. Added change to run user sync internals as internal user. verified that user sync was accepted and runs

### DIFF
--- a/cloud-common/src/main/java/com/sequenceiq/cloudbreak/concurrent/CompositeTaskDecorator.java
+++ b/cloud-common/src/main/java/com/sequenceiq/cloudbreak/concurrent/CompositeTaskDecorator.java
@@ -1,0 +1,36 @@
+package com.sequenceiq.cloudbreak.concurrent;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.List;
+
+import org.springframework.core.task.TaskDecorator;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Task decorator that composes multiple task decorators.
+ */
+public class CompositeTaskDecorator implements TaskDecorator {
+
+    private final ImmutableList<TaskDecorator> decorators;
+
+    /**
+     * Create a CompositeTaskDecorator from the list of task decorators.
+     *
+     * @param decorators a list of task decorators to apply, from innermost to outermost.
+     */
+    public CompositeTaskDecorator(List<TaskDecorator> decorators) {
+        requireNonNull(decorators, "decorators is null");
+        this.decorators = ImmutableList.copyOf(decorators);
+    }
+
+    @Override
+    public Runnable decorate(Runnable runnable) {
+        Runnable decoratedRunnable = requireNonNull(runnable, "runnable is null");
+        for (TaskDecorator decorator : decorators) {
+            decoratedRunnable = decorator.decorate(decoratedRunnable);
+        }
+        return decoratedRunnable;
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
@@ -114,19 +114,23 @@ public class UserSyncService {
         Operation operation = operationService
                 .startOperation(accountId, OperationType.USER_SYNC, environmentCrns, union(userCrnFilter, machineUserCrnFilter));
 
-        LOGGER.info("Starting operation [{}] with status [{}]", operation.getOperationId(), operation.getStatus());
+        String operationId = operation.getOperationId();
+        OperationState operationState = operation.getStatus();
+        LOGGER.info("Starting operation [{}] with status [{}]", operationId, operationState);
 
-        if (operation.getStatus() == OperationState.RUNNING) {
-            boolean fullSync = userCrnFilter.isEmpty() && machineUserCrnFilter.isEmpty();
-            if (fullSync) {
-                long currentTime = Instant.now().toEpochMilli();
-                stacks.forEach(stack -> {
-                    UserSyncStatus userSyncStatus = userSyncStatusService.getOrCreateForStack(stack);
-                    userSyncStatus.setLastFullSyncStartTime(currentTime);
-                    userSyncStatusService.save(userSyncStatus);
-                });
-            }
-            asyncSynchronizeUsers(operation.getOperationId(), accountId, actorCrn, stacks, userCrnFilter, machineUserCrnFilter, fullSync);
+        if (operationState == OperationState.RUNNING) {
+            tryWithOperationCleanup(operationId, accountId, () -> {
+                boolean fullSync = userCrnFilter.isEmpty() && machineUserCrnFilter.isEmpty();
+                if (fullSync) {
+                    long currentTime = Instant.now().toEpochMilli();
+                    stacks.forEach(stack -> {
+                        UserSyncStatus userSyncStatus = userSyncStatusService.getOrCreateForStack(stack);
+                        userSyncStatus.setLastFullSyncStartTime(currentTime);
+                        userSyncStatusService.save(userSyncStatus);
+                    });
+                }
+                asyncSynchronizeUsers(operation.getOperationId(), accountId, actorCrn, stacks, userCrnFilter, machineUserCrnFilter, fullSync);
+            });
         }
 
         return operation;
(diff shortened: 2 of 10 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -117,16 +117,22 @@ public Operation synchronizeUsers(String accountId, String actorCrn, Set<String>
...
+                if (fullSync) {
+                    long currentTime = Instant.now().toEpochMilli();
+                    stacks.forEach(stack -> {
+                        UserSyncStatus userSyncStatus = userSyncStatusService.getOrCreateForStack(stack);
+                        userSyncStatus.setLastFullSyncStartTime(currentTime);
+                        userSyncStatusService.save(userSyncStatus);
+                    });
+                }
+                asyncSynchronizeUsers(operation.getOperationId(), accountId, actorCrn, stacks, userCrnFilter, machineUserCrnFilter, fullSync);
+            } catch (RuntimeException e) {
```
giladwolff: I would catch Exceeption

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -184,7 +190,7 @@ private void internalSynchronizeUsers(String operationId, String accountId, Stri
             operationService.completeOperation(accountId, operationId, success, failure);
         } catch (RuntimeException e) {
```
giladwolff: Why not Exception?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
handavid: I catch RuntimeException because there isn't anything in the code that throws a checked exception. I can go either way on this one.

Catch RuntimeException - we narrow the exceptions classes that we catch. if the try code is changed in the future to throw a checked exception then we will know at compile time that we need to handle that.

Catch Exception - the sematics of this block are to clean up if "anything" goes wrong and Exception better represents "anything" than RuntimeException does. As a counterpoint, then what about Throwable? It may make sense here to catch Throwable because we still want to clean up if an Error occurs.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
handavid: same reasoning as above

handavid:
> How about tests to make sure you are running as the internal user?

I'm trying to follow how the CRN is passed through the async calls. I suspect that I'm missing something here and may need to revise this PR

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
handavid: I changed these to catch Throwable and attempt to save the operation failure to the database.

handavid:
> > How about tests to make sure you are running as the internal user?
> 
> I'm trying to follow how the CRN is passed through the async calls. I suspect that I'm missing something here and may need to revise this PR

I realized that the user crn wasn't propagated to the submitted Runnable. I added another decorator to propagate the crn as well as more testing.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -117,22 +121,37 @@ public Operation synchronizeUsers(String accountId, String actorCrn, Set<String>
...
+            } catch (Throwable t) {
+                try {
+                    LOGGER.error("Failed to start operation [{}]. Attempting to mark failure in database then re-throwing", operation, t);
+                    operationService.failOperation(accountId, operation.getOperationId(),
+                            "Failed to start operation: " + t.getLocalizedMessage());
+                } catch (Exception e) {
+                    LOGGER.error("Failed to mark operation [{}] as failed in database.", operation, e);
+                } finally {
+                    throw t;
+                }
```
lacikaaa: this looks almost the same like in `internalSynchronizeUsers`, could you move these into a separate method?

cloud-common/src/main/java/com/sequenceiq/cloudbreak/concurrent/UserCrnTaskDecorator.java
```diff
@@ -0,0 +1,19 @@
...
+ * the running task.
+ */
+public class UserCrnTaskDecorator implements TaskDecorator {
+    @Override
+    public Runnable decorate(Runnable runnable) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        return () -> {
+            ThreadBasedUserCrnProvider.doAs(userCrn, runnable);
+        };
+    }
```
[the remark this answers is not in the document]
handavid: ThreadBasedUserCrnProvider has a ThreadLocal to hold the actor CRN. The UserCrnTaskDecorator (to be renamed) gets the actor crn from the calling thread threadlocal and returns a runnable that sets the actor on the executor thread using doAs. doAs will save the actor crn from the executor thread threadlocal, set that threadlocal to the specified actor, and restore the threadlocal to its original actor at the end of execution.

handavid:
> With this solution I am not 100% that the right actor is used at the right time.

agreed. the actor management is somewhat mixed. db authorization calls retrieve the actor from the Threadlocal. application calls to the UMS sort of have to manage the actors themselves. I think it would be nice to re-do the GrpcUmsClient class so that the application wouldn't need to be cognizant of the actor.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
handavid: done

cloud-common/src/main/java/com/sequenceiq/cloudbreak/concurrent/CompositeTaskDecorator.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+
+import org.springframework.core.task.TaskDecorator;
+
+/**
+ * Task decorator that composes multiple task decorators.
+ */
+public class CompositeTaskDecorator implements TaskDecorator {
+
+    private final List<TaskDecorator> decorators;
```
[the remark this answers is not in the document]
handavid: done

cloud-common/src/main/java/com/sequenceiq/cloudbreak/concurrent/CompositeTaskDecorator.java
```diff
@@ -0,0 +1,34 @@
...
+     * @param decorators a list of task decorators to apply, from innermost to outermost.
+     */
+    public CompositeTaskDecorator(List<TaskDecorator> decorators) {
+        requireNonNull(decorators, "decorators is null");
+        this.decorators = List.copyOf(decorators);
+    }
+
+    @Override
+    public Runnable decorate(Runnable runnable) {
+        Runnable decoratedRunnable = runnable;
```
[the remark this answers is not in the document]
handavid: added requireNonNull

handavid:
@lacikaaa can you take another look at this review? thanks!

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -67,6 +69,8 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UserSyncService.class);
 
+    private static final String INTERNAL_USER_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
```
lacikaaa: why is it `IAM` and not `FREEIPA` as this service is not `IAM`?

lacikaaa:
CDPCP-1549: Are you sure this is the right fix for this? As I can see the main problem here is that `READ` permission is enough to write into `Operation` but you need `WRITE` for `UserSyncStatus`. I think this inconsistency should be fixed.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
handavid: Thunderhead only recognizes the internal user if the service is IAM.

handavid:
> CDPCP-1549: Are you sure this is the right fix for this? As I can see the main problem here is that `READ` permission is enough to write into `Operation` but you need `WRITE` for `UserSyncStatus`. I think this inconsistency should be fixed.

I think that the current fix is needed whether or not we change the permissions for UserSyncStatus. I can also include a permissions change if you would like.

There was definitely a gap where an exception could kill the sync operation w/o marking the operation as failed. That is fixed by one of these commits.

I also think it makes sense to run the user sync internals as the internal user. We already specify the internal user as the actor for most calls to the UMS so I think that setting the actor in the ThreadLocal is consistent with that behavior.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
keyki: There are other places as well where we use the internal crn to talk to UMS with FREEIPA service type

keyki: Verified locally and it doesn't work when `freeipa` is used instead of `iam`.

PR #7436, opened 2020-03-02
Nifi deregister

doktoric:


common/src/main/resources/messages/messages.properties
```diff
@@ -135,6 +135,7 @@ cm.cluster.services.starting=Starting Cloudera Manager services.
 cm.cluster.services.stopping=Stopping Cloudera Manager services.
 cm.cluster.services.stopped=Cloudera Manager services have been stopped.
 cm.cluster.securitygroup.too.strict=The security group of the Cloudera Manager node is probably too strict. Please check if all the necessary ports are reachable. Error message: {0}.
+cm.cluster.service.deregister.failed=Failed to deregister service(s) from Cloudera Manager. It's possible that CM or the instance is not running.
```
doktoric: I think it would be useful to mention to the enduser that they need to do this manually

attilapalfi92: Good idea

PR #7437, opened 2020-03-02
CB-5332 fix multiple termination trigger

lacikaaa:


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/UpdateHostsValidator.java
```diff
@@ -0,0 +1,128 @@
+package com.sequenceiq.cloudbreak.service.cluster.flow;
```
keyki: I'm not sure what would be a good name, but I don't like something being called *FlowService and the methods are not really related to flow operations.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterFlowService.java
```diff
@@ -0,0 +1,499 @@
...
+import com.sequenceiq.cloudbreak.structuredevent.event.CloudbreakEventService;
+import com.sequenceiq.cloudbreak.template.validation.BlueprintValidator;
+import com.sequenceiq.cloudbreak.util.UsageLoggingUtil;
+import com.sequenceiq.cloudbreak.workspace.model.User;
+import com.sequenceiq.cloudbreak.workspace.model.Workspace;
+import com.sequenceiq.common.api.type.InstanceGroupType;
+import com.sequenceiq.common.api.type.ResourceType;
+
+@Service
+public class ClusterFlowService {
```
keyki: I'm not sure what would be a good name, but I don't like something being called *FlowService and the methods are not really related to flow operations.

lacikaaa: renamed

lacikaaa:
retest this please

PR #7441, opened 2020-03-02
CB-5582 CB-5583 enforce authorization annotations

horadla23:


authorization-common/src/main/java/com/sequenceiq/authorization/service/AbstractPermissionCheckerService.java
```diff
@@ -60,34 +65,39 @@ protected Object checkPermission(ProceedingJoinPoint proceedingJoinPoint, Method
...
 
         List<? extends Annotation> annotations = getPossibleMethodAnnotations().stream()
                 .map(c -> methodSignature.getMethod().getAnnotation(c))
                 .collect(Collectors.toList());
 
-        Annotation methodAnnotation = validateNumberOfAnnotations(methodSignature, annotations);
-        if (methodAnnotation instanceof DisableCheckPermissions) {
+        Optional<Annotation> methodAnnotation = validateNumberOfAnnotations(methodSignature, annotations);
+        if (!methodAnnotation.isPresent()) {
+            LOGGER.info("Your Controller ({}) method {} does not have any authorization related annotation, " +
```
lnardai: This might be better as a Warn message.

authorization-common/src/main/java/com/sequenceiq/authorization/service/AbstractPermissionCheckerService.java
```diff
@@ -60,34 +65,39 @@ protected Object checkPermission(ProceedingJoinPoint proceedingJoinPoint, Method
...
-        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
-        PermissionChecker<? extends Annotation> permissionChecker = permissionCheckerMap.get(methodAnnotation.annotationType());
-        AuthorizationResourceType resource = classAuthorizationResource.type();
-        return permissionChecker.checkPermissions(methodAnnotation, resource, userCrn, proceedingJoinPoint, methodSignature, startTime);
+        PermissionChecker<? extends Annotation> permissionChecker = permissionCheckerMap.get(methodAnnotation.get().annotationType());
+        return permissionChecker.checkPermissions(methodAnnotation.get(), resource, userCrn, proceedingJoinPoint, methodSignature, startTime);
     }
 
-    private Annotation validateNumberOfAnnotations(MethodSignature methodSignature, List<? extends Annotation> annotations) {
+    private Optional<Annotation> validateNumberOfAnnotations(MethodSignature methodSignature, List<? extends Annotation> annotations) {
```
lnardai: What happens if someone specifies another annotation, should we verify an exact list of Annotations we are looking for? Would it be an error in the permission checking util or cause some RunTimeError?

horadla23: Child class (PermissionCheckerService) has a static list about possible authz annotations and we are checking those.

PR #7444, opened 2020-03-03
CB-3857: add Kudu web UI to list of access-controlled Knox endpoints

adembo:
Right now, any user may access the Kudu web UI; we need to ensure that Knox
enforces its access control mechanisms appropriately.

Even though commit 9e33314bf99 added all endpoints en masse, the Kudu web UI
endpoint was added two weeks prior in commit 409a0cf8c9, so it's conceivable
that the first commit missed the second.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -85,6 +85,10 @@
                 <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
             </param>
             <param>
+                <name>kuduui.acl</name>
+                <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
+            </param>
+            <param>
                 <name>jobhistoryui.acl</name>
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -83,6 +83,10 @@
             <param>
                 <name>impalaui.acl</name>
                 <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
+	    </param>
```
abhi2196: Can we add a space here for better indentation ?

adembo: Done. I had some issues with tabs and spaces, it would seem.

keyki:
2.19 is on dev, ~1,5 weeks to reach prod with this version. If you need this fix earlier then you can put it to 2.18, but it could be that this change will reach prod with 2.19 rather than a 2.18 patch version. Unless it's really urgent in that case we can upgrade prod with a newer version of 2.18.

doktoric:
I merge it then to rc2.18 and then merging back

PR #7452, opened 2020-03-03
CB-5896: Add CDP_FREEIPA_HA entitlement check.

jamisonbennett:
Check the CDP_FREEIPA_HA entitlement when creating FreeIPA HA
clusters.

Fixed the readme for the running the mock-caas.

Closes #CB-5896

### DIFF
--- a/mock-caas/src/main/resources/application.yml
+++ b/mock-caas/src/main/resources/application.yml
@@ -14,3 +14,4 @@ auth:
     baseimage.enable: true
     event-generation:
       expiration-minutes: 10
+    freeipa.ha.enable: true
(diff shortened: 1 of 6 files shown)


mock-caas/src/main/resources/application.yml
```diff
@@ -13,4 +13,5 @@ auth:
     ccm.enable: false
     baseimage.enable: true
     event-generation:
-      expiration-minutes: 10
\ No newline at end of file
+      expiration-minutes: 10
+    freeipa.ha.enable: true
```
doktoric: this means that in local mode we will always use ha IPA right ? I dont think that this is required in local dev

jamisonbennett: It means that when HA is requested it can be provisioned. It doesn't require it. The choice of HA vs non-HA is made in the create environment request.

jamisonbennett:
I am closing this PR and posting it against the rc-2.19 branch. https://github.com/hortonworks/cloudbreak/pull/7466

PR #7465, opened 2020-03-04
CB-5901 Lookup stacks by child environment crn as well

pkedvessy:
Lookup stacks for user sync and password setup by child environment crn as well.


pkedvessy:
retest this please

PR #7467, opened 2020-03-04
DISTX-397 Create data engineering HA template

cegganesh84:
1. Design : https://docs.google.com/document/d/1oN8l9p74gKNps1aeG2NCjdh1_6yRo5YZA4eTolS9dVw/edit
2. Added all the templates in preview mode as noted in the design.
3. The requirement was to use smaller instance type for quorum nodes. At the moment not possible because, a unit test fails with m5.xlarge for quorum nodes. To compensate that moved to Standard_D8_v3 for all the DE master nodes.
4. Tested that cluster definitions and blueprints came up fine in UI via moonlander private stack https://gk1-console.cdp-priv.mow-dev.cloudera.com/cloud/environments/list.

(cherry picked from commit 49b286a622eeb9219d4600707332edd5e6342e89)

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -211,6 +211,7 @@ cb:
     cm:
       defaults: >
         CDP 1.2 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie=cdp-data-engineering-702;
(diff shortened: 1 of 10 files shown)


cegganesh84:
I added this to 7.0.2 while it was supported. The fix for CB-5660 went in master only. So to save cloud cost, optimized DE HA template is proposed in master with this PR https://github.com/hortonworks/cloudbreak/pull/7712. Will close this PR for 2.19. Thanks!

PR #7471, opened 2020-03-05
CB-5938 Empty tag value breaks environment creation

doktoric:
unit test added

### DIFF
--- a/template-manager-tag/src/main/java/com/sequenceiq/cloudbreak/tag/DefaultCostTaggingService.java
+++ b/template-manager-tag/src/main/java/com/sequenceiq/cloudbreak/tag/DefaultCostTaggingService.java
@@ -57,12 +57,22 @@ public class DefaultCostTaggingService implements CostTagging {
     public Map<String, String> mergeTags(CDPTagMergeRequest request) {
         LOGGER.debug("About to merge tag(s)...");
         Map<String, String> result = new HashMap<>();
-        result.putAll(request.getRequestTags());
-        for (String key : request.getEnvironmentTags().keySet()) {
-            addTagIfNotPresented(result, request, key);
+        for (Map.Entry<String, String> entry : request.getRequestTags().entrySet()) {
+            if (!keyOrValueIsEmpty(entry.getKey(), entry.getValue())) {
+                result.put(entry.getKey(), entry.getValue());
+            }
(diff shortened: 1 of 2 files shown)


template-manager-tag/src/main/java/com/sequenceiq/cloudbreak/tag/DefaultCostTaggingService.java
```diff
@@ -57,14 +57,24 @@
     public Map<String, String> mergeTags(CDPTagMergeRequest request) {
         LOGGER.debug("About to merge tag(s)...");
         Map<String, String> result = new HashMap<>();
-        result.putAll(request.getRequestTags());
-        for (String key : request.getEnvironmentTags().keySet()) {
-            addTagIfNotPresented(result, request, key);
+        for (Map.Entry<String, String> entry : request.getRequestTags().entrySet()) {
```
keyki: request.getRequestTags() can be null

template-manager-tag/src/main/java/com/sequenceiq/cloudbreak/tag/DefaultCostTaggingService.java
```diff
@@ -57,14 +57,24 @@
...
+            if (!keyOrValueIsEmpty(entry.getKey(), entry.getValue())) {
+                addTagIfNotPresented(result, request, entry.getKey());
+            }
         }
         LOGGER.debug("The following requested tag(s) will be applied prepared: {}", result);
         return result;
     }
 
+    private boolean keyOrValueIsEmpty(String key, String value) {
+        return (key == null || key.isBlank() || key.isEmpty()) || (value == null || value.isBlank() || value.isEmpty());
```
keyki: do we need isBlank and isEmpty?

doktoric: yes

template-manager-tag/src/main/java/com/sequenceiq/cloudbreak/tag/DefaultCostTaggingService.java
doktoric: it can not!

PR #7473, opened 2020-03-05
CB-5584 Make environment accessible for creator only

foldik:
Add permission checks by resource


horadla23:
we will move back to account level check for now, no need for role assignment, but every method have to have checkpermissionbyaccount annotation

foldik:
@horadla23  I changed back the authz to `@CheckPermissionByAccount`.

PR #7475, opened 2020-03-05
CB-5582 CB-5583 generate correct credential owner crn

horadla23:


horadla23:
tested with mowdev ums, create worked

PR #7480, opened 2020-03-05
CB-5913 flow fixes

sodre90:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudbreakFlowService.java
```diff
@@ -35,18 +35,30 @@ public void getAndSaveLastCloudbreakFlowChainId(SdxCluster sdxCluster) {
...
+            return FlowState.UNKNOWN;
         } catch (Exception e) {
             LOGGER.error("Exception occured during checking if there is a flow for cluster {} in CB: {}", sdxCluster.getClusterName(), e.getMessage());
-            return true;
+            return FlowState.UNKNOWN;
         }
-        return false;
+    }
+
+    public enum FlowState {
```
lnardai: What's wrong with StateStatus? Why introduce a new Enum?

lnardai:
I see a lot of this repeating pattern:
```
                if (RUNNING.equals(flowState)) {

                } else {
                    
                }
```
Might be more readable with Optional:
`flowService.hasFlowWithState(Running).ifPresent().orElse();`

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudbreakFlowService.java
sodre90: I would like to handle unknown state like missing flow id or flow id is not known by cloudbreak. Also we call hasFlowRunningByChainId method which can only return a boolean.

PR #7481, opened 2020-03-05
CB-5896: Use the IAM service for the CDP_FREEIPA_HA entitlement

jamisonbennett:
Fix the CDP_FREEIPA_HA entitlement check to use the IAM service as the
internal actor's service. The UMS requires the internal actor to be
IAM.

This was tested locally with the mock-caas.

Closes #CB-5896

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
@@ -19,7 +19,7 @@ import com.sequenceiq.freeipa.util.CrnService;
 @Component
 public class CreateFreeIpaRequestValidator implements Validator<CreateFreeIpaRequest> {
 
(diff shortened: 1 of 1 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/controller/validation/CreateFreeIpaRequestValidator.java
```diff
@@ -19,7 +19,7 @@
 @Component
 public class CreateFreeIpaRequestValidator implements Validator<CreateFreeIpaRequest> {
 
-    static final String FREEIPA_INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.FREEIPA).getInternalCrnForServiceAsString();
+    static final String INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
```
giladwolff: So yes, the internal actor is currently a single, fixed CRN. In that fixed string the IAM is the service type. Until we change that I think that the InternalCrnBuilder should not take any parameters. It's most probably not something that you introduced or need to fix. It's just a thought.

PR #7484, opened 2020-03-06
CB-5099: FreeIPA changes to support AWS instance recovery

holleyism:
This change adds an automatic cloudwatch alarm during
the instance creation for AWS and removes it during
termination. This is only necessary for AWS as Azure
enables this by default. This enables an instance to
automatically recover when the underlying hardware,
owned by AWS, is faulty.

This was tested manually and verified that only FreeIPA instances receive the cloudwatch alarm.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -1,0 +1,90 @@
+package com.sequenceiq.cloudbreak.cloud.aws.connector.resource;
+
+import static com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts.CLOUDWATCH_CREATE_PARAMETER;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
+import com.amazonaws.services.cloudwatch.model.AmazonCloudWatchException;
+import com.amazonaws.services.cloudwatch.model.DeleteAlarmsRequest;
+import com.amazonaws.services.cloudwatch.model.Dimension;
+import com.amazonaws.services.cloudwatch.model.PutMetricAlarmRequest;
+import com.sequenceiq.cloudbreak.cloud.aws.AwsClient;
+import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
+import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
+import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
+import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+
+@Service
+public class AwsCloudWatchService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AwsCloudWatchService.class);
+
+    @Value("${freeipa.aws.cloudwatch.suffix:-Status-Check-Failed-System}")
+    private String alarmSuffix;
+
+    @Value("${freeipa.aws.cloudwatch.period:60}")
+    private int cloudwatchPeriod;
+
+    @Value("${freeipa.aws.cloudwatch.evaluationPeriods:2}")
+    private int cloudwatchEvaluationPeriods;
+
+    @Value("${freeipa.aws.cloudwatch.threshold:1.0}")
+    private double cloudwatchThreshhold;
+
+    @Inject
+    private AwsClient awsClient;
+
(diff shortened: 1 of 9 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudStack.java
```diff
@@ -33,8 +33,11 @@
 
     private final Optional<SpiFileSystem> fileSystem;
 
+    private final boolean createCloudWatchAlarm;
```
lacikaaa: could you drop this field in favor of adding only an entry to the parameter map? As it is very spi specific I think we should not include this as a field in `CloudStack`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
```diff
@@ -163,9 +176,35 @@
 
         awsTaggingService.tagRootVolumes(ac, amazonEC2Client, instances, stack.getTags());
 
+        if (stack.isCreateCloudWatchAlarm()) {
+            addCloudWatchAlarmsForSystemFailures(instances, regionName, credentialView);
+        }
         return awsResourceConnector.check(ac, instances);
     }
 
+    private void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances, String regionName, AwsCredentialView credentialView) {
```
lacikaaa: could you move cloudwatch related code/methods to a separate component?

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudStack.java
holleyism: fixed

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
holleyism: moved

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudStack.java
```diff
@@ -13,6 +13,8 @@
  */
 public class CloudStack {
 
+    public static final String CLOUDWATCH_CREATE_PARAMETER = "createCloudWatchAlarm";
```
lacikaaa: I think this would be better in `PlatformParametersConsts`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AwsCloudWatchService.class);
+
+    private  static final String ALARM_SUFFIX = "-Status-Check-Failed-System";
+
+    private static final int CLOUDWATCH_PERIOD = 60;
+
+    private static final int CLOUDWATCH_EVALUATION_PERIODS = 2;
+
+    private static final double CLOUDWATCH_THRESHOLD = 1.0;
```
lacikaaa: Could you move these to `@Values` so we can easily override them?

freeipa/src/main/java/com/sequenceiq/freeipa/converter/cloud/StackToCloudStackConverter.java
```diff
@@ -96,8 +97,9 @@ public CloudStack convert(Stack stack, Collection<String> deleteRequestedInstanc
         Network network = buildNetwork(stack);
         InstanceAuthentication instanceAuthentication = buildInstanceAuthentication(stack.getStackAuthentication());
 
-        return new CloudStack(instanceGroups, network, image, Collections.emptyMap(), getUserDefinedTags(stack), stack.getTemplate(),
-                instanceAuthentication, instanceAuthentication.getLoginUserName(), instanceAuthentication.getPublicKey(), null);
+        return new CloudStack(instanceGroups, network, image, Map.of(CLOUDWATCH_CREATE_PARAMETER, Boolean.TRUE.toString()),
```
lacikaaa: this should be also come from a `@Value` so we can turn it off/on as we wish

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudStack.java
holleyism: moved

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
holleyism: done

freeipa/src/main/java/com/sequenceiq/freeipa/converter/cloud/StackToCloudStackConverter.java
holleyism: done

PR #7485, opened 2020-03-06
Rc219master

horadla23:


horadla23:
I haven't noticed but @keyki already created the smae PR with unit test fix here: https://github.com/hortonworks/cloudbreak/pull/7485
so I close this ons

PR #7490, opened 2020-03-06
CB-5917 use parent environment virtual groups for workloads in child environments

biharitomi:


core/src/main/java/com/sequenceiq/cloudbreak/service/environment/EnvironmentConfigProvider.java
```diff
@@ -22,4 +22,18 @@ public boolean isChildEnvironment(String environmentCrn) {
             throw new CloudbreakServiceException("Failed to get environment details from Environment service due to:", ex);
         }
     }
+
+    public String getParentEnvironmentCrn(String environmentCrn) {
+        try {
+            String result = environmentCrn;
+            DetailedEnvironmentResponse environmentResponse = environmentClientService.getByCrn(environmentCrn);
+            String parentEnvironmentCrn = environmentResponse.getParentEnvironmentCrn();
```
doktoric: some log would be nice ere

PR #7492, opened 2020-03-06
CB-5902: use azure region display name instead of region value

tiborpopovics:


keyki:
Is this not needed on 2.18?

lajosrodek:
2.18: #7488

keyki:
Then this should be a merge.

PR #7493, opened 2020-03-06
CB-5902: use azure region display name instead of region value

tiborpopovics:


keyki:
Should be a merge from rc branch

PR #7494, opened 2020-03-06
CB-5583 CB-5584 Differentiate resource dependent and independent rights

horadla23:
it seems we need to differentiate completely rights in case of account level check and in case of resource level check
related logic: https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/usermanagement/src/main/java/com/cloudera/thunderhead/service/usermanagement/server/AuthorizationService.java
https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/libs/common/src/main/java/com/cloudera/thunderhead/service/common/authorization/Rights.java


horadla23:
Related thunderhead change: https://review.infra.cloudera.com/r/108016/
Reviewbot is failing because of some dependency issue, not related to the change

PR #7500, opened 2020-03-06
CB-5556 second attempt at making Sdx Repair internals better, introdu…

lnardai:


datalake/src/main/java/com/sequenceiq/datalake/settings/SdxRepairSettings.java
```diff
@@ -0,0 +1,36 @@
...
+    public static SdxRepairSettings from(SdxRepairRequest request) {
+        if (StringUtils.isNotBlank(request.getHostGroupName()) && CollectionUtils.isNotEmpty(request.getHostGroupNames())) {
+            throw new BadRequestException("Please send only one hostGroupName in the 'hostGroupName' field " +
+                    "or multiple hostGroups in the 'hostGroupNames' fields");
+        }
+        SdxRepairSettings settings = new SdxRepairSettings();
+        if (Strings.isNotEmpty(request.getHostGroupName())) {
+            settings.hostGroupNames = List.of(request.getHostGroupName());
+        }
+        settings.hostGroupNames = request.getHostGroupNames();
```
lajosrodek: I think this statement shall be in the `else` clause of the `if` above, otherwise `settings.hostGroupNames` might get overwritten (with `null` or whatever is passed in `request.getHostGroupNames()`) even if a single host group was to be repaired. Alternatively, the statement can be also moved before the `if`, whichever approach you prefer. 😄

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxRepairService.java
```diff
@@ -76,15 +75,15 @@ public FlowIdentifier triggerRepairByName(String userCrn, String clusterName, Sd
         return sdxReactorFlowManager.triggerSdxRepairFlow(cluster.getId(), clusterRepairRequest);
     }
 
-    public void startSdxRepair(Long id, SdxRepairRequest repairRequest) {
+    public void startSdxRepair(Long id, SdxRepairSettings repairRequest) {
```
lajosrodek: Minor: rename `repairRequest` to `repairSettings`. Similarly in the other two changes below.

PR #7502, opened 2020-03-06
Merge rc-2.19 into master

handavid:


horadla23:
@handavid we had to revert your change in rc219 since it has compile error (https://github.com/hortonworks/cloudbreak/pull/7503)

handavid:
ok. I'll close this and fix that one.

PR #7505, opened 2020-03-07
CB-5904: Fix FreeIPA provisioning error 400 No required SSL ...

jamisonbennett:
certificate was sent

Wait for nginx to start on all FreeIPA gateway nodes. Then save the
certificate for each of them. Previously a random node was checked
one time for each node in the cluster and that node's certificate was
saved.

This was tested locally by deploying 50 clusters without the the
"400 No required SSL certificate was sent" error. Although the
issue was intermittent, it occurred regularly enough that the problem
has likely been resolved.

Closes #CB-5904


jamisonbennett:
I tagged the wrong JIRA number...I will post a new PR.

PR #7511, opened 2020-03-09
CB-5710 - added CDP CLI request converters on Environment and Datahub APIs

attilapalfi92:


doktoric:
I think overall we should some specific term for publi api converters like using "PublicAPI" in the class name what do you think @attilapalfi92 ?

attilapalfi92:
@doktoric I moved them to specific packages which contain "cli". I could rename cli to publicapi?

doktoric:
I am fine with cli

PR #7513, opened 2020-03-09
CB-5790 Cleanup freeipa during cluster creaton

lacikaaa:


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationActions.java
```diff
@@ -107,11 +112,33 @@ protected Selectable createRequest(StackContext context) {
...
+                Set<String> ips = instanceMetaData.stream().map(InstanceMetaData::getPrivateIp).collect(Collectors.toSet());
+                return new CleanupFreeIpaEvent(context.getStack().getId(), hostNames, ips);
+            }
+        };
+    }
+
     @Bean(name = "BOOTSTRAPPING_PUBLIC_ENDPOINT_STATE")
     public Action<?, ?> bootStrappingPublicEndpointAction() {
-        return new AbstractStackCreationAction<>(HostMetadataSetupSuccess.class) {
+        return new AbstractStackCreationAction<>(StackEvent.class) {
```
handavid: Why is this StackEvent instead of CleanupFreeIpaEvent? Is it because we don't actually care about anything in the event?

PR #7515, opened 2020-03-09
CB-5520: Health check provide a message in instance issues when the node is UNREACHABLE

holleyism:
This patch adds an issue entry when the node is UNREACHABLE.  This is to provide better end user messaging with the health check.

Updated and ran unit tests as well as verified with local freeipa against AWS instance.


wonderslug:
lgtm

PR #7516, opened 2020-03-09
CB-5339 - Adding automated cloud backups to FreeIPA

wonderslug:
This adds support for automated backup of the FreeIPA
service.  It backups up to the cloud provider defined
by the telemetry and uses the same logging bucket or
container as well as the Identity.

It will determine the backup directory root pathing
similarly to the way telemetry does.

It stores the backup configuration with the stack so
that as we add its own bucket later on, backward
compatibility can be maintained.

Closes #CB-5339
Closes #CB-5336
Closes #CB-5095

Co-authored-by: holleyism <git@holleyism.com>

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
@@ -1,0 +1,59 @@
+package com.sequenceiq.freeipa.converter.backup;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.common.api.cloudstorage.old.AdlsGen2CloudStorageV1Parameters;
+import com.sequenceiq.common.api.cloudstorage.old.S3CloudStorageV1Parameters;
+import com.sequenceiq.common.api.telemetry.request.LoggingRequest;
+import com.sequenceiq.common.api.telemetry.request.TelemetryRequest;
+import com.sequenceiq.freeipa.api.model.Backup;
+import com.sequenceiq.freeipa.configuration.BackupConfiguration;
+
+@Component
+public class BackupConverter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BackupConverter.class);
+
+    private final boolean freeIpaBackupEnabled;
+
+    private final BackupConfiguration backupConfiguration;
+
+    public BackupConverter(BackupConfiguration backupConfiguration,
+            @Value("${freeipa.backup.enabled:true}") boolean freeIpaBackupEnabled) {
+        this.backupConfiguration = backupConfiguration;
+        this.freeIpaBackupEnabled = freeIpaBackupEnabled;
+    }
+
+    public Backup convert(TelemetryRequest request) {
+        Backup backup = null;
+        if (freeIpaBackupEnabled && request != null && request.getLogging() != null) {
+            backup = new Backup();
+            backup.setMonthlyFullEnabled(backupConfiguration.isMonthlyFullEnabled());
+            backup.setInitialFullEnabled(backupConfiguration.isRunInitialFullAfterInstall());
+            backup.setHourlyEnabled(backupConfiguration.isHourlyEnabled());
+            decorateBackupFromLoggingRequest(backup, request.getLogging());
+        }
+        return backup;
+    }
+
+    private void decorateBackupFromLoggingRequest(Backup backup, LoggingRequest loggingRequest) {
+        if (loggingRequest != null) {
+            backup.setStorageLocation(loggingRequest.getStorageLocation());
+            if (loggingRequest.getS3() != null) {
+                S3CloudStorageV1Parameters s3Params = new S3CloudStorageV1Parameters();
+                s3Params.setInstanceProfile(loggingRequest.getS3().getInstanceProfile());
+                backup.setS3(s3Params);
+            } else if (loggingRequest.getAdlsGen2() != null) {
+                AdlsGen2CloudStorageV1Parameters adlsGen2Params = new AdlsGen2CloudStorageV1Parameters();
+                AdlsGen2CloudStorageV1Parameters adlsGen2FromRequest = loggingRequest.getAdlsGen2();
+                adlsGen2Params.setAccountKey(adlsGen2FromRequest.getAccountKey());
+                adlsGen2Params.setAccountName(adlsGen2FromRequest.getAccountName());
+                adlsGen2Params.setSecure(adlsGen2FromRequest.isSecure());
+                adlsGen2Params.setManagedIdentity(adlsGen2FromRequest.getManagedIdentity());
+                backup.setAdlsGen2(adlsGen2Params);
+            }
+        }
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
@@ -1,0 +1,61 @@
+package com.sequenceiq.freeipa.service.freeipa.backup.cloud;
+
+import java.nio.file.Paths;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
+
+@Component
+public class AdlsGen2BackupConfigGenerator extends CloudBackupConfigGenerator<AdlsGen2BackupConfig> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AdlsGen2BackupConfigGenerator.class);
+
+    private static final String[] ADLS_GEN2_SCHEME_PREFIXES = {"abfs://", "abfss://"};
+
+    private static final String AZURE_DFS_DOMAIN_SUFFIX = ".dfs.core.windows.net";
+
+    private static final String AZURE_BLOB_STORAGE_SUFFIX = "blob.core.windows.net";
+
+    private static final String AZURE_BLOB_STORAGE_SCHEMA = "https://";
+
+    @Override
+    public String generateBackupLocation(String location, String clusterType,
+            String clusterName, String clusterId) {
+        AdlsGen2BackupConfig adlsGen2BackupConfig = generateBackupConfig(location);
+        String logFolder = resolveBackupFolder(adlsGen2BackupConfig, clusterType, clusterName, clusterId);
+        String hostPart = String.format("%s.%s", adlsGen2BackupConfig.getAccount(), AZURE_BLOB_STORAGE_SUFFIX);
+        String generatedLocation = String.format("%s%s", AZURE_BLOB_STORAGE_SCHEMA,
+                Paths.get(hostPart, adlsGen2BackupConfig.getFileSystem(), logFolder));
+        LOGGER.info("The following ADLS Gen2 base folder location is generated: {} (from {})",
+                generatedLocation, location);
+        return generatedLocation;
+    }
+
+    private AdlsGen2BackupConfig generateBackupConfig(String location) {
+        if (StringUtils.isNotEmpty(location)) {
+            String locationWithoutScheme = getLocationWithoutSchemePrefixes(location, ADLS_GEN2_SCHEME_PREFIXES);
+            String[] locationSplit = locationWithoutScheme.split("@");
+            String[] storageWithSuffix = locationSplit[0].split("/", 2);
+            String folderPrefix = storageWithSuffix.length < 2 ? "" :  "/" + storageWithSuffix[1];
+            if (locationSplit.length < 2) {
+                return new AdlsGen2BackupConfig(folderPrefix, storageWithSuffix[0], null);
+            } else {
+                String[] splitByDomain = locationSplit[1].split(AZURE_DFS_DOMAIN_SUFFIX);
+                String account = splitByDomain[0];
+                if (splitByDomain.length > 1) {
+                    String folderPrefixAfterDomain = splitByDomain[1];
+                    if (StringUtils.isNoneEmpty(folderPrefix, folderPrefixAfterDomain)) {
+                        throw new CloudbreakServiceException(String.format("Invalid ADLS Gen2 path: %s", location));
+                    }
+                    folderPrefix = StringUtils.isNotEmpty(folderPrefixAfterDomain) ? folderPrefixAfterDomain : folderPrefix;
+                }
+                return new AdlsGen2BackupConfig(folderPrefix, storageWithSuffix[0], account);
+            }
+        }
+        throw new CloudbreakServiceException("Storage location parameter is missing for ADLS Gen2");
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupConfigGenerator.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupConfigGenerator.java
@@ -1,0 +1,37 @@
+package com.sequenceiq.freeipa.service.freeipa.backup.cloud;
+
+import java.nio.file.Paths;
+
+import org.apache.commons.lang3.StringUtils;
+
+public abstract class CloudBackupConfigGenerator<T extends CloudBackupStorageConfig> {
+
+    protected static final String CLUSTER_BACKUP_PREFIX = "cluster-backups";
+
+    public abstract String generateBackupLocation(String location, String clusterType,
+            String clusterName, String clusterId);
+
+    String getLocationWithoutSchemePrefixes(String input, String... schemePrefixes) {
+        for (String schemePrefix : schemePrefixes) {
+            if (input.startsWith(schemePrefix)) {
+                String[] splitted = input.split(schemePrefix);
+                if (splitted.length > 1) {
+                    return splitted[1];
+                }
+            }
+        }
+        return input;
+    }
+
+    String resolveBackupFolder(CloudBackupStorageConfig cloudBackupStorageConfig, String clusterType,
+            String clusterName, String clusterId) {
+        String clusterIdentifier = String.format("%s_%s", clusterName, clusterId);
+
+        if (StringUtils.isNotEmpty(cloudBackupStorageConfig.getFolderPrefix())) {
+            return Paths.get(cloudBackupStorageConfig.getFolderPrefix(), CLUSTER_BACKUP_PREFIX, clusterType,
+                    clusterIdentifier).toString();
+        } else {
+            return Paths.get(CLUSTER_BACKUP_PREFIX, clusterType, clusterIdentifier).toString();
+        }
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaConfigView.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaConfigView.java
@@ -1,0 +1,165 @@
+package com.sequenceiq.freeipa.service.freeipa.config;
+
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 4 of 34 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+        LOGGER.info("The following ADLS Gen2 base folder location is generated: {} (from {})",
+                generatedLocation, location);
+        return generatedLocation;
+    }
+
+    private AdlsGen2Config generateBackupConfig(String location) {
+        if (StringUtils.isNotEmpty(location)) {
+            boolean secure = location.startsWith(ADLS_GEN2_SCHEME_PREFIXES[1]);
+            String locationWithoutScheme = getLocationWithoutSchemePrefixes(location, ADLS_GEN2_SCHEME_PREFIXES);
+            String[] split = locationWithoutScheme.split("@");
```
lacikaaa: could you change it to a more meaningful name?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+            String[] storageWithSuffix = split[0].split("/", 2);
+            String folderPrefix = storageWithSuffix.length < 2 ? "" :  "/" + storageWithSuffix[1];
+            if (split.length < 2) {
+                return new AdlsGen2Config(folderPrefix, storageWithSuffix[0], null, secure);
+            } else {
+                String[] splitByDomain = split[1].split(AZURE_DFS_DOMAIN_SUFFIX);
+                String account = splitByDomain[0];
+                if (splitByDomain.length > 1) {
+                    String folderPrefixAfterDomain = splitByDomain[1];
```
lacikaaa: so many magic numbers. why `2`?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaConfigView.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @SuppressWarnings("ExecutableStatementCount")
+    public Map<String, Object> toMap() {
+        Map<String, Object> map = new HashMap<>();
+        map.put("backup_enabled", this.backupEnabled);
+        map.put("backup_location", ObjectUtils.defaultIfNull(this.backupLocation, EMPTY_CONFIG_DEFAULT));
+        map.put("monthly_full_backup_enabled", this.monthlyFullBackUpEnabled);
+        map.put("hourly_backup_enabled", this.hourlyBackUpEnabled);
+        map.put("initial_full_backup_enabled", this.initialFullBackupEnabled);
+        map.put("backup_platform", ObjectUtils.defaultIfNull(this.backupPlatform, EMPTY_CONFIG_DEFAULT));
```
lacikaaa: backup related stuff could be under backup as they logically stick together

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupFolderResolverServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import com.sequenceiq.cloudbreak.auth.altus.CrnParseException;
+import com.sequenceiq.cloudbreak.telemetry.fluent.FluentClusterType;
+import com.sequenceiq.common.api.cloudstorage.old.AdlsGen2CloudStorageV1Parameters;
+import com.sequenceiq.common.api.cloudstorage.old.S3CloudStorageV1Parameters;
+import com.sequenceiq.freeipa.api.model.Backup;
+
+public class CloudBackupFolderResolverServiceTest {
```
lacikaaa: I think `S3BackupConfigGenerator` and `AdlsGen2BackupConfigGenerator` should be tested in different test, and here you should test the logic related to `CloudBackupFolderResolverService`. I have a feeling that not all the cases are covered here, and it would be too complex to test it through this class

freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final BackupConfiguration backupConfiguration;
+
+    public BackupConverter(BackupConfiguration backupConfiguration,
+            @Value("${freeipa.backup.enabled:true}") boolean freeIpaBackupEnabled) {
+        this.backupConfiguration = backupConfiguration;
+        this.freeIpaBackupEnabled = freeIpaBackupEnabled;
+    }
+
+    public Backup convert(TelemetryRequest request) {
```
lacikaaa: it's a bit strange that we use the telemetry settings for backup. Is it really a good idea?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
```diff
@@ -39,4 +39,53 @@ restart_krb5kdc:
...
+  file.managed:
+    - source: salt://freeipa/templates/freeipa_backup.conf.j2
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 640
+
+{% if salt['pillar.get']('freeipa:monthly_full_backup_enabled') %}
+/etc/cron.monthly/freeipa_backup_monthly:
+  file.managed:
```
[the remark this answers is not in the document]
wonderslug: No because it is not using cron it is using anacron.  This is necessary because anacron puts in some randomization of runtime for parts like monthly.  We don't want all the monthly backups to kick off at the same time in an HA environment because the monthly full backup brings down the freeipa for about 2 minutes.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/templates/freeipa_backup_monthly.j2
```diff
@@ -0,0 +1,5 @@
+{%- from 'freeipa/settings.sls' import freeipa with context -%}
+#!/bin/bash
+if [ ! -f /tmp/freeipa_backup_monthly_bypass ]; then
```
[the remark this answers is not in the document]
wonderslug: same answer.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/templates/freeipa_backup_hourly.j2
```diff
@@ -0,0 +1,6 @@
+{%- from 'freeipa/settings.sls' import freeipa with context -%}
+#!/bin/bash
+if [ ! -f /tmp/freeipa_backup_hourly_bypass ]; then
```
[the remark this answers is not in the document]
wonderslug: The file does not exist by default.  It is looked for to allow support to disable the hourly and monthly backups temporarily while a restore is done.  The freeipa backup only supports a specific /var/lib/ipa/backup directory.  We want to remove old backups from the local.  If the backups are running while someone is trying to do a restore it will remove the local restore they are trying to do.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
wonderslug: This is a direct copy of the Telemetry processing of the storage locations for logging.  Do we want to change this overall?

freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
wonderslug: This is a shorter-term decision so we did not have to introduce using another s3 bucket or azure container.  We do not have time to introduce all the UI and flow components to support the additional storage location at the moment.  The Logging bucket/container is currently the only one available to the freeipa at the moment.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
wonderslug: which part?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaCreationService.java
```diff
@@ -115,8 +120,15 @@ public DescribeFreeIpaResponse launchFreeIpa(CreateFreeIpaRequest request, Strin
         Telemetry telemetry = stack.getTelemetry();
         cloudStorageFolderResolverService.updateStorageLocation(telemetry,
                 FluentClusterType.FREEIPA.value(), stack.getName(), stack.getResourceCrn());
+
```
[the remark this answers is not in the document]
wonderslug: done

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
```diff
@@ -39,4 +39,53 @@ restart_krb5kdc:
...
+freeipa_initial_full_backup:
+  cmd.run:
+    - name: /usr/local/bin/freeipa_backup -t FULL -f "{{salt['grains.get']('fqdn')}}/full" && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/freeipa_initial_backup-executed
+    - failhard: False
+    - unless: test -f /var/log/freeipa_initial_backup-executed
+    - require:
+        - file: /usr/local/bin/freeipa_backup
+        - file: /etc/freeipa_backup.conf
+{% endif %}
+{% endif %}
```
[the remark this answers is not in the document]
wonderslug: fair enough.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
+import com.sequenceiq.cloudbreak.telemetry.fluent.cloud.AdlsGen2Config;
+
+@Component
+public class AdlsGen2BackupConfigGenerator extends CloudBackupConfigGenerator<AdlsGen2Config> {
```
oleewere: why duplicate this kind of classes/implementations from common ? most of the code looks duplication

wonderslug: Its not an outright copy.  Right now it is using very similar logic on the parsing of the adls container and s3 buckets paths, but the functions are slightly different with the base backups folder being different and a changed call path. 

As well, later on we will be moving to a dedicated selection of the backup storage location to a stand-alone bucket/container but the UI and flow work was unable to be done in the time needed for that call.  This allows for the seperate parsing mechanism for the backup once it moves away from the telementry storage location.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
lacikaaa: the variable split

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
lacikaaa: then comes the question, why do we have to copy if it's the same?

wonderslug: The functions and flow is not an outright copy. Right now it is using very similar logic on the parsing of the adls container and s3 buckets paths, but the functions are slightly different with the base backups folder being different and a changed call path.

As well, later on, we will be moving to a dedicated selection of the backup storage location to a stand-alone bucket/container but the UI and flow work was unable to be done in the time needed for that call. This allows for the separate parsing mechanism for the backup once it moves away from the telemetry storage location.

freeipa/src/test/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupFolderResolverServiceTest.java
wonderslug: Fair enough.   don.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
wonderslug: Done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/AdlsGen2BackupConfigGenerator.java
oleewere: you can make things more abstract, but fine...btw it's nothing to do with Fluent but it has many references for it, so that looks logically invalid for me (for example: using fluent cluster type that is used as a string in fluentd related configs - so if i would change that for logging it would affect this code as well)

wonderslug: Fair enough.  I refactored it to be more independent.  There is a single value enum because It will be added to later on when it gets more options.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaConfigView.java
wonderslug: Ill see if I can do this

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupConfigGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+public abstract class CloudBackupConfigGenerator<T extends CloudBackupStorageConfig> {
+
+    protected static final String CLUSTER_BACKUP_PREFIX = "cluster-backups";
+
+    public abstract String generateBackupLocation(String location, String clusterType,
+            String clusterName, String clusterId);
+
+    String getLocationWithoutSchemePrefixes(String input, String... schemePrefixes) {
```
oleewere: im wondering why is this needed here. I had to split and store different parts of the storage location parts (scheme + host + path) because of fluentd configuration requirements, i think in your case the path will work as is as you are using s3 tool for the upload (probably same is true for azure)...so probably most of these path generator code are not required at all...just pass the storage location as is (with appending the cluster-backup and cluster name/id prefixes)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaConfigView.java
wonderslug: Changed the structure

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/backup/cloud/CloudBackupConfigGenerator.java
wonderslug: Actually this needs the exact same sort of segmentation as logs.  Adding additional user-defined folders as well as segmenting the into the per cluster based folders is all still necessary.

wonderslug: An example of this is right now in out SDX environments we have a shared bucket which we put in the environment name as the top-level under the bucket in order to allow us to put the logs and data both under that bucket on a per-environment basis

oleewere: but is the splitting really needed? concat segments would be enough instead of the generator service usage as schema/host parts can be included in (at least) s3 tool (so transformations can be avoided)

freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
```diff
@@ -4,3 +4,11 @@ freeipa:
   realm: testrealm
   admin_user: admin
   freeipa_to_replicate:
+  backup:
+    enabled: true
```
lacikaaa: I would set it here to false, and override it from java if enabled

freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
```diff
@@ -0,0 +1,47 @@
...
+    - user: root
+    - group: root
+    - mode: 750
+{% endif %}
+
+{% if salt['pillar.get']('freeipa:backup:initial_full_enabled') %}
+freeipa_initial_full_backup:
+  cmd.run:
+    - name: /usr/local/bin/freeipa_backup -t FULL -f "{{salt['grains.get']('fqdn')}}/full" && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/freeipa_initial_backup-executed
+    - failhard: False
```
lacikaaa: this option is not needed, by default it's false

freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
holleyism: The desire from PM and exec is to have the default be enabled.  We don't have mechanisms in the UI to enable this so enabling for customers will be difficult.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
holleyism: removed

freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
lacikaaa: enable by default from java. usually we place empty or test values into pillars so it won't contain anything that could cause wrong behavior. I just want to be sure that when this feature is enabled, it's intentional. Like if somebody touch the code and doesn't push down the pillar related to backup when it is disabled, it would cause backup being enabled, because the value here says it.

holleyism: changed.

keyki:
@wonderslug do you have an estimate on how much data are we going to store on S3? I'm asking because on mow-dev and int we have ~500 environments and if it's a lot then we should think about disabling it on some of these environments.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
oleewere: i would suggest to support this from the API (env or probably better for freeipa) in this PR already, as backup is nothing to do with telemetry, especially with logging. use this "temporal" solution (to fill data from telemetry) only if backup request is not filled in the freeipa request. so if the UI will support it you can remove the logging usage (but you will have everything that is needed for the UI to implement that)

holleyism: We can change this when we add support for UI configuration.  I've added https://jira.cloudera.com/browse/CB-6146

holleyism:
> @wonderslug do you have an estimate on how much data are we going to store on S3? I'm asking because on mow-dev and int we have ~500 environments and if it's a lot then we should think about disabling it on some of these environments.

In dev environments, full backup is 10M, incremental is 1M. 5G storage is $0.023 / GB = $0.115/month for the full backups.

PR #7517, opened 2020-03-09
CB-3864 round robin for preffered subnet

doktoric:


environment/src/test/java/com/sequenceiq/environment/network/service/SubnetIdProviderTest.java
```diff
@@ -109,6 +109,26 @@ void testProvideShouldReturnAnySubnetWhenResultIsEmptyAndNoError() {
         Assertions.assertNotNull(actual);
     }
 
+    @Test
+    void testProvideShouldReturnAnySubnetWhenResultIsEmptyAndNoError1() {
```
topolyai5: What does it mean the `1` at the end of the test?

PR #7530, opened 2020-03-10
CB-5590 spark3 added

doktoric:
We dont have spark3 template yet


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -480,6 +480,17 @@
     {%- endif %}
     {%- endif %}
 
+    {% if 'SPARK3_YARN_HISTORY_SERVER' in salt['pillar.get']('gateway:location') -%}
+    {% if 'SPARK3HISTORYUI' in exposed -%}
+    <service>
+        <role>SPARK3HISTORYUI</role>
```
keyki: I think `ACLs` are missing from the topology.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/spark/Spark3OnYarnRoleConfigProvider.java
```diff
@@ -0,0 +1,52 @@
...
+import com.cloudera.api.swagger.model.ApiClusterTemplateConfig;
+import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateProcessor;
+import com.sequenceiq.cloudbreak.cmtemplate.configproviders.AbstractRoleConfigProvider;
+import com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils;
+import com.sequenceiq.cloudbreak.template.TemplatePreparationObject;
+
+@Component
+public class Spark3OnYarnRoleConfigProvider extends AbstractRoleConfigProvider {
+
+    private static final String SPARK3_CONF_CLIENT_SAFETY_VALVE = "spark-conf/spark-defaults.conf_client_config_safety_valve";
```
risdenk: I thought we were moving away from safety valves? This uses another one?

doktoric: @risdenk unfortunately we can only moving aways if we have fix property in CM isde

risdenk: Yea agreed - CB-4191 is being used to track a bunch of this. Is there a CM jira to add a specific property and a CB one to remove this use of safety valve? Otherwise we are going to keep fighting up a hill.

This looks like a new Spark CSD so it should be possible to add the specific config and then CB wouldn't need to deal w/ this safety valve.

doktoric: @risdenk this is https://jira.cloudera.com/browse/OPSAPS-55172

PR #7544, opened 2020-03-11
CB-5727	Solve CM clock offset issues

Bajzathd:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerSetupService.java
```diff
@@ -227,7 +219,7 @@ public Cluster buildCluster(
             installCluster(cluster, apiClusterTemplate, clouderaManagerResourceApi, prewarmed);
             clouderaManagerMgmtLaunchService.startManagementServices(stack, apiClient);
             if (CloudPlatform.YARN.equalsIgnoreCase(stack.getCloudPlatform())) {
```
biharitomi: This condition could also go to the yarn related implementation.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/util/ConfigUtils.java
```diff
@@ -25,4 +26,14 @@ public static ApiConfig makeApiConfig(String name, String value) {
         apiConfig.setValue(value);
         return apiConfig;
     }
+
+    public static void addConfig(ApiRole apiRole, ApiConfig apiConfig) {
```
biharitomi: I think it could be part of CmConfigServiceDelegate interface as a default method.

PR #7547, opened 2020-03-11
CB-6035: FMS should relogin on failures

holleyism:
This change adds an automatic retry on the FreeIpaClientFactory
getFreeIpaClient* methods.  These methods perform a login.
Manually verified by blocking http/https ports.  This is necessary
in FreeIPA HA environments where there might be a failover and
the client should retry.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientException.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientException.java
@@ -1,19 +1,27 @@
 package com.sequenceiq.freeipa.client;
 
+import java.util.OptionalInt;
+
 public class FreeIpaClientException extends Exception {
+
+    private final OptionalInt statusCode;
+
     public FreeIpaClientException(String message) {
         super(message);
+        statusCode = OptionalInt.empty();
+    }
+
+    public FreeIpaClientException(String message, int statusCode) {
+        super(message);
+        this.statusCode = OptionalInt.of(statusCode);
     }
 
     public FreeIpaClientException(String message, Throwable cause) {
         super(message, cause);
+        statusCode = OptionalInt.empty();
     }
 
-    public FreeIpaClientException(Throwable cause) {
-
-    }
-
-    public FreeIpaClientException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
-
+    public OptionalInt getStatusCode() {
+        return statusCode;
     }
 }
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
@@ -71,8 +71,6 @@ public class FreeIpaClientBuilder {
 
     private final String pass;
 
-    private final String realm;
-
     private final SSLContext sslContext;
 
     private final int port;
@@ -83,11 +81,10 @@ public class FreeIpaClientBuilder {
 
     private Map<String, String> additionalHeaders;
 
-    public FreeIpaClientBuilder(String user, String pass, String realm, HttpClientConfig clientConfig, int port, String basePath,
+    public FreeIpaClientBuilder(String user, String pass, HttpClientConfig clientConfig, int port, String basePath,
             Map<String, String> additionalHeaders, JsonRpcClient.RequestListener rpcRequestListener) throws Exception {
         this.user = user;
         this.pass = pass;
-        this.realm = realm;
         this.clientConfig = clientConfig;
         this.port = port;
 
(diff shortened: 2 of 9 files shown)


holleyism:
Hold off on review at this point.  Might be changing.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientException.java
```diff
@@ -1,19 +1,25 @@
 package com.sequenceiq.freeipa.client;
 
 public class FreeIpaClientException extends Exception {
+
+    private final int statusCode;
```
jamisonbennett: I think the statusCode should be `OptionalInt` since its not always present.

lacikaaa:
Could you remove the label when it's ready for review @holleyism ?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientException.java
holleyism: changed

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -97,11 +114,25 @@ public FreeIpaClient getFreeIpaClientForStackWithPing(Stack stack) throws Except
...
                 } else {
-                    return getFreeIpaClientBuilder(stack).buildWithPing();
+                    Exception lastException = new FreeIpaClientException("No instances exist in the stack.");
+                    for (InstanceMetaData instanceMetaData: stack.getInstanceGroups().stream().flatMap(
+                            instanceGroup -> instanceGroup.getInstanceMetaData().stream()).collect(Collectors.toList())) {
+                        try {
+                            return getFreeIpaClientBuilder(stack, instanceMetaData).buildWithPing();
+                        } catch (FreeIpaClientException | IOException e) {
+                            lastException = e;
+                            if (e instanceof FreeIpaClientException) {
```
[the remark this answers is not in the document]
holleyism: moved

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -193,21 +187,14 @@ private String connect(String user, String pass, String apiAddress, int port)
                 throw new FreeIpaClientException(String.format("Encountered unexpected response from "
                         + "FreeIPA; details:%n%n"
                         + "code: %s%n"
-                        + "headers: %s", response.getStatusLine().getStatusCode(), response.getAllHeaders()));
+                        + "headers: %s", response.getStatusLine().getStatusCode(), response.getAllHeaders()),
+                        response.getStatusLine().getStatusCode());
```
lacikaaa: I don't think this `response.getStatusLine().getStatusCode()` is necessary

freeipa/src/main/java/com/sequenceiq/freeipa/service/TlsSecurityService.java
```diff
@@ -103,9 +103,8 @@ private void generateSaltPassword(SaltSecurityConfig saltSecurityConfig) {
         saltSecurityConfig.setSaltPasswordVault(saltPassword);
     }
 
-    public GatewayConfig buildGatewayConfig(Long stackId, InstanceMetaData gatewayInstance, Integer gatewayPort,
+    public GatewayConfig buildGatewayConfig(Stack stack, InstanceMetaData gatewayInstance, Integer gatewayPort,
```
lacikaaa: if we get stack here instead of id we can omit the gw port from parameters as it's in the stack already

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -70,42 +73,48 @@ private String toClusterProxyBasepath(String freeIpaClusterCrn) {
...
                 } else {
-                    return getFreeIpaClientBuilder(stack).build();
+                    Exception lastException = new FreeIpaClientException("No instances exist in the stack.");
+                    for (InstanceMetaData instanceMetaData: stack.getInstanceGroups().stream().flatMap(
+                            instanceGroup -> instanceGroup.getInstanceMetaData().stream()).collect(Collectors.toList())) {
+                        try {
+                            return getFreeIpaClientBuilder(stack, instanceMetaData).build(withPing);
+                        } catch (FreeIpaClientException | IOException e) {
+                            lastException = e;
+                            if (e instanceof FreeIpaClientException) {
```
lacikaaa: this wouldn't be necessary if there is 2 catch for the 2 exception

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -70,42 +73,48 @@ private String toClusterProxyBasepath(String freeIpaClusterCrn) {
...
         Status stackStatus = stack.getStackStatus().getStatus();
         if (!stackStatus.isFreeIpaUnreachableStatus()) {
             try {
                 if (clusterProxyService.isCreateConfigForClusterProxy(stack)) {
-                    return getFreeIpaClientBuilderForClusterProxy(stack).build();
+                    return getFreeIpaClientBuilderForClusterProxy(stack).build(withPing);
                 } else {
-                    return getFreeIpaClientBuilder(stack).build();
+                    Exception lastException = new FreeIpaClientException("No instances exist in the stack.");
+                    for (InstanceMetaData instanceMetaData: stack.getInstanceGroups().stream().flatMap(
```
lacikaaa: please move out the stream into a variable and/or to a method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -70,42 +73,48 @@ private String toClusterProxyBasepath(String freeIpaClusterCrn) {
...
+        stack = stackService.getByIdWithListsInTransaction(stack.getId());
         Status stackStatus = stack.getStackStatus().getStatus();
         if (!stackStatus.isFreeIpaUnreachableStatus()) {
             try {
                 if (clusterProxyService.isCreateConfigForClusterProxy(stack)) {
-                    return getFreeIpaClientBuilderForClusterProxy(stack).build();
+                    return getFreeIpaClientBuilderForClusterProxy(stack).build(withPing);
                 } else {
-                    return getFreeIpaClientBuilder(stack).build();
+                    Exception lastException = new FreeIpaClientException("No instances exist in the stack.");
```
lacikaaa: this exception driven logic and return inside the for loop makes the code hard to read.
I think it would be more readable if you do something like have an `Optional<FreeIpaClient> client = Optional.empty()` and iterate through the instances while it's empty and you have more instances. In the end you could return with the client or throw an exception

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
holleyism: It's necessary for the calling function to determine if the exception was due to not being authorized.  We want to retry logging it to different IPA servers for failover, unless the exception is because we're unauthorized.

freeipa/src/main/java/com/sequenceiq/freeipa/service/TlsSecurityService.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
holleyism: fixed

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
holleyism: moved

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
holleyism: changed.

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
lacikaaa: ok, I missed the parenthesis

jamisonbennett:
I repushed on behalf of @holleyism so that the tests could be rerun. It had been too long since they had run so the links were no longer available for the test results.

PR #7548, opened 2020-03-11
CDPCP-1601. set-password now uses credentials from control plane

aarman-cloudera:
The setPassword call now uses the password hash and keys from UMS
when possible. Note that we plan to deprecate / remove this endpoint
soon, but for now the UI is still calling it and this behavior
will keep the UMS keys consistent with the keys in FreeIPA, which
is desirable for user keytab generation.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
@@ -1,0 +1,19 @@
+package com.sequenceiq.freeipa.client;
+
+import com.sequenceiq.freeipa.client.model.Config;
+
+import java.util.Set;
+
+public class FreeIpaCapabilities {
+
+    // Attribute name to store users passwords (enabled through FreeIPA plugin)
+    private static final String CDP_USER_ATTRIBUTE = "cdpUserAttr";
+
+    private FreeIpaCapabilities() {
+    }
+
+    public static boolean hasSetPasswordHashSupport(Config ipaConfig) {
+        Set<String> ipauserobjectclasses = ipaConfig.getIpauserobjectclasses();
+        return ipauserobjectclasses != null && ipauserobjectclasses.contains(CDP_USER_ATTRIBUTE);
+    }
+}
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
@@ -9,14 +9,17 @@ public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> 
 
     private final String username;
 
+    private final String userCrn;
+
     private final String password;
 
     private final Optional<Instant> expirationInstant;
 
-    public SetPasswordRequest(Long stackId, String environment, String username, String password, Optional<Instant> expirationInstant) {
+    public SetPasswordRequest(Long stackId, String environment, String username, String userCrn, String password, Optional<Instant> expirationInstant) {
         super(stackId);
         this.environment = environment;
         this.username = username;
+        this.userCrn = userCrn;
         this.password = password;
         this.expirationInstant = expirationInstant;
     }
@@ -27,6 +30,10 @@ public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> 
 
     public String getUsername() {
         return username;
+    }
+
+    public String getUserCrn() {
+        return userCrn;
     }
 
     public String getPassword() {
@@ -43,6 +50,7 @@ public class SetPasswordRequest extends FreeIpaClientRequest<SetPasswordResult> 
                 + "stackId='" + getResourceId() + '\''
                 + "environment='" + environment + '\''
                 + "username='" + username + '\''
+                + "userCrn='" + userCrn + '\''
                 + "expirationInstant=" + expirationInstant
                 + '}';
     }
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
@@ -2,6 +2,14 @@ package com.sequenceiq.freeipa.flow.freeipa.user.handler;
 
 import javax.inject.Inject;
 
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
+import com.sequenceiq.cloudbreak.logger.MDCUtils;
+import com.sequenceiq.freeipa.client.FreeIpaCapabilities;
+import com.sequenceiq.freeipa.client.FreeIpaClientException;
+import com.sequenceiq.freeipa.service.freeipa.user.UmsCredentialProvider;
+import com.sequenceiq.freeipa.service.freeipa.user.kerberos.KrbKeySetEncoder;
+import com.sequenceiq.freeipa.service.freeipa.user.model.WorkloadCredential;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
@@ -18,10 +26,16 @@ import com.sequenceiq.freeipa.service.stack.StackService;
 
 import reactor.bus.Event;
 
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Optional;
+
 @Component
 public class SetPasswordHandler implements EventHandler<SetPasswordRequest> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SetPasswordHandler.class);
+
+    private static final String IAM_INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.IAM).getInternalCrnForServiceAsString();
 
     @Inject
     private StackService stackService;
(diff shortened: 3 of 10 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredential.java
```diff
@@ -32,4 +33,12 @@ public String getHashedPassword() {
     public Optional<Instant> getExpirationDate() {
         return expirationDate;
     }
+
+    public static WorkloadCredential fromGetActorWorkloadCredentialsResponse(GetActorWorkloadCredentialsResponse response) {
```
lacikaaa: few things:
- method name should start with verb, like create
- I don't like this static method, because it will make it hard to mock during testing. Maybe moving this logic into a converter component would make it easier to mock and test
- missing tests

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -197,6 +197,11 @@ public User userSetPasswordHash(String user, String hashedPassword,
         return userMod(user, params);
     }
 
+    public boolean hasSetPasswordHashSupport() throws FreeIpaClientException {
```
handavid: doesn't quite feel right in the FreeIpaClient class. We will also likely start checking other object classes as well.

I suggest pulling this out into a separate class. Maybe FreeIpaObjectClasses or FreeIpaCapabilities

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -43,7 +70,13 @@ public void accept(Event<SetPasswordRequest> setPasswordRequestEvent) {
             MDCBuilder.buildMdcContext(stack);
 
             FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientForStack(stack);
-            freeIpaClient.userSetPasswordWithExpiration(request.getUsername(), request.getPassword(), request.getExpirationInstant());
+            if (freeIpaClient.hasSetPasswordHashSupport()) {
+                LOGGER.info("IPA has password hash support, credentials information from UMS will be used.");
+                setPasswordHashFromUms(freeIpaClient, request.getUsername(), request.getUserCrn(), request.getExpirationInstant());
+            } else {
+                freeIpaClient.userSetPasswordWithExpiration(
```
handavid: Let's add a log message here also. This will let us search to logs for whether anyone is using the "old way"

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -270,14 +269,12 @@ private void applyStateDifferenceToIpa(String environmentCrn, FreeIpaClient free
 
     private void processUsersWorkloadCredentials(
             String environmentCrn, UmsUsersState umsUsersState, FreeIpaClient freeIpaClient) throws IOException, FreeIpaClientException {
-        Config config = freeIpaClient.getConfig();
-        if (config.getIpauserobjectclasses() == null || !config.getIpauserobjectclasses().contains(Config.CDP_USER_ATTRIBUTE)) {
-            LOGGER.debug("Doesn't seems like having config attribute, no credentials sync required for env:{}", environmentCrn);
+        if (!freeIpaClient.hasSetPasswordHashSupport()) {
+            LOGGER.debug("IPA doesn't have password hash support, no credentials sync required for env:{}", environmentCrn);
```
handavid: I like the re-wording. it's clearer

aarman-cloudera:
> so it will be possible that the user sets the password from the UI but we won't change it to that because we will take it from UMS? I think it could cause some confusion.

No, when a user sets password from the UI. It first makes an api call in IAM/UMS to set workload password. It then makes this api call.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
aarman-cloudera: Moved it into a FreeIpaCapabilities class. I assume you meant we still want it in the freeipa client package.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredential.java
aarman-cloudera: Fixed. Also, instead of moving to a provider I ended up put the logic in a UmsCredentialProvider which also makes the grpc call.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
```diff
@@ -0,0 +1,20 @@
+package com.sequenceiq.freeipa.client;
+
+import com.sequenceiq.freeipa.client.model.Config;
+
+import java.util.Set;
+
+public class FreeIpaCapabilities {
+
+    // users for password setting in FreeIPA Plugin
```
giladwolff: the comment seems wrong. Maybe say "Attribute name to store users passwords ...."

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
```diff
@@ -0,0 +1,20 @@
...
+    private static final String CDP_USER_ATTRIBUTE = "cdpUserAttr";
+
+    private FreeIpaCapabilities() {
```
giladwolff: why have it at all?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
```diff
@@ -9,14 +9,17 @@
...
     private final String username;
 
+    private final String userCrn;
+
     private final String password;
 
     private final Optional<Instant> expirationInstant;
 
-    public SetPasswordRequest(Long stackId, String environment, String username, String password, Optional<Instant> expirationInstant) {
+    public SetPasswordRequest(Long stackId, String environment, String username, String userCrn, String password, Optional<Instant> expirationInstant) {
```
giladwolff: I assume that the userCrn is the actor for which the password is set? and that is extracted from the authentication context? Maybe call this actorCrn as the caller maybe a machine user? Also, how do you know the username at this point? This is the workload username, right?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
```diff
@@ -43,6 +50,7 @@ public String toString() {
                 + "stackId='" + getResourceId() + '\''
                 + "environment='" + environment + '\''
                 + "username='" + username + '\''
+                + "userCrn='" + userCrn + '\''
```
giladwolff: So not you, but there are nicer ways to do this, like using guava object helper.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsCredentialProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient.INTERNAL_ACTOR_CRN;
+
+@Component
+public class UmsCredentialProvider {
+
+    @Inject
+    private GrpcUmsClient grpcUmsClient;
+
+    public WorkloadCredential getCredentials(String userCrn, Optional<String> requestId) {
```
giladwolff: I could swear that I've seen this code before.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
aarman-cloudera: fixed

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaCapabilities.java
aarman-cloudera: This class only provides static methods and should not be instantiated, hence private constructor. There are other utility classes in this package do the same. Actually I'll remove the comment so that it's consistent with those classes.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
aarman-cloudera: Yes userCrn is the actor for which the password is set and is pulled from authentication context. Changing it to actorCrn does make sense though I'm seeing userCrn be used for both user and machine user in other places in this package. I think we can leave at as userCrn for now, since we will probably deprecate this path once single user sync is in.

The username is the workload username, the existing code calls ums to get the workload username from ums. See PasswordService.getUserIdFromUserCrn(...)

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/event/SetPasswordRequest.java
aarman-cloudera: Ah nice one. I'll try it out on new classes. And I'll keep this one the same for consistency.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsCredentialProvider.java
aarman-cloudera: Well this was extracted off of UmsUsersStateProvider, and thanks to your comment I just realized I forgot remove it from there since it's no longer used (weird that code quality doesn't complain about unused functions).

PR #7549, opened 2020-03-12
CB-5956: enable cdh images with extra parcels

tiborpopovics:


keyki:
Branch should be 2.19

PR #7564, opened 2020-03-12
CB-6066 - Remove WebAppSec cors.enabled from topology.xml.j2

risdenk:
This removes cors.enabled=true from the topology.xml. I am working on testing this on some real clusters.


keyki:
Shall we wait for @yhemanth to comment on the JIRA?

risdenk:
Hemanth confirmed should be ok on jira.

PR #7566, opened 2020-03-12
CB-5582 Differentiate environment related credential API

horadla23:


horadla23:
We need to separate (independent) credential API and environment related credential API because we need to do different authorization on them. API paths will not change.

horadla23:
We discussed with @lnardai  that separating credential authorization into credential based and environment based authorization can cause anomalies, so I close this for now.

PR #7568, opened 2020-03-12
CB-5921 Make CM setup restartable

daszabo:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerSecurityService.java
```diff
@@ -128,36 +130,61 @@ public void disableSecurity() {
...
+            throw new ClusterClientInitException(e);
+        } catch (ApiException e) {
+            if (org.springframework.http.HttpStatus.UNAUTHORIZED.value() == e.getCode()) {
+                LOGGER.debug("Cloudera Manager already running, old admin user's password has been changed.");
+                Cluster cluster = stack.getCluster();
+                String user = cluster.getCloudbreakAmbariUser();
+                String password = cluster.getCloudbreakAmbariPassword();
+                return clouderaManagerApiClientProvider.getClient(stack.getGatewayPort(), user, password, clientConfig);
+            }
+            LOGGER.debug("Cloudera Manager is not running.");
```
keyki: This log should include the exception stacktrace.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerSetupService.java
```diff
@@ -160,6 +156,30 @@ public void waitForServer() throws CloudbreakException, ClusterClientInitExcepti
         }
     }
 
+    private ApiClient createApiClient() throws ClusterClientInitException {
```
keyki: There are 2 methods that look the same with minor changes. I'm not sure if this can be moved to some common place or worth generalizing it. It's up to you, just wanted to make a note of it.

daszabo: Their behaviour is quite different. I cannot really come up with a common method for them.

PR #7572, opened 2020-03-12
CB-5245. Add account telemetry settings

oleewere:
Adding new account level telemetry settings:
- some global feature options: log collection + workload analytics (these can be overridden by environment level inputs) - defaults are false
- anonymiziation rules that are used for log collection (defined only on account level) - if nothing is attached to the account - use default ones (email,card numbers - same patterns as in CM)
- removed altus anonymization service and replace with this account level telemetry settings
- adding endpoints to create/edit account telemetry, and also endpoints for features or rules (standalone)

accounttelemetry - new table
implemented in a similar way as it was done for account tag

TODO/next steps:
- implement pattern testing for anonymiziation rules


doktoric:
overall looks ok but more log would be fine! @oleewere

PR #7582, opened 2020-03-13
CB-5768 Add entitlement and enable feature flag for IDBroker mapping validation

risdenk:
Adds IDBroker mapping validation as an entitlement and defaults the feature flag of environment request to true for IDBroker mapping validation. This lays the way to enable this per tenant in CDP in case there are issues with SimulatePrincipalPolicy due to IAM limits or other AWS issues (altus found that SimulatePrincipalPolicy is not always accurate).

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/validation/cloudstorage/CloudStorageValidator.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/validation/cloudstorage/CloudStorageValidator.java
@@ -5,12 +5,17 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.providerservices.CloudProviderServicesV4Endopint;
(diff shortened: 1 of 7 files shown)


risdenk:
Rebased to resolve merge conflict with master.

datalake/src/main/java/com/sequenceiq/datalake/service/validation/cloudstorage/CloudStorageValidator.java
```diff
@@ -21,28 +26,42 @@
 @Component
 public class CloudStorageValidator {
 
+    private static final String DATALAKE_INTERNAL_ACTOR_CRN = new InternalCrnBuilder(Crn.Service.DATALAKE).getInternalCrnForServiceAsString();
```
jamisonbennett: You may need to use the `IAM` service rather than `DATALAKE` in order to get the internal actor CRN to validate in the UMS. See https://github.com/hortonworks/cloudbreak/pull/7481#discussion_r388555752.

PR #7584, opened 2020-03-13
CDPCP-1533. Add operations to usersyncstatus

handavid:
First part to the rolling update of the freeipa database schema
for tracking user sync status.

This commit adds the requested and successful operations to the
usersync status. The new lastrequestedoperation field is preferred
but lastsyncstarttime field is still used as a
fallback when calculating whether a stack is "cool" for the purposes
of automatic user sync polling.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
@@ -32,6 +32,12 @@ public class UserSyncStatus implements AccountIdAwareResource {
 
     private Long lastFullSyncEndTime;
 
+    @OneToOne
+    private Operation lastStartedFullSync;
+
+    @OneToOne
+    private Operation lastSuccessfulFullSync;
+
     public UserSyncStatus() {
     }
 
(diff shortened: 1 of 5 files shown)


handavid:
> please rebase this to master as we will cut a new rc very soon

ok. will do!

handavid:
@lacikaaa Hi! I've rebased and re-ran all the checks.

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -32,6 +32,12 @@
 
     private Long lastFullSyncEndTime;
 
+    @OneToOne
+    private Operation lastRequestedFullSync;
```
giladwolff: so this does not get updated when a request is rejected because there is an ongoing user sync right?

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -32,6 +32,12 @@
...
+
+    @OneToOne
+    private Operation lastSuccessfulFullSync;
```
giladwolff: So the idea is that this holds the last successful operation and the lastRequestedFullSync holds the last full sync? Maybe rename it to reflect that (drop the requested?)

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -63,22 +69,58 @@ public void setUmsEventGenerationIds(Json umsEventGenerationIds) {
...
+    /**
+     * Sets the last full sync end time.
+     * @deprecated information set is redundant with {@link #setLastSuccessfulFullSync(Operation)}
+     */
+    @Deprecated
     public void setLastFullSyncEndTime(Long lastFullSyncEndTime) {
         this.lastFullSyncEndTime = lastFullSyncEndTime;
     }
 
+    public Operation getLastRequestedFullSync() {
```
giladwolff: javadoc

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -63,22 +69,58 @@ public void setUmsEventGenerationIds(Json umsEventGenerationIds) {
...
+        return lastRequestedFullSync;
+    }
+
+    public void setLastRequestedFullSync(Operation lastRequestedFullSync) {
```
giladwolff: javadoc

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -63,22 +69,58 @@ public void setUmsEventGenerationIds(Json umsEventGenerationIds) {
...
+        this.lastRequestedFullSync = lastRequestedFullSync;
+    }
+
+    public Operation getLastSuccessfulFullSync() {
```
giladwolff: javadoc

freeipa/src/main/java/com/sequenceiq/freeipa/entity/UserSyncStatus.java
```diff
@@ -63,22 +69,58 @@ public void setUmsEventGenerationIds(Json umsEventGenerationIds) {
...
+        return lastSuccessfulFullSync;
+    }
+
+    public void setLastSuccessfulFullSync(Operation lastSuccessfulFullSync) {
```
giladwolff: javadoc

lacikaaa:
@giladwolff why would these methods require javadoc? I won't add any if not really really necessary.

handavid:
I'll rename lastRequestedFullSync to lastStartedFullSync. That is more accurate.

giladwolff:
> @giladwolff why would these methods require javadoc? I won't add any if not really really necessary.

Well, the other methods have javadoc so either remove the javadoc from all the methods or add javadoc to these.

handavid:
> > @giladwolff why would these methods require javadoc? I won't add any if not really really necessary.
> 
> Well, the other methods have javadoc so either remove the javadoc from all the methods or add javadoc to these.

ah. checkStyle (or spotBugs, I forget which) requires javadoc when marking something @Deprecated. That's why those other methods have javadoc.

PR #7586, opened 2020-03-13
Cb 5341 Update DNS entries at cluster start to new IP addresses

biharitomi:


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/start/ClusterStartService.java
```diff
@@ -44,6 +47,22 @@
     @Inject
     private InstanceMetaDataService instanceMetaDataService;
 
+    @Inject
+    private ClusterPublicEndpointManagementService clusterPublicEndpointManagementService;
+
+    public void updateDnsEntriesInPem(Stack stack) {
+        if (clusterPublicEndpointManagementService.manageCertificateAndDnsInPem()) {
+            clusterPublicEndpointManagementService.start(stack);
```
akanto: Looks good, can you add a few loglines in this if statement, please?

PR #7593, opened 2020-03-16
CB-6105 delete kubernetes and mpack deadcode (in order to keep compat…

akanto:
- Dead kubernetes and mpack code is eliminated 
- I have kept Vault + Kubernetes integration
- I have kept the DB tables in order to keep compatibility


doktoric:
I guess this is also should be updated on the UI cc @zoltanzseli

PR #7596, opened 2020-03-16
CB-6084 add verifyPermissions flag to credential

sodre90:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
```diff
@@ -135,11 +98,53 @@ private CloudCredentialStatus verifyAccessKeySecretKeyIsAssumable(CloudCredentia
...
+            try {
+                awsCredentialVerifier.validateAws(awsCredential);
+            } catch (AwsPermissionMissingException e) {
+                return new CloudCredentialStatus(cloudCredential, CredentialStatus.PERMISSIONS_MISSING, new Exception(e.getMessage()), e.getMessage());
+            }
         }
         return new CloudCredentialStatus(cloudCredential, CredentialStatus.VERIFIED);
     }
+
+    @Override
```
doktoric: as I see this method did not changed but on an other. Can we put back? This is just noise in the PR

environment/src/main/resources/schema/app/20200316133300_CB-6084_add_verify_to_credential.sql
```diff
@@ -0,0 +1,10 @@
+-- // CB-6084 add verify column to credential
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE credential ADD COLUMN IF NOT EXISTS verifyPermissions boolean default true;
```
doktoric: not null ?

sodre90: good catch, thanks

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
sodre90: yes, idea rearranged the code and I didn't recognize, thanks

PR #7597, opened 2020-03-16
CB-6121 Fix child environment stop

Bajzathd:


environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
```diff
@@ -33,7 +35,9 @@ public String selector() {
     public void accept(Event<EnvironmentDto> environmentDtoEvent) {
         EnvironmentDto environmentDto = environmentDtoEvent.getData();
         try {
-            freeIpaPollerService.stopAttachedFreeipaInstances(environmentDto.getId(), environmentDto.getResourceCrn());
+            if (Objects.isNull(environmentDto.getParentEnvironmentCrn())) {
```
doktoric: Strings.isNullOrEmtpy?

Bajzathd: Nice catch, done both here and in the other PR

PR #7598, opened 2020-03-16
CB-5904: Use CA DNS name for Kerberos and LDAP

jamisonbennett:
The kerberos configuration and ldap configuration will use a DNS load
balanced entry. This will allow a failed client to reconnect to a
different server.

Some of the unit tests were updated and it was manually tested with an
LDAP client. The LDAP client tries the first DNS entry and on failure,
it can retry. FreeIPA returns the entries in a random order, so it
will eventually try a working node. Java can be configured to cache
good and bad nodes.

Closes #CB-5904

### DIFF
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaModelDescriptions.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaModelDescriptions.java
@@ -89,5 +89,7 @@ public class FreeIpaModelDescriptions {
         public static final String HOSTNAME = "Base hostname for FreeIPA servers";
         public static final String ADMIN_GROUP_NAME = "Name of the admin group to be used for all the services.";
         public static final String TAGS = "Tags on freeipa.";
+        public static final String FREEIPA_HOST = "A DNS load balanced FQDN to the FreeIPA servers";
+        public static final String FREEIPA_PORT = "The port for the load balanced FQDN to the FreeIPA servers";
     }
 }
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
@@ -468,6 +468,19 @@ public class FreeIpaClient {
         return (DnsZone) invoke("dnszone_add", flags, params, DnsZone.class).getResult();
     }
 
+    public DnsRecord getDnsRecord(String dnsZoneName, String recordName) throws FreeIpaClientException {
+        List<String> flags = List.of(dnsZoneName, recordName);
+        Map<String, Object> params = Map.of();
+        return (DnsRecord) invoke("dnsrecord_show", flags, params, DnsRecord.class).getResult();
+    }
+
+    public DnsRecord addDnsCnameRecord(String dnsZoneName, String recordName, String cnameRecord) throws FreeIpaClientException {
+        List<String> flags = List.of(dnsZoneName, recordName);
+        Map<String, Object> params = Map.of(
+                "cnamerecord", Set.of(cnameRecord + "."));
+        return (DnsRecord) invoke("dnsrecord_add", flags, params, DnsRecord.class).getResult();
+    }
+
     public RPCResponse<Object> deleteDnsZone(String... dnsZoneNames) throws FreeIpaClientException {
         List<String> flags = List.of(dnsZoneNames);
         Map<String, Object> params = Map.of();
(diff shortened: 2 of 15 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
```diff
@@ -23,4 +23,13 @@ ipa-server-install \
...
           --unattended
 
+# Setup basic CNAME records for pointing to FreeIPA services
+echo "$FPW" | kinit admin
+ipa dnsrecord-add $DOMAIN kdc --cname-rec="ipa-ca.$DOMAIN."
+ipa dnsrecord-add $DOMAIN kerberos --cname-rec="ipa-ca.$DOMAIN."
+ipa dnsrecord-add $DOMAIN ldap --cname-rec="ipa-ca.$DOMAIN."
+ipa dnsrecord-add $DOMAIN freeipa --cname-rec="ipa-ca.$DOMAIN."
+ipa dnsrecord-add $DOMAIN ipa --cname-rec="ipa-ca.$DOMAIN."
+kdestroy
```
lacikaaa: please do it from java

jamisonbennett:
> whats the difference between `freeIpaHost` and `ipaHost`? Same for the port

I got rid of "ipa". The thought was it could prepare for another type of IPA server other than FreeIPA. But now I am thinking that there will be too much else that would need to change. So I removed "ipa".

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/doc/FreeIpaModelDescriptions.java
```diff
@@ -89,5 +89,9 @@ private FreeIpaModelDescriptions() {
         public static final String HOSTNAME = "Base hostname for FreeIPA servers";
         public static final String ADMIN_GROUP_NAME = "Name of the admin group to be used for all the services.";
         public static final String TAGS = "Tags on freeipa.";
+        public static final String FREEIPA_HOST = "A DNS load balanced FQDN to the FreeIPA servers";
+        public static final String FREEIPA_PORT = "The port for the load balanced FQDN to the FreeIPA servers";
+        public static final String IPA_HOST = "A DNS load balanced FQDN to the FreeIPA servers";
+        public static final String IPA_PORT = "The port for the load balanced FQDN to the FreeIPA servers";
```
lacikaaa: please remove these too

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -468,6 +468,20 @@ public DnsZone addReverseDnsZone(String cidr) throws FreeIpaClientException {
...
     }
 
+    public DnsRecord getDnsRecord(String dnsZoneName, String recordName) throws FreeIpaClientException {
+        List<String> flags = List.of(dnsZoneName, recordName);
+        Map<String, Object> params = Map.of();
+        return (DnsRecord) invoke("dnsrecord_show", flags, params, DnsRecord.class).getResult();
+    }
+
+    public DnsRecord addDnsCnameRecord(String dnsZoneName, String recordName, String cnameRecord) throws FreeIpaClientException {
+        LOGGER.error("ASDJ CNAME " + dnsZoneName + " -- " + recordName + "  -- " + cnameRecord + " -- ");
```
lacikaaa: why error? `asdj`?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -95,4 +104,35 @@ private void modifyAdminPasswordExpirationIfNeeded(FreeIpaClient client) throws
...
+        try {
+            client.getDnsRecord(domain, cname);
+            return true;
+        } catch (FreeIpaClientException e) {
+            if (FreeIpaClientExceptionUtil.isNotFoundException(e)) {
+                return false;
+            }
+            throw e;
+        }
+    }
```
lacikaaa: could you move these to a separate class?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
jamisonbennett: I removed this log statement. The caller of this method logs debug messages.

lacikaaa:
retest this please

jamisonbennett:
I rebased to the lastest master because the integration tests that had previously passed were failing.

jamisonbennett:
I added `DnsRecordAddResponse.java`  and `DnsRecordShowResponse.java` for the integration test.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
wonderslug: I would like to understand why we needed to take something that was 8 lines in a salt script, was only going to be run once on an install of the server and move it to Java.  Now we have a lot more code to maintain.  Isn't this what using salt as the node management supposed to give us.  I am just wanting to understand where the demarkation oh what is supposed to be controlled by java and what the salt scripts are supposed to do.

keyki: Right now we cannot update the Salt states on an existing cluster. If we need to change here something that we cannot do.

keyki: Think of it as Cloudera Manager. We could install a cluster from Salt also, but we use its API. We would lose lots of capabilities if we would implement everything in Salt.

PR #7609, opened 2020-03-18
CB-6125 creator of an environment will automaticly get enviromentAdmin

schfeca75:
@keyki or @horadla23 pls review

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
@@ -106,6 +106,8 @@ public class EnvironmentCreationService {
             // To keep backward compatibility, if somebody passes the group name, then we shall just use it
             environmentService.setAdminGroupName(environment, creationDto.getAdminGroupName());
         }
+        environmentService.assignEnvironmentAdminRole(creationDto.getCreator(), environmentCrnForVirtualGroups);
+
         CloudRegions cloudRegions = setLocationAndRegions(creationDto, environment);
         Map<String, CloudSubnet> subnetMetas = networkService.retrieveSubnetMetadata(environment, creationDto.getNetwork());
         validateCreation(creationDto, environment, cloudRegions, subnetMetas);
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -534,8 +534,8 @@ public class GrpcUmsClient {
                         .build());
     }
 
-    private UmsClient makeClient(ManagedChannel channel, String accountId) {
-        return new UmsClient(channel, accountId, umsClientConfig);
+    private UmsClient makeClient(ManagedChannel channel, String actorCrn) {
+        return new UmsClient(channel, actorCrn, umsClientConfig);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
```diff
@@ -59,6 +59,19 @@ public void cleanupVirtualGroups(String accountId, String environmentCrn) {
...
 
+    public void assignEnvironmentAdminRole(String userCrn, String environmentCrn) {
+        try {
+            grpcUmsClient.assignResourceRole(userCrn, environmentCrn, grpcUmsClient.getBuiltInEnvironmentAdminResourceRoleCrn(), MDCUtils.getRequestId());
+            LOGGER.debug("EnvironmentAdmin role of {} environemnt is assigned to the {} user", environmentCrn, userCrn);
+        } catch (StatusRuntimeException ex) {
+            if (Code.ALREADY_EXISTS.equals(ex.getStatus().getCode())) {
+                LOGGER.debug("EnvironmentAdmin role of {} environemnt is already assigned to the {} user", environmentCrn, userCrn);
+            } else {
+                throw ex;
```
keyki: I don't think that we should make the provisioning fail here. Simply logging the error would do.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -106,6 +106,8 @@ public EnvironmentDto create(EnvironmentCreationDto creationDto) {
             // To keep backward compatibility, if somebody passes the group name, then we shall just use it
             environmentService.setAdminGroupName(environment, creationDto.getAdminGroupName());
         }
+        virtualGroupService.assignEnvironmentAdminRole(creationDto.getCreator(), environmentCrnForVirtualGroups);
```
keyki: This is not related to virtual group at all. It should be moved to some other proper place.

keyki:
There is no un-assignment here. I don't know if it's needed or not. @horadla23 might know.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
```diff
@@ -59,6 +59,19 @@ public void cleanupVirtualGroups(String accountId, String environmentCrn) {
         }
     }
 
+    public void assignEnvironmentAdminRole(String userCrn, String environmentCrn) {
+        try {
+            grpcUmsClient.assignResourceRole(userCrn, environmentCrn, grpcUmsClient.getBuiltInEnvironmentAdminResourceRoleCrn(), MDCUtils.getRequestId());
+            LOGGER.debug("EnvironmentAdmin role of {} environemnt is assigned to the {} user", environmentCrn, userCrn);
+        } catch (StatusRuntimeException ex) {
+            if (Code.ALREADY_EXISTS.equals(ex.getStatus().getCode())) {
```
horadla23: I didn't know about this code, I would move this exception handling into GrpcUmsClient assignResourceRole method

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
horadla23: I agree, assignment logic should be in the service of the resource, in this case in some EnvironmentService class I assume

schfeca75: Moved to the environmentService

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/VirtualGroupService.java
schfeca75: I would  left this code here as it is specific during environment creation, maybe later for other resources it could be an error case.

horadla23: ok, it is fine for me

PR #7619, opened 2020-03-20
CB-6161 Fix HDFS user home directory creation for HA configuration

lacikaaa:


orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -71,25 +71,47 @@ PATH_PREFIX="/user"
...
+  NN_HOST=$(hdfs haadmin -getAllServiceState | grep active | cut -d':' -f 1)
+  echo "Active NAMENODE host: $NN_HOST"
+  if [ "$NN_HOST" != "$(hostname -f)" ]; then
+    echo "This is not the active host. Script will run on the active one"
+    exit 0
+  fi
+  if [ "$WEBHDFS_HTTP_POLICY" == "HTTP_ONLY" ]; then
+    NAMENODE_PORT=$(hdfs getconf -confkey dfs.namenode.http-address | cut -d':' -f 2)
+    echo "NN port: $NAMENODE_PORT"
+    WEBHDFS_URL=http://$NN_HOST$NAMENODE_PORT/webhdfs/v1
```
risdenk: I'm guessing you need the colon here between host and port if I'm reading this correctly?

lacikaaa: yep, fixed. good catch!

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -71,25 +71,47 @@ PATH_PREFIX="/user"
...
+    NAMENODE_HTTP_ADDRESS=$(hdfs getconf -confkey dfs.namenode.https-address)
+    WEBHDFS_URL=https://$NAMENODE_HTTP_ADDRESS/webhdfs/v1
+  fi
 else
-  NAMENODE_HTTP_ADDRESS=$(hdfs getconf -confkey dfs.namenode.https-address)
-  WEBHDFS_URL=https://$NAMENODE_HTTP_ADDRESS/webhdfs/v1
+  echo "HA configuration"
+  NN_HOST=$(hdfs haadmin -getAllServiceState | grep active | cut -d':' -f 1)
+  echo "Active NAMENODE host: $NN_HOST"
+  if [ "$NN_HOST" != "$(hostname -f)" ]; then
```
risdenk: Do we need this check in the case of non HA? This to me looks like this script will be installed on all the nodes or maybe just the namenodes?

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -71,25 +71,47 @@ PATH_PREFIX="/user"
...
+if [[ $? -ne 0 ]]; then
+  kinit_as_hdfs
+fi
+
 WEBHDFS_HTTP_POLICY=$(hdfs getconf -confkey dfs.http.policy)
 
-if [ "$WEBHDFS_HTTP_POLICY" == "HTTP_ONLY" ]; then
-  NAMENODE_HTTP_ADDRESS=$(hdfs getconf -confkey dfs.namenode.http-address)
-  WEBHDFS_URL=http://$NAMENODE_HTTP_ADDRESS/webhdfs/v1
+hdfs haadmin -getAllServiceState
```
risdenk: nit: you can put this in the `if` statement itself to check status code instead of a separate line if you wanted.

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
lacikaaa: this will be on every, but only namenodes, and will run there, so if they switch the active one for any reason we don't have to do anything from cb side

petersomogyi:
Is it possible to cherry-pick this commit to CB 2.19?

lacikaaa:
if it's need for 2.19 I will open against it, but in the jira the version is 2.20

petersomogyi:
The bug is there on 2.19 as well, probably even on earlier releases. @akanto set the fix version to 2.20 but I this bug is severe enough to land in 2.19.

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
lacikaaa: I fixed this, could you check and resolv if you find it ok?

risdenk: Looks good - didn't run it - but surface looks good.

lacikaaa:
retest this please

orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
```diff
@@ -71,25 +71,47 @@ PATH_PREFIX="/user"
...
+    WEBHDFS_URL=https://$NAMENODE_HTTP_ADDRESS/webhdfs/v1
+  fi
 else
-  NAMENODE_HTTP_ADDRESS=$(hdfs getconf -confkey dfs.namenode.https-address)
-  WEBHDFS_URL=https://$NAMENODE_HTTP_ADDRESS/webhdfs/v1
+  echo "HA configuration"
+  NN_HOST=$(hdfs haadmin -getAllServiceState | grep active | cut -d':' -f 1)
+  echo "Active NAMENODE host: $NN_HOST"
+  if [ "$NN_HOST" != "$(hostname -f)" ]; then
+    echo "This is not the active host. Script will run on the active one"
```
toddlipcon: this doesn't seem like a robust approach. I have two concerns:
1) is it possible that this runs concurrently with the namenodes starting, in which case neither NN is active yet?
2) it's definitely possible that one of the NNs fails around the same time these two scripts are running. In that case either neither or both of the copies of this script will run.

it seems this script should be idempotent. How often does it get re-run? eg if for some reason we hit one of the above issues, would it run again a minute later and just be a transient problem, or would we be left in a persistent unsynced state?

lacikaaa: it will start on all namenodes, but will exit if it's not the active one. this is in case of HA. (if there is no active host it won't run)
if not HA it will work as it used to.

it's in cron, so it runs periodically.

I haven't touched the core logic, only the address calculation in case of HA. it was written mostly by Deveraj and you. But if we need more modification I can make it.

PR #7629, opened 2020-03-23
CB-6203 only sync workload admin groups that are related to environment

keyki:


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -166,7 +165,13 @@ private void handleUser(UmsUsersState.Builder umsUsersStateBuilder, UsersState.B
                     }
                 });
 
-                workloadAdministrationGroups.getWorkloadAdministrationGroupNameList().forEach(groupName -> {
+                // Only add workload admin groups that belong to this environment
+                workloadAdminGroups.stream()
+                        .filter(wag -> wag.getResource().equalsIgnoreCase(environmentAccessRights.getEnvironmentCrn()))
```
handavid: this doesn't work. WAGs may be associated with resources that are not environments. We don't currently have a mapping of resources to environments, so we can't tell if a WAG is applicable to a given environment.

If we could use the WAG resource, then we could do the filtering on the wags list in `getEnvToUmsUsersStateMap` instead of in `handleUser`

keyki: So this means we cannot filter at all? I didn't know that there could be other wags, but in this case, it only works for those that CB creates, right?

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/MockGroupManagementService.java
```diff
@@ -16,120 +16,151 @@
...
 
     private static final String CM_ADMIN_RIGHT = "environments/adminClouderaManager";
 
+    private static final int NUM_USER_GROUPS = 5;
+
     @Inject
     private MockCrnService mockCrnService;
 
-    private final Map<String, Map<String, UserManagementProto.Group>> accountGroups = new ConcurrentHashMap<>();
+    private final Map<String, Map<String, Group>> accountWorkloadGroups = new ConcurrentHashMap<>();
```
handavid: WAGs should be modeled as `WorkloadAdministrationGroup`s rather than `Group`s. WAGs don't have crns, only names, accountids, rights, and resources. Moving to `WorkloadAdministrationGroup` would probably make the WAG-related mock code more sensible.

PR #7632, opened 2020-03-23
CB-6206 nifi.web.proxy.context.path needs to include all Knox topology names

risdenk:
Adds all the Knox topology names to the `nifi.web.proxy.context.path`


risdenk:
Rebased on master to rerun tests.

PR #7633, opened 2020-03-23
CB-6225: health check not returning status of all hosts.

holleyism:
This patch changes the health check to return individual
health of all HA nodes. The cn of the server_conncheck API
must be the host that the client is connected to. So we
exposed the connected hostname to use as the first parameter
of the API call.

Updated unit test.

Manually validated against 3 node FreeIPA and various levels of failure.

Also fixes:
CB-6226

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
@@ -1,7 +1,7 @@
 package com.sequenceiq.freeipa.service.stack;
 
+import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -13,7 +13,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.DetailedStackStatus;
-import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceGroupType;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.health.HealthDetailsFreeIpaResponse;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.health.NodeHealthDetails;
@@ -21,7 +20,6 @@ import com.sequenceiq.freeipa.client.FreeIpaClient;
 import com.sequenceiq.freeipa.client.FreeIpaClientException;
 import com.sequenceiq.freeipa.client.model.RPCMessage;
 import com.sequenceiq.freeipa.client.model.RPCResponse;
-import com.sequenceiq.freeipa.entity.InstanceGroup;
 import com.sequenceiq.freeipa.entity.InstanceMetaData;
 import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.service.freeipa.FreeIpaClientFactory;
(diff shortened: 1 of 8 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -45,6 +45,10 @@
 
     private static final Pattern NEW_NODE_PATTERN = Pattern.compile("Check connection from master to remote replica '(.[^\']*)");
 
+    private static final String WRONG_MASTER_MESSAGE = "invalid 'cn': must be";
+
+    private static final int MAX_RETRIES = 3;
```
lacikaaa: why 3?
also it should be configurable

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +57,69 @@
 
     public HealthDetailsFreeIpaResponse getHealthDetails(String environmentCrn, String accountId) {
         Stack stack = stackService.getByEnvironmentCrnAndAccountIdWithLists(environmentCrn, accountId);
-        InstanceMetaData master = findMaster(stack);
-        Optional<RPCResponse<Boolean>> rpcResponse = Optional.empty();
-        try {
-            rpcResponse = Optional.ofNullable(checkFreeIpaHealth(stack, master));
-        } catch (FreeIpaClientException e) {
-            LOGGER.error("Unable to check the health of FreeIPA.", e);
+        List<InstanceMetaData> instances = stack.getInstanceGroups().stream()
```
lacikaaa: you could use Stack#getNotDeletedInstanceMetaDataList if you don't need the deleted ones, but if you really need terminated instances I think we should add this stream as a method to Stack as it seems to me we might need it multiple times

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +57,69 @@
...
+                .flatMap(instanceGroup -> instanceGroup.getInstanceMetaData().stream()).collect(Collectors.toList());
+        HealthDetailsFreeIpaResponse response = new HealthDetailsFreeIpaResponse();
+        // The code here is weird because the server_conncheck API requires a specific node
+        // and we don't know what determines what that node needs to be.  But if we use
+        // the wrong master, we get an exception that says
+        //   invalid 'cn': must be "ipaserver0.xx.wl.cloudera.site"
+        // so we parse that, set that as master, and then retry the node.
```
lacikaaa: why don't we know which instance we should use?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +57,69 @@
...
+        String masterCn = findMaster(stack).getDiscoveryFQDN();
+        int retries = 0;
+        for (int idx = 0; idx < instances.size(); idx++) {
```
lacikaaa: you should use `for (InstanceMetaData instance : instances)` as you don't use the index elsewhere

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +57,69 @@
...
+            if (!isInstanceTerminatedOrDeleted(instance, response)) {
+                try {
+                    RPCResponse<Boolean> rpcResponse = checkFreeIpaHealth(stack, masterCn, instance);
+                    parseMessages(rpcResponse, response);
+                } catch (FreeIpaClientException e) {
+                    String msg = e.getLocalizedMessage();
+                    if (msg.contains(WRONG_MASTER_MESSAGE) && retries < MAX_RETRIES) {
+                        masterCn = msg.substring(msg.indexOf('"') + 1);
+                        masterCn = masterCn.substring(0, masterCn.indexOf('"'));
+                        instances.add(instance);
```
lacikaaa: so you add it to the list to check it again or is it a bug?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +57,69 @@
...
+                        retries++;
+                    } else {
+                        NodeHealthDetails nodeResponse = new NodeHealthDetails();
```
lacikaaa: if we get into this branch, shouldn't we exit the for loop?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: I want to report status on all of them.  I'll add it to Stack.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: It's not stored in the stack data, and I haven't been able to determine which one it should be.  If we have multi-master setup, it's not clear which one is the real master.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: I cannot update the list of instances if using this style for loop.  Per next comment, I add it back to the list to retry since I cannot idx--.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: Yes.  I add it back to the list to retry because we don't have a proper health yet.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: No because this is just one instance.  If we can't reach one instance, we still need to contact the others to get their status.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: I think this is something we'll need to add in the next round of HA work.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: Made configurable.  I made it three in case we're in the middle of a replication change or something else that switches the value.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +58,49 @@
...
-        response.setCrn(stack.getResourceCrn());
-        if (rpcResponse.isPresent()) {
-            response.setName((String) rpcResponse.get().getValue());
-            parseMessages(rpcResponse.get(), response);
-            if (isOverallHealthy(response)) {
-                response.setStatus(DetailedStackStatus.PROVISIONED.getStatus());
-            } else {
-                response.setStatus(DetailedStackStatus.UNHEALTHY.getStatus());
-            }
+    private boolean isInstanceAvailable(InstanceMetaData instance, HealthDetailsFreeIpaResponse response) {
```
lacikaaa: the method name doesn't reflect what it does. I think `instance.isAvailable()` should be in the caller if, and the else branch from there should call the body of this method (wo the if here oc)

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -53,35 +58,49 @@
...
+                try {
+                    RPCResponse<Boolean> rpcResponse = checkFreeIpaHealth(stack,  instance);
+                    parseMessages(rpcResponse, response);
+                } catch (FreeIpaClientException e) {
+                    NodeHealthDetails nodeResponse = new NodeHealthDetails();
+                    response.addNodeHealthDetailsFreeIpaResponses(nodeResponse);
+                    nodeResponse.setName(instance.getDiscoveryFQDN());
+                    nodeResponse.setStatus(InstanceStatus.UNREACHABLE);
+                    nodeResponse.addIssue(e.getLocalizedMessage());
+                    LOGGER.error(String.format("Unable to check the health of FreeIPA instance: %s", instance.getInstanceId()), e);
```
lacikaaa: refactor into a method pls

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -45,6 +45,11 @@
 
     private static final Pattern NEW_NODE_PATTERN = Pattern.compile("Check connection from master to remote replica '(.[^\']*)");
 
+    private static final String WRONG_MASTER_MESSAGE = "invalid 'cn': must be";
+
+    @Value("${freeipa.healthcheck.max-retries:3}")
+    private int maxRetries;
```
lacikaaa: this is not needed anymore I think

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: changed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: refactored

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: removed

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
holleyism: changed.

PR #7634, opened 2020-03-23
CDPCP-1560. Allow retrieving keytab for different users

aarman-cloudera:
This allows a user to retrieve keytab that belong to other users
as long as they have the given right. This type of authorization
check doesn't seem to fit the existing authorization annotations
so I've instead put the logic in an AuthorizationUtil class.


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.inject.Inject;
+
+@Component
+public class AuthorizationUtil {
+
+    @Inject
+    private GrpcUmsClient grpcUmsClient;
+
+    /**
+     * An authorization check that verifies that a calling actor can perform an operation against a
```
giladwolff: I would rephrase a bit saying that the check checks that the calling actor is either perform an action against themselves or have the right.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
```diff
@@ -0,0 +1,34 @@
...
+     * target user. The calling actor only requires the right when the target user is different.
+     */
+    public void checkCallerIsSelfOrHasRight(String actorCrnStr, String targetUserCrnStr, String requiredRight) {
+        Crn actorCrn = Crn.safeFromString(actorCrnStr);
```
giladwolff: assert the parameters?

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
```diff
@@ -0,0 +1,34 @@
...
+        Crn targetUserCrn = Crn.safeFromString(targetUserCrnStr);
+        if (actorCrn.equals(targetUserCrn)) {
+            return;
+        }
+        if (!actorCrn.getAccountId().equals(targetUserCrn.getAccountId())) {
+            throw new AccessDeniedException("Unauthorized to run this operation in a different account");
+        }
+        if (!grpcUmsClient.checkRight(GrpcUmsClient.INTERNAL_ACTOR_CRN, actorCrn.toString(), requiredRight, MDCUtils.getRequestId())) {
```
giladwolff: I think that checkRight throws if the check fails.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
```diff
@@ -0,0 +1,34 @@
...
+public class AuthorizationUtil {
+
+    @Inject
+    private GrpcUmsClient grpcUmsClient;
+
+    /**
+     * An authorization check that verifies that a calling actor can perform an operation against a
+     * target user. The calling actor only requires the right when the target user is different.
+     */
+    public void checkCallerIsSelfOrHasRight(String actorCrnStr, String targetUserCrnStr, String requiredRight) {
```
horadla23: we have `UmsAuthorizationService` class, which has methods for checking rights through UMS, you should move this method to that class and maybe you can also reuse those existing methods

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
```diff
@@ -43,6 +45,9 @@
     @Inject
     private UserKeytabService userKeytabService;
 
+    @Inject
+    private AuthorizationUtil authorizationUtil;
```
horadla23: as mentioned, please move your logic to `UmsAuthorizationService` and use that

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
aarman-cloudera: The safeFromString does a null check for us.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
aarman-cloudera: So the grpcUmsClient checks for exception and returns a boolean instead.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/AuthorizationUtil.java
aarman-cloudera: Thanks! didn't know about that service. I've moved thee method over.

PR #7637, opened 2020-03-24
CB-6218 remove stackowner

doktoric:


lacikaaa:
could you fix CloudStack#toString also as I mentioned in the jira?

doktoric:
> could you fix CloudStack#toString also as I mentioned in the jira?

next please open a separate ticket for that :)

PR #7640, opened 2020-03-24
CDPCP-1669. Verify user exists in an environment before getting keytab

aarman-cloudera:
Currently it's possible for the user to provide a user that isn't
present (hasn't been synced) into an environment. In this case we
throw an appropriate exception.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
@@ -4,9 +4,15 @@ import static com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient.INTERNAL_ACTOR_
 import static com.sequenceiq.freeipa.controller.exception.NotFoundException.notFound;
 
 import java.util.List;
+import java.util.Optional;
 
 import javax.inject.Inject;
 
+import com.sequenceiq.freeipa.client.FreeIpaClient;
+import com.sequenceiq.freeipa.client.FreeIpaClientException;
+import com.sequenceiq.freeipa.client.model.User;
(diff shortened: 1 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
```diff
@@ -48,6 +56,21 @@ private void validateSameAccount(String userAccountId, String environmentCrn) {
...
     }
 
+    private void validateWorkloadUserInEnvironment(String workloadUsername, String environmentCrn) {
+        String accountId = Crn.safeFromString(environmentCrn).getAccountId();
+        FreeIpaClient freeIpaClient;
+        try {
+            freeIpaClient = freeIpaClientFactory.getFreeIpaClientByAccountAndEnvironment(environmentCrn, accountId);
+            Optional<User> user = freeIpaClient.userFind(workloadUsername);
+            if (user.isEmpty()) {
+                throw new BadRequestException(String.format("Could not find workload user %s within environment %s. Ensure that the user has " +
```
giladwolff: Maybe just say: Actor %s has not been synced into environment %s"

Btw, what BadRequestException translates to the user?

lacikaaa: yeah maybe a notfoundex would be better

aarman-cloudera: BadRequestException is 400 bad request. I've updated the message (though I used workload user instead of actor her). And changed to NotFoundException.

aarman-cloudera:
> LGTM. Are we concern at all with the additional load on the FreeIPA?

I don't expect this endpoint/service to be called too often so I'd say no.

PR #7641, opened 2020-03-24
CB-6248: Add subject alt names to DNS load balanced certificates

jamisonbennett:
Fix LDAP and FreeIPA x509v3 certificates to have the subject
alternative names. This fixes errors with clients that use the DNS
load balance services.

The LDAP certificate now has ldap.domain entry.
The FreeIPA API now has freeipa.domain, kdc.domain, and
kerberos.domain.

Closes #CB-5904
Closes #CB-6147
Closes #CB-6248


jamisonbennett:
Updated the salt script to be idempotent for adding the cnames.

PR #7643, opened 2020-03-25
CB-6251: Could not reboot stopped FreeIPA

holleyism:
This change adjusts the reboot to properly reboot only started instances and
to call start on stopped instances. Because of AWS lifecycle
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
other statuses cannot be started or rebooted.

Manually tested against 3 node FreeIPA with instances in various states, including
started, stopped, terminated.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
@@ -133,18 +133,63 @@ public class AwsInstanceConnector implements InstanceConnector {
     public List<CloudVmInstanceStatus> reboot(AuthenticatedContext ac, List<CloudInstance> vms) {
         AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
                 ac.getCloudContext().getLocation().getRegion().value());
+        List<CloudInstance> affectedVms = new ArrayList<>();
         try {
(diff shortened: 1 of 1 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -133,18 +133,52 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
         try {
-            Collection<String> instances = vms.stream().map(CloudInstance::getInstanceId).collect(Collectors.toSet());
-            if (!instances.isEmpty()) {
-                amazonEC2Client.rebootInstances(new RebootInstancesRequest().withInstanceIds(instances));
+            if (!vms.isEmpty()) {
+                List<CloudVmInstanceStatus> statuses = check(ac, vms);
+                doReboot(affectedVms, amazonEC2Client, statuses.stream().filter(status -> status.getStatus() == InstanceStatus.STARTED)
+                        .map(status -> status.getCloudInstance()).collect(Collectors.toList()));
+                doStart(affectedVms, ac, statuses.stream().filter(status -> status.getStatus() == InstanceStatus.STOPPED)
+                        .map(status -> status.getCloudInstance()).collect(Collectors.toList()));
```
lacikaaa: could you move these stream parameters into variables or even to methods?

holleyism: moved

PR #7644, opened 2020-03-25
CB-5487 - Knox - Oozie API doesn't show in gateway tab

gregito:
CB-5487 - Knox - Oozie API doesn't show in gateway tab


core-api/src/main/java/com/sequenceiq/cloudbreak/api/service/ExposedServiceCollector.java
```diff
@@ -99,6 +99,10 @@ public ExposedService getResourceManagerWebService() {
         return exposedServices.get("RESOURCEMANAGER_WEB");
     }
 
+    public ExposedService getOozie() {
```
doktoric: do we using this method ?

gregito: nope, got rid of it

doktoric:
It failed

PR #7645, opened 2020-03-25
CB-5660 Respect declared Role Config Groups in Data Hub template gene…

topolyai5:
I made some simple refactor. The new logic is here:
https://github.com/hortonworks/cloudbreak/compare/cb5660-unify-component-by-blacklist?expand=1#diff-34a9a5ca63de6b681c8d82d9af485f9bR91

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.cmtemplate;
 
+import static com.sequenceiq.cloudbreak.util.FileReaderUtils.readFileFromClasspath;
+import static java.util.Arrays.asList;
 import static java.util.Optional.ofNullable;
 import static java.util.stream.Collectors.toMap;
 
@@ -10,6 +12,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
 
+import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
 import org.apache.commons.lang3.tuple.Pair;
@@ -23,6 +26,9 @@ import com.cloudera.api.swagger.model.ApiClusterTemplateInstantiator;
 import com.cloudera.api.swagger.model.ApiClusterTemplateRoleConfigGroup;
 import com.cloudera.api.swagger.model.ApiClusterTemplateRoleConfigGroupInfo;
 import com.cloudera.api.swagger.model.ApiClusterTemplateService;
+import com.google.common.annotations.VisibleForTesting;
+import com.sequenceiq.cloudbreak.cmtemplate.sharedcomponent.SharedComponent;
+import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.template.TemplatePreparationObject;
 import com.sequenceiq.cloudbreak.template.model.ServiceComponent;
 import com.sequenceiq.cloudbreak.template.views.HostgroupView;
@@ -35,6 +41,19 @@ public class CmHostGroupRoleConfigProviderProcessor {
     @Inject
     private List<CmHostGroupRoleConfigProvider> providers;
 
+    private List<SharedComponent> sharedComponents = new ArrayList<>();
+
+    @PostConstruct
+    private void parseSharedComponents() {
+        try {
+            String configJson = readFileFromClasspath("shared-components-config.json");
+            sharedComponents = asList(new Json(configJson).get(SharedComponent[].class));
+            LOGGER.info("shared components are loaded: {}", sharedComponents);
+        } catch (Exception e) {
+            LOGGER.warn("Cannot read shared components from: `shared-components-config.json`. Message: {}", e.getMessage());
+        }
+    }
+
     public void process(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
         if (!getHostTemplates(templateProcessor).isEmpty()) {
             updateConfigsInTemplate(templateProcessor, generateConfigs(templateProcessor, source));
@@ -45,11 +64,10 @@ public class CmHostGroupRoleConfigProviderProcessor {
         return ofNullable(templateProcessor.getTemplate().getHostTemplates()).orElseGet(List::of);
     }
 
-    private Map<String, Map<String, List<ApiClusterTemplateConfig>>> generateConfigs(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
+    @VisibleForTesting
+    Map<String, Map<String, List<ApiClusterTemplateConfig>>> generateConfigs(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
         Map<String, Map<String, List<ApiClusterTemplateConfig>>> configsByRoleConfigGroup = new HashMap<>();
-
-        Map<String, HostgroupView> hostGroups = source.getHostgroupViews().stream()
-                .collect(toMap(HostgroupView::getName, Function.identity()));
+        Map<String, HostgroupView> hostGroups = source.getHostgroupViews().stream().collect(toMap(HostgroupView::getName, Function.identity()));
         List<ApiClusterTemplateHostTemplate> hostTemplates = getHostTemplates(templateProcessor);
         Map<String, ServiceComponent> serviceComponents = templateProcessor.mapRoleRefsToServiceComponents();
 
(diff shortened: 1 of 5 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
```diff
@@ -35,6 +38,23 @@
...
+    @Value("#{'${cb.cm.sharedComponents:}'.split(';')}")
+    private List<String> sharedComponentsList;
+
+    private Map<String, String> sharedComponents = new HashMap<>();
+
+    @PostConstruct
+    private void parseSharedComponents() {
+        sharedComponentsList.forEach(s -> {
+            try {
+                String[] split = s.split("=");
```
doktoric: I think it would be nice to see somewhere what are the configs

core/src/main/resources/application.yml
```diff
@@ -323,6 +323,8 @@ cb:
     user.role: "ROLE_USER"
     heartbeat.interval: 3
     missed.heartbeat.interval: 30
+    sharedComponents: >
+      HDFS=JOURNALNODE
```
doktoric: Just a question butmaybe we should use a config json ?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
topolyai5: Do you mean a log message?

core/src/main/resources/application.yml
attilapalfi92: I think it would add unnecessary complexity. Also it would require a rebuild to run with a new config. On the other hand this property can be modified without building a new CB version.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
```diff
@@ -45,35 +65,40 @@ public void process(CmTemplateProcessor templateProcessor, TemplatePreparationOb
...
+                    Map<String, List<ApiClusterTemplateConfig>> configs = configsByRoleConfigGroup.computeIfAbsent(roleConfigGroup, __ -> new HashMap<>());
+                    configs.computeIfAbsent(hostGroupName, __ -> new ArrayList<>())
+                            .addAll(provider.getRoleConfigs(serviceComponent.getComponent(), hostgroupView, source));
                 }
             }
         }
-        return configsByRoleConfigGroup;
+    }
+
+    private boolean isSameByServiceTypeAndRoleType(CmHostGroupRoleConfigProvider provider, ServiceComponent serviceComponent) {
```
doktoric: maybe isEqual ?

core/src/main/resources/application.yml
topolyai5: which is easier to update on the deployed version? in app.yml or in a json?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
```diff
@@ -45,35 +65,40 @@ public void process(CmTemplateProcessor templateProcessor, TemplatePreparationOb
...
-                        configsByRoleConfigGroup.computeIfAbsent(roleConfigGroup, __ -> new HashMap<>())
-                                .computeIfAbsent(hostGroupName, __ -> new ArrayList<>())
-                                .addAll(provider.getRoleConfigs(serviceComponent.getComponent(), hostgroupView, source));
-                    }
+    private void groupByHostGroupName(TemplatePreparationObject source, Map<String, Map<String, List<ApiClusterTemplateConfig>>> configsByRoleConfigGroup,
+            Map<String, ServiceComponent> serviceComponents, String hostGroupName, List<String> roleConfigGroups, HostgroupView hostgroupView) {
+        for (String roleConfigGroup : roleConfigGroups) {
+            for (CmHostGroupRoleConfigProvider provider : providers) {
+                ServiceComponent serviceComponent = serviceComponents.get(roleConfigGroup);
+                if (serviceComponent != null
```
attilapalfi92: Very large if condition. Can you please extract it?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
doktoric: yep

core/src/main/resources/application.yml
doktoric: I think it is the same. I am just saying that somehow it is not clear what is the key and the value here ?

attilapalfi92: json is built into the artifact. app.yml properties can be overwritten with env vars.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
topolyai5: what is your proposal? I need to check all the three condition in the same if

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
topolyai5: changed

topolyai5: is this better? `isEqualsByServiceTypeAndRoleType`

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
attilapalfi92: just extract this to a method an name it like areComponentsEqual or something like that

topolyai5: done

core/src/main/resources/application.yml
doktoric: in cb deployer we have capabilities to override any json which contains config. It is true that we can not do the same on the hosted service but the question here is that do we really need such flexibility ? :)

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
doktoric: I think so yes

topolyai5: ok, changed

core/src/main/resources/application.yml
topolyai5: If we do not want to change on the hosted service, in this case, we do not need to this flexibility.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmHostGroupRoleConfigProviderProcessor.java
```diff
@@ -45,35 +64,50 @@ public void process(CmTemplateProcessor templateProcessor, TemplatePreparationOb
...
+    private boolean isEqualsAndNotSharedComponent(CmHostGroupRoleConfigProvider provider, ServiceComponent serviceComponent) {
+        return serviceComponent != null
+                && isEqualsByServiceTypeAndRoleType(provider, serviceComponent)
+                && !isSharedComponent(serviceComponent);
+    }
+
+    private boolean isSharedComponent(ServiceComponent serviceComponent) {
+        return sharedComponents.stream().anyMatch(s -> s.getComponentName().equals(serviceComponent.getComponent())
+                && s.getServiceType().equals(serviceComponent.getService()));
+//        return serviceComponent.getComponent().equals(sharedComponents.get(serviceComponent.getService()));
```
doktoric: can we remove this?

PR #7646, opened 2020-03-25
CB-6217 Remove redundancy from API E2E test cases

aszegedi:


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImageValidatorTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.testng.annotations.Test;
+
+import com.sequenceiq.it.cloudbreak.client.DistroXTestClient;
+import com.sequenceiq.it.cloudbreak.context.Description;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+import com.sequenceiq.it.cloudbreak.testcase.e2e.ImageValidatorE2ETest;
+
+public class DistroXImageValidatorTest extends ImageValidatorE2ETest {
```
horadla23: I think this class should not extend `ImageValidatorE2ETest`, since `InternalSdxDistroxTests` also creates DistroX cluster

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/InternalSdxTests.java
```diff
@@ -19,15 +19,6 @@
     @Inject
     private SdxTestClient sdxTestClient;
 
-    @Override
-    protected void setupTest(TestContext testContext) {
```
horadla23: This class also should not extend `ImageValidatorE2ETest`, since it is not used in imagevalidation test suite

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImageValidatorTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Inject
+    private DistroXTestClient distroXTestClient;
+
+    @Test(dataProvider = TEST_CONTEXT)
+    @Description(
+            given = "there is a running cloudbreak",
+            when = "a valid DistroX create request is sent",
+            then = "DistroX cluster is created")
+    public void testCreateDistroX(TestContext testContext) {
```
horadla23: Since image validation does not use this test currently, I think you can move it back to the DistroxTests class, it does not need to be an imagevalidation test, the only class used for image validation is `InternalSdxDistroxTests `

horadla23: Or you can simply remove this test

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/InternalSdxDistroxTest.java
```diff
@@ -17,23 +17,14 @@
 import com.sequenceiq.sdx.api.model.SdxClusterStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxDatabaseRequest;
 
-public class InternalSdxDistroxTests extends ImageValidatorE2ETest {
+public class InternalSdxDistroxTest extends ImageValidatorE2ETest {
```
horadla23: please modify class name in the image validation suites as well

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
```diff
@@ -449,13 +451,37 @@ private boolean checkSdxInstanceState(SdxClient sdxClient, String sdxName, Strin
         List<InstanceGroupV4Response> instanceGroups = sdxClient.getSdxClient().sdxEndpoint().getDetail(sdxName, Set.of())
                 .getStackV4Response()
                 .getInstanceGroups();
-        return checkInstanceState(hostGroup, desiredState, instanceGroups);
+        SdxClusterResponse sdxResponse = sdxClient.getSdxClient().sdxEndpoint().get(sdxName);
```
gergopapi2: This method and checkDistroxInstanceState are almost identical. I would try to refactor them to have one method. I think it is only the log and exception messages that differ within method body.

integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
```diff
@@ -66,7 +66,7 @@
     @Value("${integrationtest.threadCount:2}")
```
lnardai: Is there some override for this variable?
Or where do we get the 8 value?

PR #7649, opened 2020-03-25
CB-6258 Use Launch Templates in CloudFormation templates

Bajzathd:


cloud-aws/src/main/resources/templates/aws-cf-stack-freeipa.ftl
```diff
@@ -120,7 +120,10 @@
         <#else>
         "VPCZoneIdentifier" : [{ "Ref" : "SubnetId" }],
         </#if>
-        "LaunchConfigurationName" : { "Ref" : "AmbariNodeLaunchConfig${group.groupName?replace('_', '')}" },
+        "LaunchTemplate" : {
+          "LaunchTemplateId": { "Ref" : "AmbariNodeLaunchTemplate${group.groupName?replace('_', '')}" },
```
doktoric: AmbariNodeLaunchTemplate -> CDPNodeLaunchTemplate

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -312,7 +312,10 @@
         "VPCZoneIdentifier" : [{ "Ref" : "SubnetId" }],
         </#if>
         </#if>
-        "LaunchConfigurationName" : { "Ref" : "AmbariNodeLaunchConfig${group.groupName?replace('_', '')}" },
+        "LaunchTemplate" : {
+          "LaunchTemplateId": { "Ref" : "AmbariNodeLaunchTemplate${group.groupName?replace('_', '')}" },
```
doktoric: CDPNodeLaunchTemplate

cloud-aws/src/main/resources/templates/aws-cf-stack-freeipa.ftl
bergerdenes: or rather ClusterManagerNodeLaunchTemplate

doktoric: Both are fine but cf template has a limit on the length this is why i think shorter is better

Bajzathd: Renamed it to ClusterManagerNodeLaunchTemplate, ran just fine

PR #7653, opened 2020-03-25
Merge 2.19

doktoric:


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -199,6 +202,51 @@ private void handleUser(UmsUsersState.Builder umsUsersStateBuilder, UsersState.B
 
     }
 
+    private boolean isEnvironmentUser(String enviromentCrn, GetRightsResponse rightsResponse) {
```
keyki: I think this is incorrect resolution

PR #7656, opened 2020-03-26
CB-6135 Unassign resource roles of a resource after deletion

horadla23:


horadla23:
Tested with mow-dev UMS, after credential deletion, deleted credential disappeared from user's page.
Screenshot 2020-03-26 at 11 47 26

PR #7667, opened 2020-03-26
CB-6043: FreeIPA management service should retry FreeIPA APIs

jamisonbennett:
When an instance of FreeIPA fails, the FreeIPA management service will
retry with a delay. When cluster proxy mode is used with FreeIPA HA,
the delay will allow cluster proxy to switch over to a working
instance of FreeIPA.

Retries are performed on the FreeIPA management service APIs, but not
flows nor user sync.

Retries are only performed with certain errors occur that have a
reasonable chance of succeeding on a retry.

Closes #CB-6043

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
@@ -1,35 +1,244 @@
 package com.sequenceiq.freeipa.client;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.springframework.http.HttpStatus;
 
 import com.googlecode.jsonrpc4j.JsonRpcClientException;
 
 public class FreeIpaClientExceptionUtil {
 
-    public static final int NOT_MAPPED_ERROR_CODE = 4000;
+    // See https://github.com/freeipa/freeipa/blob/master/ipalib/errors.py
+    public enum ErrorCodes {
+        PUBLIC_ERROR(900),
+        VERSION_ERROR(901),
+        UNKNOWN_ERROR(902),
+        INTERNAL_ERROR(903),
+        SERVER_INTERNAL_ERROR(904),
+        COMMAND_ERROR(905),
+        SERVER_COMMAND_ERROR(906),
+        NETWORK_ERROR(907),
+        SERVER_NETWORK_ERROR(908),
+        JSON_ERROR(909),
+        XML_RPC_MARSHALL_ERROR(910),
+        REFERER_ERROR(911),
+        ENVIRONMENT_ERROR(912),
+        SYSTEM_ENCODING_ERROR(913),
+        AUTHENTICATION_ERROR(1000),
+        KERBEROS_ERROR(1100),
+        CCACHE_ERROR(1101),
+        SERVICE_ERROR(1102),
+        NO_CACHE_ERROR(1103),
+        TICKET_EXPIRED(1104),
+        BAD_CCACHE_PERMS(1105),
+        BAD_CCACHE_FORMAT(1106),
+        CANNOT_RESOLVE_KDC(1107),
+        SESSION_ERROR(1200),
+        INVALID_SESSION_PASSWORD(1201),
+        PASSWORD_EXPIRED(1202),
+        KRB_PRINCIPAL_EXPIRED(1203),
+        USER_LOCKED(1204),
+        AUTHORIZATION_ERROR(2000),
+        ACI_ERROR(2100),
+        INVOCATION_ERROR(3000),
+        ENCODING_ERROR(3001),
+        BINARY_ENCODING_ERROR(3002),
+        ZERO_ARGUMENT_ERROR(3003),
+        MAX_ARGUMENT_ERROR(3004),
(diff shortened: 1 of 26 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/RetryableFreeIpaClientException.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.OptionalInt;
+
+public class RetryableFreeIpaClientException extends FreeIpaClientException {
+
+    public static final int MAX_RETRIES = 5;
+
+    public static final int BACKOFF_DELAY = 5000;
+
+    public static final double BACKOFF_MULTIPLIER = 1.5;
```
holleyism: These should probably be configurable with @Value.

jamisonbennett: These are used for annotation configuration in @Retryable. @Retryable requires the fields to be static which can't be configured using @Value. The @Retryable also doesn't support configuration directly via @Value. See https://stackoverflow.com/questions/38088449/how-can-i-make-spring-retryable-configurable for why this can't be done.

jamisonbennett: Done

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
```diff
@@ -1,35 +1,243 @@
...
+
+    private static boolean isRetryable(FreeIpaClientException e) {
+        return isExceptionWithErrorCode(e, RETRYABLE_HTTP_CODES) ||
+                (e.getStatusCode().isPresent() &&
+                retryableHttpCodes.contains(e.getStatusCode().getAsInt()));
+    }
+
+    public static Throwable getAncestorCauseBeforeFreeIpaClientExceptions(FreeIpaClientException e) {
+        while (e.getCause() instanceof FreeIpaClientException) {
+            e = (FreeIpaClientException) e.getCause();
```
lacikaaa: assigning new value to the method parameter seems bad practice to me. could you introduce new variable here?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/RetryableFreeIpaClientException.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.OptionalInt;
+
+public class RetryableFreeIpaClientException extends FreeIpaClientException {
+
+    public static final String MAX_RETRIES_EXPRESSION = "#{${freeipa.client.retry.retries:5}}";
+
+    public static final String DELAY_EXPRESSION = "#{${freeipa.client.retry.delay:5000}}";
+
+    public static final String MULTIPLIER_EXPRESSION = "#{${freeipa.client.retry.multiplier:1.5}}";
```
lacikaaa: you should default values only in one place, I think the yaml is fine

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
```diff
@@ -25,21 +26,25 @@
...
+        try {
+            return roleRequest == null ||
+                    roleRequest.getPrivileges().stream().allMatch(privilegeName -> {
+                        try {
+                            ipaClient.showPrivilege(privilegeName);
+                            return true;
+                        } catch (FreeIpaClientException e) {
+                            if (!FreeIpaClientExceptionUtil.isNotFoundException(e)) {
+                                LOGGER.error("Privilege [{}] show error", privilegeName, e);
+                                throw new FreeIpaClientExceptionWrapper(e);
```
lacikaaa: why is it necessary to wrap it?

freeipa/src/main/java/com/sequenceiq/freeipa/retry/RetryService.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.springframework.retry.annotation.Backoff;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.cloudbreak.util.CheckedFunction;
+import com.sequenceiq.freeipa.client.RetryableFreeIpaClientException;
+
+@Service
+public class RetryService {
```
lacikaaa: I don't know if you are aware of, but we already have a retryservice in thhe common. You might rename this just to avoid confusion later

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
jamisonbennett: `FreeIpaClientException` is a checked exception, so I wrap it with an unchecked exception which is thrown from the lambda passed into `matchAll()`. I then catch the unchecked exception and rethrow the original exception outside of the lambda.

jamisonbennett:
I also adjusted the retry delays to reduce the time to a total of about 15 seconds, which may help some with clients timing out but hopefully still allow the time for cluster proxy to connect to a different FreeIPA instance.

PR #7668, opened 2020-03-26
CDPCP-1735. Error message when getting user keytab on older FreeIPA

aarman-cloudera:
The user keytabs we generate are only valid on FreeIPAs that have password
hash support. This change returns an appropriate error based on this.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
@@ -8,10 +8,12 @@ import java.util.Optional;
 
 import javax.inject.Inject;
 
+import com.sequenceiq.freeipa.client.FreeIpaCapabilities;
 import com.sequenceiq.freeipa.client.FreeIpaClient;
 import com.sequenceiq.freeipa.client.FreeIpaClientException;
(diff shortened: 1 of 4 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/UserKeytabService.java
```diff
@@ -57,11 +59,14 @@ private void validateSameAccount(String userAccountId, String environmentCrn) {
...
     }
 
-    private void validateWorkloadUserInEnvironment(String workloadUsername, String environmentCrn) {
+    private void validateFreeIPAState(String workloadUsername, String environmentCrn) {
         String accountId = Crn.safeFromString(environmentCrn).getAccountId();
         FreeIpaClient freeIpaClient;
         try {
             freeIpaClient = freeIpaClientFactory.getFreeIpaClientByAccountAndEnvironment(environmentCrn, accountId);
+            if (!FreeIpaCapabilities.hasSetPasswordHashSupport(freeIpaClient.getConfig())) {
+                throw new UnsupportedException("User keytab retrieval requires a newer environment and IPA version");
```
lacikaaa: could you change `IPA` to `FreeIPA`?

aarman-cloudera: done

PR #7675, opened 2020-03-27
CB-6318 - Image change image should log progress

pdarvasi:


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/image/update/StackImageUpdateService.java
```diff
@@ -213,8 +213,25 @@ public boolean isValidImage(Stack stack, String newImageId, String imageCatalogN
...
                 StatedImage newImage = getNewImage(newImageId, imageCatalogName, imageCatalogUrl, currentImage);
-                return isCloudPlatformMatches(stack, newImage) && isOsVersionsMatch(currentImage, newImage) && isStackMatchIfPrewarmed(stack, newImage)
-                        && checkPackageVersions(stack, newImage).getStatus() == EventStatus.OK;
+                boolean cloudPlatformMatches = isCloudPlatformMatches(stack, newImage);
+                boolean osVersionsMatch = isOsVersionsMatch(currentImage, newImage);
+                boolean stackMatchIfPrewarmed = isStackMatchIfPrewarmed(stack, newImage);
+                CheckResult status = checkPackageVersions(stack, newImage);
+                boolean aggregatedValidationResult = cloudPlatformMatches && osVersionsMatch && stackMatchIfPrewarmed
+                        && status.getStatus() == EventStatus.OK;
+                if (!aggregatedValidationResult) {
```
lacikaaa: why not log anyway?

pdarvasi: Because it takes up quite some log-space to log the successful entries, too, with no real benefit.

PR #7680, opened 2020-03-30
CFM-913 Fix CFM-1.0.1.0 cluster creation

dam4rus:
This fix is introduced to enable cluster creation with CFM-1.0.0.0 and CFM-1.0.1.0 which is required by the DiM sustaining team.


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/nifi/NifiVolumeConfigProvider.java
```diff
@@ -26,6 +26,17 @@
 
     @Override
     public List<ApiClusterTemplateConfig> getRoleConfigs(String roleType, HostgroupView hostGroupView, TemplatePreparationObject source) {
+
+        // https://jira.cloudera.com/browse/CFM-913
```
keyki: Can you please remove the comment?

PR #7688, opened 2020-03-30
CB-6035: FMS should relogin on failures

holleyism:
This PR is cherry pick of 2 commits that are already in master, and 1 PR that has been approved but not merged.

This change adds goes through all instances in the
stack when trying to get a FreeIPA client unless the
loging is incorrect.

Co-authored-by: Jamison Bennett <jamison.bennnett@gmail.com>
(cherry picked from commit fc6106fd3014918669aa5ab19ec4f56e0d527072)

CB-6251: Could not reboot stopped FreeIPA

This change adjusts the reboot to properly reboot only started instances and
to call start on stopped instances. Because of AWS lifecycle
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
other statuses cannot be started or rebooted.

(cherry picked from commit 50f8a95712e150c95c005a19f225e95f23e0b5cc)

CB-6225: health check not returning status of all hosts.

This patch changes the health check to return individual
health of all HA nodes. The cn of the server_conncheck API
must be the host that the client is connected to. So we
exposed the connected hostname to use as the first parameter
of the API call.

(cherry picked from commit c16131f88d4ed6c4962b158792f19c8cb199d57e)

Ran unit tests and manually verified with 3 node cluster.


keyki:
This change will go out with CB 2.20. We're working on making it happen to mow.

holleyism:
Reopening per meeting discussion.

PR #7693, opened 2020-03-31
CB-6324 azure vm status set to terminated instead of stopped

topolyai5:


keyki:
I cannot verify from IT's point of view. The code change looks good so I would say merge as it fixes a real bug.

PR #7697, opened 2020-03-31
CB-6366 FreeIPA stacks are not deleted, they remain in DELETE_COMPLETED state

lacikaaa:


freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
```diff
@@ -54,22 +54,22 @@
...
+                Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+                String terminatedName = stack.getName() + DELIMITER + currentTimeMillis;
                 stack.setName(terminatedName);
-                stack.setTerminated(clock.getCurrentTimeMillis());
+                stack.setTerminated(currentTimeMillis);
                 terminateInstanceGroups(stack);
                 terminateMetaDataInstances(stack);
                 cleanupVault(stack);
+                stackUpdater.updateStackStatus(stack, DetailedStackStatus.DELETE_COMPLETED, "Stack was terminated successfully.");
                 stackService.save(stack);
```
handavid: is this save necessary since the stackUpdater also saves the stack? If we need to save it again, shouldn't we use the Stack that is returned from updateStackStatus instead? By my understanding, this stack should be detached at this point.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
```diff
@@ -54,22 +54,22 @@
...
-                stackUpdater.updateStackStatus(stackId, DetailedStackStatus.DELETE_COMPLETED, "Stack was terminated successfully.");
                 return null;
             });
-        } catch (TransactionService.TransactionExecutionException ex) {
-            LOGGER.info("Failed to terminate cluster infrastructure. Stack id {}", stack.getId());
+        } catch (TransactionExecutionException ex) {
+            LOGGER.info("Failed to terminate cluster infrastructure.");
```
handavid: nit: do we refer to freeipa stacks as "cluster" elsewhere? I think it would be clearer to refer to this as "freeipa infrastructure"

stackId is still available as a method parameter if you want to continue logging it.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
lacikaaa: you are right, I was thinking about it, but as it went into rc-2.19 I wanted to minimize the change. My issue is if updateStackStatus change and don't save the stack only saves a new StackStatus, then we don't save here. But yeah, the returned value should be saved.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
lacikaaa: it's a copy-paste error. yeah I could have added it, but hopefully the logcontext already contains that information, so I didn't think much about deleting it

PR #7700, opened 2020-03-31
CB-6376 azure_instance_msi missing from backup configuration for FreeIPA

wonderslug:
This adds the azure_instance_msi to the pillar that was missing.

Closes #CB-6376

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaBackupConfigView.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaBackupConfigView.java
@@ -75,6 +75,7 @@ public class FreeIpaBackupConfigView {
         map.put("hourly_enabled", this.hourlyEnabled);
         map.put("initial_full_enabled", this.initialFullEnabled);
         map.put("platform", ObjectUtils.defaultIfNull(this.platform, EMPTY_CONFIG_DEFAULT));
+        map.put("azure_instance_msi", ObjectUtils.defaultIfNull(this.azureInstanceMsi, EMPTY_CONFIG_DEFAULT));
         return map;
     }
 


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/config/FreeIpaBackupConfigView.java
```diff
@@ -75,6 +75,7 @@ public String getAzureInstanceMsi() {
         map.put("hourly_enabled", this.hourlyEnabled);
         map.put("initial_full_enabled", this.initialFullEnabled);
         map.put("platform", ObjectUtils.defaultIfNull(this.platform, EMPTY_CONFIG_DEFAULT));
+        map.put("azure_instance_msi", ObjectUtils.defaultIfNull(this.azureInstanceMsi, EMPTY_CONFIG_DEFAULT));
```
gergopapi2: Yes, I tested it with multiple subscriptions and this line did indeed heal the broken freeipa provisioning (freeipa initial backup).

One note though, for freeipa_backup.sh: if azcopy gets an empty identity (i.e. "" in the script) then command
{code}
/usr/local/bin/azcopy login --identity --identity-resource-id [id]
{code}
still responses with auth ok (this is what I see in the logs and I also tested it), a strange behavior in my opinion from azcopy.
So it might make sense to explicitly log the msi in the script. I know it is there in the config file but the occasional reader might not discover it first (and it can be overwritten later).
Not very important so I leave it to you if you agree and if you want to change it.

gergopapi2: Yes, this fixes it (tested with cross-subscription setup).

One note though: in the freeipa_backup.sh command `/usr/local/bin/azcopy login --identity --identity-resource-id [azure-msi]` will report auth ok even if azure-msi is "". It might make sense to add a log line what the used msi was. I know it can be read in the config file but could help clarity.

keyki:
PR is failing, because of a change in CBD. I'll retrigger if it's fixed.

PR #7706, opened 2020-04-01
CB-6164 Remove entitlement check for CCM

lacikaaa:


lacikaaa:
please wait for UI change

doktoric:
@lacikaaa  I think we should check how the local dev will work. I guess in local you should use ccm = false right ?

keyki:
This will just enable the option. It does not set it true by default.

lacikaaa:
@doktoric the default will be `DIRECT`, that won't change

PR #7708, opened 2020-04-01
CB-6311 DistroX termination hangs if cluster is STOPPED

foldik:
Fix termination with stopped nodes.


doktoric:
do we need this change on 2.19 as well ?

foldik:
The issue is there as well.

PR #7710, opened 2020-04-01
CB-6384 7.1.1 templates are visible even the feature flag is disabled…

akanto:


core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
```diff
@@ -162,7 +161,7 @@ public Blueprint create(Workspace workspace, Blueprint blueprint, Collection<Map
     }
 
     public PlatformRecommendation getRecommendation(Long workspaceId, String blueprintName, String credentialName,
-        String region, String platformVariant, String availabilityZone, CdpResourceType cdpResourceType) {
+            String region, String platformVariant, String availabilityZone, CdpResourceType cdpResourceType) {
```
sodre90, line 164: a small thing, but there is a plus tab character

akanto: Hm that is strange, my IDEA says that this  the right formatting and it has changed it automatically

core/src/main/java/com/sequenceiq/cloudbreak/service/runtimes/SupportedRuntimes.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.common.base.Strings;
+import com.sequenceiq.cloudbreak.cloud.VersionComparator;
+
+@Component
+public class SupportedRuntimes {
+
+    // Defines what is the latest version what is supported, every former version shall just work
+    @Value("${cb.runtimes.latest}")
+    private String latestSupportedRuntime;
```
doktoric, line 14: I guess it should be just runtime without s

PR #7712, opened 2020-04-02
DISTX-397 DE HA template co-locate JN and ZK nodes to masters and manager

cegganesh84:
1. CB-5660 added the ability to have JN across node groups.
2. Tested the template and the fix for CB-5660 works for the DE HA template in a private stack.


cegganesh84:
Thanks @keyki

PR #7713, opened 2020-04-02
CB-6264 Add Spot Percentage to DistroX request

bergerdenes:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsInstanceView.java
```diff
@@ -74,15 +74,7 @@ public String getKmsKey() {
...
-        if (sv instanceof Double) {
-            return (Double) sv;
-        } else if (sv instanceof Integer) {
-            return ((Number) sv).doubleValue();
-        } else if (sv instanceof String) {
-            return Double.valueOf((String) sv);
-        }
-        return null;
+    public Integer getSpotPercentage() {
+        return instanceTemplate.getParameter("spotPercentage", Integer.class);
```
doktoric, line 78: isn't it just percentage?

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -299,7 +299,7 @@
           },
           "InstancesDistribution": {
             "OnDemandBaseCapacity": 0,
-            "OnDemandPercentageAboveBaseCapacity": 100,
+            "OnDemandPercentageAboveBaseCapacity": ${100 - (group.spotPercentage!0)},
```
doktoric, line 302: I think we should use here the OnDemandPercentage and calculate the value in the java class

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsInstanceView.java
bergerdenes: on api level it is, but as internal parameter accompanying the instanceTemplate sitting in a map it is more clear, I guess

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
bergerdenes: well, yes, let's do it that way

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsInstanceView.java
doktoric: ok

PR #7720, opened 2020-04-02
CB-5734 Additional Validations for DistroX Autoscale API

smaniraju:
CB-5734 Additional Validations
1. Load Based Scaling Enabled only for clusters with cluster-proxy tunnel.
2. Additional Validation for invalid crn or name.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
@@ -35,6 +35,7 @@ import com.sequenceiq.periscope.converter.PrometheusAlertResponseConverter;
 import com.sequenceiq.periscope.converter.TimeAlertRequestConverter;
 import com.sequenceiq.periscope.converter.TimeAlertResponseConverter;
 import com.sequenceiq.periscope.domain.BaseAlert;
+import com.sequenceiq.periscope.domain.Cluster;
 import com.sequenceiq.periscope.domain.LoadAlert;
 import com.sequenceiq.periscope.domain.MetricAlert;
 import com.sequenceiq.periscope.domain.PrometheusAlert;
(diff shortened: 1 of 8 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
```diff
@@ -229,14 +224,34 @@ private LoadAlertResponse createLoadAlertResponse(LoadAlert loadAlert) {
...
+        alertId.ifPresentOrElse(
+                updateAlert -> {
+                    validateAlertForUpdate(clusterId, updateAlert, AlertType.LOAD);
+                }, () -> {
+                    if (!cluster.getTunnel().useClusterProxy()) {
+                        throw new BadRequestException(String.format("Cluster '%s' is not configured with Cluster Proxy Tunnel, " +
+                                "Cluster Tunnel is %s. Load Based Scaling not supported", cluster.getStackCrn(), cluster.getTunnel()));
+                    }
+
+                    String hostGroup = json.getScalingPolicy().getHostGroup();
```
sidseth, line 248: Do we need to check for the existence of the hostGroup in the cluster?

autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
```diff
@@ -229,14 +224,34 @@ private LoadAlertResponse createLoadAlertResponse(LoadAlert loadAlert) {
...
 
-    private void validateLoadAlert(Long clusterId, Optional<Long> alertId, LoadAlertRequest json) throws ParseException {
-        validateAlertForUpdate(clusterId, alertId, AlertType.LOAD);
+    private void validateLoadAlert(Long clusterId, Optional<Long> alertId, LoadAlertRequest json) {
+        Cluster cluster = clusterService.findById(clusterId);
+        alertId.ifPresentOrElse(
+                updateAlert -> {
+                    validateAlertForUpdate(clusterId, updateAlert, AlertType.LOAD);
+                }, () -> {
+                    if (!cluster.getTunnel().useClusterProxy()) {
```
sidseth, line 243: This looks good based on the assumption that we're going with w.r.t all clusters using the ClusterProxy. Will have to include this in the documentation eventually.

autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
```diff
@@ -265,30 +280,27 @@ private PrometheusAlertResponse createPrometheusAlertResponse(PrometheusAlert al
...
-                    alert = null;
-
-            }
-            if (alert == null) {
-                throw new NotFoundException(String.format("Could not find %s alert with id: '%s', for cluster: '%s'", alertType, alertId, clusterCrn));
-            }
-        });
+    private void validateAlertForUpdate(Long clusterId, Long alertId, AlertType alertType) {
+        BaseAlert alert;
+        String clusterCrn = clusterService.findStackCrnById(clusterId);
```
sidseth, line 285: Nit: Move into the (alert == null) block.

smaniraju: ack

autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
smaniraju: we would also need the hostGroup services validation, which can be added based on our other discussion.

PR #7721, opened 2020-04-02
CB-6412 Tie balanced dns to app version

lacikaaa:


jamisonbennett:
I think another check is needed in `StackToDescribeFreeIpaResponseConverter::convert()` for the load balanced DNS names.

PR #7724, opened 2020-04-02
CB-6217 Intorduce test case timeout for all the E2E test cases

aszegedi:
@lnardai please review

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
@@ -63,11 +63,14 @@ public class IntegrationTestApp implements CommandLineRunner {
     @Value("${integrationtest.outputdir:.}")
     private String outputDirectory;
 
-    @Value("${integrationtest.threadCount:2}")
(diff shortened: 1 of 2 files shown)


integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
```diff
@@ -168,6 +171,7 @@ private XmlSuite loadSuite(String suitePath, InputStreamSource resource) throws
             XmlSuite xmlSuite = parser.parse(suitePath, inputStream, true);
             xmlSuite.setParallel(XmlSuite.ParallelMode.getValidParallel(parallel));
             xmlSuite.setThreadCount(threadCount);
+            xmlSuite.setTimeOut(timeOut);
```
lnardai, line 174: Is this a suite timeout or individual testcase timeout?

aszegedi: Hi, Based on my knowledge the timeout for tests. You can define this at suite or test level. However this always means test timeout. If you define a different timeout for a specified test, you can override the test suite setting also.

PR #7726, opened 2020-04-02
CB-6374 do not remove blueprints which has templates

doktoric:


akanto:
I am not sure that this solves the problem, I think it needs to put back to DEFAULT those that are DEFAULT_DELETED

PR #7728, opened 2020-04-03
CB-874 Make FreeIPA and RDS provisioning parallel

nadamdb:


datalake/src/main/java/com/sequenceiq/datalake/flow/create/event/EnvWaitRequest.java
```diff
@@ -2,19 +2,27 @@
 
 import com.sequenceiq.datalake.flow.SdxContext;
 import com.sequenceiq.datalake.flow.SdxEvent;
+import com.sequenceiq.redbeams.api.endpoint.v4.databaseserver.responses.DatabaseServerStatusV4Response;
 
 public class EnvWaitRequest extends SdxEvent {
 
-    public EnvWaitRequest(Long sdxId, String userId) {
+    private DatabaseServerStatusV4Response databaseServerResponse;
```
lacikaaa, line 9: passing this response along is not necessary. If I'm not mistaken it's used in `ProvisionerService#startStackProvisioning`, but that info is already in `SdxCluster`, so I would drop it entirely if possible

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateFlowConfig.java
```diff
@@ -24,24 +24,35 @@
...
+            .event(SdxCreateEvent.RDS_WAIT_EVENT).defaultFailureEvent()
+
             .from(SDX_CREATION_WAIT_RDS_STATE)
-            .to(SDX_CREATION_START_STATE)
+            .to(SDX_CREATION_WAIT_ENV_STATE)
             .event(SdxCreateEvent.RDS_WAIT_SUCCESS_EVENT).defaultFailureEvent()
+
+            .from(SDX_CREATION_WAIT_ENV_STATE)
+            .to(SDX_CREATION_START_STATE)
+            .event(SdxCreateEvent.ENV_WAIT_SUCCESS_EVENT).defaultFailureEvent()
```
lacikaaa, line 42: I think it's possible that we have the old version running, and it's in `SDX_CREATION_WAIT_ENV_STATE`. We shut it down and fire up this new version. When it's finished with the `SDX_CREATION_WAIT_ENV_STATE` the new version will move to `SDX_CREATION_START_STATE` and will skip `SDX_CREATION_WAIT_RDS_STATE`. Could you check?

PR #7729, opened 2020-04-03
CB-6404 Secure permissions on cert and private key for user facing TLS

lacikaaa:


keyki:
For the sake of PR history, change it to root rather than nginx.

PR #7732, opened 2020-04-03
CB-6394 EnvironmentService validation fails when it has validation error

afarsang:
EnvironmentService validation fails when it has validation error

Closes 6394


keyki:
Isn't it the whole point of validation? :)

PR #7740, opened 2020-04-03
CB-6450 Add CFM template and definitions for 7.2.0 runtime versions

jtstorck:
…versions

Closes #CB-6450


doktoric:
@jtstorck  you need to add the blueprints here as well: 
https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/application.yml#L243

PR #7756, opened 2020-04-07
CB-6494: Update FreeIPA backup scripts to provide status

holleyism:
This change enables the backup script to report a single
line status. This enables easier parsing of the last
backup status for the health checks.

Local deployment and validation.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_backup
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_backup
@@ -6,6 +6,13 @@ set -x
 
 CONFIG_FILE=/etc/freeipa_backup.conf
 
+if [ -z "$CRON_BACKUP" ]
+then
+    BACKUP_MODE="MANUAL"
+else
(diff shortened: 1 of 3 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_backup
```diff
@@ -83,10 +92,19 @@ doLog(){
     echo "`date "+%Y-%m-%dT%H:%M:%SZ"` $type_of_msg ""$msg" >> $LOGFILE
 }
 
+doStatus(){
+    type_of_msg=$(echo $*|cut -d" " -f1)
+    msg=$(echo "$*"|cut -d" " -f2-)
+    [[ $type_of_msg == INFO ]] && type_of_msg="INFO " # one space for aligning
```
lacikaaa, line 98: so don't we need the same for error?

holleyism: No.  This code is just adding an addition space next to INFO so the logging lines up visually.

PR #7761, opened 2020-04-08
CB-6380. Do not persist telemetry component for stack with template type

oleewere:
- do not add telemetry component for stacks with type 'TEMPLATE'
- remove already existing ones from DB

tested against dtahub deployments with default cluster definitions + custom ones

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
@@ -122,6 +122,7 @@ public class StackV4RequestToStackConverter extends AbstractConversionServiceAwa
             convertAsStack(source, stack);
             updateCloudPlatformAndRelatedFields(source, stack, environment);
             setNetworkIfApplicable(source, stack, environment);
+            stack.getComponents().add(getTelemetryComponent(stack, source));
         }
         Map<String, Object> asMap = providerParameterCalculator.get(source).asMap();
         if (asMap != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/resources/schema/app/20200408152731_CB-6380_Do_not_persist_telemetry_for_stack_template.sql
```diff
@@ -0,0 +1,7 @@
+-- // CB-6380 Do not persist telemetry component for stack with template type
+-- Migration SQL that makes the change goes here.
+
+DELETE FROM component WHERE componenttype = 'TELEMETRY' AND stack_id IN (SELECT id FROM stack WHERE type = 'TEMPLATE')
```
lnardai, line 4: Should this be done only for stacks that are already deleted?

oleewere: for deleted stacks, those are not template ones, and telemetry components are deleted correctly there. for template types it's not needed having connected telemetry, as if it becomes a real stack, telemetry component will be populated (based on environment + sdx response). - basically that's what i have tested, like if im not having this telemetry component, for a UI based install, the right telemetry values are populated (without having anything in the stack templates)

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -139,8 +139,9 @@ public Stack convert(StackV4Request source) {
         stack.setInstanceGroups(convertInstanceGroups(source, stack));
         measure(() -> updateCluster(source, stack, workspace),
                 LOGGER, "Converted cluster and updated the stack in {} ms for stack {}", source.getName());
-        stack.getComponents().add(getTelemetryComponent(stack, source));
-
+        if (!isTemplate(source)) {
```
topolyai5, line 142: can you move into the `else` branch? or inside to `convertAsStack(source, stack)`

PR #7765, opened 2020-04-08
CB-6514 AWS Datalake deployment on mow-dev is failing with HV000028 U…

horadla23:
@bergerdenes I saw you put the annotation on this field, but in case of sdx internal, name comes from pathparam, people usually don't put it into the input json but validation is on it within stackv4request, thus lot of people experienced issue with sdx internal. Since I do not want to make people fill this info twice, I removed the annotation for now. (cloudbreak will validate the stackrequest anyway)

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
@@ -1,12 +1,9 @@
 package com.sequenceiq.sdx.api.model;
 
-import javax.validation.Valid;
-
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.StackV4Request;
 
 public class SdxInternalClusterRequest extends SdxClusterRequest {
 
-    @Valid
     private StackV4Request stackV4Request;
 
     public StackV4Request getStackV4Request() {


horadla23:
cloudbreak will validate the stackrequest anyway

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
```diff
@@ -1,12 +1,9 @@
 package com.sequenceiq.sdx.api.model;
 
-import javax.validation.Valid;
-
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.StackV4Request;
 
 public class SdxInternalClusterRequest extends SdxClusterRequest {
 
-    @Valid
```
akanto, line 9: Is it just a quick fix? It is strange that we are removing the validation.

horadla23: i am not sure, in case of sdx internal, cloudbreak validates stackrequest anyway, but if we want to validate in earlier, we can do it, but we have to make everyone to put the name into the input json, and that will duplicate data on api, so maybe this worth a discussion

bergerdenes: I put up the validation a few days back, it is fairly safe to remove it. I did not calculate with the consequences that this object of the request is used in such a relaxed way.

PR #7767, opened 2020-04-09
CB-6337 Add local profile to datalake with NON_HA database.

foldik:


foldik:
@akanto ok, I simply removed the profile based configuration.

Issue #4675, opened 2019-03-01
Cloudbreak should run preAmbari/postAmbari/postClusterInstall recipes in newly CREATED nodes


thaibui:
This is a feature/brainstorming request.

We have a use case at Bazaarvoice where we constantly upscale/downscale CB's clusters constantly, multiple times a day, 20-40 nodes at a time, based on workload from different teams. This has exposed some of the bugs in CB such as those mentioned in https://github.com/hortonworks/cloudbreak/issues/4669.

Furthermore, we have realized that the way CB + Salt handles recipes upscale for new nodes is inefficient. For example, the current logic would request pre-ambari-start, post-ambari-start and post-cluster-install salt state to be executed in **all nodes**. However, this should only be executed in **newly CREATED nodes**.

We have a patch for this ready if you guys would like to accept this new idea. This would help upscaling a large cluster very stable. For instance, in our case, with clusters having over 100 nodes, upscaling an additional 20-40 nodes is consistently around 7-15 min. Previously, this was not the case since ever upscale would require the participation of all minons in all salt states.

lacikaaa:
Could you share your solution? It makes sense to limit execution to new nodes only.

Thanks

thaibui:
@lacikaaa Just saw your reply today. Sorry for the very late response.

The current approach I took is not pretty, there's not an easy way to find which instances were still provisioning by looking at the database so we took an approach of looking at instances that were created within the last few hours. 

See https://github.com/bazaarvoice/cloudbreak/commit/f85d9e2c19b335652d33c206aaf2a16d9ca02d99

PR #7778, opened 2020-04-12
DISTX-415 Reduce the amount of time it takes to add new nodes

cegganesh84:
1. In this commit, we optimize the salt execution times by removing the `high state apply` after each phase of recipe execution.
2. There is only one high state apply in all the nodes after a scale-up.
3. We could still optimize the individual scripts like metering which always removes and adds the same package in all the nodes of the cluster for every scale-up.
4. This is the first time modifying something in salt, so I will focus on further optimization after getting initial feedback.
5. Also removed the deprecated pre/post-execution phases. Made sure that there are no existing salt scripts for such phases.

./gradlew build && test scale-up in a private stack to see the time getting reduced from 5 mins to 4 mins.

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
@@ -855,34 +855,25 @@ public class SaltOrchestrator implements HostOrchestrator {
     @SuppressFBWarnings("REC_CATCH_EXCEPTION")
     private void executeRecipes(GatewayConfig gatewayConfig, Set<Node> allNodes, ExitCriteriaModel exitCriteriaModel, RecipeExecutionPhase phase, boolean forced)
             throws CloudbreakOrchestratorFailedException, CloudbreakOrchestratorTimeoutException {
-        boolean postRecipe = phase.isPostRecipe();
         int maxRetry = forced ? maxRetryRecipeForced : maxRetryRecipe;
         try (SaltConnector sc = createSaltConnector(gatewayConfig)) {
             // add 'recipe' grain to all nodes
             Set<String> targetHostnames = allNodes.stream().map(Node::getHostname).collect(Collectors.toSet());
             saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", phase.value()), exitCriteriaModel, maxRetry,
                     exitCriteria);
-
-            // Add Deprecated 'PRE/POST' recipe execution for backward compatibility (since version 2.2.0)
-            if (postRecipe) {
-                saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", RecipeExecutionPhase.POST.value()),
-                        exitCriteriaModel, maxRetry, exitCriteria);
-            } else {
-                saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", RecipeExecutionPhase.PRE.value()),
-                        exitCriteriaModel, maxRetry, exitCriteria);
-            }
-
-            Set<String> allHostnames = allNodes.stream().map(Node::getHostname).collect(Collectors.toSet());
-            saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(allHostnames, allNodes), exitCriteriaModel, maxRetry, exitCriteria);
-            runNewService(sc, new HighStateRunner(allHostnames, allNodes), exitCriteriaModel, maxRetryRecipe, true);
+            saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(targetHostnames, allNodes), exitCriteriaModel, maxRetry, exitCriteria);
+            StateAllRunner stateAllRunner = new StateAllRunner(targetHostnames, allNodes, "recipes." + phase.value());
+            OrchestratorBootstrap saltJobIdTracker = new SaltJobIdTracker(sc, stateAllRunner);
+            Callable<Boolean> saltJobRunBootstrapRunner = saltRunner.runner(saltJobIdTracker, exitCriteria, exitCriteriaModel, maxRetry, false);
+            saltJobRunBootstrapRunner.call();
         } catch (CloudbreakOrchestratorTimeoutException e) {
             LOGGER.info("Recipe execution timeout. {}", phase, e);
             throw e;
         } catch (CloudbreakOrchestratorFailedException e) {
-            LOGGER.info("Orchestration error occurred during executing highstate (for recipes).", e);
+            LOGGER.info("Orchestration error occurred during execution of recipes.", e);
             throw e;
         } catch (Exception e) {
-            LOGGER.info("Unknown error occurred during executing highstate (for recipes).", e);
+            LOGGER.info("Unknown error occurred during execution of recipes.", e);
             throw new CloudbreakOrchestratorFailedException(e);
         } finally {
             try (SaltConnector sc = new SaltConnector(gatewayConfig, restDebug)) {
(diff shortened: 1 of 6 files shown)


cegganesh84:
The integration tests failed because of the optimizations. I could fix them if this patch is generally agreed upon.

```
[36mtest_1                |[0m 23:21:32.919 [TestNG-test=newway mock testcases-2] ERROR c.s.i.cloudbreak.context.TestContext [RecipeClusterTest.testRecipePreTerminationRecipeHasGotHighStateOnCluster] - MockVerification
[36mtest_1                |[0m java.lang.RuntimeException: /saltapi/run with bodyBody: state.highstate Times: -1 request should have been invoked exactly 5 times, but it was invoked 3 times
```

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -841,50 +841,28 @@ private void runNewService(SaltConnector sc, BaseSaltJobRunner baseSaltJobRunner
...
         int maxRetry = forced ? maxRetryRecipeForced : maxRetryRecipe;
         try (SaltConnector sc = createSaltConnector(gatewayConfig)) {
             // add 'recipe' grain to all nodes
             Set<String> targetHostnames = allNodes.stream().map(Node::getHostname).collect(Collectors.toSet());
             saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", phase.value()), exitCriteriaModel, maxRetry,
                     exitCriteria);
-
-            // Add Deprecated 'PRE/POST' recipe execution for backward compatibility (since version 2.2.0)
-            if (postRecipe) {
-                saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", RecipeExecutionPhase.POST.value()),
```
keyki, line 854: Can you please remove these 2 types from RecipeExecutionPhase as well?

keyki:
So we add recipe grains and then we remove them. If there is no highstate in between what is going to execute the recipes?

lacikaaa:
I think it would make a real difference if we would target only new nodes during upscale, like it's mentioned here: https://github.com/hortonworks/cloudbreak/issues/4675

oleewere:
@cegganesh84 , for metering, it's not the same package, so if it removes something and add something, that's totally valid to keep. (i thought probably it's time to remove the old package cleanup, but if it's still executing the remove, we cannot do that...as those different rpms using the same folders)

cegganesh84:
> So we add recipe grains and then we remove them. If there is no highstate in between what is going to execute the recipes?

So @keyki I thought this line executes the recipes `            saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(targetHostnames, allNodes), exitCriteriaModel, maxRetry, exitCriteria);
`. I thought this applies only to the newly added grains as opposed to the full state. If that is not the case it is my misunderstanding and would like to know the method to apply only the existing state.

cegganesh84:
> I think it would make a real difference if we would target only new nodes during upscale, like it's mentioned here: #4675

That is a very good suggestion. It would certainly remove the overhead for the salt masters, and in few cases like external repo like pypi.

cegganesh84:
> @cegganesh84 , for metering, it's not the same package, so if it removes something and add something, that's totally valid to keep. (i thought probably it's time to remove the old package cleanup, but if it's still executing the remove, we cannot do that...as those different rpms using the same folders)

Ah, I see. I did not look at it more closely. Thanks! Some investigation is still needed as to why I see the `rpm -i` is still getting executed on new nodes.

keyki:
> > So we add recipe grains and then we remove them. If there is no highstate in between what is going to execute the recipes?
> 
> So @keyki I thought this line executes the recipes ` saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(targetHostnames, allNodes), exitCriteriaModel, maxRetry, exitCriteria);`. I thought this applies only to the newly added grains as opposed to the full state. If that is not the case it is my misunderstanding and would like to know the method to apply only the existing state.

So `SyncAllRunner` just makes sure that all states and pillars and caches are up to date, but it does not execute anything. To target only the new instances with highstate would be like salt 'host1,host2,host3' state.highstate. The target needs to change from '*' to specific minion IDs. However, I'm afraid we'll run into troubles as some things need to be done on the master instance as well when new nodes are added to the cluster. Like we generate the TLS agent tokens on the master instance for all instances and we copy them over from there.

lacikaaa:
> .... However, I'm afraid we'll run into troubles as some things need to be done on the master instance as well when new nodes are added to the cluster. ...

I meant it only for recipe execution

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
cegganesh84: Thanks, done.

cegganesh84:
> > > So we add recipe grains and then we remove them. If there is no highstate in between what is going to execute the recipes?
> > 
> > 
> > So @keyki I thought this line executes the recipes ` saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(targetHostnames, allNodes), exitCriteriaModel, maxRetry, exitCriteria);`. I thought this applies only to the newly added grains as opposed to the full state. If that is not the case it is my misunderstanding and would like to know the method to apply only the existing state.
> 
> So `SyncAllRunner` just makes sure that all states and pillars and caches are up to date, but it does not execute anything. To target only the new instances with highstate would be like salt 'host1,host2,host3' state.highstate. The target needs to change from '*' to specific minion IDs. However, I'm afraid we'll run into troubles as some things need to be done on the master instance as well when new nodes are added to the cluster. Like we generate the TLS agent tokens on the master instance for all instances and we copy them over from there.

Thank you, that was helpful. I have added a new command StateAllRunner to run the specific recipe execution phase. Also will take up targeting the specific scaled up host in another PR.

keyki:
LGTM. I'll merge once it's green. Re-triggered IT.

akanto:
For larger node count it turned out that the highstate is inefficient, there are multiple reasons for it:
- it seems that many things happening on master side and basically Salt master is consuming a lot of  CPU, even if the machine has 8 cores we are saturating it
- we have a lot of inefficient states, which take a lot of time even they do nothing
- it seems that minions are downloading files from master and it is also slow
- we have a few states that are continuously doing something like fluentd states (although they are not slow in general)

What we learned that the pillar size is what has caused a lot of performance issues. If the pillar size is large then even if you execute a single state the master is struggling.

We tried to do state caching, changing pillar structures from yaml to json etc. but basically they didn't help that much. I think we need to put focus again and analyze the states and check the caching again, maybe even updating Salt from 2017 version to a newer one to see whether there are some performance differences. But I think I can raise a separate jira for this.

Further optimization possibility: completely avoid recipe execution if the end-user has not attached any recipe.

Although separating recipes probably ok, I wouldn't like to go down further on this path, I mean separating things out instead of just executing a highstate, because that makes the orchestration really simple, and it is super easy to keep compatibility.

cegganesh84:
Thanks @akanto for looking into this.

@keyki thanks for the review. I see that I could scale up fine with this PR, but not scale down, so looking into the issue. This is the error I am getting,

```
ew node(s) could not be removed from the cluster. Reason com.sequenceiq.cloudbreak.orchestrator.exception.CloudbreakOrchestratorFailedException: Failed: Orchestrator component went failed in 3 min(s), message: There are missing nodes from job (jid: 20200417010357866451), target: [gk1custom-compute2.gk1priv.xcu2-8y8x.cloudera.site]
Node: gk1custom-compute2.gk1priv.xcu2-8y8x.cloudera.site Error(s): 
Name: ipa host-del gk1custom-compute2.gk1priv.xcu2-8y8x.cloudera.site --updatedns && ipa-client-install --uninstall -U
Comment: Command "ipa host-del gk1custom-compute2.gk1priv.xcu2-8y8x.cloudera.site --updatedns && ipa-client-install --uninstall -U" run
Stderr: ipa: ERROR: did not receive Kerberos credentials
```

cegganesh84:
@keyki I have fixed one issue with respect to how salt state command was executed.

Still, the node decommission does not work.

I see that at some places there is a kinit before ipa commands, and it is missing in other places.

Present

```
    - unless: test -f /var/log/dnsrecord-add-executed || { echo $PW | kinit {{salt['pillar.get']('sssd-ipa:principal')}} && ipa dnsrecord-find {{salt['pillar.get']('sssd-ipa:domain')}} --a-rec=$(hostname -i); }

```
Absent

```
    - name: runuser -l root -c 'ipa host-del {{ salt['grains.get']('fqdn') }} --updatedns && ipa-client-install --uninstall -U'
```

Could this be the reason for the failure with, ```Stderr: ipa: ERROR: did not receive Kerberos credentials
```

For the context, I scale up the cluster and immediately scale it down.

cegganesh84:
@keyki scale down worked after I added kinit in ipa_leave command. I will look at the integration test failures if any later today.

keyki:
Thank you.

cegganesh84:
@keyki I have fixed the failing tests. Please review!

PR #7779, opened 2020-04-13
CB-5736 load based monitors evaulators

smaniraju:
Introduce LoadBased Evaluators and Monitors  for Autoscaling.
Integrate with Yarn Scaling API to query yarn recommendations.
Default tunnel for cluster communication changed to ClusterProxy.

Closes #CB-5736


autoscale/src/main/resources/application.yml
```diff
@@ -44,6 +45,19 @@ periscope:
...
   cloudbreak.url: http://localhost:9091
+  enabledAutoscaleMonitors:
+    time-monitor:
+      enabled: true
+    load-monitor:
+      enabled: true
+    rejected-thread-monitor:
+      enabled: true
+    cluster-manager-host-health-monitor:
+      enabled: true
```
smaniraju, line 56: cluster-manager-host-health-monitor is required for periscope functionality, it removes orphaned clusters in periscope which are deleted in cloudbreak.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+            Map<String, String> hostFqdnsToInstanceId = stackResponseUtils
+                    .getCloudInstanceIdsForHostGroup(stackV4Response, loadAlert.getScalingPolicy().getHostGroup());
+            yarnResponse.getScaleUpCandidates().ifPresentOrElse(
+                    scaleUpCandidates -> handleScaleUp(hostGroupInstanceType, scaleUpCandidates, hostFqdnsToInstanceId.size()),
+                    () -> {
+                        yarnResponse.getScaleDownCandidates().ifPresent(
+                                scaleDownCandidates -> handleScaleDown(scaleDownCandidates, hostFqdnsToInstanceId));
+                    });
+
+            LOGGER.debug("Finished loadEvaluator for cluster {} in {} ms", cluster.getStackCrn(), System.currentTimeMillis() - start);
```
sidseth, line 125: From testing, this is fairly quick, correct?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+        } catch (Exception ex) {
+            LOGGER.info("Failed to process load alert for Cluster {}, exception {}", cluster.getStackCrn(), ex);
+            eventPublisher.publishEvent(new UpdateFailedEvent(clusterId));
```
sidseth, line 128: Can more context be sent along with this event. What caused the failrue? (Inability to talk to YARN, something else)

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+                newNMCandidates.getCandidates().stream()
+                        .filter(candidate -> candidate.getModelName().equalsIgnoreCase(hostGroupInstanceType))
+                        .findFirst().map(NewNodeManagerCandidates.Candidate::getCount)
+                        .orElseThrow(() ->
+                                new RuntimeException(String.format("Yarn Scaling API Response does not contain recommended host count " +
+                                                "for hostGroupInstanceType %s in Cluster %s, Yarn Response %s",
+                                        hostGroupInstanceType, cluster.getStackCrn(), newNMCandidates)));
+
+        int maxAllowedScaleUp = loadAlert.getLoadAlertConfiguration().getMaxResourceValue() - existingHostGroupSize;
+        maxAllowedScaleUp = maxAllowedScaleUp < 0 ? 0 : maxAllowedScaleUp;
```
sidseth, line 145: As a future enhancement, we likely need to introduce a parameter which indicates how many nodes we will add in 1 iteration.
e.g. if YARN says add 200 nodes (which is within max node count) - we should not attempt to add 200 nodes in 1 invocation. Instead add 50. Nexy upscale can add the next 50 and so on.

autoscale/src/main/java/com/sequenceiq/periscope/domain/BaseAlert.java
```diff
@@ -32,6 +35,9 @@
     @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
     private ScalingPolicy scalingPolicy;
 
+    @Transient
+    private List<String> decommissionNodeIds;
```
sidseth, line 39: This does not belong here, even as a transient entity.
Think it is better to change the event that is being published to contain any dynamic information of this kind.

So instead of sending the BaseAlert in the SaclingEvent - send a new entity, which could internally contain the BaseAlert if required, but is more focussed on what scaling action needs to be taken. (decomissionNodeIds, newNodeCount, etc)

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+
+    protected void handleScaleDown(List<DecommissionCandidate> decommissionCandidates, Map<String, String> hostGroupFqdnsToInstanceId) {
+        Set<String> hostGroupFqdns = hostGroupFqdnsToInstanceId.keySet();
+        int maxAllowedScaleDown = hostGroupFqdns.size() - loadAlert.getLoadAlertConfiguration().getMinResourceValue();
+        maxAllowedScaleDown = maxAllowedScaleDown < 0 ? 0 : maxAllowedScaleDown;
+
+        List<String> decommissionHostGroupNodeIds = decommissionCandidates.stream()
+                .sorted(Comparator.comparingInt(DecommissionCandidate::getAmCount))
+                .map(DecommissionCandidate::getNodeId)
+                .map(nodeFqdn -> nodeFqdn.split(":")[0])
```
sidseth, line 165: How efficient / inefficient is this approach? i.e. will it end up iterating over the same lists multiple times over (including string split, etc) for different nodes which need to be removed?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+                .limit(maxAllowedScaleDown)
+                .map(nodeFqdn -> hostGroupFqdnsToInstanceId.get(nodeFqdn))
+                .collect(Collectors.toList());
+
+        LOGGER.info("ScaleDown NodeCount {} for Cluster {}, HostGroup {], NodeIds {}",
+                decommissionHostGroupNodeIds.size(), cluster.getStackCrn(), loadAlert.getScalingPolicy().getHostGroup(),
+                decommissionHostGroupNodeIds);
+
+        if (!decommissionHostGroupNodeIds.isEmpty()) {
+            loadAlert.setDecommissionNodeIds(decommissionHostGroupNodeIds);
```
sidseth, line 176: See comment in BaseAlert.

autoscale/src/main/resources/definitions/aws-vm.json
```diff
@@ -0,0 +1,174 @@
+{
```
sidseth, line 1: Can this and azure-vm.json use the same values that are already available in Cloudbreak?
Maintaining 2 copies of the same / similar files will lead to confusion sometime in the future when instances are changed.

autoscale/src/main/resources/application.yml
sidseth: It seems to do this by talking to the cluster itself, instead of getting the state from Cloudbreak? Letting Cloudbreak talk to the cluster for health, and then getting this information from Cloudbreak I think will work better. Future enhancement: Not related to this change at all.

autoscale/src/main/java/com/sequenceiq/periscope/utils/ClusterUtils.java
```diff
@@ -2,12 +2,16 @@
 
 import java.text.DecimalFormat;
 
-public final class ClusterUtils {
+public class ClusterUtils {
```
sidseth, line 5: Why remove the final?

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/CloudInstanceTypeService.java
```diff
@@ -0,0 +1,84 @@
...
+
+            TypeReference<HashMap<String, Set<CloudInstanceType>>> typeRef = new TypeReference<>() { };
+            Set<CloudInstanceType> cloudInstanceTypes = JsonUtil.readValue(cloudInstanceTypeConfig, typeRef).get("items");
+
+            for (CloudInstanceType cloudInstanceType : cloudInstanceTypes) {
+                vmInstanceTypes.put(cloudInstanceType.getInstanceName(), cloudInstanceType);
+            }
+
+            Set<String> configuredCloudInstances = cloudInstanceTypes.stream().map(CloudInstanceType::getInstanceName).collect(Collectors.toSet());
+            Set<String> cbSupportedInstances = internalCrnClient.withInternalCrn()
```
sidseth, line 72: Can the CB returned instance types be used directly? Why define them again in Periscope.
Alternately, if a library exists which can do this conversion - Periscope can rely on the DB entry directly.

environment/src/main/resources/application.yml
```diff
@@ -90,7 +90,8 @@ environment:
   enabledParentPlatforms: AWS, MOCK
   enabledChildPlatforms: YARN, MOCK
   tunnel:
-    default: DIRECT
+    default: CLUSTER_PROXY
```
sidseth, line 93: I'm still confused about whether CLUSTER_PROXY is the default or not. Maybe this gets changed to be the default in the production deployment.

core/src/main/resources/application.yml
```diff
@@ -177,24 +177,24 @@ cb:
     host.name.prefix.length: 255
     distrox:
       enabled.instance.types: >
-        Standard_D8_v3,
```
sidseth, line 180: This looks like an accidental formatting change?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -56,21 +60,27 @@
...
         MDCBuilder.buildMdcContext(cluster);
         try {
             int scalingAdjustment = desiredNodeCount - totalNodes;
-            if (scalingAdjustment > 0) {
+            if (!decommissionNodeIds.isEmpty()) {
+                scaleDownByNodeIds(decommissionNodeIds);
+            } else if (scalingAdjustment > 0) {
                 scaleUp(scalingAdjustment, totalNodes);
-            } else {
+            } else if (scalingAdjustment < 0) {
```
sidseth, line 83: This will be triggered in case of TimeAlerts. Will eventually have to change that over to work with the YARN API as well - so that nodes are not removed randomly. Separate jira / patch.

sidseth: Also, I suspect this gets triggered if the cluster has somehow managed to go above the maximum size. e.g. the max node count is reconfigured.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+    protected void pollYarnMetricsAndScaleCluster() {
+        long start = System.currentTimeMillis();
+        try {
+            MDCBuilder.buildMdcContext(cluster);
+
+            StackV4Response stackV4Response = cloudbreakCommunicator.getByCrn(cluster.getStackCrn());
+
+            String hostGroupInstanceType =
+                    stackResponseUtils.getHostGroupInstanceType(stackV4Response, loadAlert.getScalingPolicy().getHostGroup());
+
```
sidseth, line 112: Check min/max node count constraints first? If this gets reconfigured - then how the YARN response is processed may need to be re-evaluated.
e.g. cluster at 20 nodes. Max reconfigured to 10. YARN replies with 'need more resources'.

Looks like scaleUp handles this - except for bringing the node count down to the maxValue.

Could be a follow up jira as well, since you have already tested this.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
```diff
@@ -45,60 +47,72 @@
...
     public void onApplicationEvent(ScalingEvent event) {
         BaseAlert alert = event.getAlert();
         Cluster cluster = clusterService.findById(alert.getCluster().getId());
         MDCBuilder.buildMdcContext(cluster);
-        scale(cluster, alert.getScalingPolicy());
+        handleClusterScaling(cluster, alert);
     }
 
-    private void scale(Cluster cluster, ScalingPolicy policy) {
-        long remainingTime = getRemainingCooldownTime(cluster);
```
sidseth, line 57: Removing this check - does it affect the evaluators other than the YarnLoadEvaluator? YarnLoadEvaluator performs this check itself. The others do not - so may end up triggering way more events?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
```diff
@@ -45,60 +47,72 @@
...
-        int maxSize = cluster.getMaxSize();
-        return desiredNodeCount < minSize ? minSize : desiredNodeCount > maxSize ? maxSize : desiredNodeCount;
+        int minSize = cluster.getHostGroupMinSize();
+        int maxSize = cluster.getHostGroupMaxSize();
+
+        return desiredHostGroupNodeCount < minSize ? minSize : desiredHostGroupNodeCount > maxSize ? maxSize : desiredHostGroupNodeCount;
     }
 
+    private int getHostGroupNodeCount(Cluster cluster, ScalingPolicy policy) {
+        StackV4Response stackV4Response = cloudbreakCommunicator.getByCrn(cluster.getStackCrn());
```
sidseth, line 115: Any chance this can be passed in to ScalingHandler? May not be possible for anything other than the YARNLoadEvaluator.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+                .map(nodeFqdn -> hostGroupFqdnsToInstanceId.get(nodeFqdn))
+                .collect(Collectors.toList());
+
+        LOGGER.info("ScaleDown NodeCount {} for Cluster {}, HostGroup {], NodeIds {}",
+                decommissionHostGroupNodeIds.size(), cluster.getStackCrn(), loadAlert.getScalingPolicy().getHostGroup(),
+                decommissionHostGroupNodeIds);
+
+        if (!decommissionHostGroupNodeIds.isEmpty()) {
+            loadAlert.setDecommissionNodeIds(decommissionHostGroupNodeIds);
+            eventPublisher.publishEvent(new ScalingEvent(loadAlert));
```
sidseth, line 177: However the event which gets published is changed - I think it'll be useful to include a timestamp in the event. That can be used to track metrics on how long from event generation to event processing - and generate alerts based on that. Will be useful for load testing as well.

autoscale-api/src/main/java/com/sequenceiq/periscope/doc/ApiDescription.java
```diff
@@ -121,6 +121,10 @@
         public static final String SCALING_CONFIGURATION = "Scaling configuration for the cluster";
     }
 
+    public static class DistroXClusterJsonsProperties {
+        public static final String DISTROX_SCALING_MODE = "Scaling Types supported for distrox clusters";
```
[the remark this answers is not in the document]
smaniraju: ack

autoscale/src/main/java/com/sequenceiq/periscope/aspects/RequestLogging.java
```diff
@@ -11,10 +11,16 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(RequestLogging.class);
 
-    public <T> T logging(Supplier<T> callback, String requestName) {
+    public <T> T logResponse(Supplier<T> callback, String requestName) {
         long start = System.currentTimeMillis();
-        T o = callback.get();
-        LOGGER.debug("Ambari '{}' finished in {} ms", requestName, System.currentTimeMillis() - start);
-        return o;
+        T response = callback.get();
+        LOGGER.debug("Request '{}' finished in {} ms, Response {}", requestName, System.currentTimeMillis() - start, response);
```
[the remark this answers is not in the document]
smaniraju: Updated to log only non sensitive information returned from toLoggableString. In interaction with other services this response metadata is required otherwise it is very difficult to identify production issues. Logs can be rotated and purged. I think kibana can handle log purge effectively. Also changed logging this to "info" from "debug" since this would be needed in production issues..

autoscale/src/main/java/com/sequenceiq/periscope/domain/Cluster.java
```diff
@@ -29,6 +29,12 @@
 @Entity
 public class Cluster implements Monitored, Clustered {
 
+    public static final int DEFAULT_HOSTGROUP_MIN_SIZE = 1;
+
+    public static final int DEFAULT_HOSTGROUP_MAX_SIZE = 50;
+
+    public static final int DEFAULT_HOSTGROUP_COOL_DOWN = 30;
```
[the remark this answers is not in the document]
smaniraju: I think 2 minutes is too less, each scale up, scale down on CB takes around 10 mins. I will change it to 15 mins.

autoscale/src/main/java/com/sequenceiq/periscope/domain/Cluster.java
```diff
@@ -29,6 +29,12 @@
 @Entity
 public class Cluster implements Monitored, Clustered {
 
+    public static final int DEFAULT_HOSTGROUP_MIN_SIZE = 1;
+
+    public static final int DEFAULT_HOSTGROUP_MAX_SIZE = 50;
```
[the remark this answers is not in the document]
smaniraju: yes removed these defaults, but added range in LoadAlertConfigurationRequest
minResourceValue : minimum 1
maxResourceValue : maximum 200
coolDownMinutes : minimum 5, maximum 200.

Each cluster upscale\downscale on CB takes around 8-10 mins, so I have kept coolDownMinutes minimum to 5 mins.

autoscale/src/main/java/com/sequenceiq/periscope/domain/Cluster.java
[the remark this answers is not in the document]
smaniraju: removed these defaults

autoscale/src/main/java/com/sequenceiq/periscope/domain/LoadAlertConfiguration.java
```diff
@@ -28,6 +31,11 @@ public Integer getCoolDownMinutes() {
         return coolDownMinutes;
     }
 
+    @JsonIgnore
+    public Long getCoolDownMillis() {
+        return coolDownMinutes  * TimeUtil.MIN_IN_MS;
```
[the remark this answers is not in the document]
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/model/yarn/NewNodeManagerCandidates.java
```diff
@@ -0,0 +1,67 @@
+package com.sequenceiq.periscope.model.yarn;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public class NewNodeManagerCandidates {
```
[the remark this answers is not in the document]
smaniraju: they didnt have a swagger spec can be taken as a seperate jira when they provide.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/LoadMonitor.java
```diff
@@ -0,0 +1,50 @@
...
+                ClusterState.RUNNING, true, getPeriscopeNodeConfig().getId());
+        return clusterIds.isEmpty() ? List.of() : getClusterService().findClustersByClusterIds(clusterIds);
+    }
+
+    protected Class<? extends EvaluatorExecutor> getLoadEvaluatorForCluster(Cluster cluster) {
+        String policyHostGroup = cluster.getLoadAlerts().stream().findFirst()
+                .get().getScalingPolicy().getHostGroup().toLowerCase();
+
+        switch (policyHostGroup) {
+            case "compute":
```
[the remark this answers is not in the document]
smaniraju: yes we can add support based on services. there is a seperate jira for that.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/RejectedThreadMonitor.java
```diff
@@ -17,6 +18,7 @@
 import com.sequenceiq.periscope.monitor.evaluator.ClusterCreationEvaluator;
 
 @Component
+@ConditionalOnProperty(prefix = "periscope.enabledAutoscaleMonitors.rejected-thread-monitor", name = "enabled", havingValue = "true")
```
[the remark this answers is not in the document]
smaniraju: Dont think so, can remove the ConditionalOnProperty

autoscale/src/main/java/com/sequenceiq/periscope/monitor/client/YarnMetricsClient.java
```diff
@@ -0,0 +1,93 @@
...
+public class YarnMetricsClient {
+    private static final Logger LOGGER = LoggerFactory.getLogger(YarnMetricsClient.class);
+
+    private static final String YARN_API_URL = "%s/proxy/%s/resourcemanager/v1/cluster/scaling";
+
+    private static final String HEADER_ACTOR_CRN = "x-cdp-actor-crn";
+
+    private static final String PARAM_UPSCALE_FACTOR_NODE_RESOURCE_TYPE = "upscaling-factor-in-node-resource-types";
+
+    private static final String DEFAULT_UPSCALE_RESOURCE_TYPE = "memory-mb,vcore";
```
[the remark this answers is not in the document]
smaniraju: ack

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: the exception is logged already which includes information like YARN unavailable or YARN 500 error, UpdateFailedEvent is used to handle other flow like tracking successive failures.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: ok.

autoscale/src/main/resources/application.yml
smaniraju: Yes will create jira for that.

autoscale/src/main/java/com/sequenceiq/periscope/domain/BaseAlert.java
smaniraju: yes this belong to ScalingEvent, so moved it to ScalingEvent itself.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: new streams are produced at result generation steps like filter, map and  there will not be observable perf difference unless the list size is in 10000s. also streams processing has become standard list processing mechanism.

autoscale/src/main/resources/definitions/aws-vm.json
smaniraju: CB does not have  cpu, memory and similar metadata for azure, and there is no direct api to fetch it from CB. The lists are partially present for aws but it is deep inside aws connector. Only supported model names are available at top level api which is validated against periscope definition. Also going forward I think periscope should reduce it's dependency for metadata on CB to avoid unnecessary CB hits.

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/CloudInstanceTypeService.java
smaniraju: commented above. I dont think this is required in DB, it is stored in resourcedefinitions, so that it can be easily updated without generating sql.

environment/src/main/resources/application.yml
smaniraju: with this change default is changed to CLUSTER_PROXY.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
```diff
@@ -45,60 +47,72 @@
     @Inject
     private ClouderaManagerTotalHostsEvaluator totalHostsEvaluatorService;
```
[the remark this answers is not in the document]
smaniraju: removed

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/CloudbreakCommunicator.java
```diff
@@ -1,20 +1,36 @@
...
 public class CloudbreakCommunicator {
 
     @Inject
     private CloudbreakInternalCrnClient cloudbreakInternalCrnClient;
 
+    @Inject
+    private RequestLogging requestLogging;
+
     public StackV4Response getByCrn(String stackCrn) {
         return cloudbreakInternalCrnClient.withInternalCrn().autoscaleEndpoint().get(stackCrn);
```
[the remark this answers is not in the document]
smaniraju: can track it as a separate jira to remove the unwanted data in stack response.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -139,7 +152,31 @@ private void scaleDown(int scalingAdjustment, int totalNodes) {
...
+                    decommissionNodeIds, hostGroup, cluster.getStackCrn());
+            cloudbreakCommunicator.decommissionInstancesForCluster(cluster, decommissionNodeIds);
+            scalingStatus = ScalingStatus.SUCCESS;
+            statusReason = "Downscale successfully triggered";
+            metricService.incrementMetricCounter(MetricType.CLUSTER_DOWNSCALE_SUCCESSFUL);
+        } catch (Exception e) {
+            scalingStatus = ScalingStatus.FAILED;
+            metricService.incrementMetricCounter(MetricType.CLUSTER_DOWNSCALE_FAILED);
+            statusReason = "Couldn't trigger downscaling due to: " + e.getMessage();
+            LOGGER.info("Couldn't trigger downscaling to remove  nodeIds(s) {} from host group '{}', cluster '{}', error '{}' ",
```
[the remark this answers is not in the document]
smaniraju: updated log message, to history node count is already sent.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -92,7 +102,8 @@ private void scaleUp(int scalingAdjustment, int totalNodes) {
         String statusReason = null;
         ScalingStatus scalingStatus = null;
         try {
-            LOGGER.debug("Sending request to add {} instance(s) into host group '{}', triggered policy '{}'", scalingAdjustment, hostGroup, policy.getName());
+            LOGGER.debug("Sending request to add {} instance(s) into host group '{}', triggered policy '{}', cluster '{}'",
+                    scalingAdjustment, hostGroup, policy.getName(), cluster.getStackCrn());
             String stackCrn = internalCrnClient.withInternalCrn().autoscaleEndpoint().getStackForAmbari(ambariAddressJson).getCrn();
```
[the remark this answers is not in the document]
smaniraju: it is not required, not sure why it was added earlier, i have removed the endpoint call.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: This is basically for TimeAlerts.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: yes this is handled in scaleUp which is more real time.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: moved it to the scalingevent.

autoscale/src/main/java/com/sequenceiq/periscope/utils/ClusterUtils.java
smaniraju: added back

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: local testing it is pretty quick.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
smaniraju: This actually doesnt belong here, since this just handles the event generated. The check should be done before sending the scalingevent. Currently it applies only to LoadEvaluator, since TimeBased Alert has to be applied at the scheduled time.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
smaniraju: have added cacheable to stackv4response, so it is cached for 5 mins.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: eventTimestamp is already present in base class.

autoscale/src/main/java/com/sequenceiq/periscope/aspects/RequestLogging.java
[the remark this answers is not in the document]
sidseth: INFO level for all YARN responses is far too much logging. With multiple clusters, and pulling data every 15odd seconds (when the windowing approach is used) - this will generate a ton of logging.
Can we summarize the results from YARN into a much shorter log message, if this is absolutely required for debugging? Otherwise debug level is better.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
sidseth: eventPublished.publishEvent() ... is this what would eventually end up showing information on the 'Activities' page on the cluster UI?, or even to fetch scaling activity via the CLI?
If that is the case, it will be useful to have at least some information (Exception name, Exception message) to help with debugging without having to hunt through logs.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
sidseth: Also noticed that this is already covered somewhere in the Cloudbreak flow. Think the default is 100 at a time.

autoscale/src/main/resources/definitions/aws-vm.json
sidseth: https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html
or https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/AbstractAmazonEC2.html#describeInstanceTypes-com.amazonaws.services.ec2.model.DescribeInstanceTypesRequest-
could probably be used for AWS.

http://azure.github.io/ref-docs/java/com/microsoft/azure/management/compute/VirtualMachines.html#sizes--
for Azure.

These obviously won't be used at Runtime - but could potentially be part of the build to generate this list dynamically. (Future fix).

Curious: How did you end up generating this list for aws and azure?

Also, would this end up breaking YARN. That's something that will need to be tracked down sooner rather than later, given that a lot of the testing is going to happen on YCloud.

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/CloudInstanceTypeService.java
sidseth: Sorry. I meant the Cloudbreak DB. So specifically.
- An instance type is retrieved from the hostgroup info in the Cloudbreak DB.
- Periscope uses a library (which at the moment is backed by aws-, azure-*.json in the repo).
The intent being never to talk to Cloudbreak.

That is pretty much the case already. This communication to Cloudbreak to fetch instance types is only for error checking.
On the error checking - I think it makes sense to get Periscope to crash if information is not available for an instance type. That forces errors to be caught early.

If this will be maintained as these json files with look up - I think we should actually add an Integration test which ensures that what CB supports and Periscope knows about match, to fail even earlier. (Integration test is a follow up jira).

environment/src/main/resources/application.yml
sidseth: All of these settings could potentially be changed via the actual deployment scripts.

Also - I know this changes the default to CLUSTER_PROXY. That is a big enough change to be made via a separate jira, instead of as part of an otherwise big patch.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
sidseth: Still think a node count will be useful in the log message as well..

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
sidseth: Also for the change in 'max-nodes' scenario?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
sidseth: Bringing the nodeCount down to a dynamically configured 'maxValue' isn't handled from what I can tell.
YARN could end up recommending adding new nodes. maxNodes has been reduced by the end user. This will trigger scaleUp since YARN has a net positive node recommendation.
scaleUp doesn't trigger a reduction in size.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
sidseth: Metrics and Proemetheus, if they are enabled, will end up breaking. Mind putting a note into those evaluators that they need to check the cooldown before sending events forward?
Or maybe we do that in a common base itself.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
sidseth: Can the time be adjusted to a lower value? 5 minutes can lead to incorrect information on the number of nodes. Potentially spanning across multiple scale events.

autoscale-api/src/main/java/com/sequenceiq/periscope/api/model/LoadAlertConfigurationRequest.java
```diff
@@ -11,12 +14,19 @@
...
+    @Digits(fraction = 0, integer = 3)
     private @Valid Integer minResourceValue;
 
     @ApiModelProperty(LoadAlertJsonProperties.LOAD_ALERT_CONFIGURATION_MAX_RESOUCE_VALUE)
+    @Max(value = 200)
+    @Digits(fraction = 0, integer = 3)
     private @Valid Integer maxResourceValue;
 
     @ApiModelProperty(LoadAlertJsonProperties.LOAD_ALERT_CONFIGURATION_COOL_DOWN_MINS_VALUE)
+    @Min(value = 5)
```
sidseth, line 27: Think the 'min' should still be about 2 minutes here, to allow for enhancements in scale-up performance.
Also scale-down can finish within 2 minutes, or even faster.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -0,0 +1,187 @@
...
+        Integer scaleUpCount = IntStream.of(yarnRecommendedHostGroupCount, DEFAULT_MAX_SCALE_UP_STEP_SIZE, maxAllowedScaleUp)
+                .filter(i -> i >= 0)
+                .min()
+                .getAsInt();
+
+        LOGGER.info("ScaleUp NodeCount {} for Cluster {}, HostGroup {]", scaleUpCount,
+                cluster.getStackCrn(), loadAlert.getScalingPolicy().getHostGroup());
+
+        if (scaleUpCount > 0) {
+            loadAlert.getScalingPolicy().setScalingAdjustment(scaleUpCount);
```
sidseth, line 158: The same comment about modifying a 'Model' object applies here. Is there any chance this gets persisted to the DB accidentally?
The 'scaleUpCount' is essentially the same as the 'decommissionNodeIds' during scaleDown. Why not move this into ScalingRequest as well?

For all the other Evaluators, ScalingAdjustment is static, and is not modified by the Evaluators.

autoscale/src/main/java/com/sequenceiq/periscope/aspects/RequestLogging.java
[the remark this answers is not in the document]
smaniraju: yes i have over-riden tostring method of these dto to include absolute minimum fields.

autoscale/src/main/resources/definitions/aws-vm.json
smaniraju: I used a java client with my cloud credentials to generate the list.

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/CloudInstanceTypeService.java
smaniraju: agree. i didnt crash periscope on error checking since in that case, periscope will never start if cloudbreak is not running. agree enabling integration testing is the correct approach here. will create a jira for that.

environment/src/main/resources/application.yml
smaniraju: agree. i was planning to exclude this in the merge to main and create seperate commit for thta.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
smaniraju: added

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: I purposefully did not downscale when the cluster has somehow gone above maximum size which was caught in my unit testing, because in this case the user might have scale up cluster manually and if autoscale reduces the node size it will override the user which i felt is not expected behaviour of autoscaling.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: the time difference between retrieving latest load alert from db and triggering the scale up\down is just a few milli-seconds ( time taken by yarn api), I didn't make another db hit to get the latest max configured value since this interval is very small and saves one more db hit.  the 'maxValue' can be reconfigured few ms after triggering periscope scaleUpEvent in YarnLoadEvaluator or few ms after sending request to CB even in that case scaleUp would go through.

However in the next evaluation cycle, if yarn doesnt recommend any decommissionNodes and the policy has maxValue which is lesser than currentHostGroup size, if we trigger down scaling it will be forced downscaling, since yarn hasn't really found unused capacity. I dont think this would be expected behaviour because we are taking away required capacity. It will be downscaled automatically when yarn observes reduced load which i think is inline with the eventually consistent nature of autoscaling. I think this would also ensure less aggressive downscaling.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
smaniraju: added

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingHandler.java
smaniraju: agree, removed the cache and used single call for a single evaluation cycle.

autoscale-api/src/main/java/com/sequenceiq/periscope/api/model/LoadAlertConfigurationRequest.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: ScaleUp\ScaleDown history on UI is not from periscope, that is shown from CB itself.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/client/YarnMetricsClient.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public YarnScalingServiceV1Response getYarnMetricsForCluster(Cluster cluster,
+            String hostGroupInstanceType,
+            CloudPlatform cloudPlatform) throws Exception {
+
+        TlsConfiguration tlsConfig = tlsSecurityService.getTls(cluster.getId());
+        Optional<String> clusterProxyUrl = clusterProxyConfigurationService.getClusterProxyUrl();
+        if (!clusterProxyUrl.isPresent() || !cluster.getTunnel().useClusterProxy()) {
+            String msg = String.format("ClusterProxy Not Configured for Cluster {}, cannot query YARN Metrics.", cluster.getStackCrn());
+            throw new RuntimeException(msg);
```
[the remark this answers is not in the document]
smaniraju: It would trigger a UpdateFailedEvent and handled in UpdateFailedHandler.

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/CloudInstanceTypeService.java
sidseth: Makes sense.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
sidseth: That is an interesting point. Think we need a little more discussion on what the behaviour should be here.
As an end user, if I scale up a cluster manually beyond maxNodes
1) If there is no load on the cluster, it will be scaled down immediately - nullifying the scale up. (e.g. scale up manually to submit a large job)
2) If there a load on the cluster already, it will not be scaled down to maxNodes.
I think it is fair to keep the cluster between minNodes and maxNodes if AutoScaling is enabled. If a user wants to change something - disable the policy and scale to whatever size is needed.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
sidseth: Same as above comment.

PR #7780, opened 2020-04-13
CB-6558: Fix FreeIPA refer URL so the DNS load balanced names work

jamisonbennett:
The httpd rules were updated to allow the DNS load balanced domain
names to be used for connecting to FreeIPA directly without using the
nginx interface.

This was tested manually to ensure clients were not rejected with a
301 error.

Closes #CB-6558


freeipa/src/main/resources/freeipa-salt/salt/freeipa/templates/ipa-rewrite.conf.j2
```diff
@@ -0,0 +1,31 @@
+# VERSION 6 - DO NOT REMOVE THIS LINE
```
holleyism, line 1: Can you add a comment at the top saying "Create by Cloudbreak salt install" or something like that?

jamisonbennett:
This was validated with CML.

PR #7784, opened 2020-04-14
CB-5595 spark3 bps and defintions added

doktoric:


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/DatabaseAvailabilityTypeToExternalDatabaseRequestConverter.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.database.DatabaseAvailabilityType;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.database.DatabaseRequest;
+import com.sequenceiq.cloudbreak.converter.AbstractConversionServiceAwareConverter;
+
+@Component
```
topolyai5, line 11: We don't use the `AbstractConversionServiceAwareConverter` for super class

PR #7791, opened 2020-04-14
CB-6569. Encode (base64) regex input for anonymization rules (log collection)

oleewere:
- encode default anonymization rules (no validation yet for inputs)
- decode anonymization rules during salt config generation
- updated tests

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigService.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigService.java
@@ -1,14 +1,18 @@
 package com.sequenceiq.cloudbreak.telemetry.fluent;
 
+import java.util.ArrayList;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/telemetry/fluent/FluentConfigService.java
```diff
@@ -68,10 +72,28 @@ public FluentConfigView createFluentConfigs(TelemetryClusterDetails clusterDetai
...
+                .withAnonymizationRules(decodeRules(anonymizationRules))
                 .build();
     }
 
+    @VisibleForTesting
+    List<AnonymizationRule> decodeRules(List<AnonymizationRule> rules) {
+        List<AnonymizationRule> finalRules = null;
+        if (CollectionUtils.isNotEmpty(rules)) {
+            finalRules = new ArrayList<>();
+            for (AnonymizationRule rule : rules) {
```
horadla23, line 84: please use `stream` with `filter` instead of `for` and `if`s

PR #7795, opened 2020-04-15
CB-6497: Update Nginx config and other setup for freeipa healthagent

holleyism:
This patch adds a new Nginx config, and several salt setup scripts
to enable the new FreeIpa health agent to run.  The agent will
be installed as an RPM and will be a separate code review. CB-6399

Manually tested on 1,2, and 3 node FreeIPA clusters.


freeipa/src/main/resources/freeipa-salt/salt/freeipa/init.sls
```diff
@@ -18,6 +18,16 @@ net.ipv6.conf.lo.disable_ipv6:
       - {{ host['fqdn'] }}
 {% endfor %}
 
+{% if salt['pillar.get']('freeipa:hosts') > 1 %}
+/cdp/ipahealthagent/freeipa_cluster_node:
```
lacikaaa, line 22: I'm not sure it's a good idea to put a new folder under `/`. Couldn't we just put it under `/opt` somewhere?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
```diff
@@ -65,3 +65,20 @@ restart_httpd:
     - watch:
       - file: /etc/httpd/conf.d/ipa-rewrite.conf
 
+install-healthagent:
+  cmd.run:
+    - name: /opt/salt/scripts/freeipa_healthagent_setup.sh
+    - env:
+        - FPW: {{salt['pillar.get']('freeipa:password')}}
+    - failhard: True
+    - unless: /opt/salt/scripts/freeipa_check_replication.sh
```
lacikaaa, line 74: do we really need a separate file for this unless?

oleewere:
@holleyism is this anything to do with `ipa-healthcheck` tool or is it something else?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
holleyism: Yes.  Because of the '|', salt seems to have problems with it.  I tried several different permutations of escaping the character and using cmd.shell, but I kept receiving this `error:Failed: Orchestrator component went failed in 4 min(s), message: Salt execution went wrong: Rendering SLS 'base:freeipa.common-install' failed: mapping values are not allowed here; line 60 --- [...] cmd.run: - name: /opt/salt/scripts/freeipa_healthagent_setup.sh - env: - FPW: gnM3.B3J3G3l+-55j?Fe6V7CzDbdmkE_ - failhard: True - unless: ldapsearch -x -h localhost -b cn=config '(objectclass=nsds5replicationagreement)' | grep "^objectClass: nsds5replicationagreement"`

holleyism:
@oleewere - Yes this is a few setup steps for the new healthcheck agent.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/init.sls
holleyism: /cdp already existed from cluster proxy so seemed like a reasonable place to put it.
These already exist:
/cdp/bin/reverse-tunnel.sh
/cdp/bin/reverse-tunnel-values.sh
/cdp/bin/update-reverse-tunnel-values.sh

oleewere:
@holleyism - ok .nice, i got some monitoring tasks (for CM we are already gathering some metrics), where i need to store metrics in json files - and send them to control plane, probably I can use that tool for gathering some data about freeipa

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_healthagent_setup.sh
```diff
@@ -0,0 +1,20 @@
...
+dn: cn="$HOSTCN",cn=mapping tree,cn=config
+changetype: modify
+add: aci
+aci: (targetattr=*)(targetfilter="(|(objectclass=nsds5replicationagreement)(objectclass=nsDSWindowsReplicationAgreement))")(version 3.0; aci "permission:Read Replication Agreements"; allow (read, search, compare) groupdn = "ldap:///anyone";)
+EOF
+
+#
+# Add the cert extraction to run when the certificate is updated
+#
+ipa-getcert start-tracking -n Server-Cert -d /etc/httpd/alias -C "/usr/libexec/ipa/certmonger/restart_httpd;systemctl restart freeipa_healthagent"
```
lacikaaa, line 20: does it cause any issue on a system which don't have the agent installed?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_healthagent_setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+
+#
+# Add anonymous access only for replication agreements
+#
+export HOSTCN=`hostname -f| sed -E 's/^|\./,dc=/g' | sed 's/^,[^,]*,//g'`
```
lacikaaa, line 8: using `hostname -d` would simplify the sed a bit I think

holleyism: changed.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_healthagent_setup.sh
holleyism: removed.  will make it part of the RPM

PR #7800, opened 2020-04-15
CB-6567. Log collection: test anonymization rule regex

oleewere:
- add jregex dependeny (to handle perl-like regex)
- implement test anonymization rule endpoint
- rename TestAnonymizationRulesRequest and TestAnonymizationRulesResponse to TestAnonymizationRuleRequest and TestAnonymizationRuleResponse
 
Endpoint: 200 success + match, no match -404 (can be different - i do not know), invalid regex: 400

It only tests one rule (not all of them with a new one - reason is, it can match on different ones - so from UI perspective, it should be like a new window if someone want to test a specific rule)


attilapalfi92:
LGTM

PR #7808, opened 2020-04-15
CB-6338 Refactor wait for instance states at WaitUtil

aszegedi:
@afarsang, @lnardai and @topolyai5 please review

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImagesTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImagesTests.java
@@ -95,9 +95,7 @@ public class DistroXImagesTests extends AbstractE2ETest {
                 .given(distrox, DistroXTestDto.class).withImageSettings(imageSettings)
                 .when(distroXTestClient.create(), key(distrox))
                 .await(STACK_AVAILABLE)
-                .then((tc, testDto, client) -> {
-                    return waitUtil.waitForDistroxInstancesStatus(testDto, client, instancesHealthy);
-                })
+                .then((tc, testDto, client) -> waitUtil.waitForDistroxInstanceStatus(testDto, tc, instancesHealthy))
                 .then((tc, dto, client) -> {
                     Log.log(LOGGER, format(" Image Catalog Name: %s ", dto.getResponse().getImage().getCatalogName()));
                     Log.log(LOGGER, format(" Image Catalog URL: %s ", dto.getResponse().getImage().getCatalogUrl()));
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
@@ -1,8 +1,5 @@
 package com.sequenceiq.it.cloudbreak.util.wait;
 
-
-import static com.sequenceiq.sdx.api.model.SdxClusterStatusResponse.DELETED;
-import static com.sequenceiq.sdx.api.model.SdxClusterStatusResponse.REQUESTED;
 import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.containsIgnoreCase;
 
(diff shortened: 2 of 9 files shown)


lnardai:
I've changed the target branch to 2.20

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImagesTests.java
```diff
@@ -96,7 +96,8 @@ public void testDistroXWithBaseImageCanBeCreatedSuccessfully(TestContext testCon
                 .when(distroXTestClient.create(), key(distrox))
                 .await(STACK_AVAILABLE)
                 .then((tc, testDto, client) -> {
-                    return waitUtil.waitForDistroxInstancesStatus(testDto, client, instancesHealthy);
+                    waitUtil.waitForDistroxInstanceStatus(testDto.getResponse().getName(), tc, instancesHealthy);
```
lnardai, line 99: I don't really like the Idea that the return type of this method was changed to void.
Breaks the flow in the testcase.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
```diff
@@ -321,27 +336,36 @@ private boolean checkFailedStatuses(EnvironmentStatus currentStatus) {
...
         for (int retryBecauseOfWrongStatusHandlingInCB = 0; retryBecauseOfWrongStatusHandlingInCB < 3; retryBecauseOfWrongStatusHandlingInCB++) {
             waitResult = waitForStatuses(freeIPAClient, name, desiredStatus);
         }
         if (waitResult == WaitResult.FAILED) {
-            StringBuilder builder = new StringBuilder("The stack has failed: ").append(System.lineSeparator());
-            DescribeFreeIpaResponse freeIpaResponse = getFreeIPAResponse(freeIPAClient, name);
-            if (freeIpaResponse != null && freeIpaResponse.getStatus() != null) {
-                builder.append("statusReason: ").append(freeIpaResponse.getStatusReason());
-            }
+            StringBuilder builder = new StringBuilder("FreeIPA has failed: ").append(System.lineSeparator());
```
lnardai, line 346: This method got even more complicated. 
The status and reason reason field always set together, please refactor that to a separate method.
Also a lot of "freeIpaResponse.getStatusReason() != null ?", This could be incorporated to that method as a default value or it could be set by default.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
```diff
@@ -379,136 +414,130 @@ private boolean checkFailedStatuses(EnvironmentStatus currentStatus) {
...
+                status = sdxResponse.getStatus().name();
+                reason = sdxResponse.getStatusReason() != null ? sdxResponse.getStatusReason()
+                        : "SDX Status Reason is not available";
+                errors = Map.of("status", status, "reason", reason);
             }
         }
         return errors;
     }
 
-    public SdxTestDto waitForSdxInstanceStatus(SdxTestDto sdxTestDto, SdxClient sdxClient, String hostGroup, InstanceStatus desiredState) {
```
lnardai, line 408: By eliminating these methods the original ones got even more complex.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXImagesTests.java
aszegedi: I could revert this change only for here (for waitForDistroxInstancesStatus method). In case of waitForSdxInstancesStatus this was part of the required change.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
aszegedi: Original methods were tidied up. Supporting methods were placed here as well.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/WaitUtil.java
aszegedi: This change was required.

PR #7817, opened 2020-04-16
CB-5966. Configure DataBus (metering/logging/telemetry publisher) to use proxy.

oleewere:
Add proxy configs:
- fluentd (dbus plugin)
- telemetry publisher (CM component)

will require some more testing (manual)


keyki:
Target version should 2.20

oleewere:
from squid access log with using fluentd databus plugin
```text
1587122455.161 1115091 10.112.20.236 TCP_TUNNEL/200 64141 CONNECT dbusapi.sigma-dev.cloudera.com:443 squid HIER_DIRECT/10.32.157.226
```
I think that one is validated
Next: validate with Telemetry Publisher

oleewere:
seems like Telemetry Publisher configured properly, and also it's started. I could not check every kind of hadoop tools around that, but if it will be any issues can be done later or TP can be disabled if it does not work with http only proxies

PR #7822, opened 2020-04-16
CB-6617 Upgrade does not work due to lazy init

akanto:


keyki:
Should go against 2.20

akanto:
Switched to 2.20, I have originally fixed it on 2.20, but accidentally opened against the master

PR #7830, opened 2020-04-17
CB-6660 Added NIFI_REGISTRY_REST as an exposed service

jtstorck:
Updated displayName for the NIFI and NIFI_REGISTRY exposed services for proper stylization
Updated knoxService value for the NIFI_REGISTRY exposed service to be consistent with other exposed services
Updated apiIncluded to be false for the NIFI_REGISTRY exposed service due to the addition of the NIFI_REGISTRY_REST (NiFi Registry API) exposed service

Closes #CB-6660


core-api/src/main/resources/definitions/exposed-services.json
```diff
@@ -296,15 +296,15 @@
     },
     {
       "name": "NIFI_REGISTRY",
-      "displayName": "Nifi Registry",
+      "displayName": "NiFi Registry",
       "serviceName": "NIFI_REGISTRY_SERVER",
-      "knoxService": "NIFI-REGISTRY",
```
risdenk, line 301: I don't think this change is right? It should be `NIFI-REGISTRY` and not with an underscore?

References:
* https://github.com/apache/knox/blob/master/gateway-service-definitions/src/main/resources/services/nifi-registry/0.5.0/service.xml#L17
* https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2#L624

core-api/src/main/resources/definitions/exposed-services.json
```diff
@@ -322,6 +322,20 @@
       "visibleForDatalake": true,
       "visibleForDatahub": true
     },
+    {
+      "name": "NIFI_REGISTRY_REST",
+      "displayName": "NiFi Registry Rest",
+      "serviceName": "NIFI_REGISTRY_SERVER",
+      "knoxService": "NIFI_REGISTRY_REST",
```
risdenk, line 329: This should be `NIFI-REGISTRY` and not `NIFI_REGISTRY_REST`. If you want this to change, most likely need to change here as well:

https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology_api.xml.j2#L202

doktoric:
I guess this should go to rc-2.20 right ?

jtstorck:
If possible, I think it would be best to put this in rc-2.20.

doktoric:
@jtstorck can you please send the pull request to 2.20 branch ?

PR #7833, opened 2020-04-19
CB-4925 Enable EBS encryption for FreeIPA and DL

lajosrodek:
As per one-pager [design doc](https://docs.google.com/document/d/1OK6COfHosg3EA6IQxuV4o1AczzBLAqk2SNgyafJ-x0w/edit) linked to [CB-4925](https://jira.cloudera.com/browse/CB-4925), this PR unconditionally enables the encryption of EBS volumes for FreeIPA and DL clusters (but note the exception below for the latter). Notes:

* This applies to root (always EBS) as well as EBS data volumes. (There is no way to define encryption settings separately in `cloud-api`; they apply to all the volumes of the whole host group when given.)
* The current logic always uses the region-default EBS encryption key. The handling of customer-provided KMS CMK is out of scope here and will be covered in a follow-up JIRA & PR.
* No change to service APIs; settings are currently hard-wired into the service logic. Any possible API enhancement (e.g. to introduce custom encryption key choices) will be, again, treated in a future JIRA & PR. In other words, users have, in general, no way to tweak or disable EBS encryption for FreeIPA and DL clusters.
  * The only exception is the DL logic; `StackRequestManifester` honors any existing `AwsEncryptionV4Parameters` in the incoming `StackV4Request` with `getType() != null`, leaving the input setting untouched. This is only possible when using the `SdxInternalEndpoint`; specifying `getType() == EncryptionType.NONE` will thus always result in unencrypted EBS. On the other hand, the `getType() == null` setting will be overwritten with `EncryptionType.DEFAULT`, which is the path normally encountered when using the public `SdxEndpoint`.
* The above logic is guarded by the new entitlement `CDP_FREEIPA_DL_EBS_ENCRYPTION`. This entitlement gets checked by both FreeIPA and DL Svc; the absence of the entitlement grant will result in the legacy behavior.
* When testing with local CB, `mock-caas` defaults to presenting `CDP_FREEIPA_DL_EBS_ENCRYPTION` always granted. The following VM argument shall be added to `MockCaasApplication` to disable this entitlement: `-Dauth.mock.freeipadlebsencryption.enable=false`.
* Some extra logging have been added to `cloud-aws` and `cloud-template` that help determine the exact time spent with various cloud infra steps due to EBS encryption.
* Unit tests have been added for all logic changes, except for logging related enhancements. These are all based on JUnit 5 and AssertJ, sometimes using `ParameterizedTest`s. The exisiting `InstanceGroupRequestToInstanceGroupConverterTest` and `MockUserManagementServiceTest` have been ported to the above libraries. (The latter also retains the `ExpectedException` `Rule` from JUnit 4 via the experimental `junit-jupiter-migrationsupport` package.)

Follow-up tasks:
* Add customization options in service APIs, including the handling of customer-managed KMS keys. \[No JIRA ticket yet.\]
* Hide cloud provider specific parts of the above FreeIPA and DL logic behind interfaces & respective provider implementations. \[[CB-7455](https://jira.cloudera.com/browse/CB-7455)\]
* The usage of magic string literals (`"encrypted"`, `"type"`, `"key"` and `"spotPercentage"`) for instance template attributes is not best practice. To be discussed how to refactor this by exposing them as public constants (e.g. in `cloud-aws` / `AwsInstanceView`). \[[CB-7365](https://jira.cloudera.com/browse/CB-7365)\]

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/encryption/EncryptedImageCopyService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/encryption/EncryptedImageCopyService.java
@@ -78,8 +78,9 @@ public class EncryptedImageCopyService {
                 .collect(Collectors.toMap(Entry::getKey, e -> e.getValue().getImageId()));
         if (!imageIdByGroupName.isEmpty()) {
             Collection<String> imageIds = new HashSet<>(imageIdByGroupName.values());
-            LOGGER.debug("Start polling the availability of the created AMIs: '{}'", String.join(",", imageIds));
+            LOGGER.debug("Start polling the availability of the created encrypted AMIs: '{}'", String.join(",", imageIds));
             checkBooleanPollTask(awsPollTaskFactory.newAMICopyStatusCheckerTask(ac, imageIds, client));
+            LOGGER.info("All created encrypted AMIs are available: '{}'", String.join(",", imageIds));
         }
         return imageIdByGroupName;
     }
@@ -121,6 +122,7 @@ public class EncryptedImageCopyService {
             AuthenticatedContext ac, PersistenceNotifier resourceNotifier) {
 
         String regionName = ac.getCloudContext().getLocation().getRegion().value();
+        LOGGER.debug("Create an encrypted copy of the source AMI '{}' in region: '{}'", selectedAMIName, regionName);
         CopyImageResult copyImageResult = client.copyImage(createCopyImageRequest(selectedAMIName, regionName, awsInstanceView));
         String imageId = copyImageResult.getImageId();
         CloudResource cloudResource = new Builder()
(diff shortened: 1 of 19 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/encryption/EncryptedImageCopyService.java
```diff
@@ -78,8 +78,9 @@
                 .collect(Collectors.toMap(Entry::getKey, e -> e.getValue().getImageId()));
         if (!imageIdByGroupName.isEmpty()) {
             Collection<String> imageIds = new HashSet<>(imageIdByGroupName.values());
-            LOGGER.debug("Start polling the availability of the created AMIs: '{}'", String.join(",", imageIds));
+            LOGGER.debug("Start polling the availability of the created encrypted AMIs: '{}'", String.join(",", imageIds));
             checkBooleanPollTask(awsPollTaskFactory.newAMICopyStatusCheckerTask(ac, imageIds, client));
+            LOGGER.debug("All created encrypted AMIs are available: '{}'", String.join(",", imageIds));
```
biharitomi, line 83: I think it could be on info level, as it is an important info. This log should have been added by me years ago :)

lajosrodek: Done.

PR #7834, opened 2020-04-20
ENGESC-1231 new base image with openjd-11.0.7 for Cloudbreak and Autoscale

biharitomi:


pdarvasi:
@biharitomi  pls. check IT result so that I can merge, I see some token exceptions there.

biharitomi:
> @biharitomi pls. check IT result so that I can merge, I see some token exceptions there.

I am checking it right now, it looks like infra issue as this is not based on the new version of UAA :)

PR #7835, opened 2020-04-20
CB-6518 proper error message if credentials are not having proper policy for cloud storage backup

sodre90:


akanto:
Can we fix it on 2.20, since the new feature that caused this error message has been introduced in 2.20 and we can reduce the support time if a nice error message is displayed.

sodre90:
> Can we fix it on 2.20, since the new feature that caused this error message has been introduced in 2.20 and we can reduce the support time if a nice error message is displayed.

sure, rebased to 2.20

PR #7841, opened 2020-04-20
CB-6620 improve pkg.installed states

keyki:


orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls
```diff
@@ -8,6 +8,8 @@ install-cloudera-manager-agent:
     - pkgs:
       - cloudera-manager-daemons
       - cloudera-manager-agent
+    - unless:
+      - rpm -q cloudera-manager-daemons cloudera-manager-agent
```
akanto, line 12: I am not sure whether it matters but it breaks ubuntu and debian

keyki: The unless execution will fail on ubuntu (due to lack of rpm command), which is going to make the state execute.

PR #7843, opened 2020-04-20
CB-6697 Don't ignore volume related error in AWS upscale

daszabo:


cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/compute/CloudFailureHandler.java
```diff
@@ -59,7 +59,7 @@ private void doRollback(AuthenticatedContext auth, List<CloudResourceStatus> fai
         }
         switch (stx.getAdjustmentType()) {
             case EXACT:
-                if (stx.getThreshold() > fullNodeCount - failures.size()) {
+                if (stx.getThreshold() < fullNodeCount - failures.size()) {
```
keyki, line 62: Is this change correct? Let's say I set the threshold to 5 instances and the fullNodeCount is 10 nodes and 3 fails. Then 5 < 7 and it will throw an error.

PR #7844, opened 2020-04-21
CB-6705: Register multiple cluster proxy service for FreeIPA HA

jamisonbennett:
Register one cluster proxy service for each FreeIPA instance and an
additional one for the primary gateway for use during provisioning.
Prefer using the primary gateway freeipa instance but if it is
offline, then pick an instance of freeipa that is working.
Use the instance specific cluster proxy connections for health status,
that way the health status doesn't require both the checked node and
primary gateway node to be online.

When reporting the health status for FreeIPA's UI status, report an
error if any instance is offline. That way there is an easy way for a
customer to know if manual repair operation is required. In the future
when the repair happens automatically, the status should only report a
failure when all nodes are offline.

In addition to the tests, this was manually tested by bringing
individual FreeIPA instances offline and validating the status is
reported correctly. FreeIPA management service operations were also
validated to work when a node was offline.

Closes #CB-6705


freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -55,7 +55,7 @@ public HealthDetailsFreeIpaResponse getHealthDetails(String environmentCrn, Stri
         HealthDetailsFreeIpaResponse response = new HealthDetailsFreeIpaResponse();
 
         for (InstanceMetaData instance: instances) {
-            if (instance.isAvailable()) {
+            if (instance.isAvailable() && !instance.isDeletedOnProvider()) {
```
lacikaaa, line 58: if it's available it can't be deleted on provider side, can it?

jamisonbennett: You are correct.

jamisonbennett:
The branch rc-2.20 is no longer accepting new features. I will move this to master.

PR #7846, opened 2020-04-21
[WIP] CB-6583 Allow ManagementConsoleUser's to use our default provided image catalog

foldik:
Add default resource checker for authorization. Add authorization to image catalog.

### DIFF
--- a/authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
+++ b/authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
@@ -1,14 +1,20 @@
 package com.sequenceiq.authorization.service;
 
+import static java.util.stream.Collectors.toMap;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Parameter;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
 import org.aspectj.lang.ProceedingJoinPoint;
@@ -18,11 +24,13 @@ import org.slf4j.LoggerFactory;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.stereotype.Component;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
-import com.sequenceiq.authorization.annotation.DisableCheckPermissions;
 import com.sequenceiq.authorization.annotation.AuthorizationResource;
+import com.sequenceiq.authorization.annotation.DisableCheckPermissions;
 import com.sequenceiq.authorization.resource.AuthorizationResourceAction;
 import com.sequenceiq.authorization.resource.AuthorizationResourceType;
+import com.sequenceiq.authorization.resource.RightUtils;
 
 @Component
 public class CommonPermissionCheckingUtils {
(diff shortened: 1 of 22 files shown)


authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
```diff
@@ -18,39 +22,95 @@
...
     @Inject
-    private UmsResourceAuthorizationService umsResourceAuthorizationService;
+    public CommonPermissionCheckingUtils(UmsAccountAuthorizationService umsAccountAuthorizationService,
+            UmsResourceAuthorizationService umsResourceAuthorizationService,
+            List<DefaultResourceChecker> defaultResourceCheckers) {
+        this.umsAccountAuthorizationService = umsAccountAuthorizationService;
+        this.umsResourceAuthorizationService = umsResourceAuthorizationService;
+        this.defaultResourceCheckers = Maps.immutableEnumMap(defaultResourceCheckers.stream()
+                .collect(toMap(DefaultResourceChecker::getResourceType, Function.identity())));
+    }
```
horadla23, line 53: Is there a reason to switch to constructor injection? I have no problem with it, just other classes in the module don't use it

authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
```diff
@@ -18,39 +22,95 @@
...
-        umsResourceAuthorizationService.checkRightOfUserOnResource(userCrn, resource, action, resourceCrn);
+        DefaultResourceChecker defaultResourceChecker = defaultResourceCheckers.get(resource);
+        if (defaultResourceChecker == null || !defaultResourceChecker.isDefault(resourceCrn)) {
+            umsResourceAuthorizationService.checkRightOfUserOnResource(userCrn, resource, action, resourceCrn);
+        } else {
+            if (!defaultResourceChecker.isAllowedAction(action)) {
+                String right = RightUtils.getRight(resource, action);
+                String msg = String.format("You have no right to perform %s on resource %s", right, resourceCrn);
+                LOGGER.error(msg);
+                throw new AccessDeniedException(msg);
```
horadla23, line 68: This block occurs twice, please move into a method

authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
```diff
@@ -18,39 +22,95 @@
...
             String userCrn, Collection<String> resourceCrns) {
-        umsResourceAuthorizationService.checkRightOfUserOnResources(userCrn, resource, action, resourceCrns);
+        DefaultResourceChecker defaultResourceChecker = defaultResourceCheckers.get(resource);
+        if (defaultResourceChecker == null) {
+            umsResourceAuthorizationService.checkRightOfUserOnResources(userCrn, resource, action, resourceCrns);
+        } else {
+            Collection<String> defaultResourceCrns = defaultResourceChecker.getDefaultResourceCrns(resourceCrns);
+            Collection<String> notDefaultResourceCrns = resourceCrns.stream()
+                    .filter(crn -> !defaultResourceCrns.contains(crn))
+                    .collect(Collectors.toList());
```
horadla23, line 82: This also appears twice, maybe it is wort to move out into a method to increase readability of the code

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
```diff
@@ -54,15 +59,20 @@
...
 
 @Component
-public class ImageCatalogService extends AbstractWorkspaceAwareResourceService<ImageCatalog> {
+public class ImageCatalogService extends AbstractWorkspaceAwareResourceService<ImageCatalog> implements ResourceBasedCrnProvider {
 
     public static final String UNDEFINED = "";
 
+    public static final String CLOUDBREAK_DEFAULT_CATALOG_CRN = "cloudbreak-default-catalog-crn";
+
+    public static final String CDP_DEFAULT_CATALOG_CRN = "cdp-default-catalog-crn";
```
horadla23, line 74: Are these the CRNs of default catalogs? Format doesn't like a CRN format

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
```diff
@@ -691,6 +692,54 @@ private String createCRN(String accountId) {
...
 
+    @Override
+    public String getResourceCrnByResourceName(String resourceName) {
+        if (CDP_DEFAULT_CATALOG_NAME.equals(resourceName)) {
+            return CDP_DEFAULT_CATALOG_CRN;
+        } else if (legacyCatalogEnabled && CLOUDBREAK_DEFAULT_CATALOG_NAME.equals(resourceName)) {
+            return CLOUDBREAK_DEFAULT_CATALOG_CRN;
+        } else {
+            return imageCatalogRepository.findResourceCrnByNameAndWorkspaceId(resourceName, workspaceService.getForCurrentUser().getId())
+                    .orElseThrow(() -> NotFoundException.notFoundException("Image catalog", resourceName));
```
horadla23, line 703: I think we should talk about these default catalogs, I am not sure I get the logic here

horadla23:
Also please update authorization README with framework changes

authorization-common/src/main/java/com/sequenceiq/authorization/service/CommonPermissionCheckingUtils.java
foldik: This way the field can be final and we don't need a silly `@PostConstruct`. Of course we can start a religious war like "follow the conventions", you could have done it from java config  or there is nothing wrong with non final fields and `@PostConstruct`. I think it's better to make it final and use an immutable EnumMap in the constructor.

horadla23: it is ok for me, but one day one team member will start that war, i'm almost 100% sure about that :) I'm just like if a module follows one direction regarding things like this injection stuff based on team agreement, but I do not want to start (or getting involved in) a long discussion about this. Anyway I suggest to involve at least one other team member to check this, because some people do not like this, because I know there were discussions regarding this topic.

foldik:
Tested with locally with mow-dev UMS. Worked fine.

PR #7862, opened 2020-04-22
CB-5584 Check list before hasRights UMS call

horadla23:


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -370,10 +371,13 @@ public boolean checkRight(String actorCrn, String userCrn, String right, Optiona
...
-            List<Boolean> retVal = client.hasRights(requestId.orElse(UUID.randomUUID().toString()), memberCrn, rightChecks);
-            LOGGER.info("member {} has rights {}", memberCrn, retVal);
-            return retVal;
+            if (!rightChecks.isEmpty()) {
+                AuthorizationClient client = new AuthorizationClient(channelWrapper.getChannel(), actorCrn);
+                List<Boolean> retVal = client.hasRights(requestId.orElse(UUID.randomUUID().toString()), memberCrn, rightChecks);
+                LOGGER.info("member {} has rights {}", memberCrn, retVal);
+                return retVal;
+            }
+            return Collections.emptyList();
```
keyki, line 380: An empty list is fine? I'm not sure what does the caller check. Does this mean it will allow or deny?

horadla23: from authorization point of view, if the input is empty then there is nothing to check, so we can return without exception, callers can handle emtpy list as far as i have seen

PR #7864, opened 2020-04-22
 CB-5952 Streaming config providers for 7.1.1 / 7.2.0

benyoka:
Streaming config providers are adjusted to changes in CDH 7.1.1 / 7.2.0. These changes are:

- Kafka: removed `producer.metrics.enable` --> `true`
- SMM: changed DB setup and removed obsolete properties
- SR: changed DB setup and modified jar storage setup:
  - For 7.1.0 and earlier versions jar storage will stay the same (local folder on the first volume)
  - For 7.1.1 and newer: jar storage will be on S3 / ABFS

Testing:
- Updated unit tests
- Tested manual cluster installation

Closes CB-5952


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/AbstractRdsRoleConfigProvider.java
```diff
@@ -26,4 +27,7 @@ protected RdsView getRdsView(TemplatePreparationObject source) {
         return new RdsView(getRdsConfig(source));
     }
 
+    public static String dataBaseTypeForCM(DatabaseVendor vendor) {
```
doktoric, line 30: this should be presented only in you configprovider

benyoka: Ok. I thought it was some generic CM mechanism, but it looks like it's specific to out stuff:
```
      "validValues": [
        "mysql",
        "oracle",
        "postgresql"
      ],
```

PR #7869, opened 2020-04-22
CB-6738 - fixing premature poller timeout

attilapalfi92:


attilapalfi92:
spotbugs main timed out, retrying

PR #7871, opened 2020-04-22
CB-5741 Support For Retrieving VM Size Information

smaniraju:
Changes to retrieve VM Size Information from running cluster.
Auto-Supports all Cloud Platforms including Yarn.
Auto-Supports all future Cloud instance types like m6.
Cached VM Details per host-group to avoid unnecessary lookups.
No metadata dependency on CB and avoids runtime errors due to missing model labels.
Also introduced forced downscaling to match loadalert policy maxResourceValue.

Closes #CB-5741


autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -168,14 +169,23 @@ protected void handleScaleDown(List<DecommissionCandidate> decommissionCandidate
...
 
         ScalingEvent scalingEvent = new ScalingEvent(loadAlert);
         if (!decommissionHostGroupNodeIds.isEmpty()) {
+            LOGGER.info("ScaleDown NodeCount '{}' for Cluster '{}', HostGroup '{}', NodeIds '{}'",
+                    decommissionHostGroupNodeIds.size(), cluster.getStackCrn(), policyHostGroup,
+                    decommissionHostGroupNodeIds);
+
             scalingEvent.setDecommissionNodeIds(decommissionHostGroupNodeIds);
             eventPublisher.publishEvent(scalingEvent);
+        } else if (existingHostGroupSize > loadAlertConfiguration.getMaxResourceValue()) {
```
sidseth, line 181: Similarly here. This check before the actual scale down. Actually the combined check likely needs to happen before processing results from YARN.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClouderaManagerCommunicator.java
```diff
@@ -0,0 +1,72 @@
...
+    public Optional<CloudInstanceType> getCloudVMDetailsForHostGroup(Cluster cluster, String hostGroup, Set<String> hostGroupFqdns) {
+        try {
+            LOGGER.debug("Retrieving CloudVMType for cluster '{}', hostGroup '{}'", cluster.getStackCrn(), hostGroupFqdns);
+            HttpClientConfig httpClientConfig = tlsHttpClientConfigurationService.buildTLSClientConfig(cluster.getStackCrn(),
+                    cluster.getClusterManager().getHost(), cluster.getTunnel());
+            ClusterManager cm = cluster.getClusterManager();
+            String user = secretService.get(cm.getUser());
+            String pass = secretService.get(cm.getPass());
+            ApiClient client = clouderaManagerApiClientProvider.getClient(Integer.valueOf(cm.getPort()), user, pass, httpClientConfig);
+            HostsResourceApi hostsResourceApi = clouderaManagerApiFactory.getHostsResourceApi(client);
```
sidseth, line 57: *Not really related to this patch*
Dug through the CM API after looking at this usage.
https://sseth-hf34092-master0.sseth-en.xcu2-8y8x.dev.cldr.work/api/v32/clusters/sseth-hf34092/services/yarn/roleConfigGroups/yarn-NODEMANAGER-WORKER/config - may provide the actual value allocated.
The problem here may be the roleConfigGroup name. Highly doubt that CB stores that in its database. Even if it does, or we could get access to the template to figure this out - determining the roleName may not be trivial.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/client/YarnMetricsClient.java
```diff
@@ -66,13 +73,17 @@ public YarnScalingServiceV1Response getYarnMetricsForCluster(Cluster cluster,
         String yarnApiUrl = String.format(YARN_API_URL, clusterProxyUrl.get(), cluster.getStackCrn());
         YarnScalingServiceV1Request yarnScalingServiceV1Request = new YarnScalingServiceV1Request();
 
-        CloudInstanceType cloudInstanceType = cloudInstanceTypeService.getCloudVMInstanceType(cloudPlatform, hostGroupInstanceType)
-                .orElseThrow(() -> new RuntimeException(String.format("CloudVmType not found for CloudPlatform %s, " +
-                        " InstanceType %s, Cluster %s ", cloudPlatform, hostGroupInstanceType, cluster.getStackCrn())));
+        CloudInstanceType cloudInstanceType = hostGroupFqdns.size() == 0 ? defaultCloudInstanceType : clouderaManagerCommunicator
```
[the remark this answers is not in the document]
smaniraju: The yarnScalingAPI uses the instance vm information we give as a shirt size to give recommended nodeCount and don't think it is used to configure 32GB internally that would be based on the actual node added to the cluster. 

Also once one node is added to hostGroup the scaling service autocorrects itself. So next scaleUpEvent corrects the node size and count if a non-default standard size is used in the template.

The 32 gb default is the standard size used in DE templates corresponding to m5.2xlarge and Standard_D8_v3. 

So these are the only options I see for this.
1. Use default DE Template Host Size and let the service auto correct itself as implemented above.
2. Above changes plus enforce minHostNodeCount to "1" in LoadAlertConfiguration whenever hostGroup needs to be auto-scaled. In autoscale run if nodeCount is 0, trigger scale to change it to "1".
3. Revert back to the earlier implementation where in all the cloud platform vms were defined.

I don't see any other possible implementation for this. Let me know what you think.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/client/YarnMetricsClient.java
```diff
@@ -66,13 +73,17 @@ public YarnScalingServiceV1Response getYarnMetricsForCluster(Cluster cluster,
...
+                .getCloudVMDetailsForHostGroup(cluster, hostGroup, hostGroupFqdns)
+                .orElseGet(() -> {
```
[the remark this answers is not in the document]
smaniraju: same as above.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -109,55 +115,50 @@ protected void execute() {
...
 
-    protected void handleScaleUp(String hostGroupInstanceType, NewNodeManagerCandidates newNMCandidates, Integer existingHostGroupSize) {
+    protected void handleScaleUp(NewNodeManagerCandidates newNMCandidates, Integer existingHostGroupSize) {
         Integer yarnRecommendedHostGroupCount =
                 newNMCandidates.getCandidates().stream()
-                        .filter(candidate -> candidate.getModelName().equalsIgnoreCase(hostGroupInstanceType))
+                        .filter(candidate -> candidate.getModelName().equalsIgnoreCase(policyHostGroup))
                         .findFirst()
                         .map(NewNodeManagerCandidates.Candidate::getCount)
                         .orElseThrow(() -> new RuntimeException(String.format(
```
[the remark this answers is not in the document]
smaniraju: ok removed exception.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -109,55 +115,50 @@ protected void execute() {
...
-                stackResponseUtils.getHostGroupInstanceType(stackV4Response, loadAlert.getScalingPolicy().getHostGroup());
+        Map<String, String> hostFqdnsToInstanceId = stackResponseUtils.getCloudInstanceIdsForHostGroup(stackV4Response, policyHostGroup);
 
         YarnScalingServiceV1Response yarnResponse = yarnMetricsClient
-                .getYarnMetricsForCluster(cluster, hostGroupInstanceType, stackV4Response.getCloudPlatform());
+                .getYarnMetricsForCluster(cluster, policyHostGroup, hostFqdnsToInstanceId.keySet());
 
-        Map<String, String> hostFqdnsToInstanceId = stackResponseUtils
-                .getCloudInstanceIdsForHostGroup(stackV4Response, loadAlert.getScalingPolicy().getHostGroup());
         yarnResponse.getScaleUpCandidates().ifPresentOrElse(
```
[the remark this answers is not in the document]
smaniraju: Logically Yarn should either give a ScaleUp Count or Decommission Node because both are mutually exclusive. So if scaleUp is asked, decommission nodes should be empty but since cluster size is already above max, scale up event "-2" forces it down. But one corner case is scaleUp count is given and decommission nodes contains unhealthy nodes, in this case decommissionUnhealthy nodes was given lower priority.

Now I have updated patch to make limit check more explicit for all the conditions. 

Also forced scaleDown is done only after yarnResponse so that we can collect any available yarn decommissionCandidates.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClouderaManagerCommunicator.java
smaniraju: I tried hitting the compute roleConfigGroup but that does not contain the memory or vcpu https://10.112.6.147/api/v32/clusters/spot-distrox-5/services/yarn/roleConfigGroups/yarn-NODEMANAGER-COMPUTE/config

Response "items":[{"name":"process_auto_restart","value":"true","sensitive":false},{"name":"yarn_nodemanager_local_dirs","value":"/hadoopfs/fs1/nodemanager","sensitive":false},{"name":"yarn_nodemanager_log_dirs","value":"/hadoopfs/fs1/nodemanager/log","sensitive":false}]}"

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: commented above.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClouderaManagerCommunicator.java
sidseth: Sounds good. Which API did you end up using to look up the nodemanager memory. IThink you had mentioned a 'hosts' api.

PR #7876, opened 2020-04-22
CB-6497: Update Nginx config and other setup for freeipa healthagent

holleyism:
This patch adds a new Nginx config, and several salt setup scripts
to enable the new FreeIpa health agent to run.  The agent will
be installed as an RPM and will be a separate code review. CB-6399

(cherry picked from commit afeb8f6cc2c695818a3335ae324bbb08688b8e9d)

Removed the separate script for unless and included the check in the freeipa_healthagent_setup.sh itself.


freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
```diff
@@ -65,3 +65,19 @@ restart_httpd:
     - watch:
       - file: /etc/httpd/conf.d/ipa-rewrite.conf
 
+install-healthagent:
```
lacikaaa, line 68: shouldn't this state have any onlyif/unless option, so we won't run it if not necessary?

holleyism: There was an unless option before, but I think there was a timing problem between the unless call and the actual script because it should not have run based on the unless, but things were failing because it was.  I moved the unless check into the script itself so the script may be called but the actual action isn't performed unless it's necessary.

lacikaaa: this will cause saltstack will always detect this as a change which is not ideal. we can still place a file somewhere from the script and check that from unless

holleyism: updated to a better check.

holleyism:
Tested this against 1 and 2 node clusters twice each.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
lacikaaa: I think you should place the unless script in the require part too

holleyism: added

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_healthagent_setup.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+#
+# Add anonymous access only for replication agreements
+#
+
+export HOSTCN=`hostname -d| sed -E 's/^|\./,dc=/g' | sed 's/^,//g'`
+
+ldapsearch -x -h localhost -D "cn=directory manager" -w $FPW -b "$HOSTCN" -s base "(objectclass=*)"
+status=$?
```
lacikaaa, line 10: I just noticed that this duplicates the content of `freeipa_check_replication.sh`. could you update this to have it only in one place, eg. like a method and use that in both place?

holleyism: changed.

PR #7877, opened 2020-04-22
CB-6705: Register multiple cluster proxy service for FreeIPA HA

jamisonbennett:
Register one cluster proxy service for each FreeIPA instance and an
additional one for the primary gateway for use during provisioning.
Prefer using the primary gateway freeipa instance but if it is
offline, then pick an instance of freeipa that is working.
Use the instance specific cluster proxy connections for health status,
that way the health status doesn't require both the checked node and
primary gateway node to be online.

When reporting the health status for FreeIPA's UI status, report an
error if any instance is offline. That way there is an easy way for a
customer to know if manual repair operation is required. In the future
when the repair happens automatically, the status should only report a
failure when all nodes are offline.

In addition to the tests, this was manually tested by bringing
individual FreeIPA instances offline and validating the status is
reported correctly. FreeIPA management service operations were also
validated to work when a node was offline.

Closes #CB-6705


freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -138,7 +138,9 @@ public FreeIpaClient build(boolean withPing) throws URISyntaxException, IOExcept
                     .build()) {
                 URI target = getIpaUrl(clientConfig.getApiAddress(), port, basePath, "/session/login_password").toURI();
                 LOGGER.debug("Ping at target: {}", target);
-                client.execute(new HttpHead(target));
+                HttpHead request = new HttpHead(target);
+                additionalHeaders.forEach(request::addHeader);
+                client.execute(request).close();
```
lacikaaa, line 143: you don't have to call the `close()` here, `try` will take care of it

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/provision/action/FreeIpaProvisionActions.java
```diff
@@ -90,12 +92,29 @@ protected Selectable createRequest(StackContext context) {
...
+            protected Selectable createRequest(StackContext context) {
+                return new ClusterProxyUpdateRegistrationRequest(context.getStack().getId());
+            }
+        };
+    }
+
     @Bean(name = "FREEIPA_POST_INSTALL_STATE")
     public Action<?, ?> postInstallFreeIpa() {
-        return new AbstractStackProvisionAction<>(InstallFreeIpaServicesSuccess.class) {
+        return new AbstractStackProvisionAction<>(ClusterProxyUpdateRegistrationSuccess.class) {
```
lacikaaa, line 114: we could use `StackEvent` here so both the old and new event would match

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -138,38 +139,52 @@ private boolean canTryAnotherInstance(boolean lastInstance, IOException e) {
...
+        return instanceMetaDataRepository.findNotTerminatedForStack(stack.getId()).stream()
+                .filter(InstanceMetaData::isAvailable)
+                .sorted((l, r) -> {
+                    if (l.getInstanceMetadataType() == InstanceMetadataType.GATEWAY_PRIMARY) {
+                        return -1;
+                    }
+                    if (r.getInstanceMetadataType() == InstanceMetadataType.GATEWAY_PRIMARY) {
+                        return 1;
+                    }
+                    return Comparator.comparing(InstanceMetaData::getDiscoveryFQDN).compare(l, r);
```
lacikaaa, line 152: nit: create a comparator class for this

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -61,6 +63,9 @@
     @Inject
     private TlsSecurityService tlsSecurityService;
 
+    @Inject
+    private InstanceMetaDataRepository instanceMetaDataRepository;
```
lacikaaa, line 67: don't include repository anywhere else than the related service

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java
```diff
@@ -70,15 +74,25 @@
     @Inject
     private SecurityConfigService securityConfigService;
 
+    @Inject
+    private InstanceMetaDataRepository instanceMetaDataRepository;
```
lacikaaa, line 78: don't use repository here

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java
```diff
@@ -88,17 +102,27 @@
...
-        HttpClientConfig httpClientConfig = tlsSecurityService.buildTLSClientConfigForPrimaryGateway(
-                stack, primaryGatewayConfig.getGatewayUrl());
+        List<GatewayConfig> gatewayConfigs = gatewayConfigService.getNotTerminatedGatewayConfigs(stack);
         ClientCertificate clientCertificate = clientCertificates(stack);
 
-        ClusterServiceConfig serviceConfig = createServiceConfig(stack, httpClientConfig, clientCertificate);
+        boolean usePrivateIpToTls = stack.getSecurityConfig().isUsePrivateIpToTls();
+        List<GatewayConfig> tunnelGatewayConfigs;
+        List<ClusterServiceConfig> serviceConfigs = new LinkedList<>();
+        serviceConfigs.add(createServiceConfig(stack, FREEIPA_SERVICE_NAME, primaryGatewayConfig, clientCertificate, usePrivateIpToTls));
```
lacikaaa, line 111: is it ok to add this `ServiceConfig` on `bootstrap` and again when we are adding the other ones? It looks like to me we double add it, and I don't know if it cause any issue on CP side

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java
```diff
@@ -183,4 +227,12 @@ private ClientCertificate clientCertificates(Stack stack) {
         return clientCertificate;
     }
 
+    private String getEndpointForRegistration(Stack stack, GatewayConfig gatewayConfig, boolean usePrivateIpToTls) {
+        String ipAddresss = gatewayConfig.getPublicAddress();
+        if (usePrivateIpToTls) {
+            ipAddresss = gatewayConfig.getPrivateAddress();
+        }
+        return String.format("https://%s:%d", ipAddresss, stack.getGatewayport());
```
lacikaaa, line 235: protocol should come from `GatewayConfig` as it is used when we are building `apiAddress`

freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -31,28 +34,38 @@
...
+            Boolean result = !responses.isEmpty() && responses.stream().allMatch(RPCResponse::getResult);
+            if (result && responses.size() == notTermiatedStackInstances.size()) {
                 status = DetailedStackStatus.PROVISIONED;
             } else {
                 status = DetailedStackStatus.UNHEALTHY;
                 postFix = "Freeipa is unhealthy, ";
             }
-            return new SyncResult(postFix + getMessage(response), status, response.getResult());
+            return new SyncResult(postFix + getMessages(responses), status, result);
         } catch (HttpHostConnectException | ConnectTimeoutException e) {
```
lacikaaa, line 69: is this still a valid option? can we receive such exceptions?

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
jamisonbennett: I think both the `CloseableHttpResponse` and the `CloseableHttpClient` should be closed.

https://hc.apache.org/httpcomponents-client-ga/tutorial/html/fundamentals.html shows the following:
```
CloseableHttpResponse response = httpclient.execute(httpget);
try {
    <...>
} finally {
    response.close();
}
```

Also `FreeIpaClientBuilder::connect()` does a close on both the `CloseableHttpResponse` and the `CloseableHttpClient`.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/provision/action/FreeIpaProvisionActions.java
jamisonbennett: The `StackEvent` is still used during the original bootstrapping for the `StackProvisionFlowConfig`. The `ClusterProxyUpdateRegistrationSuccess` is for the new step in the `FreeIpaProvisionFlowConfig`. The `StackProvisionFlowConfig` events remain unchanged. Does. it still make sense to have a `StackEvent` here? If so I will make the change.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java
jamisonbennett: Yes, I confirmed this with the cluster proxy team. The registration is a "force" registration and it overwrites the old config.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java
jamisonbennett: The gateway config is updated to use the protocol of cluster proxy. This needs to be the NGINX endpoint protocol not the protocol of cluster proxy. When registering an endpoint in cluster proxy the needs to be the final NGINX endpoint. I will rename the method to make this a little more clear.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
jamisonbennett: I noticed this yesterday too. You are correct, it is impossible to receive these exceptions. I am in the process of fixing this in CB-6745. I would prefer to fix this in a separate pull request since this wasn't affected by this pull request and this PR is pretty large already.

PR #7880, opened 2020-04-23
CB-6748 Skip executing high state for recipe phases other than pre-cl…

cegganesh84:
…oudera-manager-start

1. Initially, we tried to skip the high state for all the recipe phases.
2. That did not work because the IPA domain names are not resolvable in the pre-cloudera-manager-start phase.
3. So we can re-enable the skip for other phases.


cegganesh84:
@keyki this can be merged now. Thanks!

PR #7886, opened 2020-04-24
CB-6347 Redbeams autosync for start/stop

pkedvessy:
CB-6347 Redbeams autosync for start/stop


redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/action/RedbeamsTerminationActions.java
```diff
@@ -99,6 +103,8 @@ protected Selectable createRequest(RedbeamsContext context) {
                 // through the termination
                 metricService.incrementMetricCounter(MetricType.DB_TERMINATION_FINISHED, context.getDBStack());
                 dbStackService.delete(context.getDBStack());
+                dbStackJobService.unschedule(context.getDBStack());
```
topolyai5, line 106: You should call the `unschedule` at the beginning of the termination. As I see, the termination statuses are not the part of the `getAutoSyncStatuses()`

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Inject
+    private CloudPlatformConnectors cloudPlatformConnectors;
+
+    @Inject
+    private DBStackStatusUpdater dbStackStatusUpdater;
+
+    public void sync(DBStack dbStack) {
+        DetailedDBStackStatus detailedDBStackStatus = getDetailedDBStackStatusFromProvider(dbStack);
+        Status status = detailedDBStackStatus.getStatus();
```
topolyai5, line 47: if any error is occurring at `External DB status lookup failed`, you will get an NPE

redbeams/src/test/java/com/sequenceiq/redbeams/sync/DBStackJobServiceTest.java
```diff
@@ -0,0 +1,68 @@
+package com.sequenceiq.redbeams.sync;
+
+import com.sequenceiq.redbeams.domain.stack.DBStack;
+import com.sequenceiq.statuschecker.service.JobService;
+import org.junit.Before;
```
topolyai5, line 5: please use junit from jupiter package

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
```diff
@@ -0,0 +1,94 @@
...
+        switch (externalDatabaseStatus) {
+            case STARTED: detailedDBStackStatus = DetailedDBStackStatus.STARTED;
+                break;
+            case STOPPED: detailedDBStackStatus = DetailedDBStackStatus.STOPPED;
+                break;
+            case STOP_IN_PROGRESS: detailedDBStackStatus = DetailedDBStackStatus.STOP_IN_PROGRESS;
+                break;
+            case START_IN_PROGRESS: detailedDBStackStatus = DetailedDBStackStatus.START_IN_PROGRESS;
+                break;
+            default: detailedDBStackStatus = DetailedDBStackStatus.UNKNOWN;
```
topolyai5, line 67: is `STARTED ` means `AVAILABLE`? What is the difference between `STARTED` and `AVAILABLE`?

redbeams/src/test/java/com/sequenceiq/redbeams/sync/DBStackJobServiceTest.java
pkedvessy: This module is using Junit 4

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
pkedvessy: Yes, STARTED means AVAILABLE but in a different context. As far as I can see AVAILABLE is used by legacy reasons.

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
pkedvessy: Good catch

redbeams/src/test/java/com/sequenceiq/redbeams/sync/DBStackJobServiceTest.java
topolyai5: I've checked and the jupiter api is added to the module

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
topolyai5: if in STARTED state, the ui show the db as available?

pkedvessy: I have not created new status messages for the UI so if it is shown up "available" after the creation, it should be "available" when the start operation is finished.

redbeams/src/test/java/com/sequenceiq/redbeams/sync/DBStackJobServiceTest.java
pkedvessy: Ahh, I see, this module contains a mixture of Junit 4 and 5 tests.

redbeams/src/main/java/com/sequenceiq/redbeams/sync/DBStackStatusSyncService.java
pkedvessy: Where do we display the status on the UI?
Anyway, the status in the DB is AVAILABLE both after start and create, but after start the detaileddbstackstatus is STARTED.

topolyai5: Oh, I see.

PR #7892, opened 2020-04-24
CB-6745: Allow FreeIPA health check when the stack is unavailable

jamisonbennett:
Fixes the health check so that it is able to run when the stack is
unavailable.

Fixes the stack status so that when it is unavailable it doesn't
bounce between unavailable and unhealthy every minute, but rather it
remains unavailable until it should actually change.

Fixes some exception handling so that it checks the exceptions that
are wrapped inside a FreeIpaClientException.

In addition to updating the tests, this was validated on a cluster
that was stuck in the unavailable status to ensure it became
available after the problem was fixed.

Closes #CB-6745

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
@@ -23,6 +23,7 @@ import com.sequenceiq.freeipa.client.ClusterProxyErrorRpcListener;
 import com.sequenceiq.freeipa.client.FreeIpaClient;
 import com.sequenceiq.freeipa.client.FreeIpaClientBuilder;
 import com.sequenceiq.freeipa.client.FreeIpaClientException;
+import com.sequenceiq.freeipa.client.FreeIpaHostNotAvailableException;
 import com.sequenceiq.freeipa.entity.FreeIpa;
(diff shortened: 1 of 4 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -142,17 +144,17 @@ private boolean canTryAnotherInstance(boolean lastInstance, IOException e) {
...
     }
 
     public FreeIpaClient getFreeIpaClientForStack(Stack stack, String freeIpaFqdn) throws FreeIpaClientException {
         LOGGER.debug("Creating FreeIpaClient for stack {} for {}", stack.getResourceCrn(), freeIpaFqdn);
-        return getFreeIpaClient(stack, false, Optional.of(freeIpaFqdn));
+        return getFreeIpaClient(stack, false, false, Optional.of(freeIpaFqdn));
     }
 
-    public FreeIpaClient getFreeIpaClientForStackWithPing(Stack stack, String freeIpaFqdn) throws Exception {
+    public FreeIpaClient getFreeIpaClientForStackWithPingAndForceCheckUnreachableStack(Stack stack, String freeIpaFqdn) throws Exception {
```
lacikaaa, line 155: this method name get a little bit out of hand :)
could you make it shorter, like getFreeIpaClientForCheck or something similar

PR #7893, opened 2020-04-24
CB-6597, 6616, 6624, 6625 - Refactor SDX upgrade endpoint - 4 call in…

pdarvasi:
…stead of 8

Contains fixes for the following issues:
CB-6254
CB-6595
CB-6597
CB-6616
CB-6624
CB-6625
CB-6718


akanto:
I think it is good if we are pushing this into master. Just to avoid any unwanted side-effect in 2.20

PR #7895, opened 2020-04-24
CDPCP-1561. Allow internal actor to call user sync endpoint

aarman-cloudera:
Previously the accountId to sync was inferred from the calling
user, which doesn't work when the calling user is the internal
actor. This is fixed by including accountId as an optional
parameter to the user sync request.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
@@ -5,8 +5,10 @@ import java.util.Set;
 import javax.inject.Inject;
 import javax.validation.constraints.NotNull;
 
+import com.sequenceiq.cloudbreak.auth.security.InternalCrnBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.security.access.AccessDeniedException;
 import org.springframework.stereotype.Controller;
 
 import com.sequenceiq.authorization.annotation.CheckPermissionByAccount;
@@ -76,7 +78,8 @@ public class UserV1Controller implements UserV1Endpoint {
     @CheckPermissionByAccount(action = AuthorizationResourceAction.WRITE)
     public SyncOperationStatus synchronizeAllUsers(SynchronizeAllUsersRequest request) {
         String userCrn = checkUserCrn();
-        String accountId = ThreadBasedUserCrnProvider.getAccountId();
+        String accountId = determineAccountId(userCrn, request.getAccountId());
+
         LOGGER.debug("synchronizeAllUsers() requested for account {}", accountId);
 
         return checkOperationRejected(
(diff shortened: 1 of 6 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -107,6 +109,16 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    private String determineAccountId(String callingActor, String requestedAccountId) {
+        if (requestedAccountId == null) {
+            return ThreadBasedUserCrnProvider.getAccountId();
+        }
+        if (!InternalCrnBuilder.isInternalCrn(callingActor)) {
```
handavid, line 116: you should also allow setting the account id to your own account

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -402,6 +402,16 @@ void validateParameters(String accountId, String actorCrn, Set<String> environme
         validateCrnFilter(environmentCrnFilter, Crn.ResourceType.ENVIRONMENT);
         validateCrnFilter(userCrnFilter, Crn.ResourceType.USER);
         validateCrnFilter(machineUserCrnFilter, Crn.ResourceType.MACHINE_USER);
+        validateSameAccount(actorCrn, accountId, Iterables.concat(environmentCrnFilter, userCrnFilter, machineUserCrnFilter));
+    }
+
+    private void validateSameAccount(String actorCrn, String accountId, Iterable<String> crns) {
```
handavid, line 408: actor crn not needed here

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
aarman-cloudera: Fixed.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -107,6 +109,17 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    private String determineAccountId(String callingActor, String requestedAccountId) {
+        if (requestedAccountId == null) {
+            return ThreadBasedUserCrnProvider.getAccountId();
+        }
+        String callingActorAccountId = Crn.safeFromString(callingActor).getAccountId();
+        if (!callingActorAccountId.equals(requestedAccountId) && !InternalCrnBuilder.isInternalCrn(callingActor)) {
+            throw new BadRequestException("Only the internal actor can set a different accountId");
```
lacikaaa, line 118: do we really want to propagate that information to customers if somehow they can invoke it with different accountid?

handavid: I think that permission denied be more appropriate than bad request then. 

The message should just say that the user isn't allowed to make this call. I agree that we don't want to reveal that internal actors exist.

aarman-cloudera: We're not exposing this to customers (they won't have an option to set an accountId through public api). They might somehow figure out how to call this endpoint without going through the public api but we don't need to give friendly message in this case. The message as it is right now is more useful to internal developers.

giladwolff: I agree that it should be Permission Denied.  I would prefer to have the same exception message as thrown by the Scrutinizer (see validateActorCrnInAccountOrInternal). You can put any other information you think is useful in a log message.

aarman-cloudera: Thanks, I've made the changes.

PR #7897, opened 2020-04-25
CB-6605 only sync env related WAGs

keyki:
Test against mow-dev
```
before
users: 1305
total groups: 2000 <- we're already missing user groups
wl groups: 1946
```
```
after
users: 1305
total groups: 437
wl groups: 48 (1 env 8 groups)
```
The original WAG list and the user member of group list is fetched at a different time
so they can be different as they can launch environments in between
(probably we need to filter there as well), but the `listWorkloadAdministrationGroupsForMember`
only returns the name of the group so there is nothing to filter on.
Locally it's much slower to connect to UMS than from mow-dev so this will not be that different there

Crn handling needed: Error during UMS operation: 'cdpp is not a valid partition value'
There are also weird WAGs
```
Not adding to filter: _c_wag_testers_34ef5c69 crn: null
Not adding to filter: _c_wag_testers_1bd31292 crn: crn:altus:iam:us-west-1:9d74eee4-1cad-45d7-b645-7ccf9edbb73d:user:59cbfafb-999b-4df9-81e2-aa00a88382e9
Not adding to filter: _c_ml_users_34ef5c69 crn: null
Not adding to filter: _c_ml_users_5c217dc6 crn: null
Not adding to filter: _c_ml_admins_34ef5c69 crn: null
Not adding to filter: _c_ml_admins_2c888e37 crn: null
Not adding to filter: _c_ml_admins_5c217dc6 crn: null
Not adding to filter: _c_ml_admins_43a040f6 crn: null
```

Similar approach #7898. However, this does not sync any other WAGs besides env related ones, if there are any.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient.INTERNAL_ACTOR_
 import static java.util.Objects.requireNonNull;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -22,6 +23,7 @@ import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.Machi
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.User;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.WorkloadAdministrationGroup;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.altus.Crn.ResourceType;
 import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
 import com.sequenceiq.cloudbreak.auth.altus.exception.UmsOperationException;
 import com.sequenceiq.freeipa.service.freeipa.user.model.EnvironmentAccessRights;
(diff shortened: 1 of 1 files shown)


keyki:
> LGTM. I don't quite understand why UMS doesn't do filtering instead of CB needing to filter everything. Right now I expect this is a LOT of unnecessary reads from UMS.

API wise it's a lot worse. We fetch details for users like workload passwords, membership as separate calls. If you multiply these per users we call UMS more than 10k times for a single environment.  https://jira.cloudera.com/browse/CDPCP-1091

cmleroy:
Can we make sure to add some unit testing for this? 

In terms of where the filtering is done: there's definitely improvements that can be made. The team has a reasonable write up on this. They are prioritizing customer impacting stuff over making improvements to support setups that have unrealistic numbers of resources, but they may make improvements outside the current fire drill.

One thing to keep in mind when looking at how things are done: they want to keep both the UMS and FMS unaware of resource relationships -- so they don't want to be in a situation where either needs to be able to take an X (workspace, datahub, etc.) and determine it's environment. At present, this is not a big deal as assignment is always done on the environment, but that is changing with more flexible RBAC and new cross-environment resource types coming on board.

keyki:
Yes. Unit test are needed. We will address that part as well, but I don't want to block pople on this this it's will be a separatete PR.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -230,4 +247,14 @@ private EnvironmentAccessChecker createEnvironmentAccessChecker(String environme
...
+
+    private Crn getCrn(WorkloadAdministrationGroup wag) {
+        Crn resourceCrn = null;
+        try {
+            resourceCrn = Crn.fromString(wag.getResource());
+        } catch (Exception e) {
+            LOGGER.debug("Invalid resource is assigned to workload admin group: {}", e.getMessage());
+        }
+        return resourceCrn;
+    }
```
giladwolff, line 259: So we have these utility functions in Crns.java and Crn.java to do things like Crns.safeFromString (that can't return null) and Crns.forTypr(String, Service, ResourceType) that ensures that a crn belongs to a service and is of a specific resource type. It may be useful to bring that in here.

keyki: Thanks. I'll bring that one in along with some tests.

PR #7898, opened 2020-04-25
CB-6605 only sync env related WAGs

keyki:
Only sync those WAGs that are related to the environment. 

#7897

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
@@ -67,6 +67,14 @@ public class UsersState {
             return this;
         }
 
+        public Builder addMemberToGroupIfGroupExists(String group, String user) {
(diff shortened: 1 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersState.java
```diff
@@ -67,6 +67,14 @@ public Builder addMemberToGroup(String group, String user) {
             return this;
         }
 
+        public Builder addMemberToGroupIfGroupExists(String group, String user) {
+            boolean groupExists = fmsGroups.stream().anyMatch(g -> g.getName().equals(group));
```
cegganesh84, line 71: I like the idea. It looks like even if the group is deleted, the associations are not deleted in IAM. So the users' group association keeps growing in IAM. This fix tries to reduce the group footprint in FreeIPA by not adding such groups.

Codewise, is it possible to create a HashSet for fmsGroups' name?

PR #7900, opened 2020-04-27
CB-5952. Handle SMM CM configs before version 7.1.1

oleewere:


oleewere:
waiting for @benyoka to make sure SMM operates as expected

oleewere:
although deployment failed for me with a different error, but the configuration part passed and it was set properly. I will merge this, if I would not do tomorrow e2e tests will fail on this anyway.I can revert changes if SMM functionally will be different for some reason

PR #7911, opened 2020-04-28
CB-6287 - Wait for ParcelUpdateService to complete before downloading

pdarvasi:
CB-6287 - Wait for ParcelUpdateService to complete before downloading… …
CB-6797 - ImageV4 response should return readable creation date

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/ClouderaManagerParcelsApiListenerTask.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/ClouderaManagerParcelsApiListenerTask.java
@@ -1,0 +1,59 @@
+package com.sequenceiq.cloudbreak.cm.polling.task;
+
+import java.util.Objects;
+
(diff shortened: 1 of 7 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/ClouderaManagerParcelsApiListenerTask.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    protected boolean doStatusCheck(ClouderaManagerPollerObject pollerObject, CommandsResourceApi commandsResourceApi) {
+        try {
+            boolean parcelsAvailable = pollParcels(pollerObject);
+            LOGGER.debug("Polling for parcel's availability returned: {}", parcelsAvailable);
+            return parcelsAvailable;
+        } catch (ApiException e) {
+            LOGGER.debug("Cloudera Manager Parcels API is not available.");
```
keyki, line 34: Can you please extend the log message with the stacktrace?

PR #7914, opened 2020-04-28
CB-6808. Removing old metering rpm is not required anymore.

oleewere:
it was needed as in older images, the rpm name changed in the same version mow CB version (with different build - it means the new rpm was not installed, but old one used the same files as the new one, so first, it was required to remove that). now it's not required anymore to remove the old one

also it includes to mode <-> file_mode fixes


cegganesh84:
Thanks!

PR #7915, opened 2020-04-28
CB-6798 Environment service should not deploy nat gateway if private …

doktoric:


keyki:
LGTM. I would wait until we merge and build #7916 to avoid accidental issues.

keyki:
I'll merge once we deployed 2.20 to prod

PR #7919, opened 2020-04-28
CB-6628 check if freeipa client is usable during usersync

lacikaaa:


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientException.java
```diff
@@ -29,4 +29,8 @@ public FreeIpaClientException(String message, Throwable cause, OptionalInt statu
     public OptionalInt getStatusCode() {
         return statusCode;
     }
+
+    public boolean isClientUnusable() {
+        return FreeIpaClientExceptionUtil.isClientUnusable(this);
```
handavid, line 34: Can we move this to the constructor and store a boolean flag? I think it would be better to compute this value once per exception.

lacikaaa: I was thinking about the same, but ideally you would call it once, and also the whole code is based on passing the `FreeIpaClientException` to the util class, and we shouldn't do that in construction time. If it becomes a use-case that we call this method multiple times for the same exception I can refactor it

handavid: sounds good. thanks!

PR #7922, opened 2020-04-28
Add backup and restore scripts for DL postgres db.

brycederriso:
Create bash scripts to run PostgreSQL dump and restore against `hive` and `ranger` databases.
* Add `backup_db.sh`
* Add `restore_db.sh`

Scripts accept:
1. Cloud Provider (azure | aws)
2. PostgreSQL host name.
3. PostgreSQL port.
4. Object Storage Service url to save backups.
5. PostgreSQL user name.
6. PostgreSQL password.

The backup script stores backups in the Cloud Object storage under a Date String prefix, e.g. `s3://backup-bucket/2020-04-28T18:12:53Z`.

Storage service URLs are different when doing backup versus restore. The parent folder is specified for backup, while the particular date folder should be specified for restore.
* Backup to `s3://backup-bucket`
* Restore from `s3://backup-bucket/2020-04-28T18:12:53Z`

The backup script:
1. Runs `pg_dump`, outputting to a local file
2. Copies the dump to Cloud Storage

Restore does the reverse:
1. Copies a dump from Cloud storage to the master instance.
2. Modifies the `list` file to remove some DB objects from restoration.
3. Runs `pg_restore` using the local dump file.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
@@ -3,7 +3,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-if [ $# -ne 5 ] || [ $# -ne 6 ]; then
+if [ $# -ne 5 ] && [ $# -ne 6 ]; then
   echo "Invalid inputs provided"
   echo "Script accepts 5 inputs:"
   echo "  1. Cloud Provider (azure | aws)"
@@ -17,9 +17,9 @@ if [ $# -ne 5 ] || [ $# -ne 6 ]; then
 fi
 
 CLOUD_PROVIDER="$1"
-HOST="$2"
-PORT="$3"
-BACKUP_LOCATION=$(echo "$4"| sed "s/\/\+$//g") # Clear trailng '/' (if present) for later path joining.
+BACKUP_LOCATION=$(echo "$2"| sed "s/\/\+$//g") # Clear trailng '/' (if present) for later path joining.
+HOST="$3"
+PORT="$4"
 USERNAME="$5"
 export PGPASSWORD="$6" # We can provide the password to pg_dump through this variable, or in ~/.pgpass
 
@@ -83,7 +83,7 @@ run_aws_backup () {
   dump_to_s3 "ranger"
 }
 
-doLog "INFO Starting backup to ${$BACKUP_LOCATION}"
+doLog "INFO Starting backup to ${BACKUP_LOCATION}"
 
 if [[ "$CLOUD_PROVIDER" = "azure" ]]; then
   run_azure_backup
@@ -93,4 +93,5 @@ else
   errorExit "Unknown cloud provider: ${CLOUD_PROVIDER}"
 fi
 
-doLog "INFO Completed backup ${$BACKUP_LOCATION}"
+doLog "INFO Completed backup ${BACKUP_LOCATION}"
+
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
@@ -3,15 +3,16 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-if [ $# -ne 6 ]; then
+if [$# -ne 5] || [ $# -ne 6 ]; then
   echo "Invalid inputs provided"
-  echo "Script accepts six inputs:"
+  echo "Script accepts 5 inputs:"
   echo "  1. Cloud Provider (azure | aws)"
-  echo "  2. Object Storage Service url to place backups."
+  echo "  2. Object Storage Service url to retrieve backups."
   echo "  3. PostgreSQL host name."
   echo "  4. PostgreSQL port."
   echo "  5. PostgreSQL user name."
-  echo "  6. PostgreSQL password."
+  echo
+  echo "  Optional: PostgreSQL password."
   exit 1
 fi
 
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
@@ -23,7 +23,6 @@ BACKUP_LOCATION="$4"
 USERNAME="$5"
 export PGPASSWORD="$6" # We can provide the password to pg_dump through this variable, or in ~/.pgpass
 
-BACKUP_DATE=$(date '+%Y-%m-%dT%H:%M:%SZ')
 LOGFILE=/var/log/dl_postgres_backup.log
 
 echo "Logs at ${LOGFILE}"
@@ -62,7 +61,7 @@ dump_to_azure() {
 }
 run_azure_backup () {
   BACKUPS_DIR="/var/tmp/" # Is this appropriate for keeping backups that are staged before uploading to Azure Blob Storage?
-  DATE_DIR=${BACKUPS_DIR}/${BACKUP_DATE}
+  DATE_DIR=${BACKUPS_DIR}/$(date '+%Y-%m-%dT%H:%M:%SZ')
   mkdir -p "$DATE_DIR" || error_exit "Could not create local directory for backups."
 
   azcopy login --identity || errorExit "Could not login to Azure"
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,98 @@
...
+fi
+
+CLOUD_PROVIDER="$1"
+HOST="$2"
+PORT="$3"
+BACKUP_LOCATION="$4"
+USERNAME="$5"
+export PGPASSWORD="$6" # We can provide the password to pg_dump through this variable, or in ~/.pgpass
+
+BACKUP_DATE=$(date '+%Y-%m-%dT%H:%M:%SZ')
```
hreeve-cloudera, line 25: The date directory part of the backup location will be coming from the datalake DR service calls, since we need to sync that across the different backup types. So you don't need to generate it dynamically in the the scripts.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,98 @@
...
+  echo "  6. PostgreSQL password."
+  exit 1
+fi
+
+CLOUD_PROVIDER="$1"
+HOST="$2"
+PORT="$3"
+BACKUP_LOCATION="$4"
+USERNAME="$5"
+export PGPASSWORD="$6" # We can provide the password to pg_dump through this variable, or in ~/.pgpass
```
hreeve-cloudera, line 23: If you're exporting this, you might want to unset it at the end of the script. I don't actually think it'll live outside the script's context, but just in case.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [ $# -ne 6 ]; then
+  echo "Invalid inputs provided"
+  echo "Script accepts five inputs:"
```
hreeve-cloudera, line 8: "Script accepts six inputs:"

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/restore_db.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [ $# -ne 6 ]; then
+  echo "Invalid inputs provided"
+  echo "Script accepts five inputs:"
```
hreeve-cloudera, line 8: "Script accepts six inputs:"

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,98 @@
...
+  dump_to_azure "hive"
+  dump_to_azure "ranger"
+
+  rmdir -v "$DATE_DIR" >>$LOGFILE 2>&1
+}
+
+dump_to_s3() {
+  SERVICE=$1
+  # todo: strip trailing slash from BACKUP_LOCATION if it's there
+  S3_LOCATION="${BACKUP_LOCATION}/${BACKUP_DATE}/${SERVICE}_backup"
```
kkalvagadda1, line 77: If the S3 location is created here, How will the Caller provide appropriate location for restore?

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,98 @@
...
+
+if [ $# -ne 6 ]; then
+  echo "Invalid inputs provided"
+  echo "Script accepts five inputs:"
+  echo "  1. Cloud Provider (azure | aws)"
+  echo "  2. PostgreSQL host name."
+  echo "  3. PostgreSQL port."
+  echo "  4. Object Storage Service url to place backups."
+  echo "  5. PostgreSQL user name."
+  echo "  6. PostgreSQL password."
```
kkalvagadda1, line 14: Re-order the inputs to be in-order. Move Object store URL to the end so that all the database configurations will be together.

brycederriso: My intent with password last was to make it easier to remove the password altogether if/when we migrate to using the recommended `.pgpass` file for passwords.

How about
```suggestion
  echo "  1. Cloud Provider (azure | aws)"
  echo "  2. Object Storage Service url to place backups."
  echo "  3. PostgreSQL host name."
  echo "  4. PostgreSQL port."
  echo "  5. PostgreSQL user name."
  echo "  6. PostgreSQL password."
```

hreeve-cloudera: Make sure you don't actually delete the part that explains they're missing parameters. And you might explicitly mark `PostgresSQL password` as `optional`. I didn't realize it was, which is why I ticked the mentioned number of parameters.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
brycederriso: I think this should really live in the `.pgpass` file for proper security. I'll think 
 (talk) about whether we need to that now or later.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
brycederriso: I'll change this to only create the `${SERVICE}_backup` portion at the end of the s3 location.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
brycederriso: Ah, you're giving me too much credit, I hadn't thought of making it optional. I'll add the accoutrements to make it so.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
brycederriso: Done, but I'm still generating a date for the temp files that Azure must create before uploading to blob storage -- if you notice the `$(date '+%Y-%m-%dT%H:%M:%SZ')` still hanging around.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
brycederriso: Postgres will attempt to read the `~/.pgpass` file if it's present. So we should see about adding that to Cloudbreak.

keyki:
There is one thing I'm not sure if we need it or not. For FreeIPA we also upload the backups to cloud storage and that script supports a proxy. I'm not sure what is the expectation from the customer side. If the cloud storage must have direct access anyway we don't need it otherwise we probably do. Both azcopy and s3 support the standard env variables. This is just a fyi to keep in mind when we invoke the scripts from salt.

orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/backup_db.sh
```diff
@@ -0,0 +1,96 @@
...
+  echo "  Optional: PostgreSQL password."
+  exit 1
+fi
+
+CLOUD_PROVIDER="$1"
+BACKUP_LOCATION=$(echo "$2"| sed 's/\/\+$//g') # Clear trailng '/' (if present) for later path joining.
+HOST="$3"
+PORT="$4"
+USERNAME="$5"
+export PGPASSWORD="$6" # We can provide the password to pg_dump through this variable, or in ~/.pgpass
```
keyki, line 24: I would not export the password as it will be part of the logs. In Salt we can simply pass passwords to the script without exposing it. Example: https://github.com/hortonworks/cloudbreak/blob/bb5ebdf54b0f987e232e381960cf6dc2d66ea8ed/freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls#L5

brycederriso: The problem actually lies in passing the password to the `pg_dump` command, there isn't any option to provide it as a command line argument -- if we stick to using a _password_ we need to either:
* Provide it interactively
* Put the password in the special `PGPASSWORD` variable
* Put the password as a connection string in `~/.pgpass`, and `chmod 0600` it for root.

brycederriso: @keyki has explained that we can use Salt to define the `PGPASSWORD` variable in the environment, so there's no need to set it + export it in the backup or restore scripts.

This prevents accidental exposure of the password by `set -x`.

PR #7924, opened 2020-04-29
CB-6729 Sdx create should not wait for stopped environment

sodre90:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -209,6 +211,12 @@ public SdxCluster getSdxByNameInAccount(String userCrn, String name) {
         return Pair.of(sdxCluster, flowIdentifier);
     }
 
+    private void validateEnv(DetailedEnvironmentResponse environment) {
+        if (ENV_STOPPED.equals(environment.getEnvironmentStatus())) {
```
gergopapi2, line 215: I think it would make sense to prepare the code for environment currently stopping. As I see env service will wait until datahubs etc. are stopped so it may take a considerable amount if time until it is stopped.

gergopapi2: And maybe env starting should also be filtered out, as the env might only partially be functional.

lnardai: Might be a good idea to collect all the env statuses to a list where the poller 
should immediately exit + we should not allow creation. I'm guessing the 2 list should be identical.

PR #7925, opened 2020-04-29
CB-6719 Disable start stop functionality on Azure

pkedvessy:
CB-6719 Disable start stop functionality on Azure

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
@@ -165,18 +165,17 @@ public class AzureResourceConnector implements ResourceConnector<Map<String, Map
 
     @Override
     public void startDatabaseServer(AuthenticatedContext authenticatedContext, String dbInstanceIdentifier) {
-        LOGGER.warn("Database server start operation is not supported for " + getClass().getName());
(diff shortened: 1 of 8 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/startstop/DatabaseStartStopSupportService.java
```diff
@@ -0,0 +1,26 @@
...
+public class DatabaseStartStopSupportService {
+
+    @Inject
+    private EnvironmentClientService environmentClientService;
+
+    public boolean supportsDatabaseStartStop(SdxCluster sdxCluster) {
+        if (sdxCluster.hasExternalDatabase() && Strings.isNotEmpty(sdxCluster.getDatabaseCrn())) {
+            DetailedEnvironmentResponse environment = environmentClientService.getByCrn(sdxCluster.getEnvCrn());
+
+            return CloudPlatform.AWS.name().equalsIgnoreCase(environment.getCloudPlatform());
```
doktoric, line 22: can this be part of the resourceconnector ?  I think that is much flexible so create a method there like databasePauseSupported()

pkedvessy: As we have discussed, it can't be a part of the ResourceConnector but it got moved to the PlatformConfig.

PR #7929, opened 2020-04-29
CB-6371 Handle spot instance terminations

Bajzathd:


keyki:
Can you please send it to master? Unless we know it's a blocker for 2.20

PR #7931, opened 2020-04-29
CB-6773 read username from crn in the test

topolyai5:


keyki:
Improves the IT usability so can be merged.

PR #7932, opened 2020-04-29
CB-6816. Do not downgrade fluent plugins on newer images.

oleewere:
it is required to avoid this situation:
- older Cluster (like 2.19)
- newer image (as some changes were required for 2.20, as we are building images from master -> for a newer 2.19, probably we will upgrade some binaries/plugins there)
- when applying 2.19 salt script -> in the newer image, it's possible there are newer plugins, this check plugins script will delete those plugins and will re-install the old versions.
(of course that will be fixed from CB 2.20...just described that problem with a 2.19 example)


oleewere:
DO NOT MERGE YET. i will testing this a bit

keyki:
You can convert to Draft. Below the reviewers, there is a button.

PR #7934, opened 2020-04-29
CB-6788 Check right - skip only read, back to account check on cred

horadla23:


horadla23:
I removed entitlement check from `hasRights` because it is used for usersync only (which is not checking for `read` rights) and other calls from code of `hasRights` (which is resource based part of authz framework) is not used in 2.20.

PR #7935, opened 2020-04-29
CB-6816. Do not downgrade fluent plugins on newer images.

oleewere:
creating a new PR again - as https://github.com/hortonworks/cloudbreak/pull/7932 was merged before deployment, it was reverted


keyki:
I'll merge once we deployed 2.20 to prod

PR #7939, opened 2020-04-30
DISTX-421 DISTX-427 Cluster Definition to contain auto-scaling specifications

cegganesh84:
* Added ScalingMode to InstanceGroup specification.
 * It specifies whether an instance group supports the scaling option.
 * Instance group with auto-scaling mode is both auto-scalable and manually scalable.
 * None means the instance group is not scalable.
 * If unspecified, the instance group can be scaled manually. This is the current behavior for backward compatibility.
 * Unless specified none, the instance groups can be scaled manually.
 * For now modified only DE HA template for DISTX-427 as a POC.
 * If the approach looks fine, will modify the DE templates and file a JIRA for UI modifications.

./gradlew build && testing in private stack.

### DIFF
--- a/common-model/src/main/java/com/sequenceiq/common/api/type/ScalingMode.java
+++ b/common-model/src/main/java/com/sequenceiq/common/api/type/ScalingMode.java
@@ -1,0 +1,20 @@
+package com.sequenceiq.common.api.type;
+
+/**
+ * Specifies whether an instance group supports scaling option.
+ * Instance group with auto scaling mode is both auto scalable and manually scalable.
+ * None means the instance group is not scalable.
+ * If unspecified, the instance group can be scaled manually. This is the current behaviour for backward compatibility.
+ * Unless specified none, the instance groups can be scaled manually.
+ */
+public enum ScalingMode {
+    MANUAL, AUTO, NONE, UNSPECIFIED;
+
+    public boolean isAutoscalable() {
+        return this == AUTO;
+    }
+
+    public boolean isManuallyScalable() {
+        return this != NONE;
+    }
+}
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
@@ -22,6 +22,7 @@ import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
 import javax.persistence.SequenceGenerator;
 
+import com.sequenceiq.common.api.type.ScalingMode;
 import org.apache.commons.lang3.StringUtils;
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.InstanceStatus;
@@ -58,6 +59,9 @@ public class InstanceGroup implements ProvisionEntity, Comparable<InstanceGroup>
 
     @Enumerated(EnumType.STRING)
     private InstanceGroupType instanceGroupType = InstanceGroupType.CORE;
+
+    @Enumerated(EnumType.STRING)
+    private ScalingMode scalingMode = ScalingMode.UNSPECIFIED;
 
     @ManyToOne
     private Stack stack;
@@ -151,6 +155,14 @@ public class InstanceGroup implements ProvisionEntity, Comparable<InstanceGroup>
 
     public void setSecurityGroup(SecurityGroup securityGroup) {
         this.securityGroup = securityGroup;
+    }
+
+    public ScalingMode getScalingMode() {
+        return scalingMode;
+    }
+
+    public void setScalingMode(ScalingMode scalingMode) {
+        this.scalingMode = scalingMode;
     }
 
     public Set<InstanceMetaData> getAllInstanceMetaData() {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java
@@ -11,6 +11,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
@@ -119,6 +120,11 @@ public class ClusterCommonService {
         if (hostGroup.isEmpty()) {
             throw new BadRequestException(String.format("Host group '%s' not found or not member of the cluster '%s'",
                     updateJson.getHostGroupAdjustment().getHostGroup(), stack.getName()));
+        }
+        InstanceGroup instanceGroup = hostGroup.get().getInstanceGroup();
+        if (!instanceGroup.getScalingMode().isManuallyScalable()) {
+            throw new BadRequestException(String.format("Instance group '%s' in stack '%s' is not enabled to scale",
+                    instanceGroup.getGroupName(), stack.getName()));
         }
         if (blueprintService.isClouderaManagerTemplate(blueprint)) {
             cmTemplateValidator.validateHostGroupScalingRequest(blueprint, hostGroup.get(),
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/UpdateNodeCountValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/UpdateNodeCountValidator.java
@@ -101,6 +101,9 @@ public class UpdateNodeCountValidator {
         if (instanceGroup == null) {
             throw new BadRequestException(format("Stack '%s' does not have an instanceGroup named '%s'.", stack.getName(), instanceGroupName));
         }
+        if (!instanceGroup.getScalingMode().isManuallyScalable()) {
+            throw new BadRequestException(format("Instance group '%s' in stack '%s' is not enabled to scale", instanceGroupName, stack.getName()));
+        }
     }
 
     public void validateScalingAdjustment(InstanceGroupAdjustmentV4Request instanceGroupAdjustmentJson, Stack stack) {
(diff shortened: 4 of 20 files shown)


core/src/main/resources/schema/app/20200428235034_DISTX-421_add_scalingmode_to_instancegroup.sql
```diff
@@ -0,0 +1,11 @@
+-- // DISTX-421 Add scalingmode column to instancegroup table
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE instancegroup ADD IF NOT EXISTS scalingmode VARCHAR(15) DEFAULT 'UNSPECIFIED';
```
keyki, line 4: I think we usually don't set the default value like this as it blocks the table during an upgrade. @lacikaaa do you have an example for this?

lacikaaa: yes, you have to add it in separate steps to avoid lock, like this:
https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/schema/app/20200406095012_CB-6385_add_version_to_stack.sql
the result should be the same

cegganesh84: Thanks. Addressed.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
```diff
@@ -59,6 +60,9 @@
     @Enumerated(EnumType.STRING)
     private InstanceGroupType instanceGroupType = InstanceGroupType.CORE;
 
+    @Enumerated(EnumType.STRING)
```
lacikaaa, line 63: please use converter as it's more safe for future changes. see: https://github.com/hortonworks/cloudbreak/blob/master/common/src/main/java/com/sequenceiq/cloudbreak/converter/TunnelConverter.java

core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java
```diff
@@ -120,6 +121,11 @@ private FlowIdentifier clusterHostgroupAdjustmentChange(Long stackId, UpdateClus
             throw new BadRequestException(String.format("Host group '%s' not found or not member of the cluster '%s'",
                     updateJson.getHostGroupAdjustment().getHostGroup(), stack.getName()));
         }
+        InstanceGroup instanceGroup = hostGroup.get().getInstanceGroup();
+        if (!instanceGroup.getScalingMode().isManuallyScalable()) {
+            throw new BadRequestException(String.format("Instance group '%s' in stack '%s' is not enabled to scale",
+                    instanceGroup.getGroupName(), stack.getName()));
+        }
```
lacikaaa, line 128: move into a method, like `validateInstancgroupScalebality`

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/UpdateNodeCountValidator.java
```diff
@@ -101,6 +101,9 @@ public void validateInstanceGroup(Stack stack, String instanceGroupName) {
         if (instanceGroup == null) {
             throw new BadRequestException(format("Stack '%s' does not have an instanceGroup named '%s'.", stack.getName(), instanceGroupName));
         }
+        if (!instanceGroup.getScalingMode().isManuallyScalable()) {
+            throw new BadRequestException(format("Instance group '%s' in stack '%s' is not enabled to scale", instanceGroupName, stack.getName()));
+        }
```
lacikaaa, line 106: this is very similar to the previous one, could you extract this logic into a class and reuse it?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceGroupV1ToInstanceGroupV4Converter.java
```diff
@@ -67,6 +73,11 @@ private InstanceGroupV1Request convert(InstanceGroupV4Request source) {
         response.setRecipeNames(source.getRecipeNames());
         response.setAws(getIfNotNull(source.getAws(), instanceGroupParameterConverter::convert));
         response.setAzure(getIfNotNull(source.getAzure(), instanceGroupParameterConverter::convert));
+        if (source.getScalingMode() == null) {
+            response.setScalingMode(ScalingMode.UNSPECIFIED);
+        } else {
+            response.setScalingMode(source.getScalingMode());
+        }
```
lacikaaa, line 80: this logic looks duplicated to me, could you extract it to a method?

core/src/main/resources/schema/app/20200428235034_DISTX-421_add_scalingmode_to_instancegroup.sql
```diff
@@ -0,0 +1,12 @@
+-- // DISTX-421 Add scalingmode column to instancegroup table
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE instancegroup ADD IF NOT EXISTS scalingmode VARCHAR(15);
```
lacikaaa, line 4: `VARCHAR(15)` this seems a little bit restrictive. I think we usually use 255 for enums

common-model/src/main/java/com/sequenceiq/common/api/type/ScalingMode.java
```diff
@@ -0,0 +1,20 @@
...
+ * Specifies whether an instance group supports scaling option.
+ * Instance group with auto scaling mode is both auto scalable and manually scalable.
+ * None means the instance group is not scalable.
+ * If unspecified, the instance group can be scaled manually. This is the current behaviour for backward compatibility.
+ * Unless specified none, the instance groups can be scaled manually.
+ */
+public enum ScalingMode {
+    MANUAL, AUTO, NONE, UNSPECIFIED;
+
+    public boolean isAutoscalable() {
```
smaniraju, line 13: Manually Scalable hostGroups would also be AutoScalable right, should Manual\Auto be combined into one enum ? Also considering custom templates would there be validation that a hostGroup marked Auto\Manual contains only scalable services ?

cegganesh84: Hi @smaniraju , Not all manually scalable hostGroups would be auto-scalable. For example gateway nodes. We do not have a mechanism to detect auto-scalability. The converse is true. Hence they are split into two different enums. Regarding your other question, the idea to put this in cluster template instead of hard coding it, is to give flexibility for the cluster shapes to choose auto-scalable components by themselves. For example, an OpDB template may choose some of the HBase node groups to be auto-scalable. I do not know how the backend would work, but this change enables such functionality.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
cegganesh84: Thanks, done!

core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java
cegganesh84: Thanks, done!

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/UpdateNodeCountValidator.java
cegganesh84: Thanks, done!

core/src/main/resources/schema/app/20200428235034_DISTX-421_add_scalingmode_to_instancegroup.sql
cegganesh84: Thanks, done!

core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java
cegganesh84: Thanks, done!

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceGroupV1ToInstanceGroupV4Converter.java
cegganesh84: The types are different, it gets more complicated to de-duplicate. Instead removed that logic!

common-model/src/main/java/com/sequenceiq/common/api/type/ScalingMode.java
cegganesh84: Removed manual and auto enums for now. This will unblock UI efforts. Will send PR for auto or algo related scalability enum to the specific branch.

lacikaaa: why did you have to remove those?

doktoric:
I think on the meeting we discussed that we should decide based on service and like this way. So we need to check the roles inside the CM template.

common-model/src/main/java/com/sequenceiq/common/api/type/ScalingMode.java
cegganesh84: So, the differentiation of manual and auto is unclear. Even within the auto, there are rule-based and schedule based differences. To solve DISTX-427, I wanted to get this in quicker so that UI for the Resize button can be adjusted in the master branch.

For the auto/algo/time/rule differences can put that in the auto-scaling branch for auto-scaling related UI efforts DISTX-421.

sidseth:
Until a component aware API is in place, this serves as a simple mechanism to determine whether a hostGroup can be auto scaled or not (and also very importantly whether it just cannot be scaled. e.g. Running master nodes). Once a component aware API exists - this would serve as an override for customers who really know what they are doing, and potentially even to facilitate testing.

Any component aware capabilities system would need to handle multiple scenarios (probably in steps) - spot vs non-spot, AutoScalable vs non-autoscalable,  scale up vs scale down, component versions, etc.

cegganesh84:
@doktoric I have put together various arguments in a doc for this https://docs.google.com/document/d/1XNLQpZVAyzOgQ77Rz_G63b3SA1dZHzDImx1ofc1kOxM/edit#

PR #7945, opened 2020-04-30
CDPD-6804 Add Data Discovery and Exploration initial template and definitions to …

gezan:
Introducing Solr in DataHub.
Discovery Data Hub is a collective name for DataHubs containing Solr and and other components improving user experience around Solr.


core/src/main/resources/application.yml
```diff
@@ -256,6 +259,7 @@ cb:
         7.2.0 - Streams Messaging Light Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-small-720;
         7.2.0 - Flow Management Light Duty with Apache NiFi, Apache NiFi Registry=cdp-flow-management-small-720;
         7.2.0 - Flow Management Heavy Duty with Apache NiFi, Apache NiFi Registry=cdp-flow-management-720;
+        7.1.0 - Discovery Data Hub: Apache Solr, Apache Spark, Hue=cdp-ddh-710;
```
risdenk, line 262: This should be `7.2.0` and `cdp-ddh-720`

core/src/main/resources/application.yml
```diff
@@ -223,6 +223,7 @@ cb:
         CDP 1.2 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha-702;
         CDP 1.2 - Streams Messaging Heavy Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-702;
         CDP 1.2 - Streams Messaging Light Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-small-702;
+        CDP 1.2 - Discovery Data Hub: Apache Solr, Apache Spark, Hue=cdp-ddh-702;
```
risdenk, line 226: Not sure it makes sense to release this template back 7.0? This is a very old code line and are we sure that this works?

gezan:
Yes it's only aws for now. After we can make tests on azure as well we will hit that too. I think it shall work on 7.0 but if it's not required to support "older" version then I remove it. It's not necessary from our perspective.

oleewere:
@gezan I can help with this. also I would like to test/try the template (we can discuss that offline, as that is a bit offtopic)

gezan:
retest please

gezan:
retest please

doktoric:
opening https://github.com/hortonworks/cloudbreak/pull/8059

PR #7947, opened 2020-04-30
CB-6578 - Single resource group - persist RG

pdarvasi:


environment/src/main/resources/schema/app/20200428154502_CB-5653_single_resource_group.sql
```diff
@@ -0,0 +1,23 @@
+-- // CB-5653 single_resource_group
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE environment_parameters
+    ADD COLUMN IF NOT EXISTS resource_group_name varchar(255);
+
+ALTER TABLE environment_parameters
```
keyki, line 7: Shouldn't use default like this. See same comment here: https://github.com/hortonworks/cloudbreak/pull/7939#discussion_r417805981

PR #7948, opened 2020-04-30
CB-6791 Null status should not be set on DB Stack.

pkedvessy:
CB-6791 Null status should not be set on DB Stack.


pkedvessy:
retest this please

PR #7954, opened 2020-05-01
DISTX-432 Fix Zookeeper issue during DE HA master node repair.

cegganesh84:
1. DE HA templates started supporting DH master node repair.
2. Zookeeper expects a new folder with -2 suffix, but it does not exist because the auto-create setting is false.
3. Tested this manually in a cluster.


cegganesh84:
Closing this in favor of https://github.com/hortonworks/cloudbreak/pull/7955. It is the same issue, with a nicer fix.

PR #7956, opened 2020-05-01
CDPCP-2043. Remove groups that don't exist in control plane from IPA

handavid:
Groups that do not exist in the control plane at all can be safely removed
from IPA. The UmsUsersState has modified to include all WAGs so we can make
this determination.

Some protections were added to the FreeIPA client to prevent adding/removing
certain users that user sync should not touch. e.g., this prevents accidentally
removing the admin user or the admins group.

Added unit tests for new behavior
Tested manually by adding a group and running user sync

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
@@ -64,12 +64,16 @@ public class UmsUsersStateProvider {
             Map<String, FmsGroup> crnToFmsGroup = grpcUmsClient.listGroups(actorCrn, accountId, List.of(), requestIdOptional).stream()
                     .collect(Collectors.toMap(Group::getCrn, this::umsGroupToGroup));
 
-            List<WorkloadAdministrationGroup> wags = grpcUmsClient.listWorkloadAdministrationGroups(INTERNAL_ACTOR_CRN, accountId, requestIdOptional);
-
+            Map<WorkloadAdministrationGroup, FmsGroup> wags = grpcUmsClient.listWorkloadAdministrationGroups(INTERNAL_ACTOR_CRN, accountId, requestIdOptional)
+                    .stream()
+                    .collect(Collectors.toMap(wag -> wag, wag -> nameToGroup(wag.getWorkloadAdministrationGroupName())));
             environmentCrns.forEach(environmentCrn -> {
-                UmsUsersState.Builder umsUsersStateBuilder = new UmsUsersState.Builder();
+                UmsUsersState.Builder umsUsersStateBuilder = new UmsUsersState.Builder()
+                        .setWorkloadAdministrationGroups(wags.values());
+
                 UsersState.Builder usersStateBuilder = new UsersState.Builder();
                 Set<String> wagNamesForOtherEnvironments = new HashSet<>();
+
 
                 crnToFmsGroup.values().forEach(usersStateBuilder::addGroup);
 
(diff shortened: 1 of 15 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -64,23 +64,27 @@
...
 
+
                 crnToFmsGroup.values().forEach(usersStateBuilder::addGroup);
 
                 // Only add workload admin groups that belong to this environment.
                 // At the same time, build a set of workload admin groups that are
                 // associated with other environments so we can filter these out in
                 // the per-user group listing in handleUser.
-                wags.forEach(wag -> {
+                wags.keySet().forEach(wag -> {
```
lacikaaa, line 84: I think you could iterate the entryset, so you won't have to make a get() against the map

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FmsGroup.java
```diff
@@ -6,8 +6,9 @@
 
     private String name;
 
-    public void setName(String name) {
+    public FmsGroup setName(String name) {
```
lacikaaa, line 9: usually set methods don't return, I think builder pattern use the `with` naming convention

lacikaaa:
if I remember well there was an issue that ranger doesn't handle well if you delete a group which it has policy for. Has it changed or do I have the wrong memories?

handavid:
> if I remember well there was an issue that ranger doesn't handle well if you delete a group which it has policy for. Has it changed or do I have the wrong memories?

I'll check with the ranger team. I know you can't create ranger policies before groups exist. I also know that we do not want to delete/re-add the same group (which is why we didn't remove groups originally).

handavid:
> > if I remember well there was an issue that ranger doesn't handle well if you delete a group which it has policy for. Has it changed or do I have the wrong memories?
> 
> I'll check with the ranger team. I know you can't create ranger policies before groups exist. I also know that we do not want to delete/re-add the same group (which is why we didn't remove groups originally).

I think we are ok here based on sailaja's response. The policy will become stale but the group will still exist in Ranger.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -5,113 +5,187 @@
...
+        ImmutableMultimap<String, String> groupMembershipToAdd = calculateGroupMembershipToAdd(umsState, ipaState);
+        LOGGER.info("groupMembershipToAdd size = {}", groupMembershipToAdd.size());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("adding group memberships {}", groupMembershipToAdd.asMap());
+        }
+
+        ImmutableMultimap<String, String> groupMembershipToRemove = calculateGroupMembershipToRemove(umsState, ipaState);
+        LOGGER.info("groupMembershipToRemove size = {}", groupMembershipToRemove.size());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("removing group memberships {}", groupMembershipToRemove.asMap());
```
lacikaaa, line 117: overall things lgtm, but I think this log could be moved into the calculate methods.
also `isDebugEnabled` is overkill I think, as we have debug enabled everywhere, and if we are planning to change that then we will have to revisit our logging anyway

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -5,113 +5,187 @@
...
+        ImmutableSet<FmsUser> usersToAdd = calculateUsersToAdd(umsState, ipaState);
+        LOGGER.info("usersToAdd size = {}", usersToAdd.size());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("adding users {}", usersToAdd.stream().map(FmsUser::getName).collect(Collectors.toSet()));
+        }
+
+        ImmutableSet<String> usersToRemove = calculateUsersToRemove(umsState, ipaState);
+        LOGGER.info("usersToRemove size = {}", usersToRemove.size());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("removing users {}", usersToRemove.stream().collect(Collectors.toSet()));
```
lacikaaa, line 93: this stream is not necessary I think

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
handavid: done

PR #7957, opened 2020-05-03
CB-6879: Redundant full backup created on Azure FreeIPA HA hosts

wonderslug:
This prevents the monthly backup from running, when enabled, until the next month after the initial install.  This is the first step to fixing CB-6717 across multiple FreeIPA HA nodes.

Closes #CB-6879

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
@@ -15,6 +15,11 @@
     - mode: 640
 
 {% if salt['pillar.get']('freeipa:backup:monthly_full_enabled') %}
+freeipa_backoff_monthly_full:
+  cmd.run:
+    - name: /sbin/anacron -u cron.monthly
+    - onlyif: test ! -s /var/spool/anacron/cron.monthly
+
 /etc/cron.monthly/freeipa_backup_monthly:
   file.managed:
     - source: salt://freeipa/templates/freeipa_backup_monthly.j2


freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
```diff
@@ -15,6 +15,10 @@
     - mode: 640
 
 {% if salt['pillar.get']('freeipa:backup:monthly_full_enabled') %}
+'/sbin/anacron -u cron.monthly':
```
lacikaaa, line 18: could you rename this state and move the command into the `name` parameter?
also this would be called every time a highstate is called, so I think you should place a file somewhere so it wouldn't run again

wonderslug: It will not be run again because of the **onlyif**.  The anacron state files (in /var/spool/anacron) start out as a zero byte file until it is run one.  From then on they are updated to a datestamp of when they were last run.   The anacron -u forces the update of the datestamp file without running the commands.

lacikaaa: thanks for the explanation!

freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
```diff
@@ -15,6 +15,11 @@
     - mode: 640
 
 {% if salt['pillar.get']('freeipa:backup:monthly_full_enabled') %}
+freeipa_backoff_monthly_full:
+  cmd.run:
+    - name: /sbin/anacron -u cron.monthly
+    - onlyif: test ! -s /var/spool/anacron/cron.monthly
```
jamisonbennett, line 21: Is it possible that the anacron monthly backup runs once before the FreeIPA backup is added to it? If so, then this file would exist but the backup would not.

lacikaaa: initial backup is a different state. also `-u` is for skipping the actual run

wonderslug: no the anacron is not going to the cron.monthly until at least 45 minutes after the initial instance creation and then have an up to 45 minute random delay added onto that.  Unless the instance creation is taking longer than that to reach that point it will be fine.  We can force this to not run by disabling anacron completely during the salt process and explicitly turn it on at the end of the salt run.  We can do this by putting 0anacron in the /etc/cron.hourly/jobs.deny file until we want to explicitly remove it.

wonderslug: And actually on thinking about this.  In this case if the cron.monthly does get run prior to this that's fine, means it wont run the monthly till next month which is what we want anyways.

PR #7958, opened 2020-05-04
CB-6885 Use credential to read env cloudplatform when parsing parameters

gergopapi2:


gergopapi2:
Underlying code was reverted.

PR #7967, opened 2020-05-04
CB-5196 GA Streaming Templates

benyoka:
Updated the streaming templates to match GA requirements. Heavy duty template now runs 2 Schema Registry instances, uses external HA database and HDFS Schema Registry jar storage. It also has one less hosts. 

Testing: manually tested cluster installation on AWS and Azure.

Closes CB-5196


core/src/main/resources/application.yml
```diff
@@ -114,7 +114,7 @@ cb:
     url: http://localhost:8087
     contextPath: /redbeams
   runtimes:
-    latest: "7.1.1"
+    latest: "7.2.0"
```
doktoric, line 117: ?

benyoka:
retest

core/src/main/resources/application.yml
benyoka: Oops I left it there accidentally. (I only could select the 7.2.0 templates from the UI after this change)

PR #7969, opened 2020-05-05
CB-6897 Improve InstanceMetadata with instance lifecycle

Bajzathd:


core/src/main/resources/schema/app/20200504164000_CB-6897_improve_instance_metadata_with_instance_lifecycle.sql
```diff
@@ -0,0 +1,9 @@
+-- // CB-6897 Improve InstanceMetadata with instance lifecycle
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE instancemetadata ADD COLUMN IF NOT EXISTS lifecycle VARCHAR(255) DEFAULT 'NORMAL';
```
bergerdenes, line 4: I've read somewhere in a PR comment that this blocks the whole table for a (relatively) long time.
Ideal solution is not putting the default into the ADD COLUMN but as a second statement and then setting the default value to the null columns.
Example here: https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/schema/app/20200406095012_CB-6385_add_version_to_stack.sql

Similarly, change the SQL script in the freeipa module as well.

PR #7976, opened 2020-05-05
CB-6884 Cluster template is not found error and deadlock occurred at …

topolyai5:


core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
```diff
@@ -245,7 +250,7 @@ public boolean isDatalakeBlueprint(Blueprint blueprint) {
         return blueprintListFilters.isDatalakeBlueprint(blueprint);
     }
 
-    private Iterable<Blueprint> saveDefaultsWithReadRight(Iterable<Blueprint> blueprints, Workspace workspace) {
+    private synchronized Iterable<Blueprint> saveDefaultsWithReadRight(Iterable<Blueprint> blueprints, Workspace workspace) {
```
attilapalfi92, line 253: why synchronized is needed?

topolyai5: The issue in shortly: If the workspace does not have any default blueprint and the users want to access it (list it, create freeipa, e.g) in parallel, the threads try to create the default blueprints.
I tried with transactions, try-catch with a recursive call, but the deadlock occurs as well. 

The only working solution is the sync block for one instance and the recursive call for more instance -> all threads finish with success.

I added the sync block to the beginning of update, but I think it is a bottleneck because the read is synced as well, so wait until the other thread finishes (I cannot test it properly, so it is an assumption). 
If I add the sync block to save only, the default blueprint's update will slower but this is called only on empty ws or new/updated default bps

attilapalfi92: What happens when you put this into a transaction? That should solve the issue. Blocking every thread in a spring service that should be stateless is not a good idea. Not only for 1 workspace, but all the workspaces of all the customers are going to be blocked here. Not to mention the case when you have 1 workspace, 2 cloudbreak instances and the listing and freeipa creation operations go to different cloudbreak instances? What happens then?

This only solves the case when there is only 1 CB instance is present, so it won't work in production. Even if there were 1 CB instance, this would be a wrong solution because the threads are blocked for evert workspace in the world.

I understand it's not easy to solve it, but I think it's about time to initialise default blueprints for workspaces properly.

attilapalfi92: Issues regarding default blueprint initialisation come up time to time. A final solution is timely

PR #7979, opened 2020-05-06
Db template fix

doktoric:


keyki:
IT is failed, because of known BP issue.

PR #7988, opened 2020-05-06
CB-6920. Use proper error message in case of fluent version incompatibility (No network)

oleewere:
in case of fluent plugin version should be upgraded without internet connection (like old image is used for newer CB), it's better if the user will read the right error message instead of td-agent start failed.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
+++ b/orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
@@ -8,7 +8,10 @@ function install_plugin() {
 
   install_command="/opt/td-agent/embedded/bin/fluent-gem install ${plugin}"
   if [[ "${custom_repo}" == "true" ]]; then
+     check_connection "{{ fluent.clouderaPublicGemRepo }}" "${plugin}" "${version}"
      install_command="${install_command} -s  {{ fluent.clouderaPublicGemRepo }}"
+  else
+     check_connection "https://api.rubygems.org" "${plugin}" "${version}"
   fi
 
   if [[ ! -z "${version}" ]]; then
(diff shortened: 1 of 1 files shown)


oleewere:
```bash
----------
          ID: check_fluentd_plugins
    Function: cmd.run
        Name: sh /etc/td-agent/check_fluent_plugins.sh
      Result: False
     Comment: Command "sh /etc/td-agent/check_fluent_plugins.sh" run
     Started: 09:26:44.150320
    Duration: 2143.976 ms
     Changes:
              ----------
              pid:
                  16859
              retcode:
                  1
              stderr:
                  Connection has failed for https://repository.cloudera.com/cloudera/api/gems/cloudera-gems/. Please check your firewall or use a newer OS image (required fluent-plugin-databus gem version is 1.0.6) or reach out Cloudera (Cloudbreak) support!
              stdout:
                  Checking plugin fluent-plugin-redaction with the right minimal version 0.1.2
                  Run check command: /opt/td-agent/embedded/bin/fluent-gem list -i fluent-plugin-redaction -v '>=0.1.2'
                  Check fluent-plugin-redaction command output: true
                  Plugin fluent-plugin-redaction exists.
                  Run check command: /opt/td-agent/embedded/bin/fluent-gem list -i fluent-plugin-detect-exceptions
                  Check fluent-plugin-detect-exceptions command output: true
                  Plugin fluent-plugin-detect-exceptions exists.
                  Checking plugin fluent-plugin-databus with the right minimal version 1.0.6
                  Run check command: /opt/td-agent/embedded/bin/fluent-gem list -i fluent-plugin-databus -v '>=1.0.6'
                  Check fluent-plugin-databus command output: false
                  Plugin fluent-plugin-databus does not exist, installing it...
```
with this the error message on UI is only "Failed to execute sh /etc/td-agent/check_fluent_plugins.sh", if we would map stderr and show it on the UI, that could be problematic for other commands if it would contain any secrets (in commands)

PR #7999, opened 2020-05-07
CB-6969 CB should ignore the username and password fields when empty string is set

schfeca75:
@keyki or @lacikaaa pls review

### DIFF
--- a/environment-common/src/main/java/com/sequenceiq/cloudbreak/service/proxy/ProxyConfigDtoService.java
+++ b/environment-common/src/main/java/com/sequenceiq/cloudbreak/service/proxy/ProxyConfigDtoService.java
@@ -9,10 +9,10 @@ import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
-import org.springframework.util.StringUtils;
 import org.springframework.vault.VaultException;
 
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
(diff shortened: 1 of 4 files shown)


environment-common/src/main/java/com/sequenceiq/cloudbreak/service/proxy/ProxyConfigDtoService.java
```diff
@@ -74,9 +74,13 @@ private ProxyConfig convert(ProxyResponse proxyResponse) {
                 .withServerHost(proxyResponse.getHost())
                 .withServerPort(proxyResponse.getPort());
         if (proxyResponse.getUserName() != null && proxyResponse.getPassword() != null) {
-            proxyConfigBuilder.withProxyAuthentication(ProxyAuthentication.builder()
-                    .withUserName(getSecret(proxyResponse.getUserName()))
-                    .withPassword(getSecret(proxyResponse.getPassword())).build());
+            String user = getSecret(proxyResponse.getUserName());
```
keyki, line 77: What if getUserName returns null?

environment/src/main/java/com/sequenceiq/environment/proxy/v1/converter/ProxyRequestToProxyConfigConverter.java
```diff
@@ -13,12 +14,16 @@
...
         proxyConfig.setProtocol(source.getProtocol());
         proxyConfig.setServerHost(source.getHost());
         proxyConfig.setServerPort(source.getPort());
-        proxyConfig.setUserName(source.getUserName());
+        if (StringUtils.isNoneBlank(source.getUserName())) {
+            proxyConfig.setUserName(source.getUserName());
+        }
+        if (StringUtils.isNoneBlank(source.getPassword())) {
+            proxyConfig.setPassword(source.getPassword());
+        }
```
lacikaaa, line 26: is there any reason to set if the other is blank? if not then it should look like:
`if (StringUtils.isNoneBlank(source.getUserName(), source.getPassword())) {
            proxyConfig.setUserName(source.getUserName());
			proxyConfig.setPassword(source.getPassword());
        }`
if yes, then you could use StringUtils.isNotBlank

schfeca75: again....

PR #8004, opened 2020-05-08
CB-5908 - Fix datamart instance types on Azure

drorke:
Fix instance types in datamart template on Azure:

Master:  D16_v3 -> E8_v3
Impala coordinator and executors:  D8_v3 -> E16_v3

Use less expensive instance type for master (but same memory size)
Increase size of coordinator and executor instances to match sizes used by Datamart on AWS and DWX


doktoric:
@drorke  we just deleted 711 can you please rebase your PR? I will merge it after that

PR #8007, opened 2020-05-08
CB-6986 Disable p3 instancetype on AWS

lacikaaa:


doktoric:
I think we should only introduce this in the deployment charts. After the soltion ready then we just easily enable back the instance types there. what do you think ?

keyki:
I have a feeling that this instance type is never going to be enabled again. We could do in helm-charts also, but in that case, we will need to keep in mind which CB version is going to support it again to re-enable it.

PR #8010, opened 2020-05-08
CB-6972 Handle backward incompatible InstanceStatus values

pkedvessy:
CB-6972 Handle backward incompatible InstanceStatus values


pkedvessy:
retest this please

PR #8013, opened 2020-05-08
CB-6720: Update Ranger configs for CM 7.2.1+

kmanamcheri:
In CM 7.2.1 and above, ranger db configs moved from
role to service level. Update the cm template to
address for this.

Closes CB-6720


doktoric:
retest please

PR #8015, opened 2020-05-11
CB-6894 Spot instance e2e test cases

Bajzathd:


Bajzathd:
retest this please

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/spot/SpotTestResultProvider.java
```diff
@@ -0,0 +1,36 @@
+package com.sequenceiq.it.cloudbreak.testcase.e2e.spot;
+
+import org.junit.Assert;
```
bergerdenes, line 3: Should use testng's Assert

PR #8017, opened 2020-05-11
CB-6871 Eliminate groovy dependency from periscope

akanto:


smaniraju:
> Attila, the changes look good to me - remove groovy by removing Ambari references.
> 
> This will conflict with the AutoScale feature branch though. cc @smaniraju for another look.

This will require rebasing the LoadBased Scaling feature branch, but don't think there will be many conflicts. I can rebase once this PR is merged.

akanto:
Thanks

PR #8018, opened 2020-05-11
CDPCP-808 Sync SSH keys to IPA

handavid:
Ssh public keys are retrieved for each user as part of the UMS
GetActorWorkloadCredentials API. The user sync code to set workload
passwords in FreeIPA has been refactored to sync both the hashed
password and the public keys.

FreeIPA may return an 4202 error code during the user-mod call that can
be safely ignored. This error indicates that the mod list is empty and
may occur if the user does not have a hashed password set in the UMS and
does not have any ssh public keys. Exception handling in the user sync
code has been improved to also ignore similar exceptions for user and
group addition and removal as well as group membership changes.
Non-ignorable exceptions are now propagated to the sync operation status.
Environments with these exceptions are now put in the "failure" list
instead of the "success" list.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
@@ -6,6 +6,7 @@ import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -17,6 +18,7 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.reflect.TypeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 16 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -207,14 +208,21 @@ public void updateUserPasswordMaxExpiration(String user) throws FreeIpaClientExc
...
-                        "cdpHashedPassword=" + hashedPassword,
-                        "cdpUnencryptedKrbPrincipalKey=" + unencryptedKrbPrincipalKey,
-                        "krbPasswordExpiration=" + passwordExpirationDate));
+    public User userSetWorkloadCredentials(String user, String hashedPassword,
+            String unencryptedKrbPrincipalKey, Optional<Instant> expiration,
+            List<String> sshPublicKeys) throws FreeIpaClientException {
+        Map<String, Object> params = new HashMap<>();
+        List<String> attributes = new ArrayList<>();
+
+        if (null != hashedPassword && !hashedPassword.isBlank()) {
```
lacikaaa, line 217: apache StringUtils.isNotBlank(hashPassword) is a shorter

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -581,12 +553,12 @@ public Host showHost(String fqdn) throws FreeIpaClientException {
...
         } catch (Exception e) {
             String message = String.format("Invoke FreeIpa failed: %s", e.getLocalizedMessage());
-            LOGGER.error(message, e);
+            LOGGER.warn(message);
             OptionalInt responseCode = extractResponseCode(e);
             throw FreeIpaClientExceptionUtil.convertToRetryableIfNeeded(new FreeIpaClientException(message, e, responseCode));
         } catch (Throwable throwable) {
             String message = String.format("Invoke FreeIpa failed: %s", throwable.getLocalizedMessage());
-            LOGGER.error(message, throwable);
+            LOGGER.warn(message);
```
lacikaaa, line 561: why not to log and why not error?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -69,10 +59,17 @@ public void accept(Event<SetPasswordRequest> setPasswordRequestEvent) {
 
             FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientForStack(stack);
             if (FreeIpaCapabilities.hasSetPasswordHashSupport(freeIpaClient.getConfig())) {
-                LOGGER.info("IPA has password hash support, credentials information from UMS will be used.");
-                setPasswordHashFromUms(freeIpaClient, request.getUsername(), request.getUserCrn(), request.getExpirationInstant());
+                WorkloadCredential workloadCredential = umsCredentialProvider.getCredentials(request.getUserCrn(), MDCUtils.getRequestId());
+
+                LOGGER.info("IPA has password hash support. Credentials information from UMS will be used.");
+                workloadCredentialService.setWorkloadCredential(freeIpaClient, request.getUsername(), workloadCredential);
+                if (workloadCredential.getHashedPassword().isBlank()) {
```
lacikaaa, line 66: if you use StringUtils.isBlank here it will work even if getHashedPassword returns null

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: thanks!

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
handavid: the throwable is set as the cause of the FreeIpaClientException and then rethrown. I feel that the caller should be responsible for logging/handling the exception. Seeing an error stacktrace in the logs at this point is misleading.

handavid:
> can we move this to master?

done

PR #8023, opened 2020-05-12
CB-7013 Delete old 7.1.1 templates / definitions

doktoric:


core/src/main/resources/schema/app/20200512100924_CB-7013_Delete_old_7.1.1_templates.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-7013 Delete old 7.1.1 templates
+-- Migration SQL that makes the change goes here.
+
+UPDATE blueprint SET status = 'DEFAULT_DELETED' WHERE status = 'DEFAULT' AND name LIKE '7.1.1%';
```
akanto, line 4: I think we don't need this anymore, the code shall hide things automatically.

doktoric: I removed the line and also fixed an issue with the loader service what I just realized. Different commit and different issue but if you want I can open two separated pull request for this.

akanto: Thanks, but I think we don't need SQL at all

doktoric: removed

core/src/main/resources/application.yml
```diff
@@ -234,15 +234,6 @@ cb:
         7.1.0 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha-710;
         7.1.0 - Streams Messaging Heavy Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-710;
         7.1.0 - Streams Messaging Light Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-small-710;
-        7.1.1 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie=cdp-data-engineering-711;
```
akanto, line 237: clouderamanager section still contains reference for 7.1.1

PR #8024, opened 2020-05-12
Fix e2e ssh assert for email like usernames

lnardai:
- this only fixes the username issue when doing an ssh login with usernames that contain @company.com endings


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/hybrid/AwsYcloudHybridCloudTest.java
```diff
@@ -161,11 +162,13 @@ public void testCreateSdxOnChildEnvironment(TestContext testContext) {
     }
 
     private void testShhAuthenticationSuccessfull(String username, String host) throws IOException, UserAuthException {
+        String sanitizedUserName = SanitizerUtil.sanitizeWorkloadUsername(username);
         SSHClient client = getSshClient(host);
-        client.authPassword(username, MOCK_UMS_PASSWORD);
+        client.authPassword(sanitizedUserName, MOCK_UMS_PASSWORD);
```
pkedvessy, line 167: Please use sanitizedUserName it the testShhAuthenticationFailure method as well.

PR #8027, opened 2020-05-12
CB-6914 Salt error in termination phase

bbihari:


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -782,20 +783,38 @@ public void stopClusterManagerAgent(GatewayConfig gatewayConfig, Set<Node> nodes
...
     }
 
+    private boolean isChangingRolesNecessary(GatewayConfig gatewayConfig, SaltConnector sc, String role) {
+        return getMemberRoles(gatewayConfig, sc).contains(role);
+    }
+
+    private Set<String> getMemberRoles(GatewayConfig gatewayConfig, SaltConnector sc) {
+        Map<String, JsonNode> roles = SaltStates.getGrains(sc, new HostList(List.of(gatewayConfig.getHostname())), "roles");
+        return roles.entrySet().stream().findFirst()
+                .map(Entry::getValue)
```
lacikaaa, line 809: why do you stream the entryset if you need only the values?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -782,20 +783,38 @@ public void stopClusterManagerAgent(GatewayConfig gatewayConfig, Set<Node> nodes
...
+                .map(JsonNode::elements)
+                .map(IteratorUtils::toList)
+                .map(l -> (Set<String>) l.stream()
+                        .map(i -> ((JsonNode) i).asText())
+                        .collect(Collectors.toSet())
```
lacikaaa, line 814: what happens here? what is `l` and `i`?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -782,20 +783,38 @@ public void stopClusterManagerAgent(GatewayConfig gatewayConfig, Set<Node> nodes
...
+                saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(allHostnames, allNodes), exitCriteriaModel, maxRetryLeave, exitCriteria);
+                runNewService(sc, new HighStateRunner(allHostnames, allNodes), exitCriteriaModel, maxRetryLeave, true);
+            }
         } catch (Exception e) {
             LOGGER.info("Error occurred during executing highstate (for recipes).", e);
             throw new CloudbreakOrchestratorFailedException(e);
         }
     }
 
+    private boolean isChangingRolesNecessary(GatewayConfig gatewayConfig, SaltConnector sc, String role) {
```
lacikaaa, line 802: it could be useful somewhere else, should we extract it? also some tests for this would be nice

bbihari: As long as we don't use this function at least one other place, I wouldn't like to make it public or extract it

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
bbihari: ok

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
bbihari: I've 'fixed' it. Personally I like to use short variables in lambda. I think in this place it wasn't confusing.

PR #8028, opened 2020-05-12
CB-7009 Azure single RG: env service create RG

gergopapi2:


environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+        String resourceGroupName = String.format("%s_%d", environmentDto.getName(), clock.getCurrentTimeMillis());
+        LOGGER.debug("Azure generated resource group name: {}.", resourceGroupName);
+        azureClient.createResourceGroup(resourceGroupName, environmentDto.getLocation().getName(), mergeTags(environmentDto));
+        parameterService.updateResourceGroupName(environment, resourceGroupName);
+        LOGGER.debug("Azure resource group created successfully.");
+    }
+
+    private Map<String, String> mergeTags(EnvironmentDto environmentDto) {
+        CDPTagMergeRequest mergeRequest = CDPTagMergeRequest.Builder
```
doktoric, line 124: maybe ewe should just put this into the DTO directly, in that case we can just use that object

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            return;
+        }
+        Optional<AzureResourceGroupDto> azureResourceGroupDtoOptional = getAzureResourceGroupDto(environmentDto);
+        if (!AZURE.name().equals(environmentDto.getCloudPlatform()) || azureResourceGroupDtoOptional.isEmpty()) {
+            LOGGER.debug("Cloudplatform not azure or no azure resource group dto defined, not creating resource group.");
+            return;
+        }
+        AzureResourceGroupDto azureResourceGroupDto = azureResourceGroupDtoOptional.get();
+        if (!CREATE_NEW.equals(azureResourceGroupDto.getResourceGroupCreation())) {
+            LOGGER.debug("Azure resource group creation not requested, not creating resource group.");
```
doktoric, line 110: Azure resource group creation not requested -> New single Azure resource group creation not requested

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            return;
+        }
+        CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environmentDto.getCredential());
+        AzureClient azureClient = azureClientService.getClient(cloudCredential);
+
+        String resourceGroupName = String.format("%s_%d", environmentDto.getName(), clock.getCurrentTimeMillis());
+        LOGGER.debug("Azure generated resource group name: {}.", resourceGroupName);
+        azureClient.createResourceGroup(resourceGroupName, environmentDto.getLocation().getName(), mergeTags(environmentDto));
+        parameterService.updateResourceGroupName(environment, resourceGroupName);
+        LOGGER.debug("Azure resource group created successfully.");
```
doktoric, line 120: do we want to catch the exception here ?

doktoric:
how can we handle the new status in case of a production rollback? Can you pleasse implement an enum converter like FeatureStateConverter.java into 2.21?

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
gergopapi2: Yes, I was wondering myself why is the combination of different tags done outside of the dto.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
gergopapi2: What exception do you mean? There is a try/catch around the createResourceGroup method, and the exception message is logged there.

gergopapi2:
> how can we handle the new status in case of a production rollback? Can you pleasse implement an enum converter like FeatureStateConverter.java into 2.21?
Great idea. 
However, do you mean as a separate PR before this one? Even then it will most probably have an effect for later flow changes only. This EnumConverter should be already rolled out before the present feature of RG create makes it to an env.

environment/src/main/java/com/sequenceiq/environment/parameters/service/ParametersService.java
```diff
@@ -62,6 +63,18 @@ public boolean isS3GuardTableUsed(String accountId, String cloudPlatform, String
...
     }
 
+    public void updateResourceGroupName(Environment environment, String resourceGroupName) {
+        Optional<BaseParameters> baseParametersOptional = baseParametersRepository.findByEnvironmentId(environment.getId());
+        if (baseParametersOptional.isEmpty()) {
+            return;
+        }
+
+        BaseParameters baseParameters = baseParametersOptional.get();
+        AzureParameters azureParameters = (AzureParameters) baseParameters;
```
pdarvasi, line 73: We should do this only for Azure cloud platform, pls use getCloudPlatform

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
```diff
@@ -0,0 +1,166 @@
...
+import reactor.bus.EventBus;
+
+@Component
+public class ResourceGroupCreationHandler extends EventSenderAwareHandler<EnvironmentDto> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ResourceGroupCreationHandler.class);
+
+    private CloudPlatformConnectors cloudPlatformConnectors;
+
+    private AzureClientService azureClientService;
```
attilapalfi92, line 47: I don't like the idea of adding direct Azure dependency (or any specific cloud provider dependency) in full generic Env creation flow steps.

attilapalfi92: Can you name this step instead InfrastructureCreationHandler and move every cloud provider specific logic behind the cloud-api? And use those APIs here?

gergopapi2: I will do it as I see that also in the case of S3Guard this approach was taken and accept that the team likes it this way. However, I would like just note that this is somewhat against polymorphism, as we try to pretend something to be generic when in reality it isn't (resourceGroup is a pure azure concept). I would have factored it behind an interface if at least one another implementation (AWS or MOCK) to that flow step would be needed.

attilapalfi92: If you name the step Infrastructure Creation it immediately becomes generic. It's a different question that for AWS this time there's nothing to do with it. But you know what features and requirements will there be in 3 months? 6 months? What if the need emerges to do some additional work on AWS during Env creation? We add a new AWS specific step? It would bloat the flow unnecessarily even more. Flows meant to be generic, not just in the Env service, but everywhere in CB, and cloud provider specific jobs should be hidden behind the cloud-api. Otherwise this whole software wouldn't work. This is a good legacy we have and we should guard it.

gergopapi2: Yes. One thought regarding the name: the whole cloudbreak is about infrastructure creation :), so the name is not really descriptive. And it is exactly this point, coming up with a good name that shows me that too much generics is premature now. 
Coming to the point what requirements will come in 3 months: well, I would be rather doing a refactor when we see what's needed (i.e. in 3 or 6 months time) then do a very generic solution with a suboptimal name that will need to be renamed anyway.
I agree that the flow is generic.

gergopapi2: I looked up S3GuardTable deletion. There the flow step is called that way and not masqueraded to be something generic, with good purpose: it is something that exists in the AWS world. Once there will be an azure parallel I am sure a heavy refactoring will come. So I will keep the name.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
pdarvasi: We will see, hopefully Azure exception will be meaningful enough to be presented in UI.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceGroupCreationHandler.java
pdarvasi: I have not hard bias pro or contra, but I am merging the PR to make progress with the feature.  
@attilapalfi92 pls open a follow-up if you'd like to change this.

PR #8030, opened 2020-05-12
CDPCP-1749. Remove usages of deprecated user sync status fields

handavid:


gitmfox:
I'm not familiar with the codebase, but the change LGTM.

PR #8032, opened 2020-05-13
CB-6967 Use DatabaseStack as resource connector parameter instead of DBInstanceIdentifier and implement Azure status lookup

pkedvessy:
CB-6967 Use DatabaseStack as resource connector parameter instead of DBInstanceIdentifier and implement Azure status lookup

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/connector/resource/AzureDatabaseResourceService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/connector/resource/AzureDatabaseResourceService.java
@@ -6,6 +6,8 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
+import com.microsoft.azure.management.resources.ResourceGroup;
+import com.sequenceiq.cloudbreak.cloud.model.ExternalDatabaseStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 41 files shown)


cloud-mock/src/main/java/com/sequenceiq/cloudbreak/cloud/mock/MockResourceConnector.java
```diff
@@ -193,17 +193,17 @@ public String getDBStackTemplate() throws TemplatingDoesNotSupportedException {
...
     @Override
-    public void stopDatabaseServer(AuthenticatedContext authenticatedContext, String dbInstanceIdentifier) {
+    public void stopDatabaseServer(AuthenticatedContext authenticatedContext, DatabaseStack stack) {
         throw new UnsupportedOperationException("Database server stop operation is not supported for " + getClass().getName());
     }
 
     @Override
-    public ExternalDatabaseStatus getDatabaseServerStatus(AuthenticatedContext authenticatedContext, String dbInstanceIdentifier) throws Exception {
+    public ExternalDatabaseStatus getDatabaseServerStatus(AuthenticatedContext authenticatedContext, DatabaseStack stack) throws Exception {
         throw new UnsupportedOperationException("Database server status lookup is not supported for " + getClass().getName());
```
gergopapi2, line 207: Question: during a mock test is the status sync running? If yes, it might then break the test.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/connector/resource/AzureDatabaseResourceService.java
```diff
@@ -141,5 +143,22 @@
...
+        CloudContext cloudContext = ac.getCloudContext();
+        AzureClient client = ac.getParameter(AzureClient.class);
+        String resourceGroupName = azureUtils.getResourceGroupName(cloudContext, stack);
+
+        try {
+            ResourceGroup resourceGroup = client.getResourceGroup(resourceGroupName);
+            if (resourceGroup == null) {
+                return ExternalDatabaseStatus.DELETED;
+            }
+            return ExternalDatabaseStatus.STARTED;
```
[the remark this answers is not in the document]
pkedvessy: Because other ExternalDatabaseStatuses are irrelevant on Azure at this point. Azure does not support simple start/stop functionality like AWS.

cloud-mock/src/main/java/com/sequenceiq/cloudbreak/cloud/mock/MockResourceConnector.java
pkedvessy: Do you mean the exception might break the test? I don't think it is the case as I'm not aware of any breaking test in connection with this exception. It got introduced more than a month ago.

PR #8033, opened 2020-05-13
CB-6751 WaitUtil and WaitForMultipleStatuses overall refactor part 2

aszegedi:
This is the second part of a sequent.

The goal is to retire the old API Test Project Wait Util and introduce the new Wait Service.


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseAwait.java
```diff
@@ -0,0 +1,72 @@
...
+                case DELETE_COMPLETED:
+                    waitForDatabaseStatus(new DatabaseTerminationChecker<>(), client, crn, testContext, desiredStatus,
+                            pollingInterval, maxRetry);
+                    break;
+                default:
+                    LOGGER.warn("Wait checker is not implemented yet for the desired database state '{}' ", desiredStatus);
+                    break;
+            }
+            if (!desiredStatus.equals(DELETE_COMPLETED)) {
+                entity.refresh(testContext, null);
```
lnardai, line 52: Why are we refreshing with null?

aszegedi: Here is the `CloudbreakClient` should be passed.
```awaitEntity.refresh(getTestContext(), (CloudbreakClient) null);}```
However that is not defined in `RedbeamsDatabaseServerTestDto await`. So we can remove this from `Redbeams Await`, because of Redbeams is
> "not refreshable resource" indeed

File: integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/cloudbreak/CloudbreakWaitObject.java [hunk 2]
```diff
@@ -0,0 +1,64 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public Map<String, Status> getDesiredStatuses() {
+        return desiredStatuses;
+    }
+
+    public boolean isFailed(Map<String, Status> distroxStatuses) {
```
foldik, line 57: I think the below alternative is more efficient.
```java
        Set<Status> failedStatuses = Set.of(UPDATE_FAILED, CREATE_FAILED, ENABLE_SECURITY_FAILED, DELETE_FAILED, START_FAILED, STOP_FAILED,
                EXTERNAL_DATABASE_CREATION_FAILED, EXTERNAL_DATABASE_DELETION_FAILED);
        return !Sets.intersection(failedStatuses, Set.copyOf(distroxStatuses.values())).isEmpty();
```

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseOperationChecker.java
```diff
@@ -0,0 +1,135 @@
...
+            if (desiredStatus.equals(status)) {
+                return true;
+            }
+        } catch (Exception e) {
+            StringBuilder builder = new StringBuilder("Database has been failed. Also failed to get database status: ")
+                    .append(System.lineSeparator())
+                    .append(e.getMessage())
+                    .append(System.lineSeparator())
+                    .append(e);
+            LOGGER.error(builder.toString());
```
foldik, line 58: Why call it twice? I guess the compiler will optimize it anyway but still strange a bit. I saw it earlier as well in other files. Can you please introduce a `message` variable?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseOperationChecker.java
```diff
@@ -0,0 +1,135 @@
...
+            Status status = database.getStatus();
+            throw new TestFailException(String.format("Wait operation timed out, database '%s' '%s' has been failed. Database status: '%s' "
+                    + "statusReason: '%s'", name, crn, status, database.getStatusReason()));
+        } catch (Exception e) {
+            StringBuilder builder = new StringBuilder("Wait operation timed out, database has been failed. Also failed to get database status: ")
+                    .append(System.lineSeparator())
+                    .append(e.getMessage())
+                    .append(System.lineSeparator())
+                    .append(e);
+            LOGGER.error(builder.toString());
```
foldik, line 82: For pure logging please use the `LOGGER.error("message {} {}", arg1,, arg2)` format. I saw it other places as well. Please avoid manual string concatenation when possible.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseWaitObject.java
```diff
@@ -0,0 +1,47 @@
...
+    public String getCrn() {
+        return crn;
+    }
+
+    public Status getDesiredStatus() {
+        return desiredStatus;
+    }
+
+    public boolean isFailed(Status databaseStatus) {
+        Collection<Status> failedStatuses = List.of(UPDATE_FAILED, CREATE_FAILED, ENABLE_SECURITY_FAILED, DELETE_FAILED, START_FAILED, STOP_FAILED);
```
foldik, line 44: Please use a Set instead.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/datalake/DatalakeWaitObject.java
```diff
@@ -0,0 +1,50 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public SdxClusterStatusResponse getDesiredStatus() {
+        return desiredStatus;
+    }
+
+    public boolean isFailed(SdxClusterStatusResponse datalakeStatus) {
+        Collection<SdxClusterStatusResponse> failedStatuses = List.of(PROVISIONING_FAILED, REPAIR_FAILED, UPGRADE_FAILED, DATALAKE_UPGRADE_FAILED,
```
foldik, line 46: Please use a Set.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/environment/EnvironmentOperationChecker.java
```diff
@@ -112,24 +113,29 @@ public boolean exitWaiting(T waitObject) {
...
         return false;
     }
 
+    @Override
+    public Map<String, String> getStatuses(T waitObject) {
+        return Map.of("status", waitObject.getEndpoint().getByCrn(waitObject.getCrn()).getEnvironmentStatus().name());
+    }
+
     private boolean isDeletionInProgress(EnvironmentStatus environmentStatus) {
         Collection<EnvironmentStatus> deleteInProgressStatuses = List.of(DELETE_INITIATED, NETWORK_DELETE_IN_PROGRESS, RDBMS_DELETE_IN_PROGRESS,
```
foldik, line 140: Please use a set.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseAwait.java
```diff
@@ -0,0 +1,67 @@
...
+import com.sequenceiq.it.cloudbreak.log.Log;
+import com.sequenceiq.it.cloudbreak.util.ResponseUtil;
+import com.sequenceiq.it.cloudbreak.util.wait.service.Await;
+import com.sequenceiq.it.cloudbreak.util.wait.service.ExceptionChecker;
+import com.sequenceiq.it.cloudbreak.util.wait.service.Result;
+import com.sequenceiq.it.cloudbreak.util.wait.service.WaitResult;
+import com.sequenceiq.redbeams.api.model.common.Status;
+
+@Component
+public class DatabaseAwait extends Await<RedbeamsDatabaseServerTestDto, Status> {
```
gergopapi2, line 20: I would rename Database* to Redbeams*, to avoid confusion with the existing Database endpoints on cloudbreak.
Or, if you want to keep the database name, then I would rename it to DatabaseServer*, to avoid confusion with Database endpoints on redbeams.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/database/DatabaseAwait.java
```diff
@@ -0,0 +1,67 @@
...
+                case STOPPED:
+                    waitForDatabaseStatus(new DatabaseOperationChecker<>(), client, crn, testContext, desiredStatus,
+                            pollingInterval, maxRetry);
+                    break;
+                case DELETE_COMPLETED:
+                    waitForDatabaseStatus(new DatabaseTerminationChecker<>(), client, crn, testContext, desiredStatus,
+                            pollingInterval, maxRetry);
+                    break;
+                default:
+                    LOGGER.warn("Wait checker is not implemented yet for the desired database state '{}' ", desiredStatus);
```
gergopapi2, line 46: I put this question here but relates to all Await implementations: if waiting for a given status is not implemented but a test developer wants to wait for it in his test, how will he be notified?
@aszegedi, you suggested in personal communications, that switch-case could be converted to plain if-else, distinguishing between delete and non-delete operations, could make sense.

PR #8034, opened 2020-05-13
CB-4785 Better error message display for environment+sdx accidental A…

topolyai5:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
```diff
@@ -175,7 +176,7 @@ public void startStackProvisioning(Long id, DetailedEnvironmentResponse environm
                 sdxClusterRepository.save(sdxCluster);
                 cloudbreakFlowService.saveLastCloudbreakFlowChainId(sdxCluster, stackV4Response.getFlowIdentifier());
                 LOGGER.info("Sdx cluster updated");
-            } catch (ClientErrorException e) {
+            } catch (ClientErrorException | InternalServerErrorException e) {
```
sodre90, line 179: hmm.. can we throw badrequest instead of internal server error from cloudbreak side?

topolyai5: We do not throw internal exception directly, It is a common jax-rs exception of HTTP code 500. But we need to handle it because the reason message was `500 internal error` on the UI

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
```diff
@@ -175,7 +176,7 @@ public void startStackProvisioning(Long id, DetailedEnvironmentResponse environm
...
                 String errorMessage = webApplicationExceptionMessageExtractor.getErrorMessage(e);
                 LOGGER.info("Can not start provisioning: {}", errorMessage, e);
                 throw new RuntimeException("Can not start provisioning, client error happened on Cloudbreak side: " + errorMessage);
```
daszabo, line 182: The exception message is misleading

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
```diff
@@ -636,12 +636,15 @@ private CloudVmTypes getCloudVmTypes(CloudCredential cloudCredential, Region reg
...
+        List<AvailabilityZone> availabilityZones = regions.getCloudRegions().get(region);
+        if (availabilityZones != null && !availabilityZones.isEmpty()) {
+            for (AvailabilityZone availabilityZone : availabilityZones) {
+                Set<VmType> types = vmTypes.get(region).stream()
+                        .filter(enabledInstanceTypeFilter)
+                        .collect(Collectors.toSet());
+                cloudVmResponses.put(availabilityZone.value(), types);
+                defaultCloudVmResponses.put(availabilityZone.value(), defaultVmTypes.get(region));
+            }
         }
```
daszabo, line 648: Should we throw an exception, if the AZ list is empty? If not, at least log the fact please.

topolyai5: Are the AZs required? Are you sure that we do not break the process with exception?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
topolyai5: is it better? `"Can not start provisioning, some error happened on Cloudbreak side: " + errorMessage`

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
sodre90: but why is it 500? why not something else? maybe a bad exceptionmapper?

topolyai5: Imagine that user error when someone would like to attach an AZURE SDX to an AWS env.
Somewhere in the CB throw a ClassCastException. Should we create a mapper all of the unexpected wrong user interaction? 
Sorry, my fault: in the `DefaultExceptionMapper` we throw an InternalServerError

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
attilapalfi92: Cannot*

topolyai5: https://www.lexico.com/grammar/cannot-or-can-not
someone prefer that format, what is the problem with that?

attilapalfi92: "the first is much more usual". Users will see this and they would also think can not looks weird and unusual

topolyai5: ok

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsPlatformResources.java
daszabo: Well, without an exception, the dropdown on the UI will be empty, which might also be a good enough clue for the user, that something is amiss.

topolyai5: the log message is ok for me, I will update with it

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
sodre90: please use WebApplicationException (it will catch both client and server errors)

PR #8035, opened 2020-05-13
CB-6973 - moving aws getRegions call to flow step and additional logging

attilapalfi92:


attilapalfi92:
@akanto 
Where is the inefficient part of the validation? 

I see 2 AWS API calls:
https://github.com/hortonworks/cloudbreak/pull/8035/files#diff-7bdd8e6c134e2cf0ae7b404ad526c5fdR56

https://github.com/hortonworks/cloudbreak/pull/8035/files#diff-b8f5d95a351670d2b7bc3332c743bce6R74

1-1 in the init and validation handler.

PR #8044, opened 2020-05-14
CB-7053. Disable FS trash for Data Engineering templates.

sidseth:
Do I need to generate PRs against other branches as welL?


amitagarwal06:
> Do I need to generate PRs against other branches as welL?

I think you need to generate it against rc-2.21 too. Or rather only against rc-2.21 and Krisztian will merge it to master too.

keyki:
> > Do I need to generate PRs against other branches as welL?
> 
> I think you need to generate it against rc-2.21 too. Or rather only against rc-2.21 and Krisztian will merge it to master too.

Right. We only need 1 PR to the target branch and we will automatically merge upstream. So if this is something that we need in 2.21 then PR should go to rc-2.21.

sidseth:
Closing this. Will create another one for rc.21

PR #8047, opened 2020-05-14
CB-7058 Cluster is not in proper state

sodre90:


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
```diff
@@ -224,7 +224,8 @@ public void waitCloudbreakClusterCreation(Long id, PollingConfig pollingConfig)
                                     if (FINISHED.equals(flowState)) {
                                         LOGGER.warn("Stack is in improper state! FlowState: {}, stackV4Response: {}",
                                                 flowState, JsonUtil.writeValueAsStringSilent(stackV4Response));
-                                        return sdxCreationFailed(sdxCluster, "cluster is not in proper state");
+                                        return sdxCreationFailed(sdxCluster, "cluster is not in proper state (" + stackV4Response.getStatus() +
```
akanto, line 227: I would gladly eliminate this part of the message: "cluster is not in proper state"

doktoric: also it would be nice to use the right term so because we are in the sdx app use Data Lake (this is what we shows to customers)

sodre90: Error message is improved, please check if you see any more misleading messages

PR #8051, opened 2020-05-14
CB-7065 Backward compatibility for Environment flow enums

gergopapi2:


common/src/main/java/com/sequenceiq/cloudbreak/converter/DefaultEnumConverter.java
```diff
@@ -24,9 +26,12 @@ public E convertToEntityAttribute(String attribute) {
...
             LOGGER.info("The {} value is not backward compatible: {}", getDefault().getClass().getSimpleName(), attribute);
         }
-        return getDefault();
+        Optional<E> convertedEnum = tryConvert(attribute);
+        return convertedEnum.orElse(getDefault());
     }
 
     public abstract E getDefault();
 
+    public abstract Optional<E> tryConvert(String attribute);
```
doktoric, line 35: why we dont add a default implementation here?

gergopapi2: Ehm, great idea.

PR #8053, opened 2020-05-14
Cb 7006 Use resource Owner assignment feature in UMS

foldik:
__Note:__
Should be merged after https://github.com/hortonworks/cloudbreak/pull/8045

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -584,6 +584,21 @@ public class GrpcUmsClient {
         }
     }
 
+    public void assignResourceOwnerRoleIfEntitled(String userCrn, String resourceCrn, String accountId) {
+        try {
+            if (isAuthorizationEntitlementRegistered(userCrn, accountId)) {
+                assignResourceRole(userCrn, resourceCrn, getBuiltInOwnerResourceRoleCrn(), MDCUtils.getRequestId());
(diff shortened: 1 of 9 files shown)


environment/src/main/java/com/sequenceiq/environment/credential/service/CredentialService.java
```diff
@@ -176,6 +176,7 @@ public Credential create(Credential credential, @Nonnull String accountId, @Nonn
         credential.setResourceCrn(credentialCrn);
         credential.setCreator(creatorUserCrn);
         credential.setAccountId(accountId);
+        grpcUmsClient.assignResourceOwnerRoleIfEntitled(creatorUserCrn, credentialCrn, accountId);
```
horadla23, line 179: please put this into the transaction block

foldik: okay

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -584,6 +584,21 @@ public void assignResourceRole(String userCrn, String resourceCrn, String resour
         }
     }
 
+    public void assignResourceOwnerRoleIfEntitled(String userCrn, String resourceCrn, String accountId) {
+        try {
+            if (isAuthorizationEntitlementRegistered(userCrn, accountId)) {
+                assignResourceRole(userCrn, resourceCrn, getBuiltInOwnerResourceRoleCrn(), MDCUtils.getRequestId());
```
horadla23, line 590: maybe we should use `INTERNAL_ACTOR_CRN` instead of `userCrn` as actor, because it is possible that user does not have the right to assign a role

PR #8055, opened 2020-05-14
CB-7063 tests could be skipped if developer find it flaky

afarsang:
when tests are could not be fixed, instead of removing or skipping the whole test ,developer can use skipWhenFail() or validate() at the end of the test code. skipWhenFail() let the test run as usual do all the reporting, but skip if exceptions were collected during the run.
If the test or functionality fixed skipWhenFail() easily may swap with validate().


doktoric:
retest please

PR #8056, opened 2020-05-14
CB-7048: Change FreeIPA ASDL backup location from older .blob. to .dfs

wonderslug:
This changes the backup location for FreeIPA backups to the older blob.core.windows.net to the same location that is used by logs (dfs.core.windows.net)  so we have the same location to be used with things like Private Link.

Closes #CB-7048


doktoric:
I guess the target branch should be 2.21

wonderslug:
OK let me just close this one and open another on it.

PR #8061, opened 2020-05-14
CB-7053. Disable FS trash for Data Engineering templates.

sidseth:


sidseth:
Thanks for the review.

PR #8065, opened 2020-05-15
CB-6473 Add postgres empty password to int test Profile

horadla23:


horadla23:
will fix this: http://build.eng.hortonworks.com:8080/job/integration-test/2332/console

PR #8069, opened 2020-05-15
CB-7078 - fix missing postgres password

gregito:
CB-7078 - fix missing postgres password


horadla23:
:) https://github.com/hortonworks/cloudbreak/pull/8065

PR #8071, opened 2020-05-15
CB-7031 removing workaround in hybrid usecase

doktoric:


lnardai:
https://gph.is/2kzhtRQ

PR #8076, opened 2020-05-15
CB-7083 - Availability zone fetching optimisations

attilapalfi92:


environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentService.java
```diff
@@ -157,7 +157,7 @@ public CloudRegions getRegionsByEnvironment(Environment environment) {
         PlatformResourceRequest platformResourceRequest = new PlatformResourceRequest();
         platformResourceRequest.setCredential(environment.getCredential());
         platformResourceRequest.setCloudPlatform(environment.getCloudPlatform());
-        return platformParameterService.getRegionsByCredential(platformResourceRequest);
+        return platformParameterService.getRegionsByCredential(platformResourceRequest, false);
```
daszabo, line 160: This seems to be the single place where false is passed as value. Since this new parameter never comes from user input or something dynamic, I would argue to introduce a a different method for this.

attilapalfi92: I can create a getRegionsWithoutAZs method in the CloudParameterService. But in the SPI I would keep a the current implementation. Is that OK for you?

daszabo: Well, the spi part would be the whole point. I am not adamant in making the change, it works as it is now, its just that it is not too pretty.

attilapalfi92: That would not be prettier. The boolean value travels in the GetPlatformRegionsRequestV2 object and in the GetRegionsV2Handler there would be an if based on the boolean value to call the `regions` and `regionsWithoutAZs`. It wouldn't be nice at all in my opinion. This was what I first wanted, but when I was there, I saw that passing the boolean from the GetPlatformRegionsRequestV2 is a nicer solution.

attilapalfi92: Or you meant completely new spi regions handler too, I think that would be an overkill.

PR #8081, opened 2020-05-17
CB-7096: Add downscaling FreeIPA for recovery of non-CRL master nodes

jamisonbennett:
When an non-CRL master fails, the recovery should perform a downscale.
The process includes:
  * Updating cluster proxy registration so that it does not include
    the node
  * Attempting to gracefully stop telemetry on the node
  * Terminating the cloud provider instance
  * Removing the server from FreeIPA replication agreements and hosts
  * Revoking the host's certificate
  * Removing the DNS entries including A records, PTR records, SRV
    records, and PTR records
  * Remove the instance from salt orchestration including both salt
    masters and salt minions
  * Update the metadata for the instance and stack

This will not clean up the /etc/hosts file which will still include
references to downscaled freeipa instances. If it were to clean up the
/etc/hosts file using the hosts.absent salt rule then the instance
still would need a list of all the failed hosts on the node. It
doesn't seem particularly problematic to keep the old hosts in the
file.

This was tested with the added tests and it was also tested manually
by downscaling a 2 node FreeIPA cluster to 1 node. The primary gateway
was the CA master & CRL master and that node was kept.

Closes #CB-7096


jamisonbennett:
@oleewere  I know this is a large PR. I wanted you to take a look at the stop telemetry section since it looks like you had been the one to make most of those changes originally.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/action/FreeIpaDownscaleActions.java
```diff
@@ -0,0 +1,331 @@
...
+    @Bean(name = "DOWNSCALE_CLUSTERPROXY_REGISTRATION_STATE")
+    public Action<?, ?> downscaleClusterProxyRegistrationAction() {
+        return new AbstractDownscaleAction<>(StackEvent.class) {
+            @Override
+            protected void doExecute(FreeIpaDownscaleContext context, StackEvent payload, Map<Object, Object> variables) throws Exception {
+                Stack stack = context.getStack();
+                stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.DOWNSCALE_IN_PROGRESS, "Updating cluster proxy registration.");
+                List<String> instanceIdsToRegister = stack.getNotDeletedInstanceMetaDataList().stream()
+                        .map(InstanceMetaData::getInstanceId)
+                        .filter(instanceId -> !context.getInstanceIds().contains(instanceId))
```
lacikaaa, line 115: like here, the `AbstractDownscaleAction` could have a method for getting the instanceIds from the `variables`, and not put it in the context, because now we have the same info available in 2 places

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/action/FreeIpaDownscaleActions.java
```diff
@@ -0,0 +1,331 @@
...
+    @Inject
+    private TerminationService terminationService;
+
+    @Bean(name = "STARTING_DOWNSCALE_STATE")
+    public Action<?, ?> startingDownscaleAction() {
+        return new AbstractDownscaleAction<>(DownscaleEvent.class) {
+            @Override
+            protected void doExecute(FreeIpaDownscaleContext context, DownscaleEvent payload, Map<Object, Object> variables) {
+                Stack stack = context.getStack();
+                List<String> instanceIds = context.getInstanceIds();
```
lacikaaa, line 98: I would just simply get the `instanceIds` from the event, and put into the `variables` here

freeipa/src/main/java/com/sequenceiq/freeipa/controller/FreeIpaV1Controller.java
```diff
@@ -194,6 +199,17 @@ public void rebootInstances(@Valid RebootInstancesRequest request) {
         rebootInstancesService.rebootInstances(accountId, request);
     }
 
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.ENVIRONMENT_WRITE)
+    @Retryable(value = RetryableFreeIpaClientException.class,
```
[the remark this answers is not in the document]
jamisonbennett: I agree the only place that it matters is in the health checks. I would rather not push it down because otherwise, in the future I think it may lead to multilevel retries. I suspect the health service is going to be called by code that we either don't want to retry or that we don't want to retry at two levels.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/handler/RemoveHostsHandler.java
```diff
@@ -0,0 +1,117 @@
...
+                    .forEach(instanceMetaData -> removeNodePrivateIPsByFQDN.put(instanceMetaData.getDiscoveryFQDN(), instanceMetaData.getPrivateIp()));
+            Set<InstanceMetaData> remainingInstanceMetaDatas = stack.getNotDeletedInstanceMetaDataList().stream()
+                    .filter(instanceMetaData -> !removeNodePrivateIPsByFQDN.containsValue(instanceMetaData.getPrivateIp()))
+                    .collect(Collectors.toSet());
+            Set<Node> remainingNodes = remainingInstanceMetaDatas.stream()
+                    .map(im -> new Node(im.getPrivateIp(), im.getPublicIp(), im.getInstanceId(),
+                            im.getInstanceGroup().getTemplate().getInstanceType(), im.getDiscoveryFQDN(), im.getInstanceGroup().getGroupName()))
+                    .collect(Collectors.toSet());
+            List<GatewayConfig> remainingGatewayConfigs = gatewayConfigService.getGatewayConfigs(stack, remainingInstanceMetaDatas);
+            hostOrchestrator.tearDown(remainingGatewayConfigs, removeNodePrivateIPsByFQDN, remainingNodes, new StackBasedExitCriteriaModel(stack.getId()));
```
lacikaaa, line 110: could you add logging before this line?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -48,6 +48,18 @@ public void deleteDnsRecordByFqdn(String environmentCrn, String accountId,  List
...
+                    .collect(Collectors.toSet());
+            for (DnsRecord dnsRecord : srvRecordsToDelete) {
+                for (String fqdn : fqdns) {
+                    List<String> srvRecords = dnsRecord.getHostRelatedSrvRecords(fqdn);
+                    if (!srvRecords.isEmpty()) {
+                        LOGGER.info("Delete DNS SRV record [{}] for [{}] in zone [{}]", dnsRecord.getIdnsname(), fqdn, dnsZone);
+                        freeIpaClient.deleteDnsSrvRecord(dnsRecord.getIdnsname(), dnsZone.getIdnsname(), srvRecords);
+                    }
+                }
+            }
```
lacikaaa, line 62: I think it's worth to refactor the body of the `for (DnsZoneList dnsZone : freeIpaClient.findAllDnsZone()) {` and introduce 1 or 2 new methods

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -0,0 +1,119 @@
...
+                .flatMap(instanceGroup -> instanceGroup.getInstanceMetaData().stream())
+                .collect(Collectors.toMap(InstanceMetaData::getInstanceId, Function.identity()));
+    }
+
+    /**
+     * If no instance passed in request, repair all bad instances
+     * If instances passed in request, repair all valid passed bad instances
+     * If force and instances passed in request, repair all valid passed instances
+     * If force and no instances passed in request, repair all instances
+     * @param accountId - The account id for the instance to reboot.
```
lacikaaa, line 101: not reboot, repair

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param request - A RepairInstanceRequest containing request parameters.
+     */
+    public void repairInstances(String accountId, RepairInstancesRequest request) {
+        Stack stack = stackService.getByEnvironmentCrnAndAccountIdWithLists(request.getEnvironmentCrn(), accountId);
+        MDCBuilder.buildMdcContext(stack);
+        Map<String, InstanceMetaData> allInstancesByInstanceId = getAllInstancesFromStack(stack);
+        Map<String, InstanceMetaData> instancesToRepair = getInstancesToRepair(allInstancesByInstanceId, accountId, request);
+
+        if (instancesToRepair.keySet().size() > 0) {
```
lacikaaa, line 110: `isEmpty` instead of `> 0`

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -0,0 +1,119 @@
...
+import com.sequenceiq.freeipa.controller.exception.NotFoundException;
+import com.sequenceiq.freeipa.entity.InstanceMetaData;
+import com.sequenceiq.freeipa.entity.Operation;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.flow.freeipa.downscale.event.DownscaleEvent;
+import com.sequenceiq.freeipa.service.freeipa.flow.FreeIpaFlowManager;
+import com.sequenceiq.freeipa.service.operation.OperationService;
+
+@Service
+public class RepairInstancesService {
```
lacikaaa, line 35: this is very similar to `RebootInstancesService`. could you refactor so the common part is in a common class?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -231,10 +234,10 @@ public void installAndStartMonitoring(List<GatewayConfig> allGateway, Set<Node>
     public void stopTelemetryAgent(List<GatewayConfig> allGateway, Set<Node> nodes, ExitCriteriaModel exitModel)
             throws CloudbreakOrchestratorFailedException {
         GatewayConfig primaryGateway = getPrimaryGatewayConfig(allGateway);
-        Set<String> gatewayTargets = getGatewayPrivateIps(allGateway);
+        Set<String> targetHostnames = nodes.stream().map(Node::getHostname).collect(Collectors.toSet());
         try (SaltConnector sc = createSaltConnector(primaryGateway)) {
-            StateAllRunner stateAllJobRunner = new StateAllRunner(gatewayTargets, nodes, "fluent.agent-stop");
-            OrchestratorBootstrap saltJobIdTracker = new SaltJobIdTracker(sc, stateAllJobRunner);
+            StateRunner stateRunner = new StateRunner(targetHostnames, nodes, FLUENT_AGENT_STOP);
+            OrchestratorBootstrap saltJobIdTracker = new SaltJobIdTracker(sc, stateRunner);
```
lacikaaa, line 240: @oleewere could you check this part as the target from gateways moved to all nodes. also switched from private IPs to hostnames when targeting

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
```diff
@@ -84,6 +85,13 @@
             nickname = "rebootV1")
     void rebootInstances(@Valid RebootInstancesRequest request);
 
+    @POST
+    @Path("repair")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = FreeIpaOperationDescriptions.REPAIR, produces = MediaType.APPLICATION_JSON, notes = FreeIpaNotes.FREEIPA_NOTES,
+            nickname = "repairV1")
+    void repairInstances(@Valid RepairInstancesRequest request) throws Exception;
```
lacikaaa, line 93: shouldn't this return with either a flowid or an operation id?

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
jamisonbennett: Since reboot is a part of repair and at the higher level API it is just called repair, I moved the `RebootInstancesService` into `RepairInstanceService`. But I kept the FreeIPA REST APIs as reboot and repair.

jamisonbennett:
The last update added cloudwatch cleanup for AWS and volume cleanup for Azure. Volumes were already being cleaned up for AWS.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsDownscaleService.java
```diff
@@ -67,17 +71,22 @@
                 instanceIds.add(vm.getInstanceId());
             }
 
+            AwsCredentialView credentialView = new AwsCredentialView(auth.getCloudCredential());
+            AuthenticatedContextView authenticatedContextView = new AuthenticatedContextView(auth);
+            String regionName = authenticatedContextView.getRegion();
+            awsCloudWatchService.deleteCloudWatchAlarmsForSystemFailures(stack, regionName, credentialView, instanceIds);
```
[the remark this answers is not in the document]
jamisonbennett: `deleteCloudWatchAlarmsForSystemFailures()` should only perform actions for FreeIPA because only FreeIPA management service sets `CLOUDWATCH_CREATE_PARAMETER`.

```
        String isCreate = stack.getParameters().get(CLOUDWATCH_CREATE_PARAMETER);
        if (isCreate != null && isCreate.equals(Boolean.TRUE.toString())) {
```

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -71,20 +72,41 @@ public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances,
...
+                throw new CloudConnectorException("Unable to delete cloud watch alarms for system failure because instances ID are not a part of the stack.");
+            }
+            deleteCloudWatchAlarmsForSystemFailures(regionName, credentialView, instanceIds);
         }
     }
+
+    private void deleteCloudWatchAlarmsForSystemFailures(String regionName, AwsCredentialView credentialView, List<String> instanceIds) {
+        instanceIds.stream().forEach(instanceId -> {
+            try {
+                DeleteAlarmsRequest deleteAlarmsRequest = new DeleteAlarmsRequest().withAlarmNames(instanceId + alarmSuffix);
```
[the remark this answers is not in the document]
jamisonbennett: Although it is possible to request up to 100 alarms to be deleted, "In the event of an error, no alarms are deleted." - https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_DeleteAlarms.html

We could do something where we partition the list into sets of 100 and try to delete them (but this is not needed in the case of FreeIPA). If there is a failure then we could fall back to this implementation, that way we ensure all alarms that can be deleted are deleted. If the alarm deletion was used by datahub, then this should be done. But its currently only enabled for FreeIPA. With FreeIPA management service, there should be at most 3 instances in a delete request since there is a 4 node max. But in most cases, I suspect we will rarely get to cases where 4 nodes are used and 3 of them have failed. I am not sure the complexity is worth it here.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/FreeIpaV1Controller.java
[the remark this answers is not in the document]
jamisonbennett: ok, I removed the retry here and also in the reboot since its basically the same situation.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsDownscaleService.java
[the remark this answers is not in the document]
lacikaaa: ok I missed this isCreate

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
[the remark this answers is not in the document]
lacikaaa: so for FreeIPA this is not an issue, but my fear is if we might introduce this for clusters for some reason, we won't check if it scales well. So my suggestion here is to make bulk delete, and partition it by a 100.
Also the current way is not too idempotent as if we have to retry this state and like half of the alarm delete it will fail when we try to delete a non existing one.
So we should use describeAlarms or similar to fetch which exists and try to delete them. Also we might want to fail deletion if we can't delete them, because we will leave trash behind.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
```diff
@@ -346,6 +340,16 @@ private void checkTemplateDeployment(List<CloudResourceStatus> result, AzureClie
         return resourcesToRemove;
     }
 
+    private void collectAttachedDisks(Map<String, Object> resourcesToRemove, AuthenticatedContext ac, CloudStack stack, AzureCredentialView azureCredentialView,
```
lacikaaa, line 343: so it's a bit misleading, I would expect here to have a collection of disks returned.
instead we add 1 item to a map. so I think this should be renamed, like `addAttachedDiskStorageForRemoval`, or refactor to return the name and call sth like `findAttachedDiskStorageName`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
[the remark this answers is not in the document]
jamisonbennett: I created CB-7340 for this.

PR #8090, opened 2020-05-18
CB-7095 some modification on integration test

doktoric:


freeipa/src/main/resources/application.yml
```diff
@@ -229,7 +229,7 @@ cb:
 
 clusterProxy:
   url: http://localhost:10180/cluster-proxy
-  enabled: true
+  enabled: false
```
doktoric, line 232: This was discussed with @keyki . It seems that if we enable then we can not run IT's locally

keyki: Yes, but I didn't say to disable it :)

doktoric: I thought we agreed that with this property the ITs do not working. What should be the solution here ?

PR #8098, opened 2020-05-19
CB-7073 azure storage assert should check only the clusters of the test

horadla23:
I noticed that azure storage assert in e2e tests is checking all blobs in the given container which is time consuming and unnecessary, so I changed to check only the cluster and freeipa of the given test


horadla23:
Screenshot 2020-05-19 at 21 34 19

horadla23:
it seems only azure storage sdk had to be updated
Screenshot 2020-05-20 at 06 58 23

PR #8100, opened 2020-05-19
DISTX-421 DISTX-427 Recommendation for host groups' scalability

cegganesh84:
1. Added recommendations for auto-scalability and manual scalability.
2. Implemented both CM service based recommendations and fallback recommendations.

./gradlew build

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
@@ -1,0 +1,58 @@
+package com.sequenceiq.cloudbreak.cloud.model;
+
+import java.util.Objects;
+import java.util.Set;
+
+public class AutoscaleRecommendation {
+
+    private Set<String> timeBasedHostGroups;
+
+    private Set<String> loadBasedHostGroups;
+
+    public AutoscaleRecommendation(Set<String> timeBasedHostGroups, Set<String> loadBasedHostGroups) {
+        this.timeBasedHostGroups = Set.copyOf(timeBasedHostGroups);
+        this.loadBasedHostGroups = Set.copyOf(loadBasedHostGroups);
+    }
+
+    public Set<String> getTimeBasedHostGroups() {
+        return timeBasedHostGroups;
+    }
+
+    public Set<String> getLoadBasedHostGroups() {
+        return loadBasedHostGroups;
+    }
+
+    public void setTimeBasedHostGroups(Set<String> timeBasedHostGroups) {
+        this.timeBasedHostGroups = timeBasedHostGroups;
+    }
+
+    public void setLoadBasedHostGroups(Set<String> loadBasedHostGroups) {
+        this.loadBasedHostGroups = loadBasedHostGroups;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || !getClass().equals(o.getClass())) {
+            return false;
+        }
+        AutoscaleRecommendation that = (AutoscaleRecommendation) o;
+        return Objects.equals(timeBasedHostGroups, that.timeBasedHostGroups)
+                && Objects.equals(loadBasedHostGroups, that.loadBasedHostGroups);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(timeBasedHostGroups, loadBasedHostGroups);
+    }
+
+    @Override
+    public String toString() {
+        return "AutoscaleRecommendation{" +
+                "timeBasedHostGroups=" + timeBasedHostGroups +
+                ", loadBasedHostGroups=" + loadBasedHostGroups +
+                '}';
+    }
+}
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
@@ -10,6 +10,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -21,6 +22,7 @@ import org.springframework.stereotype.Service;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
+import com.sequenceiq.cloudbreak.cloud.model.AutoscaleRecommendation;
 import com.sequenceiq.cloudbreak.cloud.model.CloudVmTypes;
 import com.sequenceiq.cloudbreak.cloud.model.DiskType;
 import com.sequenceiq.cloudbreak.cloud.model.DiskTypes;
@@ -29,6 +31,7 @@ import com.sequenceiq.cloudbreak.cloud.model.InstanceCount;
 import com.sequenceiq.cloudbreak.cloud.model.Platform;
 import com.sequenceiq.cloudbreak.cloud.model.PlatformDisks;
 import com.sequenceiq.cloudbreak.cloud.model.PlatformRecommendation;
+import com.sequenceiq.cloudbreak.cloud.model.ResizeRecommendation;
 import com.sequenceiq.cloudbreak.cloud.model.VmRecommendation;
 import com.sequenceiq.cloudbreak.cloud.model.VmRecommendations;
 import com.sequenceiq.cloudbreak.cloud.model.VmType;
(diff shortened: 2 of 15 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
```diff
@@ -0,0 +1,41 @@
+package com.sequenceiq.cloudbreak.cloud.model;
+
+import java.util.Objects;
+import java.util.Set;
+
+public class AutoscaleRecommendation {
+
+    private final Set<String> hostGroups;
```
smaniraju, line 8: Splitting this into loadBasedScaling \ timeBasedScaling will be good.

smaniraju: Also will AutoscaleRecommendation be exposed as top level api BlueprintUtilV4Endpoint::getAutoscaleRecommendation(workspaceId,blueprintName) for validation in periscope against Autoscale CLI requests?  Something similar to BlueprintUtilV4Endpoint::getServicesByBlueprint. BlueprintUtilV4Endpoint::createRecommendation will not be usable from periscope since parameters like credentialName, region are not available.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
```diff
@@ -166,6 +177,56 @@ private boolean fallbackGatewayFilter(String hostGroupName) {
...
+        }
+
+        return autoscaleRecommendation;
+    }
+
+    private boolean fallbackAutoscaleFilter(String hostGroupName) {
+        String lowerCaseName = hostGroupName.toLowerCase();
+        return lowerCaseName.contains("compute")
+                || lowerCaseName.contains("gateway");
+    }
```
smaniraju, line 228: please include only compute.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
cegganesh84: Split is done.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
cegganesh84: Split it into time and load. Included only compute for load.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/BlackListedLoadBasedAutoscaleRole.java
```diff
@@ -0,0 +1,12 @@
...
+
+public enum BlackListedLoadBasedAutoscaleRole {
+    DATANODE,
+    ZEPPELIN_SERVER,
+    KAFKA_BROKER,
+    NIFI_NODE,
+    NAMENODE,
+    // This is counter intuitive. Gateway and compute host groups have identical GATEWAY roles. So compute host group will be blacklisted initially.
+    // The fallback filter will add only compute host group for load based auto-scale role.
+    GATEWAY
```
sidseth, line 11: All of these will likely need to be qualified as ServiceName.RoleName at some point.
e.g. Both HDFS and YARN have a Role called 'GATEWAY' - and I'm guessing other services do as well.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/BlackListedLoadBasedAutoscaleRole.java
```diff
@@ -0,0 +1,12 @@
+package com.sequenceiq.cloudbreak.cmtemplate;
+
+public enum BlackListedLoadBasedAutoscaleRole {
```
sidseth, line 3: Using a blacklist will need all services from CM to be added in here; or make a new release of CB when we add new services.
If I'm not mistaken, with this set of services - the OperationDB (HBase) clusters will qualify for both Time and Load based AutoScaling.

A Whitelist is an alternate, which comes with its own set of problems. i.e. if "NODEMANAGER" is the only role there, the GATEWAY roles that go along with the NODEMANAGER in a hostgroup will cause trouble.

Maybe the GATEWAYs just need to be removed from processing. i.e. A GATEWAY (unless it is special) does not block a whitelisted role from being scaled, or in case of BLACKLISTING - GATEWAY roles are just filtered out. If that leads to no roles left for a hostGroup - the hostGroup is not scalable (but maybe we should allow this to be time scalable).

I had a slightly different take on how this would work - based more on a 'Role' -> Capability mechanism - with all roles on a hostGroup combined together. A separate enum per capability (this patch) works as well.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/BlackListedLoadBasedAutoscaleRole.java
cegganesh84: Yup, `GATEWAY` wipes off everything from the slate, rendering the service-based logic meaningless.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
cegganesh84: [This](https://gk1-console.cdp-priv.mow-dev.cloudera.com/cloud/cloudbreak/v4/0/blueprints_util/recommendation?blueprintName=7.1.0%20-%20Data%20Engineering:%20HA:%20Apache%20Spark,%20Apache%20Hive,%20Apache%20Oozie&credentialName=gk1cred&region=us-west-2&resourceType=DATAHUB) returns,

```
"autoscaleRecommendation":{"timeBasedHostGroups":["compute","gateway"],"loadBasedHostGroups":["compute"]},"resizeRecommendation":{"scaleUpHostGroups":["worker","gateway","compute"],"scaleDownHostGroups":["worker","gateway","compute"]}}
```

I will check the generated swagger definition and see what apis to add.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/BlackListedLoadBasedAutoscaleRole.java
cegganesh84: ### Regarding OpDB

[Gateway, Master, Leader, Worker] of these,

Time based -> [Gateway, Leader] # Master has NAMENODE and Worker has DATANODE
Load based -> [] # Gateway and Leader has GATEWAY

### General problem

I was thinking of Blacklist filtering, followed by forceful whitelisting. Which is kind of acheived now using the fallback filter, just that it is counter-intuitive. 

We have two fields in the CM template, roleType, refName. We see the problems with roleType. This is existing implementation for Nifi cluster, which works. However when force fitted into other templates, it struggles. RefName is interesting and what you asked for in the previous comment. NODEMANAGER for worker gets yarn-NODEMANAGER-WORKER and compute gets yarn-NODEMANAGER-COMPUTE. I am not sure if we can rely on this naming, and how it will change for custom templates.

sidseth: roleType is what I was referring to. Service.RoleType fully qualifies an entity.
YARN.NODEMANAGER
YARN.GATEWAY
HDFS.GATEWAY

Yup, OpDB will be handled because of the presence of the DataNode. Think it is easier to reason with 'Scaling supported only for NodeManager'.

IAC, this works. The API for the UI, and whatever will be added for Periscope, remains unchanged irrespective of blacklist vs whitelist.

Any thoughts on ignoring GATEWAYs in the processing?

Also, do we want to allow schedule policies to be defined for GATEWAY nodes? (What if Hue is running on one of these nodes - is that safe?. Should Hue be in the blacklist?)

cegganesh84: > Service.RoleType fully qualifies an entity.
> YARN.NODEMANAGER

For me it did not. I want to whitelist YARN.NODEMANAGER-COMPUTE. Here compute is placement. If I whitelist YARN.NODEMANAGER then worker is also whitelisted. We can over come that with first blacklist and then whitelist. Remove the ones with DATANODE then add the ones with NODEMANAGER.

I chose DATANODE over NODEMANAGER for this reason.

> Any thoughts on ignoring GATEWAYs in the processing?

Yes, it depends on GATEWAY node support.

We can ignore GATEWAY nodes if it simplifies things. Hue is not a blocker, it can scale well. The Knox topology does not change. Hue sits behind CM controlled load-balancer. DAS, and Livy could be added as blacklists. Since we have not thought through this, we can ignore it for now.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
smaniraju: Yes this response looks good. All autoscaling related CB API  are accessed  via AutoscaleV4Endpoint. It would be nice if this is also exposed in that.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
smaniraju: 👍

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/BlackListedLoadBasedAutoscaleRole.java
sidseth: YARN.NODEMANAGER-COMPUTE is a dynamic name, and can be renamed to anything in a custom template. Think you mentioned that further up as well. YARN.NODEMANAGER, however, is determined by CM.

We are not going to be testing gateway node auto-scaling. Also, I don't think we're allowing more than one autoScaling policy for a cluster at the moment. Think it is better to make sure the GATEWAY is not auto-scalable for the moment. There's also limited value in this, since the hostnames will change each time a scale operation is run.

So - we're going with a BLACKLIST for now, and trying to put enough in there to ensure that only the compute group can be automatically scaled.

sidseth: Does it make sense to disable 'gateway' hosts from AutoScaling?, since we are not testing them.

cegganesh84: Yep, I will remove both gateway role and gateway node group.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/AutoscaleRecommendation.java
cegganesh84: I will add the endpoint via a new PR.

cegganesh84:
ping @keyki

cegganesh84:
Thanks @bergerdenes for taking a look. I have addressed those comments.

cegganesh84:
@bergerdenes could you please merge this?

bergerdenes:
Merged

PR #8101, opened 2020-05-19
CDPCP-1941. Allow setting ranger azure cloud id configs in FMS

aarman-cloudera:
This introduces a service component (CloudIdentityRangerSyncService) that
user sync will call in order to set azure cloud identity configurations
(oid mappings) into the datalake instance for a specified environment.

This is an initial implementation/poc. In future commits we'll need to
hook this up to user sync service, trigger CM role refresh, better
exception handling and error propagation, etc.

Also note that the necessary CM configs will be made available in
CM 7.2.1, I've tested my changes against a manually patched CM node.
    
Testing Done

```
String envCrn = "crn:cdp:environments:us-west-1:cloudera:environment:c2874f4b-6ac1-4d2e-b28d-c243d5ac60b0";
var azureUserMapping = Map.of("K1","V1","K2","V2");
var azureGroupMapping = Map.of("K3","V3");
cloudIdentityRangerSyncService.updateAzureCloudIdentityMapping(envCrn, azureUserMapping, azureGroupMapping);
```

Verified correct CM Configs were set

```
{
  "items": [
    ...,
    {
      "name": "ranger_usersync_azure_group_mapping",
      "value": "K3=V3",
      "sensitive": false
    },
    {
      "name": "ranger_usersync_azure_user_mapping",
      "value": "K2=V2;K1=V1",
      "sensitive": false
    },
    ...
  ]
}
```

### DIFF
--- a/freeipa/build.gradle
+++ b/freeipa/build.gradle
@@ -50,6 +50,7 @@ dependencies {
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jaxrs2',                   version: opentracingJaxrs2Version
   implementation     group: 'io.opentracing.contrib',    name: 'opentracing-jdbc',                     version: opentracingJdbcVersion
 
+  implementation     group: 'com.cloudera.api.swagger',    name: 'cloudera-manager-api-swagger',       version: cmClientVersion
 
   implementation     group: 'org.apache.kerby',          name: 'kerb-util',                            version: '2.0.0'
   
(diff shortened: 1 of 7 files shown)


keyki:
I don't think the FreeIPA management service should call anything other than FreeIPA itself. If we're going to call Datalake then probably the Datalake should handle the calls towards CM.

aarman-cloudera:
> I don't think the FreeIPA management service should call anything other than FreeIPA itself. If we're going to call Datalake service then probably the Datalake service should handle the calls towards CM. In FMS we cannot do the necessary checks and handling of CM API calls, like the runtime version depending calls.

I've put a more complete review up here that includes updating the user sync service in FMS : https://github.com/hortonworks/cloudbreak/pull/8230

In that PR I've moved some of the logic into Datalake service, though it's still going through cluster proxy. Can you have a look at that one instead, I'll close this one out.

PR #8109, opened 2020-05-20
CB-7146 [E2E] Save Java VM dumps on out-of-memory conditions

aszegedi:


integration-test/docker-compose_template.yml
```diff
@@ -23,7 +23,7 @@ services:
     volumes:
       - ./:/it
       - /etc/localtime:/etc/localtime:ro
-    command: java -jar /it/build/libs/cloudbreak-integration-test.jar com.sequenceiq.it.IntegrationTestApp --integrationtest.command=suiteurls --integrationtest.suiteFiles=${INTEGRATIONTEST_SUITEFILES} | tee cbtest.log
+    command: java -Xmx1g -XX:+UseG1GC -XX:+UseStringDeduplication -XX:OnOutOfMemoryError="kill -9 %p" -XX:+HeapDumpOnOutOfMemoryError -jar /it/build/libs/cloudbreak-integration-test.jar com.sequenceiq.it.IntegrationTestApp --integrationtest.command=suiteurls --integrationtest.suiteFiles=${INTEGRATIONTEST_SUITEFILES} | tee cbtest.log
```
lnardai, line 26: - Why do we need to use UseStringDeduplication?
- Why are we specifying the GC algorithm? (I think G1GC is the default form Java 9 and upwards.)
- `-Xmx1g` why do we use this specific value?

aszegedi: **-XX:+UseG1GC -XX:+UseStringDeduplication**
> Modern Java applications do a lot of string manipulations due to web service API calls (i.e. JSON, REST, SOAP, …), external data source calls (SQL, data returned back from DB, …), text parsing, text building, etc. Thus, string objects can easily occupy at least 30% of memory. Apparently, the majority of those String objects are duplicates. Because of string duplication, a considerable amount of memory is wasted. Thus, to optimize the memory wasted by duplicate string objects, JEP 192 has become a welcome enhancement to Java.
> 
> What Does JEP 192 Do?
> When the G1 GC algorithm runs, it removes garbage objects from memory. It also removes duplicate string objects from memory — string deduplication. This feature can be activated by passing the following JVM arguments: -XX:+UseG1GC -XX:+UseStringDeduplication
> Note 1: In order to use this feature, you need to run on Java 8 update 20 or later versions.
> Note 2: In order to use ‘-XX:+UseStringDeduplication’, you need to be using the G1 GC algorithm.
[by Ram Lakshmanan](https://dzone.com/articles/easy-change-to-reduce-memory-by-20)

**-Xmx<size>: Maximum Java heap size.** 
> The value is 25% of the available memory with a maximum of 25 GB. However, where there is 2 GB or less of physical memory, the value set is 50% of available memory with a minimum value of 16 MB and a maximum value of 512 MB.
> Start of content that applies only to Java 8 (LTS) If you have set the -XX:+OriginalJDK8HeapSizeCompatibilityMode option for compatibility with earlier releases, the value is half the available memory with a minimum of 16 MB and a maximum of 512 MB.
by [Default settings for the OpenJ9 VM](https://www.eclipse.org/openj9/docs/openj9_defaults/)

lnardai: I understand what the switches are for, I'm curios based on what information are we adding these values. If we would see from the ThreadDump that there are a lot of duplicate strings, sure try that.

I still think the G1GC is the default GC for Java11.

PR #8113, opened 2020-05-20
CB-7085 - Env: moving UMS api calls from nio thread to flow

attilapalfi92:


environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/EnvironmentInitHandler.java
```diff
@@ -52,6 +62,34 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+        String environmentCrnForVirtualGroups = environment.getResourceCrn();
+        if (Objects.nonNull(environment.getParentEnvironment())) {
+            environmentCrnForVirtualGroups = environment.getParentEnvironment().getResourceCrn();
+        }
+        return environmentCrnForVirtualGroups;
+    }
+
+    private boolean createVirtualGroups(Environment environment, String envCrn) {
+        boolean result = false;
+        if (StringUtils.isEmpty(environment.getAdminGroupName())) {
```
doktoric, line 85: isNullorEmpty ?

bergerdenes: isblank? :D

attilapalfi92: dont worry they all do the same. :D I like the isEmpty named version

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/EnvironmentInitHandler.java
```diff
@@ -52,6 +62,34 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+            environmentService.setAdminGroupName(environment, environment.getAdminGroupName());
+        }
+        environmentService.assignEnvironmentAdminAndOwnerRole(environment.getCreator(), environmentCrnForVirtualGroups);
+        setLocationAndRegions(environment);
+    }
+
+    private String getEnvironmentCrnForVirtualGroups(Environment environment) {
+        String environmentCrnForVirtualGroups = environment.getResourceCrn();
+        if (Objects.nonNull(environment.getParentEnvironment())) {
+            environmentCrnForVirtualGroups = environment.getParentEnvironment().getResourceCrn();
```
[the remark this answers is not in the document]
attilapalfi92: It is not an api call or anything, returns a reference. I can cache that reference but I think the JVM will do that.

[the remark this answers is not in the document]
attilapalfi92: (I moved this code without modification from the env creation service)

PR #8116, opened 2020-05-20
CB-7149 Clusterproxy should not be applied on Mock provider

doktoric:


cluster-proxy/src/main/java/com/sequenceiq/cloudbreak/clusterproxy/ClusterProxyEnablementService.java
```diff
@@ -0,0 +1,18 @@
...
+import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
+
+@Component
+public class ClusterProxyEnablementService {
+
+    @Inject
+    private ClusterProxyConfiguration clusterProxyConfiguration;
+
+    public boolean isClusterProxyApplicable(String cloudPlatform) {
+        return clusterProxyConfiguration.isClusterProxyIntegrationEnabled() && !CloudPlatform.MOCK.name().equals(cloudPlatform);
```
lacikaaa, line 16: instead of a hard coded value, could you use a configurable one and instead of a single item it could be a set/list so we can disable it on multiple platforms if necessary

doktoric: fair point. let me do that

PR #8119, opened 2020-05-20
CDPCP-2094. Add API to retrieve user sync state for an environment

handavid:
This commit adds a new endpoint for retrieving user sync state for an environment.
The state is calculated by looking at the last started full user sync. If the last
started full user sync completed successfully, then the event generation ids for
that sync are compared against the current event generation ids from the UMS.


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+    private UserSyncState calculateUserSyncState(String accountId, String envCrnString, UserSyncStatus userSyncStatus) {
+        Operation lastSync = userSyncStatus.getLastStartedFullSync();
+        UserSyncState state = UserSyncState.STALE;
+        switch (lastSync.getStatus()) {
+            case RUNNING:
+                state = UserSyncState.SYNCING;
+                break;
+            case COMPLETED:
+                if (lastSync.getSuccessList().stream()
+                        .anyMatch(details -> envCrnString.equals(details.getEnvironment()))) {
```
lacikaaa, line 68: move this to a method/variable

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+                    UmsEventGenerationIds currentEventGenerationIds = umsEventGenerationIdsProvider.getEventGenerationIds(accountId, MDCUtils.getRequestId());
+                    if (eventGenerationIdsChecker.isInSync(userSyncStatus, currentEventGenerationIds)) {
+                        state = UserSyncState.IN_SYNC;
+                    }
+                }
```
lacikaaa, line 73: move the whole to a method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+                break;
+            case REQUESTED:
+            case REJECTED:
+                // REQUESTED or REJECTED operations will never be saved as part of the UserSyncStatus
+                LOGGER.error("UserSyncStatus.lastStartedFullSync '{}' in unexpected state {} for environment '{}'",
+                        lastSync, lastSync.getStatus(), envCrnString);
+                throw new IllegalStateException(
+                        String.format("Last sync operation [%s] for environment '%s' is in unexpected state %s",
+                                lastSync.getOperationId(), envCrnString, lastSync.getStatus()));
```
lacikaaa, line 82: move to a method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+            case TIMEDOUT:
+                LOGGER.warn("UserSyncStatus.lastStartedFullSync '{}' is timed out for environment '{}'", lastSync.getOperationId(), envCrnString);
+                break;
+            case FAILED:
+            default:
+                // state already set to STALE
```
lacikaaa, line 88: it would be easier to read if every case would set the state instead of we initialize with a value and we have to go through all case to check if it's overridden

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -109,6 +114,14 @@ public SyncOperationStatus getSyncOperationStatus(@NotNull String operationId) {
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.ENVIRONMENT_READ)
+    public EnvironmentUserSyncState getUserSyncState(@NotNull String environmentCrn) {
+        LOGGER.debug("getUserSyncState() requested for environment '{}'", environmentCrn);
+
+        return environmentUserSyncStateCalculator.calculateEnvironmentUserSyncState(Crn.safeFromString(environmentCrn));
```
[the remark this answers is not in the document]
handavid: I disagree with your suggestion here. I've updated the PR to match current behavior of the other APIs, but will start an email discussion about the API usage of the actor crn since this issue is larger than this PR.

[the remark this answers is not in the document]
lacikaaa: sure, @horadla23 knows this stuff better

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+    private UserSyncStatusService userSyncStatusService;
+
+    @Inject
+    private UmsEventGenerationIdsProvider umsEventGenerationIdsProvider;
+
+    @Inject
+    private EventGenerationIdsChecker eventGenerationIdsChecker;
+
+    public EnvironmentUserSyncState calculateEnvironmentUserSyncState(Crn environmentCrn) {
+        String accountId = environmentCrn.getAccountId();
```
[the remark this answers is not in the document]
horadla23: i guess it is called with internal crn, thus we need to find out account id differently

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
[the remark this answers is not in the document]
horadla23: @handavid as i mentioned, with `@TenantAwareParam` annotation, you can get the actual account id from `ThreadBasedUserCrnProvider`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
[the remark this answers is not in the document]
horadla23: but @handavid you can use `@TenantAwareParam` as described

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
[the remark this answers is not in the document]
handavid: @horadla23  I looked at the code to the `InternalCrnModifier` and I find it strange that it changes the CRN stored in the `ThreadBasedUserCrnProvider` into a mixture of the original user crn and the account from the annotated field. I don't think this is desirable because you may end up with an invalid actor that would fail authz on subsequent API calls.

Instead, I think I need the changes you have in CB-7104 and to use the `@CheckPermissionByResourceCrn` annotation instead since the request includes an environment crn.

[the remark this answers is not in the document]
horadla23: @handavid we are doing that only in case of internal actpr and we do not call ums with internal actor since everything is allowed with that, so it is working, other team:s service can call environment service this way without problem. You can still use @CheckPermissionByResourceCrn, as you can see from example from my email

handavid:
@horadla23 @lacikaaa I updated the review to use the `@TenantAwareParam` and `CheckPermissionByResourceCrn`. I'd appreciate a re-review. thanks!

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
```diff
@@ -0,0 +1,115 @@
...
+            case RUNNING:
+                state = UserSyncState.SYNC_IN_PROGRESS;
+                break;
+            case COMPLETED:
+                state = calculateStateForCompletedOperation(accountId, envCrnString, userSyncStatus);
+                break;
+            case REQUESTED:
+            case REJECTED:
+                // REQUESTED or REJECTED operations will never be saved as part of the UserSyncStatus
+                state = calculateStateForUnexpectedOperationStatus(envCrnString, userSyncStatus);
```
lacikaaa, line 77: this name is a bit misleading as I would expect it would set the state, but instead it will throw an exception. also this method should be void as it won't return anything ever.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/UserV1Controller.java
```diff
@@ -102,13 +113,22 @@ public SyncOperationStatus setPassword(SetPasswordRequest request) {
...
-        checkUserCrn();
+        checkActorCrn();
         String accountId = ThreadBasedUserCrnProvider.getAccountId();
         LOGGER.debug("getSyncOperationStatus() requested for operation '{}' in account '{}'", operationId, accountId);
         return operationToSyncOperationStatus.convert(
                 operationService.getOperationForAccountIdAndOperationId(accountId, operationId));
     }
 
+    @Override
+    @CheckPermissionByResourceCrn(action = AuthorizationResourceAction.ENVIRONMENT_READ)
```
horadla23, line 124: I think you should use `@CheckPermissionByAccount` here (similarly to other methods), now every permission check on API methods happens on account level in freeipa, internal calls ( with `@TenantAwareParam`) still will work, resource based authz implementation is in progress,

handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/EnvironmentUserSyncStateCalculator.java
handavid: I renamed the method but I also changed it to return the. the case statement needs to throw the exception instead of the method. Otherwise, the compiler can't tell that the method will always throw and will complain about some codepaths not setting the `state` variable.

lacikaaa: thanks for the change!

PR #8120, opened 2020-05-21
CB-7162: FreeIPA AWS backup CLI command should use the regional endpoint

wonderslug:
This adds support to the AWS FreeIPA backup S3 uploader to set the region
to allow for the backup to use a regional endpoint.  It uses the region
defined for the stack.

This will result in upload url of s3.<region>.amazonaws.com instead of
s3.amazonaws.com as talked about in the documentation:
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

Closes #CB-7162

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
+++ b/freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
@@ -14,3 +14,4 @@ freeipa:
     location:
     azure_instance_msi:
     http_proxy:
+    aws_region:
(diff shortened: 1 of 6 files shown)


freeipa/src/main/resources/freeipa-salt/pillar/freeipa/init.sls
```diff
@@ -13,4 +13,5 @@ freeipa:
     platform:
     location:
     azure_instance_msi:
-    http_proxy:
\ No newline at end of file
+    http_proxy:
+    aws_regions:
```
keyki, line 17: minor: extra `s`

PR #8127, opened 2020-05-21
CB-6730. List instance groups even if all metadata are terminated.

oleewere:
### Problem
- on resize, if you have downscaled all the nodes for an instance groups you cannot get back the instance group itself with 0 instance metadata (if ALL terminated)

### Changes
- removed left join fetch query
- get instance groups by stack and filter in java on the non-terminated instance metadata

### Why like this?
- condition (not terminated) should go into the right side of the join -> BUT you cannot include any with/on-clause if you are using fetch
- without using fetch, we end up to using many queries (around 8)
- with the actual (removed) query, we still load stack anyway in the objects, so it contains all instance groups/instance metadata -> in terms of space/load query only the instance group with all metadata should not cause larger heap space

### DIFF
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
@@ -105,6 +105,12 @@ public class InstanceGroup implements ProvisionEntity, Comparable<InstanceGroup>
         this.stack = stack;
     }
 
+    public InstanceGroup replaceInstanceMetadata(Set<InstanceMetaData> instanceMetaData) {
+        this.instanceMetaData.clear();
+        Optional.ofNullable(instanceMetaData).ifPresent(this.instanceMetaData::addAll);
+        return this;
+    }
+
     public Set<InstanceMetaData> getNotTerminatedInstanceMetaDataSet() {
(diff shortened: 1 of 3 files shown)


sodre90:
@Query("SELECT i FROM InstanceGroup i JOIN FETCH i.instanceMetaData im "
            + "WHERE i.stack.id = :stackId AND ((im IS NULL OR im.instanceStatus <> 'TERMINATED') OR i IS NOT NULL)")

is works as well, could you please try it?

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
```diff
@@ -105,6 +105,12 @@ public void setStack(Stack stack) {
         this.stack = stack;
     }
 
+    public InstanceGroup instanceMetaData(Set<InstanceMetaData> instanceMetaData) {
```
topolyai5, line 108: this name is more readable and meaningful: `replaceInstanceMetadata` or stuff like this

oleewere: I have followed the same pattern as its used for other entities (like the instanceGroups in Stack object), but that can be done of course

PR #8131, opened 2020-05-21
CB-7009 Azure single RG: env service create RG - rename flow step

gergopapi2:


attilapalfi92:
thank you for reaching understanding and the excellent work @gergopapi2

PR #8132, opened 2020-05-21
CB-7181 HDFS create home script won't run on datalake

lacikaaa:


keyki:
Can you add some tests please?

PR #8134, opened 2020-05-22
CB-7190 Rename razenabed to razenabled

doktoric:


akanto:
I think it is worth investigating why the test has not caught it.

doktoric:
with empty db it works fine

doktoric:
raised CB-7202

PR #8138, opened 2020-05-22
CB-7189 Define sdx runtimeversion explicitly for e2e

horadla23:


horadla23:
I cherry-picked @doktoric's commit which simplifies runtimeversion configuration in e2e tests

PR #8139, opened 2020-05-22
CB-7143 - DistroX should validated with DB but not in imagevalidator test

gregito:
CB-7143 - DistroX should be validated with DB but not in imagevalidator test

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXEncryptedVolumeTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXEncryptedVolumeTest.java
@@ -1,23 +1,29 @@
 package com.sequenceiq.it.cloudbreak.testcase.e2e.distrox;
 
-import java.util.List;
-
-import javax.inject.Inject;
-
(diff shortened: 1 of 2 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXEncryptedVolumeTest.java
```diff
@@ -34,18 +41,24 @@ protected void setupTest(TestContext testContext) {
...
                 .build();
-        testContext.given(DistroXTestDto.class)
+        testContext.given(externalDB, DistroXExternalDatabaseTestDto.class)
+                .withAvailabilityType(DistroXDatabaseAvailabilityType.NON_HA)
+                .given(DistroXTestDto.class)
                 .withInstanceGroupsEntity(distroXInstanceGroupTestDtos)
+                .withExternalDatabase(externalDB)
                 .when(distroXTestClient.create())
                 .await(STACK_AVAILABLE)
+                .then(validateTemplateContainsExternalDatabaseHostname(Actor.defaultUser(testParameter)))
```
topolyai5, line 56: you can use the default user from the test context

PR #8149, opened 2020-05-23
DISTX-471 Periscope Entitlement Service Integration

smaniraju:
1. Introduce Entitlement Validations
2. Introduce SuspendedClusterMonitor which monitors suspended clusters and purges them if they are deleted in CB
3. Disable AlertEndpoint
4. Introduce DistroXAutoscaleEndpointTests which runs endpoint tests against embedded tomcat and embedded db.

Closes #DISTX-471

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ClusterStateEvaluator.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ClusterStateEvaluator.java
@@ -1,0 +1,44 @@
+package com.sequenceiq.periscope.monitor.evaluator;
+
+import javax.annotation.Nonnull;
+import javax.inject.Inject;
+
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.periscope.monitor.context.ClusterIdEvaluatorContext;
+import com.sequenceiq.periscope.monitor.context.EvaluatorContext;
+import com.sequenceiq.periscope.monitor.event.UpdateFailedEvent;
+
+@Component("ClusterStateEvaluator")
+@Scope("prototype")
+public class ClusterStateEvaluator extends EvaluatorExecutor {
+
+    private static final String EVALUATOR_NAME = ClusterStateEvaluator.class.getName();
+
+    @Inject
+    private EventPublisher eventPublisher;
+
+    private long clusterId;
+
+    @Override
+    public void setContext(EvaluatorContext context) {
+        clusterId = (long) context.getData();
+    }
+
+    @Override
+    @Nonnull
+    public EvaluatorContext getContext() {
+        return new ClusterIdEvaluatorContext(clusterId);
+    }
+
+    @Override
+    public String getName() {
+        return EVALUATOR_NAME;
+    }
+
+    @Override
+    public void execute() {
+        eventPublisher.publishEvent(new UpdateFailedEvent(clusterId));
+    }
+}
--- a/autoscale/src/main/java/com/sequenceiq/periscope/service/EntitlementValidationService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/service/EntitlementValidationService.java
@@ -1,0 +1,44 @@
+package com.sequenceiq.periscope.service;
+
+import java.util.Map;
+import java.util.Optional;
+
+import javax.inject.Inject;
+
+import org.springframework.beans.factory.annotation.Value;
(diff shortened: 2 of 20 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ClusterStateEvaluator.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return EVALUATOR_NAME;
+    }
+
+    @Override
+    public void execute() {
+        eventPublisher.publishEvent(new UpdateFailedEvent(clusterId));
```
sidseth, line 42: The handler for this ends up putting a cluster in the SUSPENDED state even if it already in the SUSPENDED state, which will end up being an unnecessary DB operation every minute for STOPPED clusters. Would be useful to avoid that.

Also, this will result in all known clusters being polled every minute from Cloudbreak. I think we should introduce a more light weigh API than the on that returns all Stack details (obviously a different PR for this).

autoscale/src/test/java/com/sequenceiq/periscope/modul/rejected/RejectedThreadContext.java
```diff
@@ -59,6 +60,7 @@
     @MockBean({Clock.class, ClusterService.class, CloudbreakClientConfiguration.class,
             MetricUtils.class, InternalCrnBuilder.class, FailedNodeRepository.class})
     @EnableAsync
+    @Profile("dev")
```
sidseth, line 63: Accidental change? or should this be test. Also other places where this change is made.

autoscale/src/test/java/com/sequenceiq/periscope/testcontext/EndpointTestContext.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.context.annotation.Profile;
+
+import com.sequenceiq.cloudbreak.service.secret.service.SecretService;
+
+@TestConfiguration
+@EntityScan(basePackages = {"com.sequenceiq.periscope"})
+@Profile("test")
```
sidseth, line 12: What should this be?

autoscale/src/main/java/com/sequenceiq/periscope/service/EntitlementValidationService.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
+
+@Service
+public class EntitlementValidationService {
+
+    private static final String ALLOWED = "ALLOWED_PLATFORM";
+
+    private static final Map<String, String> PLATFORM_ENTITLEMENTS = Map.of(
+            "AWS", "CDP_AWS_AUTOSCALING",
```
sidseth, line 20: Names changes to DATAHUB_ instead of CDP_

smaniraju: ack

autoscale/src/test/java/com/sequenceiq/periscope/testcontext/EndpointTestContext.java
smaniraju: It is test and runs in unit test cycle.

autoscale/src/test/java/com/sequenceiq/periscope/modul/rejected/RejectedThreadContext.java
smaniraju: This is correct, this context is used to setup specific thread pool size for dev mode testing.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ClusterStateEvaluator.java
smaniraju: Removed DB Hit. For CB Sync optimization I have already created this jira. https://jira.cloudera.com/browse/CB-6619

smaniraju:
> Left some minor comments. Would be better to get this in as 2 commits (entitlements / suspend-handling) - but doesn't make a big difference.
> 
> Assuming we can remove AlertController in a subsequent commit.

CB requires squashing of commits into a single commit in PR, so individual commits in PR or branch does not make much difference.

PR #8151, opened 2020-05-26
CB-7221 fixed azure npe in case of new resourcegroup

doktoric:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureNetworkConnector.java
```diff
@@ -78,12 +78,12 @@ public CreatedCloudNetwork createNetworkWithSubnets(NetworkCreationRequest netwo
                 Lists.newArrayList(networkRequest.getPublicSubnets()),
                 Lists.newArrayList(networkRequest.getPrivateSubnets()),
                 networkRequest.isPrivateSubnetEnabled());
-        String template = azureNetworkTemplateBuilder.build(networkRequest, subnetRequests);
 
         Deployment templateDeployment;
         ResourceGroup resourceGroup;
         try {
             resourceGroup = getOrCreateResourceGroup(azureClient, networkRequest);
+            String template = azureNetworkTemplateBuilder.build(networkRequest, subnetRequests, resourceGroup);
```
pdarvasi, line 86: we only need the name of the RG in the function

doktoric: I think it is better to use a real object than a simple string but yeah you can change because it seems already merged

PR #8152, opened 2020-05-26
ENGESC-2389. general.clusterManagerIp parameter is not resolved as parameter is not resolved as part of a recipe

oleewere:
Simple fix for resolving clouderaManagerIp for general configs (recipe) without touching anything on flow side.

Value is only filled if it does not exist already in the general configs (same for PrimaryGatewayInstanceDiscoveryFQDN)

As generating these configs are called from multiple places, it's not always needed

added unit tests as well

workaround without the fix, adding these to recipe scripts:
```bash
server_host=$(cat /etc/cloudera-scm-agent/config.ini | grep server_host | cut -d'=' -f2)
server_ip=`dig +short $server_host`
```


oleewere:
(note: the unit test failure nothing to do with the change: one test in ParcelAccessibilityTest fails, it seems some of our UTs are depending on that an external cloudera url is accessible or not)

PR #8161, opened 2020-05-27
CB-7163 Add spot max price capability

Bajzathd:


environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -178,16 +177,18 @@ private ParametersDto azureParamsToParametersDto(AzureEnvironmentParameters azur
...
-                .withFreeIpaSpotPercentage(Optional.ofNullable(awsFreeIpa)
-                        .map(AwsFreeIpaParameters::getSpot)
-                        .map(AwsFreeIpaSpotParameters::getPercentage)
-                        .orElse(0))
-                .build();
+                        .orElse(null));
+        Optional.ofNullable(awsFreeIpa)
+                .map(AwsFreeIpaParameters::getSpot)
+                .ifPresent(awsFreeIpaSpotParameters -> {
+                    builder.withFreeIpaSpotPercentage(awsFreeIpaSpotParameters.getPercentage())
```
bergerdenes, line 188: what if spotPercentage is not present? the orElse(0) logic is changed here

Bajzathd: null is treated as a default 0 value, so it was not necessary in the first place

PR #8173, opened 2020-05-28
CB-5738 Disable ClusterManagerHostHealthEvaluator

smaniraju:
1. Disable ClusterManagerHostHealthEvaluator.
2. Including autoscaling disabled clusters for removable check.

Closes CB-5738

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/RemovableClusterMonitor.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/RemovableClusterMonitor.java
@@ -5,17 +5,16 @@ import java.util.List;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
-import com.sequenceiq.periscope.api.model.ClusterState;
 import com.sequenceiq.periscope.domain.Cluster;
 import com.sequenceiq.periscope.monitor.evaluator.ClusterStateEvaluator;
 
-@Component("SuspendedClusterMonitor")
-@ConditionalOnProperty(prefix = "periscope.enabledAutoscaleMonitors.suspended-cluster-monitor", name = "enabled", havingValue = "true")
-public class SuspendedClusterMonitor extends ClusterMonitor {
+@Component("RemovableClusterMonitor")
+@ConditionalOnProperty(prefix = "periscope.enabledAutoscaleMonitors.removable-cluster-monitor", name = "enabled", havingValue = "true")
+public class RemovableClusterMonitor extends ClusterMonitor {
 
     @Override
     public String getIdentifier() {
-        return "suspended-cluster-monitor";
+        return "removable-cluster-monitor";
     }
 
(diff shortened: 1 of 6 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/RemovableClusterMonitor.java
```diff
@@ -0,0 +1,42 @@
...
+    protected List<Cluster> getMonitored() {
+        //To monitor Suspended & Clusters with AutoScaling disabled and purge them in periscope when they are deleted in CB.
+        List removableClustersCheck = getClusterService()
+                .findAllByStateAndNode(ClusterState.SUSPENDED, getPeriscopeNodeConfig().getId());
+        removableClustersCheck.addAll(getClusterService()
+                        .findAllByStateAndNode(ClusterState.PENDING, getPeriscopeNodeConfig().getId()));
+        removableClustersCheck.addAll(getClusterService()
+                .findAllForNode(ClusterState.RUNNING, false, getPeriscopeNodeConfig().getId()));
+        return removableClustersCheck;
+    }
```
smaniraju, line 41: Including subset of clusters to monitor instead of ClusterManagerHostHealthEvaluator's all running clusters.
Also one corner case was not covered in existing flow related to "PENDING" cluster deleted in CB.

sidseth: I may be missing something here.
What sends a 'Periscope Cluster' into the SUSPENDED or PENDING state?

ClusterCreationEvaluator seems to be the one setting state to 'PENDING' when it first pulls in a cluster.

The only place that seems to move a cluster into SUSPENDED state is UpdateFailedeHandler.
The UpdateFailedHandler will only be invoked by YarnLoadEvaluator (which we can ignore since LoadAlerts may not be enabled for a cluster) or ClusterStateEvaluator, which will only be invoked based on this 'getMonitored' list.

Won't we need to look up pretty much all clusters here?

Also, didn't understand the need for "RUNNING" + scalingEnabled=false. Why not RUNNING + scalingEnabled=true (which goes back to the previous point about monitoring all clusters).

smaniraju: Ideally RUNNING + scalingEnabled=true would not be required since it would be tested while in their autoscale runs, but realized CronTimeEvaluator does not check if Cluster is online.  Fixed it.

sidseth: Even if the LoadEvaluator made this change - this particular component should not be making any assumptions about the actual triggers. It is a mechanism to keep cluster state in sync with Cloudbreak, and should behave independently of other components - which I think the latest commit does.

smaniraju: ack

PR #8175, opened 2020-05-28
CB-7165 [ASRG] Implement Redbeams CloudResource based termination flow

gergopapi2:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/connector/resource/AzureDatabaseResourceService.java
```diff
@@ -112,52 +116,83 @@
...
+            return List.of();
+        }
+        String databaseServerId = dbServerResourceOptional.get().getReference();
+        try {
+            azureUtils.deleteDatabaseServer(client, databaseServerId);
+        } catch (CloudException e) {
+            errorHandling(force, e, "Database server", databaseServerId);
+        }
+
+        return Lists.newArrayList(new CloudResourceStatus(CloudResource.builder()
```
pdarvasi, line 163: We should consider deleting the actual resource as it won't be reused in future.

PR #8178, opened 2020-05-28
CB-7259 Ensure HBASEJARS is listed as an ExposedService

joshelser:
New unit test which demonstrates failure prior to correcting the
visibleForDatalake/visibleForDatahub both being false, and then passing
upon fixing exposed-services.json.

Copied the relevant text from other sections in topology_api.xml.j2,
but I'm not sure if there's some way to validate these changes are
correct prior to getting a deploy in dev.

Should cherry-pick cleanly to rc-2.21, if this change is acceptable. Would like to see this get deployed to validate it once and for all :)

### DIFF
--- a/core-api/src/main/resources/definitions/exposed-services.json
+++ b/core-api/src/main/resources/definitions/exposed-services.json
@@ -577,10 +577,12 @@
       "serviceName": "MASTER",
       "knoxService": "HBASEJARS",
       "knoxUrl": "/hbase/jars",
+      "port": 16010,
+      "tlsPort": 16010,
       "apiOnly": true,
       "apiIncluded": true,
       "visibleForDatalake": false,
-      "visibleForDatahub": false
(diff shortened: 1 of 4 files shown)


joshelser:
> should show the checkstyle failures without running all the tests.

Thanks sir! Super helpful. Fixed in 796c2f0

lacikaaa:
next time please don't create a separate commit to fix issues, we follow if possible a one commit/PR strategy so it's easy to revert/cherry pick if necessary

joshelser:
> next time please don't create a separate commit to fix issues, we follow if possible a one commit/PR strategy so it's easy to revert/cherry pick if necessary

Two commits was not intentional. Fully expected the first commit to do what I had intended it to do. We only had to come back here because the original fix did not work.

Are you suggesting that I should've reverted the original commit and re-applied + the extra changes?

PR #8185, opened 2020-05-29
CB-6236 do not add RAZ service to SDX template if not enabled

keyki:


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazConfigProvider.java
```diff
@@ -35,11 +35,14 @@
 
     @Override
     public Map<String, ApiClusterTemplateService> getAdditionalServices(CmTemplateProcessor cmTemplateProcessor, TemplatePreparationObject source) {
-        ApiClusterTemplateService coreSettings = createTemplate();
-        Set<HostgroupView> hostgroupViews = source.getHostgroupViews();
-        return hostgroupViews.stream()
-                .filter(hg -> InstanceGroupType.CORE.equals(hg.getInstanceGroupType()))
-                .collect(Collectors.toMap(HostgroupView::getName, v -> coreSettings));
+        if (isConfigurationNeeded(cmTemplateProcessor, source)) {
```
akanto, line 38: It looks like that this was not properly covered with test, please cover it in this fix.

keyki:
> Hmm, for me it looks like we wanted to control the RAZ in request time, but somehow that does not work...
> 
> My preference would be to figure out why this does not work and fix that part:
> https://github.com/hortonworks/cloudbreak/blob/a04709441071729e4019c2c487f2f7f0c9ee5fd9/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazConfigProvider.java#L60
> 
> And I think since you have done this in this pull, I think we don't need the RAZ entitlement, and that can be just deleted.

The entitlement is used for validation only. Even if you send raz enabled in the request it will reject if you are not entitled to. Based on this the entitlement makes sense, but no need to disable it by default.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazConfigProvider.java
keyki: Done

PR #8189, opened 2020-06-02
DISTX-421 Separate endpoint for autoscale recommendation

cegganesh84:
1. Autoscale needs a recommendation endpoint for the backend verification in periscope.
2. Tested by changing the auth and setting the workspace and blueprint manually. 

Depends on PR #8100 so sent to that PRs branch instead of master.

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
@@ -100,8 +100,7 @@ public interface AutoscaleV4Endpoint {
     ClusterProxyConfiguration getClusterProxyconfiguration();
 
     @GET
-    @Path("recommendation")
+    @Path("/stack/crn/{crn}/recommendation")
     @Produces(MediaType.APPLICATION_JSON)
-    AutoscaleRecommendationV4Response getRecommendation(@QueryParam("workspaceId") Long workspaceId,
-                                                        @QueryParam("blueprintName") String blueprintName);
+    AutoscaleRecommendationV4Response getRecommendation(@PathParam("crn") String clusterCrn);
 }
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
@@ -22,6 +22,7 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.autoscales.response.AuthorizeFo
 import com.sequenceiq.cloudbreak.api.endpoint.v4.autoscales.response.AutoscaleStackV4Responses;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.autoscales.response.CertificateV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.autoscales.response.ClusterProxyConfiguration;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.connector.responses.AutoscaleRecommendationV4Response;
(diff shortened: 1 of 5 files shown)


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
```diff
@@ -106,4 +107,10 @@ void decommissionInstancesForClusterCrn(@PathParam("crn") String clusterCrn,
     @Path("clusterproxy")
     @Produces(MediaType.APPLICATION_JSON)
     ClusterProxyConfiguration getClusterProxyconfiguration();
+
+    @GET
+    @Path("recommendation")
+    @Produces(MediaType.APPLICATION_JSON)
+    AutoscaleRecommendationV4Response getRecommendation(@QueryParam("workspaceId") Long workspaceId,
+                                                        @QueryParam("blueprintName") String blueprintName);
 }
```
smaniraju, line 116: Can this take only clusterCrn and internally use blueprintName ?

cegganesh84: Doing it by blueprintName is more cache-friendly. At the same time, clusterCrn is a nice abstraction, particularly caller need not worry about workspaceId.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
```diff
@@ -157,8 +157,7 @@ public ClusterProxyConfiguration getClusterProxyconfiguration() {
     }
 
     @Override
-    @DisableCheckPermissions
-    @PreAuthorize("hasRole('AUTOSCALE')")
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.DATAHUB_READ)
     public AutoscaleRecommendationV4Response getRecommendation(Long workspaceId, String blueprintName) {
```
smaniraju, line 161: This api is meant for internal periscope use which uses internalCrn for communication. It should use the earlier PreAuthorize checks  and not CheckPermissionByAccount.

cegganesh84: That makes sense. How to test those internal APIs (in general)?

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/autoscales/AutoscaleV4Endpoint.java
smaniraju: I would prefer it using clusterCrn, since it avoids additional CB call to retrieve blueprintname associated with the cluster. Also workspaceId is internal id of CB, so  would  prefer to avoid making api calls based on that.

cegganesh84:
IT failed with an error like this, restarted the build,

```
./scripts/docker-compose.sh: line 17: integcb/.deps/bin/docker-compose: No such file or directory
make: *** [docker-compose] Error 1
```

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CloudResourceAdvisor.java
```diff
@@ -149,6 +146,20 @@ public PlatformRecommendation createForBlueprint(Long workspaceId, String bluepr
         return new PlatformRecommendation(vmTypesByHostGroup, availableVmTypes, diskTypes, instanceCounts, gateway, autoscale, resize);
     }
 
+    public AutoscaleRecommendation getAutoscaleRecommendation(Long workspaceId, String blueprintName) {
+        LOGGER.debug("Autoscale advice for blueprintName: {}.", blueprintName);
+
+        BlueprintTextProcessor blueprintTextProcessor = getBlueprintTextProcessor(workspaceId, blueprintName);
+
+        return recommendAutoscale(blueprintTextProcessor);
```
smaniraju, line 154: Nit: empty lines

PR #8196, opened 2020-06-02
CB-7268 - Implemented tracing support for flow

attilapalfi92:
screenshots:

https://drive.google.com/drive/folders/15lytU6cUsu5EbDzN3tPcGBDAFXLHTxBQ?usp=sharing


flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
```diff
@@ -74,24 +81,28 @@ public void execute(StateContext<S, E> context) {
...
-                    String flowStateName = String.valueOf(variables.get(FLOW_STATE_NAME));
-                    long executionTime = Math.max(execElapsed, flowElapsed);
-                    LOGGER.debug("Stack: {}, flow state: {}, phase: {}, execution time {} sec", payload.getResourceId(),
-                            flowStateName, execElapsed > flowElapsed ? "doExec" : "service", executionTime);
-                    metricService.submit(FlowMetricType.FLOW_STEP, executionTime, Map.of("name", flowStateName.toLowerCase()));
+                String flowStateName = String.valueOf(variables.get(FLOW_STATE_NAME));
+                Span activeSpan = tracer.activeSpan();
+                String operationName = context.getEvent().name();
+                SpanContext spanContext = flowParameters.getSpanContext();
+                if (TracingUtil.isActiveSpanReusable(activeSpan, spanContext, operationName)) {
```
lacikaaa, line 88: could you refactor so both the if and else body is in a method?

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
```diff
@@ -74,24 +81,28 @@ public void execute(StateContext<S, E> context) {
...
+                    flowContext = createFlowContext(flowParameters, context, payload);
+                    doExecute(context, payload, flowContext, variables, flowStateName);
+                } else {
+                    Span span = TracingUtil.getSpan(tracer, operationName, spanContext, flowParameters.getFlowId(),
+                            null, flowParameters.getFlowTriggerUserCrn());
+                    if (spanContext == null) {
+                        spanContext = span.context();
+                    }
```
lacikaaa, line 96: if we create above a span with spanContext, do we really need this if? we could go with
`flowParameters.setSpanContext(span.context());`
if I'm not mistaken

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
lacikaaa: also some log would be nice to see what's happening around here

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
```diff
@@ -71,14 +75,37 @@
...
+        Span activeSpan = tracer.activeSpan();
+        SpanContext spanContext = event.getHeaders().get(FlowConstants.SPAN_CONTEXT);
+        String operationName = event.getKey().toString();
+        if (TracingUtil.isActiveSpanReusable(activeSpan, spanContext, operationName)) {
+            doAccept(event, key, payload, flowId, flowChainId, flowTriggerUserCrn, spanContext);
+        } else {
+            Span span = TracingUtil.getSpan(tracer, operationName, spanContext, flowId, flowChainId, flowTriggerUserCrn);
+            if (spanContext == null) {
+                spanContext = span.context();
+            }
```
lacikaaa, line 97: same here, do we really need this?

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
```diff
@@ -102,6 +113,23 @@ public void execute(StateContext<S, E> context) {
         });
     }
 
+    private void doExecute(StateContext<S, E> context, P payload, C flowContext, Map<Object, Object> variables, String flowStateName) throws Exception {
```
lacikaaa, line 116: this `doExecute` is a bit strange, doesn't have any extra info. Could you come up with a better naming, also we already have a `doExecute` call inside this method

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
```diff
@@ -71,14 +75,37 @@
...
+            try (Scope scope = tracer.activateSpan(span)) {
+                doAccept(event, key, payload, flowId, flowChainId, flowTriggerUserCrn, spanContext);
+            } finally {
+                span.finish();
+            }
+        }
+    }
+
+    private void doAccept(Event<? extends Payload> event, String key, Payload payload, String flowId, String flowChainId,
```
lacikaaa, line 106: similar here, can we come up with a meaningful name? like `acceptWithSpanContext` or whatever

flow/src/main/java/com/sequenceiq/flow/reactor/FlowParametersAspects.java
```diff
@@ -29,16 +39,37 @@ public Object setFlowTriggerUserCrnForReactorHandler(ProceedingJoinPoint proceed
...
+            Span activeSpan = tracer.activeSpan();
+            if (TracingUtil.isActiveSpanReusable(activeSpan, spanContext, operationName)) {
+                return doProceed(proceedingJoinPoint, flowTriggerUserCrn, event, spanContext);
+            } else {
+                Span span = TracingUtil.getSpan(tracer, operationName, spanContext, event.getHeaders().get(FlowConstants.FLOW_ID),
+                        event.getHeaders().get(FlowConstants.FLOW_CHAIN_ID), flowTriggerUserCrn);
+                try (Scope ignored = tracer.activateSpan(span)) {
+                    if (spanContext == null) {
+                        spanContext = span.context();
+                    }
```
lacikaaa, line 53: same

attilapalfi92:
@lacikaaa It does not interfere with already running flows. If a flow is restarted with this code, the remaining part of the flow will be traced and that's it.

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
attilapalfi92: I think doAccept is better, this is a common java terminology for calling a private method that does the actual think, like accepting in this case.

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
attilapalfi92: shit, there is one indeed. no extra info is needed, it does the execution. any idea what should I name it?

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
attilapalfi92: Yes it is needed because we always reuse the original and first span context. This way the trace will be usable, otherwise spans would be recursively coming from each other which is untrue for the trace. 
If a context is present we use it, else if this is the first step of a flow, we have to use the current span's context.

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
attilapalfi92: its already hard to come up with a name for doExecute. Can you help me out with it?

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
attilapalfi92: same answer

flow/src/main/java/com/sequenceiq/flow/reactor/FlowParametersAspects.java
attilapalfi92: same answer

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
attilapalfi92: didnt find a way to do it. multiple exit points all the time

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
attilapalfi92: renamed to executeAction

flow/src/main/java/com/sequenceiq/flow/core/AbstractAction.java
schfeca75: you must find one, try it harder please!

PR #8202, opened 2020-06-04
CB-7329 Cannot run aws-e2e job

topolyai5:


topolyai5:
> The test class and the related yaml were deleted purposefully at [CB-6493 this testcase is redundant with DistroXScaleTest so removing it](https://github.com/hortonworks/cloudbreak/commit/d779aa96b64afec6d1fc6a7deeea34706ce33d1f) by @lnardai committed on 17 Apr. So I think we should wait for @lnardai approval on this.

Without this class the aws-e2e-tests are failing. I discussed with Laci here: https://cloudera.slack.com/archives/CQ1R8L37D/p1591191577065200

PR #8206, opened 2020-06-04
CB-7294 [API E2E] Data Lake creation failed: Provisioning failed (AVAILABLE)

foldik:
Flow and flow chain marked completed with a little delay

### DIFF
--- a/flow/src/main/java/com/sequenceiq/flow/service/FlowStateDelayedEvaluator.java
+++ b/flow/src/main/java/com/sequenceiq/flow/service/FlowStateDelayedEvaluator.java
@@ -1,0 +1,32 @@
+package com.sequenceiq.flow.service;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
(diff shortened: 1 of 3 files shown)


flow/src/main/java/com/sequenceiq/flow/service/FlowStateDelayedEvaluator.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Value("${flow.completion.delay:10000}")
+    private long delayInMillis;
+
+    private final Map<String, Long> completedAt = new ConcurrentHashMap<>();
+
+    public boolean isComplete(String flowOrFlowChainId, boolean completed) {
+        if (completed) {
+            long currentTime  = System.currentTimeMillis();
+            long completionTime = completedAt.computeIfAbsent(flowOrFlowChainId, key -> currentTime);
```
sodre90, line 20: this solution can cause problems in multi node environment

PR #8207, opened 2020-06-04
CB-7290 - No services listed for 7.2.0 Datahub blueprints

gregito:
CB-7290 - No services listed for 7.2.0 Datahub blueprints


bergerdenes:
please make sure that the version numbers are correct (what is the source of those btw?)

PR #8214, opened 2020-06-04
CB-7343 [opdb] Add Hue in the opdb blueprint

jatsakthi:
Hue talks to HBase via the Thriftserver and can be used to view/modify data.
This change adds Hue to the opdb blueprint and enables thriftserver.

Testing: Created cluster from custom blueprint and verified that Hue turns up.
Tried CRUD operations on a sample table from the Hue UI.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
@@ -155,6 +155,13 @@
               }
             ],
             "base": true
+          },
+          {
+            "refName": "hbase-HBASETHRIFTSERVER-BASE",
+            "roleType": "HBASETHRIFTSERVER",
+            "displayName": null,
+            "configs": [],
+            "base": true
           }
         ]
       },
@@ -215,6 +222,37 @@
             "base": true
           }
         ]
+      },
+      {
+        "refName": "hue",
+        "serviceType": "HUE",
+        "serviceConfigs": [
+          {
+            "name": "hue_service_safety_valve",
+            "value": "[desktop]\napp_blacklist=impala,security,filebrowser,rdbms,jobsub,pig,sqoop,zookeeper,metastore,spark,oozie,indexer,hive,documents,indexes,search",
+            "ref": null,
+            "variable": null,
+            "autoConfig": false
+          }
+        ],
+        "roleConfigGroups": [
+          {
+            "refName": "hue-HUE_SERVER-BASE",
+            "roleType": "HUE_SERVER",
+            "displayName": null,
+            "configs": [],
+            "base": true
+          },
+          {
+            "refName": "hue-HUE_LOAD_BALANCER-BASE",
+            "roleType": "HUE_LOAD_BALANCER",
+            "displayName": null,
+            "configs": [],
+            "base": true
+          }
+        ],
+        "displayName": null,
+        "roles": []
       }
     ],
     "hostTemplates": [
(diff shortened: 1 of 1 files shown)


petersomogyi:
Have you checked the memory consumption for the added services (Thrift, Hue)?
Currently the memory on the worker nodes are max out from CM's point of view and it might show  "Host memory overcommitment" warning with Thrift and Hue.

jatsakthi:
The gateway and leader nodes were less burdened, hence added to these nodes. Haven't added anything in the worker nodes. Do you think we should still check if the gateway and leader nodes would bbe overloaded? if yes than maybe we should increase their allocated memories maybe in a following jira?

core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
```diff
@@ -225,7 +263,9 @@
           "hbase-GATEWAY-BASE",
           "hdfs-GATEWAY-BASE",
           "knox-KNOX-GATEWAY-BASE",
-          "yarn-GATEWAY-BASE"
+          "yarn-GATEWAY-BASE",
+          "hue-HUE_SERVER-BASE",
+          "hue-HUE_LOAD_BALANCER-BASE"
```
joshelser, line 268: If we only have one Hue server, do we still need the Hue LB?

joshelser:
> Do you think we should still check if the gateway and leader nodes would be overloaded?

Please check this before we commit this change. If memory is overcommitted, we might introduce instability/crashes for our clusters which is no good. I have no reason to expect that memory would be overcommitted, but we should be certain that CM is not doing something unexpected.

jatsakthi:
> > Do you think we should still check if the gateway and leader nodes would be overloaded?
> 
> Please check this before we commit this change. If memory is overcommitted, we might introduce instability/crashes for our clusters which is no good. I have no reason to expect that memory would be overcommitted, but we should be certain that CM is not doing something unexpected.

@joshelser & @petersomogyi , I checked the memory commitment statuses of the gateway, leader, master & the worker nodes. As Peter pointed out all 3 worker nodes are indeed overcommitted.(using 27gb out the 30 gb available memory, where they should only use 24 gb). Whereas Master - 21/30, Leader - 12/30, Gateway - 13/30. The Leader & Gateway numbers are with inclusion of the Hue(Server+LB) & ThriftServer. So I think we are fine over here concerning the memory commitments.

core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
jatsakthi: Short answer: Yes
Long answer: As we discussed Josh, CM doesn't have an issue with having just one Hue Server neither does Hue have any problem bringing up a single server without a load balancer. But the **Hue Web UI via KNOX** wouldn't be accessible as it depends on the load balancer to be up. So, yes.

jatsakthi:
a recommended setup for high availability is at-least 2 Hue servers with 1 LB. But I think we are fine here with starting off with 1 Hue Server and 1 LB (for Knox). Also, Hue Server & Hue LB can co-exist on the same node. Gateway & leader nodes are fine to have them memory wise. So I think this change is good to go. What do you think @joshelser ? :)

jatsakthi:
Can someone help me here with the merge please?

joshelser:
> So I think we are fine over here concerning the memory commitments.

Thanks for sharing that info!

> So I think this change is good to go.

Yep. I think we're just fine for now. Thanks for the extra work you put in, Sakthi!

PR #8218, opened 2020-06-05
CB-7352 Introduce 7.2.1 SDX templates

akanto:
https://cloudera.atlassian.net/wiki/spaces/ENG/pages/596149312/How+to+introduce+a+new+Cloudera+Runtime+version


keyki:
Do we need a similar file to 7.2.1? https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/resources/cloudera-manager-template/cdh/7.2.json

bergerdenes:
yes, we do need the component list json

akanto:
Yes, we need, that file, but this is only SDX, and for SDX it is not needed.

PR #8220, opened 2020-06-05
CB-7336: Add upscaling FreeIPA for recovery of non-CRL master nodes

jamisonbennett:
When a non-CRL master fails, the recovery should perform a downscale
followed by an upscale.
The upscale process includes:
  * Adding the new cloud instances
  * Validating the cloud instances
  * Retrieving the metadata for the new instances
  * Saving the metadata for the new instances
  * Setting up TLS
  * Bootstrapping salt on the new instances
  * Collecting the new hostnames using salt
  * Saving the new hostnames in the database
  * Installing FreeIPA and replicating from an existing node
  * Updating cluster proxy registration
  * Post installation steps for FreeIPA to update the replication
    topology

Repair was tested with the added tests and it was also tested manually
by repairing a 2 node FreeIPA cluster. The primary gateway was the CA
master & CRL master and that node was kept.

Closes #CB-7336


freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/repair/RepairInstancesRequest.java
```diff
@@ -25,6 +25,9 @@
     @ApiModelProperty(ModelDescriptions.INSTANCE_ID)
     private List<String> instanceIds;
 
+    @ApiModelProperty(ModelDescriptions.FREEIPA_INSTANCE_COUNT_BY_GROUP)
+    private Integer instanceCountByGroup;
```
lacikaaa, line 29: it's a bit odd and confusing for me. So with repair endpoint/request will we be able to scale the size of the FreeIPA cluster? Is it really a repair functionality? also the naming, `instanceCountByGroup` a bit odd, as we don't really specify any group in the request

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/upscale/action/FreeIpaUpscaleActions.java
```diff
@@ -0,0 +1,482 @@
...
+                            context.getCloudContext(), context.getCloudCredential(), context.getCloudStack(), ResourceLists.transform(list));
+                    UpscaleStackResult result = new UpscaleStackResult(request.getResourceId(), ResourceStatus.CREATED, list);
+                    sendEvent(context, result.selector(), result);
+                } else {
+                    Stack updatedStack = instanceMetaDataService.saveInstanceAndGetUpdatedStack(context.getStack(), newInstances);
+                    List<CloudResource> cloudResources = resourceService.findAllByStackId(stack.getId()).stream()
+                            .map(resource -> resourceConverter.convert(resource))
+                            .collect(Collectors.toList());
+                    CloudStack updatedCloudStack = cloudStackConverter.convert(updatedStack);
+                    UpscaleStackRequest request = new UpscaleStackRequest<UpscaleStackResult>(
```
lacikaaa, line 149: should be: `UpscaleStackRequest<UpscaleStackResult> request = new UpscaleStackRequest<>(`

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/upscale/action/FreeIpaUpscaleActions.java
```diff
@@ -0,0 +1,482 @@
...
+    @Bean(name = "UPSCALE_ADD_INSTANCES_STATE")
+    public Action<?, ?> addInstancesAction() {
+        return new AbstractUpscaleAction<>(StackEvent.class) {
+            @Override
+            protected void doExecute(StackContext context, StackEvent payload, Map<Object, Object> variables) {
+                Stack stack = context.getStack();
+                stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.UPSCALE_IN_PROGRESS, "Adding instances");
+
+                List<CloudInstance> newInstances = buildNewInstances(context.getStack(), getInstanceCountByGroup(variables));
+                if (newInstances.isEmpty()) {
```
lacikaaa, line 137: could you refactor the if-else branches to methods?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/upscale/action/FreeIpaUpscaleActions.java
```diff
@@ -0,0 +1,482 @@
...
+
+            private final Set<InstanceStatus> unusedInstanceStatuses = Set.of(InstanceStatus.CREATE_REQUESTED, InstanceStatus.CREATED);
+
+            @Override
+            protected void doExecute(StackContext context, StackEvent payload, Map<Object, Object> variables) {
+                Stack stack = context.getStack();
+                stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.UPSCALE_IN_PROGRESS, "Extending metadata");
+
+                List<CloudInstance> allKnownInstances = cloudStackConverter.buildInstances(context.getStack());
+                List<Resource> resources = resourceService.findAllByStackId(context.getStack().getId());
```
lacikaaa, line 260: nit: you could use `stack` instead of `context.getStack()`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaPostInstallService.java
```diff
@@ -65,11 +65,14 @@
...
-    public void postInstallFreeIpa(Long stackId) throws Exception {
-        LOGGER.debug("Performing post-install configuration for stack {}", stackId);
+    public void postInstallFreeIpa(Long stackId, boolean fullPostInstall) throws Exception {
+        LOGGER.debug("Performing post-install configuration for stack {}. {}.", stackId, fullPostInstall ? "Full post install" : "Partial post install");
         Stack stack = stackService.getStackById(stackId);
         FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientForStack(stack);
         freeIpaTopologyService.updateReplicationTopology(stackId, freeIpaClient);
+        if (!fullPostInstall) {
+            return;
+        }
```
lacikaaa, line 75: could you refactor this? like move the rest of the method to a separate one, and put that call in the if.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -65,6 +70,49 @@
...
+                .map(instanceGroup ->
+                    instanceGroup.getInstanceMetaData().stream()
+                            .map(InstanceMetaData::getInstanceId)
+                            .filter(instanceId -> remainingInstanceIds.contains(instanceId))
+                            .count())
+                .filter(count -> count > instanceCountByGroup)
+                .count();
+        if (numInstanceGroupsWithTooMandNodes > 0) {
+            throw new BadRequestException("The instance count by group cannot be reduced during a repair.");
+        }
```
lacikaaa, line 113: this might worth to be refactored into it's own method

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -65,6 +70,49 @@
...
+                .collect(Collectors.toSet());
+        long numInstanceGroupsWithTooMandNodes = stack.getInstanceGroups().stream()
+                .map(instanceGroup ->
+                    instanceGroup.getInstanceMetaData().stream()
+                            .map(InstanceMetaData::getInstanceId)
+                            .filter(instanceId -> remainingInstanceIds.contains(instanceId))
+                            .count())
+                .filter(count -> count > instanceCountByGroup)
+                .count();
+        if (numInstanceGroupsWithTooMandNodes > 0) {
```
lacikaaa, line 111: typo: `Mand -> Many`

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/action/FreeIpaDownscaleActions.java
```diff
@@ -238,8 +239,12 @@ protected void doExecute(StackContext context, StackEvent payload, Map<Object, O
                 Stack stack = context.getStack();
                 SuccessDetails successDetails = new SuccessDetails(stack.getEnvironmentCrn());
                 successDetails.getAdditionalDetails().put("Hosts", getHosts(variables));
-                stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.DOWNSCALE_COMPLETED, "Downscale complete");
-                operationService.completeOperation(stack.getAccountId(), getOperationId(variables), List.of(successDetails), Collections.emptyList());
+                if (isRepair(variables)) {
+                    stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.DOWNSCALE_IN_PROGRESS, "Downscale complete");
```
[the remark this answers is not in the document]
lacikaaa: I think we should still replace this with something more appropriate. I'm open for ideas. My first thoughts are `REPAIR_IN_PROGRESS` or `WAITING_FOR_UPSCALE`

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/upscale/StackUpscaleService.java
```diff
@@ -206,21 +206,12 @@ private void validateResourceResults(StackScalingFlowContext context, Exception
...
-                        privateId, metaData.getInstanceGroupName(), metaData.getInstanceId(), metaData.getInstanceStatus());
-                if (privateId == null) {
-                    continue;
-                }
-                if (privateId > highest) {
-                    highest = privateId;
-                }
-            }
-        }
+        long highest = instanceGroups.stream()
```
lacikaaa, line 209: so this will produce invalid output. if we have highest ID `0` then we will return `0` as first valid id. this should be an `Optional<Long>` and the return look like:
```
return highest.isPresent() ? highest.get() + 1 : 0;
```
also we might have a test for this method to be sure we don't break it in the future

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/upscale/action/FreeIpaUpscaleActions.java
```diff
@@ -0,0 +1,507 @@
...
+                    int remainingInstances = instanceCountByGroup - instanceGroup.getNotDeletedInstanceMetaDataSet().size();
+                    for (long i = 0; i < remainingInstances; ++i) {
+                        newInstances.add(cloudStackConverter.buildInstance(stack, null, instanceGroup,
+                                stack.getStackAuthentication(), privateId++, InstanceStatus.CREATE_REQUESTED));
+                    }
+                }
+                return newInstances;
+            }
+
+            private Long getFirstValidPrivateId(Set<InstanceGroup> instanceGroups) {
```
lacikaaa, line 180: this seems to me to be a duplicate. we might move this method to a separate class. it would make testing easier

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/action/FreeIpaDownscaleActions.java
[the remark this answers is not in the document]
jamisonbennett: I updated downscale/upscale to set the status to REPAIR_IN_PROGRESS when its a repair operation. It will mark it as complete when upscale finishes.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/upscale/action/FreeIpaUpscaleActions.java
jamisonbennett: I agree its a duplicate. Right now, we have cloudbreak core and freeipa versions of InstanceGroup. They don't share common interfaces. This is the same for a bunch of other classes. While it would be nice to refactor this with a shared API, a shared private instance probably isn't too valuable by itself.

PR #8221, opened 2020-06-05
CB-7340: Delete cloudwatch alerts in batches of 100

jamisonbennett:
Batch cloudwatch alarm deletions into batches of 100. Before deleting
    an alarm, check that it exists. If the deletion fails, then throw an
    exception rather than ignoring the error.
    
    This was tested with the unit tests and the skipping of already
    deleted alarms was tested with cloudbreak on AWS.
    
    This uses the AWS permissions cloudwatch:DescribeAlarms which was
    added to the CDP permissions request in cloudbreak 2.21.0. The initial
    cloudwatch alarms permissions cloudwatch:PutMetricAlarm and
    cloudwatch:DeleteAlarms were added in cloudbreak 2.20.0.

Closes #CB-734

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -4,7 +4,9 @@ import static com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts.CLOUDWATC
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


jamisonbennett:
I have currently marked this as DO_NOT_MERGE because this uses the AWS permissions cloudwatch:DescribeAlarms which was added to the CDP permissions request in cloudbreak 2.21.0. The initial cloudwatch alarms permissions cloudwatch:PutMetricAlarm and cloudwatch:DeleteAlarms were added in cloudbreak 2.20.0. I am thinking it would be a good idea to fall back to just deleting the individual alarms if there isn't permissions to describe the alarms (failures for a bulk delete result in no changes being made).

jamisonbennett:
I removed DO_NOT_MERGE since I made the cloudwatch describe optional. If there aren't permissions to describe then the alarms will be deleted one by one.

keyki:
If it fails it is going to set the status to delete failed. Should we honor the force flag to bypass the issue in case of error?

PR #8222, opened 2020-06-05
CB-7057: fix null status reason

tiborpopovics:


orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/OrchestratorBootstrapRunner.java
```diff
@@ -157,9 +157,9 @@ private Boolean checkResult(Boolean success, int retryCount, Exception actualExc
             nodesWithErrors = ((CloudbreakOrchestratorException) actualException).getNodesWithErrors();
         }
         if (retryCount >= maxRetryCount) {
-            throw new CloudbreakOrchestratorTimeoutException(cause, nodesWithErrors, elapsedTimeRounded);
+            throw new CloudbreakOrchestratorTimeoutException(errorMessage, nodesWithErrors, elapsedTimeRounded);
```
keyki, line 160: We should not log the last message if it is null

orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/OrchestratorBootstrapRunner.java
```diff
@@ -143,23 +143,28 @@ private Boolean checkResult(Boolean success, int retryCount, Exception actualExc
...
+        } else {
+            messageTemplate = success == null
+                    ? "Timeout: Orchestrator component failed to finish in %d min(s)"
+                    : "Failed: Orchestrator component went failed in %d min(s)";
         }
-        String messageTemplate = success == null
-                ? "Timeout: Orchestrator component failed to finish in %d min(s), last message: %s"
-                : "Failed: Orchestrator component went failed in %d min(s), message: %s";
-        long elapsedTimeRounded = Math.max(Math.round((double) retryCount * SLEEP_TIME / MS_IN_SEC / SEC_IN_MIN), MINIMUM_DISPLAYED_TIME_IN_MIN);
         String errorMessage = String.format(messageTemplate, elapsedTimeRounded, cause);
```
keyki, line 158: What happens if the template requires 1 param when the actualException is null, but we pass 2 params with the cause to String.format?

tiborpopovics: According to the Java Doc: `If there are more arguments than format specifiers, the extra arguments are ignored.`

keyki: Cool.

PR #8224, opened 2020-06-05
CB-6952 Generate list about UMS right - CB API method pairs

horadla23:
- generate permission needed for API in /info endpoint of the given service
- move classes related to default resource permission checks into a package
- move classes related to filtering list APIs based on permissions into a package


horadla23:
- calling generate method once
- mock/spring boot test for info page

PR #8227, opened 2020-06-05
CB-6720 Fix a case-sensitivity bug in Ranger configs

kmanamcheri:
This is a workaround fix for a CM bug which introduces case-sensitivity for ranger_database_type. Tested with unit-tests.


akanto:
Yout chnage has been merged from 2.21 to 2.23, so we can close this pull

PR #8228, opened 2020-06-05
CB-7217 Add necessary salt stack changes to invoke the database backup restore scripts.

brycederriso:
Invoke PostgreSQL backup and restore salt scripts.
* Add Salt States for:
  * Copying scripts to master node
  * Copying .pgpass for credentials
  * Running backup and restore scripts
* Move PostgreSQL scripts to `disaster_recovery` directory.
* Add Pillar value for backup location

Closes CB-7217

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
+++ b/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
@@ -1,0 +1,3 @@
+disaster_recovery:
+  object_storage_url:
+
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
@@ -1,0 +1,16 @@
+{% set configure_remote_db = salt['pillar.get']('postgres:configure_remote_db', 'None') %}
+
+{% if 'None' != configure_remote_db %}
+
+include:
+  - postgresql.disaster_recovery
+
+backup_postgresql_db:
+  cmd.run:
+    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('platform')}} {{salt['pillar.get']('disaster_recovery:object_storage_url')}} {{salt['pillar.get']('postgres:remote_db_url')}} {{salt['pillar.get']('postgres:remote_db_port')}} {{salt['pillar.get']('postgres:remote_admin')}}
+    - require:
+        - sls: postgresql.disaster_recovery
+
+{%- else %}
+{# Intentionally left blank, we're not handling backup of non-remote (embedded) DBs #}
+{% endif %}
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/init.sls
@@ -1,0 +1,27 @@
+/opt/salt/scripts/backup_db.sh:
+  file.managed:
+    - makedirs: True
+    - mode: 750
+    - source: salt://postgresql/disaster_recovery/scripts/backup_db.sh
+    - template: jinja
+
+/opt/salt/scripts/restore_db.sh:
+  file.managed:
+    - makedirs: True
+    - mode: 750
+    - source: salt://postgresql/disaster_recovery/scripts/restore_db.sh
+    - template: jinja
+
+/opt/salt/postgresql/.pgpass:
+  file.managed:
+    - makedirs: True
+    - mode: 600
+    - source: salt://postgresql/disaster_recovery/.pgpass
+    - template: jinja
+
+set_pgpass_file:
+  environ.setenv:
+    - name: PGPASSFILE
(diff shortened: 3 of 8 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
```diff
@@ -0,0 +1,9 @@
+include:
+  - postgresql.disaster_recovery
```
keyki, line 2: What does this include? I only see init, backup and restore sls files.

keyki: Nvm, I see it's the init.sls that you are including.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/init.sls
```diff
@@ -0,0 +1,27 @@
+/opt/salt/scripts/backup_db.sh:
+  file.managed:
+    - makedirs: True
+    - mode: 755
```
keyki, line 4: Please use `750` for these files.

orchestrator-salt/src/main/resources/salt/salt/top.sls
```diff
@@ -28,6 +28,7 @@ base:
 
   'G@roles:manager_server':
     - postgresql
+    - postgresql.disaster_recovery
```
keyki, line 31: I don't think we need to include it here. Whenever we invoke restore or backup it will include the init.sls and create the scripts.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/restore.sls
```diff
@@ -0,0 +1,9 @@
+include:
+  - postgresql.disaster_recovery
+
+restore_postgresql_db:
+  cmd.run:
+    - name: /opt/salt/scripts/restore_db.sh {{salt['pillar.get']('platform')}} {{salt['pillar.get']('disaster_recovery:object_storage_url')}} {{salt['pillar.get']('postgres:remote_db_url')}} {{salt['pillar.get']('postgres:remote_db_port')}} {{salt['pillar.get']('postgres:remote_admin')}}
```
keyki, line 6: Not all of our clusters use external DB. Many tests use embedded DB where these pillar configs will be missing. We should add a condition to only execute this if it's external DB. See: [example](https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls#L3-L5)

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
```diff
@@ -0,0 +1,9 @@
...
+
```
keyki, line 3: Similarily check if it's external DB, otherwise, this state should not do anything.

keyki:
Can you please squash your commits and follow the commit message pattern as other commits?

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
```diff
@@ -0,0 +1,3 @@
+disaster_recovery:
+  object_storage_url:
```
keyki, line 2: The value of this variable will be passed in a later commit?

brycederriso: Yes, the intent is that we'll pass in a different value for the `object_storage_url` with each backup request. It was my impression that we could provide a different Pillar value per invocation, something like [this](https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/PostgresConfigService.java#L34)

orchestrator-salt/src/main/resources/salt/salt/top.sls
brycederriso: Ok, this `top` file controls what is _initially_ moved to a minion then?
We'll still receive the states when we use Java to run the `state.apply` function, so we don't have to include it here?

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/restore.sls
brycederriso: Should we expect that customers will be running with clusters that are using embedded DBs? 
This may be a question for @kkalvagadda1 as well.

 **Are we performing DR with embedded DBs as well?**

orchestrator-salt/src/main/resources/salt/salt/top.sls
keyki: This top file describes what will run during highstate. We call highstate very frequently to make sure everything is in place. Backup and restore will be called occasionally so we don't have to include in the top file, it will just create the files so it's not much, but it adds some time to the execution.

keyki:
LGTM. Can you squash your changes and follow the comment pattern of the commits and I can merge.

brycederriso:
Thanks @keyki, let me know if my squashed commit follows our conventions.

PR #8229, opened 2020-06-06
CB-7335: Fix intermittent FreeIPA replica installation failure

jamisonbennett:
When FreeIPA HA installed, the primary FreeIPA instance was updating
certificates and restarting services at the same time as the replicas
were installing. This caused the replica installation to fail.

The salt high state was updated to be able to target specific nodes.
This prevents salt from updating the primary node when the replicas
are being installed.

This was tested using the tests and also it was manually tested a few
times.

Closes #CB-7335


keyki:
Closing in favor of #8235

PR #8230, opened 2020-06-08
CDPCP-1941. Sync Azure OID Mappings as part of user sync

aarman-cloudera:
Azure cloud identity information (Azure OID mapping) are now synced from
UMS into all appropriate datalake instances. At high level the
implementation is:

- SDX service has a new endpoint that allows setting azure ranger
configs (azure user and group mappings). The SDX service will make the
necessary CM api call (through cluster proxy) to set the config and
trigger a role refresh.

- The FMS user sync service will pull the Azure Object IDs (cloud
identity mapping) from UMS and call the new SDX endpoint to set the
configs.

Follow on work items include:
- Poll the status of the refresh after triggering it
- Individual user level and group level sync of OIDs
- Moving some of the CM logic into cluster-cm and client-cm

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
@@ -4,6 +4,7 @@ import java.util.List;
 import java.util.Set;
 
 import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
@@ -15,6 +16,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
+import com.sequenceiq.sdx.api.model.SetRangerCloudIdentityMappingRequest;
 import org.springframework.validation.annotation.Validated;
 
 import com.sequenceiq.cloudbreak.jerseyclient.RetryAndMetrics;
@@ -196,4 +198,11 @@ public interface SdxEndpoint {
     @ApiOperation(value = "Get the status of datalake database restore operation", produces = MediaType.APPLICATION_JSON, nickname = "restoreDatabaseStatus")
     SdxDatabaseRestoreStatusResponse getRestoreDatabaseStatusByName(@PathParam("name") String name,
             @QueryParam("operationId") String operationId);
+
+    @POST
+    @Path("/envcrn/{envCrn}/ranger_cloud_identity_mapping")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "set ranger cloud identity mapping", produces = MediaType.APPLICATION_JSON, nickname = "setRangerCloudIdentityMapping")
+    void setRangerCloudIdentityMapping(@PathParam("envCrn") @ValidCrn String envCrn, @NotNull @Valid SetRangerCloudIdentityMappingRequest request);
+
 }
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
@@ -1,0 +1,45 @@
+package com.sequenceiq.sdx.api.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import javax.validation.constraints.NotNull;
+import java.util.Map;
+
+@ApiModel
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SetRangerCloudIdentityMappingRequest {
+
+    @ApiModelProperty
+    @NotNull
+    private Map<String, String> azureUserMapping;
+
+    @ApiModelProperty
+    @NotNull
+    private Map<String, String> azureGroupMapping;
+
+    public Map<String, String> getAzureUserMapping() {
+        return azureUserMapping;
+    }
+
+    public void setAzureUserMapping(Map<String, String> azureUserMapping) {
+        this.azureUserMapping = azureUserMapping;
+    }
+
+    public Map<String, String> getAzureGroupMapping() {
+        return azureGroupMapping;
+    }
+
+    public void setAzureGroupMapping(Map<String, String> azureGroupMapping) {
+        this.azureGroupMapping = azureGroupMapping;
+    }
+
+    @Override
+    public String toString() {
+        return "SetRangerCloudIdentityMappingRequest{" +
+                "azureUserMapping=" + azureUserMapping +
+                ", azureGroupMapping=" + azureGroupMapping +
+                '}';
+    }
+}
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
@@ -1,0 +1,34 @@
+package com.sequenceiq.datalake.cm;
+
+import com.cloudera.api.swagger.client.ApiClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+// TODO Use ClouderaManagerApiClientProvider in client-cm instad of relying on this class
+@Component
+public class ClouderaManagerProxiedClientFactory {
+
+    private static final String API_VERSION_31 = "v31";
+
+    private static final int CM_READ_TIMEOUT_MS =  60 * 1000;
+
+    private static final int CM_CONNECT_TIMEOUT_MS = 60 * 1000;
+
+    @Value("${clusterProxy.url:}")
+    private String clusterProxyUrl;
+
+    private String getClusterProxyCloderaManagerBasePath(String clusterCrn) {
+        return String.format("%s/proxy/%s/cloudera-manager/api/%s", clusterProxyUrl, clusterCrn, API_VERSION_31);
+    }
+
+    public ApiClient getProxiedClouderaManagerClient(String clouderaManagerStackCrn) {
+        ApiClient apiClient = new ApiClient();
+        apiClient.setBasePath(getClusterProxyCloderaManagerBasePath(clouderaManagerStackCrn));
+        apiClient.setConnectTimeout(CM_READ_TIMEOUT_MS);
+        apiClient.getHttpClient().setReadTimeout(CM_CONNECT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        return apiClient;
+    }
+
+}
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
@@ -1,0 +1,124 @@
+package com.sequenceiq.datalake.cm;
+
+import com.cloudera.api.swagger.ClustersResourceApi;
+import com.cloudera.api.swagger.RoleCommandsResourceApi;
+import com.cloudera.api.swagger.RolesResourceApi;
+import com.cloudera.api.swagger.client.ApiClient;
+import com.cloudera.api.swagger.client.ApiException;
+import com.cloudera.api.swagger.model.ApiBulkCommandList;
+import com.cloudera.api.swagger.model.ApiClusterList;
+import com.cloudera.api.swagger.model.ApiCommand;
+import com.cloudera.api.swagger.model.ApiConfig;
+import com.cloudera.api.swagger.model.ApiConfigList;
+import com.cloudera.api.swagger.model.ApiRole;
+import com.cloudera.api.swagger.model.ApiRoleNameList;
+import com.google.common.base.Joiner;
+import com.google.common.base.Joiner.MapJoiner;
+import com.google.common.collect.ImmutableSortedMap;
+import com.google.common.collect.Iterables;
+
+import com.sequenceiq.cloudbreak.cm.client.retry.ClouderaManagerApiFactory;
+import com.sequenceiq.datalake.controller.exception.RangerCloudIdentitySyncException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Component
+public class ClouderaManagerRangerUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClouderaManagerRangerUtil.class);
+
+    private static final String RANGER_SERVICE_NAME = "ranger";
+
+    private static final String RANGER_USER_SYNC_ROLE_TYPE = "RANGER_USERSYNC";
+
+    private static final String AZURE_USER_MAPPING = "ranger_usersync_azure_user_mapping";
+
+    private static final String AZURE_GROUP_MAPPING = "ranger_usersync_azure_group_mapping";
+
+    private static final MapJoiner CLOUD_IDENTITY_CONFIG_MAP_JOINER =
+            Joiner.on(";").withKeyValueSeparator("=");
+
+    @Inject
+    private ClouderaManagerProxiedClientFactory clouderaManagerProxiedClientFactory;
+
+    @Inject
+    private ClouderaManagerApiFactory clouderaManagerApiFactory;
+
+    private String getClusterName(ApiClient client) throws ApiException {
+        ClustersResourceApi clustersResource = clouderaManagerApiFactory.getClustersResourceApi(client);
+        ApiClusterList clusterList = clustersResource.readClusters(null, null);
+        return Iterables.getOnlyElement(clusterList.getItems()).getName();
+    }
+
+    private String getRangerUserSyncRoleName(ApiClient client, String clusterName) throws ApiException {
+        RolesResourceApi rolesResourceApi = clouderaManagerApiFactory.getRolesResourceApi(client);
+        List<ApiRole> apiRoleList =  rolesResourceApi.readRoles(clusterName, RANGER_SERVICE_NAME, null, null)
+                .getItems()
+                .stream()
+                .filter(apiRole -> RANGER_USER_SYNC_ROLE_TYPE.equals(apiRole.getType()))
+                .collect(Collectors.toList());
+        return Iterables.getOnlyElement(apiRoleList).getName();
+    }
+
+    private ApiConfig newCloudIdentityConfig(String configName, Map<String, String> configValues) {
+        // NOTE: We sort the configs first. This isn't really necessary but is more consistent.
+        ImmutableSortedMap<String, String> configValuesSorted = ImmutableSortedMap.copyOf(configValues);
+        ApiConfig config = new ApiConfig();
+        config.setName(configName);
+        config.setValue(CLOUD_IDENTITY_CONFIG_MAP_JOINER.join(configValuesSorted));
+        return config;
+    }
+
+    private void triggerRoleRefresh(ApiClient client, String clusterName, String serviceName, String roleName) {
+        LOGGER.info("Attempting to trigger role refresh on clusterName = {}, serviceName = {}, roleName = {}", clusterName, serviceName, roleName);
+        ApiRoleNameList roleNameList = new ApiRoleNameList();
+        roleNameList.addItemsItem(roleName);
+        RoleCommandsResourceApi roleCommandsResourceApi = clouderaManagerApiFactory.getRoleCommandsResourceApi(client);
+        try {
+            ApiBulkCommandList bulkResponse = roleCommandsResourceApi.refreshCommand(clusterName, serviceName, roleNameList);
+            ApiCommand response = Iterables.getOnlyElement(bulkResponse.getItems());
+            LOGGER.info("ApiCommand response for role refresh = {}", response);
+            if (response != null && (response.getActive() || response.getSuccess())) {
+                LOGGER.info("Successfully triggered role refresh");
+            } else {
+                LOGGER.debug("Failed to trigger role refresh");
+                throw new RangerCloudIdentitySyncException("Role refresh was not successfully trigerred");
+            }
+        } catch (ApiException e) {
+            LOGGER.error("Encountered ApiException on role refresh", e);
+            throw new RangerCloudIdentitySyncException("Encountered ApiException on role refresh", e);
+        }
+    }
+
+    private boolean isCloudIdMappingSupported(RolesResourceApi rolesResourceApi, String clusterName, String rangerUserSyncRole) throws ApiException {
+        ApiConfigList configList = rolesResourceApi.readRoleConfig(clusterName, rangerUserSyncRole, RANGER_SERVICE_NAME, "full");
+        return configList.getItems().stream().map(ApiConfig::getName).anyMatch(configName -> configName.equals(AZURE_USER_MAPPING));
+    }
+
+    public void setAzureCloudIdentityMapping(String stackCrn, Map<String, String> azureUserMapping, Map<String, String> azureGroupMapping) throws ApiException {
+        // NOTE: The necessary configs changed here are only available in CM7.2-1
+        // TODO Skip setting role and trigerring refresh if the configs haven't changed
+        ApiClient client = clouderaManagerProxiedClientFactory.getProxiedClouderaManagerClient(stackCrn);
+        String clusterName = getClusterName(client);
+        String rangerUserSyncRoleName = getRangerUserSyncRoleName(client, clusterName);
+        RolesResourceApi rolesResourceApi = clouderaManagerApiFactory.getRolesResourceApi(client);
+        if (!isCloudIdMappingSupported(rolesResourceApi, clusterName, rangerUserSyncRoleName)) {
+            LOGGER.info("This version of CM does not support cloud identity mapping. Skipping.");
+        } else {
+            ApiConfigList configList = new ApiConfigList();
+            configList.addItemsItem(newCloudIdentityConfig(AZURE_USER_MAPPING, azureUserMapping));
+            configList.addItemsItem(newCloudIdentityConfig(AZURE_GROUP_MAPPING, azureGroupMapping));
+            rolesResourceApi.updateRoleConfig(clusterName, rangerUserSyncRoleName, RANGER_SERVICE_NAME,
+                    "Updating Azure Cloud Identity Mapping through Cloudbreak",
+                    configList);
+            triggerRoleRefresh(client, clusterName, RANGER_SERVICE_NAME, rangerUserSyncRoleName);
+        }
+    }
+
+}
--- a/datalake/src/main/java/com/sequenceiq/datalake/controller/exception/RangerCloudIdentitySyncException.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/controller/exception/RangerCloudIdentitySyncException.java
@@ -1,0 +1,12 @@
+package com.sequenceiq.datalake.controller.exception;
+
+public class RangerCloudIdentitySyncException extends RuntimeException {
+
+    public RangerCloudIdentitySyncException(String errorMessage) {
+        super(errorMessage);
+    }
+
+    public RangerCloudIdentitySyncException(String errorMessage, Throwable err) {
+        super(errorMessage, err);
+    }
+}
--- a/datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
@@ -7,6 +7,9 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.validation.Valid;
 
+import com.sequenceiq.authorization.annotation.InternalOnly;
+import com.sequenceiq.datalake.cm.RangerCloudIdentityService;
+import com.sequenceiq.sdx.api.model.SetRangerCloudIdentityMappingRequest;
 import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.stereotype.Controller;
 
@@ -75,6 +78,9 @@ public class SdxController implements SdxEndpoint {
 
     @Inject
     private SdxDatabaseDrService sdxDatabaseDrService;
+
+    @Inject
+    private RangerCloudIdentityService rangerCloudIdentityService;
 
     @Override
     @CheckPermissionByAccount(action = AuthorizationResourceAction.CREATE_DATALAKE)
@@ -277,4 +283,10 @@ public class SdxController implements SdxEndpoint {
         String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
         SdxCluster sdxCluster = sdxService.getSdxByNameInAccount(userCrn, name);
         return sdxDatabaseDrService.getDatabaseRestoreStatus(sdxCluster, operationId);    }
+
+    @Override
+    @InternalOnly
+    public void setRangerCloudIdentityMapping(String envCrn, SetRangerCloudIdentityMappingRequest request) {
+        rangerCloudIdentityService.setAzureCloudIdentityMapping(envCrn, request.getAzureUserMapping(), request.getAzureGroupMapping());
+    }
 }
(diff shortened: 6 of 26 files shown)


datalake/src/main/java/com/sequenceiq/datalake/cm/CloudIdentityRangerSyncService.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.Map;
+
+@Service
+public class CloudIdentityRangerSyncService {
```
handavid, line 15: nit: I think a better name would be RangerCloudIdentityService

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
```diff
@@ -165,4 +166,10 @@ SdxClusterResponse create(@ValidStackNameFormat @ValidStackNameLength @PathParam
     @ApiOperation(value = "list advertised datalake versions", produces = MediaType.APPLICATION_JSON, nickname = "advertisedruntimes")
     List<AdvertisedRuntime> advertisedRuntimes();
 
+    @POST
+    @Path("/crn/{crn}/ranger_cloud_identity_mapping")
```
handavid, line 170: it looks like you want to do this by env crn instead of cluster crn. You should change the path and the parameter name. see getByEnvCrn method above for comparison.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    private ApiConfig newCloudIdentityConfig(String configName, Map<String, String> configValues) {
+        // NOTE: We sort the configs first. This isn't really necessary but is more consistent.
+        ImmutableSortedMap<String, String> configValuesSorted = ImmutableSortedMap.copyOf(configValues);
+        ApiConfig config = new ApiConfig();
+        config.setName(configName);
+        config.setValue(CLOUD_IDENTITY_CONFIG_MAP_JOINER.join(configValuesSorted));
+        return config;
+    }
+
+    @Retryable(value = RetryException.class, maxAttempts = 5, backoff = @Backoff(delay = 500))
```
handavid, line 79: why is this the only API that needs retrying?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -217,11 +219,26 @@ private void addMemberToInternalTrackingGroup(UsersState.Builder usersStateBuild
...
+        String azureObjectId = Iterables.getOnlyElement(azureCloudIdentities).getCloudIdentityName().getAzureCloudIdentityName().getObjectId();
+        return Optional.of(azureObjectId);
+    }
+
     private FmsUser umsUserToUser(User umsUser) {
         FmsUser fmsUser = new FmsUser();
         fmsUser.withName(umsUser.getWorkloadUsername());
         fmsUser.withFirstName(getOrDefault(umsUser.getFirstName(), "None"));
         fmsUser.withLastName(getOrDefault(umsUser.getLastName(), "None"));
+        fmsUser.withAzureObjectId(getOptionalAzureObjectId(umsUser.getCloudIdentitiesList()));
```
handavid, line 241: The AzureObjectId belongs in the UmsUsersState, not the UsersState because it's information that is provided by the UMS with no matching information from the FreeIPA side. We use equality to compare the FmsUser and FmsGroup models generated by the UMS against those generated by FreeIPA. Adding in asymmetric information will break the equality and the state difference calculation.
see how we handle workload credentials instead. I think that you can pass the List<CloudIdentity> into the handleUser method and store it there. This will also let you only store OIDs for users that have access to the environment instead of all users.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -241,12 +258,14 @@ private FmsUser umsMachineUserToUser(MachineUser umsMachineUser) {
...
         fmsUser.withFirstName("Machine");
         fmsUser.withLastName("User");
+        fmsUser.withAzureObjectId(getOptionalAzureObjectId(umsMachineUser.getCloudIdentitiesList()));
         return fmsUser;
     }
 
     private FmsGroup umsGroupToGroup(Group umsGroup) {
         FmsGroup fmsGroup = new FmsGroup();
         fmsGroup.withName(umsGroup.getGroupName());
+        fmsGroup.withAzureObjectId(getOptionalAzureObjectId(umsGroup.getCloudIdentitiesList()));
```
handavid, line 268: FmsGroups suffer the same issue as FmsUsers.
you'll need to calculate the group mapping somewhere around line 71 and add it into the UmsUsersState somewhere around line 78

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -274,6 +290,30 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
         }
     }
 
+    private void syncAzureObjectIds(Stack stack, UmsUsersState umsUsersState, BiConsumer<String, String> warnings) {
+        String envCrn = stack.getEnvironmentCrn();
+        LOGGER.info("Syncing Azure Object IDs for environment {}", envCrn);
+
+        UsersState userState = umsUsersState.getUsersState();
+        Map<String, String> usersToAzureObjectIdMap = userState.getUsers().stream()
```
handavid, line 298: These maps should just be stored in the UmsUsersState directly.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FmsGroup.java
```diff
@@ -1,20 +1,32 @@
 package com.sequenceiq.freeipa.service.freeipa.user.model;
 
 import java.util.Objects;
+import java.util.Optional;
 
 public class FmsGroup {
 
     private String name;
 
+    private Optional<String> azureObjectId = Optional.empty();
```
handavid, line 10: see comment on UmsUsersStateProvider for why we can't store this info in this model

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FmsUser.java
```diff
@@ -10,6 +11,8 @@
 
     private String lastName;
 
+    private Optional<String> azureObjectId = Optional.empty();
```
handavid, line 14: see comment on UmsUsersStateProvider for why we can't store this info in this model

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Component
+public class ClouderaManagerRangerUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClouderaManagerRangerUtil.class);
+
+    private static final String RANGER_SERVICE_NAME = "ranger";
```
keyki, line 37: Maybe you could include the cluster-cm module and move all the CM related code there and use it from Data Lake.

datalake/src/main/resources/application.yml
```diff
@@ -100,4 +100,7 @@ cb:
     disabled: false
 
 notification:
-  urls: http://localhost:3000/notifications
\ No newline at end of file
+  urls: http://localhost:3000/notifications
+
+clusterProxy:
```
keyki, line 105: Note: we need to add cluster proxy URL in CBD and helm-charts as well

freeipa/src/main/resources/application.yml
```diff
@@ -64,6 +64,9 @@ freeipa:
   environment:
     url: http://localhost:8088
     contextPath: /environmentservice
+  sdx:
```
keyki, line 67: This is a new dependency of FreeIPA, thus we need to introduce it in CBD and helm-charts as well.

datalake/src/main/java/com/sequenceiq/datalake/cm/CloudIdentityRangerSyncService.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Inject
+    private SdxService sdxService;
+
+    public void setAzureCloudIdentityMapping(String environmentCrn, Map<String, String> azureUserMapping, Map<String, String> azureGroupMapping) {
+        LOGGER.info("Setting Azure cloud id mappings for environment = {}", environmentCrn);
+        List<SdxCluster> sdxClusters = sdxService.listSdxByEnvCrn(environmentCrn);
+        if (sdxClusters.isEmpty()) {
+            LOGGER.info("Environment has no datalake clusters to sync");
+        }
```
lacikaaa, line 30: instead of logging when empty we should log sdxclusters we will sync, so the `if` won't be necessary

datalake/src/main/java/com/sequenceiq/datalake/cm/CloudIdentityRangerSyncService.java
```diff
@@ -0,0 +1,41 @@
...
+        sdxClusters.forEach(sdxCluster -> {
+            String stackCrn = sdxCluster.getStackCrn();
+            LOGGER.info("Updating azure cloud id mappings for datalake stack crn = {}, environment = {}", stackCrn, environmentCrn);
+            try {
+                clouderaManagerRangerUtil.setAzureCloudIdentityMapping(stackCrn, azureUserMapping, azureGroupMapping);
+            } catch (ApiException e) {
+                throw new RuntimeException("Encountered api exception", e);
```
lacikaaa, line 37: I think you should use a more specific exception here, and might log it

datalake/src/main/java/com/sequenceiq/datalake/cm/CloudIdentityRangerSyncService.java
```diff
@@ -0,0 +1,41 @@
...
+            }
+        });
```
lacikaaa, line 39: could be refaactored into a separate method

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class ClouderaManagerProxiedClientFactory {
+
+    private static final String API_VERSION_32 = "v32";
+
+    private static final int CM_READ_TIMEOUT_MS =  60 * 1000;
+
+    private static final int CM_CONNECT_TIMEOUT_MS = 20 * 1000;
```
lacikaaa, line 16: why 20sec? I think we use 1 min in CB for this

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import com.cloudera.api.swagger.client.ApiClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class ClouderaManagerProxiedClientFactory {
+
+    private static final String API_VERSION_32 = "v32";
```
lacikaaa, line 12: why this version? in CB we use 31 or 40. I think this should be in sync

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
```diff
@@ -0,0 +1,32 @@
+package com.sequenceiq.datalake.cm;
+
+import com.cloudera.api.swagger.client.ApiClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class ClouderaManagerProxiedClientFactory {
```
lacikaaa, line 10: altogether it would be nice if you could use https://github.com/hortonworks/cloudbreak/blob/master/client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/ClouderaManagerApiClientProvider.java instead of this class

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -0,0 +1,124 @@
...
+        // NOTE: The necessary configs changed here are only available in CM7.2-1
+        // TODO Skip setting role and trigerring refresh if the configs haven't changed
+        ApiClient client = clouderaManagerProxiedClientFactory.getProxiedClouderaManagerClient(stackCrn);
+        String clusterName = getClusterName(client);
+        String rangerUserSyncRoleName = getRangerUserSyncRoleName(client, clusterName);
+        RolesResourceApi rolesResourceApi = clouderaManagerApiFactory.getRolesResourceApi(client);
+        if (!supportsCloudIdMapping(rolesResourceApi, clusterName, rangerUserSyncRoleName)) {
+            LOGGER.info("This version of CM does not support cloud identity mapping. Skipping.");
+            return;
+        }
```
lacikaaa, line 114: please don't use in middle of method return

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
```diff
@@ -239,4 +248,19 @@ public FlowIdentifier stopByCrn(@ResourceCrn String crn) {
         return cdpConfigService.getAdvertisedRuntimes();
     }
 
+    // custom authorization check
+    @Override
+    @DisableCheckPermissions
+    public void setRangerCloudIdentityMapping(String envCrn, SetRangerCloudIdentityMappingRequest request) {
+        checkIsInternalActor();
```
lacikaaa, line 255: isn't already there a general solution for this @horadla23 ?

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
```diff
@@ -239,4 +248,19 @@ public FlowIdentifier stopByCrn(@ResourceCrn String crn) {
         return cdpConfigService.getAdvertisedRuntimes();
     }
 
+    // custom authorization check
```
lacikaaa, line 251: unnecessary comment

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -509,6 +509,12 @@ private void setTagsSafe(SdxClusterRequest sdxClusterRequest, SdxCluster sdxClus
         return sdxClusterRepository.findByAccountIdAndEnvCrnAndDeletedIsNull(accountIdFromCrn, envCrn);
     }
 
+    public List<SdxCluster> listSdxByEnvCrn(String envCrn) {
+        LOGGER.info("Listing SDX clusters by environment crn {}", envCrn);
```
lacikaaa, line 513: this might be debug if we really need it

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
```diff
@@ -0,0 +1,27 @@
+package com.sequenceiq.sdx.api.model;
+
+import java.util.Map;
+
+public class SetRangerCloudIdentityMappingRequest {
```
lacikaaa, line 5: I think some annotations are missing here, like ignore unknown etc

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -0,0 +1,124 @@
...
+        // NOTE: We sort the configs first. This isn't really necessary but is more consistent.
+        ImmutableSortedMap<String, String> configValuesSorted = ImmutableSortedMap.copyOf(configValues);
+        ApiConfig config = new ApiConfig();
+        config.setName(configName);
+        config.setValue(CLOUD_IDENTITY_CONFIG_MAP_JOINER.join(configValuesSorted));
+        return config;
+    }
+
+    @Retryable(value = RetryException.class, maxAttempts = 5, backoff = @Backoff(delay = 500))
+    public void triggerRoleRefresh(ApiClient client, String clusterName, String serviceName, String roleName) {
```
lacikaaa, line 80: this could be private

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -217,11 +219,26 @@ private void addMemberToInternalTrackingGroup(UsersState.Builder usersStateBuild
...
+                .filter(cloudIdentity -> cloudIdentity.getCloudIdentityDomain().hasAzureCloudIdentityDomain())
+                .collect(Collectors.toList());
+        if (azureCloudIdentities.isEmpty()) {
+            return Optional.empty();
+        }
+        if (azureCloudIdentities.size() > 1) {
+            throw new IllegalStateException(String.format("List contains multiple azure cloud identities = %s", cloudIdentities));
+        }
+        String azureObjectId = Iterables.getOnlyElement(azureCloudIdentities).getCloudIdentityName().getAzureCloudIdentityName().getObjectId();
+        return Optional.of(azureObjectId);
```
lacikaaa, line 233: this should go into an else branch, maybe into a separate method

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
keyki: The cm module have these already configured to use proxy and direct connections, I think that should be used instead of creating new CM related code in other modules.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
aarman-cloudera: Refresh specifically might need retrying because another refresh is occurring. Though I've found that there is automatic retry of failed api calls, so I don' think a retry here is necessary anymore. I've removed it.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
aarman-cloudera: Thanks, I've moved cloud identify information into the UmsUserState.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
aarman-cloudera: Do you mind if we look at this in a follow up task? Moving the logic into cluster-cm is straight forward, but pulling in cluster-cm as dependency seems problematic. I ran into dependency conflicts (hibernate, CachingConfig) as well as it requiring additional dependencies (I think it required bringing an implementation of CloudbreakEventService). The  refactoring required might be beyond the scope of this task. I can file a ticket to see if the CM portion can be easily be moved into cluster-cm and client-cm.

datalake/src/main/resources/application.yml
aarman-cloudera: Good point, I'm looking into it.

datalake/src/main/java/com/sequenceiq/datalake/cm/CloudIdentityRangerSyncService.java
aarman-cloudera: I've added a new exception type.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
aarman-cloudera: I think we used 20 seconds in thunderhead repo. I've updated this to 60 sec. As mentioned in previous comment, I think it makes sense to move the logic into other modules as a separate ticket.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
aarman-cloudera: Updated to 31

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
aarman-cloudera: For now both of these fields are required. I think once we add support for other cloud providers we might want to loosen that requirement.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerProxiedClientFactory.java
aarman-cloudera: As mentioned in previous comment, can we look into this as a separate ticket.

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
```diff
@@ -165,4 +166,10 @@ SdxClusterResponse create(@ValidStackNameFormat @ValidStackNameLength @PathParam
     @ApiOperation(value = "list advertised datalake versions", produces = MediaType.APPLICATION_JSON, nickname = "advertisedruntimes")
     List<AdvertisedRuntime> advertisedRuntimes();
 
+    @POST
+    @Path("/crn/{envCrn}/ranger_cloud_identity_mapping")
```
handavid, line 170: See the path on line 68. it looks like
/envcrn/{envCrn}
This should match to distinguish between the resource crn and the environment crn

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -274,6 +292,51 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
...
+            }
+        });
+        return azureObjectIdMap.build();
+    }
+
+    private void syncAzureObjectIds(Stack stack, UmsUsersState umsUsersState, BiConsumer<String, String> warnings) {
+        String envCrn = stack.getEnvironmentCrn();
+        LOGGER.info("Syncing Azure Object IDs for environment {}", envCrn);
+
+        Map<String, String> userToAzureObjectIdMap = getAzureObjectIdMap(umsUsersState.getUserToCloudIdentityMap());
```
handavid, line 324: this adds cloud identities for all users, regardless of if they have access to the environment. Is this desirable? I would think that you would only care about users that have access to the environment. Imagine if an account has 1000 user but the environment only has 100. You would be syncing the other 900 users cloud identities.

You can filter this map by the users that are in the umsUsersState.getUserState().getUsers()

aarman-cloudera: Good point, I've added the filtering part. Also I've moved most of this logic out of UserSyncService and into CloudIdentitySyncService.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -283,6 +290,12 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
                 workloadCredentialService.setWorkloadCredentials(freeIpaClient, umsUsersState.getUsersWorkloadCredentialMap(), warnings::put);
             }
 
+            boolean hasCloudIdentityEntitlement = entitlementService.cloudIdentityMappingEnabled(INTERNAL_ACTOR_CRN, stack.getAccountId());
+            // TODO For now we only sync cloud ids during full sync. We should eventually allow more granular syncs (actor level and group level sync).
+            if (hasCloudIdentityEntitlement && fullSync) {
```
handavid, line 295: nit: if you reverse the order of conditions and inline the entitlement check here, then we can skip the entitlement check if it's not a full sync.

aarman-cloudera: nice one

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
horadla23: yes, there is, last week we introduced `@InternalOnly` annotation, that should be used here instead of `@DisabledCheckPermissions` (I need to update authz readme)

horadla23: and `@InternalOnly` will guarantee that the method can be called only by internal actor

aarman-cloudera: Thanks, I've switched the new annotation.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -0,0 +1,99 @@
...
+            // TODO The SDX endpoint currently sets the config and triggers refresh. The SDX endpoint should also be updated
+            //      to allow polling the status of the refresh.
+            sdxEndpoint.setRangerCloudIdentityMapping(envCrn, setRangerCloudIdentityMappingRequest);
+        } catch (Exception e) {
+            LOGGER.warn("Failed to set cloud identity mapping for environment {}", envCrn, e);
+            warnings.accept(envCrn, "Failed to set cloud identity mapping");
+        }
+    }
+
+    private Map<String, String> getAzureObjectIdMap(Map<String, List<CloudIdentity>> cloudIdentityMapping) {
```
lacikaaa, line 60: some log here and in `getOptionalAzureObjectId` might be useful to investigate if something is not as desired

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private Optional<String> getOptionalAzureObjectId(List<CloudIdentity> cloudIdentities) {
+        List<CloudIdentity> azureCloudIdentities = cloudIdentities.stream()
+                .filter(cloudIdentity -> cloudIdentity.getCloudIdentityDomain().hasAzureCloudIdentityDomain())
+                .collect(Collectors.toList());
+        if (azureCloudIdentities.isEmpty()) {
+            return Optional.empty();
+        } else if (azureCloudIdentities.size() > 1) {
+            throw new IllegalStateException(String.format("List contains multiple azure cloud identities = %s", cloudIdentities));
```
lacikaaa, line 78: shouldn't we handle this exception somewhere? or is it ok if `UserSyncService#internalSynchronizeStack` handles it? 
cc @handavid

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -0,0 +1,123 @@
...
+        return config;
+    }
+
+    public void triggerRoleRefresh(ApiClient client, String clusterName, String serviceName, String roleName) {
+        LOGGER.info("Attempting to trigger role refresh on clusterName = {}, serviceName = {}, roleName = {}", clusterName, serviceName, roleName);
+        ApiRoleNameList roleNameList = new ApiRoleNameList();
+        roleNameList.addItemsItem(roleName);
+        RoleCommandsResourceApi roleCommandsResourceApi = clouderaManagerApiFactory.getRoleCommandsResourceApi(client);
+        try {
+            ApiBulkCommandList bulkResponse = roleCommandsResourceApi.refreshCommand(clusterName, serviceName, roleNameList);
```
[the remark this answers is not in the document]
sodre90: Good point, it might be better to do role refresh in a flow.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
aarman-cloudera: Thanks! (I don't think I saw similar annotations in the other sdx requets)

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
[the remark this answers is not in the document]
aarman-cloudera: This just triggers role refresh, it doesn't wait around for the role refresh to succeed, so FreeIPA won't hang. I somewhat described this in the commit description. So in a follow up commit we'll want to take the ApiCommand response and poll the status of the command. So the async part will come in a follow up commit (I've discussed with Mark and David and we weren't considering using flow, but maybe we should).

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
aarman-cloudera: Sorry thought I made it private, fixed now.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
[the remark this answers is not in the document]
handavid: The actual activity is tracked in CM. I don't think we even need a poller in the datalake service and can just let the FMS poll the datalake service. We can add a poller if we have a use-case where the datalake service needs to push the completion status to another service.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
handavid: good point. I think we can expand the try block up on line 50 to start on line 42 instead. That will catch this exception and register a warning.

aarman-cloudera: I don't think we need to fail super cleanly in this situation. This is a very unexpected scenario.

aarman-cloudera: That said, I've taken David's advice.

keyki:
Should we also add the entitlement to the mock UMS service so we can test this locally?

aarman-cloudera:
> Should we also add the entitlement to the mock UMS service so we can test this locally?

That makes sense, I've updated mock UMS

mock-caas/src/main/java/com/sequenceiq/caas/grpc/service/auth/MockUserManagementService.java
```diff
@@ -531,6 +533,7 @@ public void getAccount(GetAccountRequest request, StreamObserver<GetAccountRespo
                                 .setWorkloadSubdomain(ACCOUNT_SUBDOMAIN)
                                 .addEntitlements(createEntitlement(CDP_AZURE))
                                 .addEntitlements(createEntitlement(CDP_AUTOMATIC_USERSYNC_POLLER))
+                                .addEntitlements(createEntitlement(CDP_CLOUD_IDENTITY_MAPPING))
```
keyki, line 536: I think we should have an env var to enable and disable the entitlement. See a couple of lines above.

datalake/src/main/java/com/sequenceiq/datalake/controller/exception/RangerCloudIdentitySyncException.java
```diff
@@ -0,0 +1,12 @@
+package com.sequenceiq.datalake.controller.exception;
+
+public class RangerCloudIdentitySyncException extends RuntimeException {
```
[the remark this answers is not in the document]
aarman-cloudera: So the semantics of the exception might change once we add polling of the refresh command in follow on commit. So I think we can hold off on an exception mapper until then.

aarman-cloudera:
> some of my comments are not resolved/answered. until they are I would like to keep the review state in changes requested

Sorry just noticed them (they got buried under hidden). I've addressed those comments, please check again.

PR #8245, opened 2020-06-09
CB-7103 Add resource based Auth for DH api

foldik:
__NOTE:__ Should be merged after tried with mow-dev UMS with this change (https://github.infra.cloudera.com/thunderhead/thunderhead/pull/509)


foldik:
Tested with mow-dev ums. Worked properly, only the creator sees the created datahub.

PR #8247, opened 2020-06-09
CB-6619-Optimize-CB-ClusterSync

smaniraju:
1. Introducing CB-Periscope Cluster Sync on-demand instead of polling all CB Clusters every 30 seconds to avoid unnecessary CB DB Load and  periscope processing already synced clusters.
2. Persicope CM Communication is defaulted to use cluster-proxy which is the recommended mechanism for all Cluster Communication since CM IP would not be reachable in all scenarios.

This was tested on private stack for periscope-cb cluster sync and autoscaling scaleup\scaledown actions.

Closes #CB-6619

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
@@ -36,7 +36,6 @@ import com.sequenceiq.periscope.repository.ClusterRepository;
 import com.sequenceiq.periscope.repository.PeriscopeNodeRepository;
 import com.sequenceiq.periscope.service.DateTimeService;
 import com.sequenceiq.periscope.service.PeriscopeMetricService;
-import com.sequenceiq.periscope.service.StackCollectorService;
 import com.sequenceiq.periscope.utils.TimeUtil;
 
 @Service
@@ -70,9 +69,6 @@ public class LeaderElectionService {
     private TransactionService transactionService;
 
     @Inject
-    private StackCollectorService stackCollectorService;
-
-    @Inject
     private DateTimeService dateTimeService;
 
     @Inject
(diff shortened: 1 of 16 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
```diff
@@ -126,7 +114,6 @@ public void leaderElection() {
                     @Override
                     public void run() {
                         try {
-                            stackCollectorService.collectStackDetails();
```
sidseth, line 129: Nit: Remove the reference and import since the stackCollectorService isn't used here.

sidseth: Maybe also remove the entire StackCollectorService class if it is not used elsewhere.

smaniraju: Removed the unused inject.
StackCollectorService and few other unused classes I will remove in a different PR.

smaniraju:
> Looks good. This essentially stops polling CB every 10/30 seconds to list all new clusters, and instead imports a cluster only when a user tries defining a new AutoScale rule, or tries to fetch autoscale rules for a cluster? (i.e. listing autoscale rules will not list clusters until a user tries defining a policy, or pulling rules for the cluster?)
> One possible change (probably not really required) there would be to not import a cluster if there's no AutoScale rules defined, to be consistent.
> 
> * listAutoScaleRules does not list a cluster.
> * If a user queries autoscale rules for a cluster (doesn't define rules), then subsequent list requests will list the cluster with an empty set of rules.

The current behaviour is same i.e.

> * listAutoScaleRules does not list clusters which have never been configured with autoscaling.
> * If a user queries autoscale rules for a cluster (doesn't define rules), then subsequent list requests will list the cluster with an empty set of rules.

PR #8257, opened 2020-06-10
CB-7405: Add a force uninstall and force-join option to FreeIPA replica install

wonderslug:
This does an ipa-server-uninstall to clean up any possible bad server install
components in case of a retry of the replica install.  As well it adds a
--force-join option to the replica install in case the client has already
been installed and registered on the main server.  This effectively makes
the replica install idempotent.

Closes #CB-7405

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
@@ -5,6 +5,8 @@ set -e
 FQDN=$(hostname -f)
 IPADDR=$(hostname -i)
 
+ipa-server-install --unattended --uninstall
+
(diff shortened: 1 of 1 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
```diff
@@ -5,6 +5,8 @@ set -e
 FQDN=$(hostname -f)
 IPADDR=$(hostname -i)
 
+ipa-server-install -U --uninstall
```
lacikaaa, line 8: nit: replace `-U` with `--unattended` so it will be easier to understand what's going on

wonderslug: changed

PR #8258, opened 2020-06-11
CB-6774 Make redbeams database creation idempotent

pkedvessy:
CB-6774 Make redbeams database creation idempotent


pkedvessy:
retest this please

redbeams/src/main/java/com/sequenceiq/redbeams/repository/DatabaseServerConfigRepository.java
```diff
@@ -32,4 +32,6 @@
             @Param("names") Set<String> names,
             @Param("workspaceId") Long workspaceId,
             @Param("environmentId") String environmentId);
+
+    Optional<DatabaseServerConfig> findByClusterCrn(String stackCrn);
```
Bajzathd, line 36: Parameter should be clusterCrn, both here and in the service layer.

pkedvessy: Good catch, they are renamed

PR #8259, opened 2020-06-11
CB-7182 IT fails intermittently

bbihari:


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/freeipa/AbstractFreeIPAAction.java
```diff
@@ -0,0 +1,45 @@
...
+
+public abstract class AbstractFreeIPAAction<U extends AbstractFreeIPATestDto> implements Action<U, FreeIPAClient> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractFreeIPAAction.class);
+
+    @Override
+    public U action(TestContext testContext, U testDto, FreeIPAClient client) throws Exception {
+        int retries = 0;
+        U result = null;
+        while (retries >= testDto.getWaitUtil().getMaxRetry()) {
```
bergerdenes, line 20: should be <= ?

bbihari: sure :)

PR #8260, opened 2020-06-11
DISTX-494-PurgeExistingPeriscopeClusterData

smaniraju:
Periscope database contains Cluster data where in each cluster creator is stored as a duplicate row in clusterpertain table. This mapping has been fixed to associate all clusters created by single user with a single entry in ClusterPertain.

Existing cluster table data needs to be purged to honour this new mapping.

Closes #DISTX-494


smaniraju:
Have tested this sql script on local database with rows in all the 12 tables.

PR #8263, opened 2020-06-11
CB-7246 - Prevent favicon.ico requests for unauthenticated UIs

moresandeep:
### Describe the change you are making here!

The changes are to nginx component. 
With this change unauthenticated favicon.ico requests (automatic requests by browser) to UIs will be denied (with 403 response). Favicon requests will only be allowed when the UI is authenticated i.e. hadoop-jwt cookie is present. This only affects requests going in via Knox and does not affect CM requests. 

**Why is this change needed**
More details can be found in [CB-7246](jira.cloudera.com/browse/CB-7246). In a nutshell this is what is happening (in Chrome browser):

1. Request is made to a UI page (Atlas, CM, Ranger etc.) from CP.
2. In case where there is no authentication KnoxSSO flow is triggered.
3. Before the previous KnoxSSO flow is completed Chrome requests favicon.ico at the root of the server (this is the nginx host).
4. This favicon.ico request rewrite the originalURL parameter in Pac4j cookie causing the UI page to land on favicon icon.
5. This only happens the first time later requests work as expected.

**What the patch does:**
This patch looks for hadoop-jwt cookie in the request, if the cookie is found there is no change to the request flow. If hadoop-jwt cookie is not found and the request is for favicon.ico then the request is blocked by the nginx and 403 response code is returned.

**Side effects:**
So far local testing has not uncovered any side effects. After successful authentication all UIs (Ranger, Atlas, CM etc.) will display their favicons like they used to.

This screenshot demonstrates the initial favicon.ico request is denied but after the UI is authenticated a new request is issues and favicon is displayed properly.

image

### Please include a short description. Check out these examples:
Prevent unauthenticated favicon.ico requests by browsers.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
+++ b/orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
@@ -28,6 +28,14 @@ server {
         rewrite ^([/](?!clouderamanager/).*$) /clouderamanager$1;
     }
 
+    # If hadoop-jwt cookie not set and favicon is requested return 403.
+    if ($cookie_hadoop-jwt !~ .+) {
+        rewrite ^/favicon.ico$ /favicon.403;
+    }
+    location ~ ^/favicon.403$ {
+        return 403;
+    }
+
     location ~ .*/clouderamanager/(.*) {
         proxy_pass         {{ protocol }}://clouderamanager/$1$is_args$args;
         proxy_redirect     http:// https://;


orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
```diff
@@ -38,6 +38,8 @@ server {
         proxy_set_header   Upgrade $http_upgrade;
         proxy_set_header   Connection $connection_upgrade;
         proxy_set_header   Referer {{ protocol }}://$host/$1;
+        # Add Security cookie flags
+        proxy_cookie_path ~(.*) "$1; SameSite=none; secure; httponly";
```
giladwolff, line 42: So cookies MAY have the SameSite set. In that case you should not overwrite it or add it.

moresandeep: That is true @biharitomi is there a way to do this? thanks!

biharitomi: You can test this change on a simple local env of CB which based on your changes. Every created cluster is going to have this config, so you are able to test this change even in VPC which as proxy to complicate the flow.

moresandeep:
@biharitomi @giladwolff I am taking a different approach here this looks a bit better and safer, no SameSite business.

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
```diff
@@ -28,6 +28,14 @@ server {
         rewrite ^([/](?!clouderamanager/).*$) /clouderamanager$1;
     }
 
+    # If hadoop-jwt cookie not set and favicon is requested return 403.
```
giladwolff, line 31: So let me see if I understand what this change is doing. Before this change, the browser made two requests to knox that is fronted by this nginx, one for the actual resource and right afterwards for the favicon.ico. When google changed the default SameSite default the pac4j cookie was not sent by the browser so knox did not know where to redirect the browser to. And then when the second request came it just returned the favicon and that is what you got. 

What this change is doing is returning a 403 for the second request. This means that the user will see a 403 UNAUTHORIZED error message on the screen instead of the favicon, right?

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
```diff
@@ -28,6 +28,14 @@ server {
...
+    if ($cookie_hadoop-jwt !~ .+) {
+        rewrite ^/favicon.ico$ /favicon.403;
+    }
+    location ~ ^/favicon.403$ {
+        return 403;
+    }
```
giladwolff, line 37: Why not just add the location to the first if statement?
if ($cookie_hadoop-jwt .. && location favicon.ico) return 403?

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
moresandeep: Not exactly, browser will request (on its own) favicon.ico for new websites. Now, what happens is when we click the UI link from Control Plane we get redirected to the UI page with a request for the resource (CM, Ranger, Atlas etc.), Chrome will, on its own also request favicon.ico. These two requests would go through full KnoxSSO cycle and for some unknown reason Altus would return back response for favicon.ico (even when favicon.ico was the send request) and drop the request for requested resource. We saw this in the SAML debugging session we had few weeks ago. How this is related to SameSite is honestly a mystery to me :(  

What the fix tries to do is block favicon.ico requests for unauthenticated requests. 
Now since favicon.ico request is initiated by the browser users will not see 403. And upon successful authentication browser will again request favicon.ico and this time it will succeed. This is demonstrated in the screenshot I pasted above.

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
moresandeep: Oh I wish ! quirks of nginx. 
Location is not allowed in `if `and `if` is frowned upon in Location, see [If is Evil ](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
giladwolff: The control plane gets exactly one SAML authentication request from Knox and it's never for favicon. I believe that after you set the samesite property in knox and atlas we could not reproduce the issue any more. That said, @risdenk thinks that there are no issues with SameSite as he is running with the new feature enabled and haven't encountered any problem. 

I would like to get to the bottom of this if possible. Let's chat offline.

giladwolff: After looking at a reproduction with you again, I agree that not triggering the authentication route for favicon requests is the right thing to do. I am not sure that returning a 403 is the right thing as it seems like the browser will never retry the request. But maybe I'm wrong and maybe it's not important. I added @arnebech and @joeheyming as reviewers as they may have more ideas/information re: what to return to the original favicon request.

arnebech:
@giladwolff not sure I have much to add here. I'd say in general I don't think there is a reason to authenticate favicon.ico requests, it could be whitelisted. Perhaps that is not technically pragmatic in this case. 

I think I have seen this before on some spring (?) project, where the backend stores (?) the first unauthenticated hit as the redirect after login (which could be the favicon). It's been a while so I might be mis-remembering this.

joeheyming:
I agree with @arnebech.  favicons are usually unauthenticated resources.

giladwolff:
> I agree with @arnebech. favicons are usually unauthenticated resources.

@moresandeep Is there a way to whitelist favicon instead of returning 403? Why is knox initiating the authentication request for this?

lmccay:
@giladwolff - We have made that change on the Knox side already. But for new clusters of existing releases, we should also address this in nginx so that we don't continue to see it. There is no real way to patch those at the moment. We can always remove it later.

giladwolff:
> @giladwolff - We have made that change on the Knox side already. But for new clusters of existing releases, we should also address this in nginx so that we don't continue to see it. There is no real way to patch those at the moment. We can always remove it later.
So the idea is that this configuration is something we can change on existing clusters but it's hard/impossible to upgrade knox? So maybe this configuration change should be gated only for old knoxes?

moresandeep:
The issue here is favicons that are requested by Chrome browsers, Chrome "assumes" that the site root will have favicon.ico file and requests it immediately on page load. IMO this does not affect functionality and does not affect performance. 

I do agree that favicons should be an unauthenticated resource future releases of Knox will handle that but this is a fix for releases that are in prod right now.

moresandeep:
@biharitomi can you look at the PR and approve it if it looks ok!
thanks!

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/clouderamanager-ssl-user-facing.conf
```diff
@@ -28,6 +28,14 @@ server {
         rewrite ^([/](?!clouderamanager/).*$) /clouderamanager$1;
     }
 
+    # If hadoop-jwt cookie not set and favicon is requested return 403.
+    if ($cookie_hadoop-jwt !~ .+) {
```
biharitomi, line 32: I suspect this rule will be applied when CM is reached directly and not through Knox.
Could you please test this behaviour when NGiNX proxies the request directly to CM?
https://github.com/hortonworks/cloudbreak#how-to-reach-cm-ui-directlynot-through-knox 

The mentioned scenario is mainly for debugging when the cluster could not be installed and may Knox is not running.
Otherwise the PR looks good to me.

moresandeep: @biharitomi thanks! looks like this is not affecting direct CM requests. This is a direct call to CM, notice there is no response from the server (not even 403). If this was going through the patched nginx code we would have seen 403. 
image

When the request goes through Knox we return 403 (nginx patch code)
image

Which makes me conclude this patch works as expected. Let me know your thoughts.

biharitomi: Thanks!

moresandeep:
Thanks @biharitomi!

PR #8267, opened 2020-06-12
CB-5890 Only select private subnets in internal tenant when public is…

Bajzathd:


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/service/subnetselector/SubnetFilterStrategyMultiplePreferPublic.java
```diff
@@ -20,16 +21,16 @@
...
     @Override
-    public SubnetSelectionResult filter(Collection<CloudSubnet> subnets, int azCount) {
-        List<CloudSubnet> result = subnetSelectorService.collectPublicSubnets(subnets);
-        Set<String> uniqueAzs = result.stream().map(e -> e.getAvailabilityZone()).collect(Collectors.toSet());
-        if (uniqueAzs.size() < azCount) {
-            LOGGER.info("There is not enough different AZ in the public subnets which {}, falling back to private subnets: {}",
+    public SubnetSelectionResult filter(Collection<CloudSubnet> subnets, int azCount, boolean internalTenant) {
+        List<CloudSubnet> result = subnetCollectorService.collectPublicSubnets(subnets);
+        Set<String> uniqueAzs = result.stream().map(CloudSubnet::getAvailabilityZone).collect(Collectors.toSet());
+        if (uniqueAzs.size() < azCount && internalTenant) {
```
gergopapi2, line 30: Just a note: in case cloudbreak is provisioning into a cloudera peered network:
1) in case of public network is wanted and there is not enough of them, private's can also be collected (present solution)
2) alternative: for internal tenant change the algorithm to preferPrivate, and delete the private network selection from the preferPublic altogether.

PR #8271, opened 2020-06-12
CB-7277 PR test fails intermittently with 404 when waiting for archived

aszegedi:
@Bajzathd and @afarsang please review


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/environment/EnvironmentTerminationChecker.java
```diff
@@ -35,13 +36,12 @@ public boolean checkStatus(T waitObject) {
...
             }
         } catch (Exception e) {
-            StringBuilder builder = new StringBuilder("Environment termination failed: ")
-                    .append(System.lineSeparator())
-                    .append(e.getMessage())
-                    .append(System.lineSeparator())
-                    .append(e);
-            LOGGER.error(builder.toString());
-            throw new TestFailException(builder.toString());
+            if (e instanceof NotFoundException) {
```
Bajzathd, line 39: Using a dedicated catch block for NotFoundException would be cleaner

PR #8276, opened 2020-06-15
CB-7294 FlowChain ends before next flow actually executed

foldik:


flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
```diff
@@ -145,7 +145,13 @@ private void handle(String key, Payload payload, FlowParameters flowParameters,
                         flow.initialize(contextParams);
                         runningFlows.put(flow, flowChainId);
                         try {
-                            flowLogService.save(flowParameters, flowChainId, key, payload, null, flowConfig.getClass(), flow.getCurrentState());
+                            transactionService.required(() -> {
+                                flowLogService.save(flowParameters, flowChainId, key, payload, null, flowConfig.getClass(), flow.getCurrentState());
+                                if (flowChainId != null) {
+                                    flowChains.removeLastTriggerEvent(flowChainId, flowParameters.getFlowTriggerUserCrn());
+                                }
+                                return null;
```
horadla23, line 153: just a minor thing: as far as i see, there is a `void` version from `required` method, is it possible to use that here?

foldik: Sure

PR #8279, opened 2020-06-15
CB-7395 Revise, restrict and POC Azure minimal permission list

gergopapi2:
Permission set for single RG use-case not fully tested => draft.

Question: given the most likely scenario from a customer is to bring its own RG, network and storage account, we could create a typical set of permissions that does not contain creation and deletion of such entities. What do you think?


cloud-azure/src/main/resources/definitions/azure-role-def-single-rg.json
```diff
@@ -0,0 +1,84 @@
...
+  "NotActions": [],
+  "DataActions": [
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/write",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/delete",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/add/action"
+  ],
+  "NotDataActions": [],
+  "AssignableScopes": [
+    "/subscriptions/{subscriptionId}"
```
pdarvasi, line 82: We should scope this on RG level, shouldn't we?

gergopapi2: Yes, definitely, thank you!

gergopapi2: Hm, but actually I am not sure if we should decrease assignableScope. What if the customer wants to use the same role definition for multiple resource groups? I think the assignableScope can be as wide as possible, whereas the real assignment should be as tight as possible.

pdarvasi: The customers who are asking for single RG feature will most probably scope their credential to have access on RG level, and the AssignableScopes can contain comma-separated list of RG-s

PR #8284, opened 2020-06-15
CB-7215 Add CB service changes to perform database backup/restore

hreeve-cloudera:
Adds new cloudbreak API methods backupDatabaseByName and restoreDatabaseByName.
These methods trigger new flows that will invoke salt on datalake master node
to run the backup_db.sh and restore_db.sh scripts.

Tested with unit tests and by running cloudbreak locally and making new API
calls.


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/StackV4Controller.java
```diff
@@ -197,11 +197,15 @@ public FlowIdentifier updateSaltByName(Long workspaceId, String name) {
 
     @Override
     public BackupV4Response backupDatabaseByName(Long workspaceId, String name, String backupLocation, String backupId) {
-        return null;
+        FlowIdentifier flowIdentifier =
+            stackOperations.backupClusterDatabase(NameOrCrn.ofName(name), workspaceId, backupLocation, backupId);
+        return new BackupV4Response(flowIdentifier);
```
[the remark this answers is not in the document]
kkalvagadda1: @jamisonbennett we are using a similar approach for this.

Here is the flow Datalake-dr service -> datalake Service -> CB Service.

Datalake service returns an operation id that can used to poll the status. Datalake service keeps polling for the stack status and will update the operation status accordingly.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
```diff
@@ -1,5 +1,45 @@
...
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+
```
kkalvagadda1, line 42: Your project setting could be different. That is any the imports are re-ordered. Can you check?

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreStatusService.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Inject
+    private ClusterService clusterService;
+
+    @Inject
+    private StackUpdater stackUpdater;
+
+    public void backupDatabase(long stackId, String backupId) {
+        stackUpdater.updateStackStatus(stackId, DetailedStackStatus.DATABASE_BACKUP_IN_PROGRESS, "Initiating database backup " + backupId);
+        clusterService.updateClusterStatusByStackId(stackId, Status.UPDATE_IN_PROGRESS);
```
kkalvagadda1, line 32: It should be BACKUP_IN_PROGRESS

kkalvagadda1: It applies to both updateClusterStatusByStackId and fireEventAndLog.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
hreeve-cloudera: I've followed the instructions in the cloudbreak README.md to import all the Intellij settings it asked me to, and this is how my IDE re-orders the imports. I can't change it; if I try to move the imports at all it will automatically move them back. My IDE didn't do this before I imported the cloudbreak project settings, so I'm not sure what to do about it now.

gergopapi2: The original authors might not have had those imports, so it could be normal that you will get a reordering.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/backup/AbstractDatabaseBackupAction.java
```diff
@@ -0,0 +1,15 @@
+package com.sequenceiq.cloudbreak.core.flow2.cluster.datalake.dr.backup;
+
+import com.sequenceiq.cloudbreak.common.event.Payload;
+import com.sequenceiq.cloudbreak.core.flow2.cluster.datalake.dr.BackupRestoreContext;
+import com.sequenceiq.flow.core.AbstractAction;
+import com.sequenceiq.flow.core.FlowEvent;
+import com.sequenceiq.flow.core.FlowState;
+
+public abstract class AbstractDatabaseBackupAction<P extends Payload>
```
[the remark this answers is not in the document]
hreeve-cloudera: The code isn't technically redundant, because one has the signature `getFailurePayload(DatabaseBackupSuccess payload...)` and the other has the signature `getFailurePayload(DatabaseBackupFailedEvent payload...)`. I tried replacing them with a generic `getFailurePayload(StackEvent payload...)` method here, but that doesn't work.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/restore/AbstractDatabaseRestoreAction.java
```diff
@@ -0,0 +1,15 @@
+package com.sequenceiq.cloudbreak.core.flow2.cluster.datalake.dr.restore;
+
+import com.sequenceiq.cloudbreak.common.event.Payload;
+import com.sequenceiq.cloudbreak.core.flow2.cluster.datalake.dr.BackupRestoreContext;
+import com.sequenceiq.flow.core.AbstractAction;
+import com.sequenceiq.flow.core.FlowEvent;
+import com.sequenceiq.flow.core.FlowState;
+
+public abstract class AbstractDatabaseRestoreAction<P extends Payload>
```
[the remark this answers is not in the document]
hreeve-cloudera: Same as the other comment about `getFailurePayload()`.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
hreeve-cloudera: My IDE reorders the imports of any file I add new imports to. That's okay if it's reordering it to what the imports are supposed to look like. And I imported the cloudbreak settings so I have to assume it is. But it's not just new imports, it's all imports in the file.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/DetailedStackStatus.java
```diff
@@ -73,7 +73,14 @@
...
     CLUSTER_UPGRADE_FAILED(Status.AVAILABLE),
-    CLUSTER_UPGRADE_FINISHED(Status.AVAILABLE);
+    CLUSTER_UPGRADE_FINISHED(Status.AVAILABLE),
+    // Database backup/restore statuses
+    DATABASE_BACKUP_IN_PROGRESS(Status.BACKUP_IN_PROGRESS),
+    DATABASE_BACKUP_FINISHED(Status.BACKUP_FINISHED),
+    DATABASE_BACKUP_FAILED(Status.BACKUP_FAILED),
+    DATABASE_RESTORE_IN_PROGRESS(Status.RESTORE_IN_PROGRESS),
+    DATABASE_RESTORE_FINISHED(Status.RESTORE_FINISHED),
+    DATABASE_RESTORE_FAILED(Status.RESTORE_FAILED);
```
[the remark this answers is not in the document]
hreeve-cloudera: Status.AVAILABLE has a StatusKind of StatusKind.FINAL. The FINISHED and FAILED enums here also have StatusKind.FINAL, which I believe indicates they're done and will put the cluster back into a state where additional operations can be run. I did run multiple backups on the same cluster as part of my testing, so each one did seem to be getting resolved.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/StackV4Controller.java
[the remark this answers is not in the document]
brycederriso: What is the difference between an _operation_ and a _flow_? Is there a `has-a` or `is-a` relationship between them? Or are they different notions of a thing going on in the datalake?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/HandlerMethods.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static java.util.Collections.singletonMap;
+
+public class HandlerMethods {
+
+    private HandlerMethods() {
+
+    }
+
+    public static SaltConfig createSaltConfig(String location, String backupId, String cloudPlatform) {
```
brycederriso, line 16: I'm of the opinion that correctly handling paths is scary/hard, but this also decently easy to test.

Could you run some unit tests against the `scheme` or `fullLocation` handling?

Refactoring to something like:
```java
public static SaltConfig createSaltConfig(String location, String backupId, String cloudPlatform) {
        String fullLocation = buildFullLocation(location, backupId, cloudPlatform); // test me

        Map<String, SaltPillarProperties> servicePillar = new HashMap<>();
        servicePillar.put("disaster-recovery", new SaltPillarProperties("/postgresql/disaster_recovery.sls",
            singletonMap("disaster_recovery", singletonMap("object_storage_url", fullLocation))));
        return new SaltConfig(servicePillar);
    }

public static String buildFullLocation (String location, String backupId, String cloudPlatform) {
        String fullLocation = location + "/" + backupId + "_database_backup";
        String scheme;
        if ("aws".equalsIgnoreCase(cloudPlatform)) {
            scheme = "s3:/";
        } else if ("azure".equalsIgnoreCase(cloudPlatform)) {
            // TODO verify this is right when azure flow is working
            scheme = "abfs:/";
        } else {
            throw new UnsupportedOperationException("Cloud platform " + cloudPlatform + "not supported for backup/restore");
        }

        if (fullLocation.startsWith("/")) {
            fullLocation = scheme + fullLocation;
        } else {
            fullLocation = scheme + fullLocation.substring(fullLocation.indexOf("//") + 1);
        }
        return fullLocation;
}

Comment (hreeve-cloudera) [reply core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/HandlerMethods.java:16]: I can write some unit tests around that, but am I handling paths incorrectly here? Outside of the requested refactor?

Comment (brycederriso) [reply core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/HandlerMethods.java:16]: I don't think we handle a relative path like `bucket/backups` correctly here:
```java
fullLocation = scheme + fullLocation.substring(fullLocation.indexOf("//") + 1);
```
We'd get `s3:/bucket/backups`, right? I'm not sure how resilient this method needs to be.

To be pedantic, we may want to define the schemes without the slash to follow the URI specification, if we ever do try to use them with the [URI constructor](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#URI-java.lang.String-java.lang.String-java.lang.String-) I think it would blow up.

I'd consider using the `URI` class to at least guarantee that we have a valid URI coming out of this, though it may not be a valid s3/abfs/hdfs uri.

hreeve-cloudera: I was using the URI class originally, but it wasn't doing exactly what I wanted so I changed it. That's no excuse for not digging into what I was doing wrong though :). I've updated the code to use URIs and added unit tests around it.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/StackV4Controller.java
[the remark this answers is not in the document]
lacikaaa: operation(id) was introduced by @handavid in FreeIPA for usersync as it's not running in a flow, and back then flowid wasn't present on the api. So if you are running a flow it should be enough to expose flowid

lacikaaa:
please don't forget to squash your commits. we usually do 1 commit/pr

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/backup/DatabaseBackupHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        DatabaseBackupRequest request = event.getData();
+        Selectable result;
+        Long stackId = request.getResourceId();
+        try {
+            Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+            Cluster cluster = stack.getCluster();
+            InstanceMetaData gatewayInstance = stack.getPrimaryGatewayInstance();
+            GatewayConfig gatewayConfig = gatewayConfigService.getGatewayConfig(stack, gatewayInstance, cluster.getGateway() != null);
+            Set<String> gatewayFQDN = Collections.singleton(gatewayInstance.getDiscoveryFQDN());
+            ExitCriteriaModel noExitModel = ClusterDeletionBasedExitCriteriaModel.nonCancellableModel();
```
lacikaaa, line 66: are we sure about this? what if termination is triggered during backup?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/backup/DatabaseBackupHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            SaltConfig saltConfig = HandlerMethods.createSaltConfig(request.getBackupLocation(), request.getBackupId(), stack.getCloudPlatform());
+            hostOrchestrator.backupDatabase(gatewayConfig, gatewayFQDN, stackUtil.collectReachableNodes(stack), saltConfig, noExitModel);
+
+            result = new DatabaseBackupSuccess(stackId);
+        } catch (Exception e) {
+            LOGGER.info("Database backup event failed", e);
```
lacikaaa, line 72: error/warn should be better level

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/restore/DatabaseRestoreHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        DatabaseRestoreRequest request = event.getData();
+        Selectable result;
+        Long stackId = request.getResourceId();
+        try {
+            Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+            Cluster cluster = stack.getCluster();
+            InstanceMetaData gatewayInstance = stack.getPrimaryGatewayInstance();
+            GatewayConfig gatewayConfig = gatewayConfigService.getGatewayConfig(stack, gatewayInstance, cluster.getGateway() != null);
+            Set<String> gatewayFQDN = Collections.singleton(gatewayInstance.getDiscoveryFQDN());
+            ExitCriteriaModel noExitModel = ClusterDeletionBasedExitCriteriaModel.nonCancellableModel();
```
lacikaaa, line 66: same here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/restore/DatabaseRestoreHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            SaltConfig saltConfig = HandlerMethods.createSaltConfig(request.getBackupLocation(), request.getBackupId(), stack.getCloudPlatform());
+            hostOrchestrator.restoreDatabase(gatewayConfig, gatewayFQDN, stackUtil.collectReachableNodes(stack), saltConfig, noExitModel);
+
+            result = new DatabaseRestoreSuccess(stackId);
+        } catch (Exception e) {
+            LOGGER.info("Database restore event failed", e);
```
lacikaaa, line 72: also the same

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
```diff
@@ -315,6 +316,26 @@ public Stack getStackByCrn(String crn) {
         return stackCommonService.getRetryableFlows(name, workspaceId);
     }
 
+    public FlowIdentifier backupClusterDatabase(@NotNull NameOrCrn nameOrCrn, Long workspaceId, String location, String backupId) {
+        LOGGER.debug("Starting cluster database backup: " + nameOrCrn);
+        if (nameOrCrn.hasName()) {
```
lacikaaa, line 321: why can't this work by crn?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
```diff
@@ -315,6 +316,26 @@ public Stack getStackByCrn(String crn) {
...
+            return databaseBackupRestoreService.backupDatabase(workspaceId, nameOrCrn.getName(), location, backupId);
+        } else {
+            LOGGER.debug("No stack name provided for backup, found: " + nameOrCrn);
+            throw new BadRequestException("Please provide a stack name for backup");
+        }
+    }
+
+    public FlowIdentifier restoreClusterDatabase(@NotNull NameOrCrn nameOrCrn, Long workspaceId, String location, String backupId) {
+        LOGGER.debug("Starting cluster database restore: " + nameOrCrn);
+        if (nameOrCrn.hasName()) {
```
lacikaaa, line 331: same, why is it restricted to name?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -856,6 +856,46 @@ public String name() {
...
 
+    @Override
+    public void backupDatabase(GatewayConfig primaryGateway, Set<String> target, Set<Node> allNodes, SaltConfig saltConfig,
+            ExitCriteriaModel exitModel) throws CloudbreakOrchestratorFailedException {
+        try (SaltConnector sc = createSaltConnector(primaryGateway)) {
+            for (Entry<String, SaltPillarProperties> propertiesEntry : saltConfig.getServicePillarConfig().entrySet()) {
+                OrchestratorBootstrap pillarSave = new PillarSave(sc, Sets.newHashSet(primaryGateway.getPrivateAddress()), propertiesEntry.getValue());
+                Callable<Boolean> saltPillarRunner = saltRunner.runner(pillarSave, exitCriteria, exitModel);
+                saltPillarRunner.call();
+            }
```
lacikaaa, line 867: this part is the same as in `restoreDatabase`, could you refactor it into a separate method?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -856,6 +856,46 @@ public String name() {
...
+
+            StateRunner stateRunner = new StateRunner(target, allNodes, DATABASE_BACKUP);
```
lacikaaa, line 869: as I see the 2 method is the same except this parameter, could you refactor them to reflect this and eliminate duplication?

hreeve-cloudera:
> please don't forget to squash your commits. we usually do 1 commit/pr

I thought that merging in a PR would automatically squash the commits in that branch into a single commit. That's what I was told when I first starting using Github PRs, in the cloud-infra-services project.

hreeve-cloudera:
@lacikaaa I've refactored the code to try and reduce code duplication and move shared logic into the parent class, per your review. Please let me know if I've missed anything.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/restore/DatabaseRestoreActions.java
```diff
@@ -0,0 +1,120 @@
...
+    @Bean(name = "DATABASE_RESTORE_FAILED_STATE")
+    public Action<?, ?> databaseRestoreFailedAction() {
+        return new AbstractDatabaseRestoreAction<>(DatabaseRestoreFailedEvent.class) {
+
+            @Override
+            protected BackupRestoreContext createFlowContext(FlowParameters flowParameters, StateContext<FlowState, FlowEvent> stateContext,
+                    DatabaseRestoreFailedEvent payload) {
+                Flow flow = getFlow(flowParameters.getFlowId());
+                Stack stack = stackService.getById(payload.getResourceId());
+                MDCBuilder.buildMdcContext(stack);
```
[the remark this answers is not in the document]
lacikaaa: I also had to check this. So what I found:
- if you use `ErrorHandlerAwareReactorEventFactory` it will set MdcContext in the headers and flow engine will persist this
- if the above is set then `LogContextAspects` will load it and set it when a flow step is starting
- if we would like to set it, we usually do it in the context creation part, which is in the flow's abstract action in most cases, as it is the same throughout that flow

if I were you I would remove that MDCBuilder line and check the logs, and check them with it, to see if there is any difference

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
lacikaaa: I see, but if you check everything else (except upgrade, which I don't know why) supports both crn and name. I think this should too

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/backup/DatabaseBackupHandler.java
lacikaaa: this `ExitCriteriaModel` is for defining if salt polling/running should be canceled. So it's like if you would like to stop CB to poll the saltorchestrator when the cluster is terminated you should use the `ClusterDeletionBasedExitCriteriaModel`. As you run your salt states in the cluster I think it would make sense to use it, so CB would free up those threads and won't poll until timeout

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+    public SaltConfig createSaltConfig(String location, String backupId, String cloudPlatform) throws URISyntaxException {
+        String fullLocation = buildFullLocation(location, backupId, cloudPlatform);
+
+        Map<String, SaltPillarProperties> servicePillar = new HashMap<>();
+        servicePillar.put("disaster-recovery", new SaltPillarProperties("/postgresql/disaster_recovery.sls",
+            singletonMap("disaster_recovery", singletonMap("object_storage_url", fullLocation))));
+        return new SaltConfig(servicePillar);
+    }
+
+    public String buildFullLocation(String location, String backupId, String cloudPlatform) throws URISyntaxException {
```
lacikaaa, line 25: this should be either private or in a separate class, as it's not called from anywhere else

lacikaaa: backupid is not used

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+        URI uri = new URI(location);
+        String fullLocation;
+        if ("aws".equalsIgnoreCase(cloudPlatform)) {
+            if (uri.getScheme() != null && !uri.getScheme().startsWith("s3")) {
+                throw new UnsupportedOperationException("Incorrect URL scheme for AWS cloud platform: " + location);
+            }
+            fullLocation = "s3://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
+        } else if ("azure".equalsIgnoreCase(cloudPlatform)) {
+            // TODO verify this is right when azure flow is working
```
lacikaaa, line 34: could you do this before we merge this pr?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
lacikaaa: there is a class, `CloudStorageConfigGenerator` maybe it worth checking out if you could use/reuse some parts of it

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+            singletonMap("disaster_recovery", singletonMap("object_storage_url", fullLocation))));
+        return new SaltConfig(servicePillar);
+    }
+
+    public String buildFullLocation(String location, String backupId, String cloudPlatform) throws URISyntaxException {
+        URI uri = new URI(location);
+        String fullLocation;
+        if ("aws".equalsIgnoreCase(cloudPlatform)) {
+            if (uri.getScheme() != null && !uri.getScheme().startsWith("s3")) {
+                throw new UnsupportedOperationException("Incorrect URL scheme for AWS cloud platform: " + location);
```
lacikaaa, line 30: I think validating this here is a bit late. This should be caught when processing the request. Just check how this works for logging/telemetry

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/cluster/dr/restore/DatabaseRestoreRequest.java
```diff
@@ -0,0 +1,15 @@
...
+public class DatabaseRestoreRequest extends BackupRestoreEvent {
+
+    public DatabaseRestoreRequest(Long stackId, String backupLocation, String backupId) {
+        super(stackId, backupLocation, backupId);
+    }
+
+    @Override
+    public String selector() {
+        return "DatabaseRestoreRequest";
+    }
```
lacikaaa, line 14: I won't go over all of your requests, but please fix them all.
So you shouldn't do this. If you check `StackEvent` it has already an implementation which will return with the same by default.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/restore/DatabaseRestoreHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    @Inject
+    private StackService stackService;
+
+    @Inject
+    private StackUtil stackUtil;
+
+    @Override
+    public String selector() {
+        return "DatabaseRestoreRequest";
+    }
```
lacikaaa, line 50: please use `EventSelectorUtil` here also. And fix in other places too if necessary

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/cluster/dr/backup/DatabaseBackupSuccess.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class DatabaseBackupSuccess extends BackupRestoreEvent {
+
+    public DatabaseBackupSuccess(Long stackId) {
+        super(stackId, null, null);
+    }
+
+    @Override
+    public String selector() {
+        return DATABASE_BACKUP_FINISHED_EVENT.event();
```
lacikaaa, line 15: so we usually we do this the other way around. I would omit this part and let `EventSelectorUtil` take care of the selector here.
I would modify `DatabaseBackupEvent` like this:
`DATABASE_BACKUP_FINISHED_EVENT(EventSelectorUtil.selector(DatabaseBackupSuccess.class)),`
could you check the other events to be in line with this pattern?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
hreeve-cloudera: Where is the logging/telemetry logic? This is my first experience in this codebase so I'm not sure exactly what you're referencing.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
hreeve-cloudera: It's called from `BackupRestoreSaltConfigGeneratorTest`. I've made it package local.
I accidentally deleted the part where backupId was used when I refactored this code. It's been added back.
I looked at `CloudStorageConfigGenerator`.  I prefer what I'm doing here, so unless there's a strong reason to change it I'm going to keep what I have.

lacikaaa: at least `"aws"` and `"azure"` should be replaced with `CloudConstants`/`CloudPlatform`

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
lacikaaa: so the validation should be done around `DatabaseBackupRestoreService#backupDatabase/restoreDatabase` and the flow shouldn't be started with invalid input
if you check `LoggingBase` it has separate input for AWS/Azure and those have validators, so it's checked on request time

hreeve-cloudera: I've added a `BackupRestoreV4RequestValidator` class and unit tests for it. It checks the location scheme, and also makes sure that required parameters are set.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/HandlerMethods.java
```diff
@@ -0,0 +1,46 @@
...
+            singletonMap("disaster_recovery", singletonMap("object_storage_url", fullLocation))));
+        return new SaltConfig(servicePillar);
+    }
+
+    public static String buildFullLocation(String location, String backupId, String cloudPlatform) throws URISyntaxException {
+        URI uri = new URI(location);
+        String fullLocation;
+        if ("aws".equalsIgnoreCase(cloudPlatform)) {
+            if (uri.getScheme() != null && !uri.getScheme().startsWith("s3")) {
+                throw new UnsupportedOperationException("Incorrect URL scheme for AWS cloud platform: " + location);
```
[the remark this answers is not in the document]
hreeve-cloudera: This is now being handled by `BackupRestoreV4RequestValidator`.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreContext.java
```diff
@@ -0,0 +1,48 @@
...
+import com.sequenceiq.flow.core.CommonContext;
+import com.sequenceiq.flow.core.FlowParameters;
+
+public class BackupRestoreContext extends CommonContext {
+
+    private Long stackId;
+
+    private String backupLocation;
+
+    private String backupId;
```
lacikaaa, line 13: these should be final

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/cluster/dr/BackupRestoreEvent.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.sequenceiq.cloudbreak.common.event.AcceptResult;
+import com.sequenceiq.cloudbreak.reactor.api.event.StackEvent;
+import reactor.rx.Promise;
+
+public class BackupRestoreEvent extends StackEvent {
+
+    private String backupLocation;
+
+    private String backupId;
```
lacikaaa, line 11: should be final

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/backup/DatabaseBackupHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    protected void doAccept(HandlerEvent event) {
+        LOGGER.debug("Accepting Database backup event...");
+        DatabaseBackupRequest request = event.getData();
+        Selectable result;
+        Long stackId = request.getResourceId();
+        try {
+            Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+            Cluster cluster = stack.getCluster();
+            InstanceMetaData gatewayInstance = stack.getPrimaryGatewayInstance();
+            GatewayConfig gatewayConfig = gatewayConfigService.getGatewayConfig(stack, gatewayInstance, cluster.getGateway() != null);
```
lacikaaa, line 68: please use `Cluster#hasGateway`

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/restore/DatabaseRestoreHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    protected void doAccept(HandlerEvent event) {
+        LOGGER.debug("Accepting Database restore event...");
+        DatabaseRestoreRequest request = event.getData();
+        Selectable result;
+        Long stackId = request.getResourceId();
+        try {
+            Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+            Cluster cluster = stack.getCluster();
+            InstanceMetaData gatewayInstance = stack.getPrimaryGatewayInstance();
+            GatewayConfig gatewayConfig = gatewayConfigService.getGatewayConfig(stack, gatewayInstance, cluster.getGateway() != null);
```
lacikaaa, line 68: please use `Cluster#hasGateway`

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/backup/DatabaseBackupHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    protected Selectable defaultFailureEvent(Long resourceId, Exception e) {
+        return new DatabaseBackupFailedEvent(resourceId, e, DetailedStackStatus.DATABASE_BACKUP_FAILED);
+    }
+
+    @Override
+    protected void doAccept(HandlerEvent event) {
+        LOGGER.debug("Accepting Database backup event...");
```
lacikaaa, line 60: this is strange. we should log de event or event data here to make it a bit meaningful. same for the other handler
also for receiving the event there should be already a log provided by `LogContextAspects`

core/src/main/java/com/sequenceiq/cloudbreak/service/DatabaseBackupRestoreService.java
```diff
@@ -0,0 +1,83 @@
...
+            String errorMsg = String.format("Database backup cannot be performed because there is an active flow running: %s",
+                flowLogs.stream().map(FlowLog::toString));
+            eventService.fireCloudbreakEvent(
+                stack.getId(),
+                DATALAKE_DATABASE_BACKUP,
+                ResourceEvent.DATALAKE_DATABASE_BACKUP_COULD_NOT_START,
+                List.of(errorMsg));
+            throw new BadRequestException(errorMsg);
+        } else {
+            return flowManager.triggerDatalakeDatabaseBackup(stack.getId(), location, backupId);
```
lacikaaa, line 63: it would be nice to have a line of log here after MDCContext is build. same for restore method

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -934,6 +934,37 @@ public String name() {
...
+                Callable<Boolean> saltPillarRunner = saltRunner.runner(pillarSave, exitCriteria, exitModel, maxRetry, true);
+                saltPillarRunner.call();
+            }
+
+            StateRunner stateRunner = new StateRunner(target, allNodes, state);
+            OrchestratorBootstrap saltJobIdTracker = new SaltJobIdTracker(sc, stateRunner);
+            Callable<Boolean> saltJobRunBootstrapRunner = saltRunner.runner(saltJobIdTracker, exitCriteria, exitModel, maxRetry, true);
+            saltJobRunBootstrapRunner.call();
+        } catch (Exception e) {
+            LOGGER.info("Error occurred during database backup/restore", e);
```
lacikaaa, line 963: this should be error level, shouldn't it?

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
```diff
@@ -141,8 +138,10 @@ protected void executeInternal(JobExecutionContext context) throws JobExecutionE
                 Status.MAINTENANCE_MODE_ENABLED,
                 Status.EXTERNAL_DATABASE_CREATION_IN_PROGRESS,
                 Status.BACKUP_IN_PROGRESS,
+                Status.BACKUP_FINISHED,
                 Status.BACKUP_FAILED,
                 Status.RESTORE_IN_PROGRESS,
+                Status.RESTORE_FINISHED,
                 Status.RESTORE_FAILED
```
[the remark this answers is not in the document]
hreeve-cloudera: It shouldn't, no.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+
+    String buildFullLocation(String location, String backupId, String cloudPlatform) throws URISyntaxException {
+        URI uri = new URI(location);
+        String suffix = '/' + backupId + "_database_backup";
+        String fullLocation;
+        if (AWS.equalsIgnoreCase(cloudPlatform)) {
+            fullLocation = "s3://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
+        } else {
+            fullLocation = "abfs://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
+        }
```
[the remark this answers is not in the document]
hreeve-cloudera: So I originally had an `else if`, and then a final `else` that threw an `UnsupportedOperationException`. You said I shouldn't be throwing an `UnsupportedOperationException`, but you didn't answer when I asked what I should be doing instead. So instead I made the `BackupRestoreV4RequestValidator`, which validates the scheme is either AWS or AZURE before the flow gets here, and removed the `UnsupportedOperationException`. Are you saying I should add it back? Is there some other exception type I should be using?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -81,8 +48,37 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
+
+import static com.sequenceiq.cloudbreak.common.type.OrchestratorConstants.SALT;
+import static com.sequenceiq.cloudbreak.common.type.RecipeExecutionPhase.PRE_CLOUDERA_MANAGER_START;
+import static com.sequenceiq.cloudbreak.common.type.RecipeExecutionPhase.convert;
+import static com.sequenceiq.cloudbreak.util.FileReaderUtils.readFileFromClasspath;
```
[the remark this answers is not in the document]
hreeve-cloudera: I've loaded the config as instructed in the cloudbreak README at least twice. I'm not sure what else to do at this point.

core/src/main/java/com/sequenceiq/cloudbreak/service/DatabaseBackupRestoreService.java
```diff
@@ -0,0 +1,83 @@
...
+    public void validate(Long workspaceId, NameOrCrn nameOrCrn, String location, String backupId) {
+        Stack stack = stackService.getByNameOrCrnInWorkspace(nameOrCrn, workspaceId);
+        ValidationResult validationResult = requestValidator.validate(stack, location, backupId);
+        if (validationResult.hasError()) {
+            throw new BadRequestException(validationResult.getFormattedErrors());
+        }
+    }
+
+    public FlowIdentifier backupDatabase(Long workspaceId, NameOrCrn nameOrCrn, String location, String backupId) {
+        Stack stack = stackService.getByNameOrCrnInWorkspace(nameOrCrn, workspaceId);
```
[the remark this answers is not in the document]
kkalvagadda1: will do

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/AbstractBackupRestoreActions.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Override
+    protected BackupRestoreContext createFlowContext(FlowParameters flowParameters, StateContext<FlowState, FlowEvent> stateContext,
+            P payload) {
+        return BackupRestoreContext.from(flowParameters, payload, payload.getBackupLocation(), payload.getBackupId());
+    }
+
+    @Override
+    protected Object getFailurePayload(P payload, Optional<BackupRestoreContext> flowContext, Exception ex) {
+        return null;
```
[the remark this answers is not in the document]
hreeve-cloudera: Used StackFailureEvent as the default here instead of null.

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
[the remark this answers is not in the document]
hreeve-cloudera: I've gone ahead and removed the FINISHED enums. If @foldik or @bbihari have any other suggestions I'll make them later.

hreeve-cloudera:
> so 2 things:
> 
>     * you might want to add your flows to `CloudbreakFlowInformation#RESTARTABLE_FLOWS` so if your flows fails they can be retried

Done

>     * I don't know this for sure but I see that you don't stop hive and ranger during backup/restore operation, and you dump/restore the databases sequentially. This makes it possible that they drift apart, like you're backing up hive, it will be consistent, but in the meanwhile something new is added which is not included in the backup. You create a new policy/whatever in ranger for that. Then you backup ranger. You will have a ranger backup which will point to something in hive which doesn't exist. I'm not so familiar with these services, so I'm not sure this scenario is valid or not.

That's an overarching decision decision for phase one of the the disaster recovery service, and it's not just limited to the ranger and hive database here. This PR isn't meant to handle or resolve that particular case, so we accept it's possible things can get out of sync. That's a longer term issue for the next phase.

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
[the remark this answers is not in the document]
bbihari: Every new status enum should be categorized as syncable/ignored/unschedulable. If it's not "syncable" then that state should be an end-state like terminated.

The FAILED states should be handled in SdxClusterStatusCheckerJob too.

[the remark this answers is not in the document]
hreeve-cloudera: Handled in what way? It looks like only a handful of states from `Status` are explicitly handled in `SdxClusterStatusCheckerJob`.

hreeve-cloudera:
Put a DO NOT MERGE flag on while discussing the end-to-end testing with @kkalvagadda1, to make sure it still looked good after the most recent changes. It does, so branch is now okay to merge.

@lacikaaa would you mind merging this in for me?

PR #8288, opened 2020-06-16
CB-7225 Usersync and setpassword APIs should authorized in case of empty list too

horadla23:
In case of freeipa usersync and setPassword, environments list should not be empty.
Need to wait for UI and CLI change.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
@@ -3,7 +3,6 @@ package com.sequenceiq.freeipa.service.freeipa.user;
 import static com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient.INTERNAL_ACTOR_CRN;
 import static java.util.Objects.requireNonNull;
 
-import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


horadla23:
> change looks fine, but we'll need to make sure the UI and the environments2 API to fully populate the environments list instead of sending an empty list.

Yes, that's why I put the do_not_merge label on this PR for now

horadla23:
@handavid according to UMS guys, we cannot add custom logic in cdp-cli, so we changed the task for this and I changed the PR, in the way that we check permissions in service layer (not with our framework), to allow interfaces to call these with empty list and also make authorization possible in this case also (meaning that we still need to authorization related environments crns), so can you (and @foldik ) please review this once again?

horadla23:
tested locally with mow-dev UMS, sync and setPassword calls checked describeEnvironment right for affected environments successfully

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -102,15 +104,25 @@
...
+            Set<String> userCrnFilter, Set<String> machineUserCrnFilter, AuthorizationResourceAction action) {
+        List<Stack> stacks = getStacksForSync(accountId, actorCrn, environmentCrnFilter, userCrnFilter, machineUserCrnFilter);
+        List<String> relatedEnvironmentCrns = stacks.stream().map(stack -> stack.getEnvironmentCrn()).collect(Collectors.toList());
+        commonPermissionCheckingUtils.checkPermissionForUserOnResources(action, actorCrn, relatedEnvironmentCrns);
+        return performSyncForStacks(accountId, actorCrn, environmentCrnFilter, userCrnFilter, machineUserCrnFilter, stacks);
+    }
 
-        List<Stack> stacks = stackService.getMultipleByEnvironmentCrnOrChildEnvironmantCrnAndAccountId(environmentCrnFilter, accountId);
-        LOGGER.debug("Found {} stacks", stacks.size());
+    private Operation performSyncForStacks(String accountId, String actorCrn, Set<String> environmentCrnFilter,
```
handavid, line 124: I don't think we need the environmentCrnFilter here because we've already gotten the relevant stacks.

horadla23: ok!

horadla23: fixed

PR #8301, opened 2020-06-16
CB-7512 ParcelValidation should happen in integration test

doktoric:


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/ParcelAccessibilityTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private UtilTestClient utilTestClient;
+
+    protected void setupTest(TestContext testContext) {
+        createDefaultUser(testContext);
+        initializeDefaultBlueprints(testContext);
+    }
+
+    @Test(dataProvider = TEST_CONTEXT_WITH_MOCK)
+    @Description(
+            given = "there is a running Cloudbreak",
```
bergerdenes, line 42: wrong description!

PR #8302, opened 2020-06-16
CB-7479: Update the service name for DEX

arunmahadevan:
Update the CRN service name for DEX to match with the thunderhead
service name (defined in Crn.java in the thunderhead repo).

Refer: https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/libs/protocols/src/main/java/com/cloudera/thunderhead/service/common/crn/Crn.java

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
@@ -159,7 +159,12 @@ public class Crn {
         FREEIPA("freeipa", NON_ADMIN_SERVICE),
         DATAHUB("datahub", NON_ADMIN_SERVICE),
         DATALAKE("datalake", NON_ADMIN_SERVICE),
+        /**
+         * @deprecated {@link #DEX} was replaced by {@link #DE} and is kept here for backward compatibility.
+         */
+        @Deprecated
         DEX("dex", NON_ADMIN_SERVICE),
+        DE("de", NON_ADMIN_SERVICE),
         ACCOUNTTAG("accounttag", NON_ADMIN_SERVICE),
         ACCOUNTTELEMETRY("accounttelemetry", NON_ADMIN_SERVICE),
         ML("ml", NON_ADMIN_SERVICE);


arunmahadevan:
I assume this is the only place service name needs to be updated for DEX to interact with FreeIPA for the keytab.

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
```diff
@@ -159,7 +159,7 @@ public static Partition safeFromString(String input) {
         FREEIPA("freeipa", NON_ADMIN_SERVICE),
         DATAHUB("datahub", NON_ADMIN_SERVICE),
         DATALAKE("datalake", NON_ADMIN_SERVICE),
-        DEX("dex", NON_ADMIN_SERVICE),
+        DE("de", NON_ADMIN_SERVICE),
```
jamisonbennett, line 162: I have concerns about backwards compatibility. This will mean that you will need to synchronize the change for our internal accounts (I know DEX is not yet used on prod). I think it might be simpler to coordinate to add DE now and then remove DEX after everything has been moved to DE. I will let the DEX team decide what they want to do, since they know better than me if they are currently using DEX anywhere or if this change is to fix something which was already updated on the DEX side.

arunmahadevan: @jamisonbennett good point. Yes this would mean we need to recreate any existing environments people may be using esp the ones used by QE team for testing. I can add DE as a new option here so that existing flows would continue to work. Once everyone migrates over raise another patch to remove DEX.

arunmahadevan:
Added DE as a new option. @jamisonbennett could you take another look?

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
```diff
@@ -159,7 +159,8 @@ public static Partition safeFromString(String input) {
         FREEIPA("freeipa", NON_ADMIN_SERVICE),
         DATAHUB("datahub", NON_ADMIN_SERVICE),
         DATALAKE("datalake", NON_ADMIN_SERVICE),
-        DEX("dex", NON_ADMIN_SERVICE),
+        DEX("dex", NON_ADMIN_SERVICE), // TODO: remove this once all DEX enviroments migrate over to DE
```
kmanamcheri, line 162: @arunmahadevan Checkstyle is failing with "Don't use trailing comments." You should follow the style we use for deprecating these. Refer to "SDX" above.

```
/**
  * @deprecated {@link #DEX} was replaced by {@link #DE} and is kept here for backwards compatibility.
  */
@Deprecated
DEX("dex", NON_ADMIN_SERVICE)
```

arunmahadevan: Updated.

PR #8306, opened 2020-06-17
CB-6329: Add recovery of the FreeIPA primary gateway

jamisonbennett:
Allow the FreeIPA primary gateway to be recovered.

When recovering the primary gateway, the primary gateway is switched
to a working instance. Then the old primary gateway is removed by
downscaling. A new FreeIPA master is created during upscale. Then the
primary gateway is switched back to the FreeIPA master.

The unused FreeIPA core instance metadata type was removed.

This was tested using the unit tests and it was manually tested using
a local instance of cloudbreak.

Closes #CB-6329


jamisonbennett:
I am thinking of adding some integration tests for repair. I will probably do that tomorrow.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/event/ChangePrimaryGatewayEvent.java
```diff
@@ -0,0 +1,32 @@
+package com.sequenceiq.freeipa.flow.freeipa.repair.changeprimarygw.event;
+
+import java.util.List;
+
+import com.sequenceiq.freeipa.flow.stack.StackEvent;
+
+public class ChangePrimaryGatewayEvent extends StackEvent {
```
lacikaaa, line 7: could you add `toString`?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_promote_replica_to_master.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+FQDN=$(hostname -f)
+
+echo "$FPW" | kinit $ADMIN_USER
+CURRENT_CA_MASTER=`ipa config-show | grep "CA renewal master" | cut -d":" -f2 | cut -d" " -f2`
```
lacikaaa, line 8: we should echo this value to be able to see in minion logs

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -520,6 +553,31 @@ public void installFreeIPA(GatewayConfig primaryGateway, List<GatewayConfig> all
         }
     }
 
+    @Override
+    public Optional<String> getFreeIpaMasterHostname(GatewayConfig primaryGateway) throws CloudbreakOrchestratorException {
```
lacikaaa, line 557: why do we have to call salt to find this out? shouldn't be this info available on java side?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -489,22 +497,47 @@ public void changePrimaryGateway(GatewayConfig formerGateway, GatewayConfig newP
     @Override
     public void installFreeIPA(GatewayConfig primaryGateway, List<GatewayConfig> allGatewayConfigs, Set<Node> allNodes,
```
lacikaaa, line 498: this method got really big, could you refactor to smaller bits

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
jamisonbennett: In a normal situation, the FreeIPA primary gateway is the FreeIPA master. The primary gateway is stored in the FreeIPA database. During a repair of the primary gateway or after a failed repair of the primary gateway, it is likely that the FreeIPA primary gateway is a different node. As far as I can tell, FreeIPA doesn't store the salt roles in the database. I need to ensure that only a single node is identified as the primary FreeIPA instance (if multiple instances generate the CRLs it will be a problem), so I am querying FreeIPA.

If we add the FreeIPA roles into the database, this will be one more thing to synchronize and it will only be used during repair anyway.

jamisonbennett:
I also added the FreeIPA HA repair integration test to on-integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/freeipa/FreeIPATests.java.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -489,22 +497,25 @@ public void changePrimaryGateway(GatewayConfig formerGateway, GatewayConfig newP
...
-            saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(primaryServerHostname, allNodes, "freeipa_primary"), exitCriteriaModel, exitCriteria);
-            runNewService(sc, new HighStateRunner(primaryServerHostname, allNodes), exitCriteriaModel);
-
-            LOGGER.debug("Set replica FreeIPA: {}", replicaServersHostnames);
-            if (!replicaServersHostnames.isEmpty()) {
-                saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(replicaServersHostnames, allNodes, "freeipa_replica"), exitCriteriaModel, exitCriteria);
-                runNewService(sc, new HighStateRunner(replicaServersHostnames, allNodes), exitCriteriaModel);
-            }
+
+            installFreeIPAUpdateExistingReplicas(sc, existingFreeIpaReplaceHostnames, allNodes, exitCriteriaModel);
```
lacikaaa, line 511: so I'm just dropping it here, but I noticed somewhere else also. We used everywhere `FreeIpa` until now (at least I hope so), and not `FreeIPA`. Could you rename methods, classes to match this?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -520,6 +531,72 @@ public void installFreeIPA(GatewayConfig primaryGateway, List<GatewayConfig> all
...
+            ExitCriteriaModel exitCriteriaModel) throws Exception {
+        LOGGER.debug("Exsting Replica FreeIPAs: [{}]", existingFreeIpaReplaceHostnames);
+        // The existing replicas need to be serialized into high state. See the comments in CB-7335 for more details.
+        for (String existingReplicaHostname : existingFreeIpaReplaceHostnames) {
+            LOGGER.debug("Applying changes to FreeIPA replica {}", existingReplicaHostname);
+            runNewService(sc, new HighStateRunner(Set.of(existingReplicaHostname), allNodes), exitCriteriaModel);
+        }
+    }
+
+    private void installFreeIPAReplacas(SaltConnector sc, Set<String> newFreeIpaReplaceHostnames, Set<Node> allNodes, ExitCriteriaModel exitCriteriaModel)
```
lacikaaa, line 544: typo: `installFreeIPAReplacas`

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
jamisonbennett: I fixed this one....but I will put the others in a separate PR. There are quite a few capitalization changes (19 files and their associated classes).

PR #8308, opened 2020-06-17
CB-7461 save format and mount logs to cloud storage

sodre90:


oleewere:
@sodre90 is that only required. for datalake/datahub clusters? if these can be used for freeipa,inputs  can be extended with these as well

PR #8309, opened 2020-06-17
CB-4925 - Setting auth.mock.freeipadlebsencryption.enable=false by default

attilapalfi92:
It's a possible trap for all developers who just want to setup an environment. For most devs encryption is not needed and they won't even know that it's enabled and slowing their work down. Who needs encryption, already knows this feature exists and can set the value to true in their dev env.


lajosrodek:
Follow-up to #7833 .

PR #8314, opened 2020-06-17
CB-7522 Beautify configuration entry for audit endpoint

bergerdenes:


audit-connector/src/main/java/com/sequenceiq/cloudbreak/audit/config/AuditConfig.java
```diff
@@ -1,28 +1,42 @@
...
     private int port;
 
-    public String getEndpoint() {
-        return endpoint;
+    @PostConstruct
+    public void init() {
+        if (isConfigured()) {
+            String[] parts = endpoint.split(":");
+            if (parts.length != 2) {
+                throw new IllegalArgumentException("altus.audit.endpoint must be in host:port format.");
```
topolyai5, line 24: much better if the default port is 80 or 443 depending on HTTP or https

bergerdenes: its not http anyway

PR #8339, opened 2020-06-19
CB-7400 Fix template builder to set skuTier based on skuName if not specified

foldik:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTemplateBuilder.java
```diff
@@ -152,6 +154,30 @@ public String build(CloudContext cloudContext, DatabaseStack databaseStack) {
...
+        } else {
+            return azureDatabaseServerView.getSkuTier();
+        }
+    }
+
+    private String getSkuTierFromSkuName(String skuName) {
+        if (Set.of("GP_Gen5_2", "GP_Gen5_4", "GP_Gen5_8", "GP_Gen5_16", "GP_Gen5_32").contains(skuName)) {
+            return "GeneralPurpose";
+        } else if (Set.of("MO_Gen5_2", "MO_Gen5_4", "MO_Gen5_8", "MO_Gen5_16", "MO_Gen5_32").contains(skuName)) {
+            return "MemoryOptimized";
```
lacikaaa, line 175: these values might be better in private static fields

foldik: Do you mean the instance types as well or just the two skuTiers (GeneralPurpose, MemoryOptimized)?

lacikaaa: yes, all of them

PR #8340, opened 2020-06-19
CB-7365 Extract magic string literals in AwsInstanceView

lajosrodek:
* Extract AWS magic strings as constants in the new class `AwsInstanceTemplate` in `cloud-api`
  * The similar magic strings used for Azure & GCP encryption scenarios have not been touched. The only exceptions are the strings `"type"` and `"key"` that are common to AWS & GCP (and might be used later for Azure as well for customer-managed encryption keys); these have been added as constants to the superclass `InstanceTemplate`.
* Replace raw strings (`"NONE"`, `"DEFAULT"`, `"CUSTOM"`) with `EncryptionType` enum constant names

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AwsInstanceTemplate.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AwsInstanceTemplate.java
@@ -1,0 +1,51 @@
+package com.sequenceiq.cloudbreak.cloud.model.instance;
+
+import java.util.Collection;
+import java.util.Map;
+
+import com.sequenceiq.cloudbreak.cloud.model.InstanceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.InstanceTemplate;
+import com.sequenceiq.cloudbreak.cloud.model.Volume;
+
+public class AwsInstanceTemplate extends InstanceTemplate {
+
+    /**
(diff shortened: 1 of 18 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AwsInstanceTemplate.java
```diff
@@ -0,0 +1,90 @@
...
+     *         <li>{@code "CUSTOM"}: Use the customer-managed symmetric KMS key specified in {@link #EBS_ENCRYPTION_KMS_KEY_ARN}.</li>
+     *     </ul>
+     * </p>
+     *
+     * @see #EBS_ENCRYPTION_ENABLED
+     * @see #EBS_ENCRYPTION_KMS_KEY_ARN
+     * @see #putParameter(String, Object)
+     * @see EncryptionType
+     */
+    public static final String EBS_ENCRYPTION_KMS_KEY_TYPE = "type";
```
biharitomi, line 50: [As it is used in case of GCP](https://github.com/hortonworks/cloudbreak/pull/8340/files#diff-89018022889ad01df1fabcd052257e5bR78) we could move this constant one layer upper to the provider generic InstanceTemplate class.

lajosrodek: Done. Also moved the similarly common `"key"` constant.

PR #7599, opened 2020-03-16
CB-5576, CM-5578 - Implement CM and CDH upgrade for SDX

pdarvasi:
Intentionally 3 commits to reflect the phases of the implementation.

PR #8345, opened 2020-06-19
CB-7558: Add FreeIPA install and uninstall logs to fluentd.

wonderslug:
In order to better diagnose install issues and track installs
from repairs the FreeIPA install and uninstall logs should
be uploaded to fluent

added logs
/var/log/ipaclient-install.log
/var/log/ipaserver-install.log
/var/log/ipaclient-uninstall.log
/var/log/ipareplica-install.log
/var/log/ipaserver-uninstall.log

Closes #7599


wonderslug:
@jamisonbennett let me know if you see any logs I missed.

PR #8346, opened 2020-06-19
CB-7574: Fix FreeIPA repair to remove dead ipa servers registration

jamisonbennett:
This fixes a case where a previously deleted freeipa server was not
fully deleted and it remained registered as a server but the host
entry was cleaned up. The freeipa server deletion was fixed to check
the ipa server list (ipa server_find) rather than the host list
(ip host_find) so that a delete is issued if the server is still
still registered.

This also changes the repair downscale to remove all previously
cleaned up servers. This was important because if the downscale leaves
some old information around, a future repair can ensure it is cleaned
up. Once the first downscale finishes, the UI no longer displays the
instance ID, so it makes it very difficult to determine the old
instance ID to specify in the repair request. For most of the
downscale operations, the deletion of previously deleted items just
skips over them.

This was tested manually on a local deployment of cloudbreak on a
cluster that was experiencing this issue.

Closes #CB-7574


freeipa/src/main/java/com/sequenceiq/freeipa/flow/chain/RepairFlowEventChainFactory.java
```diff
@@ -22,10 +26,13 @@ public String initEvent() {
 
     @Override
     public Queue<Selectable> createFlowTriggerEventQueue(RepairEvent event) {
+        Set<String> terminatedOrRemovedInstanceIdsSet = new HashSet<>(event.getInstanceIds());
+        terminatedOrRemovedInstanceIdsSet.addAll(event.getAdditionalTermiantedInstanceIds());
+        List<String> terminatedOrRemovedInstanceId = terminatedOrRemovedInstanceIdsSet.stream().collect(Collectors.toList());
```
wonderslug, line 31: nit: plural naming on list?

PR #8358, opened 2020-06-22
CB-7576: Reduce the max nodes of FreeIPA to 3

jamisonbennett:
This was merged into CB-2.24.0 https://github.com/hortonworks/cloudbreak/pull/8347

When FreeIPA HA has 3 out of 4 failed instances, there were several
issues that occurred such as CB-6842, CB-7523, CDPSDX-1793, and
CDPSDX-1869. These issues do not occur when 2 out of 3 nodes have
failed. So the maximum number of instances of FreeIPA is being capped
at 3 nodes.

Closes #CB-7576

### DIFF
--- a/freeipa/src/main/resources/application.yml
+++ b/freeipa/src/main/resources/application.yml
@@ -77,7 +77,7 @@ freeipa:
       initial-delay-millis: 60000
       fixed-delay-millis: 60000
   max:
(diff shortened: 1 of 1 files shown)


lacikaaa:
you don't need a separate pr for master, it will come with a merge from 2.24

PR #8364, opened 2020-06-22
CB-7553 Change HBASEJARS to use cdp-proxy-api

petersomogyi:
Add `-api` to HBASEJARS URL. Modify unit tests to verify /hbase/jars and /avatica endpoints.


doktoric:
@petersomogyi thank you

PR #8365, opened 2020-06-22
CB-7608: Modify FreeIPA backup to check if azure backup location host resolves

wonderslug:
CB-7608: Modify FreeIPA backup to check if azure backup location host resolves

This modifies the FreeIPA backup script to check if the backup location host is valid.  This is for azure only because the storage location is a customer provided hostname based on the storage account being used.

Closes #CB-7608


wonderslug:
@keyki Let me know if you want this in 2.24 instead

keyki:
Yes. It can go to 2.14 along with `--check-length=false`

wonderslug:
Going to resubmit to take care of conflicts

PR #8375, opened 2020-06-23
CB-7615 Add resource based authz for recipes

horadla23:


horadla23:
- ums notify deletion always uses internal actor (UMS also accept only internal actor in case this notify)
- resource based authz for recipes (tested locally with mowdev ums)
- typo fix for @agov

lnardai:
Can you also add an integration test to Autorization package that uses a recipe? It might require to create a Datahub, but we should check if everything is accessible.

horadla23:
> Can you also add an integration test to Autorization package that uses a recipe? It might require to create a Datahub, but we should check if everything is accessible.

Ok, I can create test for recipe without datahub. Since we do not have authz tests for datahub yet, I would prefer recipe test without datahub, later, when we have tests for datahub also, we can extend it with recipe assertions.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/RecipesV4Controller.java
```diff
@@ -67,33 +81,38 @@ public RecipeV4Response post(Long workspaceId, RecipeV4Request request) {
...
-    public RecipeV4Responses deleteMultiple(Long workspaceId, Set<String> names) {
+    @CheckPermissionByResourceNameList(action = AuthorizationResourceAction.DELETE_RECIPE)
+    public RecipeV4Responses deleteMultiple(Long workspaceId, @ResourceNameList Set<String> names) {
         Set<Recipe> deleted = recipeService.deleteMultipleByNameFromWorkspace(names, workspaceId);
         notify(ResourceEvent.RECIPE_DELETED);
         return new RecipeV4Responses(converterUtil.convertAllAsSet(deleted, RecipeV4Response.class));
     }
 
     @Override
+    @DisableCheckPermissions
```
lnardai, line 108: We should remove this endpoint?
Can we do a lookup in kibana if it's still used or not? There should be a controller interceptor for it.

horadla23: I am not sure, I haven't checked, I just think that this endpoint is not a security risk anyway

lnardai: Well, you can read anybody's recipe and it can contains sensitive information.

horadla23: yes, you're right, fixed

PR #8377, opened 2020-06-23
CB-7626. Update Azure storage plugin version for VM nodes (support write only mode)

oleewere:
- added write_only option for azure plugin
- use upload timestamp instead of index in the filename (format: 'HHSSmmff')
- set buffer chunk size to 4MB (as you need to upload in multiple parts per abfs blocks, the block size is 4MB, and as during upload knowing the position is required we cannot continue the uploading if for some reason the app is stopped and we cannot gather the blob properties through a head operation)

edit: also add an update to s3 path to use hms_slice instead of hex random

edit2: it's possible freeipa backup will need read anyway, if that's the case, maybe this fix is not a must


lajosrodek:
Still LGTM

PR #8378, opened 2020-06-23
Relabeling DDE templates. Add new configs to DDE blueprint services

gezan:


core/src/main/resources/defaults/blueprints/cdp-dde-720.bp
```diff
@@ -42,7 +42,48 @@
...
+            "name": "hdfs_data_dir",
+            "value": "/solr-dde"
+          },
+          {
+            "name": "ranger_audit_hdfs_dir",
+            "value": "${ranger_base_audit_url}/solr-dde"
+          },
+          {
+            "name": "ranger_audit_hdfs_spool_dir",
+            "value": "/var/log/solr-dde/audit/hdfs/spool"
```
doktoric, line 76: I guess we should not write anything to the root volume

doktoric:
probably we should delete the old template with name without version

doktoric:
2.24 should be the target branch

PR #8382, opened 2020-06-23
CB-7632 - Make sure Storage Account name does not change prior to CB …

pdarvasi:


keyki:
Can you add a unit test please to make sure we don't change it in the future for legacy Storage Accounts?

PR #8383, opened 2020-06-23
CB-7636: Rename "FreeIPA" to "FreeIpa" in files and methods

jamisonbennett:
NOTE: I avoided changing `@ApiModel("FreeIPAKerberosV1Descriptor")` so that it doesn't affect downstream APIs.

Closes #CB-7636


jamisonbennett:
@keyki I want to ensure that this is what we really want. It seems like there is quite a lot that changed. This was in response to a code review comment on https://github.com/hortonworks/cloudbreak/pull/8306 suggesting we correct the FreeIPA naming and I think we didn't know it was as big of a changeset.

jamisonbennett:
I talked with @keyki and he was good with this, so I will merge it.

PR #8386, opened 2020-06-24
CB-7640: Require non-repaired FreeIPA instance to be available dur...

jamisonbennett:
...ing a repair

When FreeIPA is repaired, the remaining non-repaired instances cannot
be stopped. If the repair operation is not forced then the
non-repaired instances must be available. This validation ensures the
repair operation will not fail from either replicating from a paused
instance or fail to establish cluster proxy links to every instance
that hasn't been repaired.

This was tested with unit tests and on a local deployment of
cloudbreak.

Closes #CB-7640

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
@@ -76,15 +76,23 @@ public class RepairInstancesService {
     @Inject
     private StackUpdater stackUpdater;
 
-    private void validate(String accountId, Stack stack, Set<InstanceMetaData> remainingInstances, Collection<InstanceMetaData> instancesToRepair) {
+    private void validate(String accountId, Stack stack, Set<InstanceMetaData> remainingGoodInstances, Set<InstanceMetaData> remainingBadInstances,
+            Collection<InstanceMetaData> instancesToRepair) {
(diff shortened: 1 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -76,15 +76,21 @@
     @Inject
     private StackUpdater stackUpdater;
 
-    private void validate(String accountId, Stack stack, Set<InstanceMetaData> remainingInstances, Collection<InstanceMetaData> instancesToRepair) {
+    private void validate(String accountId, Stack stack, Set<InstanceMetaData> remainingGoodInstances, Set<InstanceMetaData> remainingBadInstances,
```
lacikaaa, line 79: I think it would help later if we log the instances parameters of this method

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/RepairInstancesService.java
```diff
@@ -190,6 +189,35 @@ public OperationStatus repairInstances(String accountId, RepairInstancesRequest
         return operationToOperationStatusConverter.convert(operation);
     }
 
+    private Set<InstanceMetaData> getRemainingGoodInstances(Map<String, InstanceMetaData> allInstancesByInstanceId,
```
lacikaaa, line 192: maybe a log here and for `getRemainingBadInstances` would be nice.

PR #8389, opened 2020-06-24
CB-7643 temporarily disable deterministic uid/gid script

keyki:


akanto:
Looks good. I think when we re-introduce it again then we shall refactor a bit to avoid copy paste and just use the include directive.

keyki:
> Looks good. I think when we re-introduce it again then we shall refactor a bit to avoid copy paste and just use the include directive.

The content is different. During upgrade we need to migrate otherwise it's just an init.

akanto:
Right, to be honest, I had to go through two times to spot that one line has a -m flag...

PR #8405, opened 2020-06-25
CB-7679 Review all DisabledPermission endpoints

horadla23:
- removed unused user API, CB relies on UMS regarding users
- removed unused workspace API, currently workspace are used only for tenant
- allow to InternalOnly annotation for controller to reduce annotations on methods
- created CustomPermissionCheck for methods which has custom authz logic in service layer
- Flow APIs should also called only by internal actor


akanto:
I think we might want to move this to master

horadla23:
Workspace API only used by int test: https://logs-dev.sre.cloudera.com/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!('@message'),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*%3Forkspace%3F4%3Fontroller*')),sort:!('@timestamp',desc))

usage of workspace API removed in this PR

User API is not used:
https://logs-dev.sre.cloudera.com/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!('@message'),index:'logstash-*',interval:auto,query:(query_string:(query:'*%3Fser%3F4%3Fontroller*')),sort:!('@timestamp',desc))

PR #8409, opened 2020-06-25
CB-7689 Mock test for datalake repair

foldik:


akanto:
I think we might want to move this to master

PR #8412, opened 2020-06-25
CB-7543 - [ASRG] Let entitlement drive default behaviour if no RG inf…

pdarvasi:


environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -304,4 +315,15 @@ public EnvironmentFeatures convertToEnvironmentTelemetryFeatures(FeaturesRequest
         features.setClusterLogsCollection(featuresRequest.getClusterLogsCollection());
         return features;
     }
+
+    private boolean azureSingleResourceGroupDeploymentEnabled() {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        boolean azureSingleResourceGroupDeploymentEnabled =
+                entitlementService.azureSingleResourceGroupDeploymentEnabled(userCrn, Crn.safeFromString(userCrn).getAccountId());
```
keyki, line 322: I think we should use the internal crn here.

PR #8413, opened 2020-06-25
CB-7636 test action name changed, ProxyController changed to tenantAware

lnardai:
- FreeIpa rename caused it


lnardai:
We need to fix this on 2.24 as well.

PR #8416, opened 2020-06-25
[CB-7630] Modified lite duty blueprints to suppress service level warnings

atxmemo:
Modified lite duty blueprints to suppress service level warnings

Closes https://jira.cloudera.com/browse/CB-7630


atxmemo:
@kmanamcheri good idea. updated!

PR #8420, opened 2020-06-26
CB-7704 [opdb] Revert "CB-7343 [opdb] Add Hue in the opdb blueprint"

jatsakthi:
Addition of hue to opdb template depends upon CM 7.2.1 version hence removing this change from the 7.2.0 blueprint. 

This reverts commit 1ef188066b2427e61c12cdf1d566bf18ca851d7e.

Testing: Created cluster from custom blueprint and verified that dh cluster turns up.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-opdb-720.bp
@@ -155,13 +155,6 @@
               }
             ],
             "base": true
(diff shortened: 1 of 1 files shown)


jatsakthi:
This change also needs to go into master branch. Kindly let me know if for that a new pr needs to be created. or this branch would get merged to the master branch in which case a new pr might not be necessary.

PR #8424, opened 2020-06-26
CB-5051 handle internal calls with internal actor correctly

horadla23:


horadla23:
- remove duplicated INTERNAL_ACTOR fields
- moved stastic internal actor field to ThreadBasedUserCrnProvider
- introduce doAsInternalActor calls
- fix internal calls for tag and telemtry endpoint, which failed previously

environment/src/main/java/com/sequenceiq/environment/tags/v1/controller/AccountTagController.java
```diff
@@ -71,6 +71,12 @@ public AccountTagController(AccountTagService accountTagService,
     @CheckPermissionByAccount(action = AuthorizationResourceAction.ENVIRONMENT_READ)
     public AccountTagResponses list() {
         String accountId = ThreadBasedUserCrnProvider.getAccountId();
+        return listInAccount(accountId);
+    }
+
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.ENVIRONMENT_READ)
```
lnardai, line 78: How is this limited to internalOnly?

horadla23:
- introduced @InternalOnly annotation to mark api which should call with internal actor only

environment/src/main/java/com/sequenceiq/environment/tags/v1/controller/AccountTagController.java
horadla23: introduced an annotation for that, let's discuss if there is a simpler way to do this

PR #8426, opened 2020-06-26
CB-7671 Extract JSON error message details from Azure exception

bergerdenes:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureCredentialConnector.java
```diff
@@ -68,7 +72,7 @@ public CloudCredentialStatus verify(AuthenticatedContext authenticatedContext) {
             }
         } catch (RuntimeException e) {
             LOGGER.warn(e.getMessage(), e);
-            return new CloudCredentialStatus(cloudCredential, CredentialStatus.FAILED, e, e.getMessage());
+            return new CloudCredentialStatus(cloudCredential, CredentialStatus.FAILED, e, exceptionExtractor.extractErrorMessage(e));
```
topolyai5, line 75: `exceptionExtractor.extractErrorMessage(e)` can be null. Could you please change to a proper message if the root cause or the description of the root cause is null?

PR #8433, opened 2020-06-26
Revert 8429: Merge CB-2.24.0 to master

aszegedi:
Previous PR has been merged incorrectly. So the original commit hashes have changed. I'm reverting this to get back to the base.


horadla23:
i think there is no need to revert them, you can merge 2.24 to master again and those commits will appear again, but code won't change i think

PR #8435, opened 2020-06-26
CB-7727: Fix FreeIPA backup script for quoting issues and ret code

wonderslug:
This fixes the FreeIPA backup script for some quoting issues via
over-zealous shellcheck compliance.  As well this sets the AWS
upload check to better handle failing return codes.

Closes #CB-7727


keyki:
I think we merged to the original change to 2.24 so this should go there as well. Sorry about the shellcheck :|

wonderslug:
closing to move to 2.24

PR #8441, opened 2020-06-28
CB-7360 Speed up handling of encrypted EBS

lajosrodek:
This PR adds the ability to provision clusters with encrypted EBS in a fast way, avoiding unnecessary resource creations. Notes:
* Applies to FreeIPA & DL (#7833) as well as DH clusters.
* Affects root (always EBS) as well as EBS data / attached volumes.
* No change to service APIs; the flag controlling the fast logic is included as a new dynamic parameter in `InstanceTemplate`. The parameter key has been added to `AwsInstanceTemplate` introduced earlier in #8340.
* The fast logic is guarded by the new entitlement `CDP_CB_FAST_EBS_ENCRYPTION`. This entitlement gets checked by both FreeIPA and Cloudbreak / Core Svc; the absence of the entitlement grant will result in the legacy behavior.
* When testing with local CB, `mock-caas` defaults to presenting `CDP_FREEIPA_DL_EBS_ENCRYPTION` (#7833) and `CDP_CB_FAST_EBS_ENCRYPTION` always granted. The following VM argument shall be added to `MockCaasApplication` to disable these entitlements: `-Dauth.mock.freeipadlebsencryption.enable=false -Dauth.mock.fastebsencryption.enable=false`. The new defaults also revert the setting made in #8309.
* Unit tests have been added for all logic changes. These are all based on JUnit 5 and AssertJ, sometimes using `ParameterizedTest`s. Any existing tests have been ported to the above libraries.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/AwsVolumeResourceBuilder.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/AwsVolumeResourceBuilder.java
@@ -44,6 +44,7 @@ import com.sequenceiq.cloudbreak.cloud.aws.context.AwsContext;
 import com.sequenceiq.cloudbreak.cloud.aws.encryption.EncryptedSnapshotService;
 import com.sequenceiq.cloudbreak.cloud.aws.service.AwsResourceNameService;
 import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
+import com.sequenceiq.cloudbreak.cloud.aws.view.AwsInstanceView;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
 import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
 import com.sequenceiq.cloudbreak.cloud.exception.CloudConnectorException;
@@ -172,6 +173,8 @@ public class AwsVolumeResourceBuilder extends AbstractAwsComputeBuilder {
 
         List<Future<?>> futures = new ArrayList<>();
         String snapshotId = getEbsSnapshotIdIfNeeded(auth, cloudStack, group);
+        boolean encryptedVolumeUsingFastApproach = isEncryptedVolumeUsingFastApproachRequested(group);
+        String volumeEncryptionKey = getVolumeEncryptionKey(group, encryptedVolumeUsingFastApproach);
         TagSpecification tagSpecification = new TagSpecification()
                 .withResourceType(com.amazonaws.services.ec2.model.ResourceType.Volume)
                 .withTags(awsTaggingService.prepareEc2Tags(auth, cloudStack.getTags()));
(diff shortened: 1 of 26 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/AwsVolumeResourceBuilder.java
```diff
@@ -226,18 +229,39 @@ private String getAvailabilityZoneFromSubnet(AuthenticatedContext auth, CloudRes
...
-    private Function<VolumeSetAttributes.Volume, CreateVolumeRequest> createVolumeRequest(String snapshotId, TagSpecification tagSpecification,
-            VolumeSetAttributes volumeSet) {
-        return volume -> new CreateVolumeRequest()
-                .withAvailabilityZone(volumeSet.getAvailabilityZone())
-                .withSize(volume.getSize())
-                .withSnapshotId(snapshotId)
-                .withTagSpecifications(tagSpecification)
-                .withVolumeType(volume.getType());
+    private boolean isFastEbsEncryptionEnabled(Group group) {
+        return new AwsInstanceView(group.getReferenceInstanceConfiguration().getTemplate()).isFastEbsEncryptionEnabled();
```
keyki, line 233: Minor, we could introduce a new method to replace `group.getReferenceInstanceConfiguration().getTemplate()` with getReferenceTemplate as it is used at many places.

cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
```diff
@@ -89,7 +89,7 @@
     "VPCSecurityGroupsParameter": {
         "Type": "List<AWS::EC2::SecurityGroup::Id>",
         "Description": "VPC security groups"
-    },
+    }
```
keyki, line 92: Is this a bug being fixed? What does it cause if we have the `,` there?

lajosrodek: Yes, it is a malformed JSON. I noticed this when adding a unit test case. I wonder if anyone has used that code path before, otherwise CF would have rejected the request (unless their JSON parser is lenient enough to ignore such extra commas).

lajosrodek: In particular, each test case of `CloudFormationTemplateBuilderDBTest` starts with the assertion `assertThat(JsonUtil.isValid(result)).overridingErrorMessage("Invalid JSON: " + result).isTrue();`. This failed due to the above extra comma.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/AwsVolumeResourceBuilder.java
lajosrodek: Good idea, will do that in a follow-up PR. 👍

lajosrodek: See #8482.

PR #8450, opened 2020-06-29
CB-7731 [ASRG] check if RG exists returns 403 and no text

gergopapi2:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -136,7 +137,15 @@ public ResourceGroups getResourceGroups() {
     }
 
     public boolean resourceGroupExists(String name) {
-        return getResourceGroups().contain(name);
+        try {
+            return getResourceGroups().contain(name);
+        } catch (CloudException e) {
+            if (e.getMessage().contains("Status code 403")) {
```
pdarvasi, line 143: This does not look that nice but I assume you've found no better way to distinguish the error type, right?

pdarvasi:
I've retriggered the timed out unit-tests

PR #8453, opened 2020-06-29
CB-7701: Capture failures and log them when dropping and creating database for restore.

brycederriso:
Print to stderr when there are problems.
Capture log output by redirecting to the LOGFILE.

Tested by running backup/restore procedure on a Data Lake.
When restoring, I did not stop the running hive services, so PostgreSQL fails the `drop` statement.

Got Salt output that looks like the following:
```
ID: restore_postgresql_db
    Function: cmd.run
        Name: /opt/salt/scripts/restore_db.sh AWS s3://eng-sdx-daily-datalake/bderriso-cranston/data/backup1 dbsvr-a40ba31f-6ff5-4323-8ef2-d758d8129ace.czhkrnzwtcwb.us-west-2.rds.amazonaws.com 5432 ucklnfysen
      Result: False
     Comment: Command "/opt/salt/scripts/restore_db.sh AWS s3://eng-sdx-daily-datalake/bderriso-cranston/data/backup1 dbsvr-a40ba31f-6ff5-4323-8ef2-d758d8129ace.czhkrnzwtcwb.us-west-2.rds.amazonaws.com 5432 ucklnfysen" run
     Started: 15:53:54.617916
    Duration: 6338.242 ms
     Changes:
              ----------
              pid:
                  26332
              retcode:
                  1
              stderr:
                  ERROR:  database "hive" is being accessed by other users
                  DETAIL:  There are 40 other sessions using the database.
              stdout:
                  Logs at /var/log/dl_postgres_restore.log
                  Completed 256.0 KiB/1.3 MiB (666.8 KiB/s) with 2 file(s) remaining
                  Completed 512.0 KiB/1.3 MiB (1.3 MiB/s) with 2 file(s) remaining
                  Completed 768.0 KiB/1.3 MiB (1.9 MiB/s) with 2 file(s) remaining
                  Completed 1009.4 KiB/1.3 MiB (2.5 MiB/s) with 2 file(s) remaining
                  download: s3://eng-sdx-daily-datalake/bderriso-cranston/data/backup1/ranger_backup to ../var/tmp/postgres_restore_staging/ranger_backup
                  Completed 1009.4 KiB/1.3 MiB (2.5 MiB/s) with 1 file(s) remaining
                  Completed 1.2 MiB/1.3 MiB (3.1 MiB/s) with 1 file(s) remaining
                  Completed 1.3 MiB/1.3 MiB (3.2 MiB/s) with 1 file(s) remaining
                  download: s3://eng-sdx-daily-datalake/bderriso-cranston/data/backup1/hive_backup to ../var/tmp/postgres_restore_staging/hive_backup
```

Verified the failure showed up in the `dl_postgres_restore.log` file:
```
2020-06-29T14:20:27Z INFO  Restoring hive
ERROR:  database "hive" is being accessed by other users
DETAIL:  There are 40 other sessions using the database.
2020-06-29T14:20:32Z ERROR Unable to drop database hive
```

**I also fixed a bug where the backup script was not exiting correctly.**
I made a change that only checked that the exit code was `1`. I've fixed it to check for exit codes that are `-ne 0`.

Closes CB-7701

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
@@ -52,27 +52,27 @@ restore_db_from_local() {
 
   doLog "INFO Restoring $SERVICE"
 
-  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "drop database ${SERVICE};"
-  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "create database ${SERVICE};"
-  psql --host="$HOST" --port="$PORT" --dbname="$SERVICE" --username="$USERNAME" < "$BACKUP" >>$LOGFILE 2>&1 || errorExit "Unable to restore ${SERVICE}"
+  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "drop database ${SERVICE};" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || errorExit "Unable to drop database ${SERVICE}"
+  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "create database ${SERVICE};" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || errorExit "Unable to re-create database ${SERVICE}"
+  psql --host="$HOST" --port="$PORT" --dbname="$SERVICE" --username="$USERNAME" < "$BACKUP" > $LOGFILE 2> >(tee -a $LOGFILE >&2) || errorExit "Unable to restore ${SERVICE}"
   doLog "INFO Succesfully restored ${SERVICE}"
 }
 
 run_azure_restore() {
   mkdir -p "$BACKUPS_DIR"
-  azcopy login --identity >>$LOGFILE 2>&1 || errorExit "Could not login to Azure"
-  azcopy copy "$CLOUD_LOCATION"/* "$BACKUPS_DIR" >>$LOGFILE 2>&1 || errorExit "Could not copy backups from Azure."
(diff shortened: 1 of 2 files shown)


hreeve-cloudera:
@brycederriso, Kalyan and I were just talking about this JIRA. We need the stderr output to keep going to stderr, so that it's propagated back up to the cloudbreak service as an error. With `2>&1`, you're sending stderr to stdout, which means we won't see any stderr in the salt response and can't extract a meaningful error message from it.

So the highest priority is that we need stderr to actually go to stderr. However, ideally we also want it to go to the log file so that we can do troubleshooting. I don't know if there's an elegant solution. This was my first thought, although you may be able to come up with something better.

1. Pipe stdout to the standard log file, and stderr to a temp log file.
2. Check if the stderr log file exists and is not empty.
3. If so, echo it's contents to the log file, and then to stderr on the console.
4. Finally, exit the script with a failure code.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -52,8 +52,8 @@ restore_db_from_local() {
 
   doLog "INFO Restoring $SERVICE"
 
-  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "drop database ${SERVICE};"
-  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "create database ${SERVICE};"
+  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "drop database ${SERVICE};" >>$LOGFILE 2>&1 || errorExit "Unable to drop database ${SERVICE}"
```
hreeve-cloudera, line 55: Commenting here primarily to do a "request changes" review. I think that'll prevent the change from being merged in accidentally. See my general comment for the requested changes.

kkalvagadda1:
Recheck

brycederriso:
recheck

brycederriso:
rebuild

brycederriso:
retest this please

PR #8455, opened 2020-06-29
CB-6588 [ASRG] e2e test

gergopapi2:
With this change the single RG parameters are introduced but it should not change anything yet.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/environment/EnvironmentTestDto.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/environment/EnvironmentTestDto.java
@@ -26,13 +26,14 @@ import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentNe
 import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentRequest;
 import com.sequenceiq.environment.api.v1.environment.model.request.LocationRequest;
 import com.sequenceiq.environment.api.v1.environment.model.request.aws.AwsEnvironmentParameters;
-import com.sequenceiq.environment.api.v1.environment.model.request.aws.S3GuardRequestParameters;
+import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureEnvironmentParameters;
 import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
 import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentStatus;
 import com.sequenceiq.environment.api.v1.environment.model.response.SimpleEnvironmentResponse;
 import com.sequenceiq.it.cloudbreak.CloudbreakClient;
 import com.sequenceiq.it.cloudbreak.EnvironmentClient;
 import com.sequenceiq.it.cloudbreak.Prototype;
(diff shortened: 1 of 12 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/environment/EnvironmentTestDto.java
```diff
@@ -212,11 +214,15 @@ public EnvironmentTestDto withNetwork(String key) {
...
     public EnvironmentTestDto withS3Guard(String tableName) {
-        AwsEnvironmentParameters awsEnvironmentParameters = new AwsEnvironmentParameters();
-        S3GuardRequestParameters s3GuardRequestParameters = new S3GuardRequestParameters();
-        s3GuardRequestParameters.setDynamoDbTableName(tableName);
-        awsEnvironmentParameters.setS3guard(s3GuardRequestParameters);
+        getCloudProvider().setS3Guard(this, tableName);
+        return this;
+    }
+
+    public EnvironmentTestDto withSingleResourceGroup(String resourceGroupUsage, String resourceGroupName) {
```
pdarvasi, line 221: I would rather call this withResourceGroupUsage because it can be used to set "multiple" model, too

gergopapi2: Thanks! I renamed it to withResourceGroup, without usage, as one parameter is called resourceGroupUsage.

gergopapi2:
retest this please

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/environment/EnvironmentStopStartTests.java
```diff
@@ -34,6 +35,9 @@
     @Inject
     private CredentialTestClient credentialTestClient;
 
+    @Inject
```
bergerdenes, line 38: I don't see this used

integration-test/src/main/resources/application.yml
```diff
@@ -138,6 +138,9 @@ integrationtest:
         assumerIdentity:
         loggerIdentity:
       secure: false
+    resourcegroup:
+      usage: MULTIPLE
+      name:
```
bergerdenes, line 143: is name not mandatory?
will single be tested as well?

gergopapi2: I added the default values as of now. We need to test the single RG scenario. Once its done, we will set the test to: SINGLE. 
The name parameter: it is not mandatory. With usage MULTIPLE it is not allowed to provide it. With single it is optional. If provided, cloudbreak will use that RG, if not, then cloudbreak will create one and use it (latter use case is a demo use case, the real advantage comes when the customer provides an existing RG)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/environment/EnvironmentStopStartTests.java
gergopapi2: I checked all imports and they are used. May I ask you to double-check?

PR #8456, opened 2020-06-29
CB-7733 allow remote connections for postgres user

keyki:


orchestrator-salt/src/main/resources/salt/salt/postgresql/scripts/init_db.sh
```diff
@@ -23,4 +23,9 @@ echo $(date +%Y-%m-%d:%H:%M:%S) >> $CONFIG_DIR/init_{{ service }}_db_executed
 
 {% endfor %}
 
+# Allow connections with postgres user remotely
+sed -i '1ihost postgres postgres 0.0.0.0/0 md5' $CONFIG_DIR/pg_hba.conf
```
doktoric, line 27: I think this should applied for customer cidr which already stored in env service

keyki: There is no point. Every other user is open to 0.0.0.0/0 as well. Besides, we don't have 5432 open at all.

PR #8457, opened 2020-06-30
Comments / questions on changes to CB for faster scale-up

sidseth:
Questions marked with  "TODO Question"

The rest talks about the various proposed changes.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsUpscaleService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsUpscaleService.java
@@ -104,6 +104,13 @@ public class AwsUpscaleService {
 
         awsTaggingService.tagRootVolumes(ac, amazonEC2Client, instances, stack.getTags());
 
+        // TODO: Reduce the polling interval from 10 seconds to something like 3-5 seconds.
+        //  Alternately a dynamic polling interval, with 'expectedRunTime' as a parameter.
+        //  e.g. 50 seconds to launch an instance - would poll every 10-15 seconds until the 40s mark,
+        //  after which it switches to 2-3 seconds.
+
+        // TODO Eventually: Try re-using connection. Each connection + call to AWS seems to take 300+ms on mow-priv.
+        //  Post new node addition - tagging etc ends up taking a couple of seconds.
         return singletonList(new CloudResourceStatus(cfStackUtil.getCloudFormationStackResource(resources), ResourceStatus.UPDATED));
     }
 
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
@@ -136,6 +136,12 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
                 List<ApiHost> hosts = clouderaManagerApiFactory.getHostsResourceApi(apiClient).readHosts(null, null, SUMMARY).getItems();
                 ApiHostRefList body = createUpscaledHostRefList(upscaleHostNames, hosts);
                 clustersResourceApi.addHosts(clusterName, body);
+                // TODO Eventually (CM): activateParcel takes at least 20 seconds in CM.
+                //  Ideally, it should not be required (it is required right now) when a new node is being added to the cluster,
+                //  or if it is required, needs to be faster.
+
+                // TODO Eventually: Any possibility of parallelizing these operations. Does CM require the new node to be
+                //  up and running for some of these API calls.
                 activateParcel(clustersResourceApi);
                 applyHostGroupRolesOnUpscaledHosts(body, hostGroup.getName());
             } else {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
@@ -173,6 +173,11 @@ public class ClusterBuilderService {
                 proxyConfig.orElse(null),
                 template);
         clusterService.save(cluster);
+        // There are not post-install recipes by default. Post-install doesn't seem to check the recipe list,
+        //  and executes always, which consistently adds 10s.
+        // TODO Question: Is there a reason for this, or is it safe to switch this to
+        //  recipeEngine.executePostInstallRecipes(stack, hostGroupService.getRecipesByCluster(cluster.getId()));
+        //  which would then execute only if recipes exist.
         recipeEngine.executePostInstallRecipes(stack);
         Set<InstanceMetaData> instanceMetaDatas = instanceMetaDataByHostGroup.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
         clusterCreationSuccessHandler.handleClusterCreationSuccess(instanceMetaDatas, stack.getCluster());
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/RecipeEngine.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/RecipeEngine.java
@@ -60,11 +60,15 @@ public class RecipeEngine {
     public void executePostAmbariStartRecipes(Stack stack, Collection<Recipe> recipes) throws CloudbreakException {
         if ((stack.getCluster() != null && ldapConfigService.isLdapConfigExistsForEnvironment(stack.getEnvironmentCrn(), stack.getName()))
                 || recipesFound(recipes, POST_CLOUDERA_MANAGER_START)) {
+            // The ldapConfigService check forces this to run ALWAYS, which adds 10 seconds.
+            // TODO Question: Is it safe to remove the ldapConfigService.isLdap configured (Doesn't seem to do anything useful in the recipe execution)?
+            // 10s saved by removing that check.
             orchestratorRecipeExecutor.postClusterManagerStartRecipes(stack);
         }
     }
 
     public void executePostInstallRecipes(Stack stack) throws CloudbreakException {
+        // TODO: Change this to accept a list of recipes, and execute only if recipes exist.
         orchestratorRecipeExecutor.postClusterInstall(stack);
     }
 
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
@@ -83,6 +83,13 @@ public class FreeIpaCleanupService {
     }
 
     private void pollCleanupOperation(OperationStatus operationStatus) {
+        // TODO Question: Is a cleanup operation required?
+        //  freeipa join for a new host uses the --force flag, which wouls presumably override old entries.
+        //  If that is the case - can this operation be skipped?
+        //  Alternately, have this executed as part of the salt ipa-client-install (if that is any faster)
+
+        //  This takes over 5 seconds due to the poll interval of 5s.
+        //  With a 1s poll interval, this operation still takes about 3.5 seconds.
         FreeIpaOperationPollerObject opretaionPollerObject = new FreeIpaOperationPollerObject(operationStatus.getOperationId(),
                 operationStatus.getOperationType().name(), operationV1Endpoint);
         Pair<PollingResult, Exception> pollingResult = freeIpaOperationChecker
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
@@ -343,6 +343,9 @@ public class SaltOrchestrator implements HostOrchestrator {
     @Override
     public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, SaltConfig saltConfig, ExitCriteriaModel exitModel)
             throws CloudbreakOrchestratorException {
+        // TODO Question: Why does this get called with allNodes set to all nodes in the cluster,
+        //  instead of just the new nodes which are being added?
+
         GatewayConfig primaryGateway = getPrimaryGatewayConfig(allGateway);
         Set<String> gatewayTargetIpAddresses = getGatewayPrivateIps(allGateway);
         Set<String> gatewayTargetHostnames = getGatewayHostnames(allGateway);
@@ -353,14 +356,33 @@ public class SaltOrchestrator implements HostOrchestrator {
             Callable<Boolean> saltPillarRunner = saltRunner.runner(hostSave, exitCriteria, exitModel);
             saltPillarRunner.call();
 
+            // Observation: The only pillar entries which seem to change when adding new nodes are disks and nodes.
+            //  PillarSave isn't very expensive - but the entire set of pillarSaves does take over 1 second.
             for (Entry<String, SaltPillarProperties> propertiesEntry : saltConfig.getServicePillarConfig().entrySet()) {
                 OrchestratorBootstrap pillarSave = new PillarSave(sc, gatewayTargetIpAddresses, propertiesEntry.getValue());
                 saltPillarRunner = saltRunner.runner(pillarSave, exitCriteria, exitModel);
                 saltPillarRunner.call();
             }
 
+            // Each grainAddRunner takes at least 0.6 seconds (combination of grains.append + grains.get). Occasionally longer than 1 second.
+            //  All of the grain additions here end up adding 5 seconds to a new node addition.
+
+            // Salt allows for multiple grains to be added in a single call. This call, howeever, will not throw an error
+            //  if a grain already exists. Instead, it will keep appending to the list of grains.
+            //  Since this operation is called with all nodes - moving over to a single call to append multiple grains
+            //  will likely break a lot of things.
+
+            // TODO Question: Back to why this is invoked with allNodes vs new nodes only. NewNodes only allows for such optimizations.
+            //  In theory, this could be a grains.get + construct the append command in node batches (i.e. existing nodes will get almost no
+            //  changes. New nodes will be sent as a single bulk operation). Not familiar enough with the flow to be confident about making
+            //  such a change. 5s is a reasonable amount of time though.
             setAdMemberRoleIfNeeded(allNodes, saltConfig, exitModel, sc, allNodeHostname);
             setIpaMemberRoleIfNeeded(allNodes, saltConfig, exitModel, sc, allNodeHostname);
+
+            // TODO Question: Can some of these grains be set ahead of time during image building, or salt-bootstrap to
+            //  avoid some of these calls.
+
+            // TODO: Add a call here to set a 'prewarmed' role (or set this up in the image itself)
 
             // knox
             if (primaryGateway.getKnoxGatewayEnabled()) {
@@ -377,7 +399,11 @@ public class SaltOrchestrator implements HostOrchestrator {
             }
             grainUploader.uploadGrains(allNodes, saltConfig.getGrainsProperties(), exitModel, sc, exitCriteria);
 
+            // TODO Question: Is SyncAllRunner needed? This takes about 1.8s. Operations seem to work fine without this.
+            //  Maybe this call is more optimal when adding a large set of nodes?
             saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(allNodeHostname, allNodes), exitModel, exitCriteria);
+
+            // TODO Question: Similarly for MineUpdateRunner. About 1 second. Are mine updates required?
             saltCommandRunner.runSaltCommand(sc, new MineUpdateRunner(gatewayTargetHostnames, allNodes), exitModel, exitCriteria);
         } catch (ExecutionException e) {
             LOGGER.warn("Error occurred during bootstrap", e);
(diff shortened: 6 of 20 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -136,6 +136,12 @@ public void initApiClient() throws ClusterClientInitException {
                 List<ApiHost> hosts = clouderaManagerApiFactory.getHostsResourceApi(apiClient).readHosts(null, null, SUMMARY).getItems();
                 ApiHostRefList body = createUpscaledHostRefList(upscaleHostNames, hosts);
                 clustersResourceApi.addHosts(clusterName, body);
+                // TODO Eventually (CM): activateParcel takes at least 20 seconds in CM.
```
keyki, line 139: We had all sorts of issues when we didn't invoke the activate parcel. What we did to make it faster is to pre-calculate the parcel hash on the image and replace the parcel itself with a .skiphash file so CM does not try to read its content.

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
```diff
@@ -173,6 +173,11 @@ public void buildCluster(Long stackId) throws CloudbreakException, ClusterClient
                 proxyConfig.orElse(null),
                 template);
         clusterService.save(cluster);
+        // There are not post-install recipes by default. Post-install doesn't seem to check the recipe list,
+        //  and executes always, which consistently adds 10s.
+        // TODO Question: Is there a reason for this, or is it safe to switch this to
+        //  recipeEngine.executePostInstallRecipes(stack, hostGroupService.getRecipesByCluster(cluster.getId()));
```
keyki, line 179: Good point. I don't think we need this phase if there are no recipes.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/RecipeEngine.java
```diff
@@ -60,11 +60,15 @@ public void executePreClusterManagerRecipes(Stack stack, Collection<Recipe> reci
     public void executePostAmbariStartRecipes(Stack stack, Collection<Recipe> recipes) throws CloudbreakException {
         if ((stack.getCluster() != null && ldapConfigService.isLdapConfigExistsForEnvironment(stack.getEnvironmentCrn(), stack.getName()))
                 || recipesFound(recipes, POST_CLOUDERA_MANAGER_START)) {
+            // The ldapConfigService check forces this to run ALWAYS, which adds 10 seconds.
```
keyki, line 63: There is a comment that we use this phase for LDAP sync, but I can't find it in the code so I guess it's not valid anymore.

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
```diff
@@ -83,6 +83,13 @@ public void cleanup(Stack stack, boolean hostOnly, boolean recover, Set<String>
     }
 
     private void pollCleanupOperation(OperationStatus operationStatus) {
+        // TODO Question: Is a cleanup operation required?
```
keyki, line 86: I think this is necessary, --force does not override the DNS entries it just appends a new IP to the existing hostname causing all sorts of issues for Kerberos.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -343,6 +343,9 @@ public void bootstrapNewNodes(List<GatewayConfig> allGatewayConfigs, Set<Node> t
     @Override
     public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, SaltConfig saltConfig, ExitCriteriaModel exitModel)
             throws CloudbreakOrchestratorException {
+        // TODO Question: Why does this get called with allNodes set to all nodes in the cluster,
```
keyki, line 346: I think this method is used for both install and upscale, hence all nodes. Certainly it can be improved.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -353,15 +356,34 @@ public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, S
             Callable<Boolean> saltPillarRunner = saltRunner.runner(hostSave, exitCriteria, exitModel);
             saltPillarRunner.call();
 
+            // Observation: The only pillar entries which seem to change when adding new nodes are disks and nodes.
```
keyki, line 359: True, but it might involve a lot of complexity to figuring out what pillars to save. Maybe someone introduces something dynamic in one of the pillar files and expects it to be refreshed during repair and it won't work. We're also working on updating the Salt states on existing clusters and for that we need to push down all the pillars as well again.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -353,15 +356,34 @@ public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, S
...
+            //  PillarSave isn't very expensive - but the entire set of pillarSaves does take over 1 second.
             for (Entry<String, SaltPillarProperties> propertiesEntry : saltConfig.getServicePillarConfig().entrySet()) {
                 OrchestratorBootstrap pillarSave = new PillarSave(sc, gatewayTargetIpAddresses, propertiesEntry.getValue());
                 saltPillarRunner = saltRunner.runner(pillarSave, exitCriteria, exitModel);
                 saltPillarRunner.call();
             }
 
+            // Each grainAddRunner takes at least 0.6 seconds (combination of grains.append + grains.get). Occasionally longer than 1 second.
```
keyki, line 367: The additional get is there to make sure the grains are added, especially on YCloud, because we observed many issues there Salt saying it's added, but actually it wasn't.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -353,15 +356,34 @@ public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, S
...
+            //  will likely break a lot of things.
+
+            // TODO Question: Back to why this is invoked with allNodes vs new nodes only. NewNodes only allows for such optimizations.
+            //  In theory, this could be a grains.get + construct the append command in node batches (i.e. existing nodes will get almost no
+            //  changes. New nodes will be sent as a single bulk operation). Not familiar enough with the flow to be confident about making
+            //  such a change. 5s is a reasonable amount of time though.
             setAdMemberRoleIfNeeded(allNodes, saltConfig, exitModel, sc, allNodeHostname);
             setIpaMemberRoleIfNeeded(allNodes, saltConfig, exitModel, sc, allNodeHostname);
 
+            // TODO Question: Can some of these grains be set ahead of time during image building, or salt-bootstrap to
```
keyki, line 382: We use the same image for all instances and all types of clusters. DL & DH, thus we can't distinguish in image build time.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -377,7 +399,11 @@ public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, S
             }
             grainUploader.uploadGrains(allNodes, saltConfig.getGrainsProperties(), exitModel, sc, exitCriteria);
 
+            // TODO Question: Is SyncAllRunner needed? This takes about 1.8s. Operations seem to work fine without this.
```
keyki, line 402: This help us to make sure the newly added pillars are properly used for the next Highstate.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -377,7 +399,11 @@ public void initServiceRun(List<GatewayConfig> allGateway, Set<Node> allNodes, S
...
+            //  Maybe this call is more optimal when adding a large set of nodes?
             saltCommandRunner.runSaltCommand(sc, new SyncAllRunner(allNodeHostname, allNodes), exitModel, exitCriteria);
+
+            // TODO Question: Similarly for MineUpdateRunner. About 1 second. Are mine updates required?
```
keyki, line 406: Mine is used as a cache and it's used for unbound configuration which is important during upscale.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -892,9 +920,12 @@ private void executeRecipes(GatewayConfig gatewayConfig, Set<Node> allNodes, Exi
         try (SaltConnector sc = createSaltConnector(gatewayConfig)) {
             // add 'recipe' grain to all nodes
             Set<String> targetHostnames = allNodes.stream().map(Node::getHostname).collect(Collectors.toSet());
+
+            // TODO Question: Execute after the recipes check says that recipes need to execute?
```
keyki, line 924: Right.

orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls
```diff
@@ -2,6 +2,7 @@
 {%- from 'cloudera/manager/settings.sls' import cloudera_manager with context %}
 {%- set manager_server_fqdn = salt['pillar.get']('hosts')[metadata.server_address]['fqdn'] %}
 
+{# Protect this via a grains[prewarmed] check. Saves 20s from an rpm database rebuild}
```
keyki, line 5: Yes, it can be.

orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls
```diff
@@ -11,6 +12,8 @@ install-cloudera-manager-agent:
     - unless:
       - rpm -q cloudera-manager-daemons cloudera-manager-agent
 
+{# Protect this via a grains[prewarmed] check}
+{# TODO Question: Can the prewarmed tag be set in the image itself?}
```
keyki, line 16: I think we can add default grains during image creation time, saying it's a pre-warmed image.

orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/start.sls
```diff
@@ -12,6 +12,7 @@ start_server:
   service.running:
     - enable: True
     - name: cloudera-scm-server
+{# Protect this via a grains[prewarmed] check}
```
keyki, line 15: Even the base image has the IPA packages installed. We could remove this, but need to validate our oldest available images in prod whether they have IPA installed as well.

orchestrator-salt/src/main/resources/salt/salt/java/init.sls
```diff
@@ -6,6 +6,9 @@
   {% set java_home = '/usr/lib/jvm/java' %}
 {% endif %}
 
+{# TODO Question: Can this move into pre-warmed images, and be protected by a check}
```
keyki, line 9: I think yes.

orchestrator-salt/src/main/resources/salt/salt/top.sls
```diff
@@ -6,6 +6,9 @@ base:
     - metadata
     - docker
     - recipes.runner
+    {# Can some of these be parallelized? fluentd, metering, ipa seem like good candidates which can work independently.}
```
keyki, line 9: We can look into it, but I expect issues with depending states are not waiting for each other which we can fix.

keyki:
We can make some changes in the polling, it can save a lot of time. 

Also we need to take a look at the proposed changes if they will remain backward compatible. We're working on updating the Salt states and pillars on existing clusters so we need to be cautious about what we change.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsUpscaleService.java
```diff
@@ -104,6 +104,13 @@
 
         awsTaggingService.tagRootVolumes(ac, amazonEC2Client, instances, stack.getTags());
 
+        // TODO: Reduce the polling interval from 10 seconds to something like 3-5 seconds.
+        //  Alternately a dynamic polling interval, with 'expectedRunTime' as a parameter.
+        //  e.g. 50 seconds to launch an instance - would poll every 10-15 seconds until the 40s mark,
+        //  after which it switches to 2-3 seconds.
```
lacikaaa, line 110: if we poll too often we have a higher chance of hitting throttling errors

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
lacikaaa: yes, this is required

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
lacikaaa: I think because we modify the pillars when new nodes are added, and we have to apply those changes on the already existing node. One thing for sure is unbound configuration

keyki:
There are few optimizations that won't worth the risk of introducing hidden problems for a couple of seconds. And there are others which definitely can be changed.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
lacikaaa: we had some intermittent, mysterious issues before adding it

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -892,9 +920,12 @@ private void executeRecipes(GatewayConfig gatewayConfig, Set<Node> allNodes, Exi
...
             saltCommandRunner.runSaltCommand(sc, new GrainAddRunner(targetHostnames, allNodes, "recipes", phase.value()), exitCriteriaModel, maxRetry,
                     exitCriteria);
             Set<String> allHostnames = allNodes.stream().map(Node::getHostname).collect(Collectors.toSet());
+            // TODO Question: Similarly here. Can the syncAllRunner be skipped?
```
lacikaaa, line 928: I wouldn't skip this either

orchestrator-salt/src/main/resources/salt/salt/unbound/init.sls
```diff
@@ -27,6 +27,10 @@ include_access_config:
   file.replace:
     - name: /etc/unbound/unbound.conf
     - pattern: '#include: "otherfile.conf"'
+{# Replace this, and any file.replace with sed}
+{# name: "sed -i 's/#include: \"otherfile\\.conf\"/include: \\\"\\/etc\\/unbound\\/access.conf\\\"/g' /etc/unbound/unbound.conf"}
+{# file.replace OR the unless check triggers a module load, which takes 5 seconds with an unpatched image, 2 seconds with a patch
```
lacikaaa, line 32: I won't skip unless, as if we don't have unless the state would always run, and salt would indicate it as if a change has happened

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
```diff
@@ -60,6 +60,7 @@ function check_and_install_plugin() {
 }
 
 function main() {
+{# Protect this via a grains[prewarmed] check}
```
oleewere, line 63: and how to check the plugin versions are right for the image (older pre-warmed) or CB version? in these cases we want to fail during deploy (if no network) or upgrade to the right versions for the plugins (if you have network)

orchestrator-salt/src/main/resources/salt/salt/top.sls
oleewere: my thoughts: I think we can achieve/win more if we would run some states parallel (salt can do that) - that are not depending hardly on each other (like in my case, telemetry related services) instead of removing checks that can be used to avoid deployment issues. as I see command states itself can be parallel (so you won't need to send multiple salt commands through CB)

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
sidseth: OK. Think this can be worked on further once the CM team has some time. Did some more multi-node tests - and at least this is independent of the number of nodes.

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
sidseth: Will add this to the patch.

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
sidseth: Ack. Would still be interested in eventually making this faster. This seems to take longer with more nodes being added (not linear growth though)

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: Thanks. Will look into this some more. Not making any changes in the initial patch, but I think this could be targeted for subsequent runs.
Am still curious about whether Unbound is still required - given IPA serves as the DNS.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: Updating the salt states on existing clusters - is there a jira where I could read about this?

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: OK. Not planning to change much here in the first patch.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: ok. in the grand scheme of adding multiple nodes - this adds time but there's much bigger items.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsUpscaleService.java
sidseth: Agreed. We're not talking orders of magnitude increase though.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: Ack

orchestrator-salt/src/main/resources/salt/salt/unbound/init.sls
sidseth: Not planning to skip the unless check. Just seems like a combination of file.replace pattern along with unless causes all modules to be loaded.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
sidseth: How often are the plugin versions changed? When did we change them last. Also, how do we know the latest version which can be used - does Cloudbreak know this?, is the change made manually, is it just the latest version of the plugin that exists in the wild?
Either way - I think prewarmed images + some kind of grains could be used to skip the check.
Oliver, I'll reach out to you offline about the td-agent package - which does slow down bootup.

sidseth:
@akanto  - yes, meant from master, and plan to keep the changes simple. No drastic changes like splitting state machines between scale-up / new install, etc.
Thanks @keyki , @lacikaaa , @oleewere for the early feedback.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
oleewere: for instance it changed yesterday. And yes also CB knows about the change, if it should be upgraded (because of CB) and there is an old image that we are using. In case of no network we should see as an error message to not hide that something won't functioning on the node.

There are 2 main use cases with version conflict:
1. if we have new image in production but a new CB is not deployed - version conflict (plugins)
- new version in the prewarmed image
- CB still uses older one (in the salt scripts)
-> outcome: here we are checking ">=", so it won't upgrade anything
2. New CB deployed, a node uses an older image (it happens often) - version conflict (plugins)
- old version in the prewarmed image
- CB requires a new plugin version (based on the salt definitions) because of some new features or fixes
outcomes:
a.) if there is network and you can access cloudera or non-cloudera gem repo -> update gems
b.) if there is no network -> install fails and tell the user they need to use a new image, update gems manually or do changes in the firewall in order to update the plugins

If we skip the second use case (because prewarm check), customer will see errors a bit later during the installation, which can be much more complicated to figure out at first (like no logs on cloud-storage or a service fails etc.)

sidseth:
Closing in favor of https://github.com/hortonworks/cloudbreak/pull/8547

PR #8463, opened 2020-06-30
DISTX-504 Optimize Schedule Based Autoscaling

smaniraju:
Time Monitor should only monitor clusters which have timeAlerts defined.

Closes #DISTX-504

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/TimeMonitor.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/TimeMonitor.java
@@ -1,8 +1,11 @@
 package com.sequenceiq.periscope.monitor;
 
+import java.util.List;
+
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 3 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/TimeMonitor.java
```diff
@@ -25,4 +28,10 @@ public String getTriggerExpression() {
         return CronTimeEvaluator.class;
     }
 
+    @Override
+    protected List<Cluster> getMonitored() {
+        List<Long> clusterIds = getClusterService().findTimeAlertClustersForNode(StackType.WORKLOAD,
```
topolyai5, line 33: is it better in performance if you query the ids first and query the clusters by the ids second?

smaniraju: yes because of the join with timealert table. also only cluster ids which have schedule based alerts are filtered in monitor instead of loading all clusters with autoscaling enabled and processing them in CronTimeEvaluator

PR #8468, opened 2020-06-30
CB-7758. Add proxy support to fluentd plugin installations.

oleewere:
in case of old images with newer CB, for limited network, it should be possible to use proxy for gem/curl based installs


lajosrodek:
LGTM

PR #8478, opened 2020-07-01
CB-7767: Fix 2 FMS to use correct cluster proxy service names for ...

jamisonbennett:
...clusters created with CB prior to 2.21.

The first fix for CB-7767 only addressed clusters that were created
with cloudbreak 2.19 and prior. This specifically fixes it for 2.20.

Fix FreeIPA management service to use the the "freeipa" service name
for clusters that were created before CB 2.21.

NOTE: I have not yet tested this. We identified an issue with what was believed to be a CB 2.20 cluster and found this. I am testing this now.

Closes #CB-7767


jamisonbennett:
I will remove the DO_NOT_MERGE label as soon as this is tested locally.

jamisonbennett:
I validated a FreeIPA cluster that was created on CB 2.19, 2.20, and 2.24. After updating to 2.24, all of the FreeIPA instances are "available" and all are able to respond to commands like creating keytabs.

PR #8479, opened 2020-07-01
CB-7794 Node decommissioning fails in CCM enabled cluster due to auto…

cegganesh84:
…ssh connection restarts

1. CDPAM-677 shows reverse tunnel autossh connection may restart due to multiple reasons.
2. This causes few API calls to the CM failing intermittently during cluster scale down. It is known that CB makes a lot of CM API calls during scale down from CDPAM-665.
3. ClouderaManagerApiFactory is used so that all those calls are retried per retry policy.
4. In the case of SocketTimeoutException, the retry logic is skipped, because SocketTimeException and SocketException are in the different class hierarchy.
5. Added toleration rule for SocketTimeoutException as well with a unit test.

./gradlew build


cegganesh84:
I could easily reproduce this error in prod. Wondering if I should set the base branch to 2.24. Open to suggestions.

keyki:
I also merged it to 2.25. We'll see how long it takes to promote 2.25 to prod.

PR #8490, opened 2020-07-06
AWS ENA check in E2E

topolyai5:
Now the ena driver is enabled on AWS instances by default. We should check this value via E2E. Two commands will be checked:
* /usr/sbin/modinfo ena
* aws ec2 describe-instances --instance-ids iid--query "Reservations[].Instances[].EnaSupport" --region region


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/ssh/action/SshEnaDriverCheckActions.java
```diff
@@ -0,0 +1,78 @@
...
+        String modinfoEnaCmd = "/usr/sbin/modinfo ena";
+        String wrongResult = "modinfo: ERROR: Module ena not found.";
+        Pair<Integer, String> result = executeCommand(instanceIP, modinfoEnaCmd);
+        if (result.getValue().startsWith(wrongResult)) {
+            throw new TestFailException("Cannot find ena driver on the instance");
+        }
+    }
+
+    private void checkEnaSupport(String instanceIP, String instanceId, String region) {
+        String modinfoEnaCmd = String.format("aws ec2 describe-instances --instance-ids %s --query \"Reservations[].Instances[].EnaSupport\" --region %s",
```
akanto, line 34: I thought this will be implemented with an API call.

topolyai5: I'm going to rewrite to an API call

doktoric:
retest this please

PR #8491, opened 2020-07-06
CB-7830: Fix NullPointerException when appversion is not present

jamisonbennett:
If the appversion is not filled out in the database, then a null
pointer exception will happen when
ClusterProxyServiceAvailabilityChecker is used.

Unit tests were added.

Closes #CB-7830


jamisonbennett:
I didn't actually experience this issue but happened to notice the fix for CB-7404 when doing the post mortem for CB-7767 and I thought it would be a good idea to fix it the second location (ClusterProxyServiceAvailabilityChecker).

PR #8497, opened 2020-07-07
CB-7829-CB User Sync in periscope

smaniraju:
Periscope cluster entities are scoped at tenant level.
Periscope workspaceId related cluster retrieval is replaced with tenant level retrieval.
This change ensures Users belonging to same tenant can configure autoscaling rules as long as they have UMS permissions.
Closes #CB-7829


autoscale/src/main/java/com/sequenceiq/periscope/filter/CloudbreakUserConfiguratorFilter.java
```diff
@@ -34,16 +34,15 @@ public CloudbreakUserConfiguratorFilter(AutoscaleRestRequestThreadLocalService r
 
     @Override
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
-        CloudbreakUser cloudbreakUser = authenticatedUserService.getCbUser();
+        CloudbreakUser cbUser = authenticatedUserService.getCbUser();
         Long workspaceId = UNINITIALIZED_WORKSPACE_ID;
```
lnardai, line 38: I'm not familiar with the UNINITIALIZED_WORKSPACE_ID and why it would have value -1.
But if you wish to call a CB API directly, you should always use workspaceID 0.
That will default to the tenant default workspace which should be already existing.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
```diff
@@ -126,6 +137,15 @@ public StackV4Response get(String crn) {
         return stackCommonService.getByCrn(crn, Collections.emptySet());
     }
 
+    @Override
+    @DisableCheckPermissions
+    @PreAuthorize("hasRole('AUTOSCALE')")
+    public WorkspaceV4Response getDefaultWorkspaceForUser(String userId, String tenant) {
```
lnardai, line 143: There should be no such endpoint, this controller will be soon removed.

horadla23: To be more precise, Workspace related API will be removed soon, so WorkspaceResponse should not be used

horadla23:
Also I'm not sure if it is a desired behaviour to allow another users to scale a cluster. For users to gain a permission to do that should be done somehow through UMS.

We are in the middle of lot of authorization changes, so I think we need a concept for this scenario, but without introducing a new API endpoint.

smaniraju:
> * Please do not try to manually get and send in the workspace ID, use the default 0 value in calls where you might need this
> * do not add to the workspace controller as it will get phased out soon
> * you can always get the userCRN from the request it self or the resource should have an actorCrn at least
> * if it's something periodic it should be run with internalCRN

ack, based on discussion will scope periscope cluster requests at tenant level and remove all workspace related IDs.

smaniraju:
> Also I'm not sure if it is a desired behaviour to allow another users to configure autoscaling for a cluster. For users to gain a permission to do that should be done somehow through UMS.
> 
> We are in the middle of lot of authorization changes, so I think we need a concept for this scenario, but without introducing a new API endpoint.

Other users are allowed to configure autoscaling only if they have AccountLevel DataHub_READ and DataHub_WRITE permission. I understand that this will be replaced with resource level permission, but still other users with resource permission can edit autoscaling rules.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
smaniraju: ack will remove this change.

autoscale/src/main/java/com/sequenceiq/periscope/filter/CloudbreakUserConfiguratorFilter.java
smaniraju: ack. as discussed will default cluster to tenant scope and remove workspace related ids.

autoscale/src/main/java/com/sequenceiq/periscope/repository/ClusterRepository.java
```diff
@@ -19,16 +19,16 @@
...
+            " WHERE c.stackCrn = :stackCrn and c.clusterPertain.tenant = :tenant")
+    Optional<Cluster> findByStackCrnAndTenant(@Param("stackCrn") String stackCrn, @Param("tenant") String tenant);
 
     @Query(" SELECT c FROM Cluster c LEFT JOIN FETCH c.clusterPertain " +
-            " WHERE c.stackName = :stackName and c.clusterPertain.workspaceId = :workspaceId")
-    Optional<Cluster> findByStackNameAndWorkspaceId(@Param("stackName") String stackName, @Param("workspaceId") Long workspaceId);
+            " WHERE c.stackName = :stackName and c.clusterPertain.tenant = :tenant")
+    Optional<Cluster> findByStackNameAndTenant(@Param("stackName") String stackName, @Param("tenant") String tenant);
 
     @Query(" SELECT c FROM Cluster c LEFT JOIN FETCH c.clusterPertain " +
```
lnardai, line 29: Where do we use this ClusterID?
I'm ok with using Name and Tenant since it might come handy on the CLI, but we are not really using clusterID anywhere.

autoscale/src/main/java/com/sequenceiq/periscope/service/ClusterService.java
```diff
@@ -102,7 +102,7 @@ public Cluster create(AutoscaleStackV4Response stack) {
         ClusterPertain clusterPertain =
                 new ClusterPertain(stack.getTenant(), stack.getWorkspaceId(), stack.getUserId(), stack.getUserCrn());
         cluster.setClusterPertain(
-                clusterPertainRepository.findByUserCrn(clusterPertain.getUserCrn())
+                clusterPertainRepository.findByUserId(clusterPertain.getUserId())
```
lnardai, line 105: Why change userCRN to userID? userID is an old concept we should rely on userCRN as much as possible.
Also I don't really get the use case why I would want to get a list based on someones CRN.

autoscale/src/main/java/com/sequenceiq/periscope/repository/ClusterRepository.java
smaniraju: This is not used now since Id based endpoints is disabled in periscope now. This was used in Id based AlertEndpoint which is disabled but not deleted.

autoscale/src/main/java/com/sequenceiq/periscope/service/ClusterService.java
smaniraju: ok updated to use userCRN.

smaniraju:
> Thanks for changes regarding querying by tenant. One thing: as Laszlo mentioned, if it is possible, please use userCrn, since we would like to use that wherever it is possible and as I see here, it doesn't matter if you are using userId or userCrn.

Ok updated to use userCrn.

PR #8509, opened 2020-07-08
DISTX-494-Purge Old Periscope Cluster Data

smaniraju:
1. Persicope old cluster data was not initialized with newly added columns like stacktype, cloudplatform..
2. Purging old data so that periscope auto-syncs new cluster data.


smaniraju:
Some of the clusters in mow-dev don't have values for newly added fields. So purging existing data so that cluster data is auto-synced from CB.

PR #8511, opened 2020-07-08
CB-7776 Technical APIs should be called with internal actor - part #1

horadla23:


horadla23:
CB DB API usage (nothing as i see):
https://logs-dev.sre.cloudera.com/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!('@message'),index:'logstash-*',interval:auto,query:(query_string:(query:'*%3Fatabase%3F4%3Fontroller*')),sort:!('@timestamp',desc))

horadla23:
CB custom info api usage (nothing as i see):
https://logs-dev.sre.cloudera.com/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!('@message'),index:'logstash-*',interval:auto,query:(query_string:(query:'*%3Floudbreak%3Fnfo%3F4%3Fontroller*')),sort:!('@timestamp',desc))

horadla23:
- CB datalake API not used by UI or CLI, other services like environment service will call is with internal actor
- situation is the same with CB cloudprovider API

horadla23:
- tested locally with env, sdx and distrox creation, looked ok

horadla23:
Also it looks cb datalake api is called with environmentcrn only, so i removed environmentname parameter: https://logs-dev.sre.cloudera.com/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!('@message'),index:'logstash-*',interval:auto,query:(query_string:(query:'*%3Fatalake%3F4%3Fontroller*')),sort:!('@timestamp',desc))

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-pull-request-builder-integration-test/10491/
https://github.com/hortonworks/cloudbreak/commit/35dfd9fdfe2543787f36c551b3ded26a907414c3
so it seems test in integration tests are passed, but because of this commit, swagger diff failed, even if I want to delete an API intentionally

horadla23:
I decided to not delete old db API but restrict it to internal only and mark as deprecated, that way I won't break compatibility but solve the authorization task.

PR #8512, opened 2020-07-08
Cb-7808 Flow events are missing from Structuctured events

topolyai5:
We filtered the structured events to datalake and datahub only but before 2.25 the CB stored as stacks. The legacy resource type added

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/decorator/responseprovider/StackResponseEventProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/decorator/responseprovider/StackResponseEventProvider.java
@@ -1,12 +1,16 @@
 package com.sequenceiq.cloudbreak.service.decorator.responseprovider;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.inject.Inject;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.api.endpoint.v4.common.StackType;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.events.responses.CloudbreakEventV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.StackResponseEntries;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackV4Response;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/decorator/responseprovider/StackResponseEventProvider.java
```diff
@@ -15,17 +19,34 @@
...
         return stackResponse;
     }
 
+    private List<CloudbreakEventV4Response> getLegacyStackType(Stack stack) {
+        List<CloudbreakEventV4Response> events = cloudbreakEventsFacade
+                .retrieveEventsByStack(stack.getId(), StackType.LEGACY, PageRequest.of(0, Integer.MAX_VALUE)).getContent();
+        if (events.isEmpty()) {
+            LOGGER.info("Cannot find any legacy events for stack: {}, crn: {}", stack.getId(), stack.getResourceCrn());
+        } else {
+            LOGGER.info("{} egacy events for stack: {}, crn: {}", events.size(), stack.getId(), stack.getResourceCrn());
```
doktoric, line 45: egacy -> legacy

doktoric:
should this go to 2.25 ?

akanto:
I think yes, it should go to 2.25, this is a regression in 2.25 and it worked well in 2.24.

core/src/main/java/com/sequenceiq/cloudbreak/service/decorator/responseprovider/StackResponseEventProvider.java
```diff
@@ -15,17 +19,34 @@
...
         List<CloudbreakEventV4Response> cloudbreakEvents = cloudbreakEventsFacade
                 .retrieveEventsByStack(stack.getId(), stack.getType(), PageRequest.of(0, Integer.MAX_VALUE)).getContent();
-        stackResponse.setCloudbreakEvents(cloudbreakEvents);
+        events.addAll(cloudbreakEvents);
+        events.addAll(getLegacyStackType(stack));
+        stackResponse.setCloudbreakEvents(events);
         return stackResponse;
     }
 
+    private List<CloudbreakEventV4Response> getLegacyStackType(Stack stack) {
```
akanto, line 39: I don't see any tests which ensure that events are not broken again.

topolyai5: I think we don't have any tests what can catch this. Here is the use case:
Start a distrox with 2.24
Update Cb to 2.25
Before the fix, the resource type with stacks was filtered out. 

If you want I can write a UT, but I think this test will not prevent similar cases.

doktoric:
retest this please

PR #8513, opened 2020-07-08
DISTX-453 Expose HBase REST API in the OpDB DataHub

petersomogyi:
Add HBase REST to OpDB blueprint
Remove WEBHBASE from cdp-proxy topology
Extend OpdbServiceEndpointCollectorTest with HBASEREST

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-opdb-721.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-opdb-721.bp
@@ -162,7 +162,14 @@
               "displayName": null,
               "configs": [],
               "base": true
-           }
+           },
+          {
+            "refName": "hbase-HBASERESTSERVER-BASE",
+            "roleType": "HBASERESTSERVER",
+            "displayName": null,
+            "configs": [],
+            "base": true
+          }
         ]
       },
       {
@@ -288,6 +295,7 @@
         "roleConfigGroupsRefNames": [
           "hbase-GATEWAY-BASE",
           "hbase-HBASETHRIFTSERVER-BASE",
+          "hbase-HBASERESTSERVER-BASE",
           "hdfs-GATEWAY-BASE",
           "hdfs-JOURNALNODE-BASE",
           "yarn-JOBHISTORY-BASE",
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -147,10 +147,6 @@
                 <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
             </param>
             <param>
-                <name>webhbase.acl</name>
-                <value>*;{{ salt['pillar.get']('gateway:envAccessGroup') }};*</value>
-            </param>
-            <param>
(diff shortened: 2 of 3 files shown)


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -261,12 +261,6 @@
                  <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
              </param>
              {%- endif %}
-             {% if 'WEBHBASE' in exposed and 'HBASERESTSERVER' in salt['pillar.get']('gateway:location') -%}
```
doktoric, line 264: is it ok if we remove this?

joshelser: Peter and I had talked about this -- we're not really sure why it was here in the first place.

The HBASERESTSERVER doesn't provide a UI. It's meant for programmatic REST access. By that stance, there's no good reason we can think of as to why it would be in the topology.xml.j2 file in addition to the topology-api.xml.j2

joshelser: Not sure is @risdenk happens to know?

risdenk: I don't know for sure. The history looks like it was added in https://github.com/hortonworks/cloudbreak/commit/f22a5279c4d916d73067b24a9800f4df88c89c3d. I don't know of any usage of this endpoint from the UI topology.

btw if this goes can also remove the `webhbase.acl` section a little further up.

core/src/main/resources/defaults/blueprints/cdp-opdb-721.bp
```diff
@@ -162,7 +162,14 @@
               "displayName": null,
               "configs": [],
               "base": true
-           }
+           },
+          {
+            "refName": "hbase-HBASERESTSERVER-BASE",
+            "roleType": "HBASERESTSERVER",
```
risdenk, line 168: Make sure that CM is setting up HBase REST in a reasonable manner - including requiring SPNEGO on the REST server endpoint. It would be REALLY bad to have it by default be exposed without authentication.

petersomogyi: Configs in CM look correct to me, e.g. `hbase.rest.ssl.enabled=true` and REST Server TLS/SSL is also enabled.
My test cluster is available if you'd like to take a look: https://console.dps.mow-dev.cloudera.com/cloud/workloads/details/psomogyi-aws-rest/events

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
petersomogyi: Nice catch @risdenk ! Removed that parameter too.

petersomogyi:
Rebased commits to retrigger builds.

petersomogyi:
@risdenk can you merge this commit? I don't have rights for it.

risdenk:
:) looks like the other "kris..." beat me to it :P

PR #8520, opened 2020-07-09
CB-7807 remove regions from environment request becuase it is a dupli…

doktoric:
…cated information with the location object. We only need the location part. The integration tests were updated.


environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -248,6 +249,15 @@ private LocationDto locationRequestToDto(LocationRequest location) {
                 .build();
     }
 
+    private Set<String> locationRequestToRegions(LocationRequest location, String cloudPlatform) {
+        if (AWS.name().equals(cloudPlatform)) {
```
topolyai5, line 253: You can merge the `if` with `else if`

PR #8521, opened 2020-07-09
CB-7517 Fix Azure db stack template render issue with tags

Bajzathd:
Revert of revert of CB-7517, and a fix for the cause of the revert


Bajzathd:
retest this please

PR #8527, opened 2020-07-09
CB-7618: Eliminate hdf and hdp from image management

jszeles:
HDP and HDF images are not supported therefore we have eliminated them and their all references from image catalog service, unit tests and the related test json files.


jszeles:
retest it please

jszeles:
retest this please

PR #8531, opened 2020-07-09
CB-7744 passing in cloud paths to SDX.

orngejaket:
This adds in passing in the cloud paths to Ranger during CM template
construction for SDX.  This is only on 7.2.2 and Azure for now.


orngejaket:
CDH/CM has been bumped to 7.2.2, so now it's good to merge.

PR #8534, opened 2020-07-10
CB-7706 Verify tags on cloud provider side

Bajzathd:
Extend tag verification so that it does not jut verify tags on CB side, but actually query the cloud provider that resources have the needed tags. Supported cloud providers are Aws and Azure.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/environment/EnvironmentStopStartTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/environment/EnvironmentStopStartTests.java
@@ -1,10 +1,13 @@
 package com.sequenceiq.it.cloudbreak.testcase.e2e.environment;
 
+import java.util.Map;
+
 import javax.inject.Inject;
 
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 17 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/environment/EnvironmentStopStartTests.java
```diff
@@ -50,24 +62,30 @@ public void testCreateStopStartEnvironment(TestContext testContext) {
...
+                    .addTags(ENV_TAGS)
                 .when(environmentTestClient.create())
                 .await(EnvironmentStatus.AVAILABLE)
+                .then(cloudProviderSideTagAssertion.verifyEnvironmentTags(ENV_TAGS))
                 .given(SdxInternalTestDto.class)
+                    .addTags(SDX_TAGS)
                 .when(sdxTestClient.createInternal())
                 .awaitForFlow(RunningParameter.key(resourcePropertyProvider().getName()))
                 .await(SdxClusterStatusResponse.RUNNING)
+                .then(cloudProviderSideTagAssertion.verifyInternalSdxTags(SDX_TAGS))
```
lnardai, line 80: Can we also verify here the Tags that are getting added by Default fro the Account Level tag configuration?
Like the default Owner tag.

PR #8536, opened 2020-07-10
CB-7902 add /hive2 and /cliservice to support PowerBI driver to conne…

keyki:
…ct through Nginx

PowerBI drivers use 2 default hard-coded paths to contact HiveServer2: /cliservice and /hive2.
CDW forwards all these paths to the corresponding Hive install through Nginx, because the
cluster name is encoded in the hostname. Knox encodes the endpoint information in the HTTP path
and it is not possible to override it in PowerBI drivers.
A proxy_pass through nginx will result in Knox rejecting /cliservice or /hive2 and it does not
work out of the box.


keyki:
> Looks good on the surface. Might want to have @t3rmin4t0r try it on a cluster manually updating the nginx config to confirm.

Yep, good idea. @t3rmin4t0r can you do that?

PR #8540, opened 2020-07-10
CDPCP-2566. Do not try to start metering application on DataLake cluster (every 5 sec)

oleewere:
as patch was reverted because of a command issue (missing echo):
- if metering is not enabled make sure the service is dead
- fix os/systed error messages - > fix missing echo from the wrong commands
+ add the metering configuration file for watch list for the service


oleewere:
retest this please

PR #8541, opened 2020-07-13
On-demand log collection for FreeIpa

attilapalfi92:


oleewere:
Descriptions from git log:

CB-7397. Implement filecollector on CB side (freeipa/datalake/datahub) for on demand diagnostic collection

    details:
    - Creating a new "filecollector" module in salt definitions (salt-common)
    - Creating a new "telemetry" module in salt definitions which can be used to store common configurations between different telemtry components (fluent,monitoring,filecollector)
    (as salt upgrade has not done yet properly for freeipa, we still uses many duplicated configs in telemetry and fluent pillars)
    - Basic usage of filecollector: there is a filecollector.py and a specific yaml configuration that is deployed/configured by salt
    - filecollection has 4 different outputs (only first 3 is implemented): local (collection only, you need to ssh into the node for that), customer cloud storage and ENG Kibana (by databus)
    - for the common configurations vm_logs.json files are creadted, in the fututre that will be used as a common source for both fluentd (monitored files) and filecollector (files that are needed to be collected)
    - also added for some default converters and services for VM logs or creating data as a map for the diagnostics flow

CB-7484. Create diagnostics collection flow (FreeIPA)

    details:
    - creating a new flow (+ API) for freeipa log collection
    - the new api has 2 endpoint: POST that starts the diagnostics flow, GET which shows files that will be collected
    - flow has 4 phases: init, collect, upload and cleanup
    - each phase executes a salt state in filecollector module like: filecollector.init or filecollector.collect etc.
    - extend salt java module to add pillar parameters for the commands and run executions in concurrent mode (see ParametrizedStateRunner and CuncurrentParameterizedStateRunner)
    - refactor hostOrchestrator a bit by extending telemetry related operations and create stack service for commonly used operations
    - add / fix some unit tests

Some other notes:
- there is a cloud storage upload script, which is a bit similar as freeipa backup scripts, in the future, we would probably need one common scripts for both, but until no good salt upgrade mechanism on freeipa we will use that (also note that no any encryption is used around there yat)
- no real validation on the request (i mean there are null checks, but like real validation, like if clusterLogCollection is disabled, you should not be able to use ENG destination for log collection) -> will be done if there will be any UI development ready

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/init.sls
```diff
@@ -0,0 +1,88 @@
+{%- from 'telemetry/settings.sls' import telemetry with context %}
+{%- from 'filecollector/settings.sls' import filecollector with context %}
+{%- from 'fluent/settings.sls' import fluent with context %}
+
+install_fluent_logger:
+  cmd.run:
+    - name: pip install fluent-logger>=0.9.6 --ignore-installed
```
keyki, line 7: We need new images with these packages otherwise the no-internet test will fail and it will be a blocker for promotion.

oleewere: these are not in highstate (you need to trigger this tasks by API) + already inside the images (see: https://github.com/hortonworks/cloudbreak-images/commit/67bec331ea94a47626f8297d338aaa0226cd07bb), will be removed later

keyki:
I haven't looked at the code yet but tried the collection on an HA IPA cluster and it worked fine.

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cloud_storage_upload.sh.j2
```diff
@@ -0,0 +1,56 @@
...
+     log "Upload file to cloud storage is not supported from platform '${platform}'"
+   fi
+}
+
+function upload_to_s3() {
+  local filename=$1
+  s3_filename="$(basename -- $filename)"
+  s3_file_path="{{ filecollector.s3BaseUrl }}/${s3_filename}"
+  # --sse AES256
+  /usr/bin/aws s3 cp --no-progress "${filename}" "${s3_file_path}"
```
keyki, line 30: We will need to deal with encryption + the region must be a parameter otherwise it won't work with S3 VPC endpoint

orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
```diff
@@ -140,6 +140,8 @@
   {% set proxy_full_url = None %}
 {% endif %}
 
+{% set forward_port = 24224 %}
```
keyki, line 143: Can this port change?

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/init.sls
```diff
@@ -0,0 +1,88 @@
...
+  cmd.run:
+    - name: pip install pid --ignore-installed
+    - unless: pip list --no-index | grep -E 'pid'
+
+/opt/filecollector:
+  file.directory:
+    - name: /opt/filecollector
+    - user: "root"
+    - group: "root"
+    - mode: 740
```
keyki, line 25: There are lots of files being managed and have 740 and 750 mode. Do any of these files contain sensitive data or should we use 700 and 600 regardless? Why would a regular user need to read these files?

keyki:
I think this change should be tested with:
* Create a new environment without these changes
* Invoke `dp cluster update-salt` to refresh the pillars and states and try to run a `highstate` to make sure we're not removing or adding something which is not available on existing clusters.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
```diff
@@ -84,7 +84,7 @@
 {% set cloudera_public_gem_repo = 'https://repository.cloudera.com/cloudera/api/gems/cloudera-gems/' %}
 {% set cloudera_azure_plugin_version = '1.0.1' %}
 {% set cloudera_azure_gen2_plugin_version = '0.3.1' %}
-{% set cloudera_databus_plugin_version = '1.0.4' %}
+{% set cloudera_databus_plugin_version = '1.0.5' %}
```
keyki, line 87: For this we will need a new image or it will break the no-internet test again. This will run on a regular highstate.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
oleewere: that's typically fluentd port, only needs to be changed if some reason someone setup a fluentd with recipes and using that port

oleewere:
> I think this change should be tested with:
> 
> * Create a new environment without these changes
> * Invoke `dp cluster update-salt` to refresh the pillars and states and try to run a `highstate` to make sure we're not removing or adding something which is not available on existing clusters.

is that for stack api right? my guess there is no such thing for freeipa yet.
because of this we have many duplicated configs at the moment. (only those were removed which were not used yet, like monitoring configs)

orchestrator-salt/src/main/resources/salt-common/salt/fluent/settings.sls
oleewere: yes we will need new images for 2.26

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cloud_storage_upload.sh.j2
oleewere: i have mentioned in the PR, encryption is not yet handled. will be done later, no UI or tests around that so that can be done later. btw, I will work on a binary lib that contains these commands with enhanced options, so that won't be final anyway (as we will have a lot of different scripts, it's better to put/use them together)

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/init.sls
oleewere: yes it should be 640, will fix that in a fix commit, but I will wait until Lajos's review, in oder to fix his opened issues together with this

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cloud_storage_upload.sh.j2
keyki: What about VPC endpoints? Do we have a jira for that as well? I'm fine if we address that later, but let's have a jira so we don't forget.

PR #8543, opened 2020-07-13
CB-7853 Too broad access rights to recipes

afarsang:
recipes access rights only for root group in /salt/srv as well in /opt/


orchestrator-salt/src/main/resources/salt/salt/recipes/post-cloudera-manager-start.sls
```diff
@@ -14,7 +14,7 @@ create_recipe_log_dir_post_start:
        - salt://pre-recipes/scripts/{{ script_name }}
        - salt://pre-recipes/scripts/pre-date.sh
      - makedirs: True
-     - mode: 755
+     - mode: 700
```
doktoric, line 17: shouldn't this be 0600 ?

afarsang: non - executable?

PR #8545, opened 2020-07-14
CB-7741: Allow upgrade with same runtime and different GBN

tiborpopovics:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeImageFilter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeImageFilter.java
@@ -32,6 +32,10 @@ public class ClusterUpgradeImageFilter {
 
     private static final String SALT_PACKAGE_KEY = "salt";
 
+    private static final String CDH_BUILD_NUMBER_KEY = "cdh-build-number";
+
+    private static final String CM_BUILD_NUMBER_KEY = "cm-build-number";
+
     @Inject
     private VersionBasedImageFilter versionBasedImageFilter;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeImageFilter.java
```diff
@@ -109,9 +116,11 @@ private boolean permitLockedComponentsUpgrade(Image currentImage, Image image) {
         return currentImagePackageVersions.equals(imagePackageVersions);
     }
 
-    private boolean permitCmAndStackUpgrade(Image currentImage, Image image, String key) {
-        return upgradePermissionProvider.permitCmAndStackUpgrade(currentImage.getPackageVersions().get(key),
-                image.getPackageVersions().get(key));
+    private boolean permitCmAndStackUpgrade(Image currentImage, Image image, String versionKey, String buildNumberKey) {
+        String currentVersion = currentImage.getPackageVersions().get(versionKey);
```
pdarvasi, line 120: pls do this in an NPE safe way

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeImageFilter.java
```diff
@@ -109,9 +116,11 @@ private boolean permitLockedComponentsUpgrade(Image currentImage, Image image) {
...
+        String newVersion = image.getPackageVersions().get(versionKey);
```
pdarvasi, line 121: pls do this in an NPE safe way

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeImageFilter.java
```diff
@@ -109,9 +116,11 @@ private boolean permitLockedComponentsUpgrade(Image currentImage, Image image) {
...
+        return upgradePermissionProvider.permitCmAndStackUpgrade(currentVersion, newVersion)
+                || (currentVersion.equals(newVersion) && componentBuildNumberComparator.compare(currentImage, image, buildNumberKey));
```
pdarvasi, line 123: logically, this should be part of permitCmAndStackUpgrade method otherwise, it might be misleading, couldn't it?

PR #8546, opened 2020-07-14
CDPCP-1941. Fix cloud identity sync filter

aarman-cloudera:
CloudIdentitySyncService was filtering by the identity domain, but
this field may be null. This is fixed by filtering based on cloud
identity name instead. This commit also adds unit tests for
CloudIdentitySyncService.

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
@@ -6,6 +6,7 @@ import io.swagger.annotations.ApiModelProperty;
 
 import javax.validation.constraints.NotNull;
 import java.util.Map;
+import java.util.Objects;
 
 @ApiModel
 @JsonIgnoreProperties(ignoreUnknown = true)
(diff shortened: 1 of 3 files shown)


datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SetRangerCloudIdentityMappingRequest.java
```diff
@@ -42,4 +43,22 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        SetRangerCloudIdentityMappingRequest that = (SetRangerCloudIdentityMappingRequest) o;
+        return azureUserMapping.equals(that.azureUserMapping) &&
```
handavid, line 56: Should we use Objects.equals for this to be safe? The current code assumes that the fields are not null. I see the fields are annotated @NotNull, but that just affects validation.

aarman-cloudera: True, I've updated the equals to be null safe.

aarman-cloudera:
> Could you move this to CB-2.26.0 branch?

Okay, changed base to CB-2.26.0

lacikaaa:
retest this please

PR #8547, opened 2020-07-14
CB-7443. Changes to make some improvements to the time it takes to scale

sidseth:
up nodes.
- Generally lowers polling intervals
- Adds a SlowStart poller which, for operations which are expected to
take time, polls at a slower interval earlier, and then gets more
aggressive.
- PostCM and PostClusterInstall recipes only executed if they exist.
- Protected various expensive commands during highState via prewarmed
checks via the 'roles' grain.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
@@ -38,6 +38,7 @@ import com.sequenceiq.cloudbreak.cloud.aws.CloudFormationStackUtil;
 import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonAutoScalingRetryClient;
 import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonCloudFormationRetryClient;
 import com.sequenceiq.cloudbreak.cloud.aws.scheduler.CustomAmazonWaiterProvider;
+import com.sequenceiq.cloudbreak.cloud.aws.scheduler.SlowStartCancellablePollingStrategy;
 import com.sequenceiq.cloudbreak.cloud.aws.scheduler.StackCancellationCheck;
 import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
@@ -50,6 +51,8 @@ public class AwsAutoScalingService {
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsAutoScalingService.class);
 
     private static final int MAX_INSTANCE_ID_SIZE = 100;
+
+    private static final int EXPECTED_SCALEUP_TIME_SECONDS = 40;
 
     @Inject
     private CloudFormationStackUtil cfStackUtil;
@@ -116,6 +119,8 @@ public class AwsAutoScalingService {
             String autoScalingGroupName, Integer requiredInstanceCount, Long stackId) throws AmazonAutoscalingFailed {
         Waiter<DescribeAutoScalingGroupsRequest> groupInServiceWaiter = asClient.waiters().groupInService();
         PollingStrategy backoff = getBackoffCancellablePollingStrategy(new StackCancellationCheck(stackId));
+        PollingStrategy slowStart = SlowStartCancellablePollingStrategy.getExpectedRuntimeCancellablePollingStrategy(
+                new StackCancellationCheck(stackId), EXPECTED_SCALEUP_TIME_SECONDS);
         try {
             groupInServiceWaiter.run(new WaiterParameters<>(new DescribeAutoScalingGroupsRequest().withAutoScalingGroupNames(autoScalingGroupName))
                     .withPollingStrategy(backoff));
@@ -123,14 +128,16 @@ public class AwsAutoScalingService {
             throw new AmazonAutoscalingFailed(e.getMessage(), e);
         }
 
+        long startTime = System.currentTimeMillis();
         Waiter<DescribeAutoScalingGroupsRequest> instancesInServiceWaiter = customAmazonWaiterProvider
                 .getAutoscalingInstancesInServiceWaiter(asClient, requiredInstanceCount);
         try {
             instancesInServiceWaiter.run(new WaiterParameters<>(new DescribeAutoScalingGroupsRequest().withAutoScalingGroupNames(autoScalingGroupName))
-                    .withPollingStrategy(backoff));
+                    .withPollingStrategy(slowStart));
         } catch (Exception e) {
             throw new AmazonAutoscalingFailed(e.getMessage(), e);
         }
+        LOGGER.info("Done with instancesInServiceWaiter. Duration={}", System.currentTimeMillis() - startTime);
 
         List<String> instanceIds = cloudFormationStackUtil.getInstanceIds(asRetryClient, autoScalingGroupName);
         if (requiredInstanceCount != 0) {
@@ -194,7 +201,7 @@ public class AwsAutoScalingService {
                 .withAutoScalingGroupName(groupName)
                 .withMaxSize(newSize)
                 .withDesiredCapacity(newSize));
-        LOGGER.debug("Updated Auto Scaling group's desiredCapacity: [to: '{}']", newSize);
+        LOGGER.debug("Updated '{}' Auto Scaling group's desiredCapacity: [to: '{}']", groupName, newSize);
     }
 
     public void terminateInstance(AmazonAutoScalingRetryClient amazonASClient, String instanceId) {
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
@@ -34,6 +34,7 @@ public class SlowStartDelayStrategy implements PollingStrategy.DelayStrategy {
     private final int expectedRuntimeSeconds;
 
     private final long msToWaitRequest0;
+
     private final long msToWaitRequest1;
 
     public SlowStartDelayStrategy(int expectedRuntimeSeconds) {
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
@@ -1,0 +1,70 @@
+package com.sequenceiq.cloudbreak.cloud.aws.scheduler;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.amazonaws.waiters.PollingStrategy;
+import com.amazonaws.waiters.PollingStrategyContext;
+
+/**
+ * A polling strategy which is less aggressive early, and gets more aggressive once the
+ * provided expectedRuntime is reached.
+ */
+public class SlowStartDelayStrategy implements PollingStrategy.DelayStrategy {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SlowStartDelayStrategy.class);
+
+    private static final Random RANDOM = ThreadLocalRandom.current();
+
+    private static final int MAX_POLLING_INTERVAL_SECONDS = 60;
+
+    private static final int STABLE_POLLING_INTERVAL_SECONDS_MAX = 5;
+
+    private static final int STABLE_POLLING_INTERVAL_SECONDS_MIN = 1;
+
+    private static final double FIRST_POLL_RATIO = 0.75;
+
+    private static final double SECOND_POLL_RATIO = 0.2;
+
+    private static final double THOUSAND = 1000.0;
+
+    private final int expectedRuntimeSeconds;
+
+    private final long msToWaitRequest0;
+    private final long msToWaitRequest1;
+
+    public SlowStartDelayStrategy(int expectedRuntimeSeconds) {
+        this.expectedRuntimeSeconds = expectedRuntimeSeconds;
+        msToWaitRequest0 = (long) (THOUSAND * Math.min(
+                Math.max(FIRST_POLL_RATIO * expectedRuntimeSeconds, STABLE_POLLING_INTERVAL_SECONDS_MIN),
+                MAX_POLLING_INTERVAL_SECONDS));
+        msToWaitRequest1 = (long) (THOUSAND * Math.min(
+                Math.max(SECOND_POLL_RATIO * expectedRuntimeSeconds, STABLE_POLLING_INTERVAL_SECONDS_MIN),
+                MAX_POLLING_INTERVAL_SECONDS));
+    }
+
+    @Override
+    public void delayBeforeNextRetry(PollingStrategyContext pollingStrategyContext) throws InterruptedException {
+        int requestNumber = pollingStrategyContext.getRetriesAttempted();
+
+        // Start polling at 75% of minExpectedRuntime.
+        // Next at 20% of minExpectedRuntime
+        // After this random between min and max
+
+        long msToWait;
+        if (requestNumber == 0) {
+            msToWait = msToWaitRequest0;
+        } else if (requestNumber == 1) {
+            msToWait = msToWaitRequest1;
+        } else {
+            msToWait = (long) (THOUSAND * Math.max(STABLE_POLLING_INTERVAL_SECONDS_MIN,
+                    RANDOM.nextInt(STABLE_POLLING_INTERVAL_SECONDS_MAX)));
+        }
+        LOGGER.info("Polling attempt: {}, Sleeping for: {}ms for request: {}. ExpectedRuntime={}s",
+                requestNumber, msToWait, pollingStrategyContext.getOriginalRequest(), expectedRuntimeSeconds);
+        Thread.sleep(msToWait);
+    }
+}
--- a/cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/UpscaleStackHandler.java
+++ b/cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/UpscaleStackHandler.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.cloud.handler;
 
+import java.time.Duration;
+import java.time.Instant;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -52,6 +54,7 @@ public class UpscaleStackHandler implements CloudPlatformEventHandler<UpscaleSta
     @Override
     public void accept(Event<UpscaleStackRequest> upscaleStackRequestEvent) {
         LOGGER.debug("Received event: {}", upscaleStackRequestEvent);
+        Instant start = Instant.now();
         UpscaleStackRequest<UpscaleStackResult> request = upscaleStackRequestEvent.getData();
         CloudContext cloudContext = request.getCloudContext();
         try {
@@ -69,10 +72,13 @@ public class UpscaleStackHandler implements CloudPlatformEventHandler<UpscaleSta
             eventBus.notify(result.selector(), new Event<>(upscaleStackRequestEvent.getHeaders(), result));
             LOGGER.debug("Upscale successfully finished for {}", cloudContext);
         } catch (Exception e) {
-            LOGGER.error("Upscaling stack failed", e);
+            LOGGER.error("Upscaling stack failed for {}", cloudContext, e);
             UpscaleStackResult result = new UpscaleStackResult(e.getMessage(), e, request.getResourceId());
             request.getResult().onNext(result);
             eventBus.notify(CloudPlatformResult.failureSelector(UpscaleStackResult.class), new Event<>(upscaleStackRequestEvent.getHeaders(), result));
+        } finally {
+            LOGGER.debug("UpscaleStackRequest for event: [{}] finished in {}ms", upscaleStackRequestEvent,
+                    Duration.between(start, Instant.now()).toMillis());
         }
     }
 
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
@@ -8,6 +8,8 @@ import static com.sequenceiq.cloudbreak.polling.PollingResult.isExited;
 import static com.sequenceiq.cloudbreak.polling.PollingResult.isTimeout;
 
 import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -133,6 +135,10 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
     @Override
     public List<String> upscaleCluster(HostGroup hostGroup, Collection<InstanceMetaData> instanceMetaDatas)
             throws CloudbreakException {
+        LOGGER.debug("Invoking CM requests for upscale for hostGroup={}, insteanceCount={}",
+                hostGroup.getName(), instanceMetaDatas == null ? 0 : instanceMetaDatas.size());
+        int hostCount = -1;
+        Instant start = Instant.now();
         ClustersResourceApi clustersResourceApi = clouderaManagerApiFactory.getClustersResourceApi(apiClient);
         try {
             String clusterName = stack.getName();
@@ -140,6 +146,7 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
             if (!upscaleHostNames.isEmpty()) {
                 List<ApiHost> hosts = clouderaManagerApiFactory.getHostsResourceApi(apiClient).readHosts(null, null, SUMMARY).getItems();
                 ApiHostRefList body = createUpscaledHostRefList(upscaleHostNames, hosts);
+                hostCount = body.getItems().size();
                 clustersResourceApi.addHosts(clusterName, body);
                 activateParcel(clustersResourceApi);
                 applyHostGroupRolesOnUpscaledHosts(body, hostGroup.getName());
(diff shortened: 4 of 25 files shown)


sidseth:
Works along with https://github.com/hortonworks/cloudbreak-images/pull/426, but both of these PRs can work independently.
Tested against CB-2.25-b124. Will tests against master after fixing review comments.

@oleewere - I've introduced a separate parameter for fluent. I think the concerns you raised on the other PR should be addressed via this. If a version changes, change the revision of the parameter here and in image creation.

sidseth:
From my limited understanding of upgrades, I don't think this will cause issues there (I believe upgrades always start with an image being replaced - so skipping installs of components should not make a difference). Any pointers on how to go about testing upgrades? - would be good to test them out anyway

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
```diff
@@ -4,7 +4,7 @@
 {% set dbus_lock_exists = salt['file.file_exists' ]('/etc/td-agent/databus_bundle.lock') %}
 
 {% if fluent.enabled %}
-
+{% if not fluent.fluent_prewarm_role in grains.get('roles', []) %}
```
oleewere, line 7: i do not think that is needed as if the /etc/td-agent folder exists, it's already sure the td-agent service is installed.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
```diff
@@ -64,13 +64,17 @@ function check_and_install_plugin() {
 }
 
 function main() {
+{% if not fluent.fluent_prewarm_role in grains.get('roles', []) %}
```
oleewere, line 67: no need extra check for the script if there is a check around creating this template + executing it

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
```diff
@@ -86,6 +90,7 @@ install_fluentd_plugins:
       - group
       - mode
 
+{% if not fluent.fluent_prewarm_role in grains.get('roles', []) %}
```
oleewere, line 93: my problem here is that i described in https://github.com/hortonworks/cloudbreak/pull/8457#discussion_r452182829
It will mean many extra support hours for us as we will hide the root cause of version incompatibility based on the pre-warm image that is used.

my suggestion for this instead pre warm role check, create a new field in image catalog (metadata), same as package_versions, (let's say fluentd_plugin_versions), then you can pass those image values as salt pillars, and you can check the versions in fluent/settings.sls -> if any of the versions are different (between CB salt script constant <-> pillar), then configure and run this script.

cc @horadla23 @keyki

horadla23: I'm not comfortable with the solution of extending the metadata of image in catalog (since it requires lot of modification in various repositories thus it is error prune), but if that is the best we can do, I'm not against it. 

To do that, you should:
- add logic in cloudbreak-images repository to generate map of fluent_plugin_versions map into metadata which is passed to cb-image-catalog-cli
- extend model in cb-image-catalog-cli and update logic regarding metadata/image copy
- extend models in cloudbreak
- implement logics regarding that map in cloudbreak

please if it possible do not modify package-versions, since it is responsible for change image/OS upgrade, i suggest a new field for this

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
```diff
@@ -46,6 +47,8 @@
 
     private static final int MAX_INSTANCE_ID_SIZE = 100;
 
+    private static final int EXPECTED_SCALEUP_TIME_MIN_SECONDS = 40;
```
keyki, line 50: MIN_SECONDS, is it a type IN_SECONDS?

keyki:
> From my limited understanding of upgrades, I don't think this will cause issues there (I believe upgrades always start with an image being replaced - so skipping installs of components should not make a difference). Any pointers on how to go about testing upgrades? - would be good to test them out anyway

The upgrade starts with an in-place upgrade. Replacing the image is the last step when everything has been upgraded. Here's the wiki about upgrade: https://cloudera.atlassian.net/wiki/spaces/ENG/pages/488638097/SDX+Upgrade. You can test with going from 7.2.0 to 7.2.1

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
sidseth: Had explicitly skipped the option of trying to add these to the image catalog. There's multiple modules associated with fluent which would need to be versioned, and the image catalog changes are far more disruptive.
The current solution is quite a bit simpler and seems adequate. Quoting from the other PR

> 1. if we have new image in production but a new CB is not deployed - version conflict (plugins)
The image, if fluent changes are required, would have changed the version on the fluent role. e.g. "fluent_prewarmed_v2" instead of "fluent_prewarmed_v1". Since CB has not been upgraded, the checks would fail (i.e. result is false), and regular attempts would be made to install fluent (i.e. the >= check will execute and effectively be a no-op). No different from what happens today.

> 2. New CB deployed, a node uses an older image (it happens often) - version conflict (plugins)
In this case, CB would have changed the fluent role in settings.sls from e.g. "fluent_prewarmed_v1" to "fluent_prewarmed_v2". Older images would still have "fluent_prewarmed_v1". The check would fail (i.e. result is false), and all the package checks would execute as they do today.

If both CB and the image are upgraded - we get the benefit of a faster install, which will likely be the majority scenario.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
sidseth: Don't think that is the case. The dir check also wraps the plugin installation, which runs without this change.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/template/check_fluent_plugins.sh.j2
sidseth: Will remove this.

sidseth:
Uploaded another patch with review comments addressed, unit tests and checkStle warnings fixed. @keyki  - upgrade testing isn't done since new environment creation is currently failing on private stacks.

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
oleewere: but because of that is wrapped inside the file check, it only runs if the dir is not there (so only runs for base images)

sidseth: That is not my observation.

sidseth:
The integration tests failed with a swagger incompatibility, which isn't related to this patch. Tried a re-run via Jenkins.
Is there a mechanism to re-trigger tests on the github PR itself?

keyki:
> The integration tests failed with a swagger incompatibility, which isn't related to this patch. Tried a re-run via Jenkins.
> Is there a mechanism to re-trigger tests on the github PR itself?

Yes, If you go here and login you can retry: http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-pull-request-builder-integration-test/10764/

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -128,13 +130,17 @@ public void initApiClient() throws ClusterClientInitException {
     @Override
     public List<String> upscaleCluster(HostGroup hostGroup, Collection<InstanceMetaData> instanceMetaDatas)
             throws CloudbreakException {
+        LOGGER.debug("Invoking CM requests for upscale");
```
lacikaaa, line 133: if we add logging here, maybe we should print the parameters

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -128,13 +130,17 @@ public void initApiClient() throws ClusterClientInitException {
...
+        int hostCount = -1;
```
lacikaaa, line 134: this is a bit odd. I think it would make sense to use Integer and we can log `null` in that case, or use `Optional<Integer>` and use `isPresent` when logging

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
```diff
@@ -114,7 +114,8 @@ public void installServicesOnNewHosts(Long stackId, String hostGroupName) throws
     public void executePostRecipesOnNewHosts(Long stackId) throws CloudbreakException {
         Stack stack = stackService.getByIdWithListsInTransaction(stackId);
         LOGGER.debug("Start executing post recipes");
-        recipeEngine.executePostInstallRecipes(stack);
+        // TODO: Question for reviewer: stackId vs clusterId - Is this usage correct?
```
lacikaaa, line 117: clusterId it is

core/src/main/java/com/sequenceiq/cloudbreak/reactor/BootstrapNewNodesHandler.java
```diff
@@ -28,13 +36,18 @@ public String selector() {
 
     @Override
     public void accept(Event<BootstrapNewNodesRequest> event) {
+        int numHosts = event.getData().getHostNames() == null ? -1 : event.getData().getUpscaleCandidateAddresses().size();
```
lacikaaa, line 39: you can use `Optional.fromNullable` here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/BootstrapNewNodesHandler.java
```diff
@@ -28,13 +36,18 @@ public String selector() {
...
+        LOGGER.debug("BootstrapNewNodes for #hosts: {}", numHosts);
+        Instant start = Instant.now();
         BootstrapNewNodesRequest request = event.getData();
         BootstrapNewNodesResult result;
         try {
             clusterBootstrapper.bootstrapNewNodes(request.getResourceId(), request.getUpscaleCandidateAddresses(), request.getHostNames());
             result = new BootstrapNewNodesResult(request);
         } catch (Exception e) {
             result = new BootstrapNewNodesResult(e.getMessage(), e, request);
+        } finally {
```
lacikaaa, line 49: `finally` could be omitted

core/src/main/java/com/sequenceiq/cloudbreak/reactor/BootstrapNewNodesHandler.java
lacikaaa: also `event.getData().getHostNames()` is checked if null, but `event.getData().getUpscaleCandidateAddresses()` size is used, so something is off here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/UpscaleClusterHandler.java
```diff
@@ -28,13 +36,17 @@ public String selector() {
...
+        LOGGER.debug("UpscaleClusterHandler for {}", event.getData().getResourceId());
+        Instant start = Instant.now();
         UpscaleClusterRequest request = event.getData();
         UpscaleClusterResult result;
         try {
             clusterUpscaleService.installServicesOnNewHosts(request.getResourceId(), request.getHostGroupName());
             result = new UpscaleClusterResult(request);
         } catch (Exception e) {
             result = new UpscaleClusterResult(e.getMessage(), e, request);
+        } finally {
```
lacikaaa, line 48: `finally` could be omitted

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/UpscaleClusterManagerHandler.java
```diff
@@ -37,6 +46,8 @@ public void accept(Event<UpscaleClusterManagerRequest> event) {
             result = new UpscaleClusterManagerResult(request);
         } catch (Exception e) {
             result = new UpscaleClusterManagerResult(e.getMessage(), e, request);
+        } finally {
```
lacikaaa, line 49: `finally` could be omitted

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/RecipeEngine.java
```diff
@@ -56,16 +53,16 @@ public void executePreClusterManagerRecipes(Stack stack, Collection<Recipe> reci
         }
     }
 
-    // note: executed when LDAP config is present, because later the LDAP sync is hooked for this salt state in the top.sls.
     public void executePostAmbariStartRecipes(Stack stack, Collection<Recipe> recipes) throws CloudbreakException {
-        if ((stack.getCluster() != null && ldapConfigService.isLdapConfigExistsForEnvironment(stack.getEnvironmentCrn(), stack.getName()))
-                || recipesFound(recipes, POST_CLOUDERA_MANAGER_START)) {
+        if (shouldExecuteRecipeOnStack(recipes, POST_CLOUDERA_MANAGER_START)) {
```
lacikaaa, line 57: is this condition is not valid anymore?
`ldapConfigService.isLdapConfigExistsForEnvironment(stack.getEnvironmentCrn(), stack.getName())`

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/RecipeEngine.java
```diff
@@ -56,16 +53,16 @@ public void executePreClusterManagerRecipes(Stack stack, Collection<Recipe> reci
...
             orchestratorRecipeExecutor.postClusterManagerStartRecipes(stack);
         }
     }
 
-    public void executePostInstallRecipes(Stack stack) throws CloudbreakException {
-        orchestratorRecipeExecutor.postClusterInstall(stack);
+    public void executePostInstallRecipes(Stack stack, Collection<Recipe> recipes) throws CloudbreakException {
+        if (shouldExecuteRecipeOnStack(recipes, POST_CLUSTER_INSTALL)) {
```
lacikaaa, line 63: if you skip this step, this script won't run: https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
from here: https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/recipes/post-cluster-install.sls#L32-L46

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
```diff
@@ -45,7 +45,7 @@
 
     private static final String ROLE_NAME_PREFIX = "hadoopadminrole-";
 
-    private static final int POLL_INTERVAL = 5000;
+    private static final int POLL_INTERVAL = 2000;
```
lacikaaa, line 48: I wouldn't change this, as it gains only 3 sec at best and increase the polling intensity by 2.5

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -91,6 +91,14 @@
 
     private static final int SLEEP_TIME = 10000;
 
+    private static final int SLEEP_TIME_MS_1000 = 1000;
+
+    private static final int SLEEP_TIME_MS_2000 = 2000;
+
+    private static final int SLEEP_TIME_MS_3000 = 3000;
+
+    private static final int SLEEP_TIME_MS_4000 = 4000;
```
lacikaaa, line 100: our opinion is that the min sleep time should be 5 sec. it's currently 10 so it still would double the poll count

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
lacikaaa: @oleewere so what do you think about this?

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
lacikaaa: @oleewere what is the agreement here?

orchestrator-salt/src/main/resources/salt/salt/unbound/init.sls
```diff
@@ -24,10 +24,9 @@
       private_address: {{ host.private_address }}
 
 include_access_config:
-  file.replace:
-    - name: /etc/unbound/unbound.conf
-    - pattern: '#include: "otherfile.conf"'
-    - repl: 'include: "/etc/unbound/access.conf"'
+# file.replace tends to take 5 seconds when run for the first time due to tons of modules being loaded in salt 3000.2. Hence a sed.
```
lacikaaa, line 27: I think if it tends to take a one time toll on us it shouldn't be an issue as we have a bunch of `file.replace`. So we should eliminate all of them or leave this one here also. I'm voting for leaving this one as it's more readable and error proof

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+
+        Double msToWait;
+        if (requestNumber == 0) {
+            msToWait = THOUSAND * Math.min(
+                    Math.max(FIRST_POLL_RATIO * expectedRuntimeSeconds, STABLE_POLLING_INTERVAL_SECONDS_MIN),
+                    MAX_POLLING_INTERVAL_SECONDS);
+        } else if (requestNumber == 1) {
+            msToWait = THOUSAND * Math.min(
+                    Math.max(SECOND_POLL_RATIO * expectedRuntimeSeconds, STABLE_POLLING_INTERVAL_SECONDS_MIN),
+                    MAX_POLLING_INTERVAL_SECONDS);
```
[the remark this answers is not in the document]
sidseth: ok.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            msToWait = THOUSAND * Math.max(STABLE_POLLING_INTERVAL_SECONDS_MIN,
+                    RANDOM.nextInt(STABLE_POLLING_INTERVAL_SECONDS_MAX));
```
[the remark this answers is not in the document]
sidseth: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html is what I could find on AWS EC2 throttling (assuming the same applies for ASGs as well).
I'm not sure how many concurrent operations a single instance of CB is supposed to be able to handle (this would be new AWS cluster launches / scale operations).
How about setting this to a constant 5 seconds (or 5 seconds + jitter in the 1s range... so 4-6 seconds). Can even remove this class and rely on the default FixedDelayStrategy which is part of the AWS SDK.
The current Exponential backoff seems wrong though - since it gets less aggressive towards the time the ASG will be ready.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
```diff
@@ -111,28 +114,33 @@ private void waitForGroup(AmazonEC2Client amClient, AmazonAutoScalingClient asCl
...
         }
+        LOGGER.info("Done with instancesInServiceWaiter. Duration={}", System.currentTimeMillis() - startTime);
 
         List<String> instanceIds = cloudFormationStackUtil.getInstanceIds(asRetryClient, autoScalingGroupName);
         if (requiredInstanceCount != 0) {
             List<List<String>> partitionedInstanceIdsList = Lists.partition(instanceIds, MAX_INSTANCE_ID_SIZE);
 
             Waiter<DescribeInstancesRequest> instanceRunningStateWaiter = amClient.waiters().instanceRunning();
             for (List<String> partitionedInstanceIds : partitionedInstanceIdsList) {
+                startTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
sidseth: Will remove. Leftover from figuring out timing of a different op.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/UpscaleStackHandler.java
```diff
@@ -69,10 +72,13 @@ public void accept(Event<UpscaleStackRequest> upscaleStackRequestEvent) {
             eventBus.notify(result.selector(), new Event<>(upscaleStackRequestEvent.getHeaders(), result));
             LOGGER.debug("Upscale successfully finished for {}", cloudContext);
         } catch (Exception e) {
-            LOGGER.error("Upscaling stack failed", e);
+            LOGGER.error("Upscaling stack failed for {}", cloudContext, e);
             UpscaleStackResult result = new UpscaleStackResult(e.getMessage(), e, request.getResourceId());
             request.getResult().onNext(result);
             eventBus.notify(CloudPlatformResult.failureSelector(UpscaleStackResult.class), new Event<>(upscaleStackRequestEvent.getHeaders(), result));
+        } finally {
```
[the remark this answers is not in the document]
sidseth: This is more for making sure the message gets logged even if their are failures (e.g. how long did the operation take before the failure)
I believe there's an effort to get some of the structured information into a queryable form. The log lines help until more info can be added to that structured effort.

If you feel strongly about not having the log lines in a finally block - will remove them. (Applies to all comments about the finally block)

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
sidseth: makes sense.

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
sidseth: There are not a lot of big rocks in terms of trying to save time for these operations... which is why there's changes to polling intervals on several different operations (roughly based on how long the operations take).
In this case the operation for a single node takes about 3 seconds, and the that guarantees a couple of seconds lost since the first poll is immediate. Multiple nodes, OTOH, can take longer.

orchestrator-salt/src/main/resources/salt/salt/unbound/init.sls
sidseth: sure ...

sidseth:
> I'm not feeling comfortable with `{%- if not "prewarmed_v1" in grains.get('roles', []) -%}` checks as it requires to keep updated every time something changes in the images. I have a feeling it would make things more fragile than it's already
This change does end up skipping a lot of unnecessary operations, and likely shaves 30+ seconds off the highstate (more if the archive changes are in place during image building).

If we're going to skip this, and roll back all the timeouts - may as well abandon the patch.

There's a few big rocks like the prewarmed_v1 check, or the extract change. Everything else beyond this is tiny bits of time which add up across an upscale operation.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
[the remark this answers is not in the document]
sidseth: Also, as part of this one thing I wanted to check but did not get a chance to - does the AWS SDK have retries built in. There's several operations outside of polling which would fail if the AWS service were down intermittently.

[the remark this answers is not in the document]
lacikaaa: the main issue here is that if a lot of people share the same account it can easily trigger throttling error, if I remember well this is the reason why the backoff strategy was introduced. maybe @sodre90 remembers better

orchestrator-salt/src/main/resources/salt-common/salt/fluent/init.sls
oleewere: @lacikaaa i guess we keep this as is in the PR, although the check should be more smart, (so if the version is higher on the image, that should not be a problem, so we need to lookup in the grains and get the version part of fluent_prewarmed_v1 value, and compare with that)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/scheduler/SlowStartDelayStrategy.java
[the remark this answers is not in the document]
keyki: Yes, this isn't an issue with a single cluster. But on mow-dev ~100 people use 3 AWS accounts where it's a constant issue for us.

keyki:
@sidseth I merged the images counterpart. Can you please rebase this branch to the top of master?

sidseth:
Thanks for merging that in. Will rebase this one shortly.

lnardai:
Is this PR still needed?

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeIpaCleanupService.java
sidseth: Reverted to 5 seconds

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
sidseth: Switching to a single 5 second sleep.

orchestrator-salt/src/main/resources/salt/salt/unbound/init.sls
sidseth: Removed. The deletion of vsphere brings such salt operations from 5s to ~2s. This would take it further down to a few milliseconds. Can be addressed later.

sidseth:
Updated the patch based on review comments, and simplified. Should be ready for review/commit.
A single timeout in salt, and doubled the retry count.
The prewarmed checks are retained ... since that does save time, especially without the un-archiving of flies in the pre-warmed image.

Tested on top of 2.31-b111

sidseth:
@keyki

doktoric:
@keyki and @sidseth please check this

sidseth:
Closed. But left the branch in place for future reference - some of the changes here can still be incorporated.

PR #8549, opened 2020-07-14
CB-7952 Add cmUserName and cmPassword to the recipe template parameters

bbihari:


pdarvasi:
@bbihari Could you pls. change the base to 2.25.0?

PR #8552, opened 2020-07-14
Introduce Audit credential

doktoric:

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/credential/v1/AuditCredentialV1Controller.java
+++ b/environment/src/main/java/com/sequenceiq/environment/credential/v1/AuditCredentialV1Controller.java
@@ -1,0 +1,100 @@
+package com.sequenceiq.environment.credential.v1;
+
+import static com.sequenceiq.common.model.CredentialType.AUDIT;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.Valid;
+
+import org.springframework.stereotype.Controller;
+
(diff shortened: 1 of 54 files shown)


environment/src/main/java/com/sequenceiq/environment/credential/v1/AuditCredentialV1Controller.java
```diff
@@ -0,0 +1,106 @@
...
+        if (auditCredentialsByPlatfom.isEmpty()) {
+            return credentialConverter.convert(credentialService.create(credential, accountId, creator, onlyAudit()));
+        } else {
+            throw new BadRequestException(String.format("Audit credential already exist for %s cloud.", credential.getCloudPlatform()));
+        }
+    }
+
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.CREATE_CREDENTIAL)
+    public CredentialResponse put(@ResourceObject @Valid CredentialRequest credentialRequest) {
```
horadla23, line 92: this should be authorized against the existing credential resource, not on account level
to do that:
- add `@CheckPermissionByResourceObject` to the method
- add `@ResourceObject` to the credentialrequest parameter (it is there as i see)
- add `@ResourceObjectField(action = AuthorizationResourceAction.EDIT_CREDENTIAL, variableType = AuthorizationVariableType.NAME)` to the `name` field of the CredentialRequest class (I think this is done too since it is the same class as used in other crednetial controller)

so I think the method annotation should be changed only

PR #8558, opened 2020-07-15
CB-7940 do not create reverse SSH tunnel on CORE type of instances

keyki:
Due to a missing filter we're configuring the reverse SSH tunnel
on every single instance. This is unnecessary as we're only
communicating with the master instance. With this change only the
instances with GATEWAY type will create the tunnel. FMS has only
1 type of instances.


keyki:
retest this please

keyki:
retest this please

PR #8565, opened 2020-07-16
CB-7853 Too broad access rights to recipes

doktoric:


pdarvasi:
Thanks @doktoric Jozsef!

PR #8572, opened 2020-07-16
CB-7930 Use DataHubCreator role for DataHub creation authz

foldik:


authorization-common/src/main/resources/actions.json
```diff
@@ -77,6 +77,12 @@
     "actionType": "RESOURCE_DEPENDENT",
     "legacyRight": "environments/read"
   },
+  "ENVIRONMENT_CREATE_DATAHUB": {
+    "right": "environments/createDatahub",
+    "resourceType": "ENVIRONMENT",
+    "actionType": "RESOURCE_DEPENDENT",
```
horadla23, line 83: please rebase this to latest master and pls remove actionType field

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/DistroXV1Request.java
```diff
@@ -26,6 +29,7 @@
 @JsonInclude(Include.NON_NULL)
 public class DistroXV1Request extends DistroXV1Base implements TaggableRequest {
 
+    @ResourceObjectField(action = AuthorizationResourceAction.ENVIRONMENT_CREATE_DATAHUB, variableType = AuthorizationVariableType.NAME)
```
horadla23, line 32: looks good, but later we'll might have issue with this, if someone add `@CheckPermissionByResourceObject` for example for a getDistrox method, because it will check create_datahub also, but i know it is a gap in authz framework. 

I also would be happier, if we would have different request object for distroxt creation, but I know it is a design issue regarding API models.

PR #8573, opened 2020-07-16
CB-7805: Avoid automatic fallback to base images

jszeles:
The problem caused by that the CB automatically fallbacks to base images if there is no prewarmed. It hides problem and eventually customers will face with the problems.
Actually, the image catalog service contained the fallback logic, therefore it was removed from source code and the related unit tests were also updated.


akanto:
Looks good, but I think it might break the e2e test

PR #8580, opened 2020-07-17
CB-7957 Delete child envs when deleting parent with cascade

Bajzathd:
- Extend environment cascading delete to delete child environments
- Set environment delete cascade flag to false by default


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
```diff
@@ -65,7 +65,7 @@
     @ApiOperation(value = EnvironmentOpDescription.DELETE_BY_NAME, produces = MediaType.APPLICATION_JSON,
             notes = ENVIRONMENT_NOTES, nickname = "deleteEnvironmentV1ByName")
     SimpleEnvironmentResponse deleteByName(@PathParam("name") String environmentName,
-        @QueryParam("cascading") @DefaultValue("true") boolean cascading,
+        @QueryParam("cascading") @DefaultValue("false") boolean cascading,
```
doktoric, line 68: if it does not affect the qaas side then I am fine with it

Bajzathd: qaas side is ready for this change
https://jira.cloudera.com/browse/QAINFRA-11810

PR #8587, opened 2020-07-17
CB-8046 Remove filtering based on permissions

horadla23:

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
@@ -7,9 +7,6 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.validation.Valid;
 
-import com.sequenceiq.authorization.annotation.InternalOnly;
-import com.sequenceiq.datalake.cm.RangerCloudIdentityService;
-import com.sequenceiq.sdx.api.model.SetRangerCloudIdentityMappingRequest;
 import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.stereotype.Controller;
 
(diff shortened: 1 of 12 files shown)


horadla23:
pls merge it after this: https://github.infra.cloudera.com/thunderhead/thunderhead/pull/1146

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
```diff
@@ -127,7 +126,7 @@ public SdxClusterResponse getByCrn(@ResourceCrn String clusterCrn) {
     }
 
     @Override
-    @FilterListBasedOnPermissions(action = AuthorizationResourceAction.DESCRIBE_DATALAKE)
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.LIST_DATALAKES)
     public List<SdxClusterResponse> list(String envName) {
```
foldik, line 130: Just a note. The list datahub page shows the create datahub button if this list sdx endpoint gives back some datalakes. So after this change a user will see this button enabled and later on will figure out he doesn't have right to create datahub beacuse there is no datalake that he can use.

horadla23: Yes, this is the same as we can click on a datalake and later we may deny to describe it. I think we already discussed  disadvantages of this solution.
Anyway authorization of buttons and UI elements is ahead of us I think.

lnardai: DatahubCreator needs be able to "list" the environments in a particular environment with it's environments/listDatalakes or environemnts/describeDatalake. Since we assign this on environment level it would be easy to solve.

horadla23:
retest this please

lnardai:
Temporary, until we figure out how to handle Hieararchy Filtered Listings.

PR #8590, opened 2020-07-17
CB-5949 Ranger authorizer support for cdp-proxy and cdp-proxy-api topologies

moresandeep:

This PR will add Ranger authorizer plugin to Knox topologies [CB-5949](https://jira.cloudera.com/browse/CB-5949). What this means is that ranger policies will be applied to  all the requests flowing through Knox. This should not break any functionality as Ranger side will have policy that will allow all access going though Knox see [OPSAPS-55862](https://jira.cloudera.com/browse/OPSAPS-55862).

**Why is this change needed**
Currently Knox and Ranger are disconnected, there is no way for Ranger to enforce policies on Knox without modifying Knox configs (and Ranger policies). This PR will enable Ranger plugin for Knox so users can update Ranger policies for Knox in on place  (Ranger)

**What the patch does:**
This patch updates the salt files to add Knox composite authorization that uses ACL and Ranger authorization. 
If Ranger is not installed the topologies should work as before and there would be no change.

**Side effects:**
No side effects to my knowledge.

**Please include a short description.** 
Ranger authorizer support for Knox

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -15,164 +15,180 @@
            </param>
         </provider>
 
+{%- if 'RANGER' in exposed and 'RANGER_ADMIN' in salt['pillar.get']('gateway:location') -%}
+    {% set ranger_installed = True %}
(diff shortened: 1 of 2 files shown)


keyki:
Hi @moresandeep, can you please rebase your branch on top of master? There are some failing swagger changes and it will fix them.

moresandeep:
Thanks @keyki! I rebased and pushed the change.

PR #8593, opened 2020-07-20
Log collection api for datahub and datalake

attilapalfi92:
See the epic for even more info: https://jira.cloudera.com/browse/CB-6994


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/diagnostics/model/DiagnosticsCollectionRequest.java
```diff
@@ -0,0 +1,33 @@
...
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel("DiagnosticsCollectionRequest")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class DiagnosticsCollectionRequest extends BaseDiagnosticsCollectionRequest {
+
+    @NotNull
+    @ApiModelProperty(ModelDescriptions.StackModelDescription.CRN)
+    @ResourceObjectField(action = AuthorizationResourceAction.DATALAKE_READ, variableType = AuthorizationVariableType.CRN)
```
horadla23, line 23: what happens if you use this on distrox api? i think it will fail because it check for datalake

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/DiagnosticsV4Controller.java
```diff
@@ -0,0 +1,49 @@
...
+    private DiagnosticsTriggerService diagnosticsTriggerService;
+
+    @Inject
+    private VmLogsService vmLogsService;
+
+    @Inject
+    private VmLogsToVmLogsResponseConverter vmlogsConverter;
+
+    @Override
+    public void collectDiagnostics(@Valid DiagnosticsCollectionRequest request) {
```
horadla23, line 38: as discussed, pls authorize this as it is done in related controller of datalake service

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/diagnostics/model/DiagnosticsCollectionRequest.java
attilapalfi92: Yea I think it would fail. I don't think it's a good idea to check datahub operations with datalake resource type

horadla23: ohh sry its an old comment, not relevant, browser kept it

PR #8594, opened 2020-07-20
Missing error handling for RedbeamsTerminationActions$4

gergopapi2:
Problem: termination went failed because of an OptimisticLockException at delete of DBStack. Nevertheless the entry was deleted and the status could not be set during the failure handling (NPE), leading to another failure. This commit fixes setting the status and tries to fix the OptimisticLockException as well.


keyki:
LGTM. Please include the commit id in the message. :)

PR #8596, opened 2020-07-20
CB-8021 Replace DnsV1Endpoint#deleteDnsRecordsByFqdn call with cleanu…

lacikaaa:
…p in CB

In case of downscale the above mentioned endpoint was invoked.
This is not an async call, which could result in timout.
This commit moves this call to cleanup, which has polling already.
Also some simple refactor is introduced to cleanup code.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/downscale/StackDownscaleService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/downscale/StackDownscaleService.java
@@ -14,14 +14,12 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
-import javax.ws.rs.WebApplicationException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.DetailedStackStatus;
-import com.sequenceiq.cloudbreak.common.exception.WebApplicationExceptionMessageExtractor;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/downscale/StackDownscaleService.java
```diff
@@ -92,15 +86,14 @@ public void finishStackDownscale(StackScalingFlowContext context, String instanc
         flowMessageService.fireEventAndLog(stack.getId(), AVAILABLE.name(), STACK_DOWNSCALE_SUCCESS, String.join(",", deletedInstanceIds));
     }
 
-    private void cleanupDnsRecords(Stack stack, List<String> fqdns) {
+    private void cleanupDnsRecords(Stack stack, Collection<InstanceMetaData> instanceMetaDatas) {
+        Set<String> fqdns = instanceMetaDatas.stream().map(InstanceMetaData::getDiscoveryFQDN).collect(Collectors.toSet());
```
jamisonbennett, line 90: I am wondering if it might be possible for `getDiscoveryFQDN` to return null some error cases. I think this might be called for cleanup of a cluster that might have a node that didn't successfully complete the registration. A `filter(Objects:nonNull)` might be helpful here.

I know later in the case there is `Objects.requireNonNull(hostNames, "Hostnames must be set");` so it will throw an exception. But I wonder if this validation is called for the cluster deletion if it will prevent a cleanup from completing without the "force".

lacikaaa: we are ignoring cleanup error, but sure, I can add the filter

PR #8598, opened 2020-07-20
CB-8013 Fix the error handling in case of forced database termination

pkedvessy:
CB-8013 Forced database termination 
* is expected to clean up the database 
* is not expected to fail the environment/sdx deletion flow 

In case of any exception while calling cloud providers' API.


pkedvessy:
Retest this please

PR #8603, opened 2020-07-20
CB-7836: Configure Azure client to create V2 type storage account

tiborpopovics:


tiborpopovics:
retest this please

PR #8604, opened 2020-07-21
CB-6359 Speedup hosts decommission

cegganesh84:
1. For autoscaling feature, we need fast scale up and scale down for better user experience and competitive edge.
2. Right now we decommission CM hosts one by one, which is slow and the decommission flow stops on the first failure.
3. The patch kicks off decommission in parallel and also it moved the delete unused credentials to post CM host deletion.
4. For a 10 node cluster just on the CM side, we reduce the host removal time from 8 mins to 3.5 mins, also we made it a constant instead of being linear.
5. The error reporting pattern of consolidating exceptions of parallel decommissions is similar to Azure host decommission.

./gradlew build && tested in private stack

1. Before : Screen Shot 2020-07-16 at 5 37 46 PM
2. After : Screen Shot 2020-07-20 at 4 48 05 PM

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
@@ -132,11 +132,25 @@ public class DecommissionHandler implements EventHandler<DecommissionRequest> {
                     .map(hostsToRemove::get)
                     .filter(Objects::nonNull)
                     .collect(Collectors.toList());
-            decommisionedInstances.forEach(clusterDecomissionService::deleteHostFromCluster);
-            updateInstanceStatuses(decommisionedInstances, InstanceStatus.DECOMMISSIONED, "instance successfully downscaled");
-            clusterDecomissionService.restartStaleServices();
+            String decommisionResult = decommisionedInstances.parallelStream().map(metaData -> {
+                try {
+                    clusterDecomissionService.deleteHostFromCluster(metaData);
+                } catch (Exception e) {
+                    return e.getMessage();
+                }
+                return null;
+            }).filter(Objects::nonNull).collect(Collectors.joining(","));
 
-            result = new DecommissionResult(request, decommissionedHostNames);
+            if (!decommisionResult.isEmpty()) {
+                String decommisionStatus = String.format("One or more instances failed to decommision : %s", decommisionResult);
+                result = new DecommissionResult(decommisionStatus, new CloudbreakServiceException(decommisionStatus), request, hostNames, UNKNOWN_ERROR_PHASE);
+            } else {
+                clusterDecomissionService.deleteUnusedCredentialsFromCluster();
+                updateInstanceStatuses(decommisionedInstances, InstanceStatus.DECOMMISSIONED, "instance successfully downscaled");
+                clusterDecomissionService.restartStaleServices();
+
+                result = new DecommissionResult(request, decommissionedHostNames);
+            }
         } catch (DecommissionException e) {
             result = new DecommissionResult(e.getMessage(), e, request, hostNames, DECOMMISSION_ERROR_PHASE);
         } catch (Exception e) {
--- a/cluster-api/src/main/java/com/sequenceiq/cloudbreak/cluster/api/ClusterDecomissionService.java
+++ b/cluster-api/src/main/java/com/sequenceiq/cloudbreak/cluster/api/ClusterDecomissionService.java
@@ -25,6 +25,8 @@ public interface ClusterDecomissionService {
 
     void deleteHostFromCluster(InstanceMetaData data);
 
+    void deleteUnusedCredentialsFromCluster();
+
     void restartStaleServices() throws CloudbreakException;
 
     Map<String, Map<String, String>> getStatusOfComponentsForHost(String host);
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
```diff
@@ -132,11 +132,25 @@ public void accept(Event<DecommissionRequest> event) {
                     .map(hostsToRemove::get)
                     .filter(Objects::nonNull)
                     .collect(Collectors.toList());
-            decommisionedInstances.forEach(clusterDecomissionService::deleteHostFromCluster);
-            updateInstanceStatuses(decommisionedInstances, InstanceStatus.DECOMMISSIONED, "instance successfully downscaled");
-            clusterDecomissionService.restartStaleServices();
+            String decommisionResult = decommisionedInstances.parallelStream().map(metaData -> {
```
lacikaaa, line 135: I think you should collect the result into `List<String>`. The `toString` method will format it well

core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
```diff
@@ -132,11 +132,25 @@ public void accept(Event<DecommissionRequest> event) {
...
+                try {
+                    clusterDecomissionService.deleteHostFromCluster(metaData);
+                } catch (Exception e) {
+                    return e.getMessage();
```
lacikaaa, line 139: this should contain the instance id and/or fqdn also, and not only the message, so we could identify easily which node failed

core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
```diff
@@ -132,11 +132,25 @@ public void accept(Event<DecommissionRequest> event) {
...
+                String decommisionStatus = String.format("One or more instances failed to decommision : %s", decommisionResult);
+                result = new DecommissionResult(decommisionStatus, new CloudbreakServiceException(decommisionStatus), request, hostNames, UNKNOWN_ERROR_PHASE);
+            } else {
+                clusterDecomissionService.deleteUnusedCredentialsFromCluster();
+                updateInstanceStatuses(decommisionedInstances, InstanceStatus.DECOMMISSIONED, "instance successfully downscaled");
+                clusterDecomissionService.restartStaleServices();
 
-            result = new DecommissionResult(request, decommissionedHostNames);
+                result = new DecommissionResult(request, decommissionedHostNames);
+            }
```
lacikaaa, line 153: could you move this whole block into a separate method?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/DecommissionHandler.java
keyki: Will it cause issues with large clusters? Like when I downscale with 100 instances that will be 100 requests parallel?

TheTinkerDad:
One concern from my side about the solution in general - with this solution in case of large clusters we're sending many, many requests to CM at once, in parallel. As far as I'm aware, this can cause problems so a maximum batch size (possibly a configurable value) should be introduced to prevent this by maximizing how many requests we can send to CM in parallel. Also AFAIK CM in many cases supports batching implemented on their side - e.g. we could call a batch operation and send 10 instances that should be decommissioned as a single request instead of sending 10 separate requests in parallel.
IMHO this is what we should be looking into in the first place and as far as I remember, we've encountered something similar in the past with Cloudera Director.

cegganesh84:
Thanks for the feedback!

Good call on large decommission.

Decommission happens in many steps, some have batch support, and some do not.

1. Decommission hosts are batchable and it is already batched prior to this PR.
2. Read services are batchable, and it already batched prior to this PR. But it was not cached so we ended up doing it N times.
3. Remove service, remove host, delete host are not batchable. The only batch API for this is to remove all services, remove all hosts and delete all hosts which can not be used in scale down. Essentially this PR parallelizes them in one shot. The ask here is to introduce BATCH_COUNT. There are three factors that could potentially determine this number,
   a. 1 will mean everything will be sequential and it will become linear. 100 will make it a constant, but there are some concerns (one of the two below). 10 or 25 will be a good middle ground. Ideally, we need to find a point of diminishing returns, but that would be a costly process, so a gut feel.
   b. The max hosts CM can deal with. CM already handles 100 commissions or decommissions in parallel, so should not be a huge factor for 3.
   c. The max Req. per sec. that Java threads in the control plane, and Nginx, and CM in workload cluster can handle. I am not sure about this number, again, there is a general comfort around the number 10 or 25.
4. Delete unused credentials is batchable, and it is introduced in this PR. Probably I can separate that out, which is a much much lower hanging fruit.

sidseth:
Apparently 'parallelStream' relies on the default ForkJoinPool, which has a default size of numCores-1 - so even on large clusters, we get automatically limited by #cores .
That said, we're probably better of creating a separate controlled thread pool here - so that the default threadpool isn't bogged down by network calls (in case there are other parts of the system using parallel calls), and there's better control on the number of threads.
The other weirdness with parallelStream is the error reporting - which exits on first failure, and leaves the currently executing threads to run to completion.

cegganesh84:
Certainly did not want to introduce a new pattern into the codebase. Saw that Azure hosts decommission also used parallel stream. For better error handling wrapped the mapper into the catch-all exceptions block. Anyway reading this to understand the problems with parallel stream framework http://coopsoft.com/dl/Blunder.pdf.

In the meantime addressed the lowest hanging fruit #8612

cegganesh84:
I tested #8612 it performed 15 seconds faster than the parallelized one. So the credentials getting deleted one-by-one seems to be the actual problem. The parallelization was not yielding good results could be due to the way parallel stream works (mentioned by Sid).

Screen Shot 2020-07-24 at 2 11 27 PM

Closing this PR.

PR #8605, opened 2020-07-21
CB-7930 Verify credential on environment level

foldik:

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
@@ -21,6 +21,7 @@ import javax.ws.rs.core.MediaType;
 import com.sequenceiq.cloudbreak.jerseyclient.RetryAndMetrics;
 import com.sequenceiq.common.api.telemetry.request.FeaturesRequest;
 import com.sequenceiq.environment.api.doc.environment.EnvironmentOpDescription;
+import com.sequenceiq.environment.api.v1.credential.model.response.CredentialResponse;
 import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentChangeCredentialRequest;
 import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentEditRequest;
 import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentRequest;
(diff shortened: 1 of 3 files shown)


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
```diff
@@ -184,6 +185,13 @@ SimpleEnvironmentResponses deleteMultipleByCrns(Set<String> crns,
             nickname = "stopEnvironmentByCrnV1")
     void postStopByCrn(@PathParam("crn") String crn);
 
+    @GET
+    @Path("/crn/{crn}/verify")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.VERIFY_BY_CRN, notes = ENVIRONMENT_NOTES,
+            nickname = "verifyEnvironmentByCrn", httpMethod = "GET")
+    CredentialResponse verifyByCrn(@PathParam("crn") String crn);
```
horadla23, line 193: pls rename it to `verifyCredentialByCrn` (or `verifyCredentialByEnvCrn`) to make it clear, also pls fix `nickname` in the annotation

foldik: Yes it makes sense to make it more specific.

horadla23: also maybe it is worth to change the path as well, because now based on path i can think that we are verifying the environment itself

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
```diff
@@ -268,6 +274,15 @@ public void postStopByCrn(@ResourceCrn @TenantAwareParam String crn) {
         environmentStopService.stopByCrn(crn);
     }
 
+    @Override
+    @CheckPermissionByResourceCrn(action = AuthorizationResourceAction.DESCRIBE_ENVIRONMENT)
+    public CredentialResponse verifyCredentialByEnvCrn(@ResourceCrn String crn) {
```
doktoric, line 279: do we need this for the /name endpoint ? most of the time we have both /crn and /name

foldik: We don't need. UI will use the version which works with crn. I can add the /name version for consistency and usability reason if I have to.

PR #8607, opened 2020-07-21
CB-7425 If I try to delete in batch, should not fail on the first error

doktoric:
Credential deletion can fail so if user initiate multiple delete then we will try to delete the next element as well

### DIFF
--- a/environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialDeleteServiceTest.java
+++ b/environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialDeleteServiceTest.java
@@ -3,7 +3,6 @@ package com.sequenceiq.environment.credential.service;
 import static com.sequenceiq.common.model.CredentialType.ENVIRONMENT;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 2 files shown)


environment/src/test/java/com/sequenceiq/environment/credential/service/CredentialDeleteServiceTest.java
```diff
@@ -98,27 +95,40 @@ void testMultipleIfAllTheCredentialsAreExistsAndAbleToArchiveThenExpectedCredent
...
+        String name2 = "something2";
+        Credential cred2 = createCredentialWithName(name2);
+        cred2.setId(2L);
+        when(credentialService.findByNameAndAccountId(eq(name2), eq(ACCOUNT_ID), any(Set.class), any()))
+                .thenThrow(new BadRequestException("anything can happen"));
 
-        verify(credentialService, times(1)).findByNameAndAccountId(anyString(), anyString(), anyCollection(), any());
+        when(environmentViewService.findAllByCredentialId(cred1.getId())).thenReturn(Set.of());
+
+        Set<Credential> result = underTest.deleteMultiple(Set.of(name1, name2), ACCOUNT_ID, ENVIRONMENT);
```
bergerdenes, line 123: change the order of deletion so exception should occur first, then delete after

PR #8613, opened 2020-07-22
CB-8072 user sync apis don't work with internal actor

horadla23:


horadla23:
https://github.com/hortonworks/cloudbreak/blob/49bbe5c28e20d2ef364184f2660dc8eb4e858345/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java#L145 this fails unfortunately

PR #8620, opened 2020-07-23
Swagger check in a separated job

doktoric:


integration-test/apidefinitions/validate.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo Cloudbreak swagger compatibility check
+IFS=,
+for val in $DataList; do
+  check $val
+done
+IFS=$Field_Separator
+
+
+[ $cstatus -eq 0 ] && echo "swagger diff finished succesfully" || echo "swagger diff failed"
```
afarsang, line 25: it might be 
 [ $cstatus -eq 0 ] && echo "swagger validation finished succesfully" || echo "swagger validation failed"

integration-test/scripts/swagger-check.sh
```diff
@@ -1 +1,74 @@
...
+done
+set -e
+IFS=$Field_Separator
+$INTEGCB_LOCATION/.deps/bin/docker-compose up swagger-diff | tee swagger-diff.out
+grep "swagger diff finished succesfully" swagger-diff.out
+swaggerdiffresult=$?
+
+$INTEGCB_LOCATION/.deps/bin/docker-compose up swagger-validation | tee swagger-validation-result.out
+
+# Check swagger validation results
```
afarsang, line 69: Previously the cloudbreak swagger was validated only, however now it is 5 different swagger yaml file will be validated, and this code will be successful when only one swagger validation is successful (expected to fail the job if ANY swagger diff is unsuccessful )

Also I am not sure, the 'status_code=1' is enough to fail the job, but I might have missed something.


grep "swagger validation finished successfully" swagger-validation-result.out
swaggervalidationresult=$?

exit $swaggerdiffresult && $swaggervalidationresult

PR #8621, opened 2020-07-23
CB-8094 Image selection fails for yarn tests

Bajzathd:
Revert "CB-7805: Avoid automatic fallback to base images"
This reverts commit f27320d11b96061cd5b62898619773f9996e433b.


Bajzathd:
Reverted without PR

PR #8622, opened 2020-07-23
CB-8098 [ASRG] error message when missing entitlement is not clear

gergopapi2:
In case the user tries to create an env with single RG but the feature is turned off from entitlement then an error message comes up. It was too technical. Also, if a RG name is specified along with single RG name then another error message comes, it was also too technical.


pdarvasi:
Pls, fix the messages in the tests also. (e.g AzureParameterValidatorTest)

PR #8630, opened 2020-07-24
CB-8126 Hard fail if audit endpoint configuration is missing

bergerdenes:
This configuration item has no default value and it was not asserted before.
Upon audit service call, the endpoint and port was always used and in case it was null, an NPE
was raised by the GRPC client library


audit-connector/src/main/java/com/sequenceiq/cloudbreak/audit/config/AuditConfig.java
```diff
@@ -50,6 +56,10 @@ public void init() {
             port = parts.length == 2
                     ? Integer.parseInt(parts[1])
                     : DEFAULT_AUDIT_PORT;
+        } else {
```
doktoric, line 59: else if ?

bergerdenes:
retest this please

PR #8637, opened 2020-07-24
CB-8135 configure paywall credentials to CM yum repo file

keyki:
Whenever a cluster is installed from https://archive.cloudera.com
a basic credential must be provided. This credential can be generated
from the CM license file that is provided by UMS. The test credential
can be used to authenticate to this paywall however, in production
each customer gets its own license file. The username is the uuid field
and the password is generated from sha256(name + uuid) and the first 12
characters.


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -414,32 +419,61 @@ private void decoratePillarWithClouderaManagerAutoTls(Cluster cluster, Map<Strin
...
+
+    }
 
+    private Map<String, Object> createCMRepoPillar(ClouderaManagerRepo clouderaManagerRepo, Optional<String> license) {
+        Map<String, Object> pillarValues = new HashMap<>();
+        pillarValues.put("repo", clouderaManagerRepo);
+        parseLicense(license).ifPresent(jsonLicense -> {
+            String username = jsonLicense.getPaywallUsername();
+            String password = jsonLicense.getPaywallPassword();
+            if (isNotEmpty(username) && isNotEmpty(password)) {
```
lacikaaa, line 455: just mentioning, don't have to replace: `StringUtils#isNoneEmpty` could work also

PR #8639, opened 2020-07-24
CB-8144: Fix Cloudbreak Core DNS registration and TTL

jamisonbennett:
Fix the cloudbreak core DNS A record and reverse registration to
always have a 30 second TTL and to be able to regsister both the
A record and reverse record with salt retries in the case of an
intermittent failure.

More Details:

Cloudbreak core had a fix put in to address scale with CB-6544. But
it turns out this does not always work. When ipa-client-install is
run, it sometimes (but for some reason it does not always) registers
the DNS and reverse DNS. If the ipa-client-install registers DNS
entries then the add_dns_record rule is not run because the unless
clause causes it to be skipped. This presents a few problems.

1. The flag file /var/log/dnsrecord-add-executed is never created. So
every time a salt high state is run the unless clauses which in turn
checks FreeIPA to see if the DNS entries are present. The intent of
CB-6544 was to fix this.

2. The actual rule is never run, so the TTL for the A record is not
set to 30. The intent of CB-7715 was to fix this.

Additionally, the reverse DNS TTL was never set so the default value
for the zone is used. See CB-7715 for more details on this.

It also appears that when the dnsrecord-add is run, it can fail with a
partially executed state (e.g. A record set but reverse not set). This
means that a retry will fail because it will not modify anything for
one of the 2 implicit commands, which makes the salt retry fail. But
this is silently happening because the retry is also checking the DNS
A record so in this case actually the reverse record never gets set.
Again, it is not clear exactly how or why this might happen.

What was tested:

A datalake was provsioned using local cloudbreak deployment. This was not validated at scale.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
@@ -12,32 +12,31 @@ ipa_packages_install:
     - unless:
       - rpm -q ipa-client openldap openldap-clients
 
+/opt/salt/scripts/join_ipa.sh:
+  file.managed:
+    - makedirs: True
+    - user: root
+    - group: root
+    - mode: 700
+    - source: salt://sssd/template/join_ipa.j2
+    - template: jinja
+
 join_ipa:
   cmd.run:
 {% if metadata.platform != 'YARN' %}
-    - name: |
-        ipa-client-install --unattended --uninstall
-        ipa-client-install --realm={{salt['pillar.get']('sssd-ipa:realm')}} \
-          --domain={{salt['pillar.get']('sssd-ipa:domain')}} --mkhomedir --principal={{salt['pillar.get']('sssd-ipa:principal')}} \
-          {%- if "ID_BROKER_CLOUD_IDENTITY_ROLE" in grains.get('roles', []) %}
-          --no-sshd --no-ssh \
-          {%- endif %}
-          --password $PW --unattended --force-join --ssh-trust-dns --no-ntp && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed
+    - name: /opt/salt/scripts/join_ipa.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed
 {% else %}
-    - name: |
-        runuser -l root -c 'ipa-client-install --unattended --uninstall'
-        runuser -l root -c 'ipa-client-install --realm={{salt['pillar.get']('sssd-ipa:realm')}} \
-          --domain={{salt['pillar.get']('sssd-ipa:domain')}} --mkhomedir --principal={{salt['pillar.get']('sssd-ipa:principal')}} \
-          {%- if "ID_BROKER_CLOUD_IDENTITY_ROLE" in grains.get('roles', []) %}
-          --no-sshd --no-ssh \
-          {%- endif %}
-          --password "{{salt['pillar.get']('sssd-ipa:password')}}" --unattended --force-join --ssh-trust-dns --no-ntp && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed'
+    - name: runuser -l root -c 'export PW="{{salt['pillar.get']('sssd-ipa:password')}}" && /opt/salt/scripts/join_ipa.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed'
 {% endif %}
-    - unless: test -f /var/log/ipa-join-executed || { echo $PW | kinit {{salt['pillar.get']('sssd-ipa:principal')}} && ipa env; }
+    - unless: test -f /var/log/ipa-join-executed
     - runas: root
     - failhard: True
+    - require:
+        - file: /opt/salt/scripts/join_ipa.sh
+{% if metadata.platform != 'YARN' %}
     - env:
         - PW: "{{salt['pillar.get']('sssd-ipa:password')}}"
+{% endif %}
 
 {% if metadata.platform == 'YARN' and not metadata.cluster_in_childenvironment %}
 dns_remove_script:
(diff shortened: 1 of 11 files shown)


jamisonbennett:
I ran the output script through shellcheck and cleaned it up.

orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
```diff
@@ -0,0 +1,43 @@
...
+# dnsrecord-add must either add the record or modify it
+if ! ipa dnsrecord-find {{ pillar['sssd-ipa']['domain'] }}. "--name=${HOSTNAME}" "--a-rec=${IPADDR}"  --ttl {{ pillar['sssd-ipa']['dns_ttl'] }} --all; then
+  ipa dnsrecord-add {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}" "--a-rec=${IPADDR}" --ttl {{ pillar['sssd-ipa']['dns_ttl'] }}
+fi
+if ! ipa dnsrecord-find {{ pillar['sssd-ipa']['domain'] }}. "--name=${HOSTNAME}" "--a-rec=${IPADDR}"  --ttl {{ pillar['sssd-ipa']['dns_ttl'] }} --all; then
+  echo "Failed to set DNS A-record for ${HOSTNAME}"
+  false
+fi
+
+for zone in $(ipa dnszone-find --raw | grep idnsname | cut -d':' -f2 | xargs)
```
lacikaaa, line 23: I think you should grep for `.in-addr.arpa.` here

lacikaaa: also I have a little concern here. if we have lets see the IP `10.10.1.2`.
And we have 2 zones:
- `10.10.in-addr.arpa`
- `10.in-addr.arpa`
this for loop will add the reverse record for both zone. Is it really what we want to do?

jamisonbennett: > I think you should grep for .in-addr.arpa. here
Done, but its technically not needed as `if echo "$REVERSE_IP" | grep -qE "$ZONE_NET$"; then` should have handled this.

also I have a little concern here. if we have

>  lets see the IP 10.10.1.2.
> And we have 2 zones:
> 
> 10.10.in-addr.arpa
> 10.in-addr.arpa
> this for loop will add the reverse record for both zone. Is it really what we want to do?
I don't think we should have cases where there are different matching zones, but maybe there is a way to do it with submitting with our current deployment of CDP. Either way, I made the reverse DNS register with the more specific zone. So it should only register once.

lacikaaa: > Done, but its technically not needed as...

Yes, but it will then loop for reverse zones only

We should not create but never know. Also customer might create it. I tested and FreeIPA will return the more exact match, but I think it's better to be prepared and add only for the right one.
Thanks for the modification!

orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
```diff
@@ -0,0 +1,45 @@
...
+do
+    ZONE_NET=${zone//.in-addr.arpa./}
+    if [[ $LONGEST_MATCHING_ZONE_FOUND -eq 0 ]] && echo "$REVERSE_IP" | grep -qE "$ZONE_NET$"; then
+        LONGEST_MATCHING_ZONE_FOUND=1
+        REVERSE_RECORD_NAME=$(echo "$REVERSE_IP" | sed "s/.$ZONE_NET$//g")
+        # dnsrecord-add must either add the record or modify it
+        if ! ipa dnsrecord-find "$zone" "--name=$REVERSE_RECORD_NAME" "--ptr-rec=${FQDN}." --ttl {{ pillar['sssd-ipa']['dns_ttl'] }}; then
+          ipa dnsrecord-add "$zone" "$REVERSE_RECORD_NAME" "--ptr-rec=${FQDN}." --ttl {{ pillar['sssd-ipa']['dns_ttl'] }}
+        fi
+        if ! ipa dnsrecord-find "$zone" "--name=$REVERSE_RECORD_NAME" "--ptr-rec=${FQDN}." --ttl {{ pillar['sssd-ipa']['dns_ttl'] }}; then
```
lacikaaa, line 34: if this `if` is false, shouldn't we exit the loop? I mean in case we find the record, so it's successfully added we shouldn't iterate any more

orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
```diff
@@ -0,0 +1,45 @@
...
+          echo "Failed to set Reverse DNS PTR-record for ${FQDN}"
+          false
+        fi
+    fi
+done
+
+# Destroy the krbtgt
+kdestroy
```
lacikaaa, line 42: due to `set -e` this won't run in case of failure. If we really want to destroy it we should use `trap` for this

orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
```diff
@@ -12,30 +12,27 @@ ipa_packages_install:
...
-          --no-sshd --no-ssh \
-          {%- endif %}
-          --password "{{salt['pillar.get']('sssd-ipa:password')}}" --unattended --force-join --ssh-trust-dns --no-ntp && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed'
+    - name: runuser -l root -c '/opt/salt/scripts/join_ipa.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed'
 {% endif %}
-    - unless: test -f /var/log/ipa-join-executed || { echo $PW | kinit {{salt['pillar.get']('sssd-ipa:principal')}} && ipa env; }
+    - unless: test -f /var/log/ipa-join-executed
     - runas: root
-    - failhard: True
+    - failhard: True:$
```
lacikaaa, line 33: it looks like you have some extra character here

jamisonbennett:
I tested this on y-cloud and AWS.

PR #8640, opened 2020-07-24
CB-6363: Create an Env api and flow which updates the Envs Stacks pillar configs

wonderslug:
This adds an API and flow to the environment service which allows for
the updating of all of an Environments Stacks pillar configs.

This is primarily needed for FreeIPA right now, when a FreeIPA DNS
server changes IP to allow all of the Clusters within an environment
to update their DNS server IP addresses.

The API added is:
POST /v1/env/crn/{crn}/update_config

These do not have authorization on them and should only ever be
called from internal to the Control Plane.  They kick off the
EnvStackConfigUpdates Flow which collects all of the stacks for
the environment and calls the Cloudbreak API and Flow created in
CB-6364.  It does not monitor these flows, it just keeps
retrying until it successfully submits a request for the pillar configs
to be updated.

If another call to update the configs for the stacks come in while
one is processing and retrying, it first call will be canceled and the
second will start the process again for all stacks.  This is to make sure
that all stacks are updated for when the latest call comes in.

It will retry submitting pillar updates every 60 seconds for 120 tries.
This is to give enough time for current running flows to finish allows
the pillar configs update to catch up the stack that did have flows running.

This is covered by the one-pager here:
https://docs.google.com/document/d/1ntWZF1WP9BVRU8mP9IEzFnN63-qggG9CNEImVbXjs9I

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
@@ -292,12 +292,21 @@ public interface StackV4Endpoint {
     FlowIdentifier updateSaltByName(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
             @AccountId @QueryParam("accountId") String accountId);
 
+    /**
+     * @deprecated Use updatePillarConfigurationByCrn instead
+     */
     @PUT
     @Path("{name}/pillar_config_update")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = UPDATE_PILLAR_CONFIG, nickname = "updatePillarConfigurationByName")
-    FlowIdentifier updatePillarConfigurationByName(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
-            @AccountId @QueryParam("accountId") String accountId);
+    @Deprecated
+    FlowIdentifier updatePillarConfigurationByName(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name);
+
+    @PUT
+    @Path("crn/{crn}/pillar_config_update")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = UPDATE_PILLAR_CONFIG, nickname = "updatePillarConfigurationByCrn")
+    FlowIdentifier updatePillarConfigurationByCrn(@PathParam("workspaceId") Long workspaceId, @PathParam("crn") String crn);
 
     @POST
     @Path("{name}/database_backup")
--- a/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/urlparsers/V4ExistingResourceByCrnOrNameRestUrlParser.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/urlparsers/V4ExistingResourceByCrnOrNameRestUrlParser.java
@@ -18,11 +18,11 @@ public class V4ExistingResourceByCrnOrNameRestUrlParser extends RestUrlParser {
     // v4/{workspaceId}/blueprints/name/{name}
     // v4/{workspaceId}/blueprints/crn/{name}
     private static final Pattern PATTERN = Pattern
-            .compile("v4/(\\d+)/(blueprints|image_catalogs|recipes|stacks|distrox|cluster_templates)/(?:name|crn)/([^/]+)");
+            .compile("v4/(\\d+)/(blueprints|image_catalogs|recipes|stacks|distrox|cluster_templates)/(?:name|crn)/([^/]+)(/\\w+)?");
 
     @Override
     protected List<String> parsedMethods() {
-        return List.of("DELETE", "GET");
+        return List.of("DELETE", "GET", "PUT");
     }
 
     @Override
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
@@ -22,6 +22,7 @@ public class EnvironmentOpDescription {
     public static final String VERIFY_CREDENTIAL_BY_CRN = "Verifies the credential used by the given environment.";
     public static final String CLI_COMMAND = "produce cli command input for environment creation";
     public static final String GET_CRN_BY_NAME = "Get the crn of an environment by name.";
+    public static final String UPDATE_CONFIG_BY_CRN = "Update the configuration for all stacks in the Environment by the Environment CRN";
 
     private EnvironmentOpDescription() {
     }
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
@@ -205,4 +205,11 @@ public interface EnvironmentEndpoint {
     @ApiOperation(value = EnvironmentOpDescription.CLI_COMMAND, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
             nickname = "getCreateEnvironmentForCli")
     Object getCreateEnvironmentForCli(@NotNull @Valid EnvironmentRequest environmentRequest);
+
+    @POST
+    @Path("/crn/{crn}/update_config")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.UPDATE_CONFIG_BY_CRN, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
+        nickname = "updateConfigsInEnvironmentByCrnV1")
+    void updateConfigsInEnvironmentByCrn(@PathParam("crn") String crn);
 }
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
@@ -1,0 +1,111 @@
+package com.sequenceiq.environment.environment.flow.config.update;
+
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FINALIZE_ENV_STACK_CONIFG_UPDATES_EVENT;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.HANDLE_FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+
+import com.sequenceiq.cloudbreak.common.event.ResourceCrnPayload;
+import com.sequenceiq.cloudbreak.event.ResourceEvent;
+import com.sequenceiq.environment.environment.EnvironmentStatus;
+import com.sequenceiq.environment.environment.domain.Environment;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesEvent;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesFailedEvent;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesHandlerSelectors;
+import com.sequenceiq.environment.environment.service.EnvironmentService;
+import com.sequenceiq.environment.environment.service.EnvironmentStatusUpdateService;
+import com.sequenceiq.environment.metrics.EnvironmentMetricService;
+import com.sequenceiq.environment.metrics.MetricType;
+import com.sequenceiq.flow.core.CommonContext;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.statemachine.action.Action;
+
+@Configuration
+public class EnvStackConfigUpdatesActions {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnvStackConfigUpdatesActions.class);
+
+    private final EnvironmentService environmentService;
+
+    private final EnvironmentStatusUpdateService environmentStatusUpdateService;
+
+    private final EnvironmentMetricService metricService;
+
+    public EnvStackConfigUpdatesActions(EnvironmentService environmentService,
+        EnvironmentStatusUpdateService environmentStatusUpdateService,
+        EnvironmentMetricService metricService) {
+        this.environmentService = environmentService;
+        this.environmentStatusUpdateService = environmentStatusUpdateService;
+        this.metricService = metricService;
+    }
+
+    @Bean(name = "STACK_CONFIG_UPDATES_START_STATE")
+    public Action<?, ?> collectClusterInfo() {
+        return new AbstractEnvStackConfigUpdatesAction<>(EnvStackConfigUpdatesEvent.class) {
+            @Override
+            protected void doExecute(CommonContext context, EnvStackConfigUpdatesEvent payload, Map<Object, Object> variables) {
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateEnvironmentStatusAndNotify(context, payload,
+                        getCurrentStatus(payload.getResourceId()),
+                        ResourceEvent.ENVIRONMENT_STACK_CONFIGS_UPDATE_STARTED,
+                        EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_START_STATE);
+
+                sendEvent(context,
+                    EnvStackConfigUpdatesHandlerSelectors.STACK_CONFIG_UPDATES_HANDLER_EVENT
+                        .selector(), environmentDto);
+            }
+        };
+    }
+
+    @Bean(name = "STACK_CONFIG_UPDATES_FINISHED_STATE")
+    public Action<?, ?> finishedAction() {
+        return new AbstractEnvStackConfigUpdatesAction<>(ResourceCrnPayload.class) {
+            @Override
+            protected void doExecute(CommonContext context, ResourceCrnPayload payload, Map<Object, Object> variables) {
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateEnvironmentStatusAndNotify(context, payload,
+                        getCurrentStatus(payload.getResourceId()),
+                        ResourceEvent.ENVIRONMENT_STACK_CONFIGS_UPDATE_FINISHED,
+                        EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FINISHED_STATE);
+                metricService.incrementMetricCounter(MetricType.ENV_STACK_CONFIG_UPDATE_FINISHED,
+                    environmentDto);
+                LOGGER.info("Flow entered into STACK_CONFIG_UPDATES_FINISHED_STATE");
+                sendEvent(context, FINALIZE_ENV_STACK_CONIFG_UPDATES_EVENT.event(), payload);
+            }
+        };
+    }
+
+    @Bean(name = "STACK_CONFIG_UPDATES_FAILED_STATE")
+    public Action<?, ?> failedAction() {
+        return new AbstractEnvStackConfigUpdatesAction<>(EnvStackConfigUpdatesFailedEvent.class) {
+            @Override
+            protected void doExecute(CommonContext context, EnvStackConfigUpdatesFailedEvent payload, Map<Object, Object> variables) {
+                LOGGER.warn(
+                    String.format("Failed to update environments stack configs '%s'. Status: '%s'.",
+                        payload.getEnvironmentDto(), payload.getEnvironmentStatus()),
+                    payload.getException());
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateFailedEnvironmentStatusAndNotify(context, payload,
+                        getCurrentStatus(payload.getResourceId()),
+                        ResourceEvent.ENVIRONMENT_STACK_CONFIGS_UPDATE_FAILED,
+                        EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FAILED_STATE);
+                metricService.incrementMetricCounter(MetricType.ENV_STACK_CONFIG_UPDATE_FAILED,
+                    environmentDto,
+                    payload.getException());
+                sendEvent(context, HANDLE_FAILED_ENV_STACK_CONIFG_UPDATES_EVENT.event(), payload);
+            }
+        };
+    }
+
+    private EnvironmentStatus getCurrentStatus(Long envId) {
+        return environmentService
+            .findEnvironmentById(envId)
+            .map(Environment::getStatus)
+            .orElseThrow(() -> new IllegalStateException(
+                String.format("Cannot get status of environment, because it does not exist: %s. ",
+                    envId)
+            ));
+    }
+}
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/config/EnvStackConfigUpdatesFlowConfig.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/config/EnvStackConfigUpdatesFlowConfig.java
@@ -1,0 +1,75 @@
+package com.sequenceiq.environment.environment.flow.config.update.config;
+
+import static com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState.FINAL_STATE;
+import static com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState.INIT_STATE;
+import static com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FAILED_STATE;
+import static com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FINISHED_STATE;
+import static com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_START_STATE;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.ENV_STACK_CONFIG_UPDATES_START_EVENT;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FINALIZE_ENV_STACK_CONIFG_UPDATES_EVENT;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FINISH_ENV_STACK_CONFIG_UPDATES_EVENT;
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.HANDLE_FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+
+import com.sequenceiq.environment.environment.flow.config.update.EnvStackConfigUpdatesState;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors;
+import com.sequenceiq.flow.core.config.AbstractFlowConfiguration;
+import com.sequenceiq.flow.core.config.RetryableFlowConfiguration;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
+public class EnvStackConfigUpdatesFlowConfig extends
+    AbstractFlowConfiguration<EnvStackConfigUpdatesState, EnvStackConfigUpdatesStateSelectors>
+    implements RetryableFlowConfiguration<EnvStackConfigUpdatesStateSelectors> {
+
+    private static final List<Transition<EnvStackConfigUpdatesState, EnvStackConfigUpdatesStateSelectors>>
+        TRANSITIONS = new Transition.Builder<EnvStackConfigUpdatesState, EnvStackConfigUpdatesStateSelectors>()
+        .defaultFailureEvent(FAILED_ENV_STACK_CONIFG_UPDATES_EVENT)
+
+        .from(INIT_STATE).to(STACK_CONFIG_UPDATES_START_STATE)
+        .event(ENV_STACK_CONFIG_UPDATES_START_EVENT).defaultFailureEvent()
+
+        .from(STACK_CONFIG_UPDATES_START_STATE).to(STACK_CONFIG_UPDATES_FINISHED_STATE)
+        .event(FINISH_ENV_STACK_CONFIG_UPDATES_EVENT).defaultFailureEvent()
+
+        .from(STACK_CONFIG_UPDATES_FINISHED_STATE).to(FINAL_STATE)
+        .event(FINALIZE_ENV_STACK_CONIFG_UPDATES_EVENT).defaultFailureEvent()
+
+        .build();
+
+    protected EnvStackConfigUpdatesFlowConfig() {
+        super(EnvStackConfigUpdatesState.class, EnvStackConfigUpdatesStateSelectors.class);
+    }
+
+    @Override
+    protected List<Transition<EnvStackConfigUpdatesState, EnvStackConfigUpdatesStateSelectors>> getTransitions() {
+        return TRANSITIONS;
+    }
+
+    @Override
+    protected FlowEdgeConfig<EnvStackConfigUpdatesState, EnvStackConfigUpdatesStateSelectors> getEdgeConfig() {
+        return new FlowEdgeConfig<>(INIT_STATE, FINAL_STATE, STACK_CONFIG_UPDATES_FAILED_STATE,
+            HANDLE_FAILED_ENV_STACK_CONIFG_UPDATES_EVENT);
+    }
+
+    @Override
+    public EnvStackConfigUpdatesStateSelectors[] getEvents() {
+        return EnvStackConfigUpdatesStateSelectors.values();
+    }
+
+    @Override
+    public EnvStackConfigUpdatesStateSelectors[] getInitEvents() {
+        return new EnvStackConfigUpdatesStateSelectors[]{ENV_STACK_CONFIG_UPDATES_START_EVENT};
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Update all Environment Clusters configuration";
+    }
+
+    @Override
+    public EnvStackConfigUpdatesStateSelectors getRetryableEvent() {
+        return HANDLE_FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+    }
+}
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
@@ -1,0 +1,37 @@
+package com.sequenceiq.environment.environment.flow.config.update.event;
+
+
+
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+
+import com.sequenceiq.cloudbreak.common.event.Selectable;
+import com.sequenceiq.environment.environment.EnvironmentStatus;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.flow.reactor.api.event.BaseFailedFlowEvent;
+
+public class EnvStackConfigUpdatesFailedEvent extends BaseFailedFlowEvent implements Selectable {
+
+    private final EnvironmentDto environmentDto;
+
+    private final EnvironmentStatus environmentStatus;
+
+    public EnvStackConfigUpdatesFailedEvent(EnvironmentDto environmentDto, Exception exception, EnvironmentStatus environmentStatus) {
+        super(FAILED_ENV_STACK_CONIFG_UPDATES_EVENT.name(), environmentDto.getResourceId(),
+                environmentDto.getName(), environmentDto.getResourceCrn(), exception);
+        this.environmentDto = environmentDto;
+        this.environmentStatus = environmentStatus;
+    }
+
+    @Override
+    public String selector() {
+        return FAILED_ENV_STACK_CONIFG_UPDATES_EVENT.event();
+    }
+
+    public EnvironmentDto getEnvironmentDto() {
+        return environmentDto;
+    }
+
+    public EnvironmentStatus getEnvironmentStatus() {
+        return environmentStatus;
+    }
+}
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/handler/StackConfigUpdatesHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/handler/StackConfigUpdatesHandler.java
@@ -1,0 +1,52 @@
+package com.sequenceiq.environment.environment.flow.config.update.handler;
+
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesEvent;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesEvent.EnvStackConfigUpdatesEventBuilder;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesFailedEvent;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesHandlerSelectors;
+import com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors;
+import com.sequenceiq.environment.environment.service.stack.StackPollerService;
+import com.sequenceiq.flow.core.FlowConstants;
+import com.sequenceiq.flow.reactor.api.event.EventSender;
+import com.sequenceiq.flow.reactor.api.handler.EventSenderAwareHandler;
+import org.springframework.stereotype.Component;
+import reactor.bus.Event;
+
+@Component
+public class StackConfigUpdatesHandler extends EventSenderAwareHandler<EnvironmentDto> {
+
+    private final StackPollerService stackPollerService;
+
+    public StackConfigUpdatesHandler(EventSender eventSender,
+        StackPollerService stackPollerService) {
+        super(eventSender);
+        this.stackPollerService = stackPollerService;
+    }
+
+    @Override
+    public String selector() {
+        return EnvStackConfigUpdatesHandlerSelectors.STACK_CONFIG_UPDATES_HANDLER_EVENT.event();
+    }
+
+    @Override
+    public void accept(Event<EnvironmentDto> event) {
+        try {
+            stackPollerService.updateStackConfigurations(event.getData().getResourceId(),
+                event.getData().getResourceCrn(), event.getHeaders().get(FlowConstants.FLOW_ID));
+
+            EnvStackConfigUpdatesEvent envStackConfigUpdatesEvent = EnvStackConfigUpdatesEventBuilder
+                .anEnvStackConfigUpdatesEvent()
+                .withSelector(
+                    EnvStackConfigUpdatesStateSelectors.FINISH_ENV_STACK_CONFIG_UPDATES_EVENT
+                        .selector())
+                .withResourceId(event.getData().getResourceId())
+                .build();
+
+            eventSender().sendEvent(envStackConfigUpdatesEvent, event.getHeaders());
+        } catch (Exception e) {
+            eventSender().sendEvent(new EnvStackConfigUpdatesFailedEvent(event.getData(), e, null),
+                event.getHeaders());
+        }
+    }
+}
--- a/environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
@@ -1,0 +1,92 @@
+package com.sequenceiq.environment.environment.poller;
+
+import com.dyngr.core.AttemptMaker;
+import com.dyngr.core.AttemptResult;
+import com.dyngr.core.AttemptResults;
+import com.dyngr.core.AttemptState;
+import com.sequenceiq.environment.environment.service.stack.StackService;
+import com.sequenceiq.flow.core.FlowConstants;
+import com.sequenceiq.flow.domain.FlowLog;
+import com.sequenceiq.flow.service.flowlog.FlowLogDBService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import javax.ws.rs.BadRequestException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class StackPollerProvider {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StackPollerProvider.class);
+
+    private final StackService stackService;
+
+    private final FlowLogDBService flowLogDBService;
+
+    public StackPollerProvider(
+        StackService stackService,
+        FlowLogDBService flowLogDBService) {
+        this.stackService = stackService;
+        this.flowLogDBService = flowLogDBService;
+    }
+
+    public AttemptMaker<Void> stackUpdateConfigPoller(List<String> stackCrns, Long envId, String flowId) {
+        List<String> mutableCrnsList = new ArrayList<>(stackCrns);
+        return () -> {
+            LOGGER.info("Attempting to update pillar information for {} clusters for environment id {}",
+                mutableCrnsList.size(), envId);
+            Optional<FlowLog> flowLog = flowLogDBService.getLastFlowLog(flowId);
+            if (flowLog.isPresent() && flowLog.get().getCurrentState().equals(FlowConstants.CANCELLED_STATE)) {
+                return AttemptResults.finishWith(null);
+            }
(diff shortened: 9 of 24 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        List<String> mutableNamesList = new ArrayList<>(stackNames);
+        return () -> {
+            Optional<FlowLog> flowLog = flowLogDBService.getLastFlowLog(flowId);
+            if (flowLog.isPresent() && flowLog.get().getCurrentState().equals(FlowConstants.CANCELLED_STATE)) {
+                return AttemptResults.finishWith(null);
+            }
+            List<String> remaining = new ArrayList<>();
+            List<AttemptResult<Void>> results = collectStackUpdateConfigResults(mutableNamesList,
+                remaining, envId);
+            mutableNamesList.retainAll(remaining);
```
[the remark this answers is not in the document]
wonderslug: No, it's not out of scope.  Its definition is outside the lambda.  So it gets maintained.  The function that is returned keeps getting run and referencing the scope with the mutableNamesList since it was defined in that scope.    Lambdas are weird.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
```diff
@@ -205,4 +205,18 @@ SimpleEnvironmentResponses deleteMultipleByCrns(Set<String> crns,
...
+    @Path("/crn/{crn}/update_config")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.UPDATE_CONFIG_BY_CRN, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
+        nickname = "updateConfigsInEnvironmentByCrnV1")
+    void updateConfigsInEnvironmentByCrn(@PathParam("crn") String crn);
+
+    @POST
+    @Path("/name/{name}/update_config")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.UPDATE_CONFIG_BY_NAME, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
```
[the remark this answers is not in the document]
wonderslug: Just have it in there to maintain consistency of how Env service functions are defined.   I'm 50/50 on it being there.  The cluster name reuse is limited by tenant anyways.   But in this case since we are removing auth checks and it won't have an x-cdp-actor-crn it might be best to remove it.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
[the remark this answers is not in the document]
jamisonbennett: Thanks. You are right. I missed that.

jamisonbennett:
LGTM but I don't since it outside of the area I usually work on I don't think my approval is sufficient.

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
```diff
@@ -0,0 +1,20 @@
+package com.sequenceiq.environment.environment.flow.config.update.event;
+
+import static com.sequenceiq.environment.environment.flow.config.update.event.EnvStackConfigUpdatesStateSelectors.FAILED_ENV_STACK_CONIFG_UPDATES_EVENT;
+
+import com.sequenceiq.cloudbreak.common.event.Selectable;
+import com.sequenceiq.flow.reactor.api.event.BaseFailedFlowEvent;
+
+public class EnvStackConfigUpdatesFailedEvent extends BaseFailedFlowEvent implements Selectable {
```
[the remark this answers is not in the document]
wonderslug: Yeah.  Removed

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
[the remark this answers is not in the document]
wonderslug: Removed the call

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
```diff
@@ -0,0 +1,75 @@
...
+        if (CollectionUtils.isNotEmpty(stackNames)) {
+            Polling.stopAfterAttempt(attemptCount)
+                .stopIfException(true)
+                .waitPeriodly(sleepTime, TimeUnit.SECONDS)
+                .run(attemptMaker);
+        }
+    }
+
+    private List<String> getUpdateableStacks(String envCrn) {
+        StackViewV4Responses stackViewV4Responses = stackV4Endpoint.list(0L, envCrn, false);
```
lacikaaa, line 69: if there is some error calling CB here, it's not really handled. Is it on purpose to leave it to the handler to handle it?

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
```diff
@@ -0,0 +1,75 @@
...
+        this.stackPollerProvider = stackPollerProvider;
+    }
+
+    public void updateStackConfigurations(Long envId, String envCrn, String flowId) {
+        List<String> stackNames = getUpdateableStacks(envCrn);
+        startStackConfigUpdatePolling(envCrn, stackNames,
+            stackPollerProvider.stackUpdateConfigPoller(stackNames, envId, flowId));
+    }
+
+    private void startStackConfigUpdatePolling(String envCrn, List<String> stackNames,
```
lacikaaa, line 58: `envCrn` is not used

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
```diff
@@ -0,0 +1,75 @@
...
+        AttemptMaker<Void> attemptMaker) {
```
lacikaaa, line 59: unnecessary linebreak

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
```diff
@@ -0,0 +1,75 @@
...
+        if (CollectionUtils.isNotEmpty(stackNames)) {
+            Polling.stopAfterAttempt(attemptCount)
```
lacikaaa, line 61: maybe we should switch to `stopAfterDelay`, os the attempt count won't have affect on how long we poll this. what do you think?

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        List<String> remaining, Long envId) {
+        return stackNames.stream()
+            .map(stackName -> fetchStackUpdateConfigResults(remaining, stackName, envId))
+            .collect(Collectors.toList());
+    }
+
+    private AttemptResult<Void> fetchStackUpdateConfigResults(List<String> remainingStacks,
+        String stackName, Long envId) {
+        try {
+            stackService.triggerConfigUpdateForStack(stackName);
```
lacikaaa, line 61: why don't we use crn here? I think crn provides some additional uniqueness

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private List<AttemptResult<Void>> collectStackUpdateConfigResults(List<String> stackNames,
+        List<String> remaining, Long envId) {
+        return stackNames.stream()
+            .map(stackName -> fetchStackUpdateConfigResults(remaining, stackName, envId))
+            .collect(Collectors.toList());
+    }
+
+    private AttemptResult<Void> fetchStackUpdateConfigResults(List<String> remainingStacks,
+        String stackName, Long envId) {
```
lacikaaa, line 59: env id not used
line break

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            stackService.triggerConfigUpdateForStack(stackName);
+            return AttemptResults.finishWith(null);
+        } catch (BadRequestException e) {
+            remainingStacks.add(stackName);
```
lacikaaa, line 64: so instead of passing in the `remainingStacks` paramater and adding stuff, why don't we return with `AttemptResult<String>`?

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    AttemptResult<Void> evaluateResult(List<AttemptResult<Void>> results) {
+        Optional<AttemptResult<Void>> error = results.stream()
+            .filter(it -> it.getState() == AttemptState.BREAK).findFirst();
+        if (error.isPresent()) {
+            return error.get();
+        }
+        long count = results.stream().filter(it -> it.getState() == AttemptState.CONTINUE).count();
+        if (count > 0) {
```
lacikaaa, line 78: I think you should use `anyMatch` here instead of `count` and `filter` if I understand the code well

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStackConfigUpdateService.java
```diff
@@ -0,0 +1,40 @@
...
+    public EnvironmentStackConfigUpdateService(
+        EnvironmentService environmentService,
+        EnvironmentReactorFlowManager reactorFlowManager) {
+        this.environmentService = environmentService;
+        this.reactorFlowManager = reactorFlowManager;
+    }
+
+    public void updateAllStackConfigsByCrn(String crn) {
+        String accountId = Crn.fromString(crn).getAccountId();
+        EnvironmentDto environment = environmentService.getByCrnAndAccountId(crn, accountId);
```
lacikaaa, line 26: why are you fetching the env here if you fetch it in `triggerStackConfigUpdates` again? this method actually uses the same `findByResourceCrnAndAccountIdAndArchivedIsFalse`, just calls a converter later

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
```diff
@@ -306,4 +311,10 @@ public Object getCreateEnvironmentForCli(EnvironmentRequest environmentRequest)
         Credential credential = credentialService.getByNameForAccountId(environmentRequest.getCredentialName(), accountId, ENVIRONMENT);
         return environmentService.getCreateEnvironmentForCli(environmentRequest, credential.getCloudPlatform());
     }
+
+    @Override
+    @DisableCheckPermissions
```
lacikaaa, line 316: are you sure about this?

wonderslug: Yes.   This is going to eventually be an automated event that happens that will have no actual end-user initiating it so it won't have someone other than "internal actor" permissions to check against.   Validation of the manual repair action is happening at the environment API endpoint.  This API call will not be exposed as a public API and will not have an issue of external calls being made to it.

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
```diff
@@ -22,6 +22,8 @@
     public static final String VERIFY_CREDENTIAL_BY_CRN = "Verifies the credential used by the given environment.";
     public static final String CLI_COMMAND = "produce cli command input for environment creation";
     public static final String GET_CRN_BY_NAME = "Get the crn of an environment by name.";
+    public static final String UPDATE_CONFIG_BY_CRN = "Update the configuration for all stacks in the Environment by the Environment CRN";
+    public static final String UPDATE_CONFIG_BY_NAME = "Update the configuration for all stacks in the Environment by the Environment name";
```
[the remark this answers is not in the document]
wonderslug: Yeap forgot to remove it when I pulled the update by name call.

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
horadla23: you can authorize this similar to changeTelemetryFeatures method

```
@CheckPermissionByResourceCrn(action = AuthorizationResourceAction.EDIT_ENVIRONMENT)
    public DetailedEnvironmentResponse changeTelemetryFeaturesByEnvironmentCrn(@ResourceCrn String crn, @Valid FeaturesRequest request)
```
unless you have a good reason for it, please do not use `@DisabledCheckPermissions`

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/config/EnvStackConfigUpdatesFlowConfig.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public EnvStackConfigUpdatesStateSelectors[] getInitEvents() {
+        return new EnvStackConfigUpdatesStateSelectors[]{ENV_STACK_CONFIG_UPDATES_START_EVENT};
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Update all Environment Stacks configuration";
```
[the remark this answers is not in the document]
wonderslug: Fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/handler/StackConfigUpdatesHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    public StackConfigUpdatesHandler(EventSender eventSender,
+        StackPollerService stackPollerService) {
+        super(eventSender);
+        this.stackPollerService = stackPollerService;
+    }
+
+    @Override
+    public String selector() {
+        return EnvStackConfigUpdatesHandlerSelectors.STACK_CONFIG_UPDATES_HANDLER_EVENT
+            .name();
```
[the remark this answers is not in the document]
wonderslug: Ok  im all for abstraction, fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+                Map<Object, Object> variables) {
+
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateEnvironmentStatusAndNotify(context, payload,
+                        getCurrentStatus(payload.getResourceId()),
+                        ResourceEvent.ENVIRONMENT_STACK_CONFIGS_UPDATE_FINISHED,
+                        EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FINISHED_STATE);
+                metricService.incrementMetricCounter(MetricType.ENV_STACK_CONFIG_UPDATE_FINISHED,
+                    environmentDto);
+                LOGGER.info("Flow entered into ENV_STOP_FINISHED_STATE");
```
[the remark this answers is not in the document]
wonderslug: Fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+        return new AbstractEnvStackConfigUpdatesAction<>(EnvStopFailedEvent.class) {
+            @Override
+            protected void doExecute(CommonContext context, EnvStopFailedEvent payload,
+                Map<Object, Object> variables) {
+                LOGGER.warn(
+                    String.format("Failed to update environments stack configs '%s'. Status: '%s'.",
+                        payload.getEnvironmentDto(), payload.getEnvironmentStatus()),
+                    payload.getException());
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateFailedEnvironmentStatusAndNotify(context, payload,
```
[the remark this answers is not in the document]
wonderslug: The environment state itself is being maintained.  
This flow does not really effect the state of the environment it is effecting the 
clusters in it.  The overall envrionment state should not change from what it is
unless its effect by the state of Environment level clustrers in it,  live FreeIPA 
which is not hit by this.

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
wonderslug: As I said above, this is going to be an automatically called function that will not have an external user.  Do we have Internal Actor support that will allow it to have permissions checked against?

horadla23: What do you mean under automatically called event? Will it be called in flow? If you are not calling it with internal actor explicitly, in threadlocal we will find the an user Crn.

regardless of this is an automatically called method, if someone make this API call manually, they can call it withouth authz using DisabledChekPermissions annotation.

horadla23: but you can check readme of authorization-common module to find out how you can make it possible to call with internal actor

wonderslug: So what I mean by automatically is that eventually, the FreeIPA HA recovery will detect a crash of a node and bring up a replacement node.  Not a flow, more like a result of an action from a "cron'ed" status check.  This recovery is manual right now but will be automatic and this API call will be made to update the pillar config across an environment.  So there will be no thread-local user outside of one for the application itself, internal actor.  

As well its an internal to the Control Plane API call, not available external.  I think this is a difference in philosophy in parts of the application.  In general from a thunderhead standpoint only publicly available API calls check, usually by the API service that fronts the internal service.  Internal services are openly available to be used by any other microservice.  From a CB standpoint, we are checking every API call every time, no matter if it is public or not.  This makes it very hard for anything like an automated service to be able to make calls to CB based services.  It has no "user" to work with except for an Internal Actor which would have access to everything anyways.

horadla23: since anyone can make a cdp cli command or ui call using cb related apis, this is not true and every cb api can become public anytime, thus we enforce authorization, cc @lnardai

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.statemachine.action.Action;
+
+@Configuration
+public class EnvStackConfigUpdatesActions {
+
+    private static final Logger LOGGER = LoggerFactory
+        .getLogger(EnvStackConfigUpdatesActions.class);
```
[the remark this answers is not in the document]
wonderslug: fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+        this.environmentStatusUpdateService = environmentStatusUpdateService;
+        this.metricService = metricService;
+    }
+
+    @Bean(name = "STACK_CONFIG_UPDATES_START_STATE")
+    public Action<?, ?> collectClusterInfo() {
+        return new AbstractEnvStackConfigUpdatesAction<>(EnvStackConfigUpdatesEvent.class) {
+
+            @Override
+            protected void doExecute(
```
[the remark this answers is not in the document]
wonderslug: fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+            @Override
+            protected void doExecute(CommonContext context, ResourceCrnPayload payload,
+                Map<Object, Object> variables) {
+
+                EnvironmentDto environmentDto = environmentStatusUpdateService
+                    .updateEnvironmentStatusAndNotify(context, payload,
+                        getCurrentStatus(payload.getResourceId()),
+                        ResourceEvent.ENVIRONMENT_STACK_CONFIGS_UPDATE_FINISHED,
+                        EnvStackConfigUpdatesState.STACK_CONFIG_UPDATES_FINISHED_STATE);
+                metricService.incrementMetricCounter(MetricType.ENV_STACK_CONFIG_UPDATE_FINISHED,
```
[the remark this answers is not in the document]
wonderslug: On the metrics side I doesnt look like there is a start on the other env service
flows,  so I followed the pattern

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
wonderslug: The proper method for this seems to be the @InternalOnly annotation.

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
```diff
@@ -0,0 +1,120 @@
...
+                    environmentDto);
+                LOGGER.info("Flow entered into ENV_STOP_FINISHED_STATE");
+                sendEvent(context, FINALIZE_ENV_STACK_CONIFG_UPDATES_EVENT.event(), payload);
+            }
+        };
+    }
+
+    @Bean(name = "STACK_CONFIG_UPDATES_FAILED_STATE")
+    public Action<?, ?> failedAction() {
+        return new AbstractEnvStackConfigUpdatesAction<>(EnvStopFailedEvent.class) {
```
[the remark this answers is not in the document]
wonderslug: Yeah  thats where the class I removed from Jamisons code review was supposed to go    
well back in there now.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        FlowLogDBService flowLogDBService) {
+        this.stackService = stackService;
+        this.flowLogDBService = flowLogDBService;
+    }
+
+    public AttemptMaker<Void> stackUpdateConfigPoller(List<String> stackNames, Long envId, String flowId) {
+        List<String> mutableNamesList = new ArrayList<>(stackNames);
+        return () -> {
+            Optional<FlowLog> flowLog = flowLogDBService.getLastFlowLog(flowId);
+            if (flowLog.isPresent() && flowLog.get().getCurrentState().equals(FlowConstants.CANCELLED_STATE)) {
```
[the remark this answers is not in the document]
wonderslug: I tried that.  Couple of issues with it.  

Biggest issue is that it will not stop the 
flow if it is running on a different instance of the environment service. 
EnvironmentInMemoryStateStore just ends up as a local ConcurrentHashMap.

Other issue is that even locally you have no way to guarantee that the CANCELED value
is picked up by the sleeping polling instance before the the new Polling instance starts 
and resets it to POLLING since it is only referenced by environment id.
   
Both of these led to just checking the state of the flow and if it has been 
canceled or not.  This allows for cross instance control and explicit stopping 
of the prior flow instance.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: Because the stack name gets added to the remainingStacks and is processed 
by reference in stackUpdateConfigPoller.  AttemptResult<String>  would not
allow it to be processed in the as I have to have a return type of AttemptResult<Void>
to handle all the cases and force a check against null in the results.

This is the same pattern used for the Datahub polling to stop multiple datahubs.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: Can do it, but it ends up being the same thing.  Not sure which is more clear. 

```
Predicate<AttemptResult<Void>> stateCheck = it -> it.getState() == AttemptState.CONTINUE;
if (results.stream().anyMatch(stateCheck)) {
```
versus

```
long count = results.stream().filter(it -> it.getState() == AttemptState.CONTINUE).count();
if (count > 0) {
```

I think the `anyMatch` adds more overhead needing an additional object, but in the scheme of things  it really makes no difference.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: This simple fronts the CB core stackV4Endpoint.updatePillarConfigurationByName which is 
name based and the CB core pattern for all of these is name based.   

Do we want to start shifting to CRN based for those APIs?

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: fixed

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackService.java
```diff
@@ -0,0 +1,65 @@
...
+            throw bre;
+        }
+    }
+
+    public void cancelRunningStackConfigUpdates(Environment environment) {
+        try {
+            FlowLog flowLog = flowLogDBService.getLastFlowLogByResourceCrnOrName(environment.getResourceCrn());
+            if (!flowLog.getFinalized() && flowLog.getFlowType() != null
+                && flowLog.getFlowType().getCanonicalName()
+                .equals(EnvStackConfigUpdatesFlowConfig.class.getCanonicalName())) {
```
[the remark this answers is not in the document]
wonderslug: yeah seems clearer

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackService.java
```diff
@@ -0,0 +1,65 @@
...
+            LOGGER.info(String
+                .format("Unable to start config update for stack %s.  Message is %s", stackName,
+                    bre.getMessage()));
+            throw bre;
+        }
+    }
+
+    public void cancelRunningStackConfigUpdates(Environment environment) {
+        try {
+            FlowLog flowLog = flowLogDBService.getLastFlowLogByResourceCrnOrName(environment.getResourceCrn());
```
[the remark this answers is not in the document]
wonderslug: I think in this case we want it to fetch from the DB again.  Since it is how we are checking for a running flow across multiple instances of the service, we need to check and not depend on a possible slow cache.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentEditRequest.java
```diff
@@ -45,6 +45,9 @@
     @ApiModelProperty(EnvironmentModelDescription.AWS_PARAMETERS)
     private AwsEnvironmentParameters aws;
 
+    @ApiModelProperty(EnvironmentModelDescription.UPDATE_STACKS_CONFIG)
+    private Boolean updateStacksConfig = false;
```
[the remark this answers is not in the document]
wonderslug: Yeah removed. old hanger on

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
wonderslug: fixed

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
wonderslug: fixed

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
wonderslug: Yeah makes sense.  Provides more percise control of how long it runs.

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackPollerService.java
wonderslug: Yes if its unable to get a list of Stacks form CB then the whole flow should fail.

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackService.java
```diff
@@ -0,0 +1,65 @@
...
+        FlowLogDBService flowLogDBService) {
+        this.stackV4Endpoint = stackV4Endpoint;
+        this.flowCancelService = flowCancelService;
+        this.flowLogDBService = flowLogDBService;
+    }
+
+    public void triggerConfigUpdateForStack(String stackName) {
+        try {
+            stackV4Endpoint.updatePillarConfigurationByName(0L, stackName);
+        } catch (BadRequestException bre) {
```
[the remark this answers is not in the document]
wonderslug: Looking at it probabbly only should catch WebApplicationException here.  
The rethrow allows the discrimination for the BadRequest to be retried 
and anything else (like a 404 or 401) to cancel the polling.  
But im open to options on it.  Thoughts?

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStackConfigUpdateService.java
wonderslug: Yeah that can be cleaned up a little.

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
wonderslug: Im not sure how "anyone" is able to do this.  This has to be a service deployed within the k8s CP or have an external API exposed via and API services for the CLI have to be defined in the CDP sdk and API documents.  This is the reason we have things like the environment service in the thunderhead fronting various things like the CB environment service or the datalake API service in thunderhead to front the CB datalake service.

What we are talking about is controlled access by internal trusted services within the control plane.  Are you saying there is some other way that we expose these APIs to the public?

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import javax.ws.rs.BadRequestException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class StackPollerProvider {
```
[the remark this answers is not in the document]
wonderslug: yeah get some in tomorrow

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
lnardai: You are correct, the cdp-cli works through the API services. But if you check the UI many API calls are not relying on the API services but go through another route, called dps-gateway. For that reason we can't just leave API's "unprotected", we need to specify at least they should used only internally, otherwise with the right URL they are accessible.

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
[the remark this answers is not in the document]
jamisonbennett: Now, this class is referenced but it is never instantiated. So the flow will never trigger a failure action because this type is the only type that is selected by `FAILED_ENV_STACK_CONIFG_UPDATES_EVENT`. This means that `STACK_CONFIG_UPDATES_FAILED_STATE` is logically unreachable.

I think the action code or handler code should have a `catch (Exception e)` which creates an `EnvStackConfigUpdatesFailedEvent` and sends that event.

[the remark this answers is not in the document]
wonderslug: Yeah. made some assumptions we talked about.  Added

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
lacikaaa: so inter service communication should use crn. see @horadla23 comment on cb developer channel on slack:

> Adam Laszlo Horvath  15:19
> I have a request to everyone who is designing, adding, modifying any APIs in CB code. If it is possible, please use CRN of a resource instead of name, since generally CRN is used in the whole platform to identify a resource and also during authorization it requires additional database query to find out the CRN based on the name, because we need CRN to authorize a call in UMS.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
lacikaaa: thanks for the clarification

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackService.java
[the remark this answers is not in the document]
lacikaaa: you misinterpreted what I said. If you check the implementation, `getLastFlowLogByResourceCrnOrName` will fetch the environment again to get it's id. While you already has the environment and could use it's id.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
lacikaaa: yeah, that's not better, but this is what I was trying to suggest:
```
		boolean continueResult = results.stream().anyMatch(result -> result.getState() == AttemptState.CONTINUE);
        if (continueResult) {
            return AttemptResults.justContinue();
        }
```
or even better moving this to a method:
```
        if (shouldContinue(results)) {
            return AttemptResults.justContinue();
        }
...
    private boolean shouldContinue(List<AttemptResult<Void>> results) {
        return results.stream().anyMatch(result -> result.getState() == AttemptState.CONTINUE);
    }
```

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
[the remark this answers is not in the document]
lacikaaa: then what's the purpose of invoking this method in these steps? only sending notification?
I'm just confused why we need to lookup the env from db, update the status to the same, save it back if it's state won't change. maybe this is the right way to send notification in case of env service, it's just looks there are some extra steps we don't need here.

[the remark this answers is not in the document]
wonderslug: Its to update the ResourceEvent status text that's associated with the Environment.  When the flow is going it updates those in the events view of the Clusters for the config.  This way there is a notification of the event to the customer without adjusting the status of the environment.

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: sounds good to me.

environment/src/main/java/com/sequenceiq/environment/environment/service/stack/StackService.java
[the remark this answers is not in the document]
wonderslug: k fixed

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StackPollerProvider.class);
```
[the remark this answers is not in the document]
wonderslug: added

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
wonderslug: K done

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
```diff
@@ -59,7 +59,9 @@
     START_FREEIPA_STARTED("Starting Free IPA"),
     START_FREEIPA_FAILED("Failed to Start Free IPA"),
 
-    FREEIPA_DELETED_ON_PROVIDER_SIDE("Free IPA deleted on cloud provider side");
+    FREEIPA_DELETED_ON_PROVIDER_SIDE("Free IPA deleted on cloud provider side"),
+
+    STACKS_CONFIG_UPDATE_IN_PROGRESS("Stack configuration updates in progress");
```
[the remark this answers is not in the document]
jamisonbennett: I still don't see this used...

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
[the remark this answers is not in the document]
lacikaaa: thanks for the clarification

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
```diff
@@ -0,0 +1,37 @@
...
+    public EnvStackConfigUpdatesFailedEvent(EnvironmentDto environmentDto, Exception exception, EnvironmentStatus environmentStatus) {
+        super(FAILED_ENV_STACK_CONIFG_UPDATES_EVENT.name(), environmentDto.getResourceId(),
+                environmentDto.getName(), environmentDto.getResourceCrn(), exception);
+        this.environmentDto = environmentDto;
+        this.environmentStatus = environmentStatus;
+    }
+
+    @Override
+    public String selector() {
+        return FAILED_ENV_STACK_CONIFG_UPDATES_EVENT.name();
```
lacikaaa, line 27: nit: this could be also `event()`

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStackConfigUpdateService.java
```diff
@@ -0,0 +1,37 @@
...
+        EnvironmentReactorFlowManager reactorFlowManager) {
+        this.environmentService = environmentService;
+        this.reactorFlowManager = reactorFlowManager;
+    }
+
+    public void updateAllStackConfigsByCrn(String envCrn) {
+        String accountId = Crn.fromString(envCrn).getAccountId();
+        if (accountId == null) {
+            throw new NotFoundException(String.format("No account id found for crn '%s'", envCrn));
+        }
```
lacikaaa, line 27: I think if you switch to `safeFromString` method it should guarantee that it's a valid crn and it has account id or an Exception would be thrown

wonderslug: in

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/event/EnvStackConfigUpdatesFailedEvent.java
wonderslug: in

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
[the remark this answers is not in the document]
wonderslug: fixed

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
```diff
@@ -275,10 +275,10 @@ FlowIdentifier setClusterMaintenanceMode(@PathParam("workspaceId") Long workspac
     FlowIdentifier updateSaltByName(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name);
 
     @PUT
-    @Path("{name}/pillar_config_update")
+    @Path("crn/{crn}/pillar_config_update")
```
doktoric, line 278: I am not sure that is it affects the UI but probably @keyki know this

doktoric: @horadla23 any authorization is required here ?

doktoric:
what is happening if the the update flow fails. I mean what is the next step ? just asking because in that case probably we need a retry ?

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
wonderslug: The whole controller was set to @DisableCheckPermissions by @keyki

wonderslug: It should not affect the UI.  I added this in the just previous PR to be called by this explicitly.  It will never have been used by the UI.

horadla23: i have another pr regarding this stack api authorization, for now it is good with disabledcheckpermissions annotation

wonderslug:
If the flow fails it is possible that some cluster is in a bad state and not received the refresh of the pillars.  Flow steps are being added to force a pillar update on start as well as a force update of the pillar info for a cluster that needs it.  The env flow will retry clusters in "working" flow states for two hours at least to try to force an update before the env flow itself fails.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
horadla23: change regarding stack api authorization is merged, please use same right to check permission as used for other stack api methods `@CheckPermissionByAccount(action = AuthorizationResourceAction.USE_INTERNALLY)` and also annotate crn param with `@TenantAwareParam` and it should work

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/urlparsers/V4ExistingResourceByCrnOrNameRestUrlParser.java
```diff
@@ -18,11 +18,11 @@
     // v4/{workspaceId}/blueprints/name/{name}
     // v4/{workspaceId}/blueprints/crn/{name}
     private static final Pattern PATTERN = Pattern
-            .compile("v4/(\\d+)/(blueprints|image_catalogs|recipes|stacks|distrox|cluster_templates)/(?:name|crn)/([^/]+)");
+            .compile("v4/(\\d+)/(blueprints|image_catalogs|recipes|stacks|distrox|cluster_templates)/(?:name|crn)/([^/]+)(/\\w+)?");
```
lacikaaa, line 21: could you add some test for this change? either via `RestUrlParserTest` or you could create a new test class
thanks

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentStackConfigUpdateService.java
```diff
@@ -0,0 +1,37 @@
...
+        EnvironmentReactorFlowManager reactorFlowManager) {
+        this.environmentService = environmentService;
+        this.reactorFlowManager = reactorFlowManager;
+    }
+
+    public void updateAllStackConfigsByCrn(String envCrn) {
+        String accountId = Crn.safeFromString(envCrn).getAccountId();
+        if (accountId == null) {
+            throw new NotFoundException(String.format("No account id found for crn '%s'", envCrn));
+        }
```
lacikaaa, line 27: nit: this is not necessary as `safeFromString` will throw an exception in case crn doesn't match the pattern, which enforces the accountid can't be empty.

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/urlparsers/V4ExistingResourceByCrnOrNameRestUrlParser.java
wonderslug: The RestUrlParserTest already catches this as part of the test.
testIfRestUrlParserMissing goes through every URL pattern defined in the controllers and makes sure all of the defined URLs have a parser.  So for every existing stack URla don't the new one get caught and the pattern applied to.  The reason for the change to V4ExistingResourceByCrnOrNameRestUrlParser.java was that the new URL pattern with crn and an action on the crn was failing.  So testIfRestUrlParserMissing checks all URL patterns against the V4ExistingResourceByCrnOrNameRestUrlParser Pattern.

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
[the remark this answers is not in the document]
lacikaaa: still present but not used

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
```diff
@@ -93,6 +93,7 @@
 
     public static final String PROXYCONFIG_NAME = "Name of the proxyconfig of the environment.";
     public static final String PROXYCONFIG_RESPONSE = "ProxyConfig attached to the environment.";
+    public static final String UPDATE_STACKS_CONFIG = "Update the configuration for all Stacks in the Environment.";
```
lacikaaa, line 96: this is not used

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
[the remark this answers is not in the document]
lacikaaa: please add some test as this class has some complex logic

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/urlparsers/V4ExistingResourceByCrnOrNameRestUrlParser.java
lacikaaa: thanks for clarification!

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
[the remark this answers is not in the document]
wonderslug: added. accidentally closed it when I went on a resolving rampage

environment/src/main/java/com/sequenceiq/environment/environment/poller/StackPollerProvider.java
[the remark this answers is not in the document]
wonderslug: added. accidentally closed it when I went on a resolving rampage

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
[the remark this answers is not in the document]
wonderslug: well its not now damnit!  :)

environment/src/main/java/com/sequenceiq/environment/environment/flow/config/update/EnvStackConfigUpdatesActions.java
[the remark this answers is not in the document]
wonderslug: fixed

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
wonderslug: added and in

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
```diff
@@ -293,11 +293,10 @@ FlowIdentifier updateSaltByName(@PathParam("workspaceId") Long workspaceId, @Pat
             @AccountId @QueryParam("accountId") String accountId);
 
     @PUT
-    @Path("{name}/pillar_config_update")
+    @Path("crn/{crn}/pillar_config_update")
```
doktoric, line 296: I think this change broke the API compatibility

lacikaaa: in that case please leave here the original one, and it should throw an exception and have deprecated annotation

wonderslug: This was added by me in CB-6364 and is not used by anything other than this call from environment service.  It has not been pushed out and I think CRN is the proper way to go for specificity.  

And as we talked about here.
https://github.com/hortonworks/cloudbreak/pull/8640#discussion_r463534811

lacikaaa: yes, but unfortunately if it's in the older branch we don't have a way to ignore this and even if we merge this, all the other PR will fail with this swagger issue. there will be a procedure to remove unused APIs, but currently the only thing we can do is to keep this one, in the implementation throw an exception, and create the right one with crn. I'm not happy with this, but there is no other way around it.

wonderslug: done

PR #8643, opened 2020-07-27
[ASRG] Azure DB is created based on Resource Group region instead of …

gergopapi2:
…ENV region

With the existing Resource Group feature, it can happen that the ENV region and the RG region is different. In this case, Redbeams should use the ENV region instead of the RG region otherwise it won't be able to create the VNet rules, because the VNets are based in the ENV region.


keyki:
Please incllude the jira id.

gergopapi2:
> Please include the jira id.

I do not know why I recently do not like jira IDs :D Thanks again for spotting it.

PR #8644, opened 2020-07-27
CB-8014: implement freeipa force delete

keyki:
CB-8014: implement freeipa force delete

- introducing new flag on FreeIpaV1Endpoint.delete() in freeipa microservice: boolean forced
- During environment deletion the forcedelete flag of environmentDeleteRequest will be passed to the freeipe deletion request
- in case of force delete the freeipa service will delete the freeipa related entities even if error occured during the deletion process


freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/StackTerminationService.java
```diff
@@ -32,7 +32,7 @@ public void finishStackTermination(StackTerminationContext context, TerminateSta
     public void handleStackTerminationError(Stack stack, StackFailureEvent payload, boolean forced) {
         if (!forced) {
             Exception errorDetails = payload.getException();
-            String stackUpdateMessage = "Termination failed: " + errorDetails.getMessage();
+            String stackUpdateMessage = " Termination failed: " + errorDetails.getMessage();
```
lacikaaa, line 35: is this space really necessary? there is no other change in this file

PR #8646, opened 2020-07-27
CDPCP-2702. Remove support for syncing OID group mappings

aarman-cloudera:


keyki:
Is this PR rebased on top of master?

aarman-cloudera:
> So the IT fails with a swagger validation error which is valid since we're removing a field from the API which is not backward compatible.
> 
> ```
> [0m - incompatible request params
> �[36mswagger-diff_1        |�[0m   - post /sdx/envcrn/{}/ranger_cloud_identity_mapping
> �[36mswagger-diff_1        |�[0m     - missing request param: azureGroupMapping (in: body, type: Hash[string, string])
> �[36mswagger-diff_1        |�[0m check to /json/datalake-swagger-2.27.0-b92.json
> �[36mswagger-diff_1        |�[0m - incompatible request params
> �[36mswagger-diff_1        |�[0m   - post /sdx/envcrn/{}/ranger_cloud_identity_mapping
> �[36mswagger-diff_1        |�[0m     - missing request param: azureGroupMapping (in: body, type: Hash[string, string])
> ```
> 
> In what version did we introduce this? If this is in 2.27 then we can ignore, if it's 2.26 we need to keep the value and deprecate it.
> 
> So this was introduced in `2.26.0-b52` so you should keep the API model and deprecate it and you can remove all of its usage.

It was added in 2.26. Okay, I've deprecated it instead. Note that this property is currently not being used anywhere, so I'd like to remove it soon.

keyki:
@aarman-cloudera we will branch out 2.27 this week, after that we can remove this field on master.

PR #8649, opened 2020-07-27
CB-8158-ForcedDownscale

smaniraju:
Forced Downscale should pass DOWNSCALE_FACTOR_IN_NODE_COUNT as QueryParam and not HeaderParam.
Forced Downscale should be limited to MandatoryDownscale NodeCount

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
@@ -2,11 +2,9 @@ package com.sequenceiq.periscope.monitor.evaluator.load;
 
 import static com.sequenceiq.periscope.monitor.evaluator.ScalingConstants.DEFAULT_MAX_SCALE_UP_STEP_SIZE;
 
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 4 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
```diff
@@ -23,33 +21,15 @@
 
     public List<String> getYarnRecommendedDecommissionHostsForHostGroup(String clusterCrn, YarnScalingServiceV1Response yarnResponse,
             Map<String, String> hostFqdnsToInstanceId, int maxAllowedDownScale, Optional<Integer> mandatoryDownScaleCount) {
```
smaniraju, line 23: mandatoryDownScale Count will always be less than maxAllowedDownScale, but based on Yarn Workaround of passing downscaleNodeCount = clusterSize, forced downscaling should be limited to mandatoryDownScale.
So that hostGroup is not brought down to LoadAlertConfiguration:minNodeCount even though there is load on cluster.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
```diff
@@ -23,33 +21,15 @@
...
-            List forcedCandidates = hostFqdnsToInstanceId.keySet().stream()
-                    .filter(nodeId -> !consideredNodeIds.contains(nodeId))
-                    .limit(forcedCandidatesCount).collect(Collectors.toList());
-
-            LOGGER.info("Forced downscaling candidates count '{}', candidates '{}' in cluster '{}' to achieve mandatory " +
-                            "downscaletarget '{}'.", forcedCandidates.size(), forcedCandidates, clusterCrn,
-                    mandatoryDownScaleCount.orElse(0));
-            decommissionNodes.addAll(forcedCandidates);
-        }
-
```
smaniraju, line 52: Based on YarnBased downscaling, this code has become dead block, so removing to avoid confusion.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
```diff
@@ -23,33 +21,15 @@
 
     public List<String> getYarnRecommendedDecommissionHostsForHostGroup(String clusterCrn, YarnScalingServiceV1Response yarnResponse,
             Map<String, String> hostFqdnsToInstanceId, int maxAllowedDownScale, Optional<Integer> mandatoryDownScaleCount) {
-        Set<String> consideredNodeIds = new HashSet<>();
-        Integer allowedDownscale = Math.max(maxAllowedDownScale, mandatoryDownScaleCount.orElse(0));
+        Integer allowedDownscale = mandatoryDownScaleCount.orElse(maxAllowedDownScale);
```
sidseth, line 24: Looking at this independently, it would seem that if mandatoryDownscale is not present, we default to maxAllowedDownscale.
That isn't actually what happens with loadBased downscaling. If there's no mandatory downscale, the request to YARN would not have maxAllowedDownscale candidates. So reading this function independently can be quite confusing.
I think it's worthwhile noting that "yarnResponse" passed to this function may or may not contain maxAllowedDownscale candidates.

smaniraju: maxAllowedDownscale is always passed, even in loadBased downscaling without mandatory downscale that is within limit, maxAllowedDownScale is passed based on policy limit.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
smaniraju: Added comment regarding yarnResponse might not have maxAllowedDownscale candidates

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnResponseUtils.java
smaniraju: Added comment regarding yarnResponse might not have maxAllowedDownscale candidates

PR #8650, opened 2020-07-27
CB-7917 certificate renewal support for clusters with generated self-signed certificates

biharitomi:


biharitomi:
I have modified a salt state to update the `FRONTEND_URL` in `cm.settings` and restart CM to take effect of this config change based on a regular expression engined replace pattern.
This way we can support certificate renewal on clusters where the certificate generation and also the public DNS entry registration has been failed at creation.

orchestrator-salt/src/main/resources/salt/salt/gateway/cm.sls
```diff
@@ -10,6 +10,22 @@ add_knox_settings_to_cm:
...
+update_frontend_url_of_cm:
+  file.replace:
+      - name: /etc/cloudera-scm-server/cm.settings
+      - pattern: 'setsettings FRONTEND_URL.*'
+      - repl: 'setsettings FRONTEND_URL https://{{ salt['pillar.get']('gateway:userfacingfqdn') }}/'
+      - unless: grep "^setsettings FRONTEND_URL https://{{ salt['pillar.get']('gateway:userfacingfqdn') }}/$" /etc/cloudera-scm-server/cm.settings
+
+restart_cm_after_fronted_url_change:
+  service.running:
+    - name: nginx
```
risdenk, line 23: This is supposed to restart CM? This looks like it checks nginx?

biharitomi: yes, @keyki has already caught that one, so fixed it, and fortunately my cluster is still running so it could be tested quickly :)

PR #8651, opened 2020-07-27
Flow documentation improvement

sodre90:


flow/README.md
```diff
@@ -51,13 +51,286 @@ You can implement the following interfaces for the flow to be more flexible:
...
+            return new HelloWorldFailedEvent(payload.getResourceId(), ex);
+        }
+    }
+}
+```
+*getFailurePayload* method is called when somehow an unexpected exception happens in your action's *doExecute*. It should return with your failure payload, 
+usually it contains an exception.
+#### Handlers
+If you have to do something that can takes long time, like a remote call or polling something, then you shouldn’t do it in an action’s *doExecute* method. 
+If you do this then it will consump one reactor thread and after a while it can happen there will be no available threads for reactor to process events for new
```
lnardai, line 294: *consume

PR #8653, opened 2020-07-28
CB-8136 Change PlatformResourceController to use environmentCrn

horadla23:


environment-api/src/main/java/com/sequenceiq/environment/api/v1/platformresource/EnvironmentPlatformResourceEndpoint.java
```diff
@@ -0,0 +1,163 @@
...
+    @GET
+    @Path("machine_types")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = OpEnvDescription.GET_VMTYPES_BY_CREDENTIAL, produces = MediaType.APPLICATION_JSON, notes = CONNECTOR_NOTES,
+            nickname = "getVmTypesByCredentialByEnv")
+    PlatformVmtypesResponse getVmTypesByCredential(
+            @QueryParam("environmentCrn") @NotEmpty String environmentCrn,
+            @QueryParam("region") String region,
+            @QueryParam("platformVariant") String platformVariant,
+            @QueryParam("availabilityZone") String availabilityZone,
```
lnardai, line 49: Aren't these (region, platformVariant, availabilityZone) redundant from environment perspective?

horadla23: Possibly, to be honest, I didn't checked, only duplicated the API with environment CRN change, we can spend more time to check it and make it easier to use

PR #8656, opened 2020-07-28
CDPCP-2673. Poll status of refresh when syncing OIDs

aarman-cloudera:
When setting cloud identity settings in ranger, the SDX service
now also returns a status object that the caller can use to
poll the status of the sync (which is just the status of the underlying
CM refresh command). The FreeIPA service polls this object to verify
that the sync is successful.

In order to simplify the design, the cloud identity sync code
has been modified to only work for single datalake per environment
and fail if there are multiple datalakes present (which we currently
don't support anyway).

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
@@ -203,8 +203,14 @@ public interface SdxEndpoint {
     @POST
     @Path("/envcrn/{envCrn}/ranger_cloud_identity_mapping")
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "set ranger cloud identity mapping", produces = MediaType.APPLICATION_JSON, nickname = "setRangerCloudIdentityMapping")
+    @ApiOperation(value = "Set ranger cloud identity mapping", produces = MediaType.APPLICATION_JSON, nickname = "setRangerCloudIdentityMapping")
     RangerCloudIdentitySyncStatus setRangerCloudIdentityMapping(@PathParam("envCrn") @ValidCrn String envCrn,
             @NotNull @Valid SetRangerCloudIdentityMappingRequest request);
+
+    @GET
+    @Path("/envcrn/{envCrn}/ranger_cloud_identity_sync_status")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Get status of a ranger cloud identity sync", produces = MediaType.APPLICATION_JSON, nickname = "getRangerCloudIdentitySyncStatus")
+    RangerCloudIdentitySyncStatus getRangerCloudIdentitySyncStatus(@PathParam("envCrn") @ValidCrn String envCrn, long commandId);
 
 }
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/RangerCloudIdentitySyncStatus.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/RangerCloudIdentitySyncStatus.java
@@ -1,4 +1,51 @@
 package com.sequenceiq.sdx.api.model;
 
+import io.swagger.annotations.ApiModelProperty;
+
+import javax.validation.constraints.NotNull;
+
 public class RangerCloudIdentitySyncStatus {
+
+    @ApiModelProperty
+    private Long commandId;
+
+    @ApiModelProperty
+    @NotNull
+    private RangerCloudIdentitySyncState state;
+
+    @ApiModelProperty
+    private String statusReason;
+
+    public Long getCommandId() {
+        return commandId;
+    }
+
+    public RangerCloudIdentitySyncState getState() {
+        return state;
+    }
+
+    public String getStatusReason() {
+        return statusReason;
+    }
+
+    public void setCommandId(Long commandId) {
+        this.commandId = commandId;
+    }
+
+    public void setState(RangerCloudIdentitySyncState state) {
+        this.state = state;
+    }
+
+    public void setStatusReason(String statusReason) {
+        this.statusReason = statusReason;
+    }
+
+    @Override
+    public String toString() {
+        return "RangerCloudIdentitySyncStatus{" +
+                ", commandId=" + commandId +
+                ", state=" + state +
+                ", statusReason='" + statusReason + '\'' +
+                '}';
+    }
 }
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
@@ -1,9 +1,12 @@
 package com.sequenceiq.datalake.cm;
 
 import com.cloudera.api.swagger.client.ApiException;
-import com.sequenceiq.datalake.controller.exception.RangerCloudIdentitySyncException;
+import com.cloudera.api.swagger.model.ApiCommand;
+import com.google.common.collect.Iterables;
 import com.sequenceiq.datalake.entity.SdxCluster;
 import com.sequenceiq.datalake.service.sdx.SdxService;
+import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncState;
+import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
@@ -11,7 +14,7 @@ import org.springframework.stereotype.Service;
 import javax.inject.Inject;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 @Service
 public class RangerCloudIdentityService {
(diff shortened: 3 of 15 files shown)


datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
```diff
@@ -24,21 +27,87 @@
...
+    }
+
+    private RangerCloudIdentitySyncStatus setAzureCloudIdentityMapping(String envCrn, SdxCluster sdxCluster, Map<String, String> azureUserMapping) {
         String stackCrn = sdxCluster.getStackCrn();
-        LOGGER.info("Updating azure cloud id mappings for datalake stack crn = {}", stackCrn);
+        LOGGER.info("Updating azure cloud id mappings for envCrn = {}, datalake stack crn = {}", envCrn, stackCrn);
         try {
-            clouderaManagerRangerUtil.setAzureCloudIdentityMapping(stackCrn, azureUserMapping);
+            if (!clouderaManagerRangerUtil.isCloudIdMappingSupported(sdxCluster.getStackCrn())) {
+                return sucessfulCloudIdSyncStatus("The datalake does not support cloud identity sync. Sync request is ignored.");
```
handavid, line 49: should we have an "unsupported" status instead of successful? successful seems misleading to me.

datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
```diff
@@ -24,21 +27,87 @@
...
+            LOGGER.error("Multiple datalakes per environment not supported, environmentCrn = {}", envCrn);
+            throw new IllegalStateException("Multiple datalakes per environment not supported");
+        }
+        return Optional.of(Iterables.getOnlyElement(sdxClusters));
+    }
+
+    public RangerCloudIdentitySyncStatus setAzureCloudIdentityMapping(String envCrn, Map<String, String> azureUserMapping) {
+        Optional<SdxCluster> sdxCluster = getSdxCluster(envCrn);
+        if (sdxCluster.isEmpty()) {
+            return sucessfulCloudIdSyncStatus("No datlakes associated with the environment.");
```
handavid, line 76: typo: datalakes
also, this doesn't sound like a successful scenario. Is there a better status to describe what happened?

datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
```diff
@@ -24,21 +27,87 @@
...
 
-    public void setAzureCloudIdentityMapping(String environmentCrn, Map<String, String> azureUserMapping) {
-        List<SdxCluster> sdxClusters = sdxService.listSdxByEnvCrn(environmentCrn);
-        List<String> sdxClusterCrns = sdxClusters.stream().map(SdxCluster::getCrn).collect(Collectors.toList());
-        LOGGER.info("Setting Azure cloud id mappings for datalake clusters = {}, environment = {}", sdxClusterCrns, environmentCrn);
-        sdxClusters.forEach(sdxCluster -> setAzureCloudIdentityMapping(sdxCluster, azureUserMapping));
+    public RangerCloudIdentitySyncStatus getRangerCloudIdentitySyncStatus(String envCrn, long commandId) {
+        Optional<SdxCluster> sdxCluster = getSdxCluster(envCrn);
+        if (sdxCluster.isEmpty()) {
+            return sucessfulCloudIdSyncStatus("No datlakes associated with the environment.");
```
handavid, line 84: typo: datalakes
also, this doesn't feel like a successful condition. maybe NOT_APPLICABLE would make sense for both this case and the case where the CM doesn't support this feature?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -26,6 +29,13 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CloudIdentitySyncService.class);
 
+    private static final int SYNC_TIMEOUT_MS = 1000 * 4;
```
handavid, line 32: lets make these timeouts configurable

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -48,16 +58,49 @@ private void syncAzureObjectIds(Stack stack, UmsUsersState umsUsersState, BiCons
...
+            LOGGER.info("syncStatus = {}", syncStatus);
+            switch (syncStatus.getState()) {
+                case SUCCESS:
+                    LOGGER.info("Successfully synced cloud identity, envCrn = {}", envCrn);
+                    return;
+                case FAILED:
+                    LOGGER.error("Failed to sync cloud identity, envCrn = {}", envCrn);
+                    warnings.accept(envCrn, "Failed to sync cloud identity into environment");
+                    return;
+                default:
```
handavid, line 86: let's explicitly call out the ACTIVE case here

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -48,16 +58,49 @@ private void syncAzureObjectIds(Stack stack, UmsUsersState umsUsersState, BiCons
...
+                    LOGGER.info("Sync is in progress, retrying");
+                    try {
+                        Thread.sleep(SYNC_SLEEP_INTERVAL_MS);
+                    } catch (InterruptedException e) {
+                        LOGGER.error("Interrupted during cloud identity sync", e);
+                        warnings.accept(envCrn, "Interrupted during cloud identity sync");
+                    }
+                    long commandId = syncStatus.getCommandId();
+                    syncStatus = sdxEndpoint.getRangerCloudIdentitySyncStatus(envCrn, commandId);
```
handavid, line 95: it's possible to make this call and then not check the returned value because we timed out. I suggest moving the clock isAfter check to before the sleep.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/RangerCloudIdentitySyncStatus.java
```diff
@@ -0,0 +1,51 @@
+package com.sequenceiq.sdx.api.model;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import javax.validation.constraints.NotNull;
+
+public class RangerCloudIdentitySyncStatus {
+
+    @ApiModelProperty
+    private Long commandId;
```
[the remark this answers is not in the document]
aarman-cloudera: Well, we can have it on a SUCCESS as well (see RangerCloudIdentityService.toRangerCloudIdentitySyncStatus).

datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
aarman-cloudera: True, I've added a NOT_APPLICABLE status like you've mentioned.

aarman-cloudera: I think we can also get rid of this part if we only make the call from FreeIPA on RAZ enabled environments. I'll look into it as a separate item though.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
aarman-cloudera: hmm true, I've moved the check up higher.

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
```diff
@@ -202,7 +203,14 @@ SdxDatabaseRestoreStatusResponse getRestoreDatabaseStatusByName(@PathParam("name
     @POST
     @Path("/envcrn/{envCrn}/ranger_cloud_identity_mapping")
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "set ranger cloud identity mapping", produces = MediaType.APPLICATION_JSON, nickname = "setRangerCloudIdentityMapping")
-    void setRangerCloudIdentityMapping(@PathParam("envCrn") @ValidCrn String envCrn, @NotNull @Valid SetRangerCloudIdentityMappingRequest request);
+    @ApiOperation(value = "Set ranger cloud identity mapping", produces = MediaType.APPLICATION_JSON, nickname = "setRangerCloudIdentityMapping")
+    RangerCloudIdentitySyncStatus setRangerCloudIdentityMapping(@PathParam("envCrn") @ValidCrn String envCrn,
+            @NotNull @Valid SetRangerCloudIdentityMappingRequest request);
+
+    @POST
```
keyki, line 210: Is it need to be a POST request?

freeipa/src/main/java/com/sequenceiq/freeipa/configuration/CloudIdSyncConfig.java
```diff
@@ -0,0 +1,23 @@
+package com.sequenceiq.freeipa.configuration;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CloudIdSyncConfig {
+
+    @Value("${freeipa.cloudidsync.poller.timeoutMs:4000}")
```
keyki, line 9: Can you please remove the default values from here? The application.yml already contains them.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -165,6 +165,17 @@ public SdxCluster getByCrn(String userCrn, String clusterCrn) {
         }
     }
 
+    public SdxCluster getByClusterCrn(String clusterCrn) {
```
keyki, line 168: This method is not used anywhere.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
```diff
@@ -48,16 +58,59 @@ private void syncAzureObjectIds(Stack stack, UmsUsersState umsUsersState, BiCons
...
+
+            // The sync status represents a cloud identity sync that may still be in progress, which we need to poll to check for completion.
+            pollSyncStatus(syncStatus, envCrn, warnings);
         } catch (Exception e) {
             LOGGER.warn("Failed to set cloud identity mapping for environment {}", envCrn, e);
             warnings.accept(envCrn, "Failed to set cloud identity mapping");
         }
     }
 
+    private void pollSyncStatus(RangerCloudIdentitySyncStatus syncStatus, String envCrn, BiConsumer<String, String> warnings) {
```
keyki, line 73: Instead of writing another polling method, we should use the one that we already have. Example: https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/ClusterProxyService.java#L296

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
aarman-cloudera: Your right, changed it to GET

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/CloudIdentitySyncService.java
aarman-cloudera: Thanks, didn't know about that one, I've moved to using it.

handavid:
The swagger compatibility test is failing with
```
[36mswagger-diff_1        |[0m check to /json/datalake-swagger-2.27.0-b184.json
[36mswagger-diff_1        |[0m - incompatible response attributes
[36mswagger-diff_1        |[0m   - post /sdx/envcrn/{}/ranger_cloud_identity_mapping
[36mswagger-diff_1        |[0m     - missing default response
```
It looks like it is picking out the change in the response of this API from `void` to `RangerCloudIdentitySyncStatus`.

I think that should be a safe change since any old callers of this API won't care about the response body. Also, the only expected caller of this API is the FMS, which is updated in the PR to match.
@keyki @lacikaaa can we merge this in or is there something we should do to pass swagger compatibility?

lacikaaa:
if you merge it all the other pr will fail. if this should be a safe change it means that the swagger check should be fixed. @doktoric

handavid:
old schema
```
responses: {
default: {
description: "successful operation"
}
}
```

new schema
```
responses: {
200: {
description: "successful operation",
schema: {
$ref: "#/definitions/RangerCloudIdentitySyncStatus"
}
}
}```

is there a way to annotate the API to include a default response?

handavid:
@doktoric @lacikaaa Do you have any further guidance on this

lacikaaa:
unfortunately I don't know much about this swagger check. try the internal channel to look for somebody who can help you

keyki:
Adding a response to a previously void method should be safe. If we can't find any solution we should just ignore the check and merge. Following PRs needs to be rebased, but that is how it is anyways.

keyki:
So according to Richard the only option is to include the response objects on 2.27 as well.

handavid:
@keyki @lacikaaa I've rebased and it's passed the swagger compatibility test. Can you re-review? I think that Asif has addressed all your comments.

keyki:
Thanks.

PR #8657, opened 2020-07-28
CB-7490 avoid duplicate key value violates unique constraint "uk_user_crn" error during credential creation

schfeca75:
- Remove long running credential verification code from the transactional block to avoid long transacition duration
- With verification removal the UserPreferencesService.getExternalId() will be removed from the transactional block. This method call in the transaction was the main reason of the duplicate key error
- Remove notification sending from the transactional block

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/credential/service/CredentialService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/credential/service/CredentialService.java
@@ -179,17 +179,18 @@ public class CredentialService extends AbstractCredentialService implements Reso
         credential.setResourceCrn(credentialCrn);
         credential.setCreator(creatorUserCrn);
         credential.setAccountId(accountId);
+        Credential verifiedCredential = credentialAdapter.verify(credential, accountId).getCredential();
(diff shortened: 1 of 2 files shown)


environment/src/main/java/com/sequenceiq/environment/credential/service/CredentialService.java
```diff
@@ -179,12 +179,12 @@ public Credential create(Credential credential, @Nonnull String accountId, @Nonn
         credential.setResourceCrn(credentialCrn);
         credential.setCreator(creatorUserCrn);
         credential.setAccountId(accountId);
+        Credential verifiedCredential = credentialAdapter.verify(credential, accountId).getCredential();
+        if (verifiedCredential.getVerificationStatusText() != null) {
+            throw new BadRequestException(verifiedCredential.getVerificationStatusText());
+        }
         try {
             return transactionService.required(() -> {
```
lacikaaa, line 187: couldn't we just drop the whole transaction?
I'm not sure about the UMS call, but that could take up some time as it's an external call. If we can skip it it would be great.

but I would definitely move the notification out from the transaction

PR #8658, opened 2020-07-28
CB-7729 Retrieve errors from Data Lake Database backup/restore failures

hreeve-cloudera:
Pulls the status reason for failed database backup/restore operations from the
salt stderr ouput, instead of from the command comment. The is achieved by
adding a new YAML file used by SaltErrorResolver, stderrcommands.yaml. When a
command listed in that file is run, the stderr output will be used as the
resolved error message. The getErrorResultSummary() method in RunnerInfo now
returns the summary in map form for simpler parsing in SaltErrorResolver.

Tested with unit tests and by testing in my local cloudbreak dev environment.


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreStatusService.java
```diff
@@ -61,7 +61,15 @@ public void restoreDatabaseFinished(long stackId) {
 
     public void handleDatabaseRestoreFailure(long stackId, String errorReason, DetailedStackStatus detailedStatus) {
         clusterService.updateClusterStatusByStackId(stackId, Status.RESTORE_FAILED, errorReason);
-        stackUpdater.updateStackStatus(stackId, detailedStatus);
+        stackUpdater.updateStackStatus(stackId, detailedStatus, extractSaltErrorIfAvailable(errorReason));
         flowMessageService.fireEventAndLog(stackId, Status.UPDATE_FAILED.name(), DATALAKE_DATABASE_RESTORE_FAILED, errorReason);
     }
+
+    private String extractSaltErrorIfAvailable(String errorReason) {
+        String errors = "Error(s): ";
```
lajosrodek, line 69: Perhaps worth extracting as a (private) constant. 😄

hreeve-cloudera: Curse you, you found the one thing where I was like, "Should I make this a constant? Eh, it's not a big deal." :laughing:

I'll fix it in the next patch, but waiting to see if there are any other comments first.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltErrorResolver.java
```diff
@@ -49,12 +55,15 @@ public void init() {
             String file = FileReaderUtils.readFileFromClasspath("salt/errormessages.yaml");
             errorMessages = new Yaml().load(file);
             LOGGER.info("Error messages for salt: {}", errorMessages);
+            file = FileReaderUtils.readFileFromClasspath("salt/stderrcommands.yaml");
+            commandsWithStderrFailures = new Yaml().load(file);
+            LOGGER.info("Salt commands that will pull the failure from stderr: {}", commandsWithStderrFailures);
```
kkalvagadda1, line 60: Do we know when the error is available in salt/errormessages.yaml and salt/stderrcommands.yaml files?

hreeve-cloudera: I'm not sure I understand the question. In errormessages.yaml, the code is basically replacing any errors run by the listed processes with the error in the YAML file. In stderrcommands.yaml we're just listing some processes, and the error comes from the stderr output. If the stderr output doesn't exist, then it defaults to a message that just says, "Failed to execute: <process string>"

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreStatusService.java
lajosrodek: Sure, definitely not a burning issue. 😛

hreeve-cloudera:
recheck

PR #8659, opened 2020-07-28
CB-8189 Adding new SDX Medium Duty config.

orngejaket:
This reworks the SDX Medium Duty HA into a new shape and configuration.
Older MDHA blueprints are removed.

This is just the beginning of these changes, I expect service config changes.


core/src/main/resources/application.yml
```diff
@@ -285,14 +283,12 @@ cb:
...
         7.2.0 - Operational Database: Apache HBase, Phoenix=cdp-opdb-720;
         7.2.0 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-720;
-        7.2.0 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha-720;
         7.2.0 - Streams Messaging Heavy Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-720;
         7.2.0 - Streams Messaging Light Duty: Apache Kafka, Schema Registry, Streams Messaging Manager=cdp-streaming-small-720;
         7.2.0 - Flow Management Light Duty with Apache NiFi, Apache NiFi Registry=cdp-flow-management-small-720;
         7.2.0 - Flow Management Heavy Duty with Apache NiFi, Apache NiFi Registry=cdp-flow-management-720;
         7.2.0 - Data Discovery and Exploration=cdp-dde-720;
         7.2.1 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-721;
-        7.2.1 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha-721;
```
kmanamcheri, line 295: Is there a checks and balances on how can someone use this? I see that there is a 7.2.2 - SDX Medium Duty is on here. Does this mean in prod I can actually spin up a medium duty data lake? If not, where is the check being done? 

I know that in UI this is being filtered and only the Light Duty ones are shown. I guess my question is where is the config which lists what are the prod-ready ones?

orngejaket: I don't believe the config here dictates when something is production ready or not.  We have to use entitlements for that.  CB-8188 will enforce the entitlement for creating a medium duty datalake.

datalake/src/main/resources/runtime/7.2.2/aws/medium_duty_ha.json
```diff
@@ -83,7 +63,7 @@
           "size": 50
         }
       },
-      "nodeCount": 2,
```
kmanamcheri, line 86: I realized that IDBroker does have client-side load balancing. Can we leave this at 2 to enable that?

orngejaket: This isn't something we have verified.  We can add this in later.

orngejaket: Also RDC has issues when spinning up a datahub with 2 nodes of idbroker.

PR #8661, opened 2020-07-29
CB-5742-EnhanceErrorHandling

smaniraju:
Removed user validation for clusterproxy since it is not User Controlled Configuration and duplicate validations.
Update periscope cluster status based on both CB Stack and Cluster Status.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
@@ -33,7 +33,6 @@ import com.sequenceiq.periscope.converter.TimeAlertRequestConverter;
 import com.sequenceiq.periscope.converter.TimeAlertResponseConverter;
 import com.sequenceiq.periscope.domain.Cluster;
 import com.sequenceiq.periscope.domain.LoadAlert;
-import com.sequenceiq.periscope.domain.ScalingPolicy;
 import com.sequenceiq.periscope.domain.TimeAlert;
 import com.sequenceiq.periscope.service.AlertService;
 import com.sequenceiq.periscope.service.AutoscaleRecommendationService;
@@ -229,16 +228,6 @@ public class AlertController implements AlertEndpoint {
                 () -> {
                     validateAccountEntitlement(cluster);
                     validateSupportedHostGroup(cluster, json.getScalingPolicy().getHostGroup(), AlertType.LOAD);
-                    String requestHostGroup = json.getScalingPolicy().getHostGroup();
-                    cluster.getLoadAlerts().stream().map(LoadAlert::getScalingPolicy).map(ScalingPolicy::getHostGroup)
-                            .filter(hostGroup -> hostGroup.equalsIgnoreCase(requestHostGroup)).findAny()
-                            .ifPresent(hostGroup -> {
-                                throw new BadRequestException(messagesService
-                                        .getMessage(MessageCode.LOAD_CONFIG_ALREADY_DEFINED, List.of(cluster.getStackName(), requestHostGroup)));
-                            });
-                    clusterProxyConfigurationService.getClusterProxyUrl()
-                            .orElseThrow(() ->  new BadRequestException(
-                                    messagesService.getMessage(MessageCode.CLUSTER_PROXY_NOT_CONFIGURED, List.of(cluster.getStackName()))));
                 });
     }
 
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
@@ -121,11 +121,12 @@ public class DistroXAutoScaleClusterV1Controller implements DistroXAutoScaleClus
     private DistroXAutoscaleClusterResponse updateClusterAutoScaleConfig(Long clusterId,
             DistroXAutoscaleClusterRequest autoscaleClusterRequest) {
 
+        alertController.validateLoadAlertRequests(clusterId, autoscaleClusterRequest.getLoadAlertRequests());
+        alertController.validateTimeAlertRequests(clusterId, autoscaleClusterRequest.getTimeAlertRequests());
+
         try {
             transactionService.required(() -> {
                 clusterService.deleteAlertsForCluster(clusterId);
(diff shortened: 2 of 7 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
```diff
@@ -229,16 +228,6 @@ private void validateLoadAlert(Long clusterId, Optional<Long> alertId, LoadAlert
...
-                            .filter(hostGroup -> hostGroup.equalsIgnoreCase(requestHostGroup)).findAny()
-                            .ifPresent(hostGroup -> {
-                                throw new BadRequestException(messagesService
-                                        .getMessage(MessageCode.LOAD_CONFIG_ALREADY_DEFINED, List.of(cluster.getStackName(), requestHostGroup)));
-                            });
-                    clusterProxyConfigurationService.getClusterProxyUrl()
-                            .orElseThrow(() ->  new BadRequestException(
-                                    messagesService.getMessage(MessageCode.CLUSTER_PROXY_NOT_CONFIGURED, List.of(cluster.getStackName()))));
                 });
     }
```
smaniraju, line 233: ClusterProxy is not a user controlled config, hence removed this user facing validation and also all clusters are registered with clusterproxy.  
Multiple load-alerts in a request is already validated in API layer, hence removed duplicate check.

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
```diff
@@ -121,11 +121,12 @@ private DistroXAutoscaleClusterResponse createClusterJsonResponse(Cluster cluste
...
+        alertController.validateTimeAlertRequests(clusterId, autoscaleClusterRequest.getTimeAlertRequests());
+
         try {
             transactionService.required(() -> {
                 clusterService.deleteAlertsForCluster(clusterId);
-                alertController.validateLoadAlertRequests(clusterId, autoscaleClusterRequest.getLoadAlertRequests());
-                alertController.validateTimeAlertRequests(clusterId, autoscaleClusterRequest.getTimeAlertRequests());
                 alertController.createLoadAlerts(clusterId, autoscaleClusterRequest.getLoadAlertRequests());
                 alertController.createTimeAlerts(clusterId, autoscaleClusterRequest.getTimeAlertRequests());
                 asClusterCommonService.setAutoscaleState(clusterId, autoscaleClusterRequest.getEnableAutoscaling());
```
smaniraju, line 132: validation need not be within the transaction.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
```diff
@@ -39,19 +40,22 @@ public void onApplicationEvent(ClusterStatusSyncEvent event) {
...
-            LOGGER.debug("Updated cluster '{}' to running, CB Cluster status '{}'.", cluster.getStackCrn(), cbClusterStatus);
-        } else if (!cbClusterStatus.isAvailable() && RUNNING.equals(cluster.getState())) {
+            LOGGER.info("Updated cluster '{}' to Running, CB Stack Status '{}', CB Cluster Status '{}'.",
+                    cluster.getStackCrn(), statusResponse.getStatus(), statusResponse.getClusterStatus());
+        } else if (!clusterAvailable && RUNNING.equals(cluster.getState())) {
             clusterService.setState(cluster.getId(), ClusterState.SUSPENDED);
-            LOGGER.debug("Suspended cluster '{}', CB Cluster status '{}'", cluster.getStackCrn(), cbClusterStatus);
+            LOGGER.info("Suspended cluster '{}', CB Stack Status '{}', CB Cluster Status '{}'",
+                    cluster.getStackCrn(), statusResponse.getStatus(), statusResponse.getClusterStatus());
         }
```
smaniraju, line 59: Periscope status should be synced based on both cluster and stack status, since next upscale is supported only when both are in available status. Even though they both are in sync in certain stimulated slow CB scenarios there is small drift between these two status.

autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
sidseth: Is ClusterProxy always on? (Is there any way to disable it, via the CLI for example?) Cloudbreak still supports DIRECT and CLUSTER_PROXY (Tunnel.java)

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
sidseth: Thank you. Was going to ask just this question on how they differ.

Wasn't there a change recently where at least downscale worked even if a previous downscale had failed (or node deleted on provider side).
That's the series of jiras Amit had opened. cc @cegganesh84

cegganesh84: > Was going to ask just this question on how they differ.

stack -> represents the cloud, cluster -> represents the cm cluster. Yes, there can be drift. The first stack will be scaled to accommodate more nodes. Then the cluster will be scaled by the means of provisioning services. https://jira.cloudera.com/browse/CB-8051?focusedCommentId=2639540&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2639540. Extending this, will apply for downscale also in the reverse order.

autoscale/src/main/java/com/sequenceiq/periscope/controller/AlertController.java
smaniraju: Yes it is always on. The CLI tunnel just decides how CB talks to cluster but cluster proxy registration is always done. Also cluster access via IP i.e DIRECT is not always guaranteed.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
smaniraju: > Wasn't there a change recently where at least downscale worked even if a previous downscale had failed (or node deleted on provider side).

amit opened jiras were basically related to CB not identifying node deleted on provider side. It is not related to this change.

PR #8664, opened 2020-07-29
CB-8201 Missing cache for GrpcUmsClient.isAuthorizationEntitlementRegistered

sodre90:
mow-dev and mow-int are very slow... this can be a cause


horadla23:
getAccountDetails is cached

sodre90:
> getAccountDetails is cached
but it is called from the same class, so it is not called through Spring proxy object. My favorite interview question so far :)

horadla23:
> > getAccountDetails is cached
> 
> but it is called from the same class, so it is not called through Spring proxy object. My favorite interview question so far :)

ahh right :)

PR #8678, opened 2020-07-30
CB-8232 remove preferedSubnetId in case of env list

doktoric:
remove preferedSubnetId because that can causee performance problems and we would like to avoid that.


doktoric:
retest this please

PR #8679, opened 2020-07-31
CB-8234: Fix downscale repair when FQDN is missing

jamisonbennett:
Fix the case where the FreeIPA management service database has the
instancemetadata.discoveryfqdn null for the node which is being
repaired.

This was tested on a local deployment of cloudbreak.


jamisonbennett:
I want to ensure that @wonderslug reviews this before it is committed. So I marked as DO_NOT_MERGE.

lacikaaa:
should we cleanup instancemetadata without FQDN during repair?

jamisonbennett:
> should we cleanup instancemetadata without FQDN during repair?

Yes.

This change will allow this cleanup to happen. It will Ωcause the instance metadata to be cleaned up so long as the instance is selected for repair. At the end of the downscale flow for repair the FreeIPA instancemetatdata table is updated to have a TERMINATED instancestatus. So from that point onwards, the instance metadata without a discoveryfqdn will no longer be used by any code which is operating on non-terminated instances. Without this change, the repair used to fail.

PR #8684, opened 2020-07-31
CB-7930 adding UMSClient to tests so we can assign resourceRoles

lnardai:
- added example of assigning resource roles
- will add more testcases and datahub creation usecase

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignUmsRoleAction.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignUmsRoleAction.java
@@ -1,0 +1,33 @@
+package com.sequenceiq.it.cloudbreak.action.ums;
+
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.it.cloudbreak.UmsClient;
+import com.sequenceiq.it.cloudbreak.action.Action;
+import com.sequenceiq.it.cloudbreak.actor.CloudbreakUser;
+import com.sequenceiq.it.cloudbreak.actor.CloudbreakUserCache;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.ums.UmsTestDto;
+
+public class AssignUmsRoleAction implements Action<UmsTestDto, UmsClient> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AssignUmsRoleAction.class);
+
+    private final String userKey;
+
+    public AssignUmsRoleAction(String userKey) {
+        this.userKey = userKey;
+    }
+
+    @Override
+    public UmsTestDto action(TestContext testContext, UmsTestDto testDto, UmsClient client) throws Exception {
+        CloudbreakUser user = CloudbreakUserCache.getInstance().getByName(userKey);
+        LOGGER.info(String.format("Assigning resourceRole %s over resource %s for user ",
+                testDto.getRequest().getRoleCrn(), testDto.getRequest().getResourceCrn()), user.getCrn());
+        client.getUmsClient().assignResourceRole(user.getCrn(), testDto.getRequest().getResourceCrn(), testDto.getRequest().getRoleCrn(), Optional.of(""));
+        return testDto;
+    }
+}
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/AbstractEnvironmentTestDto.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/AbstractEnvironmentTestDto.java
@@ -11,6 +11,7 @@ import javax.ws.rs.BadRequestException;
 import org.apache.commons.lang3.NotImplementedException;
 
 import com.sequenceiq.it.cloudbreak.EnvironmentClient;
+import com.sequenceiq.it.cloudbreak.UmsClient;
 import com.sequenceiq.it.cloudbreak.action.Action;
 import com.sequenceiq.it.cloudbreak.assertion.Assertion;
 import com.sequenceiq.it.cloudbreak.context.RunningParameter;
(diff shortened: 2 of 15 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignUmsRoleAction.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private String userRoleCrn;
+
+    public AssignUmsRoleAction(String userKey, String userRoleCrn) {
+        this.userKey = userKey;
+        this.userRoleCrn = userRoleCrn;
+    }
+
+    @Override
+    public EnvironmentTestDto action(TestContext testContext, EnvironmentTestDto testDto, UmsClient client) throws Exception {
```
afarsang, line 28: if you can use the it framework log message too then it will be on the test report as well (it is logWhen, or something, I'll check for you if you couldn't find)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/AbstractEnvironmentTestDto.java
```diff
@@ -45,6 +46,10 @@ public T when(Action<T, EnvironmentClient> action) {
         return getTestContext().when((T) this, EnvironmentClient.class, action, emptyRunningParameter());
     }
 
+    public T assignRole(Action<T, UmsClient> action) {
```
afarsang, line 49: I wouldn't enhance the basic vocabulary (given, when, then, ...), isn't it a kind of "when" at least "whenAssignRole". So the basic idea is "given an entity/resource, when action, then assertion".

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/DatalakeDatahubCreateAuthTest.java
```diff
@@ -81,11 +81,11 @@ public void testCreateEnvironment(MockedTestContext testContext) {
                 .when(sdxTestClient.createInternal(), key(sdxInternal))
                 .awaitForFlow(key(sdxInternal))
                 .await(SdxClusterStatusResponse.RUNNING)
-                .when(sdxTestClient.describeInternal(), RunningParameter.who(Actor.useRealUmsUser(AuthUserKeys.MGMT_CONSOLE_ADMIN_B)))
+                .when(sdxTestClient.describeInternal(), RunningParameter.who(Actor.useRealUmsUser(AuthUserKeys.MGMT_CONSOLE_ADMIN_B, testContext)))
```
afarsang, line 84: Frankly I am not fond of the testContext argument in the test. It feels somehow strange (I'm already calling testContext's method, and shall I pass a testContext). Could we find some way to solve this?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/AbstractEnvironmentTestDto.java
lnardai: Sure, let's have a quick call about this. The generic interface was not working since the UmsClient can't inherit from the Microservices client, since it's a Grpc client, but if we can find out something happy to do so.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/DatalakeDatahubCreateAuthTest.java
lnardai: Agree, it's ugly but it's working, we should find a better way about this.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java
```diff
@@ -0,0 +1,97 @@
...
+import com.sequenceiq.it.cloudbreak.dto.credential.CredentialTestDto;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+import com.sequenceiq.it.cloudbreak.dto.environment.EnvironmentTestDto;
+import com.sequenceiq.it.cloudbreak.dto.ums.UmsTestDto;
+import com.sequenceiq.it.cloudbreak.mock.ITResponse;
+import com.sequenceiq.it.cloudbreak.mock.freeipa.FreeIpaRouteHandler;
+import com.sequenceiq.it.cloudbreak.spark.DynamicRouteStack;
+import com.sequenceiq.it.cloudbreak.testcase.AbstractIntegrationTest;
+
+public class CreateDhWithDatahubCreator extends AbstractIntegrationTest {
```
afarsang, line 26: it seems like a test although no test in the name, and testannotation on the method

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/environment/EnvironmentTestDto.java
```diff
@@ -370,4 +371,8 @@ private String getResourceGroupUsage() {
         return getTestParameter().get(ResourceGroupTest.AZURE_RESOURCE_GROUP_USAGE);
     }
 
+    @Override
+    public String getCrn() {
+        return getResponse().getCrn();
```
afarsang, line 376: getResponsecan be null

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/ums/UmsTestDto.java
```diff
@@ -0,0 +1,49 @@
...
+        setRequest(new AssignResourceRequest());
+    }
+
+    public UmsTestDto withDatahubCreator() {
+        getRequest().setRoleCrn(DH_CREATOR_CRN);
+        return this;
+    }
+
+    public UmsTestDto assignTarget(String key) {
+        // TODO castExceptionHandling
```
afarsang, line 35: agree :)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java
lnardai: I'll pick this up in another PR.

PR #8685, opened 2020-07-31
CB-8236: Fix repair to ensure freeipa servers are removed

jamisonbennett:
Ensure old deleted FreeIPA servers are cleaned up during repair. In
addition the the expected hostnames, check FreeIPA for unexpected
hostnames and add these hostnames to the list of hostnames are are
cleaned up during the downscale process of FreeIPA repair.

This was tested locally on a cluster that was failing due to
previously deleted FreeIPA servers still being listed in FreeIPA when
the 'ipa server-find' commmand was run.


freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/handler/CollectAdditionalHostnamesHandler.java
```diff
@@ -0,0 +1,67 @@
...
+import reactor.bus.EventBus;
+
+import javax.inject.Inject;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Component
+public class CollectAdditionalHostnamesHandler implements EventHandler<CollectAdditionalHostnamesRequest> {
```
lacikaaa, line 26: `EventHandler` is deprecated, could you switch to `ExceptionCatcherEventHandler`?

jamisonbennett:
I am marking this DO_NOT_MERGE since it depends on https://github.com/hortonworks/cloudbreak/pull/8700 and I was unsure of my interpretation of the intent of ExceptionCatcherEventHandler in https://github.com/hortonworks/cloudbreak/pull/8700.

PR #8686, opened 2020-07-31
CB-8207: Add --region flag to aws s3 cp command for datalake backup/restore

brycederriso:
Closes [CB-8207](https://jira.cloudera.com/browse/CB-8207)
* Add `aws_region` pillar value
* Add `--region` parameter to backup and restore script `aws` invocations 
* Add `aws_region` to the `BackupRestoreSaltConfigGenerator`
* Update test for `BackupRestoreSaltConfigGenerator`
* Pass in `Stack` object to `BackupRestoreSaltConfigGenerator`

Tested by:
* Running salt `state.apply postgresql.disaster_recovery.bakup` and `state.apply postgresql.disaster_recovery.restore` manually in Data Lake with external DB.
* Triggering backup + restore using the Datalake API while running with an external DB.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
@@ -5,6 +5,7 @@ import static com.sequenceiq.cloudbreak.common.type.CloudConstants.AZURE;
 
 import static java.util.Collections.singletonMap;
 
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.orchestrator.model.SaltConfig;
 import com.sequenceiq.cloudbreak.orchestrator.model.SaltPillarProperties;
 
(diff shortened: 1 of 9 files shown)


orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
```diff
@@ -1,3 +1,4 @@
 disaster_recovery:
   object_storage_url:
+  aws_region: '' # empty string to prevent salt from passing the literal 'None'
```
hreeve-cloudera, line 3: When/where is this actually filled in with the region, if applicable?

kkalvagadda1: you are defining a new pillar for this, right. Where is the application change that passes this information?

brycederriso: y'all are correct. I had originally written the issue to just include the salt script changes, but obviously we need a way to provide the region value.

I'm investigating how to determine the appropriate region and add it to the pillar.

brycederriso: I've pushed updated changes to the salt pillar config.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -17,17 +18,31 @@
 
 @Component
 public class BackupRestoreSaltConfigGenerator {
+    // these values are tightly coupled to the `postgresql/disaster_recovery.sls` salt pillar in `orchestrator-salt`
+    public static final String POSTGRESQL_DISASTER_RECOVERY_PILLAR_PATH = "/postgresql/disaster_recovery.sls";
```
keyki, line 22: This diesn't need to be public, others visibility can be lowered also to package level.

keyki: I'm merging the PR. We can address this later.

PR #8689, opened 2020-07-31
CB-7963 CB work for deploying RAZ to Datahub cluster

doktoric:


doktoric:
@orngejaket can you please help to test out the change ?

PR #8693, opened 2020-08-03
CB-8250 invoke the CM setuid script with the migration flag during up…

keyki:
…grade

As part of Upgrade we need to call the setuid script to configure deterministic
user and group ids. It can happen that the original setuid script is called
without the migration flag so I added an extra guard check to make sure it's
not called during upgrade. If the original installation called the script
alreeady we don't need to migrate anyways. Also in the Upgrade handler
we only updated the CM repo pillar and not the CM settings. However,
the CM settings contain an important upgrade to only call the setuid
script if the CM version is 7.2.1 or above. This was not updated previously
so upgrading from 7.2.0 to 7.2.1 didn't invoke the script.

This commit also fixes another issue when the incorrect Stack CRN was used
as ENV CRN and we didn't properly check for workloads.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -483,7 +483,7 @@ public class ClusterHostServiceRunner {
                         singletonMap("csd-urls", csdUrls))));
     }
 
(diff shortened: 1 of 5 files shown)


keyki:
I think the swagger check can be ignored. This PR is not touching anything API related.

doktoric:
the swagger failed because Execution failed for task ':autoscale:test'.

pdarvasi:
This has resolved this issue, too: https://jira.cloudera.com/browse/CB-8253

PR #8694, opened 2020-08-03
CB-7799 Fix missing information in Structured Events

nadamdb:
Fixed version inconsistencies, deleted ambari references, added tunnel type field to show if CCM is used or not, fixed database type always showing "embedded" and added more detailed information about the used databases.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/converter/RdsConfigToRdsDetailsConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/converter/RdsConfigToRdsDetailsConverter.java
@@ -4,6 +4,7 @@ import javax.inject.Inject;
 
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/converter/RdsConfigToRdsDetailsConverter.java
```diff
@@ -23,6 +24,11 @@ public RdsDetails convert(RDSConfig source) {
         rdsDetails.setConnectorJarUrl(source.getConnectorJarUrl());
         rdsDetails.setCreationDate(source.getCreationDate());
         rdsDetails.setDatabaseEngine(source.getDatabaseEngine().name());
+        if (DatabaseVendor.EMBEDDED == source.getDatabaseEngine()) {
```
topolyai5, line 27: next time you can use a simpler version of this:
`rdsDetails.setExternal(DatabaseVendor.EMBEDDED != source.getDatabaseEngine())`
if you want you can extract into a field as well.

PR #8700, opened 2020-08-03
CB-8261: Fix ExceptionCatcherEventHandler to set the event's data

jamisonbennett:
Fix ExceptionCatcherEventHandler to use the defaultFailureEvent's for
the event's data.

Added unit test.


jamisonbennett:
I want to ensure @sodre90 reviews this because I am making a guess as to the intended functionality. I tried to use ExceptionCatcherEventHandler for https://github.com/hortonworks/cloudbreak/pull/8685 and discovered that the defaultFailureEvent isn't being used in the data. I am not sure how this works for the many other places it is used, so maybe I am missing something.

sodre90:
> I want to ensure @sodre90 reviews this because I am making a guess as to the intended functionality. I tried to use ExceptionCatcherEventHandler for #8685 and discovered that the defaultFailureEvent isn't being used in the data. I am not sure how this works for the many other places it is used, so maybe I am missing something.

Ah, I see... In the second parameter of eventBus.notify the event was wrong and it was the original event instead of the new one what we want to send. Good catch, thanks!

PR #8708, opened 2020-08-04
CB-8138 - Lock-components should only lock used parcels

pdarvasi:
Lock components from now on only means locking down components that are actually used (have their parcel activated) on the cluster, eg. by default SDX only uses CDH and if Data Hub might use Nifi parcel then lock components should lock the Nifi version also.


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -507,6 +509,16 @@ public String getStackRepositoryJson(StackRepoDetails repoDetails, String stackR
         return null;
     }
 
+    @Override
+    public Map<String, String> gatherInstalledComponents(String stackName) {
+        try {
+        return clouderaManagerParcelService.getActivatedParcels(apiClient, stackName);
```
keyki, line 515: Missing indent. Please format the code.

PR #8709, opened 2020-08-04
CB-8213 [ASRG] dl and dh might leak resources in case of provisioning…

gergopapi2:
… error

In case datalake of datahub is provisioned but some error comes up during stack provisioning then the created cloud resources might not be persisted into cloudbreak resources table. This leads to resource leakage in the azure single RG use case during deletion.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
@@ -8,7 +8,6 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
-import org.apache.commons.collections4.ListUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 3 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
```diff
@@ -141,6 +141,24 @@
         return resources;
     }
 
+    private List<CloudResource> persistCloudResources(
+            AuthenticatedContext ac, CloudStack stack, PersistenceNotifier notifier, CloudContext cloudContext, String stackName,
+            String resourceGroupName, Deployment templateDeployment) {
+        List<CloudResource> allResourcesToPersist = new ArrayList<>();
+        List<CloudResource> instances;
+        try {
+            List<CloudResource> templateResources = azureCloudResourceService.getDeploymentCloudResources(templateDeployment);
```
keyki, line 150: We should add log messages to this method about what's happening.

gergopapi2: Yes, I added some messages.

keyki:
retest this please

PR #8712, opened 2020-08-04
CB-8275: Fix FreeIPA downscale to remove SSHFP DNS records

jamisonbennett:
When FreeIPA is downscaled or repaired, the SSHFP DNS records were fix
to be removed.

This was tested with unittests and using a local deployment of
cloudbreak.


jamisonbennett:
@wonderslug I chose to not do a host-del because the host-del will always fail and this means there no confirmation that the deletion actually deleted the SSHFP records. I would rather know for sure the delete has happened so I chose to delete the record specifically.

PR #8718, opened 2020-08-05
DISTX-527 Cleanup unused code based on CM Based Autoscaling.

smaniraju:
1. Removed code related to StackCollectorService,ClusterCreationEvaluator which has been replaced with ClusterSyncOnDemand.
2. Removed ClusterManagerHostHealthMonitor since Cluster Health Status is now being synced based on CB Cluster Status.
3. Removed FailedNode related code since monitoring failed nodes has been moved out of Periscope.
4. Removed unused clusterId based endpoints since clusterId based api will not be supported. And also since all endpoints are expected to be integrated with UMS Authorization which does not support local db-id based authorization.
5. Removed MetricsAlert and PrometheusAlerts related code since they are not supported and are also linked with Ambari Definition  and Ambari support for these alerts. If they are supported in future in CM, the structure, CM Integratation API and monitoring mechanism will differ. This also improve query performance since loading Cluster joins additional two tables related to metrics and prometheusalerts even though they are not supported.


autoscale-api/src/main/java/com/sequenceiq/periscope/api/endpoint/v2/AutoScaleClusterV2Endpoint.java
```diff
@@ -18,9 +18,13 @@
...
 import io.swagger.annotations.ApiOperation;
 
+/**
+ * @deprecated Database ID based endpoints are deprecated for removal.
+ */
 @Path("/v2/clusters")
 @Consumes(MediaType.APPLICATION_JSON)
 @Api(hidden = true, value = "/v2/clusters", description = CLUSTERS_DESCRIPTION, protocols = "http,https", consumes = MediaType.APPLICATION_JSON)
+@Deprecated(since = "CB 2.26.0", forRemoval = true)
 public interface AutoScaleClusterV2Endpoint {
```
smaniraju, line 29: This endpoint is used in prometheus service related metrics unittest, hence not removed.

smaniraju:
> Looks good. Could you please move the jira over to the CB project so that a fix version can be set properly.
> Couple of clarifications
> 
> * ClusterManagerHostHealthEvaluator not needed because CB already keeps track of this information?, and scaling will just fail if a cluster is unhealthy?
> * ClusterCreation* not required since this is looked up on demand when a policy is defined?

ClusterManagerHostHealthEvaluator, yes CB tracks cluster health and periscope syncs cluster state from CB. Related jira(CB-5738).
ClusterSync from CB is done on demand when the cluster is first accessed byName or byCrn. Related jiras(DISTX-447,CB-6619)

PR #8723, opened 2020-08-05
CB-4915 Respect region disablement policies of customer AWS accounts

biharitomi:
Regarding CB-4915 a new mechanism has been implemented that verifies that the credential which is under creation is able to query regions. If regions could not be described using the globally set default region which is `us-east-1` then it tries to describe regions in different EC2 regions. This list of possible default is configurable, but the defaults are stored in environment service's `application.yml`. Once the describe of regions works from a region then the region is persisted on credential side as a new attribute. This means that AWS related services will picked the credential specific region as default when it is possible/necessary.

During the tests it turned out this describe region process takes some time due to a bug in AWS Java SDK:
https://github.com/aws/aws-sdk-java/issues/2204
The SDK's retrying client identifies 401 `AuthFailure` responses as skew errors which is retryable this way and results in 30 failures/retries from the client side. But this doesn't prevent the customer to have a credential with proper default region configured, just the creation takes much more time then it's expected.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudCredentialStatus.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudCredentialStatus.java
@@ -10,6 +10,8 @@ public class CloudCredentialStatus {
 
     private final Exception exception;
 
+    private final boolean defaultRegionChanged;
+
     public CloudCredentialStatus(CloudCredential cloudResource, CredentialStatus status) {
         this(cloudResource, status, null, null);
     }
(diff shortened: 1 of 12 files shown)


environment/src/main/resources/application.yml
```diff
@@ -145,7 +145,9 @@ cb:
         services: s3,dynamodb
       interface:
         services: autoscaling,cloudformation,ec2,ecr.api,ecr.dkr,elasticfilesystem,elasticloadbalancing,sts
-    credential.cache.ttl: 60
+    credential:
+      cache.ttl: 60
+      default.regions: us-west-1,us-west-2,eu-central-1,eu-north-1,sa-east-1,ap-northeast-1,eu-west-1,ap-northeast-2,ap-southeast-2,ap-southeast-1
```
doktoric, line 150: hmmm
I am not sure this is the approach what we would like to follow. How about using aws-zone.json for this ?

biharitomi: hm, that is not configurable without a new version

biharitomi:
Retest this please

PR #8726, opened 2020-08-05
CB-8210 Store HBase WAL file in a directory different from HBase root…

LinaAtAustin:
… directory

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
@@ -223,6 +223,10 @@
         ],
         "serviceConfigs": [
           {
+            "name": "hbase_wal_dir",
+            "value": "/hbase-wals"
+          },
+          {
             "name": "service_config_suppression_hadoop_secure_web_ui",
             "value": "true"
           }


core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
```diff
@@ -222,6 +222,10 @@
           }
         ],
         "serviceConfigs": [
+          {
```
kmanamcheri, line 225: Are there unit-tests for this? We should update them.

You have made changes to 7.2.1. Should we make this change in the next version 7.2.2? That way we avoid breaking changes.

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
```diff
@@ -222,6 +222,10 @@
...
+            "name": "hbase_wal_dir",
+            "value": "/hbase-wals"
```
kmanamcheri, line 227: Does this directory need to exist in HDFS? Is it automatically created? If not, should we add code to make sure that this directory exists?

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
LinaAtAustin: I am getting familiar with Cloudbreak code. Will find where the test code is and add test code later.

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
LinaAtAustin: Blueprint just needs to add that configuration. CM will create this directory automatically. I have verified this at the host. With this change only, the folder "/hbase-wals" was created, and it contains data.

[root@lina-0804-aws-s3-1-master1 bin]# hdfs dfs -ls /
Found 4 items
drwxr-xr-x   - hbase hbase               0 2020-08-04 11:41 /hbase-wals
drwxrwxr-x   - solr  solr                0 2020-08-04 11:40 /solr-infra
drwxrwxrwt   - hdfs  supergroup          0 2020-08-04 11:40 /tmp

[root@lina-0804-aws-s3-1-master1 bin]# hadoop fs -du -s -h /hbase-wals
584.6 K  512.6 M  /hbase-wals
drwxr-xr-x   - hdfs  supergroup          0 2020-08-04 15:43 /user

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
keyki: This will go to the root disk which is not something we want I think. There are dynamic config providers for HBase [here](https://github.com/hortonworks/cloudbreak/tree/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase). 
Configuring the attached disk, here is an [example](https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/zookeeper/ZooKeeperVolumeConfigProvider.java#L23) where we configure the first disk only for Zookeper. Here is another example to configure [multiple disks](https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hdfs/HdfsVolumeConfigProvider.java#L28) for Hdfs

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
kmanamcheri: Ok sounds good.

core/src/main/resources/defaults/blueprints/cdp-sdx-721.bp
kmanamcheri: @keyki Hmm.. this goes into HDFS. Can we configure HDFS for different disks?

keyki: Oh, sorry. I didn't know it's on HDFS. In that case, I'm fine. Thanks for the clarification.

PR #8728, opened 2020-08-05
CB-8278: Backup using hdfs cli

brycederriso:
Closes [CB-8278](https://jira.cloudera.com/browse/CB-8278)
* Restore db using HDFS cli to transfer backups. Fix styling.
* Hack together a fix for embedded database backups.
* Adjust url construction and region passing to account for HDFS cli usage.
* Fix style issues.

Tested by triggering backup and restore using HTTP requests and against external and embedded database.
**Ranger admin group changes were merged in as well** after which I retested by triggering via HTTP requests.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
@@ -24,7 +24,7 @@ public class BackupRestoreSaltConfigGenerator {
 
     public static final String OBJECT_STORAGE_URL_KEY = "object_storage_url";
 
-    public static final String AWS_REGION_KEY = "aws_region";
+    public static final String DATABASE_BACKUP_POSTFIX = "_database_backup";
 
     public static final String RANGER_ADMIN_GROUP_KEY = "ranger_admin_group";
 
@@ -35,7 +35,6 @@ public class BackupRestoreSaltConfigGenerator {
 
         Map<String, String> disasterRecoveryValues = new HashMap<>();
         disasterRecoveryValues.put(OBJECT_STORAGE_URL_KEY, fullLocation);
-        disasterRecoveryValues.put(AWS_REGION_KEY, stack.getRegion());
         disasterRecoveryValues.put(RANGER_ADMIN_GROUP_KEY, rangerAdminGroup);
 
         servicePillar.put("disaster-recovery", new SaltPillarProperties(POSTGRESQL_DISASTER_RECOVERY_PILLAR_PATH,
@@ -46,10 +45,10 @@ public class BackupRestoreSaltConfigGenerator {
 
     private String buildFullLocation(String location, String backupId, String cloudPlatform) throws URISyntaxException {
         URI uri = new URI(location);
-        String suffix = '/' + backupId + "_database_backup";
+        String suffix = '/' + backupId + DATABASE_BACKUP_POSTFIX;
         String fullLocation;
         if (AWS.equalsIgnoreCase(cloudPlatform)) {
-            fullLocation = "s3://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
+            fullLocation = "s3a://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
         } else if (AZURE.equalsIgnoreCase(cloudPlatform)) {
             fullLocation = "abfs://" + uri.getSchemeSpecificPart().replaceAll("^/+", "");
         } else {
--- a/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
+++ b/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
@@ -1,4 +1,2 @@
 disaster_recovery:
   object_storage_url:
-  aws_region: '' # empty string to prevent salt from passing the literal 'None'
-
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -1,47 +1,36 @@
 #!/bin/bash
 # backup_db.sh
 # This script uses the 'pg_dump' utility to dump the contents of hive and ranger PostgreSQL databases as plain SQL.
-# After PostgreSQL contents are dumped, the SQL is uploaded to AWS or Azure using their CLI clients, 'aws s3 cp' and 'azcopy copy' respectively.
-# For the Azure upload, we must write the SQL commands to a local disk before running the `azcopy copy` command.
-# For AWS, the cli supports piping file contents to stdin, so we do that and avoid writing to a local file.
+# After PostgreSQL contents are dumped, the SQL file is uploaded using the hdfs cli.
+# For AWS or Azure, the url should start with `s3a://` or `abfs://`, respectively.
 
 set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ $# -ne 6 && $# -ne 7 ]]; then
+if [[ $# -ne 5 && $# -ne 6 ]]; then
   echo "Invalid inputs provided"
   echo "Script accepts 5 inputs:"
-  echo "  1. Cloud Provider (azure | aws)"
-  echo "  2. Object Storage Service url to place backups."
-  echo "  3. PostgreSQL host name."
-  echo "  4. PostgreSQL port."
-  echo "  5. PostgreSQL user name."
-  echo "  6. AWS Region option."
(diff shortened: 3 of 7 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
```diff
@@ -34,7 +34,6 @@ public SaltConfig createSaltConfig(String location, String backupId, Stack stack
 
         Map<String, String> disasterRecoveryValues = new HashMap<>();
         disasterRecoveryValues.put(OBJECT_STORAGE_URL_KEY, fullLocation);
-        disasterRecoveryValues.put(AWS_REGION_KEY, stack.getRegion());
```
keyki, line 37: Why is the region removed?

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
```diff
@@ -1,4 +1,2 @@
 disaster_recovery:
   object_storage_url:
-  aws_region: '' # empty string to prevent salt from passing the literal 'None'
```
doktoric, line 3: we can not really remove pillar property because then the older clusters will not work .

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
brycederriso: We got the suggestion to use the `hdfs` cli instead of s3 to upload/download.

It was my understanding that we don't have to specify the region in that case, the hdfs configurations should take care of things?

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
brycederriso: Ok, I'll readd it and make a comment that it is not used anywhere.

There's _no_ way to remove a pillar property once it's been added? Not without breaking older clusters?

keyki: We can remove it now since you just introduced it on master and it is not deployed anywhere. It only gets problematic if we would have released it.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
kkalvagadda1: @keyki  AWS CLI would depend on permissions granted to the instance profile of the datalake node which is currently LOG_ROLE. Grating more permissions to LOG_ROLE was not a good idea.
With this change, we will be using hdfs client to write to S3. This uses S3a. All the services which to S3, internally use S3a. With the configuration we are using, S3a fetches the region information of the bucket and then constructs the request accordingly.

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
kkalvagadda1: as @keyki  mentioned, this functionality is not yet released to customers so it this be fine.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -54,69 +42,50 @@ errorExit() {
...
+  fi
+}
+
+move_backup_to_cloud () {
+  LOCAL_BACKUP="$1"
+  kinit_as_hdfs
+  doLog "INFO Uploading to ${BACKUP_LOCATION}"
+
+  hdfs dfs -mkdir -p "$BACKUP_LOCATION"
+  HDFS_FILE_LOCATION="${BACKUP_LOCATION}/${SERVICE}_backup"
```
kkalvagadda1, line 62: It's not a HDFS location. It is the location of the object-store. Please rename it accordingly.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -54,69 +42,50 @@ errorExit() {
...
+    errorExit "Couldn't kinit as hdfs"
+  fi
+}
+
+move_backup_to_cloud () {
+  LOCAL_BACKUP="$1"
+  kinit_as_hdfs
+  doLog "INFO Uploading to ${BACKUP_LOCATION}"
+
+  hdfs dfs -mkdir -p "$BACKUP_LOCATION"
```
kkalvagadda1, line 61: here are were are creating a directory and accessing directly. S3 provides an eventual guarantee. 
In short, if you create a directory and try to write to it immediately, it might fail.

we use S3 guard to handle this issue with S3. S3 guard tracks all the operations and changes the metadata.

I'm not sure if S3 guard is enabled by default. Is there anything that has to be done for it. Can you check with @risdenk  on this?

risdenk: So there are potentially a few options here:
* retry on failure to upload
* Upload the whole folder instead of trying to create it then put files in it
* in theory s3guard is enabled so this isn't even something that needs to be worried about

brycederriso: I'm going to be lazy and rely on S3 guard to protect me from the eventual consistency problems.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
brycederriso: Changed to `OBJECT_STORE_PATH`

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/dr/BackupRestoreSaltConfigGenerator.java
brycederriso: I'm resolving this conversation, given that we have a reason and there's been no further concern.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -1,47 +1,37 @@
 #!/bin/bash
 # restore_db.sh
 # This script uses the 'psql' cli to drop hive and ranger databases, create them, then read in a plain SQL file to restore data.
-# We use the `azcopy copy` and `aws s3 cp` commands to copy backups from Azure or AWS respectively.
-# When using either cloud provider, the plaintext SQL files are first copied to the local filesystem, then fed into psql.
+# We retrieve the SQL files from a valid HDFS URL, which should be an `s3a://` or `abfs://` for AWS or Azure clouds respectively.
```
kkalvagadda1, line 4: URL is not an HDFS URL . You can just say " We retrieve the SQL files from a valid URL, ..."

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -1,47 +1,36 @@
 #!/bin/bash
 # backup_db.sh
 # This script uses the 'pg_dump' utility to dump the contents of hive and ranger PostgreSQL databases as plain SQL.
-# After PostgreSQL contents are dumped, the SQL is uploaded to AWS or Azure using their CLI clients, 'aws s3 cp' and 'azcopy copy' respectively.
-# For the Azure upload, we must write the SQL commands to a local disk before running the `azcopy copy` command.
-# For AWS, the cli supports piping file contents to stdin, so we do that and avoid writing to a local file.
+# After PostgreSQL contents are dumped, the SQL file is uploaded to an HDFS URL using the hdfs cli.
```
kkalvagadda1, line 4: URL is not an HDFS URL . You can just say " .... the SQL file is uploaded to the cloud storage using the hdfs cli."

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
brycederriso: Done

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
brycederriso: Done

PR #8730, opened 2020-08-06
CB-8312: Fix FreeIPA repair to set deleted node status to termiated

jamisonbennett:
When a FreeIPA repair is performed on an instance that was deleted on
the provider, the instance metadata's status was fixed so that it will
be set to terminated. This prevents the old instance for being
displayed on the Web UI.

The status checking was also suspended during the repair operation.
This ensures the status checker doesn't overwrite the terminated
status in the case where both the status check and repair are
concurrentlu updating the instance's metadata status.


lacikaaa:
please rebase your pr

lacikaaa:
I think it would be better to modify statuschecker to skip running is there is an active flow. the same has been implemented in CB/datalake

jamisonbennett:
> I think it would be better to modify statuschecker to skip running is there is an active flow. the same has been implemented in CB/datalake. this way we don't have to add this step to every flow

I missed this comment. I must have had my browser displaying the approval without refreshing to see this comment. I will see about making that in a separate PR.

PR #8732, opened 2020-08-06
CB-8324 Authz actions refactor to use json only for legacy rights

horadla23:


horadla23:
validated using old json and old model with this code:

```
@Test
    public void write() throws IOException {
        String actionsJson = FileReaderUtils.readFileFromClasspath("actions.json");
        JsonNode tree = JsonUtil.readTree(actionsJson);
        underTest.readActions();
        Arrays.stream(AuthorizationResourceAction.values()).forEach(action -> {
            AuthorizationResourceActionModel actionModel = new ObjectMapper().convertValue(tree.get(action.name()), AuthorizationResourceActionModel.class);
            System.out.println(action);
            assertEquals(action.getRight(), actionModel.getRight());
            assertEquals(action.getAuthorizationResourceType(), actionModel.getResourceType());
            assertEquals(actionModel.getLegacyRight(), underTest.getLegacyRight(action));
        });
    }
```

PR #8735, opened 2020-08-06
ENGESC-3405 Fix mpack validation issue for urls behind paywall

bergerdenes:
Neither on the fly validation on the UI upon entering (or changing) the
mpack URL, nor for the check upon registration did not send the basic
authorization header with the request thus getting a HTTP 401.
This had to be fixed in 2 places.


bergerdenes:
retest this please

PR #8737, opened 2020-08-06
DISTX-528 Periscope API Resource Based Authorizations.

smaniraju:
1. Periscope endpoints updated to use resource based UMS authorization.
2. Periscope API are authorized for AuthorizationResourceAction.SCALE_DATAHUB
3. Authorization annotation check enabled in periscope.
4. Unused DB-Id based AlertEndpoint removed.


autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
```diff
@@ -58,91 +66,102 @@
...
+
+    @Inject
+    private EntitlementValidationService entitlementValidationService;
+
     @Override
-    @CheckPermissionByAccount(action = AuthorizationResourceAction.DATAHUB_READ)
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.SCALE_DATAHUB)
     public List<DistroXAutoscaleClusterResponse> getClusters() {
         return distroXAutoscaleClusterResponseConverter.convertAllToJson(asClusterCommonService.getDistroXClusters());
     }
```
smaniraju, line 82: getClusters is still an AccountLevel Permission Check.

smaniraju:
@lnardai  The periscope api authorization has been changed from account level to resource level in-line with other datahub api authorization model. Please take a look.

autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterV2Controller.java
```diff
@@ -17,7 +18,8 @@
 /**
  * @deprecated Database ID based endpoints are deprecated for removal.
  */
-@Component
+@Controller
+@DisableCheckPermissions
```
horadla23, line 22: I think this should be changed to `@InternalOnly` to ensure that noone will call this

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
horadla23: i think scaledatahub right will be present only in resource roles, so it wont work with account level, so this will work only with poweruser, so it is better to change this `@DisabledCheckPermissions` since we decided to disable authz on list for now, right @lnardai ?

autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterV2Controller.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
smaniraju: ok added `@DisabledCheckPermissions` for listApi in sync with other listApi like distrox.

PR #8739, opened 2020-08-06
CB-7984 Proxies,Tags,Telemetry should be PowerUser Only APIs

sodre90:


environment/src/main/java/com/sequenceiq/environment/proxy/v1/controller/ProxyController.java
```diff
@@ -55,39 +55,45 @@ public ProxyController(ProxyConfigService proxyConfigService,
...
         String accountId = ThreadBasedUserCrnProvider.getAccountId();
         ProxyConfig config = proxyConfigService.getByNameForAccountId(name, accountId);
         return proxyConfigToProxyResponseConverter.convert(config);
     }
 
     @Override
-    @CheckPermissionByAccount(action = AuthorizationResourceAction.ENVIRONMENT_READ)
-    public ProxyResponse getByEnvironmentCrn(@ResourceCrn @TenantAwareParam String environmentCrn) {
+    @InternalOnly
+    public String getCrnByAccountIdAndName(String accountId, String name) {
```
horadla23, line 75: `accountId` should be annotated with `@AccountId`, if the method is internal only

sodre90: Thanks!

oleewere:
are you sure about this? like for account telemetry, you won't be able to see details about Telemetry (i mean because of the restricted read operations, i guess putting that to writes are fine)

horadla23:
> are you sure about this? like for account telemetry, you won't be able to see details about Telemetry (i mean because of the restricted read operations, i guess putting that to writes are fine)

yes, this is based on authz sharing design agreement

PR #8741, opened 2020-08-06
CB-8328. Implemented service type tagging for DataHub

oleewere:
later after the new metering heartbeat application is ready and burned on our images, we will use this custom tag to fill MeteringConfigView object to fill its serviceType field

more details in the commit message


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/servicetype/ServiceTypeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+        serviceTypes.remove(ServiceType.DATAHUB);
+        if (serviceTypes.size() == 0) {
+            LOGGER.debug("Service type only matched for dummy/default DataHub provider. Default DATAHUB service type will be used.");
+        } else if (serviceTypes.size() == 1) {
+            result = serviceTypes.iterator().next();
+            LOGGER.debug("The following service type will be selected based on the service type providers: {}", result.value());
+        } else {
+            LOGGER.warn("You cannot match on multiple service types based on the service type providers. " +
```
doktoric, line 56: maybe in case the user could not specify the servicetype then we should check the stack.type. What do you think?

oleewere: default service type is DATAHUB here, is the same as the stack type (that is not used for DATALAKE). metering v1 event supports only that. better to use that default enum (in case of DATAHUB name will change in the future or something), so i think it should be fine to use the default ServiceType enum instead of the stack type

petersomogyi:
Looks good from the OpDB side! Thanks @oleewere!

PR #8742, opened 2020-08-06
CB-8311 - Ranger authorizer should be enabled for datahub clusters

moresandeep:
CB-5949 Adds ranger authorization support for requests flowing through Knox. This only works for SDX cluster and not for DistroX clusters. This is because the logic to enable ranger authorizer checks if Ranger is installed and ranger is not installed on DistroX clusters but rather installed in SDX cluster. This PR proposes eliminating condition that checks whether ranger is installed on the cluster. By default all public cloud clusters have Ranger installed so this change will not affect any functionality.

**Testing**
This patch was locally tested (tested to see that the topology file is generated correctly)

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -295,7 +295,6 @@
              {%- endif %}
              {% if 'HUE' in exposed and 'HUE_LOAD_BALANCER' in salt['pillar.get']('gateway:location') -%}
              <param>
-             <param>
                  <name>HUE</name>
                  <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
              </param>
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -300,6 +294,7 @@
              </param>
              {%- endif %}
              {% if 'HUE' in exposed and 'HUE_LOAD_BALANCER' in salt['pillar.get']('gateway:location') -%}
+             <param>
```
keyki, line 297: Is this extra param needed?

moresandeep: ugh, good catch, that looks an artifact of rebase. I will fix it and resubmit the patch. Thanks

keyki:
@moresandeep can you please rebase your branch on top of master?

moresandeep:
> @moresandeep can you please rebase your branch on top of master?

Sure, just rebased my changes.

PR #8743, opened 2020-08-06
CB-8326 livy 3 added as a knox service and also added as supported se…

doktoric:
…rvice. Now we will be able to attach the right parcel to the cluster in case of livy 3


dabondor:
+1 LGTM (knox changes: https://gerrit.sjc.cloudera.com/c/cdh/knox/+/93119)

PR #8745, opened 2020-08-07
CB-8040 Parcels are activated without distribution in replaced VM

schfeca75:


cluster-api/src/main/java/com/sequenceiq/cloudbreak/cluster/api/ClusterApi.java
```diff
@@ -76,6 +76,10 @@ default int startCluster(Set<InstanceMetaData> hostsInCluster) throws Cloudbreak
         return clusterModificationService().gatherInstalledComponents(hostname);
     }
 
+    default void deactivateUnUsedComponents(Set<ClusterComponent> usedComponents) {
```
keyki, line 79: Unused instead of UnUsed

PR #8748, opened 2020-08-07
CB-8342 Cluster product components should not come from DefaultCDHInfo

pkedvessy:
DefaultCDHInfo was filled out with parcels in 2.18 last time from the application.yml. Parcel configurations got removed in 2.19 but the related java code remained in the codebase. As a part of CB-7647 parcels got filled out again in the DefaultCDHInfo based on the latest greatest prewarmed image - it was necessary for the stack matrix - but the remained parcel fillup caused that cluster product components are coming from the DefaultCDHInfo. As a part of this PR I'm simply remove the DefaultCDHInfo based product fill-up as it was a dead code since 2.19.


keyki:
I think there are other code paths as well where we add the default parcels. https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/service/ClouderaManagerClusterCreationSetupService.java#L192-L194

keyki:
Merged it as it solves the QA code path. We can look into the other path if needed in a different PR.

pkedvessy:
Thanks @keyki for calling out that there is an other code path. Follow up PR:
https://github.com/hortonworks/cloudbreak/pull/8761

PR #8749, opened 2020-08-07
CB-8346 modify RetryAndMetricsJerseyClientAspect to log 404 responses on debug level

biharitomi:


lacikaaa:
if you are already there, wouldn't it make sense to edit `JerseyClientRetryPolicy` and do not retry on not found? like why would we check if a stack/proxy config/whatever exists 5 times? if we have to poll for something we won't rely on this logic anyway

biharitomi:
> if you are already there, wouldn't it make sense to edit `JerseyClientRetryPolicy` and do not retry on not found? like why would we check if a stack/proxy config/whatever exists 5 times? if we have to poll for something we won't rely on this logic anyway

I think this client works as you expect it because it doesn't mark Bad requests and client errors 40x responses as retryable by default.

PR #8756, opened 2020-08-07
CB-8329: Fix DNS forwarding on FreeIPA replicas

jamisonbennett:
Fixed the DNS forwarding to use the DNS server provided by the
original DHCP lease rather than pointing to the master which points to
the one from the cloud provider.

Use the same logic for the master too eventhough it didn't matter.

This was validated by deploying a FreeIPA HA cluster with a local
cloudbreak deployment and validating the dns forwarder settting in
LDAP.


jamisonbennett:
I wanted to call out that shell check will report a warning for $FORWARDERS not being double-quoted. This is intentional so it will split into multiple arguments to support multiple DNS forwarders if needed. I made it so that the runtime provisioning will fail if someone does double quote it, that way the integration tests catch the error if someone inadvertently "corrects" the shell checker warning.

PR #8761, opened 2020-08-10
CB-8342 Cluster product components should not come from DefaultCDHInfo

pkedvessy:
DefaultCDHInfo was filled out with parcels in 2.18 last time from the application.yml. Parcel configurations got removed in 2.19 but the related java code remained in the codebase. As a part of CB-7647 parcels got filled out again in the DefaultCDHInfo based on the latest greatest prewarmed image - it was necessary for the stack matrix - but the remained parcel fillup caused that cluster product components are coming from the DefaultCDHInfo. As a part of this PR I'm simply remove the DefaultCDHInfo based product fill-up as it was a dead code since 2.19.

It is a followup PR on https://github.com/hortonworks/cloudbreak/pull/8748


keyki:
Should this go to 2.27 as well?

pkedvessy:
retest this please

PR #8762, opened 2020-08-10
CB-8243: Implement mock test related to the Ycloud hybrid cloud usecase

jszeles:
The mock tests are extended with a new Ycloud hybrid cloud related use case, which is similar to the handled scenario and parameters of AwsYcloudHybridCloudTest e2e test. The test is executed against mock provider and validates the content of request that must include the instance groups, parcel and repository information.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/YcloudHybridCloudTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/YcloudHybridCloudTest.java
@@ -1,0 +1,257 @@
+package com.sequenceiq.it.cloudbreak.testcase.mock;
+
+import static com.sequenceiq.it.cloudbreak.cloud.HostGroupType.IDBROKER;
+import static com.sequenceiq.it.cloudbreak.cloud.HostGroupType.MASTER;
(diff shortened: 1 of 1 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/YcloudHybridCloudTest.java
```diff
@@ -0,0 +1,252 @@
...
+        validateRepository(cmRequest.getRepository());
+        validateProducts(cmRequest.getProducts());
+
+        return dto;
+    }
+
+    private void validateInstanceGroups(List<InstanceGroupV4Request> instanceGroups) {
+        assertNotNull(instanceGroups);
+        assertEquals("The result instance group list has " + instanceGroups.size() + " items instead of 2", 2, instanceGroups.size());
+        assertTrue(instanceGroups.stream().anyMatch(ig -> ig.getName().equals(MASTER_INSTANCE_GROUP) || ig.getName().equals(IDBROKER_INSTANCE_GROUP)));
```
pkedvessy, line 236: I think you should test that both MASTER and IDBROKER instance groups are available at the same time.

PR #8771, opened 2020-08-10
CB-6362: Update FreeIPA IP addresses during repair

jamisonbennett:
During a FreeIPA repair, the FreeIPA management service now calls the
environment service to update the IP addresses for the FreeIPA
servers.

This was tested on a local cloudbreak deployment.


freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/downscale/action/FreeIpaDownscaleActions.java
```diff
@@ -349,6 +353,31 @@ protected void doExecute(StackContext context, StackEvent payload, Map<Object, O
...
+                Stack stack = context.getStack();
+                stackUpdater.updateStackStatus(stack.getId(), getInProgressStatus(variables), "Updating environment stack config");
+                try {
+                    if (!isRepair(variables)) {
+                        ThreadBasedUserCrnProvider.doAsInternalActor(() -> {
+                            environmentEndpoint.updateConfigsInEnvironmentByCrn(stack.getEnvironmentCrn());
+                        });
+                    }
+                    sendEvent(context, DOWNSCALE_UPDATE_ENVIRONMENT_STACK_CONFIG_FINISHED_EVENT.selector(), new StackEvent(stack.getId()));
+                } catch (Exception e) {
```
lacikaaa, line 373: shouldn't we log the exception here? and if it's a webapplicationexception we might need to use `WebApplicationExceptionMessageExtractor` to get the real message.
same applies to upscale

PR #8773, opened 2020-08-10
CB-8332 ensure that flow gets the right user crn in case of internal call

horadla23:


horadla23:
@lacikaaa can you review just the second commit?

horadla23:
tested locally with:
- sdx create
- sdx stop
- sdx start
- sdx master repair

horadla23:
tested with:
- env + sdx creation
- env stop
- env start
- sdx master node repair

PR #8776, opened 2020-08-11
CB-8300 and CB-8347. Add e2e diagnostics tests for FreeIpa/Sdx (+ abstrac…

oleewere:
…tions) - extended basic test cases with those

Also extended awaitFlow calls in order to make it work against freeipa + new dto
Added a distrox diagnostics dto as well, that is not used yet. (not clear how exactly use that as related tests are more like old-style stack tests not really distrox api tests, so i have to find a way to fit there)

Im using existing test cases -> it won't slow that much down the e2e test time + plan is to run diagnostics for every test case if possible (+ use recipe to gather some data from the nodes and put that as well in the diagnostics)

The PR is in draft until i did not checked against Azure

also adding a custom folder to gitignore, in order to use own/custom testsuites, as those needs to be on the classpath

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
@@ -130,8 +130,10 @@ public class AzureResourceConnector extends AbstractResourceConnector {
         } finally {
             if (!resourcesPersisted) {
                 Deployment templateDeployment = client.getTemplateDeployment(resourceGroupName, stackName);
-                LOGGER.debug("Get template deployment to persist created resources: {}", templateDeployment.exportTemplate().template());
-                persistCloudResources(ac, stack, notifier, cloudContext, stackName, resourceGroupName, templateDeployment);
+                if (templateDeployment != null && templateDeployment.exportTemplate() != null) {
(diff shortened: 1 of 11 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/sdx/SdxCloudStorageTest.java
```diff
@@ -48,6 +50,10 @@ public void testSDXWithDataLakeAndFreeIpaStorageCanBeCreatedSuccessfully(TestCon
                 .awaitForFlow(key(sdx))
                 .await(SdxClusterStatusResponse.RUNNING)
                 .awaitForInstance(getSdxInstancesHealthyState())
+                .given(diagnostics, SdxDiagnosticsTestDto.class)
```
aszegedi, line 53: I think `.withSdx(sdx)` is missing after this step.

oleewere:
> I can see `FreeIpa` and `SDX` diagnostics steps in test cases. However I cannot any `Distrox` related tests here? This is intentionally or accidentally?

it's in the PR description, that distrox tests more like stack api tests, so not really sure where it will fit at the moment. so i have created the dto for that but not used yet. could be a new test case will be needed there

aszegedi (approved):

PR #8777, opened 2020-08-11
CB-7812 decrease the error count of cert renewal poller and improve error status reasons

biharitomi:


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -265,6 +265,22 @@ public String changePrimaryGateway(Stack stack) throws CloudbreakException {
         }
     }
 
+    public void redeployGatewayCertificate(@Nonnull Stack stack, @Nonnull Cluster cluster, List<String> candidateAddresses) {
```
topolyai5, line 268: Where do you use the `candidateAddresses` param?

biharitomi: Nowhere, so removing it. Thanks! 👍

PR #8780, opened 2020-08-11
CB-8395. VmLogsService is not used for DataHub/Datalake therefore it won't col…

oleewere:
…lect the right logs during diagnostics collection


keyki:
Can you please include the jira id and extend the commit message with some description about this change?

PR #8783, opened 2020-08-11
DISTX-526 Cleanup DB ID from periscope response

smaniraju:
1.  Remove DB ID from Periscope Response which is replaced with CRN.
2. Remove unused classes from periscope api layer.


smaniraju:
Swagger-Compatibility test failure is expected since DB ID is removed from periscope response. It has been verified with api consumers CLI and UI that the removed DB ID is not used.

smaniraju:
Swagger-Compatibility test failure is expected since DB ID is removed from periscope response. It has been verified with periscope api consumers CLI and UI that the removed DB ID is not used.

PR #8787, opened 2020-08-11
CB-8355: Update FreeIPA health check agent salt scripts to work with fixed agent

wonderslug:
This updates the salt config for FreeIPA to work with the new FreeIPA health
check agent.  It copies the config scripts to the image locations, sets up
the LDAP ACIs in order to have anonymous access to the replication agreements
for the health agent, copies the certificates for use by the health agent, and
registers the script to update the certs for the health agent on cert renewal.

As well it renames the main script to cdp-freeipa-healthagent to allow for
distinction from freeipa scripts.


lacikaaa:
is this change compatible with the old agent/existing images or it depends on a new set of images?

wonderslug:
The "old" agent has not been working or running.  It was not able to run because of missing libraries.

These changes depend on a new image with an updated RPM. This image is currently being pushed through the builds.

wonderslug:
Images are now in for dev for AWS and Azure.

PR #8790, opened 2020-08-11
CB-8031: Update pillar configuration when starting cluster

jamisonbennett:
Co-authored-by: Brian Towles <btowles@cloudera.com>

When a datalake or datahub is started, update the pillar
configuration. This ensures that if the FreeIPA IP addresses changed
while the datalake was stopped then cloudbreak will update the FreeIPA
IP addresses when starting the stopped instances.

This was tested manually by ensuring the pillar configuration was
called on a local instance of cloudbreak.


biharitomi:
I checked the changes, the flow modifications look ok. But I believe this pillar refresh should happen earlier in the start flow. Because now it would refresh the pillars and configs via `HighState` after the start of CM server, agent and all the services. 
I think the services couldn't be started if a config change is necessary in the Freeipa related configs especially if the server address has been changed then they are not able to gather new auth tokens from Freeipa.
cc @lacikaaa

jamisonbennett:
I updated the start flow to update the DNS IP addresses at the first step in the flow.

jamisonbennett:
@wonderslug wanted me to check 1 more thing in regards to ensuring the node is fully online when doing the pillar update so I am marking it DO_NOT_MERGE until that is manually validated.

jamisonbennett:
`StartFlowEventChainFactory` has a chain with `STACK_START_EVENT` and `CLUSTER_START_EVENT`. The stack will always be started before the pillar update starts.

This can be committed after @lacikaaa reviews it.

PR #8792, opened 2020-08-12
CB-8341. Increase CM <-> (CP <->) Nginx proxy timeouts and make CM timeouts co…

oleewere:
…nfigurable

timeouts: CB  > (cluster proxy) >  nginx  > CM


client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/ClouderaManagerApiClientProvider.java
```diff
@@ -25,7 +26,17 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClouderaManagerApiClientProvider.class);
 
-    private static final Integer CLUSTER_PROXY_TIMEOUT = 90000;
+    @Value("${cm.client.cluster.proxy.timeout:90000}")
```
keyki, line 29: Can you please move the default values to the application.yml for the related modules?

PR #8798, opened 2020-08-12
CB-8427: Use flows to determine if FreeIPA health checker is disabled

jamisonbennett:
The StackStatusCheckerJob already verifies that the stack does not
have an active flow before it update the status. It is not necessary
to start and stop the status checker during the repair flows.

The logic was simplified to not disable or enable the status checker
during a FreeIPA repair.

This was tested manually on cloudbreak.


jamisonbennett:
I wanted to provide the except of the existing code which is relevant to this code review:
```
public class StackStatusCheckerJob extends StatusCheckerJob {
    @Override
    protected void executeInternal(JobExecutionContext context) throws JobExecutionException {
        Long stackId = getStackId();
        Stack stack = stackService.getStackById(stackId);
        prepareMdcContextWithStack(stack);
        if (flowLogService.isOtherFlowRunning(stackId)) {
            LOGGER.debug("StackStatusCheckerJob cannot run, because flow is running for stack: {}", stackId);
            return;
        }
        syncAStack(stack);
        MDCBuilder.cleanupMdc();
    }
}
```

PR #8799, opened 2020-08-12
CB-7975 - [ASRG] Save and manage managed images and storage accounts …

pdarvasi:
…if CB created RG

Currently, we do not delete the storage account and managed image components in single RG use-cases. 

This PR does this:

- save storage accounts and managed images as Resources, too
- if they got created via an SDX or FreeIPA creation then they get deleted by deleting them
- otherwise, we do not touch them

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureStorage.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureStorage.java
@@ -36,6 +36,7 @@ import com.sequenceiq.cloudbreak.cloud.azure.view.AzureCredentialView;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
 import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
 import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 
 @Service
 public class AzureStorage {
@@ -75,23 +76,24 @@ public class AzureStorage {
         return ArmAttachedStorageOption.valueOf(attachedStorageOption);
     }
 
-    public String getCustomImageId(AzureClient client, AuthenticatedContext ac, CloudStack stack) {
-        return getCustomImageId(client, ac, stack, stack.getImage().getImageName());
+    public AzureImage getCustomImage(AzureClient client, AuthenticatedContext ac, CloudStack stack) {
+        return getCustomImage(client, ac, stack, stack.getImage().getImageName());
     }
 
-    public String getCustomImageId(AzureClient client, AuthenticatedContext ac, CloudStack stack, String imageName) {
+    public AzureImage getCustomImage(AzureClient client, AuthenticatedContext ac, CloudStack stack, String imageName) {
         String imageResourceGroupName = azureResourceGroupMetadataProvider.getImageResourceGroupName(ac.getCloudContext(), stack);
         AzureCredentialView acv = new AzureCredentialView(ac.getCloudCredential());
         String imageStorageName = getImageStorageName(acv, ac.getCloudContext(), stack);
         String imageBlobUri = client.getImageBlobUri(imageResourceGroupName, imageStorageName, IMAGES_CONTAINER, imageName);
         String region = ac.getCloudContext().getLocation().getRegion().value();
-        return getCustomImageId(imageBlobUri, imageResourceGroupName, region, client);
+        return getCustomImage(imageBlobUri, imageResourceGroupName, region, client);
     }
 
-    private String getCustomImageId(String vhd, String imageResourceGroupName, String region, AzureClient client) {
-        String customImageId = client.getCustomImageId(imageResourceGroupName, vhd, region);
+    private AzureImage getCustomImage(String vhd, String imageResourceGroupName, String region, AzureClient client) {
+        AzureImage image = client.getCustomImageId(imageResourceGroupName, vhd, region);
+        String customImageId = image.getId();
         LOGGER.debug("Custom image id: {}", customImageId);
-        return customImageId;
+        return image;
     }
 
     public String getImageStorageName(AzureCredentialView acv, CloudContext cloudContext, CloudStack cloudStack) {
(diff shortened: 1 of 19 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureStorageAccountService.java
```diff
@@ -40,13 +45,22 @@
...
         StorageAccount storageAccount = client.getStorageAccountByGroup(resourceGroup, storageName);
         if (storageAccount == null) {
             try {
                 LOGGER.info("Creating storage account: {}", storageName);
-                armStorage.createStorage(client, storageName, LOCALLY_REDUNDANT, resourceGroup, region, isEncryptionNeeded(stack), stack.getTags());
+                StorageAccount storage = armStorage.createStorage(client, storageName, LOCALLY_REDUNDANT, resourceGroup, region, isEncryptionNeeded(stack),
+                        stack.getTags());
                 pollStorageAccountCreation(ac, new StorageAccountCheckerContext(client, resourceGroup, storageName));
+                CloudResource cloudResource = azureCloudResourceService.buildCloudResource(storage.name(), storage.id(), ResourceType.AZURE_STORAGE);
+                azureCloudResourceService.saveCloudResources(resourceNotifier, ac.getCloudContext(), List.of(cloudResource));
```
gergopapi2, line 63: I would put this into a finally clause. Since the storage account has more than 1 resource it could happen that the storage account is created but for some reason the blobService not, and an exception is thrown.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
```diff
@@ -95,18 +95,31 @@
...
+
+            // deleting images
+            List<String> imageIds = getResourceIdsByResourceType(resourcesToRemove, ResourceType.AZURE_MANAGED_IMAGE);
+            azureUtils.deleteImages(client, imageIds);
+            deleteCloudResourceList(ac, resourcesToRemove, ResourceType.AZURE_MANAGED_IMAGE);
+
+            // deleting storage account
+            List<String> accountIds = getResourceIdsByResourceType(resourcesToRemove, ResourceType.AZURE_STORAGE);
+            azureUtils.deleteStorageAccounts(client, accountIds);
+            deleteCloudResourceList(ac, resourcesToRemove, ResourceType.AZURE_STORAGE);
```
gergopapi2, line 122: I am not convinced this is the right place. What if a datahub is being terminated, will it delete the images and the storage account? What about concurrent termination of cluster A and provisioning of cluster B? I might be missing something, sorry for that.

gergopapi2: We had a conversation with @pdarvasi. This would cause a performance hit in case a customer is creating a large number of short-lived datahubs: the image would be copied frequently, causing 10-20 minutes of extra datahub provisioning time. If this potential performance hit is not a problem then this PR can be merged.

gergopapi2: As clarified, it will not cause any perf hit.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureStorageAccountService.java
pdarvasi: You are right, but we do not know the storage acc id in the "finally" clause, we could go and fetch the deployment from Azure there but it is cumbersome as we don't have the deployment's id.
in such cases, the whole deployment will fail, the user will be notified about the failing storage account and at most a storage account will not be deleted, which is free. As this is a temp solution, I would not make a major effort to fix this.

pdarvasi:
We have another conversation with @gergopapi2 and clarified that this PR affects the images and not the VHD-s underhood so the performance degradation should not be an issue.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureStorage.java
```diff
@@ -108,14 +110,15 @@ public String getAttachedDiskStorageName(ArmAttachedStorageOption armAttachedSto
...
+    public StorageAccount createStorage(AzureClient client, String osStorageName, AzureDiskType storageType, String storageGroup,
+            String region, Boolean encrypted, Map<String, String> tags)
             throws CloudException {
         if (!storageAccountExist(client, osStorageName)) {
             StorageAccountParameters storageAccountParameters = new StorageAccountParameters(
                     storageGroup, osStorageName, region, skuTypeResolver.resolveFromAzureDiskType(storageType), encrypted, tags);
-            azureStorageAccountBuilderService.buildStorageAccount(client, storageAccountParameters);
+            return azureStorageAccountBuilderService.buildStorageAccount(client, storageAccountParameters);
         }
+        throw new CloudbreakServiceException(String.format("Trying to delete non-existing storage account %s", osStorageName));
```
gergopapi2, line 121: This is create, not delete - can go to follow-up.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
```diff
@@ -95,18 +95,31 @@
...
             List<String> networkIds = getResourceIdsByResourceType(resourcesToRemove, ResourceType.AZURE_NETWORK);
             azureUtils.deleteNetworks(client, networkIds);
             deleteCloudResourceList(ac, resourcesToRemove, ResourceType.AZURE_NETWORK);
             deleteCloudResourceList(ac, resourcesToRemove, ResourceType.AZURE_SUBNET);
 
+            // deleting security groups
             List<String> securityGroupIds = getResourceIdsByResourceType(resourcesToRemove, ResourceType.AZURE_SECURITY_GROUP);
             azureUtils.deleteSecurityGroups(client, securityGroupIds);
             deleteCloudResourceList(ac, resourcesToRemove, ResourceType.AZURE_SECURITY_GROUP);
+
```
gergopapi2, line 113: I would put sections labelled with comments into separate methods with descriptive names. Can go into follow-up.

PR #8801, opened 2020-08-12
DISTX-530 DE HA template does not work with telemetry publisher

cegganesh84:
1. Telemetry publisher needs to have HDFS redaction enabled.
2. With these settings being false, the customer sees that the telemetry publisher does not start.
3. Since normal templates are working fine, removing the config to leave it to default.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-ha-702.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-ha-702.bp
@@ -20,10 +20,6 @@
         "serviceType": "HDFS",
         "serviceConfigs": [
           {
-            "name": "redaction_policy_enabled",
-            "value": "false"
-          },
-          {
             "name": "zookeeper_service",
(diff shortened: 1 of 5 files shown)


cegganesh84:
> Does this essentially means that telemetry publisher will now start and be able to push results to WXM (which already happens for the non-HA templates?)?

Yes.

Also, looks like a limited cluster admin can not enable or disable this setting. Customer does not really have a workaround. So thinking of putting this in 2.27 branch, any objections.

keyki:
On the UI there is an action to enable workload analytics. Is it what we're looking for?

cegganesh84:
@keyki Yes, that button will not work for DE HA template because the HDFS redaction is disabled. The customer will get a warning/error in CM like this.
image (13)

For now, I have asked the customer to go through what other customers went through (admin role and all that) to change CM settings. I do not have a response yet. Probably we can put it in master and only go through the hotfix process if it becomes an immediate issue.

PR #8802, opened 2020-08-13
ENGESC-3627 Disable thrash related warnings in DE clusters.

cegganesh84:
1. Trash interval was set to zero, and skip trash was set to false for performance reasons consciously CB-7053.
2. This triggered a couple of warnings in CM.
3. Suppressing these because the customer can not do anything about it, and in-fact wondering what those were.
4. The change here was convoluted so will explain what is going on,
   a. I did not copy the trash interval settings in the HA template in 702. That is why it is missing some configs in some templates.
   b. So the suppressions were added to the config only if there would have been a warning otherwise.
5. These settings were manually modified in the CM and then the template was exported so that we know what setting to add and at which level.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-702.bp
@@ -25,6 +25,10 @@
             "base": true,
 	    "configs": [
               {
+                "name": "role_config_suppression_fs_trash_interval_minimum_validator",
+                "value": "true"
+              },
+              {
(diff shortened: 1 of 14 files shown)


amitagarwal06:
@cegganesh84 Change looks good. I have only one suggestion for future template changes that its better to set boolean configs as boolean true rather than as a string value "true".

Just FYI, another way to find out what to change in template is turn on browser developer tools while manually modifying the config in CM. It clearly shows CM's internal name for the config and under which role & roleConfigGroup to add that.

PR #8803, opened 2020-08-13
CB-8435 StackStatusCheckerJob should always cleanup MDC

lacikaaa:
Currently we set the MDCContext before checking if a flow is running. If a flow is running we skip sync, but the code will skip MDCContext cleanup also.
This commit modifies the code to ensure cleanup is happening.


jamisonbennett:
@lacikaaa It appears this method can throw an exception, should ` MDCBuilder.cleanupMdc()` be called in a finally block?

lacikaaa:
which one?

jamisonbennett:
The status checker catches and swallows all exceptions during a status check, but I wonder if a runtime exception could be thrown from `flowLogService.isOtherFlowRunning(stackId)` when it checks the database via the repository.

lacikaaa:
I think only if the DB is not accessible. but if you think we should be more cautious I can put it into a finally block.

jamisonbennett:
I had to fix a few other related logging items. So I addressed this in, https://github.com/hortonworks/cloudbreak/pull/8811.

lacikaaa:
thanks

PR #8806, opened 2020-08-13
ENGESC-3681 Cluster expansion hung after hitting EC2 limit. An attemp…

topolyai5:
…t to increase the size of a data hub cluster failed due to an AWS quota limit, but the operation in CB isn't timing out or canceling. In this PR, I'm decreasing the timeout from 13,5H to 0,5H

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
@@ -7,6 +7,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.inject.Inject;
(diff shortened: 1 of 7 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
```diff
@@ -143,6 +149,27 @@ private void waitForGroup(AmazonEC2Client amClient, AmazonAutoScalingClient asCl
...
+            Date timeBeforeASUpdate) throws AmazonAutoscalingFailed {
+        try {
+            Waiter<DescribeScalingActivitiesRequest> autoscalingActivitesWaiter = customAmazonWaiterProvider
+                    .getAutoscalingActivitesWaiter(asClient, timeBeforeASUpdate);
+            autoscalingActivitesWaiter.run(new WaiterParameters<>(new DescribeScalingActivitiesRequest().withAutoScalingGroupName(autoScalingGroupName)));
+        } catch (Exception e) {
+            LOGGER.error("Failed to list activities: {}", e.getMessage(), e);
+            throw new AmazonAutoscalingFailed(e.getMessage(), e);
+        }
+        DescribeScalingActivitiesResult describeScalingActivitiesResult = asRetryClient
```
biharitomi, line 162: Wouldn't it be better to have a solid try-catch on this provider call?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsAutoScalingService.java
```diff
@@ -143,6 +149,27 @@ private void waitForGroup(AmazonEC2Client amClient, AmazonAutoScalingClient asCl
...
+                .describeScalingActivities(new DescribeScalingActivitiesRequest().withAutoScalingGroupName(autoScalingGroupName));
+        Optional<Activity> firstActivity = describeScalingActivitiesResult.getActivities().stream().findFirst()
```
biharitomi, line 164: Earlier we had a similar logic and support cases where it turned out that looking for a failed activity is not a good approach. As activities could have gone wrong in the past for a scaling activity which is not related to the actual/latest one. I think we need to check the state of the latest activity only.

PR #8807, opened 2020-08-13
CDPCP-2673. Fix role refresh on cloud identity sync

handavid:
The previous behavior for syncing azure oids into CM was to only
trigger role refresh if CM reported the configs to be stale. This
does not always work as expected because the staleness status that
CM reports is eventually consistent.

To fix this, instead of relying on staleness check we only set
the configs if they are different than the existing configs. Under
rare circumstances we still need to fallback on checking staleness
status (as explained in the code comment).

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
@@ -26,6 +26,8 @@ import org.springframework.stereotype.Component;
 
 import javax.inject.Inject;
 import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -77,7 +79,7 @@ public class ClouderaManagerRangerUtil {
     }
 
     private ApiCommand triggerRoleRefresh(ApiClient client, String clusterName, String serviceName, String roleName) throws ApiException {
-        LOGGER.info("Attempting to trigger role refresh on clusterName = {}, serviceName = {}, roleName = {}", clusterName, serviceName, roleName);
+        LOGGER.info("Trigerring role refresh on clusterName = {}, serviceName = {}, roleName = {}", clusterName, serviceName, roleName);
         ApiRoleNameList roleNameList = new ApiRoleNameList();
         roleNameList.addItemsItem(roleName);
         RoleCommandsResourceApi roleCommandsResourceApi = clouderaManagerApiFactory.getRoleCommandsResourceApi(client);
@@ -90,23 +92,11 @@ public class ClouderaManagerRangerUtil {
         return configList.getItems().stream().map(ApiConfig::getName).anyMatch(configName -> configName.equals(AZURE_USER_MAPPING));
     }
 
-    private boolean isRoleRefreshNeeded(RolesResourceApi rolesResourceApi, String clusterName, String rangerUserSyncRole) throws ApiException {
+    private boolean isRoleStale(RolesResourceApi rolesResourceApi, String clusterName, String rangerUserSyncRole) throws ApiException {
         ApiRole role = rolesResourceApi.readRole(clusterName, rangerUserSyncRole, RANGER_SERVICE_NAME, "summary");
         ApiConfigStalenessStatus stalenessStatus = role.getConfigStalenessStatus();
         LOGGER.debug("Ranger user sync ApiConfigStalenessStatus = {}", stalenessStatus);
         return stalenessStatus.equals(ApiConfigStalenessStatus.STALE_REFRESHABLE);
-    }
-
-    private Optional<ApiCommand> refreshRoleIfStale(ApiClient client, RolesResourceApi rolesResourceApi, String clusterName, String rangerUserSyncRoleName)
-            throws ApiException {
-        if (isRoleRefreshNeeded(rolesResourceApi, clusterName, rangerUserSyncRoleName)) {
-            LOGGER.info("Role refresh required, trigerring role refresh");
-            ApiCommand command = triggerRoleRefresh(client, clusterName, RANGER_SERVICE_NAME, rangerUserSyncRoleName);
-            return Optional.of(command);
-        } else {
-            LOGGER.info("No role refresh required");
-            return Optional.empty();
-        }
     }
 
     public boolean isCloudIdMappingSupported(String stackCrn) throws ApiException {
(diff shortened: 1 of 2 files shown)


datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -118,17 +109,61 @@ public boolean isCloudIdMappingSupported(String stackCrn) throws ApiException {
         return isCloudIdMappingSupported(rolesResourceApi, clusterName, rangerUserSyncRoleName);
     }
 
-    public Optional<ApiCommand> setAzureCloudIdentityMapping(String stackCrn, Map<String, String> azureUserMapping) throws ApiException {
+    private Map<String, String> userMappingStrToMap(String userMappingStr) {
+        userMappingStr = userMappingStr.trim();
+        if (userMappingStr.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        Map<String, String> ret = new HashMap<>();
```
lacikaaa, line 117: could you give this variable a meaningful name?

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -118,17 +109,61 @@ public boolean isCloudIdMappingSupported(String stackCrn) throws ApiException {
         return isCloudIdMappingSupported(rolesResourceApi, clusterName, rangerUserSyncRoleName);
     }
 
-    public Optional<ApiCommand> setAzureCloudIdentityMapping(String stackCrn, Map<String, String> azureUserMapping) throws ApiException {
+    private Map<String, String> userMappingStrToMap(String userMappingStr) {
+        userMappingStr = userMappingStr.trim();
+        if (userMappingStr.isEmpty()) {
+            return Collections.emptyMap();
```
lacikaaa, line 115: I don't prefer returns in the middle of a method. Could you add an else branch for the rest f the code, so it's obvious there is 2 exit point?

lacikaaa: shall we log here it's empty?

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -118,17 +109,61 @@ public boolean isCloudIdMappingSupported(String stackCrn) throws ApiException {
...
+        }
+        Map<String, String> ret = new HashMap<>();
+        String[] mappings = userMappingStr.split(";");
+        Arrays.stream(mappings).forEach(mapping -> {
+            String[] entry = mapping.split("=");
```
lacikaaa, line 120: what happens if there is no `=` sign? I think it would throw an exception when you try to assign value to `val`. Shouldn't we handle this somehow?

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -118,17 +109,61 @@ public boolean isCloudIdMappingSupported(String stackCrn) throws ApiException {
...
+        return ret;
+    }
+
+    private Map<String, String> getExistingAzureUserMapping(RolesResourceApi rolesResourceApi, String clusterName, String rangerUserSyncRoleName)
+            throws ApiException {
+        ApiConfigList roleConfigList = rolesResourceApi.readRoleConfig(clusterName, rangerUserSyncRoleName, RANGER_SERVICE_NAME, "summary");
+        ApiConfig azureUserMappingConfig = roleConfigList.getItems().stream()
+                .filter(apiConfig -> apiConfig.getName().equals(AZURE_USER_MAPPING))
+                .findFirst()
+                .get();
```
lacikaaa, line 134: what if none found? are we sure there is at least one?

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
gitmfox: Renamed.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
gitmfox: Fixed first issue, added log statement as suggested.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
gitmfox: Right. Now handles the lack of an equals sign by logging a warning and skipping the entry.

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
gitmfox: I am not sure if we can be sure that the config is always present--probably not, but in any case we should protect against this possibility. Now checks that the optional value is present before getting it. If it is absent, an empty map is returned.

gitmfox:
run tests

lacikaaa:
could you squash your commit? otherwise LGTM

gitmfox:
Squashed. @lacikaaa, I see that 2.28 is branching today; is it possible to get this commit into 2.28? We are trying to release this feature in early September. Thanks.

lacikaaa:
it's already branch, could you change the base to CB-2.28.0 in that case?

gitmfox:
@lacikaaa I'm happy to do that, but I'm not entirely sure I know how. Would I close this PR,  then create a new PR with the compare branch set to CB-2.28.0?

gitmfox:
@lacikaaa , I've rebased onto CB-2.28.0

lacikaaa:
if it's green you can merge it

gitmfox:
@lacikaaa , I don't have permissions to merge in this repo.

gitmfox:
@lacikaaa , @keyki , I'm not sure what I need to do to complete this PR. The build failures don't seem to have anything to do with these specific changes.

pdarvasi:
@gitmfox I've re-triggered the failing tests, swagger test had issues recently, but the IT should be green to be able to merge

pdarvasi:
I am merging as the IT is now green.

gitmfox:
@pdarvasi, thanks! Does this change need to be merged into master as well?

PR #8811, opened 2020-08-14
CB-8401: Add the operation ID to the FreeIPA logs

jamisonbennett:
Add the operation ID to the FreeIPA logs for repair.

Change the user sync to use the operationId for the operation ID
rather than the flowId.

Ensure the MDC cleanup happens if an exception is thrown.

### DIFF
--- a/freeipa/src/main/resources/logback.xml
+++ b/freeipa/src/main/resources/logback.xml
@@ -12,7 +12,7 @@
         <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
             <layout class="com.sequenceiq.cloudbreak.logger.MaskingPatternLayout">
                 <loggerNameFilter>com.sequenceiq</loggerNameFilter>
-                <pattern>%date{ISO8601} [%thread] %method:%line %-5level %logger{5} - [type:%X{resourceType:-springLog}] [crn:%X{resourceCrn:-}] [name:%X{resourceName:-}] [flow:%X{flowId:-}] [requestid:%X{requestId:-}] [tenant:%X{tenant:-}] [userCrn:%X{userCrn:-}] [environment:%X{environmentCrn:-}] [traceId:%X{traceId:-}] [spanId:%X{spanId:-}] %msg%n</pattern>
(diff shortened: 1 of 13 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/operation/OperationService.java
```diff
@@ -99,6 +102,14 @@ public Operation getOperationForAccountIdAndOperationId(String accountId, String
         return operationOptional.get();
     }
 
+    public static void addMdcOperationId(String operationId) {
+        MDCBuilder.addMdcField(OPERATION_ID_MDC_NAME, operationId);
+    }
+
+    public static void removeMdcOperationId() {
+        MDCBuilder.removeMdcField(OPERATION_ID_MDC_NAME);
+    }
```
lacikaaa, line 111: I think you should add these methods to `MDCBuilder` as we have it like `addEnvCrn`

freeipa/src/main/resources/logback.xml
```diff
@@ -12,7 +12,7 @@
         <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
             <layout class="com.sequenceiq.cloudbreak.logger.MaskingPatternLayout">
                 <loggerNameFilter>com.sequenceiq</loggerNameFilter>
-                <pattern>%date{ISO8601} [%thread] %method:%line %-5level %logger{5} - [type:%X{resourceType:-springLog}] [crn:%X{resourceCrn:-}] [name:%X{resourceName:-}] [flow:%X{flowId:-}] [requestid:%X{requestId:-}] [tenant:%X{tenant:-}] [userCrn:%X{userCrn:-}] [environment:%X{environmentCrn:-}] [traceId:%X{traceId:-}] [spanId:%X{spanId:-}] %msg%n</pattern>
+                <pattern>%date{ISO8601} [%thread] %method:%line %-5level %logger{5} - [type:%X{resourceType:-springLog}] [crn:%X{resourceCrn:-}] [name:%X{resourceName:-}] [flow:%X{flowId:-}] [requestid:%X{requestId:-}] [tenant:%X{tenant:-}] [userCrn:%X{userCrn:-}] [environment:%X{environmentCrn:-}] [traceId:%X{traceId:-}] [spanId:%X{spanId:-}] [operationId:%X{operationId:-}] %msg%n</pattern>
```
lacikaaa, line 15: @oleewere does this change break kibana log parsing?

oleewere: @lacikaaa nope, we are using JSON formatter on mow envs

PR #8815, opened 2020-08-14
CB-7520 platformresources should work with audit credential as well

doktoric:

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/platformresource/PlatformParameterService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/platformresource/PlatformParameterService.java
@@ -1,10 +1,12 @@
 package com.sequenceiq.environment.platformresource;
 
 import static com.sequenceiq.cloudbreak.cloud.service.CloudParameterService.ACCESS_CONFIG_TYPE;
+import static com.sequenceiq.common.model.CredentialType.AUDIT;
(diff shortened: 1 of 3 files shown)


environment/src/main/java/com/sequenceiq/environment/platformresource/PlatformParameterService.java
```diff
@@ -152,9 +154,17 @@ public PlatformResourceRequest getPlatformResourceRequest(
         PlatformResourceRequest platformResourceRequest = new PlatformResourceRequest();
 
         if (!Strings.isNullOrEmpty(credentialName)) {
-            platformResourceRequest.setCredential(credentialService.getByNameForAccountId(credentialName, accountId, ENVIRONMENT));
+            Credential credential = credentialService.getByNameForAccountId(credentialName, accountId, ENVIRONMENT);
```
topolyai5, line 157: Can you do in the same request? filter for all type or something

topolyai5: I mean, in the same sql query

doktoric: when we implemented this feature we decided to follow the approach that only one type can passed to the repository

PR #8816, opened 2020-08-14
CB-8460 Duplicate hostnames for datahub cluster after series of downscale, force remove and upscale which prevents future downscale

foldik:


foldik:
Tested upscale, downscale multiple times.

PR #8820, opened 2020-08-14
CB-7876 Fail API Cleanup Jenkins build in case of left resources

aszegedi:
This is the final improvement of the API Cleanup App:

- Provided resource JSON files are driving the cleanup instead of the found resources in the given tenant.
- If JSON file(s) is(are) present at `integrationtest.outputdir` path, listed resources are going to be checked. Otherwise all the found resources are going to be deleted/terminated as fallback.
- If a test generates more than one resource per type, that will generate an Array instead of Object:
  ```
  {
    "sdxName": "yarn-test-8206e51801004830a4a90e730a1ffc",
    "blueprintName": "yarn-test-94f1651394dd491a9a03a6714e66e9",
    "environmentName": [
      "yarn-test-18f9dd8b60ed492388",
      "aws-test-6f1adc4012e94028987"
    ],
    "stackName": [
      "yarn-test-af9ea",
      "aws-test-902939"
    ],
    "credentialName": [
      "aws-test-1b8a1afafded4daca3e1cc5e9533f37",
      "yarn-test-54daadca706347818431edc9c6b560"
    ]
  }
  ```
- If other resource(s) is(are) present at the given tenant, that won't be touched by the Cleanup App except resource JSON file cannot be found.
- If the Cleanup finds one or more credential(s) present from the resource JSON at the given tenant, Cleanup exits with error:
`ERROR c.s.it.util.cleanup.CleanupUtil [nolabel] - End To End cleanup have been failed, because of credential(s) '[aws-test-1b8a1afafded4daca3e1cc5e9533f37, yarn-test-54daadca706347818431edc9c6b560]' found left behind!` after all the found resources have been deleted/terminated.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
@@ -1,16 +1,29 @@
 package com.sequenceiq.it.util.cleanup;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
 
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackViewV4Response;
@@ -18,6 +31,7 @@ import com.sequenceiq.cloudbreak.client.CloudbreakClient;
 import com.sequenceiq.environment.api.v1.credential.model.CredentialBase;
 import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentBaseResponse;
 import com.sequenceiq.environment.client.EnvironmentClient;
+import com.sequenceiq.it.cloudbreak.MicroserviceClient;
 import com.sequenceiq.it.cloudbreak.util.WaitResult;
 import com.sequenceiq.sdx.api.model.SdxClusterResponse;
 import com.sequenceiq.sdx.client.SdxClient;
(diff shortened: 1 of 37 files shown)


integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
```diff
@@ -172,21 +191,213 @@ public void cleanupCredentials() {
...
     }
 
-    private void deleteEnvironments(Set<String> environmentNames, EnvironmentClient environmentClient) {
-        environmentNames.forEach(environmentName -> LOG.info("Environment with name: {} will be deleted!", environmentName));
+    private List<String> getResourcesFromFile(String resourceNameType, Path filePath) {
+        List<String> resourceNames = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(filePath.toString()))) {
+            StringBuilder builder = new StringBuilder();
+            String line;
+            while ((line = reader.readLine()) != null) {
```
foldik, line 199: Please use this standard library method for file reading:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#readString(java.nio.file.Path)

integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
```diff
@@ -172,21 +191,213 @@ public void cleanupCredentials() {
...
+            }
+        } catch (Exception e) {
+            LOG.error("Cannot find resource file at path: '{}', because of: {}", Paths.get(outputDirectory).toAbsolutePath().normalize(), e.getMessage(), e);
+            throw new RuntimeException(String.format("Cannot find resource file at path: '%s', because of: %s",
+                    Paths.get(outputDirectory).toAbsolutePath().normalize(), e.getMessage()));
+        }
+        fileList.forEach(filePath -> {
+            LOG.info("Processing resource file: '{}'", filePath.getFileName());
+            List<String> resourcesName = Optional.ofNullable(getResourcesFromFile(resourceNameType, filePath))
+                    .orElse(List.of(""));
```
foldik, line 253: Why not an empty list with `List.of()`?

integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
```diff
@@ -172,21 +191,213 @@ public void cleanupCredentials() {
...
+            resourcesName.forEach(resourceName -> {
+                if (foundResources.stream()
```
foldik, line 255: Cool, functional, but I guess `foundResource.contains(resourceName)` will be enough instead of `foundResources.stream().anyMatch(resourceName::equals)`.

integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
```diff
@@ -172,21 +191,213 @@ public void cleanupCredentials() {
...
+                        case "distroxName":
+                        case "stackName":
+                            deleteDistrox(getCloudbreakClient(), resourceName);
+                            break;
+                        case "sdxName":
+                            deleteSdx(getSdxClient(), resourceName);
+                            break;
+                        case "credentialName":
+                            deleteCredential(getEnvironmentClient(), resourceName);
+                            e2eCleanupFailed.set(true);
```
foldik, line 269: This is strange. You try to delete credentials, then you immediately say it is failed. Where does this information comes from?

lnardai (approved):
I've only checked the TestContext -> MockedTestContext conversion.
Looks like a nice cosmetic change, then I started wondering how the MeasuredTestContext comes into place.

foldik (changes requested):

integration-test/src/main/java/com/sequenceiq/it/util/cleanup/CleanupUtil.java
aszegedi: Answer in the description:
>If the Cleanup finds one or more credential(s) present from the resource JSON at the given tenant, Cleanup exits with error:
ERROR c.s.it.util.cleanup.CleanupUtil [nolabel] - End To End cleanup have been failed, because of credential(s) '[aws-test-1b8a1afafded4daca3e1cc5e9533f37, yarn-test-54daadca706347818431edc9c6b560]' found left behind! after all the found resources have been deleted/terminated.

foldik: I see. What about other resources? Don't we expect them to be deleted as well?

aszegedi: Credential(s) is(are) the final in the sequence of deleting resources. So we can rest assured if we need to delete any credential from the created ones, that means the related E2E job cleanup was not successful.

We can sure the final deletable resource from the list is the credential. If we need to delete any listed credential from the resources Json file, we need to investigate the related E2E build: 
- What was happened at the Cleanup phase? Is this started at the end of the test execution?
- Why the E2E cleanup cannot fulfil its goal?
In this situation the amount and type (distrox, sdx, environment, credential) of the remained resources are not important for the Cleanup App. The fact: we found deletable resource(s) is the vital information and need to highlight for reviewers.

foldik (approved):

PR #8823, opened 2020-08-14
CB-8478: Fix freeipa_healthagent_setup.sh to work with FreeIPA HA

jamisonbennett:
Fix the freeipa_healthagent_setup.sh so that it works with multiple
instances of FreeIPA.

This was tested with a local deployment of cloudbreak by deploying a
2 node FreeIPA cluster.


jamisonbennett:
NOTE CB-8479 is not addressed yet so this fix will not without it.

lacikaaa:
int tests are failing because 7.2.1 version is missing from rc image catalog for 2.28
swagger is failing for every 2.28 pr, so it's not related and you haven't done any api change so I'm merging this

PR #8834, opened 2020-08-18
CB-8480 internal calls broke tags and audit

sodre90:

### DIFF
--- a/authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
+++ b/authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
@@ -25,8 +25,13 @@ import com.sequenceiq.authorization.annotation.FilterListBasedOnPermissions;
 import com.sequenceiq.authorization.annotation.InternalOnly;
 import com.sequenceiq.authorization.service.list.ListPermissionChecker;
 import com.sequenceiq.authorization.util.AuthorizationAnnotationUtils;
-import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.auth.InternalCrnBuilder;
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.security.CrnUserDetailsService;
+import com.sequenceiq.cloudbreak.auth.security.internal.InitiatorUserCrn;
+import com.sequenceiq.cloudbreak.auth.security.internal.InternalUserModifier;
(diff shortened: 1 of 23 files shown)


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
```diff
@@ -285,6 +351,13 @@ UpgradeV4Response checkForClusterUpgradeByName(@PathParam("workspaceId") Long wo
     FlowIdentifier upgradeClusterByName(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name, String imageId,
             @AccountId @QueryParam("accountId") String accountId);
 
+    @POST
+    @Path("internal/{name}/cluster_upgrade")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = STACK_UPGRADE_INTERNAL, nickname = "upgradeClusterByNameInternal")
+    FlowIdentifier upgradeClusterByNameInternal(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name, String imageId,
```
lnardai, line 358: Can't we skip the whole workspace ID from these new API's?

lnardai: Or use accountId instead?

authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
```diff
@@ -54,6 +68,12 @@ public Object hasPermission(ProceedingJoinPoint proceedingJoinPoint) {
         MethodSignature methodSignature = (MethodSignature) proceedingJoinPoint.getSignature();
         LOGGER.debug("Permission check started at {} (method: {})", startTime,
                 methodSignature.getMethod().getDeclaringClass().getSimpleName() + "#" + methodSignature.getMethod().getName());
+        Optional<Object> initiatorUserCrn = reflectionUtil.getParameter(proceedingJoinPoint, methodSignature, InitiatorUserCrn.class);
```
lnardai, line 71: Can we check if the endpoint has the @InternalOnly annotation?

lnardai (approved):

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
sodre90: I'd like to, but we have some precheck code for this and fails if there is no workspaceId. We can create a followup task for this but I didn't want to deal with this in this PR.

PR #8844, opened 2020-08-24
CB-8505-AssociatePeriscopeRepositoryWithTransaction

smaniraju:
All existing Periscope Repository interfaces should be annotated with Transactional attribute similar to CB Repository interfaces, since without that Query Results will not be pre-flushed and stale data will be returned in certain scenarios.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/repository/ClusterPertainRepository.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/repository/ClusterPertainRepository.java
@@ -1,14 +1,15 @@
 package com.sequenceiq.periscope.repository;
 
-import java.util.Optional;
-
+import com.sequenceiq.cloudbreak.workspace.repository.EntityType;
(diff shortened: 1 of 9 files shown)


smaniraju:
> Does this end up making every DB operation a transaction?
> Is there an alternate way to prevent caching?

By default all DB Operations including read are transactional at spring jpa level. But when we extend CrudRepository to write custom query api, the Transaction annotation is not inherited and has to be explicitly defined since annotation are not inherited for interfaces. If we use Spring "@Transactional" there is some optimization where in "readOnly=true" can be specified for read only queries, but this differs from CB Repository Classes which uses Javax "@Transactional"  instead of Spring "@Transacational" and I didnt want to change this mechanism only for periscope repositories.

smaniraju:
Swagger compatibility checks failure ignored because of CDPCP-2673 based on comments in eng_cb_dev_internal slack channel.

PR #8848, opened 2020-08-24
CB-8517 added crn field to the cluster definition response because it…

doktoric:
… was not part of the response before. Also added the unit test just to be consistent


agov (approved):
LGTM, tested locally

PR #8851, opened 2020-08-25
CDPCP-2822. Ensure datalake is running before triggerring OID sync

aarman-cloudera:
This ensures that we only make the CM api calls to set cloud
identity configs on CM's that are running.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
@@ -3,8 +3,11 @@ package com.sequenceiq.datalake.cm;
 import com.cloudera.api.swagger.client.ApiException;
 import com.cloudera.api.swagger.model.ApiCommand;
 import com.google.common.collect.Iterables;
+import com.sequenceiq.datalake.entity.DatalakeStatusEnum;
 import com.sequenceiq.datalake.entity.SdxCluster;
+import com.sequenceiq.datalake.entity.SdxStatusEntity;
 import com.sequenceiq.datalake.service.sdx.SdxService;
+import com.sequenceiq.datalake.service.sdx.status.SdxStatusService;
 import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncState;
 import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncStatus;
 import org.slf4j.Logger;
@@ -26,6 +29,9 @@ public class RangerCloudIdentityService {
 
     @Inject
     private SdxService sdxService;
+
+    @Inject
+    private SdxStatusService sdxStatusService;
 
     private RangerCloudIdentitySyncStatus newSyncStatus(RangerCloudIdentitySyncState state, String message) {
         RangerCloudIdentitySyncStatus status = new RangerCloudIdentitySyncStatus();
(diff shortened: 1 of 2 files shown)


datalake/src/main/java/com/sequenceiq/datalake/cm/RangerCloudIdentityService.java
```diff
@@ -64,11 +70,20 @@ private RangerCloudIdentitySyncStatus setAzureCloudIdentityMapping(String envCrn
...
+    }
+
     public RangerCloudIdentitySyncStatus setAzureCloudIdentityMapping(String envCrn, Map<String, String> azureUserMapping) {
         Optional<SdxCluster> sdxCluster = getSdxCluster(envCrn);
         if (sdxCluster.isEmpty()) {
             return newSyncStatus(RangerCloudIdentitySyncState.NOT_APPLICABLE, "No datalakes associated with the environment.");
         }
+        if (!isDatalakeRunning(sdxCluster.get())) {
+            return newSyncStatus(RangerCloudIdentitySyncState.NOT_APPLICABLE, "Datalake is not running for the environment.");
+        }
```
lacikaaa, line 86: this is a duplicate of the `getRangerCloudIdentitySyncStatus`. could you refactor this to a common part to reduce code duplication?

aarman-cloudera: Thanks, I've added the `else if` branches as per your chat suggestion. I think making it more DRY would require wrapping the operation in a functional interface (see below example) which seemed less readable to me, so I didn't go for it. Let me know if that's okay, or if you had some other idea.


```java
public RangerCloudIdentitySyncStatus runCloudIdSyncOperation(String envCrn, Function<SdxCluster, RangerCloudIdentitySyncStatus> operation) {
    Optional<SdxCluster> sdxCluster = getSdxCluster(envCrn);
    if (sdxCluster.isEmpty()) {
        return newSyncStatus(RangerCloudIdentitySyncState.NOT_APPLICABLE, "No datalakes associated with the environment.");
    } else if (!isDatalakeRunning(sdxCluster.get())) {
        return newSyncStatus(RangerCloudIdentitySyncState.NOT_APPLICABLE, "Datalake is not running for the environment.");
    } else {
        return operation.apply(sdxCluster.get());
    }
}

public RangerCloudIdentitySyncStatus setAzureCloudIdentityMapping(String envCrn, Map<String, String> azureUserMapping) {
    return runCloudIdSyncOperation(envCrn, sdxCluster -> setAzureCloudIdentityMapping(envCrn, sdxCluster, azureUserMapping));
}

public RangerCloudIdentitySyncStatus getRangerCloudIdentitySyncStatus(String envCrn, long commandId) {
    return runCloudIdSyncOperation(envCrn, sdxCluster -> {
        try {
            ApiCommand apiCommand = clouderaManagerRangerUtil.getApiCommand(sdxCluster.getStackCrn(), commandId);
            return toRangerCloudIdentitySyncStatus(apiCommand);
        } catch (ApiException e) {
            LOGGER.error("Encountered cloudera manager api exception", e);
            return newSyncStatus(RangerCloudIdentitySyncState.FAILED, "Encountered cloudera manager api exception");
        }
    });
}
```

lacikaaa (approved):

PR #8855, opened 2020-08-25
CB-8357 - added global ROLE_LIMITED_CLUSTER_ADMIN for CM users

attilapalfi92:

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerLdapService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerLdapService.java
@@ -42,6 +42,12 @@ public class ClouderaManagerLdapService {
     @Value("${cb.cm.limited.admin.role}")
     private String limitedAdminRole;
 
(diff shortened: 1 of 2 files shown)


attilapalfi92:
datalake creation in progress 4 trying this out

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerLdapService.java
```diff
@@ -36,6 +36,8 @@
 public class ClouderaManagerLdapService {
     private static final Logger LOGGER = LoggerFactory.getLogger(ClouderaManagerLdapService.class);
 
+    private static final String ROLE_LIMITED_CLUSTER_ADMIN = "ROLE_LIMITED_CLUSTER_ADMIN";
```
keyki, line 39: Should be an env var as the others.

keyki:
According to the ticket we need a dashboard user as well.

harshy-cloudera:
cluster admin role addition looks good.

PR #8856, opened 2020-08-25
CB-8470 Update AwsYcloudHybridCloudTest child environment with Telemetry steps

aszegedi:
In case of failing Hybrid Cloud E2E test we have no chance to check the issue in Kibana. So we need to introduce the related Telemetry configuration steps to the Yarn Environment creation.

First of all Cloudbreak should [allow Telemetry with no Logging Object](https://github.com/hortonworks/cloudbreak/pull/8846), this is needed only for YCloud where the Cloud Storage is not available. Beyond this we should modify the related E2E test case [AwsYcloudHybridCloudTest](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/hybrid/AwsYcloudHybridCloudTest.java) as well.


oleewere (approved):
lgtm

PR #8858, opened 2020-08-25
CB-8118 adding proto file for generate event data. This helps to the …

doktoric:
…core CP team to generate the audit documentation


bhavanki:
Cool, this is a great start. I think with some modifications it will work out as the first pass at this.

[index.html.gz](https://github.com/hortonworks/cloudbreak/files/5126547/index.html.gz)

I have attached the documentation that is generated from the proto file. The only real problem with it is the big blob of text for flowState. I experimented, and found that you can include HTML tags and they are obeyed. So, for example, you can add line breaks to make the text easier to read. I recommend trying that.

```
  // The state of the resource creation flow<br>
  // States for Data Hubs:<br>
  // INIT_STATE<br>
  // START_CREATION_EVENT<br>
...
```

Since there are different flow states for datalake vs. datahub, can you create two separate proto files, one for datalake and one for datahub? Then, the big list of flow states can be broken up, and I can have the generator provide one doc for datalake and another for datahub. You can continue to use just one of the generated Java classes in code, since they have the same structure.

Is this details structure used for environments as well / do environment flows emit audit events?

doktoric:
@bhavanki 
- I created only one file because we do the auditing in the core modul where everything is a cluster. Also I just using the generated java from proto in this case I can force to update the proto file first when somebody wants to do a modification (in this case the documentation is always updated as well).
- currently we dont have auditing for environment service.

bhavanki:
> @bhavanki
> 
>     * I created only one file because we do the auditing in the core modul where everything is a cluster. Also I just using the generated java from proto in this case I can force to update the proto file first when somebody wants to do a modification (in this case the documentation is always updated as well).
> 

Architecturally your choice makes sense. I'm just trying to think of a way to have two documents, one for datalake and another for datahub.

I could have the generator just make a second copy of the single proto file, but then it would need a way to select out just the datalake flow states and datahub flow states for each copy that it works from. How about including tokens in the comment that the scripting can process? Something like __DATAHUB_START__ and __DATAHUB_END__ for the section of text that only applies to datahub, and similar for datalake?

>     * currently we dont have auditing for environment service.

Great, that makes things simpler.

doktoric:
thanks @bhavanki

PR #8864, opened 2020-08-26
CB-7930 during SdxCreateAction SdxResponse was not set in the testcontext

lnardai:
SdxResponse was not set in the testcontext resulting in NPE


biharitomi (approved):
LGTM

doktoric (approved):

PR #8865, opened 2020-08-26
CB-7040 Upgrade salt states during Datalake upgrade

lacikaaa:
Salt upgrade has been added in 2 places:
- before starting the runtime upgrade the states pushed down to the GW
nodes and a highstate is called. This is a separate flow which is
added to the upgrade flowchain as first step. Also the component table
is updated with the new states
- as replacing the instances is implemented using full repair, this
commit adds a step to `upgradeOs` calls where it replaces the salt
states stored in `ClusterComponent` table. In this case we don't
initiate a highstate as the new GW node will come up with the new
states during the upgrade/repair. If the upgrade/repair flow cannot be
initiated the `ClusterComponent` is rolled back to previous version as
it is a versioned entity

As a result we don't check if salt version is matching in the future.

Notable refactors:
- `ClusterRepairService#validateRepair` had a big, ugly transaction.
This has been removed and where it was necessary separate repository
calls have been introduced
- there were some hand written checks for running flows which has been
eliminated as flow engine won't accept new flows if there is one
already running


foldik (approved):
lgtm

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/cluster/upgrade/ClusterUpgradeRequest.java
```diff
@@ -7,9 +7,4 @@
     public ClusterUpgradeRequest(Long stackId) {
         super(stackId);
     }
-
-//    @Override
```
keyki, line 11: Why is this commented out?

lacikaaa: it's not commented. the commented code is deleted

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageService.java
```diff
@@ -208,7 +155,8 @@ private StatedImage checkIfBasePermitted(StatedImage image, boolean baseImageEna
         return image;
     }
 
-    private Set<String> getSupportedOperationSystems(Long workspaceId, ImageSettingsV4Request imageSettings, String clusterVersion, String platform) throws Exception {
+    private Set<String> getSupportedOperationSystems(Long workspaceId, ImageSettingsV4Request imageSettings, String clusterVersion, String platform)
```
foldik, line 158: Can you please rename it to `getSupportedOperatingSystems`?

lacikaaa: sure

PR #8871, opened 2020-08-26
CB-8547: Use FreeIPA healthcheck for cluster proxy

jamisonbennett:
Register the FreeIPA healthcheck with cluster proxy for new clusters.
Continue using the old login page as the status check for clusters
that were provisioned prior to the healthcheck's existence.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
@@ -8,15 +8,15 @@ import com.sequenceiq.cloudbreak.common.type.Versioned;
 import com.sequenceiq.freeipa.entity.Stack;
 
 @Component
-public class ClusterProxyServiceAvailabilityChecker {
+public class HealthCheckAvailabilityChecker {
 
-    // feature supported from 2.21
-    private static final Versioned DNS_BASED_SERVICE_NAME_AFTER_VERSION = () -> "2.20.0";
+    // feature supported from 2.29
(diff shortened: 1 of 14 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class HealthCheckAvailabilityChecker {
+
+    // feature supported from 2.29
+    private static final Versioned CDP_FREEIPA_HEALTH_AGENT_AFTER_VERSION = () -> "2.28.0";
+
+    private HealthCheckAvailabilityChecker() {
+    }
+
+    public static boolean isCdpFreeIpaHeathAgentAvailable(Stack stack) {
```
lacikaaa, line 17: could you drop static and convert this class to `@Component` and use inject in `ClusterProxyService` so we can mock this class easily if we add unit tests for it?

lacikaaa (changes requested):
`TlsSecurityService` and `InstanceMetaDataRepository` are not used any more in the class but not removed. Could you remove them? thanks

freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
jamisonbennett: Sure. I also updated the other availability checkers to be components.

lacikaaa (approved):
LGTM
one note: these app version dependent checkers looks the same to me, only the version differs, so we might create one app version checker which expects a stack and a version number so we won't have the same logic all over again. We can still have a facade component if we need the same version check, like in case of `balancedDnsAvailabilityChecker` I would keep this component but it would just proxy the call to the app version checker with the exact version number.

PR #8874, opened 2020-08-27
CB-6185-Consistent HostNames

smaniraju:
The hostgroups are sorted to follow same privateId generation pattern for a given template.

 Gateway hostgroups start with privateId 0 (the hosgroup which maps to various endpoints like CM-API, HiveServer..).
 If multiple hostGroups are marked gateway then they are processed alphabetically so privateId remains same and hence hostName remaing same for a given template.

 Gateway hostgroups as defined in various default templates
 1. DE : master
 2. DE HA  : manager
 3. Datamart : master
 4. Flink : manager
 5. FlowManagement : gateway

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
@@ -69,6 +69,7 @@ import org.springframework.util.CollectionUtils;
 
 import javax.inject.Inject;
 import java.io.IOException;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -81,6 +82,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 
 import static com.sequenceiq.cloudbreak.service.metrics.MetricType.STACK_PREPARATION;
 import static com.sequenceiq.cloudbreak.util.Benchmark.measure;
@@ -370,9 +372,13 @@ public class StackCreatorService {
         }
     }
 
-    private void fillInstanceMetadata(Stack stack) {
+    void fillInstanceMetadata(Stack stack) {
         long privateIdNumber = 0;
-        for (InstanceGroup instanceGroup : stack.getInstanceGroups()) {
+        //Gateway HostGroups are sorted first to start with privateIdNumber 0.
+        List<InstanceGroup> sortedInstanceGroups = stack.getInstanceGroups().stream()
+                .sorted(Comparator.comparing(InstanceGroup::getInstanceGroupType)
+                        .thenComparing(InstanceGroup::getGroupName)).collect(Collectors.toList());
+        for (InstanceGroup instanceGroup : sortedInstanceGroups) {
             for (InstanceMetaData instanceMetaData : instanceGroup.getAllInstanceMetaData()) {
                 instanceMetaData.setPrivateId(privateIdNumber++);
                 instanceMetaData.setInstanceStatus(InstanceStatus.REQUESTED);
--- a/core/src/test/java/com/sequenceiq/cloudbreak/controller/StackCreatorServiceTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/controller/StackCreatorServiceTest.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.controller;
 
+import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


smaniraju:
Sample Hostname generated for Operational Database: Apache HBase, Phoenix template as below.

 smani-opdb-gateway0.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-leader1.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-leader1.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-master3.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-worker4.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-worker5.smani-lo.xcu2-8y8x.wl.cloudera.site
 smani-opdb-worker6.smani-lo.xcu2-8y8x.wl.cloudera.site

lacikaaa (changes requested):
please move this to master

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -370,9 +372,13 @@ private void setStackTypeAndValidateDatalake(Stack stack, Blueprint blueprint) {
         }
     }
 
-    private void fillInstanceMetadata(Stack stack) {
+    void fillInstanceMetadata(Stack stack) {
         long privateIdNumber = 0;
-        for (InstanceGroup instanceGroup : stack.getInstanceGroups()) {
+        //Gateway HostGroups are sorted first to start with privateIdNumber 0.
+        List<InstanceGroup> sortedInstanceGroups = stack.getInstanceGroups().stream()
+                .sorted(Comparator.comparing(InstanceGroup::getInstanceGroupType)
```
sidseth, line 379: I don't think this will provide the Knox node(s) (which effectively is the endpoint for all services) with a consistent hostname. e.g. If the number of 'CORE' nodes is dynamic - the 'GATEWAY' instance will get a number which follows the count of all 'CORE' nodes.
This needs better understanding of 'CORE' vs 'GATEWAY'. If I'm not mistaken, there can be one GATEWAY hostGroup, and this is the instance on which KNOX is installed (not sure if there can be more than one such node). I could be completely wrong about this though. @cegganesh84 - I think you have a much better understanding of CORE vs GATEWAY and cardinality given the HA templates that you've worked on.
If GATEWAY==KNOX node and there can be only 1 gateway hostGroup, explicitly assigning numbers to this hostGroup first probably works.
Alternately, may need to make this service aware.

sidseth: Looking some more, looks like GATEWAY is the first in the sort order, so this works as expected. May still be better to explicitly look at GATEWAY nodes first.
Still need to get an understanding of where Knox goes, and how many GATEWAY type groups can exist.

biharitomi (approved):
LGTM

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
biharitomi: Knox is always deployed to the CM's machine which group always have the `GATEWAY` as InstanceGroupType. In case of HA clusters where could be multiple gateway instance/host group due to the comparison on group name the private id selection should be consistent - I think

keyki: Previously, we had multiple groups with GATEWAY type for Ambari HA. At some point, the same will come to CM HA as well.

smaniraju: Yes, if multiple GATEWAY host-groups are defined in future (current template validation fails multiple gateway host-groups), the naming convention will still be consistent for the same template due to comparison on InstanceGroupType which is followed by groupName. so nodes will masterA0, masterB1  if masterA, masterB are gateway hostgroups.

Also tested cluster launch by changing gateway hostGroup from "master"  to "worker"  and then "worker" nodes started with worker0.

CB-6185 concern is basically about different\random node sequenceId and hence node name for same template and cardinality and that issue would be addressed by this patch .

cegganesh84:
Just looking at this, wondering if a manager host in HA template is repaired, will it still get the same manager0 hostname. I am not saying, if it is not, then address it right here. Just want to have a shared understanding, that is all.

keyki:
> Just looking at this, wondering if a manager host in HA template is repaired, will it still get the same manager0 hostname. I am not saying, if it is not, then address it right here. Just want to have a shared understanding, that is all.

Yes, repaired hosts get the same hostname.

smaniraju:
The swagger compatibility test failure is not related to this PR.

PR #8881, opened 2020-08-28
CB-8555 Introduce Await For Flow exceptions to TestContext

aszegedi:
Await for flow errors/exceptions should be present in the TestContext's ExceptionMap as well. So we can see these issues on the test suite's logs and HTML reports.

We can see these errors in the HTML report as:

|Step | Details|
|---|---|
| Await | Sdx internal await for flow SdxInternalTestDto |
|           | Sdx internal await for flow 'SdxInternalTestDto[name: aws-test-654a07c3f43d4ab497577e79b82dd2b]' is failed for 'aws-test-654a07c3f43d4ab497577e79b82dd2b', because of Error during polling flow. Crn=crn:cdp:datalake:us-west-1:qe-aws:datalake:838ac751-2493-4b2c-baaa-5ea4b72c6201, FlowId=da25d4fc-fc27-4a3c-a7d2-dae90cd568e3 , FlowChainId=null, Message=The template variable 'flowId' has no value |
|          | Datalake Internal await should be skipped because of previous error. await [RUNNING] |

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/freeipa/FreeIpaRefreshAction.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/freeipa/FreeIpaRefreshAction.java
@@ -1,0 +1,24 @@
+package com.sequenceiq.it.cloudbreak.action.freeipa;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.it.cloudbreak.FreeIpaClient;
+import com.sequenceiq.it.cloudbreak.action.Action;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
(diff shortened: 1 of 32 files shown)


oleewere:
@aszegedi , as i had refactor in the flow util, can we merge this one first: https://github.com/hortonworks/cloudbreak/pull/8883 i have split the original PR for diagnostics into.2 parts, so this PR does not contain any test changes just the dto-s/actions and client changes.
if that's merged i can fix the conflicts here

i would like to merge that part as that's done for long time but as validation the tests itself requires that much time, when im finishing it always have new conflicts because of the flow changes, so it would speed me up if i would not need to always re-check everhthing and fixing conflicts over and over again

oleewere:
thanks @aszegedi , i have updated/rebased your changes against master (+ adding changes to TestContext object for diagnostics DTOs, similar as it was done in your commit)

PR #8884, opened 2020-08-29
CB-7976 Modify framework for dual Right checking

sodre90:

### DIFF
--- a/environment/src/test/java/com/sequenceiq/environment/service/integration/CredentialAuthorizationIntegrationTest.java
+++ b/environment/src/test/java/com/sequenceiq/environment/service/integration/CredentialAuthorizationIntegrationTest.java
@@ -1,5 +1,8 @@
 package com.sequenceiq.environment.service.integration;
 
+import static com.sequenceiq.authorization.resource.AuthorizationResourceAction.DELETE_CREDENTIAL;
+import static com.sequenceiq.authorization.resource.AuthorizationResourceAction.DESCRIBE_CREDENTIAL;
+import static com.sequenceiq.authorization.resource.AuthorizationResourceAction.EDIT_CREDENTIAL;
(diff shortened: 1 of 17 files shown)


environment/src/test/java/com/sequenceiq/environment/service/integration/CredentialAuthorizationIntegrationTest.java
```diff
@@ -122,8 +122,8 @@ public void testCredentialPermissions() {
         credentialRepository.save(getAwsCredential(FIRST_CRED_NAME, ACCOUNT_ID, FIRST_USER_CRN));
         credentialRepository.save(getAwsCredential(SECOND_CRED_NAME, ACCOUNT_ID, SECOND_USER_CRN));
 
-        testUnhappyPaths(firstUserClient, SECOND_CRED_NAME);
-        testUnhappyPaths(secondUserClient, FIRST_CRED_NAME);
+//        testUnhappyPaths(firstUserClient, SECOND_CRED_NAME);
+//        testUnhappyPaths(secondUserClient, FIRST_CRED_NAME);
```
horadla23, line 126: What was the issue with these? Credential should work like before I think

horadla23 (commented):
LGTM, I had only one question regarding credential authz springboot test changes

environment/src/test/java/com/sequenceiq/environment/service/integration/CredentialAuthorizationIntegrationTest.java
sodre90: It was a mistake, I forgot to uncomment these two lines! Thanks Adam!

horadla23 (approved):

lnardai (approved):

PR #8887, opened 2020-08-31
CB-8589 if credential has no namein request then we are generating a …

doktoric:
…uuid. Also you are able to delete any audit credential which is a new endpoint.

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/endpoint/AuditCredentialEndpoint.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/endpoint/AuditCredentialEndpoint.java
@@ -5,6 +5,7 @@ import static com.sequenceiq.environment.api.doc.credential.CredentialDescriptor
 
 import javax.validation.Valid;
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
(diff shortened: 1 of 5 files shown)


environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/endpoint/AuditCredentialEndpoint.java
```diff
@@ -74,4 +75,18 @@
             notes = CredentialDescriptor.CREDENTIAL_NOTES, nickname = "getAuditPrerequisitesForCloudPlatform", httpMethod = "GET")
     CredentialPrerequisitesResponse getPrerequisitesForCloudPlatform(@PathParam("cloudPlatform") String platform,
         @QueryParam("deploymentAddress") String deploymentAddress);
+
+    @DELETE
+    @Path("name/{name}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = CredentialOpDescription.DELETE_BY_NAME, produces = MediaType.APPLICATION_JSON,
```
topolyai5, line 82: Does the credential note relate to env cred, right?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/CredentialBase.java
```diff
@@ -29,7 +29,6 @@
     @Size(max = 100, min = 5, message = "The length of the credential's name has to be in range of 5 to 100")
     @Pattern(regexp = "(^[a-z][-a-z0-9]*[a-z0-9]$)",
             message = "The name of the credential can only contain lowercase alphanumeric characters and hyphens and has start with an alphanumeric character")
-    @NotNull
```
topolyai5, line 32: Hm, the env cred cannot be null

doktoric: the name can be https://github.com/hortonworks/cloudbreak/pull/8887/files?file-filters%5B%5D=#diff-592e562d2b48995f6e0279a1ef5606acR106

topolyai5 (approved):

PR #8891, opened 2020-08-31
CB-8586 Check and fill FMS log context

lacikaaa:
We have missed setting the `MDCContext` in many places. This commit
tries to fix those cases.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
@@ -20,6 +20,7 @@ import com.googlecode.jsonrpc4j.JsonRpcClient.RequestListener;
 import com.sequenceiq.cloudbreak.ccm.endpoint.ServiceFamilies;
 import com.sequenceiq.cloudbreak.client.HttpClientConfig;
 import com.sequenceiq.cloudbreak.clusterproxy.ClusterProxyConfiguration;
+import com.sequenceiq.cloudbreak.logger.MDCBuilder;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status;
 import com.sequenceiq.freeipa.client.ClusterProxyErrorRpcListener;
 import com.sequenceiq.freeipa.client.FreeIpaClient;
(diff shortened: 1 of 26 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -87,6 +88,7 @@ public FreeIpaClient getFreeIpaClientForStackId(Long stackId) throws FreeIpaClie
 
     public FreeIpaClient getFreeIpaClientByAccountAndEnvironment(String environmentCrn, String accountId) throws FreeIpaClientException {
         Stack stack = stackService.getByEnvironmentCrnAndAccountId(environmentCrn, accountId);
+        MDCBuilder.buildMdcContext(stack);
```
handavid, line 91: would it make sense to move this to the private `getFreeIpaClient` method? Then it would work for all the various public methods of this class.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -36,6 +37,7 @@
 
     public void deleteDnsRecordByFqdn(String environmentCrn, String accountId,  List<String> fqdns) throws FreeIpaClientException {
         Stack stack = stackService.getByEnvironmentCrnAndAccountId(environmentCrn, accountId);
+        MDCBuilder.buildMdcContext(stack);
         FreeIpa freeIpa = freeIpaService.findByStack(stack);
         FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientByAccountAndEnvironment(environmentCrn, accountId);
```
handavid, line 42: we already have the Stack so we can avoid a redundant read by calling `getFreeIpaClientForStack` instead of `getFreeIpaClientByAccountAndEnvironment`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
lacikaaa: I have added MDC building where the context was insufficient. I started from the endpoints and followed the call chain to check if the context is set at the earliest point where `Stack` is available. Most probably the calls to the `FreeIpaClientFactory` come from flows, which have the context already set (at least I hope so, haven't checked all of them). But it's possible I missed some places so I'm open to suggestions :)

handavid: it makes sense to set the MDC at the earliest possible moment. The `FreeIpaClientFactory` should be able to rely on callers of `getFreeIpaClientForStack` to have already set the MDC.

PR #8893, opened 2020-08-31
CB-8248: FreeIPA reboot flow finishes before instances reboot

christmasferret:
Use combined stop/start operations in the FreeIPA instance reboot flow so that FreeIPA repair will wait for the AWS/Azure nodes to return to an actual running state. Currently, during a reboot operation, the AWS/Azure nodes always report a running state (this is so that the billing doesn't start billing a new session). With this fix, the reboot flow will actually finish. AWS will only bill for a minimum of 60 seconds each time a new instance is started.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
@@ -14,6 +14,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
@@ -30,7 +31,6 @@ import com.amazonaws.services.ec2.model.DescribeInstancesResult;
 import com.amazonaws.services.ec2.model.GetConsoleOutputRequest;
 import com.amazonaws.services.ec2.model.GetConsoleOutputResult;
 import com.amazonaws.services.ec2.model.Instance;
-import com.amazonaws.services.ec2.model.RebootInstancesRequest;
 import com.amazonaws.services.ec2.model.Reservation;
 import com.amazonaws.services.ec2.model.StartInstancesRequest;
 import com.amazonaws.services.ec2.model.StopInstancesRequest;
@@ -39,7 +39,6 @@ import com.sequenceiq.cloudbreak.cloud.InstanceConnector;
 import com.sequenceiq.cloudbreak.cloud.aws.poller.PollerUtil;
 import com.sequenceiq.cloudbreak.cloud.aws.util.AwsInstanceStatusMapper;
 import com.sequenceiq.cloudbreak.cloud.aws.view.AuthenticatedContextView;
-import com.sequenceiq.cloudbreak.cloud.aws.view.AwsCredentialView;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
 import com.sequenceiq.cloudbreak.cloud.exception.CloudOperationNotSupportedException;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
@@ -56,9 +55,6 @@ public class AwsInstanceConnector implements InstanceConnector {
 
     @Inject
     private PollerUtil pollerUtil;
-
-    @Inject
-    private AwsClient awsClient;
 
     @Value("${cb.aws.hostkey.verify:}")
     private boolean verifyHostKey;
@@ -131,25 +127,31 @@ public class AwsInstanceConnector implements InstanceConnector {
     )
     @Override
     public List<CloudVmInstanceStatus> reboot(AuthenticatedContext ac, List<CloudInstance> vms) {
-        AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
-                ac.getCloudContext().getLocation().getRegion().value());
-        List<CloudInstance> affectedVms = new ArrayList<>();
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
+
         try {
             if (!vms.isEmpty()) {
                 List<CloudVmInstanceStatus> statuses = check(ac, vms);
-                doReboot(affectedVms, amazonEC2Client, getStarted(statuses));
-                doStart(affectedVms, ac, getStopped(statuses));
-                logInvalidStatuses(getNotStoppedOrStarted(statuses));
+                doStop(ac, getStarted(statuses));
+                statuses = check(ac, vms);
+                logInvalidStatuses(getNotStopped(statuses), InstanceStatus.STOPPED);
+                rebootedVmsStatus = doStart(ac, getStopped(statuses));
+                logInvalidStatuses(getNotStarted(statuses), InstanceStatus.STARTED);
             }
         } catch (SdkClientException e) {
             LOGGER.warn("Failed to send reboot request to AWS: ", e);
             throw e;
         }
-        return pollerUtil.waitFor(ac, affectedVms, Sets.newHashSet(InstanceStatus.STARTED, InstanceStatus.FAILED));
+        return rebootedVmsStatus;
     }
 
-    private List<CloudVmInstanceStatus> getNotStoppedOrStarted(List<CloudVmInstanceStatus> statuses) {
-        return statuses.stream().filter(status -> status.getStatus() != InstanceStatus.STOPPED && status.getStatus() != InstanceStatus.STARTED)
+    private List<CloudVmInstanceStatus> getNotStarted(List<CloudVmInstanceStatus> statuses) {
+        return statuses.stream().filter(status -> status.getStatus() != InstanceStatus.STARTED)
+                .collect(Collectors.toList());
+    }
+
+    private List<CloudVmInstanceStatus> getNotStopped(List<CloudVmInstanceStatus> statuses) {
+        return statuses.stream().filter(status -> status.getStatus() != InstanceStatus.STOPPED)
                 .collect(Collectors.toList());
     }
 
@@ -163,32 +165,35 @@ public class AwsInstanceConnector implements InstanceConnector {
                 .map(status -> status.getCloudInstance()).collect(Collectors.toList());
     }
 
-    private void doReboot(List<CloudInstance> affectedVMs, AmazonEC2Client amazonEC2Client, List<CloudInstance> instances) {
+    private List<CloudVmInstanceStatus> doStart(AuthenticatedContext ac, List<CloudInstance> instances) {
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
         for (CloudInstance instance: instances) {
             try {
-                amazonEC2Client.rebootInstances(new RebootInstancesRequest().withInstanceIds(List.of(instance.getInstanceId())));
-                affectedVMs.add(instance);
-            } catch (AmazonEC2Exception e) {
-                LOGGER.warn(String.format("Unable to reboot instance %s", instance), e);
-            }
-        }
-    }
-
-    private void doStart(List<CloudInstance> affectedVMs, AuthenticatedContext ac, List<CloudInstance> instances) {
-        for (CloudInstance instance: instances) {
-            try {
-                start(ac, null, List.of(instance));
-                affectedVMs.add(instance);
+                rebootedVmsStatus.addAll(start(ac, null, List.of(instance)));
             } catch (AmazonEC2Exception e) {
                 LOGGER.warn(String.format("Unable to start instance %s", instance), e);
             }
         }
+        return rebootedVmsStatus;
     }
 
-    private void logInvalidStatuses(List<CloudVmInstanceStatus> instances) {
-        for (CloudVmInstanceStatus instance: instances) {
-            LOGGER.warn(String.format("Unable to reboot instance %s because of invalid status %s.",
-                    instance.getCloudInstance().getInstanceId(), instance.getStatus().toString()));
+    private void doStop(AuthenticatedContext ac, List<CloudInstance> instances) {
+        for (CloudInstance instance: instances) {
+            try {
+                stop(ac, null, List.of(instance));
+            } catch (AmazonEC2Exception e) {
+                LOGGER.warn(String.format("Unable to stop instance %s", instance), e);
+            }
+        }
+    }
+
+    public void logInvalidStatuses(List<CloudVmInstanceStatus> instances, InstanceStatus targetStatus) {
+        if (CollectionUtils.isNotEmpty(instances)) {
+            StringBuilder warnMessage = new StringBuilder("Unable to reboot ");
+            warnMessage.append(instances.stream().map(instance -> String.format("instance %s because of invalid status %s",
+                    instance.getCloudInstance().getInstanceId(), instance.getStatus().toString())).collect(Collectors.joining(", ")));
+            warnMessage.append(String.format(". Instances should be in %s state.", targetStatus.toString()));
+            LOGGER.warn(warnMessage.toString());
         }
     }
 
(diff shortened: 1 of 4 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureInstanceConnector.java
```diff
@@ -78,7 +78,8 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
             AzureClient azureClient = ac.getParameter(AzureClient.class);
             String resourceGroupName = azureResourceGroupMetadataProvider.getResourceGroupName(ac.getCloudContext(), vm.getCloudInstance());
             if (vm.getStatus() == InstanceStatus.STARTED) {
-                doReboot(completables, vm, statuses, () -> azureClient.rebootVirtualMachineAsync(resourceGroupName, vm.getCloudInstance().getInstanceId()));
+                doReboot(completables, vm, statuses, () -> azureClient.stopVirtualMachineAsync(resourceGroupName, vm.getCloudInstance().getInstanceId())
+                        .andThen(azureClient.startVirtualMachineAsync(resourceGroupName, vm.getCloudInstance().getInstanceId())));
```
jamisonbennett, line 82: I am a little confused as to why there is an async stop followed by an async start. Does `andThen()` wait for the async to complete? If so, why even use async (unless it is referenced elsewhere).

christmasferret: Thank you very much for the review. I verified that andThen() is a blocking call and wait for the first async completable to finish. Also it will re-throw any error that is thrown by the first completable. The link to the snippet I use to verify that : https://gist.github.com/christmasferret/3c0a185160ef420526bf882090517ca7

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -134,18 +134,21 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
         AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
```
lacikaaa, line 134: this is not used anymore

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -185,6 +188,28 @@ private void doStart(List<CloudInstance> affectedVMs, AuthenticatedContext ac, L
         }
     }
 
+    private void doStart(List<CloudInstance> affectedVMs, AuthenticatedContext ac, List<CloudInstance> instances, List<CloudVmInstanceStatus> rebootedVmsStatus) {
```
lacikaaa, line 191: why did you introduce a new `doStart`? the old one is not used anymore

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -134,18 +134,21 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
         AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
                 ac.getCloudContext().getLocation().getRegion().value());
         List<CloudInstance> affectedVms = new ArrayList<>();
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
+
         try {
             if (!vms.isEmpty()) {
                 List<CloudVmInstanceStatus> statuses = check(ac, vms);
-                doReboot(affectedVms, amazonEC2Client, getStarted(statuses));
```
lacikaaa, line 140: `doReboot` is not used anymore

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -134,18 +134,21 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
         AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
                 ac.getCloudContext().getLocation().getRegion().value());
         List<CloudInstance> affectedVms = new ArrayList<>();
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
```
lacikaaa, line 137: why do we need this new list of statuses? shouldn't the `affectedVms` fill the same purpose?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -134,18 +134,21 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
+
         try {
             if (!vms.isEmpty()) {
                 List<CloudVmInstanceStatus> statuses = check(ac, vms);
-                doReboot(affectedVms, amazonEC2Client, getStarted(statuses));
-                doStart(affectedVms, ac, getStopped(statuses));
+                doStop(new ArrayList<>(), ac, getStarted(statuses));
+                statuses = check(ac, vms);
+                doStart(affectedVms, ac, getStopped(statuses), rebootedVmsStatus);
```
lacikaaa, line 144: instead of passing `affectedVms` list to `doStop` and `doStart` it would be nicer if they return with a list and we can use that

lacikaaa: I miss some action here. what happens here is:
- issue stop for VMs which are running
- collect the VM statuses
- start VMs which are stopped

In theory all VMs should be stopped at the point we are collecting the statuses after stop. I think we should at least match that the requested VMs for reboot are all stopped and have a log here if any of them is not stopped

lacikaaa (changes requested):
could you add some unit test? especially for `AwsInstanceConnector` as it has stop, start, check already covered. thanks

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -134,18 +134,21 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
                 logInvalidStatuses(getNotStoppedOrStarted(statuses));
             }
         } catch (SdkClientException e) {
             LOGGER.warn("Failed to send reboot request to AWS: ", e);
             throw e;
         }
-        return pollerUtil.waitFor(ac, affectedVms, Sets.newHashSet(InstanceStatus.STARTED, InstanceStatus.FAILED));
+        return rebootedVmsStatus;
```
lacikaaa, line 151: you have dropped polling here. are we sure about `AmazonEC2Client#stopInstances` and `AmazonEC2Client#startInstances` will return only after the state is changed to stopped/started? if not we might need some polling there

lacikaaa (changes requested):

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Thank you for pointing this out. It is removed.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Thank you for the suggestion! I removed _affectedVms_. Also, the _rebootedVmsStatus_ will be returned from doStart() method as suggested in another comment.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Thank you! I removed the line.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Yes. I returned _rebootedVmsStatus_ from doStart() method. Thanks for the suggestion.

I also added a specific log check to see if any requested VMs for reboot were not stopped  
-_logInvalidStatuses(getNotStopped(statuses));_

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: The reason to drop the polling outside here is that we already specifically do polling inside doStop() and doStart() as below. The polling will take 1000 seconds total. In the case of a hang blocking call, we will wait twice as much long. It is not necessary to poll and then immediately poll again. I added unit test and also did integration testing with AWS. 

```
doStop(...) { 
	stop(...) {
		setCloudVmInstanceStatuses(...){
			pollerUtil.waitFor(...);
		}
	}
}
```

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: The reason to introduce new doStart() is we want to return the polling results from doStart(). The polling will take 1000 seconds total. In the case of a hang blocking call, we will wait twice as much long. It is not necessary to poll inside doStart() and then after return from doStart() immediately poll again.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
lacikaaa: ok I missed that! thanks for pointing out

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -131,25 +126,31 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
-        List<CloudInstance> affectedVms = new ArrayList<>();
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
+
         try {
             if (!vms.isEmpty()) {
                 List<CloudVmInstanceStatus> statuses = check(ac, vms);
-                doReboot(affectedVms, amazonEC2Client, getStarted(statuses));
-                doStart(affectedVms, ac, getStopped(statuses));
-                logInvalidStatuses(getNotStoppedOrStarted(statuses));
+                doStop(new ArrayList<>(), ac, getStarted(statuses));
```
lacikaaa, line 134: `doStop` first parameter is never used, could you drop it?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -131,25 +126,31 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
+                statuses = check(ac, vms);
+                logInvalidStatuses(getNotStopped(statuses));
+                rebootedVmsStatus = doStart(new ArrayList<>(), ac, getStopped(statuses));
```
lacikaaa, line 137: `doStart` first parameter case is the same as `doStop`. could you drop it?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -131,25 +126,31 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
+                logInvalidStatuses(getNotStarted(statuses));
```
lacikaaa, line 138: I think `logInvalidStatuses` would need some extra parameter which is also logged. Like "instances should be in `STOPPED` state" and here comes the list of instance ids and states. Otherwise we won't know if the log is coming from line 136 or 138.
Also `logInvalidStatuses` should log only one line, and not separate lines per instance. Could you fix that?

lacikaaa (changes requested):
please don't forget to squash your commits into 1 commit

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Thanks. I dropped it.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: Thanks. I dropped it.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
christmasferret: I fixed both here. 

Now it will only log one line. Also there is a targetStatus parameter to log the expected status.

Below is a sample log message. 
`
Unable to reboot instance i-1 because of invalid status STARTED, instance i-2 because of invalid status STARTED. Instances should be in STOPPED state.`

christmasferret: I have also squashed all my commit into one commit.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -163,33 +164,37 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
...
         }
     }
 
-    private void logInvalidStatuses(List<CloudVmInstanceStatus> instances) {
-        for (CloudVmInstanceStatus instance: instances) {
-            LOGGER.warn(String.format("Unable to reboot instance %s because of invalid status %s.",
-                    instance.getCloudInstance().getInstanceId(), instance.getStatus().toString()));
+    public void logInvalidStatuses(List<CloudVmInstanceStatus> instances, InstanceStatus targetStatus) {
+        if (instances == null || instances.size() < 1) {
+            return;
```
lacikaaa, line 191: multiple issues here: 
- please don't use return like this, if possible avoid in the middle of returns in methods, or use `else` to indicate there are multiple exit points in the method
- if you would like to check if a collection is empty, please use `isEmpty()`
- I would negate the if condition and put the body of the method inside. You can use `CollectionUtils` from apache (`CollectionUtils.isNotEmpty(instances)`) or spring (`!CollectionUtils.isEmpty(instances)`) or write the whole condition by hand (`instances != null && !instances.isEmpty()`). It would still result in the same logic but there won't be that unnecessary return

lacikaaa (changes requested):

lacikaaa (approved):
please don't forget to squash your commits

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsInstanceConnector.java
```diff
@@ -131,25 +127,31 @@ public String getConsoleOutput(AuthenticatedContext authenticatedContext, CloudI
     )
     @Override
     public List<CloudVmInstanceStatus> reboot(AuthenticatedContext ac, List<CloudInstance> vms) {
-        AmazonEC2Client amazonEC2Client = awsClient.createAccess(new AwsCredentialView(ac.getCloudCredential()),
-                ac.getCloudContext().getLocation().getRegion().value());
-        List<CloudInstance> affectedVms = new ArrayList<>();
+        List<CloudVmInstanceStatus> rebootedVmsStatus = new ArrayList<>();
```
lacikaaa, line 130: one more nit, but we can merge it without even if you don't change this:
you can declare this variable right in the line where it gets value, at `doStart`. also the return could be moved inside the try, right after `logInvalidStatuses(getNotStarted(statuses), InstanceStatus.STARTED);`

lacikaaa: oh I missed `if (!vms.isEmpty())`, nevermind....

PR #8894, opened 2020-09-01
Revert "CB-8547: Use FreeIPA healthcheck for cluster proxy"

jamisonbennett:
This reverts commit 21613be7e6c5eb70fd471fb842ed1edde6d785f2.


jamisonbennett:
I am reverting this because there was an issue with the e2e tests where the health check would intermittently fail during provisioning.

akanto (approved):

PR #8896, opened 2020-09-01
CLOUDQE-257: Add gateway hostgroup for Data Engineering Spark 3 template in Public Cloud

roczei:
Hi Cloudbreak team,

The gateway hostgroup is missing from this Spark 3 cluster template. It is necessary for the Public Cloud QE testing. Could you please add it?

CC @dabondor

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-spark3-722.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-spark3-722.bp
@@ -278,7 +278,19 @@
           "yarn-NODEMANAGER-COMPUTE",
           "yarn-GATEWAY-BASE"
         ]
+      },
+      {
+        "refName": "gateway",
+        "cardinality": 0,
+        "roleConfigGroupsRefNames": [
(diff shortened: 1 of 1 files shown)


roczei:
Hi @doktoric and @dabondor,

I did some further validations and the Spark 3 testing is working without the gateway hostgroup. The compute hostgroup is enough which contains the gateway roles as well.

If we want to add a new hostgroup, this requires that we update these json files as well and we have to define the related extra host:

- https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/defaults/clustertemplates/aws/aws-dataengineering-spark3-722.json
- https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/defaults/clustertemplates/azure/azure-dataengineering-spark3-722.json

I have opened a new jira which is related to this. Seems like we only have this extra gateway hostgroup by the YCloud test environment: https://jira.cloudera.com/browse/CLOUDQE-266

==========
Summary
==========

Based on the above findings and I do not want to add this gateway hostgroup. I am going to cancel this pull request.

PR #8903, opened 2020-09-01
CB-8577 - cdp-proxy-api topology files needs ACLs for hbasejars service

moresandeep:
HBASEJARS services in cdp-proxy-api topology needs acls just like other services defined in topology_api.xml.j2
This is a very simple change and should not have any side effects.


moresandeep:
Awesome, thanks @keyki !!

joshelser:
Thanks Sandeep! Looks great :)

PR #8907, opened 2020-09-02
CB-8597, CB-8604 Remove create new Single Resource Group option from …

gergopapi2:
…backend, AND Storage Account deletion causes issues with other ENVs in the same Resource Group

On azure in case of cloudbreak created single RG termination required the RG to be empty. This meant that storage account and image had to be deleted as well. This, however had serious implications in other cases, like another environment using that storage account would be crippled. Partly because of these issues it was decided to remove the whole feature and leave images and the storage account in place.


environment/src/test/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverterTest.java
```diff
@@ -174,35 +172,6 @@ void testInitEditDto() {
         verify(networkRequestToDtoConverter).convert(request.getNetwork());
     }
 
-    @ParameterizedTest
-    @ValueSource(booleans = { false, true })
-    void testAzureSingleRgEnabledWithEmptyAzureRequest(Boolean azureSingleResourceGroupDeploymentEnabled) {
```
pdarvasi, line 179: May I ask, why was this deleted?

gergopapi2: Because now, even if entitlement for single RG exists, resources will not be deployed into a single automatically, only if the name of the RG and SINGLE usage is provided. From now on cloudbreak needs a name I think.

gergopapi2: After personal communication we agreed to reintroduce the test. I also added another one.

pdarvasi (approved):

PR #8909, opened 2020-09-02
CB-8620 added minimal set of permission to audit credential policy. D…

doktoric:
…escribe region is required because of the validation.

### DIFF
--- a/cloud-aws/src/main/resources/definitions/aws-audit-policy.json
+++ b/cloud-aws/src/main/resources/definitions/aws-audit-policy.json
@@ -1,24 +1,38 @@
 {
+  "Version": "2012-10-17",
   "Statement": [
     {
-      "Sid": "SimpleStorageServiceFull",
+      "Effect": "Allow",
       "Action": [
-        "s3:AbortMultipartUpload",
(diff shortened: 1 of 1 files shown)


topolyai5 (approved):

cloud-aws/src/main/resources/definitions/aws-audit-policy.json
```diff
@@ -1,24 +1,38 @@
...
-        "s3:PutBucketPublicAccessBlock"
+        "s3:DeleteObject"
+      ],
+      "Resource": [
+        "arn:aws:s3:::<your-bucket-name-for-audit-archiving>/*"
+      ]
+    },
+    {
+      "Action": [
+        "ec2:DescribeRegions"
```
jadair-cloudera, line 30: I don't have this in my own policy, and audit archiving is working fine. I think this is a relatively harmless permission, though, so if you have reason to believe it is necessary, it is fine to leave it in.

doktoric: this is for validate the credential on our side

PR #8910, opened 2020-09-02
CB-8617 if the gw node has a valid hostname (called userfacingfqdn) t…

akanto:
…hen CB is trying configure and restart CM on every node where knox is present. This issue has been caused by CB-7917 Certificate renewal should be able to change the generated self-signed certificates to valid ones

This commits fixes the described problem with a simple refactor, that ensures:
- CM is not started/restarted multiple times since all CM start command are after each other with properly defined watch conditions
- CM reconfiguration should happen only when the certificate has changed and only on CM node
- CM configuration is done only on those nodes where knox is present

Code can be improved further by ensuring that the "knox" role is always appended to the node which contains knox instances, that improvemnt is not part of this commit, but I have added a comment message into the top.sls file to remind ourselves.


orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/start.sls
```diff
@@ -8,6 +9,27 @@ init-cloudera-manager-db:
...
+  file.replace:
+      - name: /etc/cloudera-scm-server/cm.settings
+      - pattern: 'setsettings FRONTEND_URL.*'
+      - repl: 'setsettings FRONTEND_URL https://{{ salt['pillar.get']('gateway:userfacingfqdn') }}/'
+      - unless: grep "^setsettings FRONTEND_URL https://{{ salt['pillar.get']('gateway:userfacingfqdn') }}/$" /etc/cloudera-scm-server/cm.settings
+
+restart_cm_after_fronted_url_change:
+  service.running:
+    - name: cloudera-scm-server
+    - enable: True
```
lacikaaa, line 24: I think we can omit this one as `start_server` already contains it, no reason to duplicate the same setting

akanto: ok

lacikaaa (approved):

PR #8912, opened 2020-09-02
CB-8211: Set HBase root directory in Datalake as S3 path

LinaAtAustin:
When the cluster is Data Lake, still set the HBase root directory to use cloud storage if the cloud storage base is configured.

In "DISTX-303 DistroX Hbase Cloud Storage", it only sets HBase root directory for workload cluster. We need to extend cloud storage for HBase to the Data Lake cluster.


keyki:
Are there any version limitations? Like this will work from CDP 7.2.1 or will it work on older versions as well?

LinaAtAustin:
@keyki "Are there any version limitations? Like this will work from CDP 7.2.1 or will it work on older versions as well?" This change works with any version that has update in CB-8210 (Store HBase WAL file in a directory different from HBase root directory). 

The reasons are
1) HBase WAL file has to be stored in HDFS for now.
2) Without setting HBase WAL directory, WAL directory is derived from HBase root. And therefore requires HBase root using HDFS
3) After configuring HBase WAL file directory specifically using HDFS, we can safely set HBase root directory using Cloud storage.

PR #8913, opened 2020-09-02
CB-8611. Implement CB flow for CM based diagnostics collection (DistroX)

oleewere:
some UTs will be needed for these changes as well, see description in the commit message

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDiagnosticsService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDiagnosticsService.java
@@ -1,0 +1,108 @@
+package com.sequenceiq.cloudbreak.cm;
+
+import static com.sequenceiq.cloudbreak.polling.PollingResult.isExited;
+import static com.sequenceiq.cloudbreak.polling.PollingResult.isTimeout;
+
+import java.math.BigDecimal;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
+import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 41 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDiagnosticsService.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new ClusterClientInitException(e);
+        }
+    }
+
+    @Override
+    public void collectDiagnostics(CmDiagnosticsParameters parameters) throws CloudbreakException {
+        ClouderaManagerResourceApi resourceApi = clouderaManagerApiFactory.getClouderaManagerResourceApi(client);
+        try {
+            ApiCommand collectDiagnostics = resourceApi.collectDiagnosticDataCommand(convertToCollectDiagnosticDataArguments(parameters));
+            PollingResult pollingResult = clouderaManagerPollingServiceProvider.startPollingCollectDiagnostics(stack, client, collectDiagnostics.getId());
```
attilapalfi92, line 81: a log would be nice here. "starting polling blabla..." or something

attilapalfi92 (commented):
other than 1 line of log, lgtm

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDiagnosticsService.java
oleewere: actually there is a log in startPollingCollectDiagnostics call itself

attilapalfi92 (approved):

PR #8928, opened 2020-09-03
CB-8667 Add all the deleted resources to the error message

aszegedi:
API Cleanup App error message is not correct:
```
Caused by: java.lang.RuntimeException: End To End cleanup have been failed, because of credential(s) '[aws-test-bd84fafad3c24769bf327e633d69d47, yarn-test-4f08108497dd4abe9d8dd47382aef9]' found left behind!
```
However there is also:
```
environmentName:aws-test-9061250527914a019dc will be deleted!
credentialName:aws-test-bd84fafad3c24769bf327e633d69d47 will be deleted!
credentialName:yarn-test-4f08108497dd4abe9d8dd47382aef9 will be deleted!
```
Need to collect the correct list of deleted resources for the final error message.


foldik (approved):
lgtm

PR #8931, opened 2020-09-04
DISTX-504 Optimize DB Cluster Update

smaniraju:
AS Cluster's lastEvaluated and lastScalingActivity are updated at a very high rate for every alert monitor run. Saving the entire cluster object hierarchy in this flow is inefficient and unnecessary. Hence only the required single field is updated.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/AbstractMonitor.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/AbstractMonitor.java
@@ -42,7 +42,7 @@ public abstract class AbstractMonitor<M extends Monitored> implements Monitor<M>
                 monitored.setLastEvaluated(System.currentTimeMillis());
                 save(monitored);
             } catch (RejectedExecutionException ignore) {
-
+                LOGGER.info("Error in processing monitor: {}", monitored, ignore);
             }
         }
     }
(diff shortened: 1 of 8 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/AbstractMonitor.java
```diff
@@ -64,7 +64,7 @@ protected EvaluatorExecutor getEvaluatorExecutorBean(M monitored) {
 
     protected abstract List<M> getMonitored();
 
-    protected abstract M save(M monitored);
+    protected abstract void save(M monitored);
```
smaniraju, line 68: The returned object is unused and avoid additional DB Hit in ClusterMonitor.

sidseth: Will this change affect HA deployments, and failover?

sidseth (commented):
One question about HA and failover. Other than that, looks good.

smaniraju:
> One question about HA and failover. Other than that, looks good.

No it wont affect HA\failover, the returned object is not used anywhere. failover is handled in a different flow in LeaderElectionService.

sidseth (approved):

PR #8932, opened 2020-09-04
CB-8643 Stabilize DistroXAutoScaleClusterV1EndpointTest

smaniraju:
1. Alert Monitors are disabled in UnitTest Context where there isn't any external service connectivity.
2. Tests should assert response of the invoked api.


smaniraju:
In some Jenkins CI runs when jobs are running very slowly, test appears flaky because of background processing by AlertMonitors.  Monitors should not be enabled in unit test context since external service connectivity is not available.

sidseth (approved):

PR #8934, opened 2020-09-04
CB-8670 Introduce withSkipOnFail to TestDTO cleanup

aszegedi:
- Introduce `withSkipOnFail(false))` to all the TestDtos' cleanUp
  - Along with the previous change introduce `testDto.setResponse` to all the related DeleteActions
- Update and standardised all the TestDtos' not implemented cleanUp with: `LOGGER.warn(String.format("Cleanup WARN: 'cleanUp' is not implemented for TestDto: %s with name: %s", this, getName()));` 
- Make `deletedResources` MultiValueMap CleanupUtil class global
- Some additional changes for Integration Tests:
  - FlowUtil: `Thread.currentThread().interrupt()`
  - EnvironmentChildTest: `checkEnvsAreNotListedByName`

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/connector/PlatformAccessConfigsTestDto.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/connector/PlatformAccessConfigsTestDto.java
@@ -4,7 +4,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.sequenceiq.environment.api.v1.platformresource.model.PlatformAccessConfigsResponse;
-import com.sequenceiq.it.cloudbreak.CloudbreakClient;
 import com.sequenceiq.it.cloudbreak.Prototype;
 import com.sequenceiq.it.cloudbreak.context.TestContext;
 import com.sequenceiq.it.cloudbreak.dto.AbstractCloudbreakTestDto;
@@ -68,11 +67,6 @@ public class PlatformAccessConfigsTestDto extends AbstractCloudbreakTestDto<Obje
     public PlatformAccessConfigsTestDto withAvailabilityZone(String availabilityZone) {
         this.availabilityZone = availabilityZone;
         return this;
-    }
-
-    @Override
-    public void cleanUp(TestContext context, CloudbreakClient cloudbreakClient) {
-        LOGGER.debug("this entry point does not have any clean up operation");
     }
 
     @Override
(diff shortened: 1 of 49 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/connector/PlatformAccessConfigsTestDto.java
```diff
@@ -72,7 +72,7 @@ public PlatformAccessConfigsTestDto withAvailabilityZone(String availabilityZone
 
     @Override
     public void cleanUp(TestContext context, CloudbreakClient cloudbreakClient) {
-        LOGGER.debug("this entry point does not have any clean up operation");
+        LOGGER.warn(String.format("Cleanup WARN: 'cleanUp' is not implemented for TestDto: %s with name: %s", this, getName()));
```
Bajzathd, line 75: I think in this case we should simply not override the default cleanUp

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/EnvironmentChildTest.java
```diff
@@ -231,4 +238,21 @@ private boolean parentNameEquals(EnvironmentTestDto environment, SimpleEnvironme
...
+            Collection<SimpleEnvironmentResponse> simpleEnvironmentV4Respons = environmentTestDto.getResponseSimpleEnvSet();
+            if (isNull(simpleEnvironmentV4Respons)) {
+                throw new TestFailException("Environment list response is missing.");
+            }
+            List<String> listedEnvironmentNames = simpleEnvironmentV4Respons.stream()
+                    .filter(response -> !response.getEnvironmentStatus().equals(EnvironmentStatus.ARCHIVED))
+                    .map(EnvironmentBaseResponse::getName)
+                    .collect(Collectors.toList());
+            if (listedEnvironmentNames.stream().anyMatch(environmentNames::contains)) {
+                throw new TestFailException("Environment is listed");
```
Bajzathd, line 253: From the message we can not tell which env was listed - the parent or the child.
I have also refactored this test case in my PR, see here: https://github.com/hortonworks/cloudbreak/pull/8923/files#diff-68f3ed9be95a1bffc91d2b7ec253c58c
I think we should merge these somehow here

aszegedi: I think we should merge your PR first, after that rebase this PR onto master with your changes...

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/connector/PlatformAccessConfigsTestDto.java
aszegedi: So we should remove all the `cleanUp' is not implemented` `Override` from the project?

Bajzathd: I think yes, and move this message to the default implementation. This way we will have less code, and truly standardized cleanUp

Bajzathd (approved):

PR #8936, opened 2020-09-07
CB-8641 Include DistroX user tags on DistroX RDS

bergerdenes:
And do the same for SDX user tags (for future compatibility). For this SDX service's SdxClusterResponse had to be extended also
as it did not return the Tags property.
Redbeams spring config was incorrect regarding Cloudbreak endpoint configuration values.
Redbeams also had to be linked to SDX service.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/provision/handler/CreateExternalDatabaseHandler.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/provision/handler/CreateExternalDatabaseHandler.java
@@ -64,9 +64,9 @@ public class CreateExternalDatabaseHandler implements EventHandler<CreateExterna
         LOGGER.debug("External database: {} for stack {}", externalDatabase.name(), stack.getName());
         Selectable result;
         try {
+            String resourceCrn = null;
             if (externalDatabase == DatabaseAvailabilityType.NONE) {
                 LOGGER.info("External database for stack {} is not requested.", stack.getName());
-                result = new CreateExternalDatabaseResult(stack.getId(), EXTERNAL_DATABASE_WAIT_SUCCESS_EVENT.event(), stack.getName(), null);
             } else {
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/provision/handler/CreateExternalDatabaseHandler.java
```diff
@@ -80,9 +80,9 @@ public void accept(Event<CreateExternalDatabaseRequest> createExternalDatabaseRe
                         stack.getName(), stack.getStatus().name(), DetailedStackStatus.PROVISION_REQUESTED.name());
                 stackUpdaterService.updateStatus(stack.getId(), DetailedStackStatus.PROVISION_REQUESTED,
                         ResourceEvent.CLUSTER_EXTERNAL_DATABASE_CREATION_FINISHED, "External database creation finished");
-                result = new CreateExternalDatabaseResult(stack.getId(), EXTERNAL_DATABASE_WAIT_SUCCESS_EVENT.event(),
-                        stack.getName(), stack.getCluster().getDatabaseServerCrn());
+                resourceCrn = stack.getCluster().getDatabaseServerCrn();
             }
+            result = new CreateExternalDatabaseResult(stack.getId(), EXTERNAL_DATABASE_WAIT_SUCCESS_EVENT.event(), stack.getName(), resourceCrn);
```
topolyai5, line 85: The resourceCrn can be null, is it ok?

bergerdenes: no, it's not ok, why would id be null?

bergerdenes: it is just a minor refactor, the functionality did not change

topolyai5 (approved):

PR #8937, opened 2020-09-07
CB-6644 Auditing environment

topolyai5:
The structured event moved to a different module, named as 'structuredevent-service-legacy'. In this modul the functionalities are untouched. For the CDP world, we created two new modules: 'structuredevent-api-cdp' and 'structuredevent-service-cdp'. We created a common endpoint, service and repository. Only one thing need to integrate: the domain SQL.


authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourcePermissionChecker.java
```diff
@@ -15,6 +16,9 @@
 
     protected Map<String, AuthorizationResourceAction> getAuthorizationActions(String resourceCrn, AuthorizationResourceAction action) {
         ResourceBasedCrnProvider resourceBasedCrnProvider = commonPermissionCheckingUtils.getResourceBasedCrnProvider(action);
+        if (resourceBasedCrnProvider == null) {
+            return Collections.emptyMap();
+        }
```
lacikaaa, line 21: could you add an else branch or restructure the code so you won't have a return in the middle of the method?

common/src/main/java/com/sequenceiq/cloudbreak/common/json/Json.java
```diff
@@ -170,4 +178,28 @@ public String toString() {
...
+            return false;
+        }
+    }
+
+    public boolean isArray() {
+        try {
+            JSONArray.fromObject(value);
+            return true;
+        } catch (JSONException e) {
+            LOGGER.debug("This json is not an object: {}", e.getMessage());
```
lacikaaa, line 197: object->array

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/BaseLegacyStructuredFlowEventFactory.java
```diff
@@ -26,18 +26,18 @@
...
 import com.sequenceiq.cloudbreak.structuredevent.event.StackDetails;
 import com.sequenceiq.cloudbreak.structuredevent.event.StructuredFlowEvent;
 import com.sequenceiq.cloudbreak.structuredevent.event.StructuredNotificationEvent;
-import com.sequenceiq.cloudbreak.structuredevent.rest.StructuredFlowEventFactory;
+import com.sequenceiq.cloudbreak.structuredevent.event.legacy.OperationDetails;
+import com.sequenceiq.cloudbreak.structuredevent.rest.LegacyStructuredFlowEventFactory;
 import com.sequenceiq.flow.ha.NodeConfig;
 
 @Component
 @Transactional
```
lacikaaa, line 37: could you check if we really need this transactional annotation?

environment/src/main/java/com/sequenceiq/environment/environment/domain/Environment.java
```diff
@@ -156,6 +157,11 @@ public void setAccountId(String accountId) {
         this.accountId = accountId;
     }
 
+    @Override
+    public String getResourceName() {
+        return null;
```
lacikaaa, line 162: why null?

environment/src/main/java/com/sequenceiq/environment/events/CredentialUrlParser.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.structuredevent.event.CloudbreakEventService;
+import com.sequenceiq.cloudbreak.structuredevent.rest.urlparser.CDPRestUrlParser;
+
+@Component
+public class CredentialUrlParser extends CDPRestUrlParser {
```
lacikaaa, line 13: missing unit test, especially for pattern matching

environment/src/main/java/com/sequenceiq/environment/events/EnvironmentStructuredFlowEventFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    @Inject
+    private NodeConfig nodeConfig;
+
+    @Value("${info.app.version:}")
+    private String serviceVersion;
+
+    @Override
+    public CDPStructuredFlowEvent createStructuredFlowEvent(Long resourceId, FlowDetails flowDetails, Boolean detailed) {
+        Optional<Environment> environmentOptional = environmentService.findEnvironmentById(resourceId);
+        Environment environment = environmentOptional.orElseThrow(notFound("Environment", resourceId));
```
lacikaaa, line 41: I would prefer if you would add an `getEnvironmentById` method to the `EnvironmentService` and that could throw an exception as I see this kind of usage a lot in the code generally

environment/src/main/java/com/sequenceiq/environment/events/EnvironmentStructuredFlowEventFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public CDPStructuredFlowEvent createStructuredFlowEvent(Long resourceId, FlowDetails flowDetails, Boolean detailed, Exception exception) {
+        return null;
+    }
+
+    @Override
+    public CDPStructuredNotificationEvent createStructuredNotificationEvent(Long resourceId, String notificationType, String message, String groupName) {
+        return null;
+    }
```
lacikaaa, line 57: I see you have created `CDPStructuredFlowEventFactory` and this is the only implementation of that interface. And 2 of 3 methods return null. Why do we need these methods?

environment/src/main/java/com/sequenceiq/environment/events/EventSenderService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private CDPStructuredNotificationDetails getNotificationDetails(ResourceEvent resourceEvent, String resourceCrn, String resourceType,
+            Object payload) {
+        String serializedPayload;
+        try {
+            serializedPayload = new Gson().toJson(payload);
+            LOGGER.debug("CDPStructuredNotificationDetails' payload has been serialized with ResourceEvent[{}], resource type[{}], CRN[{}]",
+                    resourceEvent.name(), resourceType, resourceCrn);
+        } catch (RuntimeException re) {
+            String msg = String.format("CDPStructuredNotificationDetails' payload couldn't be serialized with ResourceEvent[%s], resource type[%s], CRN[%s]",
```
lacikaaa, line 117: I see no reason to introduce `msg` here, you could just use `serializedPayload`

structuredevent-api-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/annotation/TransformGetterType.java
```diff
@@ -0,0 +1,13 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface TransformGetterType {
```
lacikaaa, line 12: there is already a `TransformGetterType` in the codebase, is it possible to reuse that one? same applies for `Immutable`

structuredevent-api-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/model/CDPEventV1Response.java
```diff
@@ -0,0 +1,183 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.sequenceiq.cloudbreak.structuredevent.event.LdapDetails;
+import com.sequenceiq.cloudbreak.structuredevent.event.RdsDetails;
+
+import io.swagger.annotations.ApiModel;
+
+@ApiModel
+@JsonInclude(Include.NON_NULL)
+public class CDPEventV1Response extends CDPEventBaseV1 {
```
lacikaaa, line 12: maybe add json ignore unknown annotation also

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPOperationDetails.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public Long getTimestamp() {
+        return timestamp;
+    }
+
+    public void setTimestamp(Long timestamp) {
+        this.timestamp = timestamp;
+    }
+
+    public String getUTCDateTime() {
```
lacikaaa, line 104: this might worth a simple unit test

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredEventContainer.java
```diff
@@ -0,0 +1,52 @@
...
+        List<CDPStructuredFlowEvent> flow,
+        List<CDPStructuredRestCallEvent> rest,
+        List<CDPStructuredNotificationEvent> notification) {
+        this.flow = flow;
+        this.rest = rest;
+        this.notification = notification;
+    }
+
+    public List<CDPStructuredFlowEvent> getFlow() {
+        return flow != null ? flow : Collections.emptyList();
```
lacikaaa, line 30: why not initialize the field with a `Collections.emptyList()` and then the getter won't have any logic. At least I think we usually handle this way. same applies to the others

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredFlowEvent.java
```diff
@@ -0,0 +1,73 @@
...
+        String state = flow.getFlowState();
+//        if ("unknown".equals(state)) {
+//            state = getOperation().getResourceType().toUpperCase()
+//            if (stack != null) {
+//                state = "STACK_" + stack.getDetailedStatus();
+//            }
+//            if (cluster != null) {
+//                state += " | CLUSTER_" + cluster.getStatus();
+//            }
+//        }
```
lacikaaa, line 41: do we need these or could be deleted? if can be then you should return in the first line and you don't need the state variaable

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredFlowEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.sequenceiq.cloudbreak.structuredevent.event.FlowDetails;
+import com.sequenceiq.cloudbreak.structuredevent.json.Base64Deserializer;
+import com.sequenceiq.cloudbreak.structuredevent.json.Base64Serializer;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CDPStructuredFlowEvent<T> extends CDPStructuredEvent {
```
lacikaaa, line 11: I think `T` should be serializable

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredFlowEvent.java
```diff
@@ -0,0 +1,73 @@
...
+//            if (cluster != null) {
+//                state += " | CLUSTER_" + cluster.getStatus();
+//            }
+//        }
+        return state;
+    }
+
+    @Override
+    public Long getDuration() {
+        return Math.max(0L, flow.getDuration());
```
lacikaaa, line 47: why do we need this max? duration can be negative?

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredNotificationEvent.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public String getStatus() {
+        return "SENT";
+    }
+
+    @Override
+    public Long getDuration() {
+        return 0L;
+    }
```
lacikaaa, line 47: these 2 could be static final variable

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredNotificationEvent.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public CDPStructuredNotificationEvent() {
+        super(CDPStructuredNotificationEvent.class.getSimpleName());
+    }
+
+    public CDPStructuredNotificationEvent(CDPOperationDetails operation, CDPStructuredNotificationDetails notificationDetails) {
+        super(CDPStructuredNotificationEvent.class.getSimpleName(), operation);
+        this.notificationDetails = notificationDetails;
+        ldapNotificationDetails = null;
+        rdsNotificationDetails = null;
```
lacikaaa, line 24: why these null initializations required? the same for those in the other constructors

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredRestCallEvent.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public CDPStructuredRestCallEvent(CDPOperationDetails operation, RestCallDetails restCall) {
+        super(CDPStructuredRestCallEvent.class.getSimpleName(), operation);
+        this.restCall = restCall;
+    }
+
+    @Override
+    public String getStatus() {
+        return restCall.getRestResponse().getStatusText() != null
```
lacikaaa, line 21: are we sure restcall has always a non null restresponse? I don't see anything in `RestCallDetails` what would ensure that

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/config/CDPStructuredEventConfig.java
```diff
@@ -0,0 +1,7 @@
+package com.sequenceiq.cloudbreak.structuredevent.config;
+
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CDPStructuredEventConfig {
```
lacikaaa, line 6: do we need this empty class?

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/CDPRestCommonService.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        if (responseJson != null && StringUtils.isEmpty(id)) {
+                id = Optional.ofNullable(getValueFromJson(responseJson, path))
+                        .orElse(getListValue(request, responseJson, pluralPath, restParams, idType));
+        }
+        return id;
+    }
+
+    private String getListValue(RestRequestDetails restRequest, Json json, String path, Map<String, String> restParams, String idType) {
```
lacikaaa, line 83: it's a bit hard to follow this method, similarly to `getResourceId`. Could you refactor this a bit to make it more readable?

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/CDPRestCommonService.java
```diff
@@ -0,0 +1,129 @@
...
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.common.anonymizer.AnonymizerUtil;
+import com.sequenceiq.cloudbreak.common.json.Json;
+import com.sequenceiq.cloudbreak.common.json.JsonUtil;
+import com.sequenceiq.cloudbreak.structuredevent.event.rest.RestCallDetails;
+import com.sequenceiq.cloudbreak.structuredevent.event.rest.RestRequestDetails;
+
+@Component
+public class CDPRestCommonService {
```
lacikaaa, line 22: please add unit tests

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
```diff
@@ -0,0 +1,383 @@
...
+
+    private static final String CRN = "crn";
+
+    private final List<String> skippedHeadersList = Lists.newArrayList("authorization");
+
+    private final Map<String, AccountAwareResourceRepository<?, ?>> pathRepositoryMap = new HashMap<>();
+
+    private final Pattern extractIdRestParamFromResponsePattern = Pattern.compile("\"" + ID + "\":([0-9]*)");
+
+    private final Pattern extractCrnRestParamFromResponsePattern = Pattern.compile("\"" + CRN + "\":\"([0-9a-zA-Z:-]*)\"");
```
lacikaaa, line 97: not used anywhere

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
```diff
@@ -0,0 +1,383 @@
...
+                    MessageUtils.getCharset(context.getMediaType())).toString();
+            Map<String, String> restParams = (Map<String, String>) context.getProperty(REST_PARAMS);
+            if (restParams == null) {
+                restParams = new HashMap<>();
+            }
+            sendStructuredEvent(restRequest, restResponse, restParams, requestTime, responseBody);
+        }
+    }
+
+    private void sendStructuredEvent(RestRequestDetails restRequest, RestResponseDetails restResponse, Map<String, String> restParams, Long requestTime,
```
lacikaaa, line 203: could you break up this method? it too long and could be refactored

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
```diff
@@ -0,0 +1,383 @@
...
+            String responseBody) {
+        boolean valid = checkRestParams(restParams);
+        try {
+            if (!valid) {
+                LOGGER.debug("Cannot create structured event, because rest params are invalid.");
+                return;
+            }
```
lacikaaa, line 210: please don't return in the middle of the method, at least add an `else` branch

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
```diff
@@ -0,0 +1,383 @@
...
+import com.sequenceiq.cloudbreak.structuredevent.rest.CDPRestCommonService;
+import com.sequenceiq.cloudbreak.structuredevent.rest.annotation.AccountEntityType;
+import com.sequenceiq.cloudbreak.structuredevent.rest.urlparser.CDPRestUrlParser;
+import com.sequenceiq.cloudbreak.structuredevent.service.CDPBaseRestRequestThreadLocalService;
+import com.sequenceiq.cloudbreak.structuredevent.service.CDPDefaultStructuredEventClient;
+import com.sequenceiq.cloudbreak.structuredevent.service.lookup.CDPAccountAwareRepositoryLookupService;
+import com.sequenceiq.flow.ha.NodeConfig;
+
+@Component
+public class CDPStructuredEventFilter implements WriterInterceptor, ContainerRequestFilter, ContainerResponseFilter {
```
lacikaaa, line 69: this is a bit complicated class with a lot of logic and hard to follow/understand. no one would dare to touch it if you don't cover this with unit tests pretty well.
also some refactor would be nice, the high number of `Autowires` indicate the same also that maybe the whole class could be refactored into smaller ones which is easier to understand, maintain and test

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/AbstractAccountAwareResourceService.java
```diff
@@ -0,0 +1,156 @@
...
+import com.sequenceiq.cloudbreak.structuredevent.repository.AccountAwareResource;
+import com.sequenceiq.cloudbreak.structuredevent.repository.AccountAwareResourceRepository;
+
+public abstract class AbstractAccountAwareResourceService<T extends AccountAwareResource>
+        implements AccountAwareResourceService<T> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractAccountAwareResourceService.class);
+
+    @Override
+    public T create(T resource, @Nonnull String accountId) {
```
lacikaaa, line 28: maybe `@NotEmpty` would fit better. Should we add this to the other public methods definition also?

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
biharitomi: Hi, yes, it makes sense, but this logic came from the core module. Good idea to refactor it, but it could be a separated task/improvement. Also my experience with this class that dependencies could not be injected into Jersey managed filters with `Inject` annotation and the only way that worked for me leave the `Autowire`s. I only could dig in a little bit, but in case of `Inject` Jersey tries to use it's own dependency management mechanism to resolve dependencies which are configured in Spring related configurations and managed by Spring.

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/CDPFlowStructuredEventHandler.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.structuredevent.event.FlowDetails;
+import com.sequenceiq.cloudbreak.structuredevent.event.cdp.CDPStructuredEvent;
+import com.sequenceiq.flow.core.FlowEventListener;
+
+@Primary
+@Component
+@Scope("prototype")
+public class CDPFlowStructuredEventHandler<S, E> extends StateMachineListenerAdapter<S, E> implements FlowEventListener<S, E> {
```
lacikaaa, line 24: please add unit tests

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/converter/CDPStructuredEventEntityToCDPStructuredEventConverter.java
```diff
@@ -0,0 +1,53 @@
...
+import com.sequenceiq.cloudbreak.common.json.JsonUtil;
+import com.sequenceiq.cloudbreak.converter.AbstractConversionServiceAwareConverter;
+import com.sequenceiq.cloudbreak.structuredevent.domain.CDPStructuredEventEntity;
+import com.sequenceiq.cloudbreak.structuredevent.event.cdp.CDPStructuredEvent;
+import com.sequenceiq.cloudbreak.structuredevent.event.cdp.CDPStructuredFlowEvent;
+import com.sequenceiq.cloudbreak.structuredevent.event.cdp.CDPStructuredNotificationEvent;
+import com.sequenceiq.cloudbreak.structuredevent.event.cdp.CDPStructuredRestCallEvent;
+
+@Component
+public class CDPStructuredEventEntityToCDPStructuredEventConverter
```
lacikaaa, line 23: unit test

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/converter/CDPStructuredEventToCDPStructuredEventEntityConverter.java
```diff
@@ -0,0 +1,36 @@
...
+            structuredEventEntity.setStructuredEventJson(new Json(source));
+            CDPOperationDetails operationDetails = source.getOperation();
+            structuredEventEntity.setEventType(operationDetails.getEventType());
+            structuredEventEntity.setResourceType(operationDetails.getResourceType());
+            structuredEventEntity.setResourceCrn(operationDetails.getResourceCrn());
+            structuredEventEntity.setTimestamp(operationDetails.getTimestamp());
+            structuredEventEntity.setAccountId(source.getOperation().getAccountId());
+            return structuredEventEntity;
+        } catch (IllegalArgumentException e) {
+            LOGGER.error("Failed to parse structured event JSON", e);
```
lacikaaa, line 32: shall we log the source to know what we can't convert?

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/lookup/CDPRepositoryLookupService.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.annotation.PostConstruct;
+
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.util.CollectionUtils;
+
+import com.sequenceiq.cloudbreak.workspace.repository.EntityType;
+
+public abstract class CDPRepositoryLookupService<T extends CrudRepository<?, ?>> {
+
+    private List<T> repositoryList;
```
lacikaaa, line 17: this field is unnecessary and misleading. could you remove this?

structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/config/LegacyStructuredEventConfig.java
```diff
@@ -0,0 +1,8 @@
+package com.sequenceiq.cloudbreak.structuredevent.config;
+
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class LegacyStructuredEventConfig {
```
lacikaaa, line 6: do we need this class?

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
lacikaaa: I don't have any issues with `Autowire` :)
I'm fine with fixing this in a followup pr, just please create a jira for that, otherwise I'm afraid this will stay the same way

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/BaseLegacyStructuredFlowEventFactory.java
topolyai5: Yes, it is necessary.

topolyai5: LazyInit exception occurred during the event firing

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredRestCallEvent.java
topolyai5: yes, here: `CDPStructuredEventFilter#326-335`

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
topolyai5: Here is the jira for tracking this: https://jira.cloudera.com/browse/CB-8777

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/filter/CDPStructuredEventFilter.java
topolyai5: I will refactor this `return;` in the follow-up jira.

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredFlowEvent.java
topolyai5: I don't want to add the `Serializable` to the generic type. Sometimes generate unnecessary changes/refactors.

authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourcePermissionChecker.java
topolyai5: I think it is much clearer then if I create an if {} else {} structure. Otherwise, we have a lot of similar code style in the codebase.

structuredevent-model/src/main/java/com/sequenceiq/cloudbreak/structuredevent/event/cdp/CDPStructuredFlowEvent.java
lacikaaa: but `CDPStructuredEvent` is serializable and a class is only serializable if all the fields are serializable or transient. In this case you don't have to modify anything else as I have checked and the constructor is used only at one place where payload is set to `null`

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/CDPFlowStructuredEventHandler.java
topolyai5: Hm, I tried to write any useful test, but I cannot. Could you please give me any suggestion?

lacikaaa: - in `public void transition(Transition<S, E> transition)` a `CDPStructuredEvent` is sent. in this meethod there are a lot of ternary operator and an if also. depending on the `transition` parameter one of the `cdpStructuredFlowEventFactory.createStructuredFlowEvent` is called with different values which could be asserted. 
you can even check the in case an error like nullpointer it is caught, so this would ensure no one removes the try catch by mistake.
- in `public void stateMachineStopped(StateMachine<S, E> stateMachine)` you could test it has only  effect if it's not complete. it also has some ternary operator, those effects could be tested also.

In the meanwhile: why do we have try-catch in `transition` and not in `stateMachineStopped`?

topolyai5: _In the meanwhile: why do we have try-catch in transition and not in stateMachineStopped?_
Good questions, these codes are 2-3 years old.

topolyai5: I don't agree with you. These tests could not catch any error. I cannot validate the content of the payload. It would be much useful than verifying the methods. 
But I will write these tests, maybe

lacikaaa (approved):

PR #8941, opened 2020-09-07
CB-8629 [ASRG] single RG and custom network: single RG is deleted whe…

gergopapi2:
…n network is deleted

On azure when deleting a network created by cloudbreak the resource group of the network is deleted. In case of single RG the network and deployment resources themselves should be deleted instead of the resource group.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureNetworkConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureNetworkConnector.java
@@ -112,23 +112,47 @@ public class AzureNetworkConnector implements NetworkConnector {
 
     @Override
     public void deleteNetworkWithSubnets(NetworkDeletionRequest networkDeletionRequest) {
-        if (!networkDeletionRequest.isExisting()) {
-            try {
-                AzureClient azureClient = azureClientService.getClient(networkDeletionRequest.getCloudCredential());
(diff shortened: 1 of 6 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureNetworkConnector.java
```diff
@@ -113,21 +113,46 @@ public CreatedCloudNetwork createNetworkWithSubnets(NetworkCreationRequest netwo
...
+                azureClient.deleteResourceGroup(networkDeletionRequest.getResourceGroup());
+            }
+        } catch (CloudException e) {
+            LOGGER.warn("Deletion error, cloud exception happened: ", e);
+            if (e.body() != null && e.body().details() != null) {
+                String details = e.body().details().stream().map(CloudError::message).collect(Collectors.joining(", "));
+                throw new CloudConnectorException(String.format("Network deletion failed, status code %s, error message: %s, details: %s",
+                        e.body().code(), e.body().message(), details));
+            } else {
+                throw new CloudConnectorException(String.format("Network deletion failed: '%s', please go to Azure Portal for detailed message", e));
```
pdarvasi, line 155: This is the third occurrence of this fragment in this class, let alone the others, could you pls extract this logic?

keyki:
This should go to CB-2.30.0

pdarvasi (approved):

PR #8945, opened 2020-09-08
CB-8605 - New SINGLE option to use Dedicated Storage Account

pdarvasi:
This change introduces a SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT option for Azure resourceGroup request which calculates the storage account and resource group names for the VHD as it used to be in the legacy "MULTIPLE" RG use-case.

The intent is to use this internally (mow-dev) to save the time and cost of many image copies.

The feature is behind entitlement CDP_AZURE_SINGLE_RESOURCE_GROUP_DEDICATED_STORAGE_ACCOUNT and also CDP_AZURE_SINGLE_RESOURCE_GROUP entitlement has to be turned on, too.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/ResourceGroupUsage.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/ResourceGroupUsage.java
@@ -2,5 +2,6 @@ package com.sequenceiq.cloudbreak.cloud.azure;
 
 public enum ResourceGroupUsage {
     SINGLE,
-    MULTIPLE
+    MULTIPLE,
+    SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT
 }
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
@@ -10,6 +10,7 @@ import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.BadRequestException;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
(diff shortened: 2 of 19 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/ResourceGroupUsage.java
```diff
@@ -2,5 +2,6 @@
 
 public enum ResourceGroupUsage {
     SINGLE,
-    MULTIPLE
+    MULTIPLE,
+    SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT
```
gergopapi2, line 6: I would add a method as isSingleResourceGroup(), as in some cases the flavor of the single RG is not important.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
```diff
@@ -234,7 +235,9 @@ public UpgradeV4Response checkForClusterUpgrade(@NotNull NameOrCrn nameOrCrn, Lo
             boolean osUpgrade = upgradeService.isOsUpgrade(request);
             UpgradeV4Response upgradeResponse = clusterUpgradeAvailabilityService.checkForUpgradesByName(workspaceId, stackName,
                     osUpgrade);
-            clusterUpgradeAvailabilityService.filterUpgradeOptions(upgradeResponse, request);
+            if (CollectionUtils.isNotEmpty(upgradeResponse.getUpgradeCandidates())) {
+                clusterUpgradeAvailabilityService.filterUpgradeOptions(upgradeResponse, request);
+            }
```
gergopapi2, line 240: I am not sure this is related to the rest of the changes. If not, it might make sense to extract it to a separate PR.

mock-thunderhead/src/main/resources/application.yml
```diff
@@ -22,6 +22,7 @@ auth:
     raz.enable: true
     freeipadlebsencryption.enable: true
     azure.single.resourcegroup.enable: false
+    azure.single.resourcegroup.dedicated.storage.account.enable: false
```
gergopapi2, line 25: It might make sense to merge it with the above line

gergopapi2 (approved):
LGTM with minor changes. Those can go into a follow-up jira if needed.

PR #8950, opened 2020-09-09
CB-8739 Consistent Hostnames Enhancements

smaniraju:
Node name generation logic enhanced so that all  hostgroup's nodes start with index 0.

Testing done.
- Tested launch of new cluster with 0 based node indexes for every hostgroup.
- Tested scale up and scale down of hostgroup nodes with hostGroup based node index.
- Tested deleted on provider node and new node added with same name on repair.
- Tested manually deleted nodes and successful scale up.
- Tested scale up\down on existing clusters with privateId based names.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
@@ -9,11 +9,14 @@ import static org.apache.commons.lang3.StringUtils.isNoneBlank;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -90,6 +93,9 @@ public class ClusterBootstrapper {
 
     @Inject
     private HostOrchestrator hostOrchestrator;
+
+    @Inject
+    private ClusterNodeNameGenerator clusterNodeNameGenerator;
 
     @Inject
     private GatewayConfigService gatewayConfigService;
@@ -244,12 +250,16 @@ public class ClusterBootstrapper {
     private Set<Node> collectNodesForBootstrap(Stack stack) {
         Set<Node> nodes = new HashSet<>();
         String domain = hostDiscoveryService.determineDomain(stack.getCustomDomain(), stack.getName(), stack.isClusterNameAsSubdomain());
+
+        Map<String, AtomicLong> hostGroupNodeIndexes = new HashMap<>();
+        Set<String> clusterNodeNames = stack.getNotTerminatedInstanceMetaDataList().stream()
+                .map(InstanceMetaData::getShortHostname).collect(Collectors.toSet());
+
         for (InstanceMetaData im : stack.getNotDeletedInstanceMetaDataSet()) {
             if (im.getPrivateIp() == null && im.getPublicIpWrapper() == null) {
                 LOGGER.debug("Skipping instance metadata because the public ip and private ips are null '{}'.", im);
             } else {
-                String generatedHostName = hostDiscoveryService.calculateHostname(stack.getCustomHostname(), im.getShortHostname(), im.getInstanceGroupName(),
-                        im.getPrivateId(), stack.isHostgroupNameAsHostname());
+                String generatedHostName = clusterNodeNameGenerator.getNodeNameForInstanceMetadata(im, stack, hostGroupNodeIndexes, clusterNodeNames);
                 String instanceId = im.getInstanceId();
                 String instanceType = im.getInstanceGroup().getTemplate().getInstanceType();
                 nodes.add(new Node(im.getPrivateIp(), im.getPublicIpWrapper(), instanceId, instanceType, generatedHostName, domain, im.getInstanceGroupName()));
@@ -269,9 +279,13 @@ public class ClusterBootstrapper {
                 .collect(Collectors.toSet());
         String clusterDomain = getClusterDomain(metaDataSet, stack.getCustomDomain());
 
+        Map<String, AtomicLong> hostGroupNodeIndexes = new HashMap<>();
+        Set<String> clusterNodeNames = stack.getNotTerminatedInstanceMetaDataList().stream()
+                .map(InstanceMetaData::getShortHostname).collect(Collectors.toSet());
+
         Iterator<String> iterator = recoveryHostNames.iterator();
         for (InstanceMetaData im : metaDataSet) {
-            Node node = createNode(stack.getCustomHostname(), im, clusterDomain, stack.isHostgroupNameAsHostname());
+            Node node = createNode(stack, im, clusterDomain, hostGroupNodeIndexes, clusterNodeNames);
             if (upscaleCandidateAddresses.contains(im.getPrivateIp())) {
                 // use the hostname of the node we're recovering instead of generating a new one
                 // but only when we would have generated a hostname, otherwise use the cloud provider's default mechanism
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -361,4 +373,21 @@ private void validatePollingResultForCancellation(PollingResult pollingResult, S
         }
     }
 
+    String getNodeNameForInstance(InstanceMetaData im, Stack stack,
```
lacikaaa, line 376: this should be private or moved to a separate class

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -361,4 +373,21 @@ private void validatePollingResultForCancellation(PollingResult pollingResult, S
...
+            Map<String, AtomicLong> hostGroupNodeIndex, Set<String> clusterNodeNames) {
```
lacikaaa, line 377: unnecessary line break

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -361,4 +373,21 @@ private void validatePollingResultForCancellation(PollingResult pollingResult, S
...
+        if (isNotBlank(im.getShortHostname())) {
+            return im.getShortHostname();
+        }
```
lacikaaa, line 380: missing else to indicate there are multiple exit point for the method

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -361,4 +373,21 @@ private void validatePollingResultForCancellation(PollingResult pollingResult, S
...
+        AtomicLong hostGroupIndex = hostGroupNodeIndex.computeIfAbsent(im.getInstanceGroup().getGroupName(),
+                instantGroup -> new AtomicLong(0));
```
lacikaaa, line 382: `0L` instead of integer `0`

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -361,4 +373,21 @@ private void validatePollingResultForCancellation(PollingResult pollingResult, S
...
+        String generatedHostName;
+        do {
+            generatedHostName = hostDiscoveryService.calculateHostname(stack.getCustomHostname(), im.getShortHostname(),
+                    im.getInstanceGroupName(), hostGroupIndex.getAndIncrement(), stack.isHostgroupNameAsHostname());
+        } while (clusterNodeNames.contains(generatedHostName));
```
lacikaaa, line 387: this doesn't scale well at first blink, if I have a 300 node cluster in a single hostgroup and upscale with 50 node what would happen?

lacikaaa (changes requested):
have you tested:
- upscale
- downscale
- upscale again
- repair/upgrade?

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
smaniraju: ok

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
smaniraju: fixed

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
smaniraju: fixed.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
smaniraju: if 300 node cluster in hostgroup, hostGroupNodeIndex would be initialized to 300 before hitting this loop. So the first new node would get 301 index and subsequent nodes would get next sequential indexes. The existing clusterNodeNames check is failSafe to ensure no duplicate name is generated as that would fail the cluster launch\scaling.

smaniraju:
> have you tested:
> 
> * upscale
> * downscale
> * upscale again
> * repair/upgrade?

Yes tested various scenarios.
1. upscale, downscale, followed by multiple upscales\downscales.
2. Scale up, Scale down, Scale up, delete node on provider and repair.
3. Manual delete nodes from CB-UI, then scale up.
4. Same name was reused in repair of DH\DL nodes, which were deleted on provider.

lacikaaa:
I have tested some scenario, here are my findings:
First test:
- create a cluster with 200 compute node. Hostnames will be: compute0-199
- Downscale the first 100 nodes, hostnames will be: compute 100-199
- upscale with 2. Hostnames will be: compute 100-201

Second case:
- create a cluster with 10 compute nodes: compute 0-9
- downscale the first 6 nodes. remains: compute 6-9
- upscale with 5 nodes. result: compute 4-12

It looks like to me, the hostnames will be very "random" in these cases. If you put `0` into `hostGroupNodeIndexes` for every hostgroup at every upscale, then it would always start from 0 and would fill up the holes created during downscale. What's your opinion?

smaniraju:
> I have tested some scenario, here are my findings:
> First test:
> 
> * create a cluster with 200 compute node. Hostnames will be: compute0-199
> * Downscale the first 100 nodes, hostnames will be: compute 100-199
> * upscale with 2. Hostnames will be: compute 100-201
> 
> Second case:
> 
> * create a cluster with 10 compute nodes: compute 0-9
> * downscale the first 6 nodes. remains: compute 6-9
> * upscale with 5 nodes. result: compute 4-12
> 
> It looks like to me, the hostnames will be very "random" in these cases. If you put `0` into `hostGroupNodeIndexes` for every hostgroup at every upscale, then it would always start from 0 and would fill up the holes created during downscale. What's your opinion?

The test results above are similar to my test results.

Initially I started with hostGroupIndex 0 so that all the gaps would be filled. But to select one index for existing 100 node hostgroup and upscale of 5, 100 names has to be tested. For subsequent scale ups the number of names to be tested would keep on increasing starting from index 0. 

For auto-scaleable hostGroups there is no guarantee if a particular nodeName would always be available because any node can be decommissioned by autoscaler. Hence when Autoscaling is enabled, auto-scaleable hostGroup nodes cannot be used for direct node connectivity. If autoscaling is disabled, nodeNames would be consistent.

Considering this and ensuring that nodeIndex does not become an ever increasing number similar to privateId( within a hostGroup) since that would result in a index of 1000 after just 5 days of 0-200-0 Scheduled Daily Upscale\Downscale, I felt using a hostGroupIndex based on notTerminatedNodes is a good tradeoff between performance and also ensuring nodeName Index is not an ever increasing number.

keyki:
I don't get the performance impact of this change. If we start with 100 compute nodes and downscale 50 (let's assume it's the first 50 nodes). We would start from 50 and check if they exist and the next would be 51. If we start from 0 we have to do the same. But overall we're comparing strings so I don't see where is the performance issue. Comparing string should not be that expensive call pls with the current implementation there is no guarantee that starting from the number of alive instances will give us the best performance either. However, starting from 0 would give a better experience to always fill the gaps.

smaniraju:
> I don't get the performance impact of this change. If we start with 100 compute nodes and downscale 50 (let's assume it's the first 50 nodes). We would start from 50 and check if they exist and the next would be 51. If we start from 0 we have to do the same. But overall we're comparing strings so I don't see where is the performance issue. Comparing string should not be that expensive call pls with the current implementation there is no guarantee that starting from the number of alive instances will give us the best performance either. However, starting from 0 would give a better experience to always fill the gaps.

Ok, will update code to start from 0.

smaniraju:
Updated code to start with index 0 even during Autoscaling Scenarios, so that new node names would fill up the holes created during downscale

lacikaaa (approved):

PR #8962, opened 2020-09-09
CB-8536 added redbeams support to gcp

doktoric:

### DIFF
--- a/redbeams/src/main/java/com/sequenceiq/redbeams/service/network/NetworkParameterAdder.java
+++ b/redbeams/src/main/java/com/sequenceiq/redbeams/service/network/NetworkParameterAdder.java
@@ -27,16 +27,21 @@ public class NetworkParameterAdder {
     @VisibleForTesting
     static final String SUBNET_ID = "subnetId";
 
+    @VisibleForTesting
+    static final String AVAILABILITY_ZONE = "availabilityZone";
+
     // These constants must match those in AzureNetworkView
 
     @VisibleForTesting
     static final String SUBNETS = "subnets";
 
-    public Map<String, Object> addSubnetIds(Map<String, Object> parameters, List<String> subnetIds, CloudPlatform cloudPlatform) {
+    public Map<String, Object> addSubnetIds(Map<String, Object> parameters, List<String> subnetIds, List<String> azs, CloudPlatform cloudPlatform) {
         switch (cloudPlatform) {
             case AWS:
+            case GCP:
             case MOCK:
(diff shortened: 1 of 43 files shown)


redbeams/src/main/java/com/sequenceiq/redbeams/service/network/NetworkParameterAdder.java
```diff
@@ -27,16 +27,21 @@
...
     static final String SUBNETS = "subnets";
 
-    public Map<String, Object> addSubnetIds(Map<String, Object> parameters, List<String> subnetIds, CloudPlatform cloudPlatform) {
+    public Map<String, Object> addSubnetIds(Map<String, Object> parameters, List<String> subnetIds, List<String> azs, CloudPlatform cloudPlatform) {
         switch (cloudPlatform) {
             case AWS:
+            case GCP:
             case MOCK:
                 parameters.put(SUBNET_ID, String.join(",", subnetIds));
+                parameters.put(AVAILABILITY_ZONE, String.join(",", azs));
```
cegganesh84, line 44: Not pressing and also a thing to consider for other cloud vendors too...

`        // GCP VPCs are global. Subnets have a global scope in region. So picking the first availability zone in the region for subnet.
        String zone = compute.regions().get(projectId, region.value()).execute().getZones().stream().findFirst().orElse(null);
`

We chose the first AZs in the region list to host the compute resources. If they talk to the database hosted in multiple AZs and if the traffic goes to a different AZ then the customer incurs a cost of 1 US cent per GB. Is it possible to pick AZs for database is in such a way that it is co-located with compute for the primary and the secondary can be any other AZ?

doktoric: currently this service requires a subnet where the sql server needs to be deployed. I thought it would be nice to use the AZ of the selected subnet and then if we would like to adopt HA then just passing multiple subnet

cegganesh84: Yeah, sorry the pre-text of the question is in GCP a single subnet spans across all the AZs in that region (quite different from other vendors). So even if it is a single subnet given by the customer, in GCP traffic can go cross AZs. To prevent that in the GcpResourceBuilder, we deliberately choose the first subnet consistently to avoid deploying FreeIPA, DL and DH in three different AZs and thereby increasing the cost of the installation.

PR #8963, opened 2020-09-09
CB-8659 restart services after repair

sodre90:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
@@ -94,21 +94,33 @@ public class ClusterUpscaleService {
         recipeEngine.uploadUpscaleRecipes(stack, hostGroup, hostGroups);
     }
 
-    public void installServicesOnNewHosts(Long stackId, String hostGroupName) throws CloudbreakException {
+    public void installServicesOnNewHosts(Long stackId, String hostGroupName, Boolean repair, Boolean restartServices) throws CloudbreakException {
         Stack stack = stackService.getByIdWithClusterInTransaction(stackId);
-        LOGGER.debug("Start installing Ambari services");
+        LOGGER.debug("Start installing CM services");
         HostGroup hostGroup = Optional.ofNullable(hostGroupService.getByClusterIdAndNameWithRecipes(stack.getCluster().getId(), hostGroupName))
                 .orElseThrow(NotFoundException.notFound("hostgroup", hostGroupName));
         Set<InstanceMetaData> runningInstanceMetaDataSet = hostGroup.getInstanceGroup().getRunningInstanceMetaDataSet();
         recipeEngine.executePostAmbariStartRecipes(stack, hostGroup.getRecipes());
         ClusterApi connector = clusterApiConnectors.getConnector(stack);
         List<String> upscaledHosts = connector.upscaleCluster(hostGroup, runningInstanceMetaDataSet);
+        if (shouldRestartServices(repair, restartServices, stack)) {
+            try {
+                LOGGER.info("Trying to restart services");
+                connector.restartAll();
+            } catch (RuntimeException e) {
+                LOGGER.info("Restart services failed", e);
+            }
+        }
         runningInstanceMetaDataSet.stream()
                 .filter(instanceMetaData -> upscaledHosts.contains(instanceMetaData.getDiscoveryFQDN()))
                 .forEach(instanceMetaData -> {
                     instanceMetaData.setInstanceStatus(InstanceStatus.SERVICES_HEALTHY);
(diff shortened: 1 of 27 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
```diff
@@ -103,6 +103,14 @@ public void installServicesOnNewHosts(Long stackId, String hostGroupName) throws
         recipeEngine.executePostAmbariStartRecipes(stack, hostGroup.getRecipes());
         ClusterApi connector = clusterApiConnectors.getConnector(stack);
         List<String> upscaledHosts = connector.upscaleCluster(hostGroup, runningInstanceMetaDataSet);
+        if (repair) {
```
keyki, line 106: Is this going to work if both master and idbroker are down? I think after we replace the master we will try to restart the services which is going to fail.

sodre90: Yes, it will fail because CM will not be able to start every service, but it won't cause any problem. It will spend some time here to restart the node for sure which can be avoided. I can try to find out somehow if it is the last node we are trying to recover. Another solution can be if we check every other node are fine and if yes, then we can do a restart.

lnardai: How much is the "spend some time" in minutes? Do we need to consider this for e2e tests?

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/upscale/ClusterUpscaleContext.java
```diff
@@ -16,14 +16,20 @@
 
     private final ClusterManagerType clusterManagerType;
 
+    private final Boolean repair;
```
foldik, line 19: Why the wrapper type instead of the primitive?

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
```diff
@@ -94,15 +94,23 @@ public void uploadRecipesOnNewHosts(Long stackId, String hostGroupName) throws C
...
         HostGroup hostGroup = Optional.ofNullable(hostGroupService.getByClusterIdAndNameWithRecipes(stack.getCluster().getId(), hostGroupName))
                 .orElseThrow(NotFoundException.notFound("hostgroup", hostGroupName));
         Set<InstanceMetaData> runningInstanceMetaDataSet = hostGroup.getInstanceGroup().getRunningInstanceMetaDataSet();
         recipeEngine.executePostAmbariStartRecipes(stack, hostGroup.getRecipes());
         ClusterApi connector = clusterApiConnectors.getConnector(stack);
         List<String> upscaledHosts = connector.upscaleCluster(hostGroup, runningInstanceMetaDataSet);
+        if (shouldRestartServices(repair, restartServices, stack)) {
+            try {
+                LOGGER.info("Trying to restart services");
+                connector.restartAll();
```
foldik, line 109: How much time does it take to restart all the services? Does it make the repair a lot longer?

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
```diff
@@ -94,15 +94,23 @@ public void uploadRecipesOnNewHosts(Long stackId, String hostGroupName) throws C
...
+            } catch (RuntimeException e) {
+                LOGGER.info("Restart services failed", e);
```
foldik, line 111: What if all restartAll fail? Will we show the cluster as healthy or the flow will fail later?

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
sodre90: around 3-4 minutes

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterUpscaleService.java
sodre90: it will be healthy

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/upscale/ClusterUpscaleContext.java
sodre90: because singlePrimaryGateway was Boolean also :D

foldik (approved):
lgtm

PR #8964, opened 2020-09-09
CDPCP-2986. Include internal machine users in user sync

handavid:
Internal machine users are needed in user sync. This commit adds
the includeInternal flag to the listMachineUsers calls.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
@@ -43,6 +43,8 @@ public class UmsUsersStateProvider {
 
     private static final String ADMIN_FREEIPA_GROUP = "admins";
 
+    private static final boolean INCLUDE_INTERNAL_MACHINE_USERS = true;
+
     @Inject
     private GrpcUmsClient grpcUmsClient;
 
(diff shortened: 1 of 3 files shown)


gitmfox (approved):
LGTM. I don't think I have approval permission in this repo, but I morally approve :-)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UmsUsersStateProvider.java
```diff
@@ -43,6 +43,8 @@
 
     private static final String ADMIN_FREEIPA_GROUP = "admins";
 
+    private static final boolean INCLUDE_INTERNAL_MACHINE_USERS = true;
```
keyki, line 46: Don't we want to make this configurable?

handavid: We do not want to make this configurable. We always want to include internal machine users. Internal machine users may be identities within the environment. Right now, single user syncs that provide a machineUser crn can sync the user to the environment but then the users will be deleted by a full sync. This is incorrect behavior that we need to fix.

I don't think there are any internal machineUsers currently in use, but the DWX team is working on creating an internal machine user for their service and potentially the DEX team.

I'll update the commit message to include more details about the behavior we are fixing

handavid:
I updated the commit message to explain why the includeInternal flag is needed and always set to true in user sync.

keyki (approved):

keyki:
Thanks. LGTM

PR #8975, opened 2020-09-10
CB-8735 set CDP_ENV to PUBLIC_CLOUD when CM server starts rather than…

keyki:
… with API

Previously, we set this env var with API call after them CM server has started.
However, CM team would like to rely on this env var during CM startup so they can
initialise the application accordingly. Making this change in the cm.settings
file is not viable as the different initialisation processes do not use this file
and it would be too late for the CM server to use this file hence the -env server
argumentum.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
@@ -30,6 +30,18 @@ setup_missed_cm_heartbeat:
     - text: setsettings MISSED_HB_BAD {{ cloudera_manager.settings.missed_heartbeat_interval }}
     - unless: grep "MISSED_HB_BAD" /etc/cloudera-scm-server/cm.settings
 
+add_settings_file_to_cfm_server_args:
+  file.replace:
+    - name: /etc/default/cloudera-scm-server
+    - pattern: "CMF_SERVER_ARGS=.*"
+{% if salt['pillar.get']('cloudera-manager:settings:set_cdp_env') == True %}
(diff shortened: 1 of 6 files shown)


schfeca75 (approved):

lacikaaa (approved):

keyki:
@lacikaaa I had to change the implementation to use `-env` as CM server arg rather than the settings file. Please re-review the changes.

orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
```diff
@@ -30,6 +30,17 @@ setup_missed_cm_heartbeat:
...
+add_settings_file_to_cfm_server_args:
+  file.replace:
+    - name: /etc/default/cloudera-scm-server
+    - pattern: "CMF_SERVER_ARGS=.*"
+{% if salt['pillar.get']('cloudera-manager:settings:set_cdp_env') == True %}
+    - repl: CMF_SERVER_ARGS="-i /etc/cloudera-scm-server/cm.settings -env PUBLIC_CLOUD"
+{% else %}
+    - repl: CMF_SERVER_ARGS="-i /etc/cloudera-scm-server/cm.settings"
+{% endif %}
+    - unless: grep "CMF_SERVER_ARGS=\"-i /etc/cloudera-scm-server/cm.settings\"" /etc/default/cloudera-scm-server
```
lacikaaa, line 42: this won't match `"-i /etc/cloudera-scm-server/cm.settings -env PUBLIC_CLOUD"`

lacikaaa (changes requested):

lacikaaa (approved):

lacikaaa (approved):

PR #8979, opened 2020-09-10
CB-8726 Introduce Cloudera Runtime 7.2.3 support.

akanto:
It turned out that L0 tests are depending on the default templates, therefore we need to introduce these templates in 2.29.0

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -124,7 +124,7 @@ cb:
     url: http://localhost:8087
     contextPath: /redbeams
   runtimes:
-    latest: "7.2.2"
+    latest: "7.2.3"
 
   host.discovery.custom.hostname.enabled: false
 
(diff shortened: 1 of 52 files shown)


core/src/main/resources/application.yml
```diff
@@ -321,6 +321,21 @@ cb:
         7.2.2 - Data Discovery and Exploration=cdp-dde-722;
         7.2.2 - Streaming Analytics Light Duty with Apache Flink=cdp-flink-light-722;
         7.2.2 - Streaming Analytics Heavy Duty with Apache Flink=cdp-flink-heavy-722;
+        7.2.3 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-722;
```
Bajzathd, line 324: I guess these should end with `-723`

akanto: Thanks good catch

shwethags (approved):

PR #8980, opened 2020-09-10
CB-8744 enable Azure Single Resource Group feature locally (UMS mock)

keyki:
Azure Single Resource Group feature is behind an Entitlement and up until
now it was disabled. In order to enable it we had to extend the feature with
an additional 'resourceGroupUsage' type to use dedicated Storage Accounts
for the images. Without this feature each and every person would copy our images
into their personal Storage Account that would generate a lot of cost for Cloudera.
Given the above we're granting this additional Entitlement locally so everyone
would use the same SA for the images as it would be a multi RG setup. This way
all of their resources will go to the selected RG except the images which will
be stored in a common SA generated per Azure subscription.

### DIFF
--- a/mock-thunderhead/src/main/resources/application.yml
+++ b/mock-thunderhead/src/main/resources/application.yml
@@ -21,8 +21,8 @@ auth:
     sshpublickey.file: key.pub
     raz.enable: true
     freeipadlebsencryption.enable: true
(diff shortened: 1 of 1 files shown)


keyki:
The PR has to wait for the UI E2E changes.

PR #8986, opened 2020-09-10
CB-8611. Missing @Valid annotiation from DiagnosticsEndpoints

oleewere:

just find out in sdx diagnostics branch, one of my change was missing

on diagnostics api call, it causes an error as that is redefined in the diagnostics controller call


oleewere:
merging it as it's straight forward change

PR #8991, opened 2020-09-11
CB-8711 [e2e] [e2e] azure fails with NPE (Create failed) -- part 2

gergopapi2:
On azure there is an e2e test utility listing tags by instance id. If, however, the instance id is null, an NPE is produced, marking a test as failed, masking the real reason for failure. There was an additional condition that needed to be fixed.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/util/CloudProviderSideTagAssertion.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/util/CloudProviderSideTagAssertion.java
@@ -105,7 +105,7 @@ public class CloudProviderSideTagAssertion {
     }
 
     private void verifyTags(List<String> instanceIds, Map<String, String> customTags, TestContext testContext, String resourceName) {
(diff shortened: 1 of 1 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/util/CloudProviderSideTagAssertion.java
```diff
@@ -105,7 +105,7 @@ public CloudProviderSideTagAssertion(CloudProviderProxy cloudProviderProxy, Tags
     }
 
     private void verifyTags(List<String> instanceIds, Map<String, String> customTags, TestContext testContext, String resourceName) {
-        if ((instanceIds != null && !instanceIds.isEmpty()) || !instanceIds.contains(null)) {
+        if ((instanceIds != null && !instanceIds.isEmpty()) && !instanceIds.contains(null)) {
```
Bajzathd, line 108: I think filtering out the nulls and verifying the non-nulls would be a better solution, maybe with a warning that there were null instance ids.

aszegedi (approved):

PR #8996, opened 2020-09-11
CB-8754 Introduce the createDefaultImageCatalog setup step to the E2E test

aszegedi:
The createDefaultImageCatalog(testContext); required for the E2E tests' setup, because of 
```
if (!resources.containsKey(key) || resources.get(key) == null) {
            throw new NoSuchElementException(
                String.format("Key: '%s' has been provided but it has no result in the Test Context's Resources map", key));
        }
```
However some of the setup methods do not contain this.

So the missing createDefaultImageCatalog(testContext) step has been introduced to these.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
@@ -12,7 +12,6 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Pattern;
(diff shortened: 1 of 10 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
```diff
@@ -319,6 +319,13 @@ protected void initializeDefaultBlueprints(TestContext testContext) {
                 .validate();
     }
 
+    protected void initializeDefaultImageCatalog(TestContext testContext) {
```
gergopapi2, line 322: Is this method used?

attilapalfi92 (approved):

gergopapi2 (changes requested):

lnardai:
So basically this is only required for the TestContext to be aware that there is an image catalog available, but in most of the testcases I assume we are not doing much just use the default cagtalog.
Can we somehow spare the additional API call?

aszegedi:
> So basically this is only required for the TestContext to be aware that there is an image catalog available, but in most of the testcases I assume we are not doing much just use the default cagtalog.
> Can we somehow spare the additional API call?

Refactored the previous solution. So we can :)

gergopapi2 (approved):

PR #8997, opened 2020-09-11
CB-8626 - purging all finalised flows scheduled with quartz

attilapalfi92:
There is no reason to store finalised flow logs even if they are connected to existing resources. When environments and stacks can run for years, they could have significant amount of flow logs.

Datalake and Freeipa services never had any flowlog cleaning mechanism so the first restart with them after this change could take a while.


attilapalfi92:
Should use quartz scheduling to do the flow removals

akanto:
Totally agree, the only thing that we have problems with startup time, therefore instead of purging things during startup, I would prefer to use a quartz sheduler and purge things periodically e.g every 24h. Quartz can handle singleton scheduling, and it can ensure that purging is executed only on a single instance.

bbihari (approved):
LGTM

PR #9000, opened 2020-09-11
CB-8755 Fix stream mark error

bergerdenes:
The stream type is BufferedInputStream and mark() method call in LegacyStructuredEventFilter.logInboundEntity method
has a parameter readlimit which is set to 64K and if stream reading goes beyond this limit, the mark position is invalidated.

In the next line IOUtils.toString(stream, charset) was called which indeed reads the whole stream to the end and possibly passing the 64K readlimit thus invalidating the mark and throwing an exception when stream.reset is called.

This is fixed by explicitly reading into a pre-determined size byte buffer.


doktoric (approved):

topolyai5 (commented):
Why do you ignore the solution of `while`?

topolyai5:
> Why do you ignore the solution of `while`?

I have seen the slack message, LGTM

topolyai5 (approved):

PR #9004, opened 2020-09-14
CB-8773 remove yarn_nodemanager_resource_memory_mb

doktoric:
CB-8773 remove yarn_nodemanager_resource_memory_mb property from opdb templates because by OPSAPS-56088 CM now autogenerate several property


petersomogyi (approved):
+1

PR #9006, opened 2020-09-14
CB-8778 validate paywall access before starting the upgrade process

keyki:
For runtime upgrade, we have to download the CM and CDP builds from paywall
(archive.cloudera.com/p/). The CM license that is configured uniquely for each
CDP account can be used to authenticate. However, it can happen that customers
do not have an updated license that can be used. In this case, it would be a
mistake to even start the upgrade process. Instead, whenever an upgrade is
requested we'll validate if we can use the configured license to access the paywall.
If it's not valid we will reject the request. Once they update the license in UMS
we will also update the license as part of the upgrade process to make sure
we can download everything. We're also NOT checking the validity of the license
when we're searching for upgrade candidates. This is because this operation happens
very frequently (each time you open up the Datalake details page on the UI). Due to
this we will show a message on the UI that upgrades are available, but when they
try to trigger it we will reject the request.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/upgrade/SdxRuntimeUpgradeService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/upgrade/SdxRuntimeUpgradeService.java
@@ -13,22 +13,35 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.inject.Inject;
+import javax.ws.rs.ProcessingException;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.glassfish.jersey.client.authentication.HttpAuthenticationFeature;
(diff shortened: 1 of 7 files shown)


keyki:
Swagger is failing, because 2.30 is not merged back to master..

datalake/src/main/java/com/sequenceiq/datalake/service/upgrade/SdxRuntimeUpgradeService.java
```diff
@@ -146,6 +174,46 @@ private void verifyRuntimeUpgradeEntitlement(String userCrn, SdxUpgradeRequest u
...
+            if (!isInternalRepoAllowedForUpgrade(userCrn)) {
+                verifyCMLicenseValidity(userCrn);
+            } else {
+                LOGGER.info("Internal repo is allowed for upgrade, skip CM license validation");
+            }
+        }
+    }
+
+    private void verifyCMLicenseValidity(String userCrn) {
+        LOGGER.info("Verify if the CM license is valid to authentichate to {}", paywallUrl);
```
pdarvasi, line 188: typo here

pdarvasi (approved):
LGTM

PR #9011, opened 2020-09-15
CDPCP-2674. Allow syncing single cloud identity

aarman-cloudera:
The datalake service has been updated with a new
`updateUserRangerCloudIdentity` endpoint that allows
updating the cloud identity values for a single user.
The user sync service in FMS has been updated to call
this endpoint for syncs involving a single user.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
@@ -47,6 +47,10 @@ public class ClouderaManagerRangerUtil {
     private static final MapJoiner CLOUD_IDENTITY_CONFIG_MAP_JOINER =
             Joiner.on(";").withKeyValueSeparator("=");
 
+    private static final Boolean MERGE_EXISTING_MAPPING = true;
+
+    private static final Boolean REPLACE_EXISTING_MAPPINNG = false;
+
     @Inject
     private ClouderaManagerProxiedClientFactory clouderaManagerProxiedClientFactory;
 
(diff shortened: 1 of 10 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -350,10 +351,15 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
...
 
-            // TODO For now we only sync cloud ids during full sync. We should eventually allow more granular syncs (actor level and group level sync).
-            if (fullSync && entitlementService.cloudIdentityMappingEnabled(INTERNAL_ACTOR_CRN, stack.getAccountId())) {
+            if (entitlementService.cloudIdentityMappingEnabled(INTERNAL_ACTOR_CRN, stack.getAccountId())) {
                 LOGGER.debug("Starting {} ...", LogEvent.SYNC_CLOUD_IDENTITIES);
-                cloudIdentitySyncService.syncCloudIdentities(stack, umsUsersState, warnings::put);
+                // For cloud identites we only allow full sync or syncing a single user
+                if (fullSync) {
+                    cloudIdentitySyncService.syncCloudIdentities(stack, umsUsersState, warnings::put);
+                } else if (umsUsersState.getUserToCloudIdentityMap().size() == 1) {
```
handavid, line 359: what if we have more than one? since user sync supports sets of actors.

handavid (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
aarman-cloudera: As per comment "For cloud identities we only allow full sync or syncing a single user". The public interfaces don't allow syncing sets of actors so I didn't think we want/need internal support for sets of actors?

datalake/src/main/java/com/sequenceiq/datalake/cm/ClouderaManagerRangerUtil.java
```diff
@@ -47,6 +47,10 @@
     private static final MapJoiner CLOUD_IDENTITY_CONFIG_MAP_JOINER =
             Joiner.on(";").withKeyValueSeparator("=");
 
+    private static final Boolean MERGE_EXISTING_MAPPING = true;
```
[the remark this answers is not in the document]
lacikaaa: I would say, if we don't have to make these configurable via environment variables (using `@Value`) then we should inline them, and the method/parameter name should be enough to reflect it's purpose.

sodre90:
Should we close this PR? Nothing happened since September, do we need this change anymore? @aarman-cloudera

aarman-cloudera:
Let me look into this. I'm not sure if we still need this. If we do, I'll create a new PR since this one is so old.

PR #9018, opened 2020-09-15
CB-8793 Failed to store Audit event, due to unknown field. Cannot ser…

topolyai5:
…ialize the circular referencies


doktoric (approved):

doktoric (approved):

audit-connector/src/main/java/com/sequenceiq/cloudbreak/audit/converter/FlowEventDataExtractor.java
```diff
@@ -33,17 +34,15 @@
 
     @Override
     public EventData eventData(StructuredFlowEvent structuredEvent) {
-        AuditEventDetailsProto.AuditEventDetailsOrBuilder auditEventDetails =
```
doktoric, line 36: if we dont use the builder then we should remove it

PR #9019, opened 2020-09-15
CB-8801 implement Redbeams and sdx force deletion. Tested with 1. env…

doktoric:
… deleted and sdx remained there with force delete 2. env is there with force delete 3. env is there and no force delete 4. env deleted and sdx remained there without force delete

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/flow/delete/SdxDeleteActions.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/flow/delete/SdxDeleteActions.java
@@ -76,7 +76,7 @@ public class SdxDeleteActions {
 
             @Override
             protected Object getFailurePayload(SdxDeleteStartEvent payload, Optional<SdxContext> flowContext, Exception ex) {
-                return SdxDeletionFailedEvent.from(payload, ex);
+                return SdxDeletionFailedEvent.from(payload, ex, payload.isForced());
             }
         };
     }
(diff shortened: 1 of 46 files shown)


datalake/src/main/java/com/sequenceiq/datalake/flow/delete/SdxDeleteActions.java
```diff
@@ -169,14 +169,24 @@ protected void doExecute(SdxContext context, SdxDeletionFailedEvent payload, Map
                 if (exception.getMessage() != null) {
                     statusReason = exception.getMessage();
                 }
-                try {
-                    SdxCluster sdxCluster = sdxStatusService.setStatusForDatalakeAndNotify(DatalakeStatusEnum.DELETE_FAILED, statusReason,
-                            payload.getResourceId());
-                    metricService.incrementMetricCounter(MetricType.SDX_DELETION_FAILED, sdxCluster);
-                } catch (NotFoundException notFoundException) {
-                    LOGGER.info("Can not set status to SDX_DELETION_FAILED because data lake was not found");
+                if (payload.isForced()) {
```
sodre90, line 172: I think it is not a good idea to check "forced" flag here. It would be a better approach if we handle it separately in stack deletion and rds deletion phase. With this implementation, it can hide many error which is not related to stack or rds delete. For example there is a nullpointer exception somewhere, then it will set sdx to delete status but it should not happen. We can move a logic like this into StackDeletionHandler and RdsDeletionHandler.

sodre90 (approved):

PR #9020, opened 2020-09-15
CB-8717 Add thunderhead datalake-dr service connector

hreeve-cloudera:
Adds the datalakedr client to the datalake service so the datalake backup/restore 
flows can query the datalake-dr service for the status of the full backup. Previously 
the datalake service was only aware of the status of the database backup, and could
not track the HBase and Solr backups that were happening simultaneously.

Tested with unit tests, and by running datalake locally and connecting to the
datalake-dr service on mow-dev.


datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+        this.datalakeDrConfig = datalakeDrConfig;
+        this.statusConverter = statusConverter;
+    }
+
+    public DatalakeDrStatusResponse getBackupStatusByBackupId(String datalakeName, String backupId, String actorCrn) {
+        checkNotNull(datalakeName);
+        checkNotNull(actorCrn);
+        checkNotNull(backupId);
+
+        actorCrn = "crn:altus:iam:us-west-1:9d74eee4-1cad-45d7-b645-7ccf9edbb73d:user:8d1e890c-8a2e-4cf9-96bd-d50478bc027e";
```
doktoric, line 44: I guess it should be removed

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+        datalakeName = "biglauer-az-dr-4-dl";
```
doktoric, line 45: I guess it should be removed

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            BackupDatalakeStatusRequest.Builder builder = BackupDatalakeStatusRequest.newBuilder()
+                .setDatalakeName(datalakeName);
+//                .setBackupId(backupId);
```
doktoric, line 50: please remove if not required

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+            );
+        }
+    }
+
+    public DatalakeDrStatusResponse getRestoreStatusByRestoreId(String datalakeName, String restoreId, String actorCrn) {
+        checkNotNull(datalakeName);
+        checkNotNull(actorCrn);
+        checkNotNull(restoreId);
+
+        actorCrn = "crn:altus:iam:us-west-1:9d74eee4-1cad-45d7-b645-7ccf9edbb73d:user:8d1e890c-8a2e-4cf9-96bd-d50478bc027e";
```
doktoric, line 64: I guess it should be removed

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+        datalakeName = "biglauer-az-dr-4-dl";
```
doktoric, line 65: I guess it should be removed

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -0,0 +1,105 @@
...
+
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            RestoreDatalakeStatusRequest.Builder builder = RestoreDatalakeStatusRequest.newBuilder()
+                .setDatalakeName(datalakeName);
+//                .setRestoreId(restoreId);
```
doktoric, line 70: please remove if not required

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class DatalakeDrConfig {
+
+    private static final int DEFAULT_DATALAKE_DR_PORT = 80;
+
+    @Value("${altus.datalakedr.endpoint:}")
```
doktoric, line 14: If we would like to connect one of the cp service then we need changes in dps-k8s repo

datalake-dr-connector/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.8'
+    }
+}
+
+dependencies {
+    compile project(':common')
+    compile project(':grpc-common')
+    compile project(':auth-connector')
+    compile project(':structuredevent-model')
+    compile project(':flow')
```
lacikaaa, line 20: are you sure these dependencies are necessary? I'm not sure about you need common here

lacikaaa (changes requested):
Could you change your commit message? It's misleading, this is not expending a flow, it's about adding a client

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
hreeve-cloudera: Thanks. Went back and cleaned all that up.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
hreeve-cloudera: I have a PR here is to add this config setting in the CSI repo: https://github.infra.cloudera.com/thunderhead/cloud-services-infra/pull/1741. Is that what you mean, or is the dps-k8s repo different?

datalake-dr-connector/build.gradle
hreeve-cloudera: I've cleaned up the build.gradle file to remove any unnecessary dependencies. I did need common for the logger, but there was a bunch of other stuff I could take out. I've also updated the header to "CB-8717 Add thunderhead datalake-dr service connector".

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
keyki: The CSI change alone isn't doing anything. If you want to wire up the 2 services you'll need to make changes in CB helm charts: https://github.infra.cloudera.com/thunderhead/dps-k8s/tree/master/cloudbreak/helm-charts/cloudbreak-umbrella/charts

hreeve-cloudera: Thanks. I'll look into that now.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/converter/GrpcStatusResponseToDatalakeDrStatusResponseConverter.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class GrpcStatusResponseToDatalakeDrStatusResponseConverter {
+
+    public DatalakeDrStatusResponse convert(datalakeDRProto.BackupDatalakeStatusResponse response) {
+        return new DatalakeDrStatusResponse(
+            DatalakeDrStatusResponse.State.valueOf(response.getOverallState()),
+            Optional.ofNullable(response.getFailureReason())
+        );
```
kkalvagadda1, line 15: we can not depend on just the getOverallState. Here you need to check the internal states as well.

If the overall status could be failed but the internal states could be still in progress.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/converter/GrpcStatusResponseToDatalakeDrStatusResponseConverter.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public DatalakeDrStatusResponse convert(datalakeDRProto.RestoreDatalakeStatusResponse response) {
+        return new DatalakeDrStatusResponse(
+            DatalakeDrStatusResponse.State.valueOf(response.getOverallState()),
+            Optional.ofNullable(response.getFailureReason())
```
kkalvagadda1, line 21: we can not depend on just the getOverallState. Here you need to check the internal states as well.

If the overall status could be failed but the internal states could be still in progress.

kkalvagadda1 (changes requested):

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/converter/GrpcStatusResponseToDatalakeDrStatusResponseConverter.java
hreeve-cloudera: My thought there was that if the overall state was failed, but some operations were still in progress, that we no longer had to block the flow. At that point it doesn't matter if something interrupts the b/r operation, because even if the remaining operations succeed, we still end up in a bad state. I'm pretty confident that's true for backup flows. However, I'm not as sure about restore flows. If part of a restore has failed, but there are ongoing operations, would we wind up in a worse position if we interrupted those operations? Or would it not matter because we would have to re-attempt the restore regardless?

kkalvagadda1: I agree with you with the internal behavior. Why do we want to report that backup-restore operations are complete even when some of the internal operations are still in progress when it can be reported accurately with a simple check?

hreeve-cloudera: It's not a question of reporting, it's a question of blocking other flows. If the Solr backup has failed, but the HBase backup is still ongoing, do we want to block other datalake flows until the HBase backup is complete? Or do we want to say that the overall backup has failed, so don't bother waiting for the HBase part to finish and go ahead and unblock the other flows?

hreeve-cloudera: I'm resolving this conversation because it is likely going to be rendered a moot point but an upcoming change in the datalake-dr service.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
lacikaaa: I think the colon should be removed and a default value must be included in the application.yaml which points to local mock, because right now it won't start locally

lacikaaa (changes requested):

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
hreeve-cloudera: The colon is in all the altus service endpoint settings (`"${altus.ums.host:}", "${altus.idbmms.host:}", "${altus.audit.endpoint:}"`, etc.) so I don't want to remove it, but I've added the altus.datalakedr.endpoint setting pointing to the datalake's application.yml file.

keyki, line 14: The value should not have `:` as it will initialize it with an empty string. All these definitions should present in the application.yamls instead.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
```diff
@@ -0,0 +1,52 @@
...
+            String[] parts = endpoint.split(":");
+            if (parts.length < 1 || parts.length > 2) {
+                throw new IllegalArgumentException("altus.datalakedr.endpoint must be in host or host:port format.");
+            }
+            host = parts[0];
+            port = parts.length == 2
+                ? Integer.parseInt(parts[1])
+                : DEFAULT_DATALAKE_DR_PORT;
+        } else {
+            throw new IllegalStateException("altus.datalakedr.endpoint is not configured");
```
keyki, line 33: Does this mean that the application won't start if we're not configuring this endpoint? If that's the case we will need to introduce the datalake dr application in our dev Kubernetes cluster as well along with the local CBD. I would rather create another endpoint to enable/disable this feature so it's not a hard dependency. If it's just not configured then we're skipping some of steps that you do as part of this PR. We do similar things with Cluster Proxy, Public Endpoint Management service and with other external services as well.

keyki: I tried running the CB application and it fails with the endpoint not being set. Probably we should create a datalakedr.enabled variable also with a `false` default value and safeguard all calls to only make them if it's enabled.

hreeve-cloudera: I've removed the datalake-dr-connector dependency from core/build.gradle. I've also removed the colon from `${altus.datalakedr.endpoint}`, included the default value of an empty string in the datalake application.yml file. Finally, I added a new setting `altus.datalakedr.enabled` that defaults to false if it's not set, and will prevent `DatalakeDrConfig` from throwing an exception on `init()` if it's not enabled. Please let me know if that's working for you.

datalake/src/main/resources/application.yml
```diff
@@ -68,6 +68,8 @@ altus:
   audit:
     enabled: true
     endpoint: localhost:8982
+  datalakedr:
+    endpoint: localhost:8982
```
lacikaaa, line 72: this port doesn't seem right to me, having the same value as audit

integration-test/docker-compose_template.yml
```diff
@@ -33,6 +33,7 @@ services:
       - INTEGRATIONTEST_SDX_SERVER=http://dev-gateway
       - INTEGRATIONTEST_REDBEAMS_SERVER=http://dev-gateway
       - ALTUS_AUDIT_ENDPOINT=thunderhead-mock
+      - ALTUS_DATALAKEDR_ENDPOINT=thunderhead-mock
```
lacikaaa, line 36: you are pointing int tests to the mock, but there is no mock implementation of this service, is there? this could be misleading

lacikaaa (changes requested):

integration-test/docker-compose_template.yml
hreeve-cloudera: Removed these settings.

datalake/src/main/resources/application.yml
hreeve-cloudera: 8982 is the service port for all thunderhead services. What exactly is this file used for? I thought I had to add this setting here when I first did the change, but I'm also adding it to the Helm charts in [this review.](https://github.infra.cloudera.com/thunderhead/dps-k8s/pull/120/files), so I'm not sure when this setting is actually going to be used.

lacikaaa: so this is usually for default settings, as we try to avoid in code defaults. in case of endpoints, we can say the default is for local development. in that case this looks a good choice, to point this to the mock, but the mock doesn't have any implementation yet.
so I think we can delete the endpoint line as it doesn't make any difference as the client is disabled and there is nothing which it can call at that port

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
```diff
@@ -0,0 +1,57 @@
...
+
+@Configuration
+public class DatalakeDrConfig {
+
+    private static final int DEFAULT_DATALAKE_DR_PORT = 80;
+
+    @Value("${altus.datalakedr.endpoint}")
+    private String endpoint;
+
+    @Value("${altus.datalakedr.enabled:false}")
```
keyki, line 17: Can you please move the default value to application.yaml?

datalake/src/main/resources/application.yml
keyki: Since we have `altus.datalakedr.enabled` set to `false` we don't' need to set any value to it.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/config/DatalakeDrConfig.java
hreeve-cloudera: Done.

kkalvagadda1 (approved):

lacikaaa (approved):
please resolve the conflict

PR #9021, opened 2020-09-15
CB-8717 Expand B/R flow to cover full backup/restore

hreeve-cloudera:
Adds a new stage to the backup/restore flow config. When the database b/r is
complete, a new event will be triggered to query the datalake-dr service for
the final status of the full operation. The flow will complete 1) when the
datalake-dr service reports the full operation is complete, 2) it runs into
an error querying the datalake-dr service, or 3) a timeout is reached (30
minutes by default, can by changed by the sdx.stack.restore.status.duration_min
setting).
    
Tested with updated unit tests and by running  a backup/restore and forwarding
requests to the datalake-dr service on mow-dev.


kkalvagadda1 (changes requested):
Why do have changes to cloud break service? cloudbreak still does what it did before which is database backup-restore operations.

PR #9023, opened 2020-09-16
CB-8822 npe check in sdxinternaltestdto investigate

afarsang:
npe check in sdxinternalTestDto investigate 
beforehand it had nullpointer exception(only in test log)

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/sdx/SdxInternalTestDto.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/sdx/SdxInternalTestDto.java
@@ -431,7 +431,7 @@ public class SdxInternalTestDto extends AbstractSdxTestDto<SdxInternalClusterReq
                 CloudbreakEventService.DATAHUB_RESOURCE_TYPE,
                 null,
                 getResponse().getCrn());
-        boolean hasSpotTermination = getResponse().getStackV4Response().getInstanceGroups().stream()
+        boolean hasSpotTermination = (getResponse().getStackV4Response() == null) ? false : getResponse().getStackV4Response().getInstanceGroups().stream()
                 .flatMap(ig -> ig.getMetadata().stream())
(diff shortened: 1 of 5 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/sdx/SdxInternalTestDto.java
```diff
@@ -423,7 +423,7 @@ public SdxRepairRequest getSdxRepairRequest() {
 
     @Override
     public Clue investigate() {
-        if (getResponse() == null || getResponse().getCrn() == null) {
+        if (getResponse() == null || getResponse().getCrn() == null || getResponse().getStackV4Response() == null) {
```
Bajzathd, line 426: If we can have a crn in the response but not stackresponse, it means that the response could have other useful info in it. So I think a better solution would be to handle the possible NPE in the hasSpotTermination assignment, something like this:
`boolean hasSpotTermination = Optional.ofNullable(getResponse().getStackV4Response())
                .map(stack -> stack.getInstanceGroups())
                .stream()
                .flatMap(ig -> ig.getMetadata().stream())
                .anyMatch(metadata -> InstanceStatus.DELETED_BY_PROVIDER == metadata.getInstanceStatus());`
This way we could add the response to the report without an NPE. Also it is possible that other Investigables have this issue, please check them too.

afarsang: so you mean audit info still useful, whether stackresponse is null?

Bajzathd: I think it can be as it already had a crn, because without it could not have caused an NPE.

afarsang: other testdtos had the potential issue too, it was a good idea to check them. Thanks

Bajzathd (approved):

lacikaaa (approved):

lnardai (approved):

PR #9024, opened 2020-09-16
CB-8811 Flow Management Template Installation Fails

vbrodetskyi:


doktoric:
@vbrodetskyi this problem came up on the current release which is CB-2.29.0

PR #9027, opened 2020-09-16
CB-8825 MockUMS fixes for thunderhead DatalakeApi service custom DL

lajosrodek:
Contents:
* Improve `MockUserManagementService.getAccount()`:
  * Grant entitlement `CDP_CM_ADMIN_CREDENTIALS`
  * Set `accountId` in response (as per request)
  * Set `externalAccountId` in response
* Implement `MockUserManagementService.listTerms()` (empty dummy response only)
* Update existing unit tests & add some new cases

### DIFF
--- a/mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
+++ b/mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
@@ -85,10 +85,12 @@ import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListM
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListMachineUsersResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListResourceAssigneesRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListResourceAssigneesResponse;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesRequest;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListRolesRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListRolesResponse;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesRequest;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesResponse;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListTermsRequest;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListTermsResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListUsersRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListUsersResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListUsersResponse.Builder;
@@ -126,16 +128,16 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.io.Resources;
 import com.google.protobuf.util.JsonFormat;
-import com.sequenceiq.thunderhead.grpc.GrpcActorContext;
-import com.sequenceiq.thunderhead.model.AltusToken;
-import com.sequenceiq.thunderhead.util.CrnHelper;
-import com.sequenceiq.thunderhead.util.IniUtil;
-import com.sequenceiq.thunderhead.util.JsonUtil;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.Crn.ResourceType;
 import com.sequenceiq.cloudbreak.auth.altus.UmsRight;
 import com.sequenceiq.cloudbreak.auth.altus.model.AltusCredential;
 import com.sequenceiq.cloudbreak.util.SanitizerUtil;
+import com.sequenceiq.thunderhead.grpc.GrpcActorContext;
+import com.sequenceiq.thunderhead.model.AltusToken;
+import com.sequenceiq.thunderhead.util.CrnHelper;
+import com.sequenceiq.thunderhead.util.IniUtil;
+import com.sequenceiq.thunderhead.util.JsonUtil;
 
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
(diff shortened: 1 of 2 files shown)


gregito:
@lajosrodek, did you apply the designated IDEA config to not reorder imports, or the original class (MockUserManagementService.java) has created with a "wrong" import order?

doktoric (approved):

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
```diff
@@ -210,6 +212,9 @@
 
     private static final String CDP_ALLOW_INTERNAL_REPOSITORY_FOR_UPGRADE = "CDP_ALLOW_INTERNAL_REPOSITORY_FOR_UPGRADE";
 
+    // See com.cloudera.thunderhead.service.common.entitlements.CdpEntitlements.CDP_CP_CUSTOM_DL_TEMPLATE
+    private static final String CDP_CP_CUSTOM_DL_TEMPLATE = "CDP_CM_ADMIN_CREDENTIALS";
```
lacikaaa, line 216: bit strange that the variable name is different from the value. would this mean `CDP_CM_ADMIN_CREDENTIALS` has the right to use/create custom templates?

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
```diff
@@ -590,7 +595,10 @@ public void getAccount(GetAccountRequest request, StreamObserver<GetAccountRespo
                                 .addEntitlements(createEntitlement(DATAHUB_AZURE_AUTOSCALING))
                                 .addEntitlements(createEntitlement(DATAHUB_AWS_AUTOSCALING))
                                 .addEntitlements(createEntitlement(LOCAL_DEV))
+                                .addEntitlements(createEntitlement(CDP_CP_CUSTOM_DL_TEMPLATE))
                                 .setPasswordPolicy(workloadPasswordPolicy)
+                                .setAccountId(request.getAccountId())
+                                .setExternalAccountId("external-" + request.getAccountId())
```
doktoric, line 601: is this ok to using a fixed external name here?

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
```diff
@@ -85,10 +85,12 @@
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListMachineUsersResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListResourceAssigneesRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListResourceAssigneesResponse;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesRequest;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListRolesRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListRolesResponse;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListServicePrincipalCloudIdentitiesRequest;
```
doktoric, line 90: the import order in your intellij is somehow different that ours please check

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
lajosrodek: Yes. The name is indeed totally confusing. I am not sure what the originally intended name was supposed to be, but wanted to keep the thunderhead constant name here for "consistency". For reference, the check in DatalakeApi svc is made [here](https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/api/datalake/src/main/java/com/cloudera/thunderhead/service/datalakeapi/server/DatalakeApiImpl.java#L1095-L1107).

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
lajosrodek: Honestly I do not know what this is supposed to be like. For most of the real UMS users, the externalAccountId is derived from the corresponding SalesForce (SFDC) account ID. For instance, mow-dev (accountId=9d74eee4-1cad-45d7-b645-7ccf9edbb73d) has externalAccountId=0018000000ik5epAAA. As far as I can tell, the `001` is a fixed prefix, while the last 3 characters are a sort-of checksum calculated from the rest of the string.

The relevant code in DatalakeApi svc making the check is located [here](https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/api/datalake/src/main/java/com/cloudera/thunderhead/service/datalakeapi/server/DatalakeApiImpl.java#L1102-L1105). The underlying implementation currently only asserts that the account response contains both accountId and externalAccountId and they are non-empty.

I could find a Java implementation for the calculation logic of "real" externalAccountId's based on the SFDC ID in [thunderhead](https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/services/libs/common-test/src/main/java/com/cloudera/thunderhead/service/common/test/sfdc/SFDCUtils.java#L30-L46), but did not feel it justified to duplicate the same behavior here until it is crucial. What is your preference?

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
lajosrodek: Shall I rather rename the constant to match the actual entitlement name?

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
lajosrodek: Pardon me but my ordering seems alright. All the imports are now in the correct lexicographical order. Why would `ListServicePrincipalCloudIdentitiesResponse` come before `ListRolesRequest`? Same below, `com.sequenceiq.thunderhead.*` seems to be greater than `com.sequenceiq.cloudbreak.*`, so the former ones shall probably come later in the import list. Could you please clarify what this should look like?

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
lacikaaa: I'm  fine with it, it just seemed interesting and could indicate some error. but @keyki should check this

keyki: It's fine with me. If we would rename it to CDP_CM_ADMIN_CREDENTIALS probably it would be totally misleading what is it supposed to guard. But, indeed I'm not sure what this entitlement supposed to do.  Seems like a bad copy-paste error: https://github.infra.cloudera.com/thunderhead/thunderhead/blob/4cf3cae30759e47df352b2b41bc8b9c30b2c43b8/services/libs/common/src/main/java/com/cloudera/thunderhead/service/common/entitlements/CdpEntitlements.java#L510.

lajosrodek: Yeah. I fear it is too late to "rename" the entitlement as it might be already in use in several tenants. 😃

lajosrodek:
@gregito Double-checked with @doktoric , the original class contained a wrong import order.

PR #9030, opened 2020-09-17
CB-8800 Increase wait time for repair on SDX service side

lacikaaa:
During E2E tests I have seen multiple times SDX service timeout while
the repair finished successfully. This commit will increase the wait
time from 40 mins to 60.


sodre90 (approved):

foldik:
Last time I came up with the idea to increase this timeout the saying was that "a datalake should come up in 40 minutes, we shouldn't increase this timeout", which obviously gives a more complicated task "Why 40 minutes not enough?". @lnardai is it fine to increase the timeout, what is the generally accepted fix for this case?

lacikaaa:
I'm investigating what changed increasing the time necessary for recovery, but it currently might hide other errors if we don't wait for it to finish.

foldik:
Okay, so is it just a temporary fix and we have a task to use 40 minutes timeout again?

lacikaaa:
if I find any issues and I can fix it, I will reset this to 40 min.

PR #9036, opened 2020-09-18
CB-8700 optimizations for IT test

sodre90:

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
@@ -82,10 +82,29 @@ public class PkiUtil {
 
     private static final Integer MAX_CACHE_SIZE = 200;
 
-    private static final Map<String, RSAKeyParameters> CACHE =
+    private static final KeyPairGenerator KEY_GEN;
+
+    static {
+        try {
+            KEY_GEN = KeyPairGenerator.getInstance("RSA");
+            KEY_GEN.initialize(KEY_SIZE, new SecureRandom());
+        } catch (Exception e) {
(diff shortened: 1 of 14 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/certificate/PkiUtil.java
```diff
@@ -118,15 +137,22 @@ public static String generateSignature(String privateKeyPem, byte[] data) {
...
-        signer.init(true, rsaKeyParameters);
-        signer.update(data, 0, data.length);
+        Signer signer = PSS_SIGNER_CACHE.get(rsaKeyParameters);
+
+        if (signer == null) {
+            signer = new PSSSigner(new RSAEngine(), new SHA256Digest(), SALT_LENGTH);
+            signer.init(true, rsaKeyParameters);
+            signer.update(data, 0, data.length);
+
+            PSS_SIGNER_CACHE.put(rsaKeyParameters, signer);
```
topolyai5, line 153: little bit looks weird. The map is static final but modifies the content of the map. Why is it static final?

sodre90: the reference is final and I can initialize it in static initializer block

topolyai5: I understand that, but why do you create the map as static final?

sodre90: 1) because if it is not static then I can not initialize it in static block
2) It is static because I wanted one instance per JVM
3) I followed the previous CACHE solution :)

lacikaaa (approved):

PR #9038, opened 2020-09-18
CB-8872 Authz framework improvements

horadla23:
- CB-8873 Remove InternalReady annotation
- CB-8874 Move internal actor related annotations into separate package


auth-internal/src/test/java/com/sequenceiq/cloudbreak/auth/security/internal/InternalCrnModifierTest.java
```diff
@@ -24,6 +24,7 @@
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
+import com.sequenceiq.cloudbreak.auth.ReflectionUtil;
```
lnardai, line 27: Do we really need this?

lnardai (approved):

auth-internal-api/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    jcenter()
+  }
+  dependencies {
+    classpath "org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}"
+  }
+}
+
+dependencies {
+  compile project(':common')
+  testCompile group: 'junit', name: 'junit', version: junitVersion
```
lnardai, line 15: I think the junit impl is defined in the parent gradle file. Please try to rely on that.

auth-internal/src/test/java/com/sequenceiq/cloudbreak/auth/security/internal/InternalCrnModifierTest.java
horadla23: yes, it is used in the test, i just moved that class into common, so we have to import it

PR #9042, opened 2020-09-18
CB-8809 verify that AWS credentials' role is assumable without external Id

biharitomi:

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
@@ -24,12 +24,15 @@ import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
 import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
 import com.sequenceiq.cloudbreak.cloud.model.CloudCredentialStatus;
 import com.sequenceiq.cloudbreak.cloud.model.CredentialStatus;
+import com.sequenceiq.cloudbreak.cloud.model.credential.CredentialVerificationContext;
 import com.sequenceiq.cloudbreak.cloud.response.AwsCredentialPrerequisites;
 import com.sequenceiq.cloudbreak.cloud.response.CredentialPrerequisitesResponse;
 import com.sequenceiq.common.model.CredentialType;
 
 @Service
 public class AwsCredentialConnector implements CredentialConnector {
+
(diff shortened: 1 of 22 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
```diff
@@ -168,6 +184,29 @@ private CloudCredentialStatus verifyAccessKeySecretKeyIsAssumable(CloudCredentia
         return credentialStatus;
     }
 
+    private void checkRoleIsAssumableWithoutExternalId(CredentialVerificationContext credentialVerificationContext, AwsCredentialView awsCredential) {
+        if (credentialVerificationContext.getCreationVerification()) {
+            String roleArn = awsCredential.getRoleArn();
+            try {
+                credentialClient.retrieveSessionCredentialsWithoutExternalId(awsCredential);
+                String message = String.format("CDP Control Pane is able to perform 'sts:AssumeRole' on '%s' role without external id. " +
```
doktoric, line 192: I am not sure but should we somehow show what is the external id which is missing ? I guess not but lets check

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
```diff
@@ -168,6 +184,29 @@ private CloudCredentialStatus verifyAccessKeySecretKeyIsAssumable(CloudCredentia
...
+                        "The role is vulnerable and could be exploited by confused deputy attacks. " +
+                        "Please update the role's trust relationship to have a condition on external id " +
+                        "or re-create the role with external id configured in the create role wizard.", roleArn);
+                LOGGER.warn(message);
+                throw new AwsPossibleConfusedDeputyException(message);
+            } catch (AmazonClientException ae) {
+                if (ae.getMessage().contains(ROLE_IS_NOT_ASSUMABLE_ERROR_MESSAGE_INDICATOR)) {
+                    String msg = String.format("Consider the specified role as secured. " +
+                            "CDP Control Pane is not authorized to perform sts:AssumeRole on '%s' role without external Id.", roleArn);
```
doktoric, line 201: I think here the problem is not the missing external id right ?

doktoric (changes requested):
we should also mention this in mow-dev channel

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsCredentialConnector.java
biharitomi: That's a happy path and only a log is called to leave a sign about it.

doktoric (approved):

PR #9050, opened 2020-09-21
CB-8889 added Ycloud defintions for every cluster clustershape

doktoric:


biharitomi (approved):
OMG

Bajzathd (approved):

PR #9051, opened 2020-09-21
CB-8878 - added tracing instrumentation for azure api calls

attilapalfi92:


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClientCredentials.java
```diff
@@ -32,24 +40,53 @@
...
     }
 
+    private Interceptor getTracingInterceptor() {
+        return chain -> {
+            Request request = chain.request();
+            String loggingContext = request.header("x-ms-logging-context").split(" ")[0];
+            Span span = tracer.buildSpan("[" + request.method() + "] " + loggingContext)
+                    .addReference(References.FOLLOWS_FROM, tracer.activeSpan() != null ? tracer.activeSpan().context() : null)
+                    .start();
+            span.setTag("component", "java-azure-sdk");
```
lacikaaa, line 72: I think these magic strings like `component`, `http.method`, `error` should come from `TracingUtil` or similar class, so when you add support to another sdk, etc we won't duplicate those.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClientService.java
```diff
@@ -23,6 +25,9 @@
     @Inject
     private AuthenticationContextProvider authenticationContextProvider;
 
+    @Inject
+    private Tracer tracer;
```
lacikaaa, line 29: I think it would be nice if you would inject a new class here, like `AzureTracer` and it would have the `public Interceptor getTracingInterceptor()` method and you could pass that along. So it would have the responsibility creating the interceptor instead of `AzureClientCredentials`. What do you think?

lacikaaa (changes requested):

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClientService.java
attilapalfi92: Im going to create TracingInterceptor as a separate class and inject it directly instead

lacikaaa (approved):

PR #9052, opened 2020-09-21
CB-8891. Prepare for using cdp-logging-agent instead of td-agent.

oleewere:

- handle that if cdp-logging-agent or td-agent (or both) exists on the nodes
-> if that code is deployed (CB 2.31):
- with new CB, if a cluster is old but has newer salt definitions, td-agent still can be used
- with new CB, if cluster and salt definitions are old, td-agent still can be used
- with new CB, and new image with cdp-logging-agent, cdp-logging-agent can be used
(overall both have similar salt states, have some changes for delayed restart or for cdp-logging-agent we wont install any plugins by salt defs + won't override systemd service file with root user/group)

was manually tested, with support only td-agent, then install cdp-logging-agent on the node and re-run state.apply (which removes td-agent and init cdp-logging-agent)


attilapalfi92 (approved):
lgtm

PR #9053, opened 2020-09-21
CB-8901: Fix FreeIPA repair with a requested instance

jamisonbennett:
Fix FreeIPA repair so that if there is an instance stuck in the
requsted state, the repair operation is able to succeed and it clean
up as much as possible.

This was tested manully using a local instance of cloudbreak.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
@@ -91,6 +91,11 @@ public class TerminationService {
         }
     }
 
+    public void finalizeTerminationForInstancesWithoutInstanceIds(Long stackId) {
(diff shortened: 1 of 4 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/termination/action/TerminationService.java
```diff
@@ -91,6 +91,19 @@ public void finalizeTermination(Long stackId, List<String> instanceIds) {
         }
     }
 
+    public void finalizeTerminationForInstancesWithoutInstanceIds(Long stackId) {
+        try {
+            transactionService.required(() -> {
```
lacikaaa, line 96: I don't think you need transaction here

lacikaaa (changes requested):

lacikaaa (approved):

jamisonbennett:
@lnardai can you please recommend who else needs to review this PR?

PR #9054, opened 2020-09-21
CB-8909: Fix FreeIPA repair of deleted instances

jamisonbennett:
In AWS if a FreeIPA instance was deleted on the cloud provider side,
the repair operation would not properly clean up the instances. For
example, associated cloud watches were not cleaned up. Additionally, it
caused problems during the upscale since the instance group still
contained the terminated instance and so the upscale would wait until
the operation timed out.

Repair of a deleted instance was validated on AWS using a local
deployment of cloudbreak.


christmasferret (commented):
LGTM!

christmasferret (approved):

PR #9055, opened 2020-09-21
CB-8910: Fix null pointer exception when repairing FreeIPA

jamisonbennett:
Fix a null pointer exception that would be thrown when repairing
FreeIPA without specifying which instances to repair. This is
supposed to automatically detect which instance to repair.

This was texted manually using a local deployment of cloudbreak.


lacikaaa (approved):
LGTM, but would be nice to cover with unit test

jamisonbennett:
I will add a unit test, please do not merge until I add it.

PR #9059, opened 2020-09-22
CB-8714 Fix OCSP and CRL URL for CM CA

lacikaaa:
When creating cloudera manager intermediate CA, it has a default
profile present. This has invalid settings for OCSP and CRL. In this
commit we are replacing those values with the right ones.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
@@ -88,6 +88,18 @@ cloudera_manager_set_parcel_validation:
 
 {% if cloudera_manager.communication.autotls_enabled == True %}
 
+replace_crl_in_cmca_profile:
+  file.replace:
+    - name: /etc/cloudera-scm-server/cmSubCaCert.profile
+    - pattern: "policyset.cmSubCaCertSet.9.default.params.crlDistPointsPointName_0=http://changeme.com/ipa/crl/MasterCRL.bin"
+    - repl: "policyset.cmSubCaCertSet.9.default.params.crlDistPointsPointName_0=http://ipa-ca.{{ metadata.cluster_domain }}/ipa/crl/MasterCRL.bin"
+
+replace_ocsp_in_cmca_profile:
+  file.replace:
+    - name: /etc/cloudera-scm-server/cmSubCaCert.profile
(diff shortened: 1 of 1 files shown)


keyki:
Can you add some more content to the description about what OCSP is and what it supposed to solve?

orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/init.sls
```diff
@@ -88,6 +88,18 @@ cloudera_manager_set_parcel_validation:
 
 {% if cloudera_manager.communication.autotls_enabled == True %}
 
+replace_crl_in_cmca_profile:
+  file.replace:
+    - name: /etc/cloudera-scm-server/cmSubCaCert.profile
+    - pattern: "policyset.cmSubCaCertSet.9.default.params.crlDistPointsPointName_0=http://changeme.com/ipa/crl/MasterCRL.bin"
+    - repl: "policyset.cmSubCaCertSet.9.default.params.crlDistPointsPointName_0=http://ipa-ca.{{ metadata.cluster_domain }}/ipa/crl/MasterCRL.bin"
```
keyki, line 95: Should we have an unless condition or it doesn't matter as replace wouldn't do anything anyway?

lacikaaa: AFAIK `file.replace` will only make a change if it has any effect. But I can test it

lacikaaa:
```
----------
          ID: cloudera_manager_set_parcel_validation
    Function: file.replace
        Name: /opt/cloudera/cm/bin/cm-server
      Result: True
     Comment: unless condition is true
     Started: 10:26:57.457679
    Duration: 20.731 ms
     Changes:
----------
          ID: replace_crl_in_cmca_profile
    Function: file.replace
        Name: /etc/cloudera-scm-server/cmSubCaCert.profile
      Result: True
     Comment: No changes needed to be made
     Started: 10:26:57.478883
    Duration: 5.332 ms
     Changes:
----------
          ID: replace_ocsp_in_cmca_profile
    Function: file.replace
        Name: /etc/cloudera-scm-server/cmSubCaCert.profile
      Result: True
     Comment: No changes needed to be made
     Started: 10:26:57.484460
    Duration: 4.205 ms
     Changes:
----------
```
the first state is with unless+grep, the other 2 is the new state. in result there is no difference, it's even faster this way

PR #9064, opened 2020-09-22
CB-8937: FreeIPA doesn't automatically restart polkit

christmasferret:
Add polkit service systemd salt sls restart setting in FreeIPA instances. Polkit is a component for controlling system-wide privileges in Unix-like operating systems. Currently, polkit service won't restart after it fails. With this fix, polkit service will restart after a three-second delay. This is in consistent with all other services running in FreeIpa instances.


christmasferret:
After the fix, polkit.service is in consistence with other services.

[Unit]
Description=Authorization Manager
Documentation=man:polkit(8)

[Service]
Restart=always
RestartSec=3

jamisonbennett (approved):

PR #9069, opened 2020-09-23
CB-8924 AWS InsufficientInstanceCapacity fix on ASG

bergerdenes:
In case InsufficientInstanceCapacity is returned on an ASG request
either on launch or upscale, we need to keep waiting for the instances
to be provisioned by AWS as this failure seems recoverable, i.e. in a few
minutes AWS will likely launch the requested number of instances in the ASG.


doktoric:
This is what we discussed so lets wait for other reviewers

pdarvasi (approved):

topolyai5 (approved):

PR #9071, opened 2020-09-23
CB-8947-filter-structured-event

topolyai5:

### DIFF
--- a/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/db/CDPStructuredEventDBService.java
+++ b/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/db/CDPStructuredEventDBService.java
@@ -13,6 +13,7 @@ import org.springframework.core.convert.ConversionService;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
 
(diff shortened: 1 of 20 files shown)


structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/db/CDPStructuredEventDBService.java
```diff
@@ -68,43 +69,43 @@ protected void prepareCreation(CDPStructuredEventEntity resource) {
...
     }
+
+    private List<StructuredEventType> getAllEventTypeIfEmpty(List<StructuredEventType> eventTypes) {
+        List<StructuredEventType> types = new ArrayList<>(eventTypes);
+        if (CollectionUtils.isEmpty(eventTypes)) {
+            LOGGER.info("We need to add all structured event types to the filter");
+            types.add(StructuredEventType.NOTIFICATION);
+            types.add(StructuredEventType.REST);
+            types.add(StructuredEventType.FLOW);
+        }
```
biharitomi, line 108: Shouldn't we do something if the eventypes is not empty??

biharitomi (changes requested):

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/db/CDPStructuredEventDBService.java
topolyai5: Yes, we do. We return with the original list

biharitomi (approved):

PR #9080, opened 2020-09-24
CB-8864 added auditproto file back to use in case of audit and help u…

doktoric:
…s generate the documentation. We had problems with the standard jsonutil java so I just changed it to gson.


structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/audit/extractor/FlowCDPEventDataExtractor.java
```diff
@@ -97,4 +108,9 @@ public boolean shouldAudit(CDPStructuredEvent structuredEvent) {
     private CDPFlowResourceAuditEventConverter getConverter(String resourceType) {
         return resourceAuditEventConverters.get(resourceType + "FlowAuditEventNameConverter");
     }
+
+    public ObjectMapper jacksonObjectMapper() {
```
topolyai5, line 112: is it used?

topolyai5 (changes requested):

topolyai5 (approved):

PR #9081, opened 2020-09-24
CB-9032 Introduce a Hibernate event listener that will try to detect inefficient queries executed by Hibernate

akanto:
CB-9032 Introduce a Hibernate event listener that will try to detect inefficient queries executed by Hibernate. The goal is to catch those cases when more queries are executed than expected, this is usually referred as N + 1 query problem in Hibernate.

If we are over a fixed treshold then log an error if too many queries are executed in a tx (useful for test environemnts).


bbihari (approved):
cool

daszabo (approved):
Cool. Have you though about overriding flushEnd too? It gets the number of entities flushed that you could also log. I am not sure but may be it would have some benefit in understanding the scale of entities inserted during stack creation.

akanto:
Ok, I will take a look at counting entities in flush

PR #9083, opened 2020-09-24
CB-8418 Introduce ResizeDatahubCluster event in Audit. We introduce i…

topolyai5:
…nformation in case of resizing what was the original and new value.


doktoric (changes requested):
unit test pls

doktoric (approved):

doktoric:
thanks

PR #9085, opened 2020-09-24
CB-8703 fix authz framework to handle null queryparam

horadla23:
If there is an API method which can be called with internal actor and @AccountId/@TenantAwareParam/@InitiatorUserCrn param is null, then the framework will throw an ugly NPE.
If this is occuring in PROD, the workaround is to fill in the parameter, so I opened this PR for 2.31, not for 2.29.


akanto (changes requested):
Missing unit test

horadla23:
> Missing unit test

right, added

akanto (approved):

PR #9086, opened 2020-09-24
CB-8904 Check CM version when filtering images for upgrade

lacikaaa:
Upgrade with lock components hasn't filtered images by Cloudera Manager build number.
In this commit an additional check has been added. Also the filtering for CM and stack
version has been refactored into smaller classes.


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/locked/ParcelMatcher.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sequenceiq.cloudbreak.service.image.PreWarmParcelParser;
+
+@Component
+public class ParcelMatcher {
+
+    @Inject
+    private PreWarmParcelParser preWarmParcelParser;
+
+    // Only compares the versions of the activated parcels
+    public boolean isMatchingParcels(Image image, Map<String, String> activatedParcels) {
```
keyki, line 25: The name should indicate that it doesn't check the CDH parcel.

keyki:
1 minor naming suggestion, otherwise LGTM.

keyki (approved):

PR #9087, opened 2020-09-24
Include "cloudPlatform" in the data lake list API

atxmemo:
Based on the cloudPlatform (AWS or Azure) we need to hit different services to interact with metadata for data residing with the cloud providers (in addition to ATLAS), e.g. AWS Glue Data Catalog in AWS.


atxmemo:
NOTE: This implementation will incur an additional API call per SDX Cluster.

lnardai:
Could you add a JIRA ticket number to the PR?

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
```diff
@@ -134,8 +134,11 @@ public SdxClusterResponse getByCrn(@TenantAwareParam @ResourceCrn String cluster
         String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
         List<SdxCluster> sdxClusters = sdxService.listSdx(userCrn, envName);
         return sdxClusters.stream()
-                .map(sdx -> sdxClusterConverter.sdxClusterToResponse(sdx))
-                .collect(Collectors.toList());
+                .map(sdxCluster -> {
+                    SdxClusterResponse sdxClusterResponse = sdxClusterConverter.sdxClusterToResponse(sdxCluster);
+                    sdxService.setAdditionalClusterResponseFields(sdxClusterResponse, sdxCluster);
```
lnardai, line 139: I'm not a fan of this change. This might have performance impact since you get the information one by one. Listing should be quick.
We need to find a more economic way to get this data, some ideas:
- can't we get a list of environments and match them by envCrn and get the Provider based on that.
- store this in the SdxCluster model in the database since this won't change and we don't have to worry about this getting out of sync
- infer the Provider from some other data from the SdxCluster Model.

We need to think of a way to make this as a batch call, or somehow infer the Data form elsewhere.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -646,6 +648,19 @@ public void updateRuntimeVersionFromStackResponse(SdxCluster sdxCluster, StackV4
         return result;
     }
 
+    public void setAdditionalClusterResponseFields(SdxClusterResponse sdxClusterResponse, SdxCluster sdxCluster) {
```
lnardai, line 651: Please add some units tests once the implementation is final.

lnardai (changes requested):

atxmemo:
It's the branch name, but sure: https://jira.cloudera.com/browse/CB-7928

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
atxmemo: Same, which is why I noted above: `NOTE: This implementation will incur an additional API call per SDX Cluster.`

Thanks for the ideas

lnardai: Yes, thanks, I saw the comment, appreciate it.

PR #9088, opened 2020-09-24
CB-xxxx Introduce a simple circuit breaker that will detect the ineff…

akanto:
…icient queries executed by Hibernate. The goal is to catch those cases when more queries are executed than expected, this is usually referred as N + 1 query problem in Hibernate.

If we are over the treshold then the circuit breaker can do multiple things:
- log an error if too many queries are executed in a tx (useful for test environemnts)
- break the tx if too many queries are executed in a tx (useful in IT test, to catch the problem as early as possible)
- do nothing (stage and prod), since it statistic counter cases some perfomance degaradation therefore probably we shall not enable it


akanto:
This is just a playground

PR #9063, opened 2020-09-22
CB-8931 Introduce Gcloud template as a first drop

doktoric:
See detailed description in the commit message.

PR #9097, opened 2020-09-29
CB-3254 Adding GCP support for dataengineering cluster creation

cegganesh84:
1. We have not fixed the minimum version for GCP. So just added 7.2.2 and 7.2.3 templates for testing.
2. Tested via dp cli.


doktoric:
@cegganesh84 similar like https://github.com/hortonworks/cloudbreak/pull/9063

cegganesh84:
Thanks, @doktoric, I will discard this in favor of #9063

PR #9107, opened 2020-09-29
DISTX-533 Set S3A buffer dir to Yarn local dir

amitagarwal06:
[DISTX-533](https://jira.cloudera.com/browse/DISTX-533)

Setting fs.s3a.buffer.dir=${env.LOCAL_DIRS:-${hadoop.tmp.dir}}/s3a
for DE, DE HA & Spark3 templates.

Testing done:

Have created a DE cluster using a modified 7.2.3 template that includes above mentioned config changes. Cluster creation was successful on AWS using m5d.2xl instance types.


keyki:
@lnardai as release owner

amitagarwal06:
@steveloughran Please review and approve it.

doktoric:
please use commit message <jira-ticket-id> <msg>

lnardai:
Judging by the ticket this is not a regression but an enhancement. We should target the master branch with this.

amitagarwal06:
@lnardai  If we need some changes on a specific branch as well as on master, Krisztian has suggested us to raise PR for that specific branch alone (here CB-2.31.0)and he will merge it to master too.
That is why it was raised against CB-2.31.0 branch.

steveloughran:
+1 to the change -the property is as we resolved in the escalation
in a YARN cluster, spark & MR will use env.LOCAL_DIRS for storing staged files, so when a container is terminated all this disk space is cleaned up. Fallback is hadoop.tmp.dir when that var is unset, which is today's default

steveloughran (approved):
I approve the config changes; leave it to others to worry about the correct branches

amitagarwal06:
@lnardai Can you pls merge this PR?

doktoric:
@amitagarwal06 can you please modify the commit message for a pattern like: DISTX-533 <problem which is solved by this PR>
Also we are mostly prefer to do not use safety valve so please raise an OPSAPS issue to introduce a new variable name in CM for this property.

amitagarwal06:
> please

amitagarwal06:
> @amitagarwal06 can you please modify the commit message for a pattern like: DISTX-533
> Also we are mostly prefer to do not use safety valve so please raise an OPSAPS issue to introduce a new variable name in CM for this property.

@doktoric I have changed commit message as per your comment. 
Regarding not using safety valve, I understand this well and have raised few OPSAPS in past (for e.g. https://jira.cloudera.com/browse/OPSAPS-55598) to expose some of the properties where it made sense. But the property fs.s3a.buffer.dir in this commit has been added to the same safety valve where all S3Guard related properties (fs.s3a.*) are added on the fly during cluster creation. Also it does not make sense to expose this as individual config in CM because we dont expect folks to change it afterwards.

I hope this helps.

amitagarwal06:
Looks like blueprints have been re-organized under version specific folders. I am closing this PR and going to create another PR.

PR #9109, opened 2020-09-30
CB-9056. Prevent restart of clusters creating new machines user acces…

oleewere:
…s keys in UMS.

details:
- create a new column (secret) for cluster in order to store databus access/private keys
(could not use component table properly with secrets and cluster attributes are not really used I think, so i have created a new one)
- get this value during salt pillar config update (or during deployment), if not set, generate a new keypair and save it in the cluster entity
- the exact same thing will be needed for freeipa as well (not needed yet as no salt pillar config upgrade there yet)

also it requires some local testing as well (against UMS, using shared credential + local setup)

update: testing with local setup + with using remote UMS has finished (first one with the logic, second one is just for the calls to make sure those will work)

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -641,6 +641,24 @@ public class GrpcUmsClient {
     }
 
     /**
+     * Check that machine user has a specific access key in UMS
+     * @param actorCrn actor for the machine user request
+     * @param accountId the account ID
+     * @param machineUserCrn machine user crn that own the access key
+     * @param accessKeyId access key id that we need to check
+     * @return result that is true if the machine user has the queried access key in UMS
+     */
+    public boolean doesMachineUserHasAccessKey(String actorCrn, String accountId,
+            String machineUserCrn, String accessKeyId) {
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
+            String requestId = UUID.randomUUID().toString();
+            List<String> accessKeys = client.listMachineUserAccessKeys(requestId, actorCrn, accountId, machineUserCrn, true);
+            return accessKeys.contains(accessKeyId);
+        }
+    }
+
+    /**
      * Gather anonymization rules for a specific account
      * NOTE: not supported yet on UMS side
      *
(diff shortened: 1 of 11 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -640,6 +640,24 @@ public void deleteMachineUserAccessKeys(String actorCrn, String accountId, Strin
...
+     * @param machineUserCrn machine user crn that own the access key
+     * @param accessKeyId access key id that we need to check
+     * @return result that is true if the machine user has the queried access key in UMS
+     */
+    public boolean doesMachineUserHasAccessKey(String actorCrn, String accountId,
+            String machineUserCrn, String accessKeyId) {
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
+            String requestId = UUID.randomUUID().toString();
+            List<String> accessKeys = client.listMachineUserAccessKeys(requestId, actorCrn, accountId, machineUserCrn, true);
```
lnardai, line 656: Isn't there a cheaper method of validating wether and access key exists works?
GetAccessKeyVerificationData? I'm not sure it's better, but listing all keys seems tedious.

lnardai: Also what happens if someone deletes the key or deactivates it? I'm not sure this check by list solves that problem.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
```diff
@@ -40,6 +40,27 @@ public AltusIAMService(GrpcUmsClient umsClient,
                         UserManagementProto.AccessKeyType.Value.ED25519)));
     }
 
+    /**
+     * Checks that machine user has a specific access key on UMS side
+     */
+    public boolean doesMachineUserHasAccessKey(String machineUserName, String accessKey, String actorCrn, String accountId,
```
lnardai, line 46: It's quite confusing that you did not kept the order of variables but used the same name for the method! Somewhere the actorCrn is first, but here machineUserName is the first sine. Since we are only using strings here it's easy to misuse these methods.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
oleewere: that's only listing the access keys that are assigned to the machine user + we need to first check the machine user existence itself anyway + with this no any new calls added. so i think that should be fine for now (can be refactored later)

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/service/AltusIAMService.java
oleewere: ok,right, im fixing it

lnardai (approved):

PR #9115, opened 2020-09-30
CB-8956 Updating Azure Medium Duty to use correct storage type instea…

orngejaket:
…d of a node type.


sxxgrc (approved):
Looks good.

PR #9116, opened 2020-09-30
CB-9055: Put NO HDFS feature under entitlement

LinaAtAustin:
We should put the code change to remove HDFS from Data Lake behind entitlement. So customer can enable this feature for new SDX cluster, and existing SDX cluster upgrade does not hit this feature by accident.

The flag is committed in thunderhead at https://github.infra.cloudera.com/thunderhead/thunderhead/commit/aefabb467c204d196a3a421aa015e8895800137d#diff-78fd6bd58147c45d045cc080c7400665

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.cmtemplate.configproviders.hbase;
 
+import static com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider.INTERNAL_ACTOR_CRN;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_2_2;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.isVersionNewerOrEqualThanLimited;
 import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.config;
@@ -7,9 +8,13 @@ import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.c
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.inject.Inject;
+
 import org.springframework.stereotype.Component;
 
 import com.cloudera.api.swagger.model.ApiClusterTemplateConfig;
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
+import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
 import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateComponentConfigProvider;
 import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateProcessor;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils;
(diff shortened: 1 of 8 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
```diff
@@ -48,9 +55,12 @@ public boolean isConfigurationNeeded(CmTemplateProcessor cmTemplateProcessor, Te
                 .orElse(false);
         String cdhVersion = getCdhVersion(source);
         boolean is722OrNewer = isVersionNewerOrEqualThanLimited(cdhVersion, CLOUDERAMANAGER_VERSION_7_2_2);
+        GeneralClusterConfigs generalClusterConfigs = source.getGeneralClusterConfigs();
+        boolean sdxHbaseCloudStorageEnabled = entitlementService == null ? true :
```
keyki, line 59: I don't think the EntitlementService can be null.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/general/GeneralClusterConfigsProvider.java
```diff
@@ -53,6 +54,9 @@ public GeneralClusterConfigs generalClusterConfigs(Stack stack, Cluster cluster)
         generalClusterConfigs.setKnoxUserFacingCertConfigured(userFacingCertHasBeenGenerated);
         generalClusterConfigs.setExternalFQDN(cluster.getFqdn());
         generalClusterConfigs.setEnableRangerRaz(cluster.isRangerRazEnabled());
+        generalClusterConfigs.setUserCrn(stack.getEnvironmentCrn());
```
keyki, line 57: I don't think it will work with the environment crn. The stack has a creator crn instead, but I still recommend to use the `com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider.INTERNAL_ACTOR_CRN` here.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
LinaAtAustin: In the test CentralCmTemplateUpdaterTest, I debug into the code, and it is null. That is why I put the condition there to let the test pass. I suspect in real end-to-end test, EntitlementService will not be null. 

What is alternative for CentralCmTemplateUpdaterTest to pass without allowing EntitlementService to be null?

keyki: You need to include the service in the test as well
```
@Mock
    private EntitlementService entitlementService;
```
and it won't be null

LinaAtAustin: The code is updated based on your suggestion.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/general/GeneralClusterConfigsProvider.java
LinaAtAustin: Now, I am using INTERNAL_ACTOR_CRN as crn and ThreadBasedUserCrnProvider.getAccountId() as account ID. This is how entitlement is checked at
https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/instancegroup/template/InstanceTemplateV4RequestToTemplateConverter.java#L66

keyki (approved):

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/general/GeneralClusterConfigsProvider.java
lajosrodek: Hmm, nice trick! I was not aware of `ThreadBasedUserCrnProvider.getAccountId()`, but it is definitely smarter this way. 😃

keyki: unless there is an issue and it contains the wrong account id :D

PR #9119, opened 2020-10-01
CB-9044 Add heartbeat success logging

bergerdenes:
as it was missing and not traceable

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/common/metrics/type/MetricType.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/common/metrics/type/MetricType.java
@@ -6,6 +6,7 @@ public enum MetricType implements Metric {
     VAULT_WRITE("vault.write"),
     VAULT_WRITE_FAILED("vault.write.failed"),
     VAULT_DELETE("vault.delete"),
+    HEARTBEAT_UPDATE_SUCCESS("heartbeat.update.success"),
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/common/metrics/type/MetricType.java
```diff
@@ -6,6 +6,7 @@
     VAULT_WRITE("vault.write"),
     VAULT_WRITE_FAILED("vault.write.failed"),
     VAULT_DELETE("vault.delete"),
+    HEARTBEAT_UPDATE_SUCCESS("heartbeat.update.success"),
```
doktoric, line 9: do we need any modification on the metrics in datadog ?

lnardai: do we have a default mapper for this Enum?

lnardai (approved):

common/src/main/java/com/sequenceiq/cloudbreak/common/metrics/type/MetricType.java
bergerdenes: - metrics modification is not needed, this new metric however can be used to calculate a success/fail ratio
- no default enum is needed in this case as the enum used only for micrometer metric publishing (which either happens or does not)

PR #9126, opened 2020-10-01
CB-9044 Flow handling race condition fix on flow redistribution

bergerdenes:
The issue was caused when a failed node resumes operation but in the
meantime another node picked up the same flow.
The fix is that the resumed node does not continue progressing the
flow if the last operation is performed by another node, rather
removing the paticular flow from the inmemory store and skipping
flow continuation.

### DIFF
--- a/flow/src/main/java/com/sequenceiq/flow/cleanup/InMemoryCleanup.java
+++ b/flow/src/main/java/com/sequenceiq/flow/cleanup/InMemoryCleanup.java
@@ -1,0 +1,40 @@
+package com.sequenceiq.flow.cleanup;
+
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.scheduler.PollGroup;
+import com.sequenceiq.cloudbreak.cloud.store.InMemoryResourceStateStore;
+import com.sequenceiq.flow.core.FlowRegister;
+import com.sequenceiq.flow.core.chain.FlowChains;
+
+@Component
+public class InMemoryCleanup {
+
+    @Inject
+    private FlowRegister runningFlows;
+
+    @Inject
+    private FlowChains flowChains;
+
+    public void cancelEveryFlowWithoutDbUpdate() {
+        for (String resourceType : InMemoryResourceStateStore.getResourceTypes()) {
+            for (Long resourceId : InMemoryResourceStateStore.getAllResourceId(resourceType)) {
+                InMemoryResourceStateStore.putResource(resourceType, resourceId, PollGroup.CANCELLED);
+            }
+        }
+        for (String id : runningFlows.getRunningFlowIds()) {
+            cancelFlowWithoutDbUpdate(id);
+        }
+    }
+
+    public void cancelFlowWithoutDbUpdate(String flowId) {
+        String flowChainId = runningFlows.getFlowChainId(flowId);
+        if (flowChainId != null) {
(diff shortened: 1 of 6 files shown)


flow/src/test/java/com/sequenceiq/flow/core/Flow2HandlerTest.java
```diff
@@ -150,6 +152,12 @@
     @Mock
     private SpanContext spanContext;
 
+    @Mock
```
doktoric, line 155: are these fields not called at all in the tests ? I guess some verification should be useful

flow/src/main/java/com/sequenceiq/flow/cleanup/InMemoryCleanup.java
```diff
@@ -0,0 +1,44 @@
...
+            for (Long resourceId : InMemoryResourceStateStore.getAllResourceId(resourceType)) {
+                InMemoryResourceStateStore.putResource(resourceType, resourceId, PollGroup.CANCELLED);
+            }
+        }
+        for (String id : runningFlows.getRunningFlowIds()) {
+            String flowChainId = runningFlows.getFlowChainId(id);
+            if (flowChainId != null) {
+                flowChains.removeFullFlowChain(flowChainId);
+            }
+            runningFlows.remove(id);
```
topolyai5, line 32: this is the same with `cancelFlowWithoutDbUpdate`

topolyai5 (changes requested):

flow/src/main/java/com/sequenceiq/flow/core/Flow2Handler.java
```diff
@@ -184,11 +193,24 @@ private void handleFlowControlEvent(String key, Payload payload, FlowParameters
...
+                    } else {
+                        LOGGER.warn("Flow {} was handled by another node {}, current node ID is {}, abandoning.",
+                                flow.getFlowId(), flowLog.getCloudbreakNodeId(), nodeId);
+                        inMemoryCleanup.cancelFlowWithoutDbUpdate(flow.getFlowId());
+                        isFlowCancelled.setTrue();
+                    }
+                });
             });
-            flow.sendEvent(key, flowParameters.getFlowTriggerUserCrn(), payload, flowParameters.getSpanContext());
+            if (!isFlowCancelled.booleanValue()) {
```
topolyai5, line 211: why don't send this inside the lamda?

sodre90: it may take some time to consume events, and it shouldn't be in a transaction then

topolyai5: I see, thanks

bergerdenes: because that would change the logic: the sendEvent is necessary in the case lastFlowLog is not present

flow/src/main/java/com/sequenceiq/flow/cleanup/InMemoryCleanup.java
bergerdenes: true, reused.

flow/src/test/java/com/sequenceiq/flow/core/Flow2HandlerTest.java
bergerdenes: wrote a test

topolyai5 (approved):

PR #9132, opened 2020-10-01
CB-9055: Put HBase Cloud Storage feature under entitlement

LinaAtAustin:
Cherry-pick the commit at master to 2.31.0 branch. https://github.com/hortonworks/cloudbreak/commit/b1f5b8a9ccb2c2b8b7e78c99fa52d640917faade

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
@@ -74,6 +74,9 @@ public class EntitlementService {
     @VisibleForTesting
     static final String CDP_UMS_USER_SYNC_MODEL_GENERATION = "CDP_UMS_USER_SYNC_MODEL_GENERATION";
 
+    @VisibleForTesting
+    static final String CDP_SDX_HBASE_CLOUD_STORAGE = "CDP_SDX_HBASE_CLOUD_STORAGE";
+
     @Inject
     private GrpcUmsClient umsClient;
 
@@ -155,6 +158,10 @@ public class EntitlementService {
 
     public boolean umsUserSyncModelGenerationEnabled(String actorCrn, String accountId) {
         return isEntitlementRegistered(actorCrn, accountId, CDP_UMS_USER_SYNC_MODEL_GENERATION);
+    }
+
+    public boolean sdxHbaseCloudStorageEnabled(String actorCrn, String accountId) {
+        return isEntitlementRegistered(actorCrn, accountId, CDP_SDX_HBASE_CLOUD_STORAGE);
     }
 
     public List<String> getEntitlements(String actorCrn, String accountId) {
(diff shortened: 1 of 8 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
```diff
@@ -74,6 +74,9 @@
     @VisibleForTesting
     static final String CDP_UMS_USER_SYNC_MODEL_GENERATION = "CDP_UMS_USER_SYNC_MODEL_GENERATION";
 
+    @VisibleForTesting
+    static final String CDP_SDX_HBASE_CLOUD_STORAGE = "CDP_SDX_HBASE_CLOUD_STORAGE";
```
doktoric, line 78: CDP_SDX_NO_HDFS ?

LinaAtAustin: That is the initial name I came up. I have discussed with Karthik offline. The reasons we use the current name are
1) We prefer to use positive tone for the entitlement name. So customers perceive they are getting benefits instead of losing some features. 
2) The main purpose of removing HDFS from Data Lake is to store HBase data to cloud storage for public cloud deployment. Therefore, we can reduce resources (such as CPU, memory) used by HDFS, and cut cost. The current name highlights this purpose.

doktoric: @LinaAtAustin I see the point but lets see a different aspect. Somebody who join to team and will see this entitlement, do you think it is expressive enough? Maybe it is only my view :)

kmanamcheri: @doktoric so the issue is that we are making this in phases. Currently its not actually "NO HDFS" because HDFS is still going to be there. This change particularly only moves HBase data onto cloud storage. Once that is done, we have to change the configurations of the services to not depend on HDFS service, move Solr storage etc.

In the long term, we don't intend to keep these entitlements around. Once all the parts are completed, we'll remove these entitlements and make this the default behavior.

> Somebody` who join to team and will see this entitlement, do you think it is expressive enough
NO_HDFS is not correct though 😄  I think that would cause confusion if something thinks, "I enabled NO_HDFS, why is there HDFS" 😅

kmanamcheri: FYI @doktoric this functional change is already merged (and has been merged for a while). We forgot to put it behind an entitlement 😨 😅

doktoric: @kmanamcheri ok so the Jira ticket confused me already this is why I thought no hdfs on the datalake cluster. @LinaAtAustin  can you please correct the commit message?

LinaAtAustin: @doktoric I have changed the title of this jira to "Put HBase Cloud Storage feature under entitlement", and the description in "https://jira.cloudera.com/browse/CB-9055". Is that enough?

lnardai (approved):
Talked with the team, we are okay with this for now, but next time we are going to revert the original feature and ask you to recreate it with another PR behind the entitlement on master.

PR #9138, opened 2020-10-02
CB-9129 We want to save the structured event if the resource crn is null

topolyai5:


horadla23 (approved):
LGTM

biharitomi (approved):
LGTM

PR #9139, opened 2020-10-02
CB-9060 Check if certificates on cluster are about to expire

lacikaaa:
We are setting up clusters with autotls, which means every host has it's own certificate.
These certificates expiry every year and we must detect it and notify the customer before expiration date.
In this commit `StackStatusCheckerJob` has been enriched with checking the result from Cloudera Manager for
host certificate expiration related health info. Although these are not stored per host, they are aggregated
and if there is at least one which is expiring we are setting a new field in `Cluster` entity, `certExpirationState`
from `VALID` to `HOST_CERT_EXPIRING`.
This field is an enum to allow further expansion if we later would like to indicate eg CMCA expiration.
The value of this field is available on DistroX API in `ClusterV4Response` and on SDX API in `SdxClusterDetailResponse`.

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
@@ -36,6 +36,8 @@ import com.cloudera.api.swagger.model.ApiRole;
 import com.cloudera.api.swagger.model.ApiRoleState;
 import com.cloudera.api.swagger.model.ApiService;
 import com.cloudera.api.swagger.model.ApiServiceState;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimaps;
 import com.google.common.collect.Sets;
 import com.sequenceiq.cloudbreak.client.HttpClientConfig;
 import com.sequenceiq.cloudbreak.cloud.model.HostName;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
```diff
@@ -185,20 +189,24 @@ private void buildMdcContext(Stack stack) {
...
-        syncInstances(stack, runningInstances, failedInstances, defaultState);
+        ifFlowNotRunning(() -> updateStates(stack, failedInstances, newFailedNodeNames, newHealtyHostNames, hostCertExpiring));
+        syncInstances(stack, runningInstances, failedInstances, InstanceSyncState.RUNNING);
+    }
+
+    private void updateStates(Stack stack, Collection<InstanceMetaData> failedInstances, Set<String> newFailedNodeNames, Set<String> newHealtyHostNames,
+            boolean hostCertExpiring) {
+        clusterService.updateClusterCertExpirationState(stack.getCluster(), hostCertExpiring);
+        clusterOperationService.reportHealthChange(stack.getResourceCrn(), newFailedNodeNames, newHealtyHostNames);
+        if (!failedInstances.isEmpty()) {
```
keyki, line 205: Can you please add a few log lines here?

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
```diff
@@ -215,16 +223,37 @@ public ClusterStatusResult getStatus(boolean blueprintPresent) {
...
 
     @Override
-    public Map<HostName, ClusterManagerState> getExtendedHostStatuses() {
-        Map<HostName, ClusterManagerState> result = new HashMap<>();
-        getHostHealth().forEach((hostname, health) ->
-                result.put(hostname, new ClusterManagerState(healthSummaryToState(health.getSummary()),
-                        getHostHealthMessage(health.getSummary(), health.getExplanation()))));
-        return result;
+    public ExtendedHostStatuses getExtendedHostStatuses() {
+        Map<HostName, Collection<ApiHealthCheck>> hostHealth = getHostHealth();
```
keyki, line 227: Can you please add a few log lines to these methods?

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
lacikaaa: added

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
lacikaaa: added

PR #9140, opened 2020-10-02
CB-9130 Single RG is deleted when using new network

gergopapi2:
On azure, when using single RG, then deleting the new network during environment deletion should delete the VNET resource but must not delete any resource groups (RG). With the recent introduction of resource group usage pattern USE_SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT the check whether using a single RG or not was broken and was assumed that it is MULTIPLE, hence the single RG was deleted.


pdarvasi (approved):

horadla23 (approved):
LGTM

PR #9141, opened 2020-10-02
CB-8547: Use entitlement for FreeIPA healthcheck for cluster proxy

jamisonbennett:
When the FreeIPA health check entitlement is enabled, then register
the FreeIPA healthcheck with cluster proxy for new clusters.

Continue using the old login page as the status check for clusters
that were provisioned prior to the healthcheck's existence or for
accounts without the entitlement.

This was already reviewed and committed here https://github.com/hortonworks/cloudbreak/pull/8871. But it was reverted due to flakey test failure. It is now under and entitlement to prevent any issues until a new version of cluster proxy and a new version of health checks are added.


christmasferret (approved):
LGTM

jamisonbennett:
This was already reviewed and approved here #8871. So I am going to commit without additional reviewers.

PR #9150, opened 2020-10-06
CB-9155 and CB-9147

doktoric:

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/NetworkDeleteHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/NetworkDeleteHandler.java
@@ -61,7 +61,7 @@ public class NetworkDeleteHandler extends EventSenderAwareHandler<EnvironmentDel
                     if (RegistrationType.CREATE_NEW == registrationType) {
                         environmentNetworkService.deleteNetwork(environmentDtoConverter.environmentToDto(environment));
                     }
-                    network.setName(network.getName() + "_DELETED_@_" + System.currentTimeMillis());
+                    network.setName(environment.getName() + "_network_DELETED_@_" + System.currentTimeMillis());
                     environmentService.save(environment);
                 }
             });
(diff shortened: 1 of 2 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/NetworkDeleteHandler.java
```diff
@@ -61,7 +61,7 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
                     if (RegistrationType.CREATE_NEW == registrationType) {
                         environmentNetworkService.deleteNetwork(environmentDtoConverter.environmentToDto(environment));
                     }
-                    network.setName(network.getName() + "_DELETED_@_" + System.currentTimeMillis());
+                    network.setName(environment.getName() + "_network_DELETED_@_" + System.currentTimeMillis());
```
topolyai5, line 64: it is the same then earlier. Much better if you check` _DELETED_@_timestamp` is in the name or not.

topolyai5 (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/NetworkDeleteHandler.java
doktoric: why is this the same?
we dont modify the env name until we can not delete

topolyai5: if you see the desc in the jira: `gtlocal1_DELETED_@1601902245809_DELETED@1601903025413_DELETED@1601903612895_DELETED@1601904217274_DELETED@1601904607194_DELETED@1601905595570_DELETED@1601906596339_DELETED@1601906833696_DELETED@1601907190970_DELETED@1601908948322_DELETED@_1601923539012`
of course, this is a timeline error but I could delete only in debug mode (I updated first with less than 255 chars and save it)

topolyai5 (approved):

PR #9151, opened 2020-10-06
CB-9149. Splitting large (> 16K) JSON logs on CB level in order to av…

oleewere:
…oid log rotation related issues by docker logging driver on K8s

details:
- in order to avoid large messages use a 12K max message length for JSON based logging (12K can be safe as we have many other fields, not only message)
- split message into chunks based on the max chunk length
- add partial metadata about the chunks, that can be handled by fluentd
- using the same field names for partial metadata that is used by docker fluentd logging driver: partial_message, partial_id, partial_ordinal and partial_last
- refactor: removed loggerFilterName from masking pattern layout

additional info:
fluent logger for docker: https://github.com/moby/moby/blob/master/daemon/logger/fluentd/fluentd.go#L116
all the keys are deleted by fluent-conat plugin if the events are merged, see:  https://github.com/fluent-plugins-nursery/fluent-plugin-concat/blob/master/lib/fluent/plugin/filter_concat.rb#L149
if this is deployed, it will split json logs, so until fluent-concat plugin config is not updated, the logs will be in splitted in kibana (!! but you can find those based on the component names)

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/logger/LoggerConfiguration.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/logger/LoggerConfiguration.java
@@ -1,0 +1,31 @@
+package com.sequenceiq.cloudbreak.logger;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class LoggerConfiguration {
+
+    private LoggerConfiguration() {
+    }
+
+    public static Boolean isJsonFormatEnabled() {
+        return Boolean.parseBoolean(getProperty("logger.format.json.enabled", null));
+    }
+
+    public static String getJsonFormatMdcName() {
+        return getProperty("logger.format.json.mdc.name", "context");
+    }
+
+    public static Integer getSplitterMaxChunLength() {
+        return Integer.parseInt(getProperty("logger.appender.splitter.max.chunk.length", "12000"));
+    }
+
+    private static String getProperty(String property, String defaultValue) {
+        String envProperty = property.toUpperCase().replace(".", "_");
+        if (StringUtils.isNotEmpty(System.getenv(envProperty))) {
+            return System.getenv(envProperty);
+        } else if (StringUtils.isNotEmpty(System.getProperty(property))) {
+            return System.getProperty(property);
+        }
+        return defaultValue;
+    }
+}
--- a/common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
@@ -1,7 +1,13 @@
 package com.sequenceiq.cloudbreak.logger.format;
 
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.google.common.base.Splitter;
+import com.sequenceiq.cloudbreak.logger.PartialMessageMetadata;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.contrib.json.classic.JsonLayout;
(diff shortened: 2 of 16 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/logger/PropertyHelper.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+public class PropertyHelper {
+
+    private PropertyHelper() {
+    }
+
+    public static Boolean isJsonFormatEnabled() {
+        return Boolean.parseBoolean(getProperty("logger.format.json.enabled", null));
```
biharitomi, line 11: Is there any reason why we are not using the @Value annotation of Spring like in all the other property holder components?

common/src/main/java/com/sequenceiq/cloudbreak/logger/PropertyHelper.java
```diff
@@ -0,0 +1,31 @@
+package com.sequenceiq.cloudbreak.logger;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class PropertyHelper {
```
biharitomi, line 5: I would rename it to indicates that this class holds the logger related configurations that comes either from property or not. For example LoggerConfiguration.java

common/src/test/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayoutTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Test;
+
+import ch.qos.logback.classic.spi.LoggingEvent;
+
+public class CustomJsonLayoutTest {
+
+    private CustomJsonLayout underTest;
+
+    @Before
+    public void setUp() {
```
biharitomi, line 16: Would you mind to create tests based on the size message bigger than the 12KB limit or not?

biharitomi (changes requested):

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
```diff
@@ -1,29 +1,65 @@
...
+        this.maxChunkLength = maxChunkLength;
     }
 
     public String doLayout(ILoggingEvent event, String fullLogMessage, String loggerNameFilter) {
-        Map map = toJsonMap(event, fullLogMessage, loggerNameFilter);
-        if (map == null || map.isEmpty()) {
+        if (StringUtils.length(fullLogMessage) > this.maxChunkLength) {
+            StringBuilder stringBuilder = new StringBuilder();
+            String chunkId = Integer.toHexString(event.hashCode());
+            Iterable<String> messages = Splitter.fixedLength(this.maxChunkLength).split(fullLogMessage);
```
lacikaaa, line 41: this `Splitter.fixedLength(this.maxChunkLength)` could be initiated in the constructor

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
```diff
@@ -1,29 +1,65 @@
...
+
+    CustomJsonLayout(String contextName, Integer maxChunkLength) {
         this.contextName = contextName;
+        this.maxChunkLength = maxChunkLength;
     }
 
     public String doLayout(ILoggingEvent event, String fullLogMessage, String loggerNameFilter) {
-        Map map = toJsonMap(event, fullLogMessage, loggerNameFilter);
-        if (map == null || map.isEmpty()) {
+        if (StringUtils.length(fullLogMessage) > this.maxChunkLength) {
```
lacikaaa, line 38: `this.` is unnecessary

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
```diff
@@ -1,29 +1,65 @@
...
+            StringBuilder stringBuilder = new StringBuilder();
+            String chunkId = Integer.toHexString(event.hashCode());
+            Iterable<String> messages = Splitter.fixedLength(this.maxChunkLength).split(fullLogMessage);
+            int currentIndex = 0;
+            Iterator<String> messageIterator = messages.iterator();
+            while (messageIterator.hasNext()) {
```
lacikaaa, line 44: you can save one line here with:
`for (int currentIndex = 0; messageIterator.hasNext(); currentIndex++) {`

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
```diff
@@ -1,29 +1,65 @@
...
+
+    private String toJsonRecordString(Map map) {
+        if (map.isEmpty()) {
             return null;
         }
         return toJsonString(map) + CoreConstants.LINE_SEPARATOR;
     }
 
-    private Map toJsonMap(ILoggingEvent event, String fullLogMessage, String loggerNameFilter) {
+    private Map toJsonMap(ILoggingEvent event, String fullLogMessage, String loggerNameFilter, PartialMessageMetadata partialMessageMetadata) {
```
lacikaaa, line 62: - `loggerNameFilter` is not used
- method definition should look like: `private Map<String, Object> toJsonMap`

`toJsonRecordString` and `toJsonString` parameter is also missing the same

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/JsonLayoutFormat.java
```diff
@@ -10,8 +10,8 @@
 
     private final CustomJsonLayout jsonLayout;
 
-    JsonLayoutFormat(String contextName) {
-        this.jsonLayout = new CustomJsonLayout(contextName);
+    JsonLayoutFormat(String contextName, Integer maxChunkLength) {
+        this.jsonLayout = new CustomJsonLayout(contextName, maxChunkLength);
```
lacikaaa, line 14: `this.` can be removed

lacikaaa (changes requested):

common/src/main/java/com/sequenceiq/cloudbreak/logger/PropertyHelper.java
oleewere: that is for logging, it's needed right before Spring bootstrapping

common/src/main/java/com/sequenceiq/cloudbreak/logger/format/CustomJsonLayout.java
oleewere: sure i can remove `loggerNameFilter`, it seems anonymiziation was removed by @biharitomi  before, and the parameter was not cleaned up from `MaskingPatternLayout`

oleewere: with the Map, I have followed the `JsonLayoutBase` class, but that the same anyway ..so whatever, i can change that

common/src/main/java/com/sequenceiq/cloudbreak/logger/LoggerConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+    public static String getJsonFormatMdcName() {
+        return getProperty("logger.format.json.mdc.name", "context");
+    }
+
+    public static Integer getSplitterMaxChunLength() {
+        return Integer.parseInt(getProperty("logger.appender.splitter.max.chunk.length", "12000"));
+    }
+
+    private static String getProperty(String property, String defaultValue) {
+        String envProperty = property.toUpperCase().replace(".", "_");
```
lacikaaa, line 23: I don't really get this. so there are 3 properties above, you wrote them with lowercase and dot separated and right after you convert them. wouldn't be it easier just to have them written the way we expect them?

lacikaaa (approved):

common/src/main/java/com/sequenceiq/cloudbreak/logger/LoggerConfiguration.java
oleewere: we use this both as a system property + environment variable, first we are checking the environment property, lower we are simply checking the system property (system property used mostly locally, env vars are used for qa/mow deployments), we can do the opposite here, but I think it does not really matter.

common/src/main/java/com/sequenceiq/cloudbreak/logger/PropertyHelper.java
biharitomi: This is a spring boot application so even the logging framework is initialised by Spring boot framework. 
I would give it a try to use @Value annotation instead of this and let spring to build the dependencies in the right order(starting with logger dependencies)
On the other hand it would be a much cleaner code and you shouldn't have to deal with env variables and properties as @lacikaaa took this into the focus in his review.
what do you think @lacikaaa ?

common/src/main/java/com/sequenceiq/cloudbreak/logger/LoggerConfiguration.java
biharitomi: That's why we should use Spring instead of implementing a Spring feature this way....

biharitomi: sry, it is not supported by spring

common/src/main/java/com/sequenceiq/cloudbreak/logger/PropertyHelper.java
biharitomi:
```

Since logging is initialized before the ApplicationContext is created, it is not possible to control logging from @PropertySources in Spring @Configuration files. The only way to change the logging system or disable it entirely is via System properties.
--
```
Sry, it is not supported, approving this PR

biharitomi (approved):

PR #9157, opened 2020-10-06
CB-9171 Associate private network and root user for GCP RDS instances

doktoric:


doktoric:
thanks for the fix @biharitomi feel free to amend the commit

topolyai5 (approved):

PR #9158, opened 2020-10-06
CB-8982 Indefinitely long loop when deleting environment. When we fet…

topolyai5:
…ched the cluster definition, we checked the default templates but this step was unnecessary, because we should delete the user managed only.

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
@@ -47,6 +47,13 @@ public interface ClusterTemplateV4Endpoint {
     ClusterTemplateViewV4Responses list(@PathParam("workspaceId") Long workspaceId);
 
     @GET
+    @Path("env/{environmentCrn}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = ClusterTemplateOpDescription.LIST_BY_ENV, produces = MediaType.APPLICATION_JSON,
+            notes = CLUSTER_TEMPLATE_NOTES, nickname = "listClusterTemplatesByEnv")
+    ClusterTemplateViewV4Responses listByEnv(@PathParam("workspaceId") Long workspaceId, @PathParam("environmentCrn") String environmentCrn);
+
+    @GET
     @Path("name/{name}")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = ClusterTemplateOpDescription.GET_BY_NAME_IN_WORKSPACE, produces = MediaType.APPLICATION_JSON, notes = CLUSTER_TEMPLATE_NOTES,
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ClusterTemplateV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ClusterTemplateV4Controller.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.controller.v4;
 
+import static com.sequenceiq.cloudbreak.util.Benchmark.measure;
 import static java.util.stream.Collectors.toSet;
 
 import java.util.Objects;
@@ -31,15 +32,16 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.clustertemplate.responses.Clust
 import com.sequenceiq.cloudbreak.api.endpoint.v4.clustertemplate.responses.ClusterTemplateV4Responses;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.clustertemplate.responses.ClusterTemplateViewV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.clustertemplate.responses.ClusterTemplateViewV4Responses;
-import com.sequenceiq.cloudbreak.api.endpoint.v4.dto.NameOrCrn;
 import com.sequenceiq.cloudbreak.api.util.ConverterUtil;
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 import com.sequenceiq.cloudbreak.common.service.TransactionService;
 import com.sequenceiq.cloudbreak.common.service.TransactionService.TransactionExecutionException;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.ClusterTemplate;
+import com.sequenceiq.cloudbreak.domain.stack.cluster.ClusterTemplateView;
 import com.sequenceiq.cloudbreak.service.blueprint.BlueprintService;
 import com.sequenceiq.cloudbreak.service.template.ClusterTemplateService;
+import com.sequenceiq.cloudbreak.service.template.ClusterTemplateViewService;
 import com.sequenceiq.cloudbreak.workspace.controller.WorkspaceEntityType;
 import com.sequenceiq.distrox.v1.distrox.service.EnvironmentServiceDecorator;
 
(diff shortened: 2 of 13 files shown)


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
```diff
@@ -46,6 +46,13 @@
             nickname = "listClusterTemplatesByWorkspace")
     ClusterTemplateViewV4Responses list(@PathParam("workspaceId") Long workspaceId);
 
+    @GET
+    @Path("get_by_env")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = ClusterTemplateOpDescription.LIST_BY_ENV, produces = MediaType.APPLICATION_JSON,
+            notes = CLUSTER_TEMPLATE_NOTES, nickname = "listClusterTemplatesByEnv")
+    ClusterTemplateViewV4Responses listByEnv(@PathParam("workspaceId") Long workspaceId, @QueryParam("environmentCrn") String environmentCrn);
```
gregito, line 54: why do we need to introduce another endpoint for this? the env crn is a query param which can be empty (seemingly). If this param would missing, tha call would go to the simple list endpoint (line 47).
IMO we shouldn't introduce this endpoint, we just need to extend the original list endpoint with this query param. since tha param itself is not a mandatory field, this endpoint extension shouldn't break the compatibility

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ClusterTemplateV4Controller.java
```diff
@@ -139,9 +148,8 @@ public ClusterTemplateV4Responses deleteMultiple(Long workspaceId, @ResourceName
         if (Objects.nonNull(names) && !names.isEmpty()) {
             clusterTemplates = clusterTemplateService.deleteMultiple(names, workspaceId);
         } else {
-            NameOrCrn environmentNameOrCrn = NameOrCrn.ofCrn(environmentCrn);
             Set<String> namesByEnv = clusterTemplateService
-                    .findAllByEnvironment(environmentNameOrCrn)
+                    .findAllByEnvironment(environmentCrn)
```
gregito, line 152: if you remove the input-independent solution (the usage of the NameOrCrn object) then please remove the unused environmentName arg also from the method arguments

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
horadla23: I asked Gergo to do that, since original endpoint is for listing in account and this is for listing by env and these are conpletely different from authorization point of view, also this endpoint will called by env service only so it is an internal call right now (later it is possible to authorize it for an env)

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ClusterTemplateV4Controller.java
topolyai5: Yes, it could be valid, but:
* only one place where this is called: `ClusterTemplateV4Controller.deleteMultiple -> ClusterTemplateService.findAllByEnvironment -> and call this function`
* if you see the deleteMultiple function, we never use the `name`:
```
NameOrCrn environmentNameOrCrn = NameOrCrn.ofCrn(environmentCrn);
Set<String> namesByEnv = clusterTemplateService
        .findAllByEnvironment(environmentNameOrCrn)
        .stream()
        .map(ClusterTemplate::getName)
        .collect(toSet());
```
* and this chain references causes this:
   `env service call multiple delete with crn -> we fetch the detailed env with this crn -> and use the name and crn to log` , does not look good.

I think nobody call with name because of last change about there January 2020.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
gregito: I see, but the original endpoint could work in both ways whether the env crn is given or not. If we'd like to follow your mentioned thought about exposing this new endpoint in a "public" way somewhere in the future, this exact same question could emerge again (namely: why do we have two endpoints for a seemingly quite identical behavior). And at that point, we may have to deprecate this endpoint which comes with the concern of breaking something that uses this endpoint (but we don't know about it, because it may be some experience or other embedding service).

Back to the AuthZ. What - component - decides that this given endpoint is an internal or an "external" endpoint?
I mean if the internal kind of behavior means only that we have to implement its mechanism in a different way one (or more) layer lower, I think my suggestion could have more weight.
If the decision of which service can call this - new - endpoint and which can't is in another component's hand then I have nothing to say here, it is fine to me.
This section is partly because most likely your general knowledge around our AuthN/AuthZ is much deeper than mine.

I would like to include more people into this if you don't mind to have/obtain a more diverse opinion.

topolyai5: yes you are right, but we need this because of the authorization. Please talk with @horadla23

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ClusterTemplateV4Controller.java
gregito: I have nothing against removing the usage of NameOrCrn from this method, my comment just pointed to the fact that we shouldn't let unused method arguments as we do in this case after this change

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/clustertemplate/ClusterTemplateV4Endpoint.java
gregito: @doktoric , @biharitomi , @daszabo could you please take a look at the abovementioned topic(s) and leave some opinion about it? thanks guys!

horadla23: ok so my point here is that even both of them are identical from developer/code point of view but it is different from user point of view. Original list gives all templates in the account/tenant, this new one gives them back by an environment . This is different from customer point of view and since we are authorizing calls on api, it is completely different from authz point of view. one of them should be authorized based on account and this new one should be authorized on env level. we had so much pain caused by these mixed endpoints during implementation of authorization so we should avoid mixing these slightly different calls into 1 method.

you can check readme of authorization-common module about this topic and also we will hold a demo about it.

we can involve team into this discussion of course.

(so this is not about that we need to differentiate them because one of them is internal, rather it is about api design which considers different kind of authorizations)

biharitomi: what if the env crn would be as a path param? Would it be feasible for both side? 😄  @horadla23 @gregito cc: @topolyai5 
Of course this endpoint's responses would be the subset of the list by tenant, but I think it is more acceptable this way.

gregito: using the crn as a path param would emphasize the difference between two endpoints thus this approach would be more acceptable to me 👍🏻

doktoric: I am also fine with an optional path param. One more thing If we add this ned function probably @andrasfelfoldi and @zoltanzseli can use this in case of DH creation page. This is a kind of optimization.

topolyai5: I don't support the env crn in the path, because we want to filter instead of separate the resource by env. I think the best solution an optional param on the list operation but this is a problem for authz team.

horadla23: if UI will use this, then that is one more reason to think about how UI will use it and how we will authorize this.

doktoric (approved):

doktoric:
restest this please

doktoric:
retest this please

PR #9159, opened 2020-10-06
CB-8209: Adding the Cloudbreak version to the environment API.

sxxgrc:
The main use case for this is to enable the describe-environment call in the CDP CLI to return the Cloudbreak version as well.
I want to specify that I am new to Cloudbreak development and am not very familiar with how to test this correctly to ensure it is set up correctly, nor am I familiar with how to update these API files on Thunderhead (I assume this is done automatically after a Cloudbreak release/promotion).

Any more information for this process would be very helpful and I am happy to answer any questions. Thanks!

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
@@ -108,6 +108,8 @@ public class EnvironmentModelDescription {
     public static final String PROXYCONFIG_NAME = "Name of the proxyconfig of the environment.";
     public static final String PROXYCONFIG_RESPONSE = "ProxyConfig attached to the environment.";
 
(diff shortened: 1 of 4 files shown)


doktoric (changes requested):
Based on the ticket this information should be persisted into the db so every environment will contains which env service version provisioned it. I only see the API part of this story here.

topolyai5:
If you want the current CB version, you can get from here: https://cloudera.cdp.mow-int.cloudera.com/cb/info
If you want the version of CB when the env is created, you should fetch from the db

sxxgrc:
> Based on the ticket this information should be persisted into the db so every environment will contains which env service version provisioned it. I only see the API part of this story here.

Hmm, to be honest I'm not exactly sure how to do that. Where can I change the information stored in the database per environment? 
And with respect to the other comment, how can I communicate with the DB? 
Do you guys mean that I should be: grabbing the Cloudbreak version from the DB, and then storing it along with the other environment information?

sxxgrc:
Okay I understand how to do this now, you can ignore my previous question. I will make a new PR when I'm done.

PR #9160, opened 2020-10-07
CDPCP-3122. Bulk user sync state retrieval fails over to default retr…

handavid:
…ieval

The UMS may protect itself by rejecting bulk computation requests for
large tenants. There is also a grpc timeout built into the system of
one minute after which the transport will be closed and the request
will fail.

In these cases, the ums user sync state model retrieval
should failover to the default approach.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/ums/UmsUsersStateProviderDispatcher.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/ums/UmsUsersStateProviderDispatcher.java
@@ -39,19 +39,38 @@ public class UmsUsersStateProviderDispatcher {
 
             boolean fullSync = userCrns.isEmpty() && machineUserCrns.isEmpty();
 
-            Map<String, UmsUsersState> envUsersStateMap;
             if (fullSync && entitlementService.umsUserSyncModelGenerationEnabled(INTERNAL_ACTOR_CRN, accountId)) {
-                envUsersStateMap = bulkUmsUsersStateProvider.get(accountId, environmentCrns, requestIdOptional);
+                return dispatchBulk(accountId, actorCrn, environmentCrns, userCrns, machineUserCrns,
+                        requestIdOptional, fullSync);
             } else {
-                envUsersStateMap = defaultUmsUsersStateProvider.get(
-                        accountId, actorCrn,
-                        environmentCrns, userCrns, machineUserCrns,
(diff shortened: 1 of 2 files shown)


horadla23:
what "default approach" means? Will it have performance impact for example on mow-dev? If yes, then we should target this change to master to fully test it .

handavid:
Our concern is that, if each request takes more than a minute, then running multiple simultaneous requests could consume all of the UMS' resources. We want the ability for the UMS to protect itself without needing to revisit the FMS code. E.g., the UMS could implement a limit on the number of bulk requests it accepts at a time or reject bulk requests for cases where we suspect we would exceed the 1 minute threshold without breaking user sync in the FMS.

We've identified a number of inefficiencies in the bulk calculation in the UMS. This is because we started with a direct translation of the work that the FMS was doing. See https://jira.cloudera.com/browse/CDPCP-3078 for more details on areas of improvement on the UMS side. Essentially there are some data retrieval operation that are still performed on a per-actor basis that can be replaced. Our intention is to target and test this feature against 2000 actors.

In the meantime, we discovered (from https://jira.cloudera.com/browse/CDPCP-2986) that there are an excessive number of actors in mow-dev due to orphaned internal machine users (some 8000+ actors total) which far exceeded our expectations. We are looking at ways to pull back or modify that change as well while we understand where all the orphan internal machine users are coming from.

cc: @giladwolff @mdzz

lacikaaa (approved):
I'm ok with this change, but raven team is the 2.31 branch guardian, they have to decide if they let it in or you have to move this to master.

lacikaaa:
retest this please

horadla23:
if current behaviour on 2.31 does not cause any issue on PROD, then I would put this change into master to have time to test it properly

horadla23 (changes requested):
Same here: I have checked related jira ticket and it seems for me that this improvement will have effect mostly on mow-dev, so I wouldn't put this into the upcoming CB release, I think master will branched soon to 2.32 and it will reach mow-dev soon, so please rebase this to master and change PR accordingly

horadla23:
@mdzz has user sync changed somehow in 2.31 ? are there such amount of machine users in prod which can affect customers negatively compared to 2.29 ? My concern is almost same here as in https://github.com/hortonworks/cloudbreak/pull/9166 that 2.31 release coming soon and I cannot tell if this change won't cause any issue in prod. It would be better to have time to test and experience this change in mow-dev before release to avoid any unexpected issues in prod.

handavid:
retest this please

handavid:
@horadla23 I retargeted this to master. can you merge if acceptable?

horadla23 (approved):

PR #9165, opened 2020-10-07
CB-8923 Introduce a custom JUnit XML Report Listener

aszegedi:
Introduce a custom JUnit XML Report Listener to override the original ReportNG one. So we can solve the issue: `we cannot check the details with the 'Show/Hide Errors'`
Beyond this I customised the `tests.name` at the test suite YAML files, because of the TestNG generated XML results are saved with the same file name. On the Jenkins these caused troubles with spaces in the names.


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -1133,6 +1139,29 @@ public void handleExceptionsDuringTest(TestErrorLog testErrorLog) {
             String errorMessage = errorLogMessageProvider.getMessage(exceptionMap, clues);
             testErrorLog.report(LOGGER, errorMessage);
 
+            ITestResult testResult = Reporter.getCurrentTestResult();
+            Throwable testFailException = errorLogMessageProvider.getException(exceptionMap);
+
+            if (testFailException != null) {
+                testResult.setThrowable(testFailException);
+                testResult.setTestName(getTestMethodName().get());
+                testResult.setStatus(2);
```
lnardai, line 1148: What does status 2 mean? Why 2 exactly?

aszegedi: This means [ITestResult.FAILURE](https://www.javadoc.io/doc/org.testng/testng/6.8.21/constant-values.html#org.testng.ITestResult.FAILURE)

Here we have to set the correct test results for test, because of the our TestContext is not [ITestContext](ITestContext) indeed. So our project generates all the errors and related messages as Reporter Output.

lnardai: So, can we use the Enum from there, or not a variable that is more descriptive?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/listener/ReportListener.java
```diff
@@ -24,23 +23,31 @@
...
         log(tr);
+        Reporter.setCurrentTestResult(tr);
+        super.onTestFailure(tr);
     }
 
     @Override
     public void onTestSuccess(ITestResult tr) {
         logUrl(tr);
         logMeasurements(tr);
+        Reporter.setCurrentTestResult(tr);
```
lnardai, line 39: Just asking, would this work correctly with parallel runs since it's static?

lnardai: Just checked in the code, it uses ThreadLocal internally.

aszegedi: Each test method and its respective before and after method is executed in a different thread. This is identified by the ID of the thread that is printed on the console.
You can find some additional details, for example: [TestNG parallel execution of tests, classes and suites](https://howtodoinjava.com/testng/testng-executing-parallel-tests/)

aszegedi: You were scratcing the ugly truth: After I've tested this out with several threads, I had to realised the original solution wasn't able to get the correct throwable to the defined test case. So I've applied the same prefixing as we already have for `searchUrl`.
So finally not the `Reporter` caused the issue as rather than `custom.results.xml` template.

aszegedi: Here is the test upload for Quanta: [https://quanta.infra.cloudera.com/?#/runDetails?gtn=6369877](https://quanta.infra.cloudera.com/?#/runDetails?gtn=6369877)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/ErrorLogMessageProvider.java
```diff
@@ -22,6 +23,13 @@
 
     private static final ObjectWriter OBJECT_WRITER = new ObjectMapper().writerWithDefaultPrettyPrinter();
 
+    public Throwable getException(Map<String, Exception> exceptionsDuringTest) {
+        return exceptionsDuringTest.values().stream()
+                .filter(exception -> exception.getCause() instanceof TestFailException)
```
lnardai, line 28: This will only show the first exception that we encounter in the test, is that correct?
Or we are specifically looking for this type of exceptions among them?
If so this method should be rather, findFirstTestFailException.

integration-test/src/main/resources/testsuites/e2e/aws-distrox-baseimage-tests.yaml
```diff
@@ -1,5 +1,5 @@
 name: "aws-distrox-baseimage-tests"
 tests:
-  - name: "aws distrox e2e base image tests"
+  - name: "aws_distrox_e2e_base_image_tests"
```
lnardai, line 3: I think this will require changes in jenkins for all jobs right?

lnardai:
Overall LGTM, check the jenkins result there is only an additional 10MB of data there.

lnardai (approved):

PR #9166, opened 2020-10-07
CDPCP-3121 - Filter internal machine users for user sync

handavid:

This changes the flag we use when retrieving machine users. Rather than retrieving all internal machine users, most of which are not workload identities, we've added another flag to the usermanagement.proto to indicate that we want all the workload machine users.

This change is needed because we've discovered that we have an unknown and likely large number of orphaned internal machine users. Processing all internal machine users negatively impacts user sync performance. E.g., in mow-dev, we move from syncing 2500 actors to 8000+ actors.

### DIFF
--- a/auth-connector/src/main/proto/usermanagement.proto
+++ b/auth-connector/src/main/proto/usermanagement.proto
@@ -137,7 +137,12 @@ service UserManagement {
 
   // Create an email based account.
   rpc CreateC1CAccount (CreateC1CAccountRequest)
-  returns (CreateC1CAccountResponse) {}
+    returns (CreateC1CAccountResponse) {}
+
+  // An endpoint called when a user verifies their email by clicking the link we sent
+  // them.
+  rpc VerifyC1CEmail (VerifyC1CEmailRequest)
+    returns (VerifyC1CEmailResponse) {}
 
   // Grant Entitlement to an Account
   rpc GrantEntitlement (GrantEntitlementRequest)
@@ -438,6 +443,10 @@ service UserManagement {
   // List cloud identity mappings for service principals.
   rpc ListServicePrincipalCloudIdentities (ListServicePrincipalCloudIdentitiesRequest)
     returns (ListServicePrincipalCloudIdentitiesResponse) {}
+
+  // Retrieves the CRN of identity provider connector used for CDP initiated logins.
+  rpc GetDefaultIdentityProviderConnector (GetDefaultIdentityProviderConnectorRequest)
+      returns (GetDefaultIdentityProviderConnectorResponse) {}
 
   // Get user sync state model, including actors, groups, workload administration groups, etc
   rpc GetUserSyncStateModel (GetUserSyncStateModelRequest)
@@ -1328,10 +1337,21 @@ message CreateC1CAccountRequest {
   string firstName = 2;
   // This may be an empty string.
   string lastName = 3;
+  // The user's password;
+  string password = 4 [(options.FieldExtension.sensitive) = true];
 }
 
 message CreateC1CAccountResponse {
   Account account = 1;
+}
+
+message VerifyC1CEmailRequest {
+  // The registration token passed by the caller
+  string registrationToken = 1 [(options.FieldExtension.sensitive) = true];
+}
+
+message VerifyC1CEmailResponse {
+  string userCrn = 1;
 }
 
 message AccountId {
@@ -1419,6 +1439,7 @@ message AssignRoleRequest {
 }
 
 message AssignRoleResponse {
+  string assigneeCrn = 1;
 }
 
 message UnassignRoleRequest {
@@ -1430,6 +1451,7 @@ message UnassignRoleRequest {
 }
 
 message UnassignRoleResponse {
+  string assigneeCrn = 1;
 }
 
 message ListAssignedRolesRequest {
@@ -1458,6 +1480,7 @@ message AssignResourceRoleRequest {
 }
 
 message AssignResourceRoleResponse {
+  string assigneeCrn = 1;
 }
 
 message UnassignResourceRoleRequest {
@@ -1470,6 +1493,7 @@ message UnassignResourceRoleRequest {
 }
 
 message UnassignResourceRoleResponse {
+  string assigneeCrn = 1;
 }
 
 message ListAssignedResourceRolesRequest {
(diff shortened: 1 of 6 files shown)


auth-connector/src/main/proto/usermanagement.proto
```diff
@@ -1703,6 +1727,13 @@ message ListMachineUsersRequest {
   bool includeInternal = 5;
   // Whether to include deleted machine users.
   bool includeDeleted = 6;
+  // Whether to include workload machine users. All regular machine users are
+  // workload machine users. This is a subset of the internal machine users,
+  // which have a password set. This field has no effect if internal machine
+  // users are requested because all workload machine users will already be
+  // included in the results. Note that this has no impact when listing by name
+  // or CRN. For those calls, internal machine users are always returned.
+  bool includeWorkloadMachineUsers = 7;
```
handavid, line 1736: this is the needed proto change. the rest of the proto is just to keep it in sync with the UMS.

horadla23:
same applies here: if current behaviour on 2.31 does not cause any issue on PROD, then I would put this change into master to have time to test it properly

horadla23 (changes requested):
I have checked related jira ticket and it seems for me that this improvement will have affect mostly on mow-dev, so I wouldn't put this into the upcoming CB release, I think master will branched soon to 2.31 and it will reach mow-dev soon, so please rebase this to master and change PR accordingly

lacikaaa (approved):

handavid:
@horadla23 We really would like this change in 2.31. Otherwise, we are looking at an unknown, potentially large jump in the load on the UMS in prod due to the orphaned internal machine users that we just found out about.
cc: @mdzz

mdzz:
@horadla23 The change is also blocking DWX (R10) release which is why we are trying to get it included into CB2.31.

horadla23:
@handavid @mdzz My concern is here that I am not sure if we need machine user to synced in Freeipa for example for logging (what do you think @oleewere ) or for QE or QAAS test (@lnardai @sodre90 what do you think?).
So in general lot of people are using machine users and I also cannot tell how customers using them.
I do not think that there are lot of machine users in prod's tenants so it should not affect prod's performance significantly, regarding mow-dev we need to clean them out.
So in general I need more information about machine user usage and I think we need to prove that this won't break anything, but based on release schedule 2.31 is not the best choice, since the plan to release it soon. At least I need managers to discuss if this can go in or not.

handavid:
Most machine users do not need to be synced. They are used to generate access keys and communicate with the control plane. We originally made the change to support DWX's use case. DWX is the first use case that we know of that requires an internal machine user to be a workload user.

subratanandi:
This is super important for DWX R10 release which is planned to go in the next couple of days.

sodre90:
This is super important for us to not break anything. It seems 2.31.0-b106 will be promoted to production and it will not contain this commit. You can create a hotfix later.

lnardai:
I agree with Peter and Adam, that changes like this should not be forced in to CB-2.31.0 last minute since we are at the last moment of promoting it too int and then to stage. I understand that it is important for some DWX feature, but it's also important for other features to reach prod in time and we are delaying those with such changes.

@handavid 
"DWX is the first use case that we know of that requires an internal machine user to be a workload user."
Could you elaborate on this? Does this machine user sync was introduced between CB 2.29 and 2.31.0? If we skip this would it degrade Production performance or would it be as of today?

horadla23:
One suggestion from @lacikaaa is that we should but this change behind entitlement and then you can turn it off by default, and after it is tested on mow-dev, etc., then you can turn it on.

giladwolff:
> I agree with Peter and Adam, that changes like this should not be forced in to CB-2.31.0 last minute since we are at the last moment of promoting it too int and then to stage. I understand that it is important for some DWX feature, but it's also important for other features to reach prod in time and we are delaying those with such changes.
> 
> @handavid
> "DWX is the first use case that we know of that requires an internal machine user to be a workload user."
> Could you elaborate on this? Does this machine user sync was introduced between CB 2.29 and 2.31.0? If we skip this would it degrade Production performance or would it be as of today?

This must go in. Applications including CB are leaking internal machine users. We have no idea how bad the situation is. This may cause large outage as user-sync requests come in and we need to scan and return large number of machine users. If this is not allowed to go in I will turn off listing of internal machine users and that will break dfx. Your call.

handavid:
@sodre90 can you tell me about the hotfix process?
If you can't accept this fix, then we need to revert the original change that requests internal machine users. Would we be able to get this change in through the mentioned hotfix process? How long does that process take? 

cc: @subratanandi heads up on delays for DWX

mdzz:
I'd like to add more details around the dependencies here:
- CDPCP-2986 is **already included** in CB2.31. This is the change that starts syncing internal machine users.
- DWX R10 release depends on CDPCP-2986 as stated in the discussion. According to @subratanandi, the dependent part in DWX will be made available to all customers (thus making adding an entitlement for CDPCP-2986 pointless).
- We discovered performance issue introduced by CDPCP-2986.
- CDPCP-3121 addresses the performance issue introduced by CDPCP-2986. Adding an entitlement for CDPCP-3121 is also not a meaningful option due to the comment above around DWX R10.

In other words this change (CDPCP-3121) is the fix to prevent the performance issue **already present** in CB2.31 (CDPCP-2986) we have identified from showing up in prod.

Given that we are late in the release cycle, the dependencies above and the seriousness of the issues involved. Our options are (as @giladwolff suggested):
1. Include CDPCP-3121 to CB2.31.
2. Back out or disable CDPCP-2986 which will delay DWX R10 until we can put back both CDPCP-2986 & CDPCP-3121 (either via hotfix or with CB2.32). Like CB2.31 I suspect there are a bunch of changes riding in DWX R10 as well. @subratanandi can comment on the impact of delay to DWX R10.

I think both options are not ideal but the less impactful and overall less risky one is option 1. @sodre90 @lnardai let us know what you think.

lnardai:
@mdzz Thank you for the detailed information. This is helpful, now we better understand the situation. The current version of CB should not go out since there is in fact a possible regression introduced somewhere between 2.29.0 and 2.31.0.

We had a long discussion with the team.
Our goal is to ship the actual version of CB more frequently. At the moment we are not doing a great job, we had to scrub 2 releases in the last 2 months.
If we keep this goal ahead of us, option2 would be the most in-line with that. We should not release new features until we are sure it's working right.

Step1: Revert the original inclusion of machine users in the user sync process  (or if possible turning it off)
Step2: Recreate the PR with this fix included and ship it as a HOTFIX or next version of CB-2.32.0.

> Would we be able to get this change in through the mentioned hotfix process? How long does that process take?
If QE is not overloaded it should take 3-5 days.

mdzz:
Updating the PR here with the the latest decision/plan. We will proceed with CB2.31 as is (with CDPCP-2986 included), we will disable CDPCP-2986 from UMS (see PR here: https://github.infra.cloudera.com/thunderhead/thunderhead/pull/2287). Next we will include CDPCP-3121 as a HOTFIX on top of CB 2.31. HOTFIX JIRA is [HOTFIX-3527](https://jira.cloudera.com/browse/HOTFIX-3527).

Given this plan we should merge this change to master first.

akanto:
Can we commit this now?

akanto:
As we agreed separately this can be merged, and when we need to fix something in CB 2.31 then we can decide how to proceed.

akanto (approved):

horadla23 (approved):

PR #9169, opened 2020-10-08
CB-9209

sodre90:
FMS syncer was updating the status of Freeipa instance while state was not yet stable or flow could start during syncronization.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
@@ -15,6 +15,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.cloud.model.CloudVmInstanceStatus;
+import com.sequenceiq.flow.core.FlowLogService;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
 import com.sequenceiq.freeipa.entity.InstanceMetaData;
 import com.sequenceiq.freeipa.entity.Stack;
@@ -31,6 +32,9 @@ public class ProviderChecker {
     @Inject
     private StackInstanceProviderChecker stackInstanceProviderChecker;
 
+    @Inject
+    private FlowLogService flowLogService;
+
     @Value("${freeipa.autosync.update.status:true}")
     private boolean updateStatus;
 
@@ -38,28 +42,35 @@ public class ProviderChecker {
         return checkedMeasure(() -> {
             List<ProviderSyncResult> results = new ArrayList<>();
             List<CloudVmInstanceStatus> statuses = stackInstanceProviderChecker.checkStatus(stack, checkableInstances);
-            statuses.forEach(s -> {
-                Optional<InstanceMetaData> instanceMetaData = checkableInstances.stream()
-                        .filter(i -> s.getCloudInstance().getInstanceId().equals(i.getInstanceId()))
-                        .findFirst();
-                if (instanceMetaData.isPresent()) {
-                    InstanceStatus instanceStatus = updateStatuses(s, instanceMetaData.get());
-                    if (instanceStatus != null) {
-                        results.add(new ProviderSyncResult("", instanceStatus, false, s.getCloudInstance().getInstanceId()));
+            if (flowLogService.isOtherFlowRunning(stack.getId())) {
+                throw new InterruptSyncingException(":::Auto sync::: interrupt syncing in updateAndGetStatuses, flow is running on freeipa stack " +
+                        stack.getName());
+            } else {
+                statuses.forEach(s -> {
+                    Optional<InstanceMetaData> instanceMetaData = checkableInstances.stream()
+                            .filter(i -> s.getCloudInstance().getInstanceId().equals(i.getInstanceId()))
+                            .findFirst();
+                    if (instanceMetaData.isPresent()) {
+                        InstanceStatus instanceStatus = updateStatuses(s, instanceMetaData.get());
+                        if (instanceStatus != null) {
+                            results.add(new ProviderSyncResult("", instanceStatus, false, s.getCloudInstance().getInstanceId()));
+                        }
+                    } else {
+                        LOGGER.info(":::Auto sync::: Cannot find instanceMetaData");
                     }
-                } else {
-                    LOGGER.info(":::Auto sync::: Cannot find instanceMetaData");
-                }
-            });
-            checkableInstances.forEach(instanceMetaData -> {
-                if (statuses.stream().noneMatch(s -> s.getCloudInstance().getInstanceId().equals(instanceMetaData.getInstanceId()))) {
-                    if (updateStatus) {
-                        setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_ON_PROVIDER_SIDE);
-                        instanceMetaDataService.save(instanceMetaData);
+                });
+                checkableInstances.forEach(instanceMetaData -> {
+                    if (statuses.stream().noneMatch(s -> s.getCloudInstance().getInstanceId().equals(instanceMetaData.getInstanceId()))) {
+                        if (updateStatus) {
+                            setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_ON_PROVIDER_SIDE);
+                            instanceMetaDataService.save(instanceMetaData);
+                        } else {
+                            LOGGER.debug("updateStatus flag is false, don't update status");
+                        }
                     }
-                }
-            });
-            return results;
+                });
+                return results;
+            }
         }, LOGGER, ":::Auto sync::: provider is checked in {}ms");
     }
 
@@ -93,7 +104,10 @@ public class ProviderChecker {
         }
         if (updateStatus) {
             instanceMetaDataService.save(instanceMetaData);
+        } else {
+            LOGGER.debug("updateStatus flag is false, don't update status");
         }
+
         return status;
     }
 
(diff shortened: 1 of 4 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -99,7 +99,9 @@ private void syncAStack(Stack stack) {
                     if (!checkableInstances.isEmpty()) {
                         SyncResult syncResult = freeipaChecker.getStatus(stack, checkableInstances);
                         List<ProviderSyncResult> results = providerChecker.updateAndGetStatuses(stack, checkableInstances);
-                        updateStackStatus(stack, syncResult, results, alreadyDeletedCount);
+                        if (results.size() > 0) {
```
lacikaaa, line 102: please use `isEmpty()`

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -53,17 +57,23 @@
             });
             checkableInstances.forEach(instanceMetaData -> {
                 if (statuses.stream().noneMatch(s -> s.getCloudInstance().getInstanceId().equals(instanceMetaData.getInstanceId()))) {
-                    if (updateStatus) {
-                        setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_ON_PROVIDER_SIDE);
-                        instanceMetaDataService.save(instanceMetaData);
+                    if (!flowLogService.isOtherFlowRunning(stack.getId())) {
```
lacikaaa, line 60: it's strange we need this. I think `StackStatusCheckerJob` already checks

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -113,7 +115,11 @@ private void updateStackStatus(Stack stack, SyncResult result, List<ProviderSync
         DetailedStackStatus status = getStackStatus(providerSyncResults, result, alreadyDeletedCount);
         if (status != stack.getStackStatus().getDetailedStackStatus()) {
             if (autoSyncConfig.isUpdateStatus()) {
-                stackUpdater.updateStackStatus(stack, status, result.getMessage());
+                if (!flowLogService.isOtherFlowRunning(stack.getId())) {
```
lacikaaa, line 118: same here, code shouldn't reach this part if flow is running

lacikaaa (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
sodre90: It can take long seconds to fetch provider side information. Meanwhile, flow can start in the background.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
sodre90: same answer as previous

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
lacikaaa: in that case it would make sense to throw an exception or similar to stop the whole sync. right now we will continue the sync just log an error.

lacikaaa: also I don't know if it's a good idea to have this check in a loop. freeipa won't have many instances but I would move these checks out from loops to ease the load on the database

sodre90: you are right:
- I created a new exception (InterruptSyncingException)
- check is moved before loops but after provider query operation

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -31,39 +32,48 @@
...
             List<CloudVmInstanceStatus> statuses = stackInstanceProviderChecker.checkStatus(stack, checkableInstances);
-            statuses.forEach(s -> {
-                Optional<InstanceMetaData> instanceMetaData = checkableInstances.stream()
-                        .filter(i -> s.getCloudInstance().getInstanceId().equals(i.getInstanceId()))
-                        .findFirst();
-                if (instanceMetaData.isPresent()) {
-                    InstanceStatus instanceStatus = updateStatuses(s, instanceMetaData.get());
-                    if (instanceStatus != null) {
-                        results.add(new ProviderSyncResult("", instanceStatus, false, s.getCloudInstance().getInstanceId()));
+            if (flowLogService.isOtherFlowRunning(stack.getId())) {
```
topolyai5, line 45: Don't need this here, we check this at the start

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -91,21 +101,25 @@ private InstanceStatus updateStatuses(CloudVmInstanceStatus vmInstanceStatus, In
...
     private void setStatusIfNotTheSame(InstanceMetaData instanceMetaData, InstanceStatus newStatus) {
-        if (instanceMetaData.getInstanceStatus() != newStatus) {
-            if (updateStatus) {
-                instanceMetaData.setInstanceStatus(newStatus);
-                LOGGER.info(":::Auto sync::: The instance status updated from {} to {}", instanceMetaData.getInstanceStatus(), newStatus);
-            } else {
-                LOGGER.info(":::Auto sync::: The instance status would be had to update from {} to {}",
-                        instanceMetaData.getInstanceStatus(), newStatus);
-            }
+        if (updateStatus) {
```
lacikaaa, line 117: this check has been removed: `instanceMetaData.getInstanceStatus() != newStatus`
is it on purpose? I mean we will log here even no change happens in this case

sodre90: thanks! it was a mistake :)

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -113,7 +115,11 @@ private void updateStackStatus(Stack stack, SyncResult result, List<ProviderSync
         DetailedStackStatus status = getStackStatus(providerSyncResults, result, alreadyDeletedCount);
         if (status != stack.getStackStatus().getDetailedStackStatus()) {
             if (autoSyncConfig.isUpdateStatus()) {
-                stackUpdater.updateStackStatus(stack, status, result.getMessage());
+                if (flowLogService.isOtherFlowRunning(stack.getId())) {
+                    throw new InterruptSyncingException(":::Auto sync::: interrupt syncing, flow is running on freeipa stack " + stack.getName());
```
topolyai5, line 119: could you please add a more accurate message? when did happened or stuff like this

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
lacikaaa: thanks for the changes!
please consider catching `InterruptSyncingException` in `executeInternal` method instead of letting it propagate to quartz. There is a try-finally block already

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -99,7 +99,9 @@ private void syncAStack(Stack stack) {
                     if (!checkableInstances.isEmpty()) {
                         SyncResult syncResult = freeipaChecker.getStatus(stack, checkableInstances);
                         List<ProviderSyncResult> results = providerChecker.updateAndGetStatuses(stack, checkableInstances);
-                        updateStackStatus(stack, syncResult, results, alreadyDeletedCount);
+                        if (!results.isEmpty()) {
+                            updateStackStatus(stack, syncResult, results, alreadyDeletedCount);
+                        }
```
topolyai5, line 104: please add a log message, why you skip the update

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -91,8 +101,14 @@ private InstanceStatus updateStatuses(CloudVmInstanceStatus vmInstanceStatus, In
             default:
                 LOGGER.info(":::Auto sync::: the '{}' status is not converted", vmInstanceStatus.getStatus());
         }
-        if (updateStatus) {
-            instanceMetaDataService.save(instanceMetaData);
+        if (!flowLogService.isOtherFlowRunning(stack.getId())) {
```
topolyai5, line 104: if the flow is finished while you iterate over the list, what will happen?

topolyai5 (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
sodre90: please see the above comments

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
sodre90: you will see it from stack trace ;) but I will add some custom thing here

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
sodre90: true!

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
sodre90: good idea!

topolyai5 (approved):

lacikaaa (approved):

lnardai (approved):

PR #9172, opened 2020-10-08
CB-8636 fix for legacy mock service generated crns which causing diff…

doktoric:
…erent creator crn then what we have in the CB user table

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/spi/CredentialToExtendedCloudCredentialConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/spi/CredentialToExtendedCloudCredentialConverter.java
@@ -1,15 +1,18 @@
 package com.sequenceiq.cloudbreak.converter.spi;
 
+import java.util.Optional;
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/converter/spi/CredentialToExtendedCloudCredentialConverter.java
```diff
@@ -31,7 +32,12 @@ public ExtendedCloudCredential convert(Credential credential) {
         if (cloudbreakUser != null) {
             user = userService.getOrCreate(cloudbreakUser);
         } else {
-            user = userService.findByUserCrn(credential.getCreator()).orElse(null);
+            Crn crn = Crn.fromString(credential.getCreator());
+            // Legacy mock generated crn:altus and not crn:cdp this causing different user crn on our side
```
keyki, line 36: It's the other way around, legacy is the crn:cdp and the new is crn:altus

keyki (approved):

PR #9181, opened 2020-10-08
CB-9103 send renewCertificate as internal actor to cloudbreak from da…

sodre90:
…talake

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
@@ -727,6 +727,8 @@ public class SdxService implements ResourceIdProvider, ResourceBasedCrnProvider 
     }
 
     public void renewCertificate(SdxCluster sdxCluster, String userCrn) {
-        stackV4Endpoint.renewCertificate(0L, sdxCluster.getClusterName(), userCrn);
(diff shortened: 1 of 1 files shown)


horadla23 (approved):

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -727,6 +727,8 @@ public SdxCluster save(SdxCluster sdxCluster) {
     }
 
     public void renewCertificate(SdxCluster sdxCluster, String userCrn) {
-        stackV4Endpoint.renewCertificate(0L, sdxCluster.getClusterName(), userCrn);
+        ThreadBasedUserCrnProvider.doAsInternalActor(() -> {
```
doktoric, line 730: should we cover this with a unit test ? I guess this is a bugfix so would be great to see some unit change for this.

sodre90: unfortunately, it seems to be hard to test this. The doAsInternalActor modifies a thread-local variable and later the ApiClientRequestFilter reads this in filterchain, so in unit test there is no chance to test this, maybe with some spring integration test.

lnardai (approved):

PR #9183, opened 2020-10-09
CB-8209: Adding the Cloudbreak version to the environment API.

sxxgrc:
The main reason for this change is to be able to retrieve the exact Cloudbreak version used during the creation of the environment, as this will help with QE purposes.

Let me know if I did all of this right, but essentially I tried to:

- Add a new column to the environment database table for the cloudbreak version
- Update the existing table to include that column
- Set the cloudbreak version when the environment is first created using the info.app.version value
- Have the cloudbreak version present in the DetailedEnvironmentResponse and the EnvironmentDto classes so that it is obtained with any describe environment calls

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
@@ -108,6 +108,8 @@ public class EnvironmentModelDescription {
     public static final String PROXYCONFIG_NAME = "Name of the proxyconfig of the environment.";
     public static final String PROXYCONFIG_RESPONSE = "ProxyConfig attached to the environment.";
 
+    public static final String ENVIRONMENT_SERVICE_VERSION = "The version of the Cloudbreak build used to create the environment.";
+
     private EnvironmentModelDescription() {
     }
 }
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentRequest.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentRequest.java
@@ -113,6 +113,9 @@ public class EnvironmentRequest extends EnvironmentBaseRequest implements Creden
     @ApiModelProperty(EnvironmentModelDescription.GCP_PARAMETERS)
     private GcpEnvironmentParameters gcp;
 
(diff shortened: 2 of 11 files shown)


attilapalfi92 (dismissed):
Needs to be tested first locally

sxxgrc:
This has now been tested locally as I verified that migrating the database up and down with this change works successfully, and used breakpoints with IntelliJ to ensure that the value was correctly being populated in the DB during creation of an environment, and then retrieved from the DB when getting information about the environment.

attilapalfi92 (approved):

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
```diff
@@ -108,6 +108,8 @@
     public static final String PROXYCONFIG_NAME = "Name of the proxyconfig of the environment.";
     public static final String PROXYCONFIG_RESPONSE = "ProxyConfig attached to the environment.";
 
+    public static final String CLOUDBREAK_VERSION = "The version of the Cloudbreak build used to create the environment.";
```
doktoric, line 111: I think we should highlight Environment service here. I know this is overall one repo and same version for each service but we should clearly separate services

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentRequest.java
```diff
@@ -272,6 +275,14 @@ public void setRegions(Set<String> regions) {
         this.regions = regions == null ? new HashSet<>() : regions;
     }
 
+    public String getCloudbreakVersion() {
+        return cloudbreakVersion;
```
doktoric, line 279: environmentServiceVersion would be better I think

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -52,6 +53,9 @@
 
     private final EntitlementService entitlementService;
 
+    @Value("${info.app.version}")
+    private String cloudbreakVersion;
```
doktoric, line 57: EnvironmentServiceVersion

sxxgrc:
I'll make the changes you requested Richard and then merge.

sxxgrc:
Made the desired changes.

doktoric:
@sxxgrc thanks

PR #9190, opened 2020-10-09
CB-8668: FreeIPA status updates should not incorrectly overwrite the …

christmasferret:
…status

Added health check to the end of a successful freeipa repair operation to update the stack status to available. Also change freeipa status from "available" to "repair failed" when a freeipa repair operation fails. Currently FreeIPA status is updated based on a flow completing rather than what the status should be regardless of whether it was in an available statue before. In case of a failed operation, the stack will report a false available status.


christmasferret:
Once repair operation succeed, it shows

Screen Shot 2020-10-09 at 3 03 57 PM

Then within 3 minutes after a health check, it shows
Screen Shot 2020-10-09 at 3 04 18 PM

christmasferret:
If repair operation fails,

Screen Shot 2020-10-09 at 3 04 51 PM

Then after health check,
Screen Shot 2020-10-09 at 3 05 04 PM

christmasferret:
The detailed status change

Screen Shot 2020-10-09 at 9 58 39 AM

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
```diff
@@ -33,10 +41,10 @@
     public static final Collection<Status> AVAILABLE_STATUSES = List.of(AVAILABLE, MAINTENANCE_MODE_ENABLED);
```
handavid, line 41: shouldn't the stack be available if it has the `UPSCALE_COMPLETED`, `DOWNSCALE_COMPLETED`, or `REPAIR_COMPLETED` status? There are places we check whether a stack is available before performing an operation on it.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
```diff
@@ -33,10 +41,10 @@
...
 
     public static final Collection<Status> REMOVABLE_STATUSES = List.of(AVAILABLE, UPDATE_FAILED, CREATE_FAILED, DELETE_FAILED,
-            DELETE_COMPLETED, STOPPED, START_FAILED, STOP_FAILED);
+            DELETE_COMPLETED, STOPPED, START_FAILED, STOP_FAILED, REPAIR_FAILED, UPSCALE_FAILED, DOWNSCALE_FAILED);
```
handavid, line 44: should `UPSCALE_COMPLETED`, `DOWNSCALE_COMPLETED`, and `REPAIR_COMPLETED` also be removable? I'm not sure what `isRemovable` is used for.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
lacikaaa: I would stick with `AVAILABLE` and won't introduce new states representing the same. What's the benefit of this?

christmasferret: Hi @handavid and @lacikaaa , thanks for the review!

The reason we are changing the finishing status of a repair operation is because an operation will always change the status to running/available even operation failed or freeipa is unhealthy. So we will have a period where the available status is falsely reported.

Screen Shot 2020-10-12 at 10 41 31 AM

One example,
Issue a reboot/repair change
Screen Shot 2020-10-12 at 10 50 01 AM

Then instances are terminated so reboot/repair operation failed. But the operation will update status to running.  
Screen Shot 2020-10-12 at 10 49 52 AM

After 3 minutes, the actually health check find out freeipa is not running. 
Screen Shot 2020-10-12 at 10 53 18 AM

With this fix, the false running status will not be reported before a health check.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
christmasferret: If it is removable status, we will delete the freeipa stack from in memory cache. I am reverting this part of code change as we can leave it in cache.

        if (newStatus.isRemovableStatus()) {
            InMemoryStateStore.deleteStack(stack.getId());
        }

lacikaaa: All new `FAILED`, `COMPLETED` statuses should be added to `REMOVABLE_STATUSES` as this indicates these are final states, no flow should be running for these stack and they can be removed from the `InMemoryStateStore`.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
lacikaaa: ok so we would let the statuschecker to determine the status? in case of failed this seems right to me. Although it's a bit strange if we run repair/upscale/downscale and the flow finishes successfully and we can't expect the cluster is in a healthy state. It means the flow exit criteria isn't checking if it's job is done well. This just seems like a workaround to me, instead of fixing the flow (if it needs fixing)
cc @jamisonbennett

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
christmasferret: Thanks for the suggestion! Now I added all the status to REMOVABLE_STATUSES. Since I have removed "repair completed", I use statuschecker to update it to available status.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/action/ChangePrimaryGatewayActions.java
```diff
@@ -150,7 +154,11 @@ protected void doExecute(ChangePrimaryGatewayContext context, StackEvent payload
                 Stack stack = context.getStack();
                 SuccessDetails successDetails = new SuccessDetails(stack.getEnvironmentCrn());
 
-                stackUpdater.updateStackStatus(stack.getId(), getChangePrimaryGatewayCompleteStatus(variables), "Finished changing the primary gateway");
+                if (isFinalChain(variables)) {
```
handavid, line 157: this `if` can be combined with the `if` starting on line 162.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
```diff
@@ -32,11 +37,11 @@
 
     public static final Collection<Status> AVAILABLE_STATUSES = List.of(AVAILABLE, MAINTENANCE_MODE_ENABLED);
 
-    public static final Collection<Status> REMOVABLE_STATUSES = List.of(AVAILABLE, UPDATE_FAILED, CREATE_FAILED, DELETE_FAILED,
+    public static final Collection<Status>  REMOVABLE_STATUSES = List.of(AVAILABLE, UPDATE_FAILED, CREATE_FAILED, DELETE_FAILED,
```
lacikaaa, line 40: still missing the failure states from here

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/action/ChangePrimaryGatewayActions.java
```diff
@@ -150,7 +154,11 @@ protected void doExecute(ChangePrimaryGatewayContext context, StackEvent payload
...
+                    stackStatusCheckerJob.syncAStack(stack);
+                } else {
+                    stackUpdater.updateStackStatus(stack.getId(), getChangePrimaryGatewayCompleteStatus(variables), "Finished changing the primary gateway");
```
lacikaaa, line 160: `getChangePrimaryGatewayCompleteStatus` implementation also checks `isFinalChain(variables)` which doesn't make too much sense after this change. most probably you should replace the method call with `DetailedStackStatus.REPAIR_IN_PROGRESS`

lacikaaa (changes requested):

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
christmasferret: I added the newly added failed states (REPAIR_FAILED, UPSCALE_FAILED, DOWNSCALE_FAILED) to removable states now. Thanks!

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/action/ChangePrimaryGatewayActions.java
christmasferret: Thanks a lot for the suggestion. I've combined the two "if"s and tested the flow again. 

After repair, the status check updates it to available. 
Screen Shot 2020-10-19 at 8 31 31 PM


Screen Shot 2020-10-19 at 8 32 12 PM

christmasferret (commented):
@lacikaaa @handavid 

Thanks for the review!

1) I added the newly added failed states (REPAIR_FAILED, UPSCALE_FAILED, DOWNSCALE_FAILED) to removable states now. 

2) I've combined the two "if"s and tested the flow again.

3) I removed the method getChangePrimaryGatewayCompleteStatus() since it is not necessary and I use DetailedStackStatus.REPAIR_IN_PROGRESS directly.

Screen Shot 2020-10-19 at 8 31 31 PM

Screen Shot 2020-10-19 at 8 32 12 PM

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/Status.java
christmasferret: I added the newly added failed states (REPAIR_FAILED, UPSCALE_FAILED, DOWNSCALE_FAILED) to removable states now. Thanks!

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/action/ChangePrimaryGatewayActions.java
christmasferret: Thanks! I removed the method getChangePrimaryGatewayCompleteStatus() since it is not necessary and I use DetailedStackStatus.REPAIR_IN_PROGRESS directly. I tested the repair flow again and attached evidence below.

lacikaaa:
please rebase your branch to master

lacikaaa (approved):
LGTM
@jamisonbennett could you also check?

jamisonbennett (approved):
Thanks, I think it looks good.

jamisonbennett:
@christmasferret I think you may need to rebase if the swagger changed since your PR was started. I retried the test and it failed again.

christmasferret:
rebased and passed testing.

PR #9193, opened 2020-10-10
CB-9227 - unable to get cluster definition on a stopped environment

gregito:
Removed the condition that – seemingly – has no additional value to the conversion but it prevents one from describing a cluster definition that is connected to a stopped environment.


akanto (changes requested):
Please ensure that we have proper test coverage for this error.

doktoric:
and please add some details to the commit message

doktoric:
please check the integration test job

doktoric (approved):

PR #9197, opened 2020-10-12
CB-8998 increased non java memory to 6gb on smon service based on OPS…

doktoric:
…APS-58273 discussion


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/config/SMONMemoryWorkaroundService.java
```diff
@@ -41,7 +41,7 @@
     @Value("${cb.cm.datahub.normal.smon.firehose.nonJavaMemoryBytes:2}")
     private long datahubNormalFirehoseNonJavaMemoryBytes;
 
-    @Value("${cb.cm.datahub.extensive.smon.firehose.nonJavaMemoryBytes:4}")
+    @Value("${cb.cm.datahub.extensive.smon.firehose.nonJavaMemoryBytes:6}")
```
keyki, line 44: We usually store these in the app yaml so it's worth to move out.

PR #9198, opened 2020-10-12
CB-9237 stack can stuck, somehow status is deleted but terminated timestamp is null

sodre90:


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/TerminationService.java
```diff
@@ -114,12 +114,18 @@ private void deleteOnlyIfForced(Stack stack, boolean force, String terminatedNam
...
-        if (stack.getType().equals(StackType.WORKLOAD)) {
-            grpcUmsClient.notifyResourceDeleted(stack.getResourceCrn(), MDCUtils.getRequestId());
+        try {
+            transactionService.required(() -> {
+                stack.setName(terminatedName);
+                stack.setTerminated(clock.getCurrentTimeMillis());
+                stackService.save(stack);
+                stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.DELETE_COMPLETED, statusReason);
+                if (stack.getType().equals(StackType.WORKLOAD)) {
+                    grpcUmsClient.notifyResourceDeleted(stack.getResourceCrn(), MDCUtils.getRequestId());
```
lacikaaa, line 124: are you sure you want to include an external call in the transaction? I'm against it

sodre90: oh god! :) thanks..

lacikaaa (approved):

biharitomi (approved):

biharitomi (changes requested):
https://github.com/hortonworks/cloudbreak/blob/75341a4ccae88235beb652b4e51ba8cc5b131152/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/terminate/handler/TerminateExternalDatabaseHandler.java#L79
based on the code one more case should be covered what could have caused the stucked stack.
Theoretically the deletion of the external database should not cause the termination of the cluster's stack 😄  Probably a different state was intended to cover those cases

sodre90:
> https://github.com/hortonworks/cloudbreak/blob/75341a4ccae88235beb652b4e51ba8cc5b131152/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/terminate/handler/TerminateExternalDatabaseHandler.java#L79
> 
> 
> based on the code one more case should be covered what could have caused the stucked stack.
> Theoretically the deletion of the external database should not cause the termination of the cluster's stack  Probably a different state was intended to cover those cases

I will create a separate ticket for this.

biharitomi (approved):

PR #9199, opened 2020-10-12
CB-9220. Account id is not passed properly for delete machine user ca…

oleewere:
…ll if internal actor is used

details:
- account id is set from user crn for delete machine user call -> for internal actor, that will be "altus"
- some refactor to use internal call everywhere (in the exact same way)


lnardai (approved):
LGTM!

PR #9201, opened 2020-10-12
CB-9156 refactor permission check logic on request object within auth…

horadla23:
…z framework


horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1423/

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test-legacy-nonpower/15/
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1507/

authorization-common/src/main/java/com/sequenceiq/authorization/service/CompositeRequestPropertyPermissionChecker.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Inject
+    private RequestPropertyPermissionChecker requestPropertyPermissionChecker;
+
+    @Override
+    public <T extends Annotation> void checkPermissions(T rawMethodAnnotation, String userCrn, ProceedingJoinPoint proceedingJoinPoint,
+            MethodSignature methodSignature, long startTime) {
+        // check fields of requestObject
+        CheckPermissionByCompositeRequestProperty methodAnnotation = (CheckPermissionByCompositeRequestProperty) rawMethodAnnotation;
+        Arrays.stream(methodAnnotation.value()).forEach(checkPermissionByRequestField -> requestPropertyPermissionChecker.
```
foldik, line 29: Not a blocker but would be better to make only one request to UMS instead of doing it in a for loop.

foldik (approved):
lgtm

authorization-common/src/main/java/com/sequenceiq/authorization/service/CompositeRequestPropertyPermissionChecker.java
horadla23: to be honest, i would not use `hasRights` once again, it is hard to use and hard to extract result from response, especially if we would use it here after collecting crns for every field of request object, conditional dual check (based on entitlement) makes it even more complex

foldik: Yes I see, the UMS API doesn't give us very usable answer, therefore our logic would become quite complex. In an ideal world for every controller method we could just collect the information what to authorize then send the whole thing to UMS in one request, and based on the annotations we could generate the authorization logic in compile time not using any kind of reflection at all. I know it's a dream so I'm not pushing this further.

PR #9203, opened 2020-10-12
CB-8687 Apply image validation before env creation to fail fast.

pkedvessy:


pkedvessy:
retest this please

Bajzathd (approved):

PR #9204, opened 2020-10-12
DISTX-534 use v1 committer for spark

amitagarwal06:
DISTX-534 As per TSB-437, FileOutputCommitter algorithm v2 may result into
bad output data. Hence it has been advised to go back to v1 version.
This commit makes following config changes in 710, 720, 721, 722, 723
DE, DE HA & Spark3 templates so that v1 algorithm is configured ootb
in DH clusters :

  fs.s3a.committer.name=directory
  mapreduce.fileoutputcommitter.algorithm.version=1
  spark.hadoop.mapreduce.fileoutputcommitter.algorithm.version=1

Testing done :
Have created a DE cluster using a modified 7.2.2 template that includes above mentioned config changes. Cluster creation was successful on AWS using m5d.2xl instance types.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/cdp-data-engineering-710.bp
+++ b/core/src/main/resources/defaults/blueprints/cdp-data-engineering-710.bp
@@ -18,6 +18,12 @@
       {
         "refName": "hdfs",
         "serviceType": "HDFS",
+        "serviceConfigs": [
+          {
+            "name": "core_site_safety_valve",
+            "value": "<property><name>fs.s3a.committer.name</name><value>directory</value></property>"
+          }
+        ],
         "roleConfigGroups": [
           {
             "refName": "hdfs-NAMENODE-BASE",
@@ -77,6 +83,10 @@
           {
             "name": "yarn_log_aggregation_IFile_remote_app_log_dir",
             "value": ""
+          },
+          {
+            "name": "yarn_service_mapred_safety_valve",
+            "value": "<property><name>mapreduce.fileoutputcommitter.algorithm.version</name><value>1</value></property>"
           }
         ],
         "roleConfigGroups": [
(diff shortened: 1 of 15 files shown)


amitagarwal06:
@keyki @lnardai Please merge it in CB-2.31.0 as well as master once its reviewed and all the tests pass.

amitagarwal06:
@steveloughran Pls review and approve it.

horadla23:
@amitagarwal06 is it needed into 2.31? Is there any issue with current solution? If not, please change this PR's target to master, since we are about to release 2.31. Merging changes like this can cause major delay in release process, so if this is required, we can still add this later in 2.31 and release it using hotfix process.

horadla23 (changes requested):
@amitagarwal06 is it needed into 2.31? Is there any issue with current solution? If not, please change this PR's target to master, since we are about to release 2.31. Merging changes like this can cause major delay in release process, so if this is required, we can still add this later in 2.31 and release it using hotfix process.

amitagarwal06:
@horadla23 As mentioned in the PR comment, there is data correctness issue in FileOutputCommitter algorithm version V2 which effects customers using Spark (on previous version of clusters too). Hence its required to configure a prior version V1 of the algorithm. It is very much needed as per TSB (https://jira.cloudera.com/browse/TSB-437), that is why I have been asked to put this in CB-2.31.0 as well as master branches.

cc @shwethags can also add her thoughts.

horadla23:
And do we know about any customer issue caused by using v2?
I'm ok with this if it impacts customers, but without it, it is little bit risky to merge this into 2.31, since it affects all DE templates and this requires retesting on AWS and Azure all DE templates enabled in 2.31. But after proper test it can go into 2.31 and can be released via hotfix process.

shwethags (approved):

shwethags:
I will coordinate with Prasad for testing. Can we get this in CB 2.31 as its for the TSB issue

horadla23 (approved):
we decided with team that this can go in but please test it with all possible runtime version on aws and azure too

PR #9211, opened 2020-10-13
CB-9136. Create scheduled job for machine user cleanup for FreeIPA/DL+DH

oleewere:
details:
We should clean up generated machine users if those were not delete during cluster destroy (no significant increase with those users, but it's useful if we will have some cleanup scripts)
that should be done for 2 components:
- freeipa service
- cloudbreak (core) service (against datalake/datahub)
machine users:
- 1 fluent user for freeipa (if cluster logs collection enabled)
- 1 fluent user for datalake (if cluster logs collection enabled)
- 1 fluent user for datahub (if cluster logs collection or metering enabled) and 1 TP user if WA enabled
(overall freeipa/datalake: max. 1-1, datahub: max 2.)
(all users have specific patterns)
-> Using quartz job scheduler to do the cleanup on every sSunday (0:00) clean users that has no attached clusters + user creation date is older than 3 days


bbihari (approved):
LGTM

PR #9213, opened 2020-10-13
CB-9299: Add FreeIPA HA repair script from runbook to FreeIPA instance

jamisonbennett:
The FreeIPA HA repair runbook has a rather large script that is
manually copied/pasted during a repair. This repair script is being
added to the FreeIPA instances so that it makes the repair runbook
simpler. The repair runbook is for repairing all instances from a
backup.


christmasferret (approved):
LGTM. Keep up the good work! : )

keyki (approved):

lacikaaa:
Although this is merged I have to notes:
- I would use salt to move this script to a more user friendly space and set the appropriate ownship and permissions
- Ideally I would add this to the image, but if we expect this to change frequently, we can have it in the FMS codebase

PR #9217, opened 2020-10-13
CB-9300 Add support for solr collections with HA

orngejaket:
This sets Atlas's Solr configuration to use 2 shards and 2 replications.


abhi2196 (changes requested):
Please update the ranger service solr configs as well for HA.

These are ranger admin role level configs.
`ranger.audit.solr.no.replica: 2`
`ranger.audit.solr.no.shards: 1`
`ranger.audit.solr.max.shards.per.node: 2`

abhi2196 (approved):
LGTM

PR #9219, opened 2020-10-14
CB-9296 Creating a new edit credential object for put requests becaus…

doktoric:
…e in case of audit credential we are using a uuid and we would like to avoid the unnecessary validation on the name object in case of put

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/request/EditCredentialRequest.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/request/EditCredentialRequest.java
@@ -1,12 +1,13 @@
 package com.sequenceiq.environment.api.v1.credential.model.request;
 
 import javax.validation.Valid;
-import javax.validation.constraints.Pattern;
-import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.sequenceiq.authorization.annotation.ResourceObjectField;
+import com.sequenceiq.authorization.resource.AuthorizationResourceAction;
+import com.sequenceiq.authorization.resource.AuthorizationVariableType;
 import com.sequenceiq.environment.api.doc.ModelDescriptions;
 import com.sequenceiq.environment.api.doc.credential.CredentialDescriptor;
 import com.sequenceiq.environment.api.doc.credential.CredentialModelDescription;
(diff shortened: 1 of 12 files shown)


environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/request/CredentialRequest.java
```diff
@@ -18,6 +24,13 @@
 @JsonInclude(Include.NON_NULL)
 public class CredentialRequest extends CredentialBase {
 
+    @Size(max = 100, min = 5, message = "The length of the credential's name has to be in range of 5 to 100")
+    @Pattern(regexp = "(^[a-z][-a-z0-9]*[a-z0-9]$)",
+            message = "The name of the credential can only contain lowercase alphanumeric characters and hyphens and has start with an alphanumeric character")
+    @ApiModelProperty(value = ModelDescriptions.NAME, required = true, allowableValues = "length range[5, 100]")
+    @ResourceObjectField(action = AuthorizationResourceAction.EDIT_CREDENTIAL, variableType = AuthorizationVariableType.NAME)
```
horadla23, line 31: if we do not use this request for editing credential, only for creation, then ResourceObjectField annotation is not needed, since for creation we are using other permission to check

horadla23 (approved):

horadla23 (changes requested):
I accidentally clicked approve, but I had one minor suggestion

environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/request/CredentialRequest.java
doktoric: thanks

horadla23 (approved):

environment-api/src/main/java/com/sequenceiq/environment/api/v1/credential/model/request/EditCredentialRequest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(description = CredentialDescriptor.CREDENTIAL_NOTES, parent = CredentialBase.class, value = "CredentialV1Request")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
+public class EditCredentialRequest extends CredentialBase {
+
+    @ApiModelProperty(value = ModelDescriptions.NAME, required = true)
```
biharitomi, line 25: would we like to allow update the name of a credential?
I think we should add constraint at least for the size as it is VARCHAR(255) in the database and creation requests allow only 100 chars as maximum length

biharitomi (approved):

biharitomi (changes requested):
would we like to allow update the name of a credential?
I think we should add constraint at least for the size as it is VARCHAR(255) in the database and creation requests allow only 100 chars as maximum length

doktoric:
we only use the name of edit request to query the credential

doktoric:
retest this please

biharitomi:
> retest this please



> we only use the name of edit request to query the credential

so the name could not be overridden, right?

doktoric:
> > we only use the name of edit request to query the credential
> 
> so the name could not be overridden, right?

exactly

biharitomi (approved):

biharitomi:
> > > we only use the name of edit request to query the credential
> > 
> > 
> > so the name could not be overridden, right?
> 
> exactly

I approved, but it would be more straightforward with validation/limitation on the length of the name in case of edit requests.

PR #9220, opened 2020-10-14
CB-8639 Stop RDS when stopping DistroX cluster

bergerdenes:
Adding new flows into the the stop/start flow chains, new
states for stack.
Challenge was to find the extra condition that checks the
proper status for stopping.
NOTE: the cert pre-validation flag is now set to Ignored in
the 'core' module!


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/start/handler/StartExternalDatabaseHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            if (StackType.WORKLOAD != stack.getType()) {
+                LOGGER.debug("External database start in Cloudbreak service is required for WORKLOAD stacks only.");
+                result = new StartExternalDatabaseResult(stack.getId(), EXTERNAL_DATABASE_STARTED_EVENT.event(),
+                        stack.getName(), null);
+            } else if (externalDatabase == DatabaseAvailabilityType.NONE) {
+                LOGGER.info("External database for stack {} is not requested. Start is not possible.", stack.getName());
+                result = new StartExternalDatabaseResult(stack.getId(), EXTERNAL_DATABASE_STARTED_EVENT.event(),
+                        stack.getName(), null);
+            } else if (!externalDatabaseConfig.isExternalDatabasePauseSupportedFor(CloudPlatform.valueOf(environment.getCloudPlatform()))) {
```
doktoric, line 83: should we just redbeams and if redbeams say it is not supported than we just ignore the start stop on the DB ?

PR #9224, opened 2020-10-14
CB-9051 [AuthZ] Improve error pop up message for non entitled users

bbihari:
The error message for authz was confusing for users.
This pull request introduce a new format for this error message and unify the messages at different locations.

New format of the message:

You have insufficient rights to perform the following action(s): "environments/describeCredential" on a(n) 'credential' type resource with resource identifier: '<crn of resource>'


horadla23 (changes requested):
Please add the description of the PR to the commit message. Also it seems you need to rebase this branch (swagger check failed because of that possibly).

horadla23:
This solution can extract resource type from CRN but still not showing the name of the resource in the message, as I see this is part of the ticket. Is this intentional?

bbihari:
I think it's too expensive

horadla23 (approved):

PR #9215, opened 2020-10-13
CB-9289 Redbeams ssl integration for supporting cloud provider relate…

doktoric:
…d certs. This PR also helps to identify that customer would like to use the offical provider cert or they preffer BYO scenario

See detailed description in the commit message.

PR #9225, opened 2020-10-14
CB-8114 Gather root certs for Redbeams database instances and save it to  a file on every node of the clusters

biharitomi:


lajosrodek:
Merged into #9215.

PR #9228, opened 2020-10-15
CB-9328 Dual right check should be performed only if authz entitlement disabled

horadla23:


horadla23:
unit test still needed

horadla23:
I have added a small smoke test to check legacy authz for user without poweruser role.

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test-legacy-nonpower/13

lnardai (approved):

sodre90 (approved):

PR #9232, opened 2020-10-17
CB-8147. Make node specific metadata (hostGroup, YARN NM attributes)

sidseth:
available on cluster nodes.

- New metadata file created under /opt/metadata/node.json on cluster
nodes, which contains the hostGroup name, and service specific
attributes.
- The service specific attributes introduced in this PR are for YARN,
and specify whether a node is a compute or worker based on the services
running on the node (instead of relying on the hostGroup name).

Tested by launching a custom DE template with hostGroupNames changed with 2.31-b111 as the base. New files created etc posted on the jira.


sidseth:
@keyki

orchestrator-salt/src/main/resources/salt/salt/metadata/init.sls
```diff
@@ -9,4 +9,11 @@ cluster_metadata:
     - name: /opt/metadata/cluster.json
     - dataset:
         {{ metadata }}
+    - formatter: json
+
+node_attributes:
```
keyki, line 14: What happens if the `hostattrs` are empty? As part of an upgrade we upgrade the salt states and pillar as well, but they won't have any attributes.

sidseth: https://jira.cloudera.com/browse/CB-8147?focusedCommentId=4095583&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4095583

has the output in case there are no attributes defined. This isn't specifically for upgrades, but I'd expect an upgraded CB with this change to exercise the same code path and write out empty attributes.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
...
+        Map<String, Set<ServiceComponent>> hgToNonGwServiceComponentsWithYarnNMs = new HashMap<>();
+        for (Entry<String, Set<ServiceComponent>> entry : hgToNonGwServiceComponents.entrySet()) {
+            for (ServiceComponent sc : entry.getValue()) {
+                if (sc.getService().equalsIgnoreCase(YarnRoles.YARN) && sc.getComponent()
+                        .equalsIgnoreCase(YarnRoles.NODEMANAGER)) {
+                    hgToNonGwServiceComponentsWithYarnNMs.put(entry.getKey(), entry.getValue());
+                    break;
+                }
+            }
+        }
```
lacikaaa, line 329: it would be nicer with stream:
```
Map<String, Set<ServiceComponent>> hgToNonGwServiceComponentsWithYarnNMs = hgToNonGwServiceComponents.entrySet().stream()
                .filter(this::isYarnNodemanager)
                .collect(toMap(Entry::getKey, Entry::getValue));

private boolean isYarnNodemanager(Entry<String, Set<ServiceComponent>> entry) {
        return entry.getValue().stream().anyMatch(sc -> YarnRoles.YARN.equalsIgnoreCase(sc.getService())
                && YarnRoles.NODEMANAGER.equalsIgnoreCase(sc.getComponent()));
    }
```

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
         return cmTemplate.getCdhVersion();
     }
 
+    @Override
+    public Map<String, Map<String, ServiceAttributes>> getHostGroupBasedServiceAttributes() {
+        // The only attributes required at the moment relate to the YARN NodeManager.
+        // This functionality can be moved to an interface / separated if additional components
+        // need this functionality.
```
lacikaaa, line 318: doesn't seem relevant, could you drop these?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
...
+            }
+        }
+        // We have the hostGroups without GATEWAY components with YARN NodeManagers
+        Map<String, Set<String>> componentsByHostGroup = hgToNonGwServiceComponentsWithYarnNMs.entrySet()
+                .stream().collect(
+                        toMap(
+                                Entry::getKey,
+                                e -> e.getValue().stream()
+                                .map(ServiceComponent::getComponent)
+                                .collect(Collectors.toUnmodifiableSet())
```
lacikaaa, line 337: this part should be moved out and named in a separate method

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
...
+        for (Entry<String, Set<ServiceComponent>> entry : hgToNonGwServiceComponents.entrySet()) {
+            for (ServiceComponent sc : entry.getValue()) {
+                if (sc.getService().equalsIgnoreCase(YarnRoles.YARN) && sc.getComponent()
+                        .equalsIgnoreCase(YarnRoles.NODEMANAGER)) {
+                    hgToNonGwServiceComponentsWithYarnNMs.put(entry.getKey(), entry.getValue());
+                    break;
+                }
+            }
+        }
+        // We have the hostGroups without GATEWAY components with YARN NodeManagers
```
lacikaaa, line 330: instead of comment could you use variable and methods name reflecting the same?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
...
+        Map<String, Set<String>> componentsByHostGroup = hgToNonGwServiceComponentsWithYarnNMs.entrySet()
+                .stream().collect(
+                        toMap(
+                                Entry::getKey,
+                                e -> e.getValue().stream()
+                                .map(ServiceComponent::getComponent)
+                                .collect(Collectors.toUnmodifiableSet())
+                ));
+        // Re-using the current LoadBasedAutoScaling recommendation along with the above YARN filter.
```
lacikaaa, line 339: this comment also seems unnecessary or I'm missing something

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
...
+            if (computeHostGroups.contains(hg)) {
+                m = Collections.singletonMap(YarnRoles.YARN, new ServiceAttributes(ServiceComponent.of(YarnRoles.YARN, YarnRoles.NODEMANAGER),
+                        Collections.singletonMap(YarnConstants.ATTRIBUTE_NAME_NODE_INSTANCE_TYPE, YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_COMPUTE)));
+            } else {
+                m = Collections.singletonMap(YarnRoles.YARN, new ServiceAttributes(ServiceComponent.of(YarnRoles.YARN, YarnRoles.NODEMANAGER),
+                        Collections.singletonMap(YarnConstants.ATTRIBUTE_NAME_NODE_INSTANCE_TYPE, YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_WORKER)));
+            }
+            result.put(hg, m);
+        }
+        LOGGER.debug("ServiceAttributes: {}", result);
```
lacikaaa, line 356:
```        Map<String, Map<String, ServiceAttributes>> result = new HashMap<>();
        for (String hg : componentsByHostGroup.keySet()) {
            String instanceType = computeHostGroups.contains(hg) ? YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_COMPUTE
                    : YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_WORKER;
            Map<String, ServiceAttributes> m = Collections.singletonMap(YarnRoles.YARN, new ServiceAttributes(ServiceComponent.of(YarnRoles.YARN, YarnRoles.NODEMANAGER),
                    Collections.singletonMap(YarnConstants.ATTRIBUTE_NAME_NODE_INSTANCE_TYPE, instanceType)));
            result.put(hg, m);
        }
```
seems a bit readable this way and could be moved to a separate method also

lacikaaa: Also `m` is not a good variable name, please rename it to something more helpful

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -291,6 +311,52 @@ public String getStackVersion() {
         return cmTemplate.getCdhVersion();
     }
 
+    @Override
+    public Map<String, Map<String, ServiceAttributes>> getHostGroupBasedServiceAttributes() {
```
lacikaaa, line 315: I ended up with this, but the method and variable names should be changed with more meaningful one I think. This is just an example:
```    
	@Override
    public Map<String, Map<String, ServiceAttributes>> getHostGroupBasedServiceAttributes() {
        Map<String, Set<String>> componentsByHostGroup = collectComponentsByHostrgoup();
        Set<String> computeHostGroups = getRecommendationByBlacklist(BlackListedLoadBasedAutoscaleRole.class,
                true, componentsByHostGroup);
        Map<String, Map<String, ServiceAttributes>> result = createServiceAttributeMap(componentsByHostGroup, computeHostGroups);
        LOGGER.debug("ServiceAttributes: {}", result);
        return result;
    }

    private Map<String, Map<String, ServiceAttributes>> createServiceAttributeMap(Map<String, Set<String>> componentsByHostGroup, Set<String> computeHostGroups) {
        Map<String, Map<String, ServiceAttributes>> result = new HashMap<>();
        for (String hg : componentsByHostGroup.keySet()) {
            String instanceType = computeHostGroups.contains(hg) ? YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_COMPUTE
                    : YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_WORKER;
            Map<String, ServiceAttributes> m = Collections.singletonMap(YarnRoles.YARN, new ServiceAttributes(ServiceComponent.of(YarnRoles.YARN, YarnRoles.NODEMANAGER),
                    Collections.singletonMap(YarnConstants.ATTRIBUTE_NAME_NODE_INSTANCE_TYPE, instanceType)));
            result.put(hg, m);
        }
        return result;
    }

    private Map<String, Set<String>> collectComponentsByHostrgoup() {
        Map<String, Set<ServiceComponent>> hgToNonGwServiceComponents = getNonGatewayServicesByHostGroup();
        Map<String, Set<ServiceComponent>> hgToNonGwServiceComponentsWithYarnNMs = hgToNonGwServiceComponents.entrySet().stream()
                .filter(this::isYarnNodemanager)
                .collect(toMap(Entry::getKey, Entry::getValue));
        Map<String, Set<String>> componentsByHostGroup = hgToNonGwServiceComponentsWithYarnNMs.entrySet()
                .stream().collect(toMap(Entry::getKey, this::collectComponents));
        return componentsByHostGroup;
    }

    private Set<String> collectComponents(Entry<String, Set<ServiceComponent>> e) {
        return e.getValue().stream()
                .map(ServiceComponent::getComponent)
                .collect(Collectors.toUnmodifiableSet());
    }

    private boolean isYarnNodemanager(Entry<String, Set<ServiceComponent>> entry) {
        return entry.getValue().stream().anyMatch(sc -> YarnRoles.YARN.equalsIgnoreCase(sc.getService())
                && YarnRoles.NODEMANAGER.equalsIgnoreCase(sc.getComponent()));
    }

**Review by lacikaaa — CHANGES_REQUESTED** [review]

File: orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java [hunk 9]
```diff
@@ -23,19 +25,33 @@
 
     private String uuids;
 
+    // Used for generic attributes associated with the node. e.g. YARN attributes when running NMs, Spot vs non-spot, etc
+    private Map<String, Map<String, String>> attributes;
```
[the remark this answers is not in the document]
sidseth: Will look into this as an option.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
sidseth: I don't like using Java streams (not as easy to read). If that's the pattern that is generally used, I can switch it over. (Applies to other comments related to streams)

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
sidseth: Sure. This was a comment to indicate potential future changes.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
sidseth: Think this is very specific to CB code style? I think comments in the code can be quite helpful.

sidseth:
Thanks for the review @lacikaaa . Will update the PR once I'm back after a short break.
One of the concerns I had is whether the patch applies to various scenarios like new nodes, scale-up, new cluster etc. I think it covers the various cases - but something I was hoping to get review feedback on.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
lacikaaa: comments tends to be outdated, misleading. Usually indicates the code is not "clean", doesn't speak for itself which could be improved by refactoring, like breaking into methods with good names, etc
We generally avoid them as much as possible and leave comments where it's really necessary.

lacikaaa:
If you drop things from the `Node` class I would check that perspective, but we push down pillars when we are scaling, so I think you are covered. I still think the calculation of the pillars should be refactored as I suggested as it would simplify that part of the code and you won't have to add it at so many places where nodes are created.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
sidseth: Goign to modify this comment a bit. It is relevant in the sense that determining a node as 'compute' means that it can be AutoScaled, and so it is important to re-use the autoscale recommendation logic.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
sidseth: Thanks for the suggestions. Have broken up the method a bit, and made some other changes.

sidseth:
@lacikaaa Haven't tested the changes on a cluster yet. Will do that tomorrow. Please do not commit before that. This is ready for review though.

sidseth:
Tested on a cliuster as well. This should be good to go in now.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -323,6 +331,26 @@ private SaltConfig createSaltConfig(Stack stack, Cluster cluster, GatewayConfig
         return new SaltConfig(servicePillar, grainPropertiesService.createGrainProperties(gatewayConfigs, cluster, nodes));
     }
 
+    @VisibleForTesting
+    void addHostAttributes(Stack stack, Map<String, SaltPillarProperties> servicePillar, Set<Node> nodes) {
```
lacikaaa, line 335: I think this should be moved to a separate class together with `getAttributesForHostGroup`.
And this method should return with a map, like:
`Map.of("hostattrs", new SaltPillarProperties("/nodes/hostattrs.sls", singletonMap("hostattrs", attributes)))`
And in line 303 you would need only:
`servicePillar.putAll(createHostAttributes(..));`

I know there is a lot of code here (even written by me) where we add stuff to this map by passing it around but honestly it's not the best solution.

Moving this method out would make `@VisibleForTesting` unnecessary which is already a sign that it should be in a separate class.

orchestrator-salt/src/main/resources/salt/salt/metadata/settings.sls
```diff
@@ -20,3 +22,5 @@
     'clusterName' : clusterName,
     'cluster_in_childenvironment' : cluster_in_childenvironment
 }) %}
+
+{% set hostattrs = salt['pillar.get']('hostattrs:'~grains['fqdn']) %}
```
lacikaaa, line 26: what is this `~` for?

lacikaaa (changes requested):
looks good, just one bit of refactor suggested

orchestrator-salt/src/main/resources/salt/salt/metadata/settings.sls
sidseth: The ~ is apparently required when accessing variables inside of a pillar get. i.e. pillar.get(blah(grains.get))

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
sidseth: Moving this into a HostAttributeDecorator, and having that set the PillarProperties instead of returning a map.
I'm going to ask you to make changes to return a map if you feel strongly about that / other code style changes. Given I don't write a lot of CB patches - was avoiding any kind of re-structuring and following existing patterns as much as possible.

PR #9238, opened 2020-10-19
CB-9227 - unable to get cluster definition on a stopped environment

gregito:
CB-9227 - unable to get cluster definition on a stopped environment, so removed the validation which prevents this conversion to happen in these cases

### DIFF
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
@@ -30,7 +30,6 @@ import com.sequenceiq.distrox.api.v1.distrox.model.DistroXV1Request;
 import com.sequenceiq.distrox.api.v1.distrox.model.network.NetworkV1Request;
 import com.sequenceiq.distrox.api.v1.distrox.model.tags.TagsV1Request;
 import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -192,14 +192,10 @@ private YarnStackV4Parameters getYarnProperties(DistroXV1Request source, Detaile
     public DistroXV1Request convert(StackV4Request source) {
         DistroXV1Request request = new DistroXV1Request();
         request.setName(source.getName());
-        DetailedEnvironmentResponse environment = null;
         if (source.getEnvironmentCrn() != null) {
-            environment = environmentClientService.getByCrn(source.getEnvironmentCrn());
-            if (environment != null && environment.getEnvironmentStatus() != EnvironmentStatus.AVAILABLE) {
-                throw new BadRequestException(String.format("Environment state is %s instead of AVAILABLE", environment.getEnvironmentStatus()));
-            }
+            Optional<DetailedEnvironmentResponse> environment = getEnvironmentName(source.getEnvironmentCrn());
```
doktoric, line 196: if the source has no env crn why we woul like to query it ?

gregito: do we? in line 194 we have a check for the env crn. if it's null then no query happens, otherwise, we set the name based on the result of line 214

doktoric: now I see that part thanks

PR #9239, opened 2020-10-19
CB-8227: Introduce a new entitlement to allow different Data Hub versions than Data Lake

tiborpopovics:


lacikaaa (approved):

core/src/main/java/com/sequenceiq/cloudbreak/controller/StackCreatorService.java
```diff
@@ -178,6 +182,7 @@ public StackV4Response createStack(User user, Workspace workspace, StackV4Reques
                 determineBlueprint(stackRequest, workspace),
                 LOGGER,
                 "Stack request converted to stack took {} ms for stack {}", stackName);
+            validateBlueprintVersion(blueprint, stackRequest);
```
keyki, line 185: I'm not sure about the blueprint version, because QE sometimes use the same blueprint for multiple runtime versions.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private Optional<String> findStackVersion(StackV4Request stackRequest) {
+        return Optional.of(stackRequest)
+                .map(StackV4Request::getCluster)
+                .map(ClusterV4Request::getCm)
+                .map(ClouderaManagerV4Request::getRepository)
+                .map(RepositoryV4Request::getVersion);
+    }
+
+    private void validateStackVersion(String stackBlueprintVersion, String sdxRuntimeVersion) {
```
keyki, line 56: We shouldn't call the method param stackBlueprintVersion

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Inject
+    private EntitlementService entitlementService;
+
+    public void validate(StackV4Request stackRequest) {
+        if (isDifferentDataHubAndDataLakeVersionAllowed()) {
+            LOGGER.debug("The Data Hub version validation has been turned off with entitlement.");
+        } else {
+            LOGGER.debug("Validating Data Hub version.");
+            findStackVersion(stackRequest).ifPresent(stackRuntimeVersion -> {
```
keyki, line 40: I think we might need to check the image as well. In the case of pre-warm image, the request won't contain the repository info, but the image id. In this case, we should check the version on the image if this is available at this point.

keyki (approved):

PR #9241, opened 2020-10-19
CB-9374: Use last FreeIPA cookie when there are multiple

jamisonbennett:
When there are multiple FreeIPA ipa_session cookies, use the last one.
This is the same behavior as FreeIPA does.
See https://github.com/freeipa/freeipa/blob/139d60d74706d2ba9855db5faefee322b23b0ba5/ipapython/cookie.py#L322-L349
for details.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
@@ -33,7 +33,6 @@ import org.apache.http.conn.socket.ConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.cookie.Cookie;
 import org.apache.http.entity.ContentType;
-import org.apache.http.impl.client.BasicCookieStore;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
@@ -54,6 +53,7 @@ import com.sequenceiq.cloudbreak.client.HttpClientConfig;
 import com.sequenceiq.freeipa.client.auth.InvalidPasswordException;
 import com.sequenceiq.freeipa.client.auth.InvalidUserOrRealmException;
 import com.sequenceiq.freeipa.client.auth.PasswordExpiredException;
+import com.sequenceiq.freeipa.util.FreeIpaCookieStore;
 
 public class FreeIpaClientBuilder {
 
@@ -201,7 +201,7 @@ public class FreeIpaClientBuilder {
         stickyHeaders.forEach(post::addHeader);
         post.setEntity(new UrlEncodedFormEntity(List.of(new BasicNameValuePair("user", user), new BasicNameValuePair("password", pass))));
 
-        CookieStore cookieStore = new BasicCookieStore();
+        CookieStore cookieStore = new FreeIpaCookieStore();
         try (CloseableHttpResponse response = execute(post, cookieStore)) {
             stickyId = getStickIdFromHeaders(response);
             if (response.getStatusLine().getStatusCode() != HttpStatus.OK.value()) {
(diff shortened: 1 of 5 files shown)


lacikaaa (commented):
Have you seen this issue? I mean it's strange if we can run into this as every time we build the client we do a login, so I think we should have only one cookie there

jamisonbennett:
@lacikaaa I was debugging some issues with the TGT generator and happened to notice the difference in behavior between the FreeIPA implementation and the cloudbreak implementation. I did not see multiple cookies in the actual payload but rather I was just trying to infer what might be different for an intermittent issue. I figured I would adjust the behavior even though we have not experienced this issue yet that I know of. I suspect that it is rare for there to be multiple cookies like maybe close to ticket renewal.

lacikaaa:
ok, but in this case I think `reduce` is not enough, it should be sorted by expiration date or similar. Or do you think these cookies are returned in the order they have been added?

jamisonbennett:
Here is a deep dive on the cookie order:

We use the `BasicCookieStore` which behaves as follows:
1. Matching cookies that are returned later in the sequence cause remove existing cookies (expiration is not considered yet).
2. If the new cookie is not expired then it is added.
3. Cookies are sorted based on the `CookieIdenityComparator`

The `CookieIdentifyComparator` sorts as follows:
1. Cookie name is the top priority
2. Domain name is the second priority
3. Cookie path is the third priority
4. Expiration is not considered

All of the cookies will have the same name (we are filtering on "ipa_session" cookie name), so the top priority sort will be the same. The domain name will be the same. So this leaves the cookie path. The cookie path should partially. According to [RFC-6265](https://tools.ietf.org/html/rfc6265), section 5.4.2 "Cookies with longer paths are listed before cookies with shorter paths...Not all user agents sort the cookie-list in this order". So when using the `CookieIdenityComparator`, the last cookie should be used.

While it is noted in [RFC-6265](https://tools.ietf.org/html/rfc6265) section 4.2.2, "servers SHOULD NOT rely upon the serialization order". We are dealing with the client-side of the implementation not the server-side.

While it is also noted in [RFC-6265](https://tools.ietf.org/html/rfc6265) section 5.4.2, "Among cookies that have equal-length path fields, cookies with earlier creation-times are listed before cookies with later creation-times." This is not how the java client we are using works. The cloudbreak implemenation will unfortunately rely on the serialization order from the server for expiration but I dont' see this as a problem either since the FreeIPA client that was written by FreeIPA does the same. https://github.com/freeipa/freeipa/blob/139d60d74706d2ba9855db5faefee322b23b0ba5/ipapython/cookie.py#L322-L349

----

I don't really have a strong opinion since I really cant concluded that this is a necessary fix since we don't know for sure that it has happened. If there are strong opinions, I can just close the PR and our client will have different behavior than the FreeIPA client from FreeIPA.

lacikaaa:
I think one thing we certainly should do: if there are multiple `ipa_session` cookie, we should log everything about them, except the sensitive `value` part.
Also we can work ahead and compare/sort these cookies by expiration time and select the first/last (depending on the sort). At least I think if we have multiple cookie with the same name here, it should only differ by the expiration date, and which one has an expiration farther in the future should be the valid one. Or we could stick with your current implementation and at least we will have a log before it so we can check if something goes sideways.
I'm fine with both solution, but a log should be added so we can investigate later.

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -231,7 +231,10 @@ private CookieAndStickyId connect(String user, String pass, String apiAddress, i
                         response.getStatusLine().getStatusCode()));
             }
         }
-        Cookie sessionCookie = cookieStore.getCookies().stream().filter(cookie -> "ipa_session".equalsIgnoreCase(cookie.getName())).findFirst().get();
+        Cookie sessionCookie = cookieStore.getCookies().stream()
```
kmanamcheri, line 234: Lets add this comment block from the canonical freeipa client for reference

> A cookie string may contain multiple cookies, parse the cookie
 string and return the last cookie in the string matching the
 cookie name or None if not found.
 method which iterates over what parse() returns looking for
 the specific cookie.
 When cookie_name appears more than once the last instance is
 returned rather than the first because the ordering sequence
 makes the last instance the current value.

kmanamcheri:
@lacikaaa Reading a little bit more on the commit in the canonical freeipa client (https://github.com/freeipa/freeipa/commit/9269e5d6ddc85716d2b03c7763cf4b8e1ca67cad) it looks like we should do this where we pick the last one because that's what the server seems to assume the client is doing.

I do agree that we need to add logging around when this happens, but even otherwise I think we should fix this, and overall doesn't seem like a risky fix.

jamisonbennett:
In order to log all duplicates with different expirations, I had to create a custom `CookieStore` since the default one did not consider the expiration as part of the key and thus it would overwrite the cookie. This is probably the safest approach but its a little more code.

freeipa/src/main/java/com/sequenceiq/freeipa/util/CookieComparator.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.http.cookie.Cookie;
+import org.apache.http.cookie.CookieIdentityComparator;
+
+// Sort preceidence:
+//   1. Cookie name (backwards - doesn't matter)
+//   2. Domain name (backwards - doesn't matter)
+//   3. Cookie path (more specific path is listed first) per RFC-6265 section 5.4.2
+//   4. Expiration (no expiratation is listed first, latest expiration is listed next)
+public class CookieComparator implements Serializable, Comparator<Cookie> {
```
kmanamcheri, line 15: Don't you have to use it somewhere? Like when you are sorting above, you have to use this comparator, no?

jamisonbennett: The `FreeIpaCookeieStore` is already sorted:
https://github.com/hortonworks/cloudbreak/blob/865c9d05b03a675e0169b5fa3f17efcbd1c5d98d/freeipa/src/main/java/com/sequenceiq/freeipa/util/FreeIpaCookieStore.java#L15
`private final Set<Cookie> cookies = new TreeSet<>(new CookieComparator());`

freeipa/src/main/java/com/sequenceiq/freeipa/util/CookieComparator.java
```diff
@@ -0,0 +1,46 @@
+package com.sequenceiq.freeipa.util;
+
+import java.io.Serializable;
+import java.util.Comparator;
+import java.util.Date;
+
+import org.apache.http.cookie.Cookie;
+import org.apache.http.cookie.CookieIdentityComparator;
+
+// Sort preceidence:
```
kmanamcheri, line 10: nit: typo

freeipa/src/main/java/com/sequenceiq/freeipa/util/CookieComparator.java
kmanamcheri: I see that this is already used in https://github.com/hortonworks/cloudbreak/blob/865c9d05b03a675e0169b5fa3f17efcbd1c5d98d/freeipa/src/main/java/com/sequenceiq/freeipa/util/FreeIpaCookieStore.java#L15

kmanamcheri (approved):
Other than the nit.. lgtm

lacikaaa (approved):

PR #9247, opened 2020-10-20
CB-9382 make sure getResourceCrnByResourceName not returns null

lnardai:
- make sure getResourceCrnByResourceName for BlueprintService is not returning null and framework also responds with 404 if resourceCrn cannot be found

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
@@ -398,7 +398,8 @@ public class BlueprintService extends AbstractWorkspaceAwareResourceService<Blue
 
     @Override
     public String getResourceCrnByResourceName(String resourceName) {
-        return blueprintRepository.findResourceCrnByNameAndAccountId(resourceName, ThreadBasedUserCrnProvider.getAccountId());
+        return blueprintRepository.findResourceCrnByNameAndAccountId(resourceName, ThreadBasedUserCrnProvider.getAccountId())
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
```diff
@@ -409,7 +410,7 @@ public String getResourceCrnByResourceName(String resourceName) {
     @Override
     public List<String> getResourceCrnListByResourceNameList(List<String> resourceNames) {
         return resourceNames.stream()
-                .map(resourceName -> blueprintRepository.findResourceCrnByNameAndAccountId(resourceName, ThreadBasedUserCrnProvider.getAccountId()))
+                .map(resourceName -> blueprintRepository.findResourceCrnByNameAndAccountId(resourceName, ThreadBasedUserCrnProvider.getAccountId()).orElse(null))
```
horadla23, line 413: maybe it is better to throw here a not found also, isn't it?

horadla23 (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/blueprint/BlueprintService.java
lnardai: I was not sure to change the behaviour, since the old way only returned a null, but I guess it's safer to exit with an error.

horadla23: i guess this method is used by authz framework only so i think we can change the behavior, and in long term, we should change this method anyway to use 1 db query

horadla23 (approved):

PR #9252, opened 2020-10-20
CB-9392. Add default annymization rule for CDP hashed password & refactor default rule loading

oleewere:
See detailed description in the commit message

add a default anonymization rule for cdp hashed password

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/configuration/telemetry/AccountTelemetryConfig.java
+++ b/environment/src/main/java/com/sequenceiq/environment/configuration/telemetry/AccountTelemetryConfig.java
@@ -1,0 +1,24 @@
+package com.sequenceiq.environment.configuration.telemetry;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.common.api.telemetry.model.AnonymizationRule;
+
+@Component
+@ConfigurationProperties(prefix = "environment.telemetry")
+public class AccountTelemetryConfig {
+
+    private List<AnonymizationRule> rules = new ArrayList<>();
+
+    public List<AnonymizationRule> getRules() {
+        return rules;
+    }
+
+    public void setRules(List<AnonymizationRule> rules) {
+        this.rules = rules;
+    }
+}
--- a/environment/src/main/java/com/sequenceiq/environment/telemetry/service/AccountTelemetryService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/telemetry/service/AccountTelemetryService.java
@@ -5,6 +5,7 @@ import java.util.Base64;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -16,6 +17,7 @@ import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.exception.BadRequestException;
 import com.sequenceiq.common.api.telemetry.model.AnonymizationRule;
 import com.sequenceiq.common.api.telemetry.model.Features;
+import com.sequenceiq.environment.configuration.telemetry.AccountTelemetryConfig;
 import com.sequenceiq.environment.telemetry.domain.AccountTelemetry;
 import com.sequenceiq.environment.telemetry.repository.AccountTelemetryRepository;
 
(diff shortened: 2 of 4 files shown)


environment/src/main/java/com/sequenceiq/environment/telemetry/service/AccountTelemetryService.java
```diff
@@ -28,28 +38,25 @@
...
+        this.rulesFileLocation = rulesFileLocation;
+    }
+
+    @PostConstruct
+    public void init() {
+        try {
+            defaultRules = loadAnonymizationRules();
+        } catch (IOException e) {
+            LOGGER.error("Error during loading default anonymization rules.", e);
+        }
```
lacikaaa, line 59: I think we won't notice this issue when something is wrong if we let the app start

environment/src/test/resources/default-anonymization-rules.yaml
```diff
@@ -0,0 +1,10 @@
+- pattern: \b([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-\._]*[A-Za-z0-9])@(([A-Za-z0-9]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])\.)+([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\b
+  replacement: "email@redacted.host"
+- pattern: \d{4}[^\w]\d{4}[^\w]\d{4}[^\w]\d{4}
+  replacement: "XXXX-XXXX-XXXX-XXXX"
+- pattern: \d{3}[^\w]\d{2}[^\w]\d{4}
+  replacement: "XXX-XX-XXXX"
+- pattern: FPW\:\s+[\w|\W].*
+  replacement: "FPW: [REDACTED]"
+- pattern: cdpHashedPassword=.*[']
+  replacement: "[CDP PWD ATTRS REDACTED]"
```
lacikaaa, line 10: we should test the same file we use in prod, and not a different one.

environment/src/main/java/com/sequenceiq/environment/telemetry/service/AccountTelemetryService.java
```diff
@@ -190,4 +178,26 @@ private String createCRN(String accountId) {
                 .build()
                 .toString();
     }
+
+    private List<AnonymizationRule> loadAnonymizationRules() throws IOException {
```
lacikaaa, line 182: instead of reading the file manually, wouldn't be it better if we would use spring for this? like in `InstanceMetadataUpdater` we use it for reading the `packages` info.

environment/src/main/resources/telemetry/default-anonymization-rules.yaml
```diff
@@ -0,0 +1,10 @@
+- pattern: \b([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-\._]*[A-Za-z0-9])@(([A-Za-z0-9]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])\.)+([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\b
+  replacement: "email@redacted.host"
+- pattern: \d{4}[^\w]\d{4}[^\w]\d{4}[^\w]\d{4}
+  replacement: "XXXX-XXXX-XXXX-XXXX"
+- pattern: \d{3}[^\w]\d{2}[^\w]\d{4}
+  replacement: "XXX-XX-XXXX"
+- pattern: FPW\:\s+[\w|\W].*
+  replacement: "FPW: [REDACTED]"
+ - pattern: cdpHashedPassword=.*[']
```
lacikaaa, line 9: unnecessary space as first char

environment/src/test/resources/default-anonymization-rules.yaml
oleewere: now the objects are mocked as the tests are not about to parsing the files (as config read from application.yml)

environment/src/main/java/com/sequenceiq/environment/configuration/telemetry/AccountTelemetryConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import com.sequenceiq.common.api.telemetry.model.AnonymizationRule;
+
+@Configuration
```
lacikaaa, line 11: the main purpose of `@Configuration` is to indicate the class declares `Beans`
I think here we should go with `@Component`

lacikaaa (approved):

PR #9254, opened 2020-10-20
CB-9154: Add ssl root cert and ssl mode to backup and restore PGSQL settings.

brycederriso:
Check if a root cert exists on the instance.
* Run backup/restore with ssl-full, using the root cert if it's there.
* Fallback to default behavior if the root cert is not present.

Tested by `rsync`ing salt scripts on to a datalake master node and manually running salt backup and restore commands.
* `rsync` salt scripts
* `rsync` rds root certs to `/hadoopfs/fs1/database-cacerts/certs.pem`
* update pillar values with `ranger_admin_group` and `backup_location`
* run backup and restore:
```sh
salt $(hostname -f) state.apply postgresql.disaster_recovery.backup
salt $(hostname -f) state.apply postgresql.disaster_recovery.restore
```

Tested backup and restore when:
* Root cert present
* Root cert not present

And when `force_ssl` rds option is `0` and `1` in AWS.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -29,6 +29,11 @@ RANGERGROUP="$5"
 LOGFILE=${6:-/var/log/}/dl_postgres_backup.log
 echo "Logs at ${LOGFILE}"
 
+if [[ -f /hadoopfs/fs1/database-cacerts/certs.pem ]]; then
+  export PGSSLROOTCERT=/hadoopfs/fs1/database-cacerts/certs.pem
+  export PGSSLMODE=verify-full
+fi
+
 doLog() {
   set +x
   type_of_msg=$(echo "$@" | cut -d" " -f1)
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -9,15 +9,16 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ $# -ne 5 && $# -ne 6 ]]; then
+if [[ $# -ne 6 && $# -ne 7 ]]; then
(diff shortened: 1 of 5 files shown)


lajosrodek (approved):
LGTM

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -29,6 +29,11 @@ RANGERGROUP="$5"
 LOGFILE=${6:-/var/log/}/dl_postgres_backup.log
 echo "Logs at ${LOGFILE}"
 
+if [[ -f /hadoopfs/fs1/database-cacerts/certs.pem ]]; then
```
doktoric, line 32: how about if we pass the folder /hadoopfs/fs1 down  in a pillar. In that case if somebody change the volume logic then we automatically get the right value.

brycederriso: Where should we set that pillar value? Is it already available in salt through the other changes @lajosrodek has made, or will I need to set it somewhere in the Java side?

lajosrodek: That would certainly need changes in my logic as well once it happens. 😃

@doktoric I agree this would be a better approach, and we shall indeed do it (not only here but also in other related parts). However, I am wondering if it is the right time to fix this now. Could we possibly postpone it to 2.33 and fix all affected things altogether?

brycederriso: I've added a new value under the disaster_recovery pillar.

brycederriso: I'll monitor the build jobs and retry them.

doktoric (approved):

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
lajosrodek: Has been solved in #9563.

PR #9257, opened 2020-10-20
CB-9250 mock idbroker mapping should be only applied if there is file…

doktoric:
…system in the cluster object


biharitomi (approved):
LGTM

PR #9258, opened 2020-10-20
CB-8945 Ability to run tests against MOW Dev

aszegedi:
In order to make sure that we can verify CB deployment was successful we shall execute our E2E test suits against MOW environments as well.

>As a first version I would only include the API E2E tests and use a static ACCESSKEY/SECRETKEY pair. Later on with the `cdpcpaccounthelper` we can even start a Chrome instance that has UI access but that would require more effort.


integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
```diff
@@ -109,14 +110,17 @@ private void validateOwnerTag(TaggedResponse response, String tag, TestContext t
     }
 
     private void validateClouderaCreatorResourceNameTag(TaggedResponse response, String tag, TestContext testContext) {
-        if (response.getTagValue(tag).contains(testContext.getActingUserName())) {
-            Log.log(LOGGER, format(" Default tag: [%s] value is: [%s] contains [%s] acting user name! ", tag, response.getTagValue(tag),
-                    testContext.getActingUserName()));
+        Crn actingUserCrn = testContext.getActingUserCrn();
+        Crn clouderaCreatorResourceName = Crn.fromString(response.getTagValue(tag));
+
+        if (clouderaCreatorResourceName.equals(actingUserCrn)) {
```
lnardai, line 116: I'm not sure about his change. At least the Owner tag should contain the creator's name not the CRN I think.

lnardai (changes requested):

integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
aszegedi: Unfortunately we had to upgrade this comparison, because of:
>The account basis value is a string that will be converted into SFDC account ID format that serves as the primary key for the account. This is distinct from the CDP-internal account ID UUID that shows up in CRNs and many API calls. 
...
> The user basis value works along similar lines but results in the primary key for the user within the account. 

from [CDPCP Account Helper > Creating Test Users](https://github.infra.cloudera.com/thunderhead/thunderhead/tree/master/tools/cdpcp-account-helper#creating-test-users)

For instance a mocked: `"Cloudera-Creator-Resource-Name":"crn:altus:iam:us-west-1:qe-aws:user:cloudbreak-qe@hortonworks.com"` and a real one: `"Cloudera-Creator-Resource-Name":"crn:altus:iam:us-west-1:9d74eee4-1cad-45d7-b645-7ccf9edbb73d:user:9e6ff6c7-7a4b-4086-8989-731c2d57fbf7"`

lnardai (approved):

PR #9261, opened 2020-10-21
CB-9174 - Creation of cluster definitions through the CLI is broken

gregito:
When a DH cluster creation has triggered with an existing custom cluster definition that contains storage location with template indicators (handlebars), the cluster creation fails because CB didn't replace the template placeholders with the actual value.


biharitomi:
@doktoric could you please check this?
I can remember a time when in this scenarios the templated values was resolved by using the storage configs from the related Datalake. So maybe that logic is not working if it should still work at all.

doktoric:
I think in case of templates we should generated the template itself right ?

doktoric (approved):

doktoric:
retest this please

doktoric:
retest this please

PR #9265, opened 2020-10-21
CB-9312 do not use stop or start with list endpoint because we are not able to authorize them at this point

sodre90:


sodre90:
I tested the env delete with multiple datahub and it worked

lnardai (approved):

biharitomi (approved):
LGTM, but please do not forget to create the follow up to eliminate these sequential calls to bulked ones

PR #9267, opened 2020-10-21
CB-9421. Use accountId (by param and not from CRN) for deleteAccessKe…

oleewere:
…ys request as well

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
@@ -425,26 +425,26 @@ public class UmsClient {
      * Remove a role (from machine user) - if role exists
      *
      * @param requestId      id of the request
-     * @param userCrn        actor user (account & assignee)
+     * @param accountId      account id for user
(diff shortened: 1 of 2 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
```diff
@@ -662,15 +662,15 @@ CreateAccessKeyResponse createAccessPrivateKeyPair(String requestId, String user
      * @param accessKeyCrns list of access key CRNs
      * @param actorCrn      user that executes the deletion
      */
-    void deleteAccessKeys(String requestId, List<String> accessKeyCrns, String actorCrn) {
+    void deleteAccessKeys(String requestId, List<String> accessKeyCrns, String actorCrn, String accountId) {
         checkNotNull(requestId);
         checkNotNull(actorCrn);
```
lnardai, line 667: This should be checked for accountID

lnardai: Also actor CRN is not needed here anymore.

lnardai (approved):

PR #9275, opened 2020-10-22
CB-9433 Cannot download the event json in zip format on event histroy…

topolyai5:
… page. When I try to download the event history in json format the response is only an empty list. We already filter by resource crn so we don't need to filter by resource type as well.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/db/LegacyStructuredEventRepository.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/db/LegacyStructuredEventRepository.java
@@ -28,7 +28,7 @@ public interface LegacyStructuredEventRepository extends WorkspaceResourceReposi
 
     List<StructuredEventEntity> findByWorkspaceAndResourceTypeAndResourceId(Workspace workspace, String resourceType, Long resourceId);
 
-    List<StructuredEventEntity> findByWorkspaceAndResourceTypeAndResourceCrn(Workspace workspace, String resourceType, String resourceCrn);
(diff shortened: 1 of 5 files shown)


doktoric:
I think some unit test would be great

topolyai5:
> I think some unit test would be great. Also we should add an integration test check for this just to make sure we will not break it.

I think the only tests can catch this kind of error, that is the UI test for the event history in zip format.

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/db/LegacyStructuredEventRepository.java
```diff
@@ -26,9 +26,9 @@
     @Query("SELECT se from StructuredEventEntity se WHERE se.workspace.id = :workspaceId AND se.id = :id")
     StructuredEventEntity findByWorkspaceIdAndId(@Param("workspaceId") Long workspaceId, @Param("id") Long id);
 
-    List<StructuredEventEntity> findByWorkspaceAndResourceTypeAndResourceId(Workspace workspace, String resourceType, Long resourceId);
+    List<StructuredEventEntity> findByWorkspaceAndResourceId(Workspace workspace, Long resourceId);
```
akanto, line 29: Unfortunately, this is not going to be unique. It can return multiple resources if the if matches. Maybe it would be better to deprecate this kind of access.

topolyai5: done

akanto (approved):

PR #9278, opened 2020-10-23
CB-9298: Instance status is incorrect on UI

christmasferret:
Integrated freeipa health check and platform instance check results to show actual instance statuses on UI. Currently instance status is incorrect on UI when the instances are unreachable because only platform instance check is used. The platform instance check doesn't check if the instance is reachable. If instances are not available, the UI will show incorrect running status. With this fix, the UI will show the instance as unreachable.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
@@ -2,8 +2,11 @@ package com.sequenceiq.freeipa.sync;
 
 import static com.sequenceiq.cloudbreak.util.Benchmark.checkedMeasure;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -29,9 +32,9 @@ public class FreeipaChecker {
     @Inject
     private FreeIpaHealthDetailsService freeIpaHealthDetailsService;
 
-    private Pair<List<DetailedStackStatus>, String> checkStatus(Stack stack, Set<InstanceMetaData> checkableInstances) throws Exception {
+    private Pair<Map<InstanceMetaData, DetailedStackStatus>, String> checkStatus(Stack stack, Set<InstanceMetaData> checkableInstances) throws Exception {
         return checkedMeasure(() -> {
-            List<DetailedStackStatus> statuses = new LinkedList<>();
+            Map<InstanceMetaData, DetailedStackStatus> statuses = new HashMap<>();
             List<RPCResponse<Boolean>> responses = new LinkedList<>();
             for (InstanceMetaData instanceMetaData : checkableInstances) {
                 try {
@@ -39,13 +42,13 @@ public class FreeipaChecker {
                             ":::Auto sync::: FreeIPA health check ran in {}ms");
                     responses.add(response);
                     if (response.getResult()) {
-                        statuses.add(DetailedStackStatus.AVAILABLE);
+                        statuses.put(instanceMetaData, DetailedStackStatus.AVAILABLE);
                     } else {
-                        statuses.add(DetailedStackStatus.UNHEALTHY);
+                        statuses.put(instanceMetaData, DetailedStackStatus.UNHEALTHY);
                     }
                 } catch (Exception e) {
                     LOGGER.info("FreeIpaClientException occurred during status fetch: " + e.getMessage(), e);
-                    statuses.add(DetailedStackStatus.UNREACHABLE);
+                    statuses.put(instanceMetaData, DetailedStackStatus.UNREACHABLE);
                 }
             }
             String message = getMessages(responses);
@@ -63,8 +66,8 @@ public class FreeipaChecker {
             Set<InstanceMetaData> notTermiatedStackInstances = stack.getAllInstanceMetaDataList().stream()
                     .filter(Predicate.not(InstanceMetaData::isTerminated))
                     .collect(Collectors.toSet());
-            Pair<List<DetailedStackStatus>, String> statusCheckPair = checkStatus(stack, checkableInstances);
-            List<DetailedStackStatus> responses = statusCheckPair.getFirst();
+            Pair<Map<InstanceMetaData, DetailedStackStatus>, String> statusCheckPair = checkStatus(stack, checkableInstances);
+            List<DetailedStackStatus> responses = new ArrayList<>(statusCheckPair.getFirst().values());
             DetailedStackStatus status;
             if (areAllStatusTheSame(responses) && !hasMissingStatus(responses, notTermiatedStackInstances)) {
                 status = responses.get(0);
@@ -72,10 +75,10 @@ public class FreeipaChecker {
                 status = DetailedStackStatus.UNHEALTHY;
             }
 
-            return new SyncResult("FreeIpa is " + status + ", " + statusCheckPair.getSecond(), status);
+            return new SyncResult("FreeIpa is " + status + ", " + statusCheckPair.getSecond(), status, statusCheckPair.getFirst());
         } catch (Exception e) {
             LOGGER.info("Error occurred during status fetch: " + e.getMessage(), e);
-            return new SyncResult("FreeIpa is unreachable, because error occurred: " + e.getMessage(), DetailedStackStatus.UNREACHABLE);
+            return new SyncResult("FreeIpa is unreachable, because error occurred: " + e.getMessage(), DetailedStackStatus.UNREACHABLE, null);
         }
     }
 
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.cloudbreak.util.Benchmark.checkedMeasure;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
@@ -16,6 +17,7 @@ import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.cloud.model.CloudVmInstanceStatus;
 import com.sequenceiq.flow.core.FlowLogService;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.DetailedStackStatus;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
 import com.sequenceiq.freeipa.entity.InstanceMetaData;
 import com.sequenceiq.freeipa.entity.Stack;
@@ -38,7 +40,8 @@ public class ProviderChecker {
     @Value("${freeipa.autosync.update.status:true}")
     private boolean updateStatus;
 
-    public List<ProviderSyncResult> updateAndGetStatuses(Stack stack, Set<InstanceMetaData> checkableInstances) {
+    public List<ProviderSyncResult> updateAndGetStatuses(Stack stack, Set<InstanceMetaData> checkableInstances,
+        Map<InstanceMetaData, DetailedStackStatus> instanceHealthStatusMap) {
         return checkedMeasure(() -> {
             List<ProviderSyncResult> results = new ArrayList<>();
             List<CloudVmInstanceStatus> statuses = stackInstanceProviderChecker.checkStatus(stack, checkableInstances);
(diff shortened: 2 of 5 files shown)


christmasferret:
Currently, the freeipa instance status checker only uses platform instance check API. If the instance is unavailable (for example, no network access), the instance status incorrectly shows running.  

Screen Shot 2020-10-23 at 9 02 05 AM

With this fix, we combine platform instance check and freeipa health check. Therefore the instance status shows unavailable as the correct status. This will help to find out the freeipa instance issue. 

Screen Shot 2020-10-23 at 9 05 30 AM

jamisonbennett:
I appreaciate the screenshots you include with your pull requests. They make it very clear.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -74,33 +77,39 @@
...
-                setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_ON_PROVIDER_SIDE);
-                status = InstanceStatus.DELETED_ON_PROVIDER_SIDE;
-                break;
-            case TERMINATED_BY_PROVIDER:
-                setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_BY_PROVIDER);
-                status = InstanceStatus.DELETED_BY_PROVIDER;
-                break;
-            default:
-                LOGGER.info(":::Auto sync::: the '{}' status is not converted", vmInstanceStatus.getStatus());
+        if (DetailedStackStatus.UNREACHABLE == instanceHealthStatusMap.get(instanceMetaData)) {
```
jamisonbennett, line 85: I think this might only work for cases where the status is unreachable. For example, if the instance health is unhealthy I suspect the UI will report "running" since it will rely on the VM status which will be started which means the instance status will be created which means the UI will report "running".

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -74,33 +77,39 @@
...
+                    status = InstanceStatus.FAILED;
+                    break;
+                case TERMINATED:
+                    setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_ON_PROVIDER_SIDE);
+                    status = InstanceStatus.DELETED_ON_PROVIDER_SIDE;
+                    break;
+                case TERMINATED_BY_PROVIDER:
+                    setStatusIfNotTheSame(instanceMetaData, InstanceStatus.DELETED_BY_PROVIDER);
+                    status = InstanceStatus.DELETED_BY_PROVIDER;
+                    break;
```
jamisonbennett, line 109: Since the unreachable's logic above is the highest precedence, it appears that unhealthy overrides stopped, failed, terminated, and terminated by the provider (all of which will report an unreachable status). I think the instance's health status should only be considered if the VM is in the running state.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -29,31 +32,31 @@
...
+                    statuses.put(instanceMetaData, DetailedStackStatus.UNREACHABLE);
                 }
             }
             String message = getMessages(responses);
             return Pair.of(statuses, message);
         }, LOGGER, ":::Auto sync::: freeipa server status is checked in {}ms");
     }
 
-    public SyncResult getStatus(Stack stack, Set<InstanceMetaData> checkableInstances) {
+    public SyncResult getStatus(Stack stack, Set<InstanceMetaData> checkableInstances, Map<InstanceMetaData, DetailedStackStatus> instanceHealthStatusMap) {
```
jamisonbennett, line 59: Instead of using an outbound parameter `instanceHealthStatusMap`, would it be better to include that as a part of the `SyncResult`?

jamisonbennett (commented):
I think the logic for status is getting very difficult to ensure correctness by looking at the code.

It seems like the responsibility of a few classes are so tightly coupled. I have thought it was a challenge before but now it seems like the coupling is after this PR is much stronger. I think the class names make it a little more confusing. So I would prefer some refactoring and we can discuss some ideas.

ProviderChecker:
Old Behavior: Only considered the instance status from the cloud provider. Stores the status of the instances to the database.
New Behavior: Requires instance health but checks the cloud provider status. Stores the status of the instances to the database. Has combiner logic which is similar to StackStatusCheckerJob::getStackStatus() in that it combines both VM cloud provider status & instance health status.

StackStatusCheckerJob:
Old Behavior: Called the cloud provider checker, called the health checks, stored the status of the stack in the database. StackStatusCheckerJob::getStackStatus() has the logic to combine the two status.
New Behavior: Same but passes around the instance status map to provider checker which has combiner similar logic.

FreeipaChecker
Old behavior: Performed health check & returned a sync result which had a single combined status based on all the instance status
New behavior: Same but also takes an outbound argument for the instance map

topolyai5:
> Currently, the freeipa instance status checker only uses platform instance check API. If the instance is unavailable (for example, no network access), the instance status incorrectly shows running.
> 
> Screen Shot 2020-10-23 at 9 02 05 AM
> 
> With this fix, we combine platform instance check and freeipa health check. Therefore the instance status shows unavailable as the correct status. This will help to find out the freeipa instance issue.
> 
> Screen Shot 2020-10-23 at 9 05 30 AM

If I understand correctly, is the instance running on the provider but the freeipa server is unreachable? AFAIK this status is point to the instance status and not the freeipa status.

topolyai5 (changes requested):
If I understand correctly, is the instance running on the provider but the freeipa server is unreachable? AFAIK this status is point to the instance status and not the freeipa status.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: Thanks for the suggestion! I added an UNHEALTHY status to the instancestatus enum to catch UNHEALTHY separately. If this is merged, I will also add the UNHEALTHY to cloudbreak-cloud UI repository to be able to display UNHEALTHY in UI (https://github.com/hortonworks/hortonworks-cloud/blob/master/web/cloudbreak-ui/src/app/helpers/freeipa.helpers.ts).

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: You are absolutely right. I made the change so it will not overwrite the failed and terminated status. Now only running instances that are unreachable or unhealthy are reported as it is. 

Tested again below,

Screen Shot 2020-10-26 at 9 59 25 AM
Screen Shot 2020-10-26 at 9 59 14 AM
Screen Shot 2020-10-26 at 9 59 09 AM

freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
christmasferret: This is great suggestion. I returned instanceHealthStatusMap in SyncResult now. It will make the code less coupled and more readable. I removed the carrier parameter.

christmasferret:
> If I understand correctly, is the instance running on the provider but the freeipa server is unreachable? AFAIK this status is point to the instance status and not the freeipa status.

Thanks for the review! You are right that the status only point to the instance status that platform sdk reports. So currently instances status will be "running" even it is unreachable. The reason we want to make it more user friendly is that user can identify which instance is actually failing when freeipa is unhealthy. Then a reboot or repair operation can be issued to fix the failed instance. Currently, developer and user will have no idea why the freeipa is not available from UI view. After we added this feature, the instance health is also reported and facilitate the user's understanding.

christmasferret:
> > Currently, the freeipa instance status checker only uses platform instance check API. If the instance is unavailable (for example, no network access), the instance status incorrectly shows running.
> > Screen Shot 2020-10-23 at 9 02 05 AM
> > With this fix, we combine platform instance check and freeipa health check. Therefore the instance status shows unavailable as the correct status. This will help to find out the freeipa instance issue.
> > Screen Shot 2020-10-23 at 9 05 30 AM
> 
> If I understand correctly, is the instance running on the provider but the freeipa server is unreachable? AFAIK this status is point to the instance status and not the freeipa status.

Thanks for the review! I replied to the comment above.

topolyai5:
> > If I understand correctly, is the instance running on the provider but the freeipa server is unreachable? AFAIK this status is point to the instance status and not the freeipa status.
> 
> Thanks for the review! You are right that now the status only point to the instance status that the platform sdk reports which is quite functionally inaccurate. So currently instances status will be "running" even it is unreachable. The reason we want to make it more user friendly is that user can identify which instance is actually failing when freeipa is unhealthy. Then a reboot or repair operation can be issued to fix the failed instance. Currently, developer and user will have no idea why the freeipa is not available from UI view. After we added this feature, the instance health is also reported and facilitate the user's understanding.

thanks, it's clear now.

topolyai5 (approved):

jamisonbennett (approved):
Thanks for addressing my comments. I would like @lacikaaa to take a look at this PR especially with respect to my comment ( https://github.com/hortonworks/cloudbreak/pull/9278#pullrequestreview-515753262) to see if additional refactoring would be helpful.

lacikaaa:
>  New Behavior: Requires instance health but checks the cloud provider status. Stores the status of the instances to the database. Has combiner logic which is similar to StackStatusCheckerJob::getStackStatus() in that it combines both VM cloud provider status & instance health status.

ok, I checked it and I know why it's a bit messy. I once filed this task: CB-7879
It's about we should check the provider first and won't try to connect freeipa if the instance is stopped/terminated.
But as we discussed it turned out it should be the other way around (like for CB and CM) and that should be implemented for FMS.

So ideally the status check should look like:
1. Try to reach FreeIPA and get health status
2.	a) If everything is ok skip provider check and set the status if there is a change
	b) If we detect we have issue with any of the nodes, most likely not reachable, check those on the provider side and refresh the statuses

does it makes sense to you guys? @jamisonbennett @christmasferret @topolyai5

jamisonbennett:
I agree with the order of CB-7879.

christmasferret:
> > New Behavior: Requires instance health but checks the cloud provider status. Stores the status of the instances to the database. Has combiner logic which is similar to StackStatusCheckerJob::getStackStatus() in that it combines both VM cloud provider status & instance health status.
> 
> ok, I checked it and I know why it's a bit messy. I once filed this task: CB-7879
> It's about we should check the provider first and won't try to connect freeipa if the instance is stopped/terminated.
> But as we discussed it turned out it should be the other way around (like for CB and CM) and that should be implemented for FMS.
> 
> So ideally the status check should look like:
> 
> 1. Try to reach FreeIPA and get health status
> 2. a) If everything is ok skip provider check and set the status if there is a change
>    b) If we detect we have issue with any of the nodes, most likely not reachable, check those on the provider side and refresh the statuses
> 
> does it makes sense to you guys? @jamisonbennett @christmasferret @topolyai5

@lacikaaa 

I made another commit here to make the change. Now it skips provider check if status is OK. I didn't squash so the difference is more clear.
https://github.com/hortonworks/cloudbreak/pull/9278/commits/85ca87fd33ec0f33c10648f5c29ee0e421760d31

I tested creating new instance, reboot instance and terminate instance as below.
Screen Shot 2020-10-28 at 2 29 34 PM
Screen Shot 2020-10-28 at 2 34 22 PM
Screen Shot 2020-10-28 at 2 40 10 PM
Screen Shot 2020-10-28 at 2 42 52 PM

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -100,11 +100,16 @@ public void syncAStack(Stack stack) {
                     }
                     if (!checkableInstances.isEmpty()) {
                         SyncResult syncResult = freeipaChecker.getStatus(stack, checkableInstances);
-                        List<ProviderSyncResult> results = providerChecker.updateAndGetStatuses(stack, checkableInstances, syncResult.getInstanceStatusMap());
-                        if (!results.isEmpty()) {
-                            updateStackStatus(stack, syncResult, results, alreadyDeletedCount);
+                        if (DetailedStackStatus.AVAILABLE == syncResult.getStatus()) {
```
jamisonbennett, line 103: I think I know the answer to this but I wanted to check. Is the reason that `DetailedStackStatus.UNHEALTHTY` can't bypass the cloud provider check because we need to catch the case where 1 VM is running and 1 VM is either stopped, deleted, etc?

christmasferret: DetailedStackStatus.AVAILABLE happens only when freeipa in all instances are available. That is skipped by this change.

lacikaaa (approved):
LGTM, just squash your commits please
thanks for the changes

PR #9279, opened 2020-10-23
CB-9449 if AWS returns that DBInstance does not exsist, then we shall…

akanto:
… consider this as the termination protection is not enabled.

Before this change if somebody went to AWS console and removed the protection and deleted the DB instance, that caused Environment/SDX/DH delete failures, since Redbeams would just thorw a CloudConnectorException for delete call and Environment/SDX/DH deletion stuck forever. After this change Redbeams realises that DB does not exsist anymore and it just proceeds and Environment/SDX/DH deletion will be successful.

I have intentionally changed the unit test as well, to make sure that it reflects the the expected behaviour.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsRdsStatusLookupService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsRdsStatusLookupService.java
@@ -37,6 +37,7 @@ public class AwsRdsStatusLookupService {
         try {
             describeDBInstancesResult = rdsClient.describeDBInstances(describeDBInstancesRequest);
         } catch (DBInstanceNotFoundException ex) {
+            LOGGER.debug("DB Instance does not exist: {}", ex.getMessage());
             return ExternalDatabaseStatus.DELETED;
         } catch (RuntimeException ex) {
             throw new CloudConnectorException(ex.getMessage(), ex);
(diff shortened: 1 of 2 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsRdsStatusLookupService.java
```diff
@@ -37,6 +37,7 @@ public ExternalDatabaseStatus getStatus(AuthenticatedContext ac, DatabaseStack d
         try {
             describeDBInstancesResult = rdsClient.describeDBInstances(describeDBInstancesRequest);
         } catch (DBInstanceNotFoundException ex) {
+            LOGGER.debug("DB Instance does not exists: {}", ex.getMessage());
```
gergopapi2, line 40: typo: exists => exist

gergopapi2 (changes requested):
LGTM, just some typos:
- commit message: thorw, exsist
- typos in log-lines

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsRdsStatusLookupService.java
akanto: Fixed

gergopapi2 (approved):

PR #9285, opened 2020-10-26
CB-9362 on aws rds postgres version is 10.14 instead of 10.6

afarsang:
CB-9362 on aws rds postgres version is 10.14 instead of 10.6

### DIFF
--- a/cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
+++ b/cloud-aws/src/main/resources/templates/aws-cf-dbstack.ftl
@@ -47,7 +47,7 @@
     },
     "EngineVersionParameter": {
         "Type": "String",
-        "Default": "10.6",
+        "Default": "10.14",
         "Description": "Engine version",
         "MinLength": 1,
         "MaxLength": 64


akanto (changes requested):
Embedded DB should default to the same version

pdarvasi:
> Embedded DB should default to the same version

@akanto  Do we specify a minor version for embedded install? https://github.com/hortonworks/cloudbreak-images/blob/76d44d0eddb60b4a64e84545c56bc53fde731a7a/saltstack/base/salt/postgresql/init.sls#L16-L18

pdarvasi (approved):

akanto:
You might be right: https://github.com/hortonworks/cloudbreak-images/blob/76d44d0eddb60b4a64e84545c56bc53fde731a7a/saltstack/base/salt/postgresql/yum/postgres10-el7.repo#L3 


And this is updated continuously: https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/

akanto (approved):

PR #9288, opened 2020-10-26
CB-9266 distroximagetest may run on gcp

afarsang:

### DIFF
--- a/integration-test/src/main/resources/application.yml
+++ b/integration-test/src/main/resources/application.yml
@@ -150,8 +150,9 @@ integrationtest:
     availabilityZone: europe-west2-a
     region: europe-west2
     location: europe-west2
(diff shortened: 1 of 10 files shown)


doktoric:
please squash the two commit

integration-test/src/main/resources/application.yml
```diff
@@ -163,6 +164,9 @@ integrationtest:
       volumeSize: 100
       volumeCount: 1
       volumeType: pd-standard
+    securityAccess:
+      knoxSecurityGroup: cloudbreak
```
doktoric, line 168: this should be empty

afarsang: ok

doktoric (approved):

PR #9290, opened 2020-10-26
CDPSDX-1981: Add retries when registering DNS A records

jamisonbennett:
Register DNS A records up to 3 times for datalake nodes waiting 1
second between attempts.

While I agree this change should not be required, in practise there
there is evidence that something isn't persisting as expected. See
CDPSDX-1981 for details and reference the linked JIRAs. This has been
"fixed" about a dozen times. While I know this will not fix the
underlying problem, I think it will likley reduce the frequency of the
intermittent issue.


jasonwmcswain (approved):
looks good to me.

jamisonbennett:
I should also mention testing....since I can't reproduce the issue reliably, I don't have much to say about testing other than it didn't make things worse.

lacikaaa:
thanks. I'm adding the do not merge label until we verify the issue and if this change is really necessary

jamisonbennett:
I am dropping this in favor of CB-9483.

PR #9292, opened 2020-10-27
CB-9325 All entitlements in one place and entitlement check result is logged

foldik:


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -511,7 +512,7 @@ public boolean isAuthorizationEntitlementRegistered(String actorCrn, String acco
         return getAccountDetails(actorCrn, accountId, MDCUtils.getRequestId()).getEntitlementsList()
                 .stream()
                 .map(e -> e.getEntitlementName().toUpperCase())
-                .anyMatch(e -> e.equalsIgnoreCase("CB_AUTHZ_POWER_USERS"));
+                .anyMatch(e -> e.equalsIgnoreCase(Entitlement.CB_AUTHZ_POWER_USERS.name()));
```
horadla23, line 515: Entitlement check for authz entitlement still not logged here. The main reason for jira task creation was to use EntitlementService here and also resolve circular dependency between GrpcUmsClient and EntitlementService

horadla23 (changes requested):

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
foldik: Yes, this is the reason why it is in do_not_merge state, but I'll put it into draft to make it more clear.

foldik: I'll leave it simply in DO_NOT_MERGE.

horadla23: oh, ok, yes, makes sense to make it draft for now

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -503,15 +504,24 @@ public String rightCheckToString(AuthorizationProto.RightCheck rightCheck) {
...
 
-    @Cacheable(cacheNames = "umsAuthorizationEntitlementRegisteredCache", key = "{ #actorCrn, #accountId }")
-    public boolean isAuthorizationEntitlementRegistered(String actorCrn, String accountId) {
-        if (StringUtils.equals(accountId, InternalCrnBuilder.INTERNAL_ACCOUNT)) {
-            return false;
+    @Cacheable(cacheNames = "umsEntitlementRegisteredCache", key = "{ #actorCrn, #accountId, #entitlement }")
+    public boolean isEntitled(String actorCrn, String accountId, Entitlement entitlement) {
+        boolean entitled;
+        if (Entitlement.CB_AUTHZ_POWER_USERS.equals(entitlement) && StringUtils.equals(accountId, InternalCrnBuilder.INTERNAL_ACCOUNT)) {
+            entitled = false;
```
horadla23, line 511: maybe it is better to fail here if account id is internal (which should not ever happen), so i would not check for that, anyway CB will fail somewhere else also if internal account id reaches service layer

horadla23 (changes requested):

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: nvm i put that kind of if statement because of an issue

horadla23 (approved):

gergopapi2 (approved):

PR #9294, opened 2020-10-27
CB-9109 Freeipa launch crippled by missing storage account

gergopapi2:
With rebase another small refactoring was necessary: 
azureUtils.getImageNameFromConnectionString(image) was moved into CustomVMImageNameProvider and into AzureImageInfo.

=====

On azure, during freeipa launch, a 1) storage account is created, 2) the corresponding vhd image file is copied into the storage account, and 3) an finally azure managed image is created from the vhd file.

If an image already exists, then the whole process should be bypassed. However, if the storage account is deleted but the image exists an error is produced.

In this commit the bug is fixed by refactoring.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
@@ -12,11 +12,8 @@ import com.microsoft.azure.CloudException;
 import com.microsoft.azure.management.compute.VirtualMachineCustomImage;
 import com.sequenceiq.cloudbreak.cloud.azure.AzureImage;
 import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClient;
-
-import com.sequenceiq.cloudbreak.cloud.azure.resource.AzureResourceIdProviderService;
 import com.sequenceiq.cloudbreak.cloud.azure.task.image.AzureManagedImageCreationCheckerContext;
 import com.sequenceiq.cloudbreak.cloud.azure.task.image.AzureManagedImageCreationPoller;
-import com.sequenceiq.cloudbreak.cloud.azure.util.CustomVMImageNameProvider;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
 import com.sequenceiq.cloudbreak.cloud.exception.CloudConnectorException;
 import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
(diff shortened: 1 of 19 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
```diff
@@ -135,7 +106,7 @@ public CloudResource buildCloudResource(String name, String id, CommonStatus sta
                 .build();
     }
 
-    private boolean isRequested(String imageId) {
-        return resourcePersistenceRetriever.notifyRetrieve(imageId, CommonStatus.REQUESTED, ResourceType.AZURE_MANAGED_IMAGE).isPresent();
+    private boolean wasCreateRequested(AzureImageDetails azureImageDetails) {
```
pdarvasi, line 109: I'd still prefer isRequested as it check for current status "REQUESTED"

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
```diff
@@ -36,96 +33,70 @@
...
         }
+
+        LOGGER.debug("Custom image found in '{}' resource group with name '{}'", azureImageDetails.getResourceGroup(), azureImageDetails.getImageName());
+        azureManagedImageCreationPoller.startPolling(ac, new AzureManagedImageCreationCheckerContext(azureImageDetails, client));
+        return Optional.of(new AzureImage(azureImageDetails.getImageId(), azureImageDetails.getImageName(), true));
     }
 
-    private AzureImage createNewAzureImageAndNotify(AuthenticatedContext ac, VirtualMachineCustomImage customImage) {
-        updateImageStatus(ac, customImage.name(), customImage.id(), CommonStatus.CREATED);
+    public AzureImage createCustomImage(AzureImageDetails azureImageDetails, String fromVhdUri, AzureClient client, AuthenticatedContext ac) {
```
pdarvasi, line 52: Generally, I don't like the image --> customImage renames, as it is an Azure Java SDK only term, the object type is "type": "Microsoft.Compute/images" and in Portal, it is called "Image" in Azure Portal, too.
Furthermore, the custom image term in Cloudbreak is used for a completely different concept. Pls. undo the renames. 
It would be best to rename all occurrences, regardless of this PR.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
```diff
@@ -36,96 +33,70 @@
...
-    }
-
-    private String getImageName(String region, String fromVhdUri) {
-        return customVMImageNameProvider.get(region, fromVhdUri);
+    private Optional<VirtualMachineCustomImage> findCustomImage(AzureImageDetails azureImageDetails, AzureClient client) {
+        return azureManagedImageService.findVirtualMachineCustomImage(azureImageDetails, client);
     }
 
-    private void saveImage(AuthenticatedContext ac, String imageName, String imageId) {
+    private void saveCustomImage(AuthenticatedContext ac, String imageName, String imageId) {
```
pdarvasi, line 89: same: pls. rename the rename

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
```diff
@@ -36,96 +33,70 @@
...
         LOGGER.debug("Persisting image with REQUESTED status: {}", imageId);
         persistenceNotifier.notifyAllocation(buildCloudResource(imageName, imageId, CommonStatus.REQUESTED), ac.getCloudContext());
     }
 
-    private void updateImageStatus(AuthenticatedContext ac, String imageName, String imageId, CommonStatus commonStatus) {
-        LOGGER.debug("Updating image status to {}: {}", commonStatus.toString(), imageId);
+    private void updateCustomImageStatus(AuthenticatedContext ac, String imageName, String imageId, CommonStatus commonStatus) {
```
pdarvasi, line 94: same: pls. rename the rename

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageDetails.java
```diff
@@ -0,0 +1,35 @@
+package com.sequenceiq.cloudbreak.cloud.azure.image;
+
+public class AzureImageDetails {
```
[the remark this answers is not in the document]
gergopapi2: While it's true they have common fields but there are some not in common, as well. They also serve slightly different purposes.

I view it as:
input: AzureImageDetails. Grouping all the needed input parameters into one class
output: AzureImage. Tells, for example, if an image exists.

I would keep it.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
gergopapi2: You mean in AzureClient as well? Doing it.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageDetails.java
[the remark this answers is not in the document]
gergopapi2: I renamed it to AzureImageInfo in hope to give it a better name.

PR #9296, opened 2020-10-27
CB-9460 add SRM to the known service list and we can provide service d…

doktoric:
…efintion for the UI


biharitomi (approved):
LGTM

PR #9299, opened 2020-10-27
Resource deletion should break test case when a resource reaches the 'DELETE_FAILED' state

biharitomi:
Changes:
 - enforce delete yarn environment from test to be able to identify yarn environment deletion issues
 - enforce TestFailException in case of DELETE_FAILED status during resource terminations

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/hybrid/AwsYcloudHybridCloudTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/hybrid/AwsYcloudHybridCloudTest.java
@@ -210,6 +210,9 @@ public class AwsYcloudHybridCloudTest extends AbstractE2ETest {
                     }
                     return dto;
                 })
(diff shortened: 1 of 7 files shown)


aszegedi:
I think the [integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/hybrid/AwsYcloudHybridCloudTest.java tearDown](https://github.com/hortonworks/cloudbreak/pull/9299/files#diff-fc2c3effe9c5b6448398444ea9a0899928f50a5f24378e256e5baabc4e54a24fR219-R230) is not necessary anymore if we move back the:
```
.given(CHILD_ENVIRONMENT_KEY, EnvironmentTestDto.class, CHILD_CLOUD_PLATFORM)
.when(environmentTestClient.cascadingDelete(), RunningParameter.key(CHILD_ENVIRONMENT_KEY))
.await(EnvironmentStatus.ARCHIVED, RunningParameter.key(CHILD_ENVIRONMENT_KEY))
```
block to the test.

aszegedi (approved):

Bajzathd (approved):

PR #9302, opened 2020-10-27
CB-9472 make authz mock tests runnable from one testsuite

horadla23:

### DIFF
--- a/integration-test/scripts/fetch-secrets.sh
+++ b/integration-test/scripts/fetch-secrets.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-secret_version="92e72b83dc3847048cb56c6f1345812f"
+secret_version="9924e0b4b5314f389a0e49c04bb21739"
 
 USER_JSON_LOCATION="./src/main/resources/ums-users/api-credentials.json"
 
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUser.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUser.java
@@ -12,6 +12,8 @@ public class CloudbreakUser {
 
     private String displayName;
 
+    private boolean admin;
+
     public CloudbreakUser() {
     }
 
@@ -25,6 +27,13 @@ public class CloudbreakUser {
         this.accessKey = accessKey;
         this.secretKey = secretKey;
         this.displayName = displayName;
+    }
+
+    public CloudbreakUser(String accessKey, String secretKey, String displayName, boolean admin) {
+        this.accessKey = accessKey;
+        this.secretKey = secretKey;
+        this.displayName = displayName;
+        this.admin = admin;
     }
 
     public String getAccessKey() {
@@ -57,6 +66,14 @@ public class CloudbreakUser {
 
     public String getCrn() {
         return crn;
+    }
+
+    public boolean getAdmin() {
+        return admin;
+    }
+
+    public void setAdmin(boolean admin) {
+        this.admin = admin;
     }
 
     public static void validateRealUmsUser(CloudbreakUser user) {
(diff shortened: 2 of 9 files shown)


lnardai:
Could we should redirect this to master, on 2.32 we can live with the 2 separate suits?

akanto:
Commit message should be extended with a detailed description of what problem is solved with this commit

horadla23:
> Could we should redirect this to master, on 2.32 we can live with the 2 separate suits?

yes

> Commit message should be extended with a detailed description of what problem is solved with this commit

this is still a draft, but of course, i'll extend it

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1603

integration-test/scripts/fetch-secrets.sh
```diff
@@ -14,7 +14,7 @@ fi
 mkdir -p ./src/main/resources/ums-users
 
 echo "Executing secret fetching from Azure 'jenkins-secret' store"
-az keyvault secret show --name "real-ums-users-dev" --vault-name "jenkins-secret" --query 'value' -o tsv | jq > $USER_JSON_LOCATION
```
lnardai, line 17: why create a new map? 
`az keyvault secret show --name "real-ums-users-dev" --vault-name "jenkins-secret" --version $secret_version --query 'value' -o tsv | jq > $USER_JSON_LOCATION`

I've a similar approach in another PR. This way we can easily update the secret.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUser.java
```diff
@@ -12,19 +12,30 @@
 
     private String displayName;
 
+    private Boolean admin;
```
lnardai, line 15: can't we just use the primitive? Constructors could be simpler also.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -336,7 +335,7 @@ T doAction(T entity, Class<? extends MicroserviceClient> clientClass, Action<T,
 
     protected <T extends CloudbreakTestDto, U extends MicroserviceClient>
     T doActionAsAdmin(T entity, Class<? extends MicroserviceClient> clientClass, Action<T, U> action, String who) throws Exception {
-        return action.action(getTestContext(), entity, getAdminMicroserviceClient(clientClass));
+        return action.action(getTestContext(), entity, getAdminMicroserviceClient(clientClass, Crn.fromString(entity.getCrn()).getAccountId()));
```
lnardai, line 338: Will this work with the old local setup for dp cli command? Since in that run there is no concept of userCrn.

lnardai (approved):

lnardai (changes requested):
Mistakenly clicked approved, wanted to request changes, sorry for that.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
horadla23: what do you mean by old local setup? i guess tests should run against same version of cbd, or is there some tests which uses older cbd than the version of integration-test.jar? here i am using the crn of the entity, not usercrn, so it should work. (side note: this method is not used, so probably i will delete it when i update this pr with secret versioning change)

integration-test/scripts/fetch-secrets.sh
horadla23: sure, makes sense, i'll wait for your pr and rebase onto that commit to use that solution

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUser.java
horadla23: sure

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
horadla23: removed

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1630/

horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1636/

aszegedi (approved):

lnardai (approved):

PR #9310, opened 2020-10-28
CB-9306 - Azure ARM templates - API version upgrade

pdarvasi:
Upgrading ARM API versions to a more current one.


gergopapi2 (approved):
LGTM. 
Question: don't we want to remove these arm files:
cloud-azure/src/test/resources/templates/arm-*.ftl

pdarvasi:
> cloud-azure/src/test/resources/templates

Those are still being used in AzureTemplateBuilderTest

PR #9314, opened 2020-10-28
ENGESC-4921 Handle Azure custom image with SAS token

Bajzathd:
When extracting image name from connection string expect query paramters to follow the image name. Also added Linux specific howto to readme.


daszabo (approved):

biharitomi (approved):
LGTM

PR #9319, opened 2020-10-28
CB-9503 - Env creation fails with private endpoint when created with …

pdarvasi:
…restricted resource group permissions

Contains 2 fixes:

1. sets the deployment status properly after exception
2. delegates the original exception to reach the UI

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureDnsZoneService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureDnsZoneService.java
@@ -106,8 +106,13 @@ public class AzureDnsZoneService {
                     updateCloudResource(authenticatedContext, deploymentName, dnsZoneDeploymentId, CommonStatus.CREATED, AZURE_PRIVATE_DNS_ZONE);
 
                 }
-            } catch (CloudException | DataAccessException e) {
+            } catch (CloudConnectorException e) {
                 LOGGER.warn("Deployment {} failed due to {}", deploymentName, e.getMessage());
+                pollForCreation(authenticatedContext, azureClient, resourceGroup, deploymentName, dnsZoneDeploymentId,
+                        enabledPrivateEndpointServices, null);
+                throw e;
+            } catch (DataAccessException e) {
+                LOGGER.warn("Polling {} deployment due to db unique constraint violation: {}", deploymentName, e.getMessage());
                 pollForCreation(authenticatedContext, azureClient, resourceGroup, deploymentName, dnsZoneDeploymentId,
                         enabledPrivateEndpointServices, null);
             }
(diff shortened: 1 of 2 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureDnsZoneService.java
```diff
@@ -106,10 +106,15 @@ public void getOrCreateDnsZones(AuthenticatedContext authenticatedContext, Azure
                     updateCloudResource(authenticatedContext, deploymentName, dnsZoneDeploymentId, CommonStatus.CREATED, AZURE_PRIVATE_DNS_ZONE);
 
                 }
-            } catch (CloudException | DataAccessException e) {
+            } catch (CloudConnectorException e) {
                 LOGGER.warn("Deployment {} failed due to {}", deploymentName, e.getMessage());
                 pollForCreation(authenticatedContext, azureClient, resourceGroup, deploymentName, dnsZoneDeploymentId,
```
gergopapi2, line 111: Just for clarification: why is poll for creation started when an exception is thrown? I have not found a catch clause in calling methods, I guess the creation will fail.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/task/dnszone/AzureDnsZoneCreationCheckerTask.java
```diff
@@ -42,10 +42,10 @@ protected Boolean doCall() {
         ResourceStatus templateDeploymentStatus = azureClient.getTemplateDeploymentStatus(resourceGroupName, deploymentName);
 
         if (templateDeploymentStatus == ResourceStatus.DELETED) {
-            throw new CloudConnectorException(String.format("Deployment %s is either deleted of does not exist", deploymentName));
+            throw new CloudConnectorException(String.format("Deployment %s is either deleted or does not exist", deploymentName));
```
gergopapi2, line 45: Another question: in case there was no permission to create a DNS zone, how come this error message was shown to the user? How did the deployment get deleted?

gergopapi2 (commented):
I just have two questions. I might be the incompetent dev as well :D

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/task/dnszone/AzureDnsZoneCreationCheckerTask.java
pdarvasi: Because this exception got propagated to the UI instead of the Azure exception, before the fix

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureDnsZoneService.java
pdarvasi: The poll here is started to handle the status change and refresh the status of the resource to the up-to-date status seen in the cloud provider.

keyki (approved):

PR #9324, opened 2020-10-28
Cb 9470

orngejaket:
This removes the find first filter in SecurityAccessManifest so that secuirty groups will be applied to all instance groups.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java
@@ -26,8 +26,7 @@ public class SecurityAccessManifester {
     public void overrideSecurityAccess(InstanceGroupType instanceGroupType, List<InstanceGroupV4Request> instanceGroups, String securityGroupId, String cidrs) {
         instanceGroups.stream()
                 .filter(ig -> ig.getType() == instanceGroupType)
-                .findFirst()
-                .ifPresent(ig -> {
+                .forEach(ig -> {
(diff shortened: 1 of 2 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java
```diff
@@ -26,8 +26,7 @@
     public void overrideSecurityAccess(InstanceGroupType instanceGroupType, List<InstanceGroupV4Request> instanceGroups, String securityGroupId, String cidrs) {
         instanceGroups.stream()
                 .filter(ig -> ig.getType() == instanceGroupType)
-                .findFirst()
```
kmanamcheri, line 29: @doktoric What was the original intention of applying this only on the first node for a given instance type?

kmanamcheri (approved):

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java
doktoric: if was not the first node but the instancegroup where the type equals

doktoric (approved):

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java
kmanamcheri: Ah sorry my bad. What was the original intention of applying this only on the first instancegroup. This assumes that there could be only one set of nodes for a given instancegroup, no?

PR #9325, opened 2020-10-29
CB-9452 Fix image import/export issue for Google cloud

cegganesh84:
1. Google for their default centos 7 deprecated SECURE_BOOT guest os feature and started using UEFI_COMPATIBLE flag https://github.com/hashicorp/packer/issues/9353#issuecomment-640484734
2. When we import such images without this flag, then instances could not be created from that
3. Tested by importing this flag with gcloud apis and made sure that the instance booted properly.


doktoric (approved):

biharitomi (approved):
LGTM

PR #9327, opened 2020-10-29
CB-9524 - Skip swagger compatibility check if there is no api change in the pull request

attilapalfi92:
- The list of non-API modules is a hardcoded array
- We iterate through all the changed files
- Check if their module is one of the non-API modules
- If yes, we add them to a collector array
- If in the the size of the collector array equals the number of changed files, no swagger check is needed

### DIFF
--- a/integration-test/scripts/swagger-check.sh
+++ b/integration-test/scripts/swagger-check.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -ex
+set -e
 
 date
 echo -e "\n\033[1;96m--- Install cbd\033[0m\n"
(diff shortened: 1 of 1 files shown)


biharitomi (approved):
LGTM, Thanks

attilapalfi92:
`Checking if there are any API changes between this and the target branch
+ apiChanged=false
++ git diff --name-status master
++ grep autoscale-api
++ wc -l
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
+ '[' 0 -gt 0 ']'
++ git diff --name-status master
 `

What is the problem here? @biharitomi

akanto:
Swagger compatibility test has failed...

Unfortunately, this might not be enough since, because we cannot ensure that there is no gradle import from other modules... e.g. incompatible changes in a pojo in structuredevents module will break API if this merge is fixed then that remains unnoticed

attilapalfi92:
@akanto I know it failed, see the previous comment. 

Regarding the example, I can add a line to check if there were changes in the structured event model or not. I can add common-model too because its used in core-api and freeipa-api. Are there anything else I should add?

attilapalfi92:
Added 
structuredevent-api-cdp, structuredevent-model and common-model to the list.

akanto:
>Are there anything else I should add?

I think we need to change the approach, and most probably we need to flip the logic and put modules to the skip list, instead of defining a list of modules that needs to be checked.

Just imagine the following scenario:
 - someone introduces a new module that has an effect on swagger API then incompatible changes remaining unnoticed since it is not on your list at the moment
 - next time when you send your unrelated change, e.g. you are introducing a compatible change into one of the API modules, your pull request will just fail, and you will be frustrated again

attilapalfi92:
I rewrote the checking logic as @akanto requested.

integration-test/scripts/swagger-check.sh
```diff
@@ -41,6 +41,129 @@ Field_Separator=$IFS
...
+  "cloud-reactor"
+  "docker-autoscale"
+  "environment"
+  "integration-test"
+  "redbeams-model"
+  "structuredevent-service"
+  "auth-connector"
+  "client-cm"
+  "cloud-template"
+  "common"
```
akanto, line 68: Looks almost good, unfortunately, common is included in environment-api, so if you change things there it might be still unnoticed

https://github.com/hortonworks/cloudbreak/blob/7959221b2102b2ac5c4ff5899dcdb687d8a968b9/environment-api/build.gradle#L17

Can we also somehow check the transitive dependencies of the modules? Maybe it does not worth it, if it is too much effort...

attilapalfi92: I cannot do the transitive dependency check :( Or it would take 1 to 2 weeks.

attilapalfi92: Removed the common from the non-API modules

akanto: Understood, probably does not worth the effort...

akanto (approved):

oleewere (approved):

PR #9331, opened 2020-10-29
CB-9529: Revert "CB-8969 Expand B/R flow to cover full backup/restore instead of just database"

kkalvagadda1:
There are multiple issues in the change merged in CB-8969.
 
Here is the list of issues

1. Database operation is set to failure when there is a failure in fetching overall backup-restore status.

2. Wrong id is used while polling overall data lake backup-restore status so the status of the database backup will always be set to failure/

3. Currently the flows end if the database backup-restore operation fails. Instead, polling for over-all status should be performed.

4. Currently, only backup-id is passed to restore flow but we also need the restore-id to perform a status lookup in datalakeFullResoreInProgress state. This needs changes to API and more testing.

I made code changes to fix most of them but one of the fixes needs API change which I will continue working on. As this issue reported in CB-8969 would break datalake-dr functionality which is already promoted to stage/prod, I prefer to first revert CB-8969 and commit again with appropriate changes.


lnardai:
> As this issue reported in CB-8969 would break datalake-dr functionality which is already promoted to stage/prod, I prefer to first revert CB-8969 and commit again with appropriate changes.

If you want to do a HOTFIX for 2.32 you should target this PR to CB-2.32.0, and once it get's merged then create a PR for master. This is how we handle Prod hotfixes.

PR #9334, opened 2020-10-29
CB-9127 refactor the current cluster template logic and separate it i…

doktoric:
…nto two flow. Default flow will not create stack entity and this means we only store the template. The other flow will add an original node count and the we dont need to collect instancemetadatas.


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/clustertemplate/ClusterTemplateToClusterTemplateV4ResponseConverter.java
```diff
@@ -32,9 +39,21 @@ public ClusterTemplateV4Response convert(ClusterTemplate source) {
...
+            clusterTemplateV4Response.setDistroXTemplate(getIfNotNull(stackV4Request, stackV4RequestConverter::convert));
+            clusterTemplateV4Response.setNodeCount(stack.getFullNodeCount());
+        } else {
+            try {
+                DefaultClusterTemplateV4Request clusterTemplateV4Request = new Json(getTemplateString(source))
+                        .get(DefaultClusterTemplateV4Request.class);
+                clusterTemplateV4Response.setDistroXTemplate(clusterTemplateV4Request.getDistroXTemplate());
+                clusterTemplateV4Response.setNodeCount(getFullNodeCount(clusterTemplateV4Request.getDistroXTemplate()));
+            } catch (IOException e) {
+                clusterTemplateV4Response.setDistroXTemplate(null);
```
topolyai5, line 54: please add some logs here

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/ResourceStatus.java
```diff
@@ -4,5 +4,13 @@
     DEFAULT,
     DEFAULT_DELETED,
     USER_MANAGED,
-    OUTDATED
+    OUTDATED;
```
akanto, line 7: Do we need this? It is going to break during CB upgrade.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/view/ClusterTemplateClusterApiView.java
```diff
@@ -0,0 +1,84 @@
...
+@SuppressFBWarnings("UWF_UNWRITTEN_FIELD")
+@Table(name = "Cluster")
+public class ClusterTemplateClusterApiView extends CompactView {
+    @OneToOne(fetch = FetchType.LAZY)
+    private ClusterTemplateStackApiView stack;
+
+    @ManyToOne(fetch = FetchType.EAGER)
+    private BlueprintView blueprint;
+
+    private String clusterManagerIp;
```
akanto, line 27: Do we need an IP in the template?

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/view/ClusterTemplateStackApiView.java
```diff
@@ -0,0 +1,184 @@
...
+    private ClusterTemplateClusterApiView cluster;
+
+    @Column(columnDefinition = "TEXT")
+    private String cloudPlatform;
+
+    @Column(columnDefinition = "TEXT")
+    private String platformVariant;
+
+    @OneToOne
+    private StackStatusView stackStatus;
```
akanto, line 41: Do we need an Status in the template?

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/view/ClusterTemplateStackApiView.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @OneToMany(mappedBy = "stack", fetch = FetchType.EAGER)
+    private Set<ClusterTemplateInstanceGroupView> instanceGroups = new HashSet<>();
+
+    private Long created;
```
akanto, line 46: I think we can eliminate a few fields/relations from this entity.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/ResourceStatus.java
doktoric: I just put a ; sign at the end

core/src/main/java/com/sequenceiq/cloudbreak/service/template/ClusterTemplateService.java
```diff
@@ -264,25 +271,42 @@ public void updateDefaultClusterTemplates(long workspaceId) {
...
-            msg = "Cluster definition should contain a cluster template!";
+        String blueprintName = null;
+        if (clusterTemplate.getStatus().isDefault()) {
+            try {
+                blueprintName = new Json(getTemplateString(clusterTemplate)).get(DefaultClusterTemplateV4Request.class)
+                        .getDistroXTemplate()
+                        .getCluster()
+                        .getBlueprintName();
+            } catch (IOException e) {
+                msg = "The cluster template in the cluster definition should be an exists!";
```
biharitomi, line 287: Please correct the message and also log it.

biharitomi (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/template/ClusterTemplateService.java
```diff
@@ -313,16 +337,17 @@ private void updateDefaultClusterTemplates(Workspace workspace) {
...
-            clusterTemplateLoaderService.loadClusterTemplatesForWorkspace(clusterTemplates, this::createAll);
+            clusterTemplateLoaderService.loadClusterTemplatesForWorkspace(clusterTemplates, workspace, this::createAll);
             LOGGER.debug("ClusterTemplate modifications finished based on the defaults for '{}' workspace.", workspace.getId());
         }
     }
 
     private Collection<ClusterTemplate> createAll(Iterable<ClusterTemplate> clusterTemplates) {
-        return StreamSupport.stream(clusterTemplates.spliterator(), false)
+        User user = userService.getOrCreate(legacyRestRequestThreadLocalService.getCloudbreakUser());
+        return StreamSupport.stream(clusterTemplates.spliterator(), true)
```
sodre90, line 347: Can we set it to false as it was? As I know it is not a very good idea to use parallal streams in a multithreaded web based application because it can cause high load for one request and the application should be able to serve many request at the same time.

biharitomi (approved):

PR #9337, opened 2020-10-29
CB-9540 Revert pillar value lookup

brycederriso:
Fixes [CB-9540](https://jira.cloudera.com/browse/CB-9540).

Revert the lookup of pillar values for cacerts. Use hardcoded values instead.


lajosrodek (approved):
LGTM

lajosrodek:
Follow-up to #9254.

PR #9338, opened 2020-10-30
CB-8962: Added a Cloudwatch alarm to each AWS datalake instance created.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-8962

What I did:

* Used the FreeIPA cloudwatch alarm knowledge to understand how to extend this to the datalake AWS instances
* Added the Cloudwatch alarm parameter to the stackv4request when it is created in the datalake setup flow
* Changed the AwsStackV4Parameters class to include the logic for this new parameter

Images of the results are shown in the following links:

* [All of the instances having alarms associated with them](https://drive.google.com/file/d/1uaAubywsGSkfQQHO5ouGSbefHAKGKbX7/view?usp=sharing)
* [The alarm associated with each instance](https://drive.google.com/file/d/111JpPJt6JuNfpbEr4_X4vPqt_E_HSExV/view?usp=sharing)

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -34,19 +34,19 @@ public class AwsCloudWatchService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsCloudWatchService.class);
 
-    @Value("${freeipa.aws.cloudwatch.suffix:-Status-Check-Failed-System}")
+    @Value("${aws.cloudwatch.suffix:-Status-Check-Failed-System}")
     private String alarmSuffix;
 
-    @Value("${freeipa.aws.cloudwatch.period:60}")
+    @Value("${aws.cloudwatch.period:60}")
     private int cloudwatchPeriod;
 
-    @Value("${freeipa.aws.cloudwatch.evaluationPeriods:2}")
+    @Value("${aws.cloudwatch.evaluationPeriods:2}")
     private int cloudwatchEvaluationPeriods;
 
-    @Value("${freeipa.aws.cloudwatch.threshold:1.0}")
+    @Value("${aws.cloudwatch.threshold:1.0}")
     private double cloudwatchThreshhold;
 
-    @Value("${freeipa.aws.cloudwatch.max-batchsize:100}")
+    @Value("${aws.cloudwatch.max-batchsize:100}")
     private int maxBatchsize;
 
     @Inject
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/base/parameter/stack/AwsStackV4Parameters.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/base/parameter/stack/AwsStackV4Parameters.java
@@ -6,20 +6,40 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts;
 import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
+import java.util.Map;
+
 @ApiModel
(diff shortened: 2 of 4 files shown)


sxxgrc:
I will look over the test errors tomorrow.

jamisonbennett (commented):
Thanks for making this change. I think this looks like it should work correctly, however, I wanted to get a second opinion on whether it is OK to rename the config settings. I agree with the rename but I want to ensure it is not overridden in a different repo which before I approve.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -34,19 +34,19 @@
...
+    @Value("${aws.cloudwatch.evaluationPeriods:2}")
     private int cloudwatchEvaluationPeriods;
 
-    @Value("${freeipa.aws.cloudwatch.threshold:1.0}")
+    @Value("${aws.cloudwatch.threshold:1.0}")
     private double cloudwatchThreshhold;
 
-    @Value("${freeipa.aws.cloudwatch.max-batchsize:100}")
+    @Value("${aws.cloudwatch.max-batchsize:100}")
     private int maxBatchsize;
```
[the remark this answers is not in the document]
lacikaaa: LGTM, but as the PR touches datalake service I have added @sodre90 and @lnardai to have a look at it

orngejaket:
This looks fine to me, but someone else must approve.  Agree with Jamison's rename comments.

sxxgrc:
Not really sure how to fix the Swagger compatibility error. If one of you has an insight on what is happening that would be helpful.

Already rebased to master and checked to see if the typo change I made was the issue.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -287,4 +298,10 @@ void setupInstanceVolumeEncryption(StackV4Request stackRequest, String cloudPlat
         }
     }
 
+    @VisibleForTesting
```
kmanamcheri, line 301: This is marked as VisibleForTesting, but I don't see any tests using this.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/base/parameter/stack/AwsStackV4Parameters.java
```diff
@@ -6,20 +6,40 @@
...
 import io.swagger.annotations.ApiModelProperty;
 
+import java.util.Map;
+
 @ApiModel
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(Include.NON_NULL)
 public class AwsStackV4Parameters extends StackV4ParameterBase {
+    @ApiModelProperty
+    private boolean createCloudwatch;
```
kmanamcheri, line 22: You have added a new model to `StackV4Request`. This modifies the final API of the datalake service. You should modify the tests to include this.

kmanamcheri (changes requested):
@sxxgrc The swagger compatibility tests are failing because you are modifying the Datalake API's `StackV4Request` object. Either make them backwards compatible, or modify the tests to include the new field you added.

If you look at the [swagger test result](http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-pull-request-builder-swagger-compatibility-test/2476/console), you'll notice this

```
[36mswagger-diff_1        |[0m check to /json/datalake-swagger-2.32.0-b196.json
[36mswagger-diff_1        |[0m - incompatible request params
[36mswagger-diff_1        |[0m   - post /internal/sdx/{}
[36mswagger-diff_1        |[0m     - missing request param: stackV4Request/aws (in: body, type: Hash[string, *])
[36mswagger-diff_1        |[0m - incompatible response attributes
[36mswagger-diff_1        |[0m   - get /sdx/crn/{}/detail
[36mswagger-diff_1        |[0m     - missing attribute from 200 response: stackV4Response/aws (in: body, type: Hash[string, *])
[36mswagger-diff_1        |[0m   - get /sdx/{}/detail
[36mswagger-diff_1        |[0m     - missing attribute from 200 response: stackV4Response/aws (in: body, type: Hash[string, *])```

You'll see that the stackV4Request and response don't match up because you added a new attribute for the map.

A bigger question:
Why are you making the cloudwatch creation a model property? Shouldn't it always be created? I don't think we should change the APIs. We should enable cloudwatch by default all the time. Thoughts @jamisonbennett ?

jamisonbennett:
> Why are you making the cloudwatch creation a model property? Shouldn't it always be created? I don't think we should change the APIs. We should enable cloudwatch by default all the time. 

@sxxgrc FreeIPA controlled it by a property config setting. You might want to consider doing that.

sxxgrc:
The scope of this has changed so I will reopen the PR when I am done with applying this to all AWS instances created by CB.

PR #9348, opened 2020-10-30
CB-9545 use JpaRepository getOne() instead of EntityManager

sodre90:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/eventbus/CloudResourcePersisterService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/eventbus/CloudResourcePersisterService.java
@@ -3,7 +3,6 @@ package com.sequenceiq.cloudbreak.service.eventbus;
 import java.util.Optional;
 
 import javax.inject.Inject;
-import javax.persistence.EntityManager;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


biharitomi (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/eventbus/CloudResourcePersisterService.java
```diff
@@ -31,7 +32,7 @@
     @Inject
     private ResourceService resourceService;
 
-    @Inject
+    @PersistenceContext
```
akanto, line 35: Please add a detailed commit message that why we need to change it, what problem is caused by the old solution. I just want to avoid that situation that a year from now somebody would accidentally/intentionally write it back.

Please also double-check whether we could eliminate the following lines from DatabaseConfig java files. 

   @Bean
    public EntityManager entityManager(EntityManagerFactory entityManagerFactory) {
        return entityManagerFactory.createEntityManager();
    }

https://github.com/hortonworks/cloudbreak/blob/fe089b5bf16558e4478f3c71165cc485d8cdc0e3/autoscale/src/main/java/com/sequenceiq/periscope/config/DatabaseConfig.java#L119

lnardai (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/eventbus/CloudResourcePersisterService.java
sodre90: Detailed commit message and java comment added. 
It seems the referred DatabaseConfig lines can be removed, applications are able to start without these lines.

akanto (approved):

akanto (changes requested):
```

 private Stack findStackById(Long stackId) {
        return entityManager.getReference(Stack.class, stackId);
    }
```

Final ask, can we completely eliminate direct access to  EntityManager, and just use Repository calls?

akanto (approved):

PR #9354, opened 2020-11-02
CB-8861. Enable diagnostics (CM & VM based) tests for AWS

oleewere:


oleewere:
note: this feature will be available from 2.33, so we should enable e2e tests

keyki (approved):

PR #9356, opened 2020-11-02
CB-9270 Upscale operation was not functioning

bergerdenes:
The setHostName method was introduced but was semantically incorrect:
it took the first group's first instance's hostname in any case.
As it turned out, setting the hostname on GCP's Instance object is not
necessary.
Manually tested repair and working fine without this.


cegganesh84:
@bergerdenes I have a different proposal that does not break FreeIPA creation https://github.com/hortonworks/cloudbreak/pull/9359

PR #9359, opened 2020-11-03
CB-9270 [GCP] Do not set hostnames for non-freeipa hosts while creati…

cegganesh84:
…ng the instances

1. If the hostname is not set at creation during FreeIPA bootstrapping they were reset by Google Cloud causing salt script failures that depend on `hostname -f`
2. But this need not be set for other host groups.
3. Also for FreeIPA do not hard code a single instance because that assumption breaks down in repair and HA scenarios.

./gradlew build

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.cloud.gcp.compute;
 
+import static com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts.CLOUD_STACK_TYPE_PARAMETER;
+import static com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts.FREEIPA_STACK_TYPE;
 import static com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil.getCustomNetworkId;
 import static com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil.getSharedProjectId;
 import static com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil.getSubnetId;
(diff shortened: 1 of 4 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -213,10 +217,15 @@ String getPublicKey(String groupPublicKey, String groupLoginUserName) {
         return publicKey;
     }
 
-    private String getHostname(CloudStack cloudStack, List<CloudResource> buildableResource) {
+    private String getHostname(CloudStack cloudStack, int privateId) {
```
doktoric, line 220: getHostName(Group group, int privateId) ?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -213,10 +217,15 @@ String getPublicKey(String groupPublicKey, String groupLoginUserName) {
...
         String hostname = null;
-        if (!cloudStack.getGroups().isEmpty() && !cloudStack.getGroups().get(0).getInstances().isEmpty()) {
-            hostname = cloudStack.getGroups().get(0).getInstances().get(0).getStringParameter(CloudInstance.DISCOVERY_NAME);
+        if (!cloudStack.getGroups().isEmpty() && !cloudStack.getGroups().get(0).getInstances().isEmpty()
+                && cloudStack.getGroups().get(0).getInstances().get(privateId) != null) {
+            hostname = cloudStack.getGroups().get(0).getInstances().get(privateId).getStringParameter(CloudInstance.DISCOVERY_NAME);
+            // Setting hostnames for only FreeIPA hosts
+            if (hostname == null || !hostname.startsWith("ipaserver")) {
```
doktoric, line 226: I think we should implement a cloudStack.type parameter which can get FREEIPA, STACK  values

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
cegganesh84: Nice suggestion! It simplified the code a lot. Thank you!

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
cegganesh84: This is a great suggestion. Could you please help me with how the stack name for freeipa got FreeIPA in it? Looks like that is the last place the information is dealt with as an Enum, beyond that they are blended as a string in names. I tried to track that yesterday, but got lost and implemented this simple solution.

cegganesh84: Hi @doktoric, I tried what seems simple and gets the job done with respect to stack type parameter suggestion. Please review.

doktoric (approved):

PR #9363, opened 2020-11-03
CB-9276: Enable audit events for SDX upgrade

tiborpopovics:


topolyai5 (changes requested):
could you please add some IT?

topolyai5 (approved):

PR #9367, opened 2020-11-03
CB-9598 use stack's accountid in HealthCheckAvailabilityChecker.java …

sodre90:
…instead of crnService.getCurrentAccountId() (which is altus in internal case)

Freeipa sync is running with internal user, so we can not use current accountId because it is 'altus' at this point.


keyki (approved):

akanto (changes requested):
We have already broken mow-dev. A test shall be implemented for this.

keyki:
Closing, as this should be fixed in 2.32.

PR #9371, opened 2020-11-03
CB-9598: Use stack's account ID in the HealthCheckAvailbilityChecker

jamisonbennett:
Fixed the account ID because FreeIpa sync is running with the internal
user, so the current account ID was altus.

Added a test to ensure the entitlement check does not use "altus" for
the account ID.

There is thread-local storage which sets the internal actor (account ID altus). It does this for the entire quartz health check. The availability checker uses the crn service to get the account ID. The CRN service reads the thread-local storage. Every flow and every API call sets the thread-local storage to the real user's account ID. Previously in 2.32, the HealthCheckAvailbiltyChecker was only called for flows/API calls so the real account ID was used. In 2.33 (302e984637775cac69e0e6b1b706341cade5fa22 tag: 2.33.0-b15) it started to be called from the quartz job which means it uses account ID altus. There was a little confusion with this since both check had been there in 2.32 but what I didn't realize is that the thread local storage would have different values depending on how it was called. This was not caught by the mock UMS since the mock UMS doesn't validate account IDs.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/util/HealthCheckAvailabilityChecker.java
@@ -17,13 +17,10 @@ public class HealthCheckAvailabilityChecker extends AvailabilityChecker {
     private static final Versioned CDP_FREEIPA_HEALTH_AGENT_AFTER_VERSION = () -> "2.31.0";
 
     @Inject
-    private CrnService crnService;
-
-    @Inject
     private EntitlementService entitlementService;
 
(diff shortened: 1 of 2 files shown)


jamisonbennett:
If you don't think the unittest is sufficient. I can add a hardcoded check to ensure the account ID is not "altus" in the mock UMS. I chose not to do this since it prevents someone from testing using that account id. Just let me know.

akanto (approved):

akanto:
I think the junit test is sufficient for this fix. @sodre90 promised to improve the Mock UMS and throw an Exception for Altus accounts. It will be tracked under different jira.

keyki (approved):

PR #9372, opened 2020-11-03
CB-9616: Have kinit code look for more HDFS node directories.

brycederriso:
Closes [CB-9616](https://jira.cloudera.com/browse/CB-9616).

In Medium duty datalakes we don't have an `hdfs-NAMENODE` directory, so trying to kinit fails.  It actually doesn't matter which keytab we use, so the regular expression to look for the directory can also include `JOURNALNODE` and `DATANODE`.

This is done in both the backup and the restore script.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -4,7 +4,6 @@
 # After PostgreSQL contents are dumped, the SQL file is uploaded using the hdfs cli.
 # For AWS or Azure, the url should start with `s3a://` or `abfs://`, respectively.
 
-set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
@@ -53,14 +52,30 @@ errorExit() {
   exit 1
 }
 
-kinit_as_hdfs() {
-  # Use a new Kerberos credential cache, to keep from clobbering the default.
-  export KRB5CCNAME=/tmp/krb5cc_cloudbreak_$EUID
-  HDFS_KEYTAB=/run/cloudera-scm-agent/process/$(ls -t /run/cloudera-scm-agent/process/ | grep hdfs-NAMENODE$ | head -n 1)/hdfs.keytab
-  doLog "kinit as hdfs using Keytab: $HDFS_KEYTAB"
-  kinit -kt "$HDFS_KEYTAB" "hdfs/$(hostname -f)"
+kinit_as() {
+  doLog "attempting kinit as $1 using Keytab: $2"
+  kinit -kt "$2" "$1/$(hostname -f)"
   if [[ $? -ne 0 ]]; then
-    errorExit "Couldn't kinit as hdfs"
+    doLog "Couldn't kinit as $1"
+    return 1
+  fi
+}
+
+run_kinit() {
+  export KRB5CCNAME=/tmp/krb5cc_cloudbreak_$EUID
+
+  HDFS_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*hdfs*" | head -n 1)
+  HBASE_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -a \( -path "*hbase-REGIONSERVER*" -o -path "*hbase-MASTER*" \) -a -type f | head -n 1)
+  SOLR_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*solr-SOLR_SERVER*" | head -n 1)
+
+  if kinit_as hdfs "$HDFS_KEYTAB"; then
+    doLog "Successful kinit using hdfs principal"
+  elif kinit_as hbase "$HBASE_KEYTAB"; then
+    doLog "Successful kinit using hbase principal"
+  elif kinit_as solr "$SOLR_KEYTAB"; then
+    doLog "Successful kinit using solr principal"
+  else
+    errorExit "Couldn't get kerberos ticket to access cloud storage."
   fi
 }
 
@@ -77,7 +92,7 @@ replace_ranger_group_before_export() {
 
 move_backup_to_cloud () {
   LOCAL_BACKUP="$1"
-  kinit_as_hdfs
+  run_kinit
   doLog "INFO Uploading to ${BACKUP_LOCATION}"
 
   hdfs dfs -mkdir -p "$BACKUP_LOCATION"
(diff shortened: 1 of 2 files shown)


brycederriso:
@kkalvagadda1 This is ready to be looked at when you get a chance.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -56,7 +56,7 @@ errorExit() {
 kinit_as_hdfs() {
   # Use a new Kerberos credential cache, to keep from clobbering the default.
   export KRB5CCNAME=/tmp/krb5cc_cloudbreak_$EUID
-  HDFS_KEYTAB=/run/cloudera-scm-agent/process/$(ls -t /run/cloudera-scm-agent/process/ | grep hdfs-NAMENODE$ | head -n 1)/hdfs.keytab
+  HDFS_KEYTAB=/run/cloudera-scm-agent/process/$(ls -t /run/cloudera-scm-agent/process/ | grep "hdfs-\(JOURNALNODE\|DATANODE\|NAMENODE\)$" | head -n 1)/hdfs.keytab
```
kkalvagadda1, line 59: This will work as long one of the service listed below are running in the gateway node where salt-master is running.

1. JOURNALNODE
2. DATANODE
3. NAMENODE

Extend this to use keytab's for Solr and Hbase services as well if the HDFS keytab's are not found.

Later we could update the logic in CB to run this command on a host where Solr/Hbase services are running instead of always picking the gatewayFQDN node.

kkalvagadda1 (changes requested):
Please add code changes to capture the failure when the keytab is not found.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
brycederriso: Changed this to use `find` and look for hbase and solr keytabs.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -3,7 +3,6 @@
 # This script uses the 'psql' cli to drop hive and ranger databases, create them, then read in a plain SQL file to restore data.
 # We retrieve the SQL files from a valid URL, which should be `s3a://` or `abfs://` for AWS or Azure clouds respectively.
 
-set -o errexit
```
kkalvagadda1, line 6: Was this the reason why the error was not logged?

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -56,9 +55,15 @@ errorExit() {
...
-  doLog "kinit as hdfs using Keytab: $HDFS_KEYTAB"
-  kinit -kt "$HDFS_KEYTAB" "hdfs/$(hostname -f)"
+
+  KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*hdfs*" -o \
+   -path "*hbase-REGIONSERVER" -o \
+   -path "*hbase-MASTER" -o \
+   -path "*solr-SOLRSERVER" \
+   | head -n 1)
+
+  doLog "kinit as hdfs using Keytab: $KEYTAB"
```
kkalvagadda1, line 65: With the above change, you might use the key tab for hdfs/hbase/solr services but when you perform kinit you are always using hdfs as the principal name.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -57,9 +56,15 @@ errorExit() {
...
-  kinit -kt "$HDFS_KEYTAB" "hdfs/$(hostname -f)"
+
+  KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*hdfs*" -o \
+   -path "*hbase-REGIONSERVER" -o \
+   -path "*hbase-MASTER" -o \
+   -path "*solr-SOLRSERVER" \
+   | head -n 1)
+
+  doLog "kinit as hdfs using Keytab: $KEYTAB"
+  kinit -kt "$KEYTAB" "hdfs/$(hostname -f)"
```
kkalvagadda1, line 67: With the above change, you might use the key tab for hdfs/hbase/solr services but when you perform kinit you are always using hdfs as the principal name.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -57,9 +56,15 @@ errorExit() {
...
-  doLog "kinit as hdfs using Keytab: $HDFS_KEYTAB"
-  kinit -kt "$HDFS_KEYTAB" "hdfs/$(hostname -f)"
+
+  KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*hdfs*" -o \
+   -path "*hbase-REGIONSERVER" -o \
+   -path "*hbase-MASTER" -o \
+   -path "*solr-SOLRSERVER" \
+   | head -n 1)
+
+  doLog "kinit as hdfs using Keytab: $KEYTAB"
```
kkalvagadda1, line 66:
```suggestion
  doLog "kinit using Keytab: $KEYTAB"
```

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
kkalvagadda1, line 65:
```suggestion
  doLog "kinit using Keytab: $KEYTAB"
```

kkalvagadda1 (changes requested):

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
brycederriso: Yes.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
brycederriso: Thank you! This is definitely a bug. I've added handling for kiniting as different principals.

brycederriso: done

kkalvagadda1 (commented):
Looks good. Did you test the cases where HDFS_KEYTAB/HBASE_KEYTAB/SOLR_KEYTAB are empty?

brycederriso:
@kkalvagadda1 I tested when they were missing, and that it fell through to the later else blocks.

kkalvagadda1 (approved):

PR #9373, opened 2020-11-04
CB-9451: Move deployment of RAZ to the master node for better supportability with medium duty datalake setups

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-9451

Basically just made it so RAZ should be deploying on master nodes instead of CORE nodes, this is subject to change in the future as we modify the medium duty shape.

I tested this by running the core cloudbreak service locally and setting up both a light duty and medium duty Azure RAZ datalake, ensuring that the code change was hit by placing a breakpoint on it. Both of these were set up successfully.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazBaseConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazBaseConfigProvider.java
@@ -15,7 +15,6 @@ import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateProcessor;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.AbstractRoleConfigProvider;
 import com.sequenceiq.cloudbreak.template.TemplatePreparationObject;
 import com.sequenceiq.cloudbreak.template.views.HostgroupView;
-import com.sequenceiq.common.api.type.InstanceGroupType;
 
 public abstract class RangerRazBaseConfigProvider extends AbstractRoleConfigProvider {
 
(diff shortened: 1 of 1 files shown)


orngejaket (approved):

kmanamcheri:
@sxxgrc in your testing, did you verify by going into CM UI that two instances of Raz is installed on the master nodes correctly?

sxxgrc:
No I did not, I did not know that I could verify it that way. I will do that today and let you know.

sxxgrc:
Not sure if I can really do this cause the way I am testing this is kind of complicated.
Specifically I am setting up the environment and datalake using the CDP CLI instead of the UI because, for whatever reason, my local UI does not have the option for setting up RAZ. 
As I used the CDP CLI, Knox is not bypassed, and I do not have any way of logging in through Knox to access the CM UI.

Please let me know if there are ways to circumvent Knox or better ways to do what I am trying to do.

kmanamcheri:
> Not sure if I can really do this cause the way I am testing this is kind of complicated.
> Specifically I am setting up the environment and datalake using the CDP CLI instead of the UI because, for whatever reason, my local UI does not have the option for setting up RAZ.
> As I used the CDP CLI, Knox is not bypassed, and I do not have any way of logging in through Knox to access the CM UI.
> 
> Please let me know if there are ways to circumvent Knox or better ways to do what I am trying to do.

You can ssh into the master nodes and confirm that Raz server process is running.

sxxgrc:
> You can ssh into the master nodes and confirm that Raz server process is running.

Not sure if I'm doing it right but I SSHd into the master node and then ran ps aux and found some processes that belonged to the user "rangerraz". I assume that works? I only did this for light duty I can check the other medium duty setup I have.

sxxgrc:
Okay yeah, I checked every single node for both light duty and medium duty, and only the master nodes had any processes being run by the "rangerraz" user.
I confirmed that both of the master nodes for HA had rangerraz processes too.

kmanamcheri (approved):

PR #9374, opened 2020-11-04
CB-8962: Added a Cloudwatch alarm to each AWS instance created.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-8962

We specifically wanted to extend the FreeIPA Cloudwatch alarm feature to all AWS nodes created. In order to do this I simply rolled back the specific FreeIPA logic that added a Cloudwatch parameter to the Cloudstacks created for the instances, and removed the check for that parameter in the Cloudwatch alarm parameter.

As far as I know, removing this check will simply cause all AWS instances created by cloudbreak to create the Cloudwatch alarm by default. Please let me know if I am forgetting or overlooking something that could possibly break by removing this check.

All of the logic for automatically removing the Cloudwatch alarm when the instances are terminated, or on some failure, already exists from the FreeIPA changes made before and is still in use.

I tested this by setting up a light duty and medium duty environment/datalake locally, running both the core cloudbreak service and the FreeIPA service locally. I ensured that both cases created the same alarm for all instances as shown in the images below:

* [All Nodes with Alarms](https://drive.google.com/file/d/16xR4eAPjHraymt-SynUI9pdjT2fRCCsH/view?usp=sharing)
* [Alarm per Node](https://drive.google.com/file/d/1VXLQzEq66qgDyYwQ3Yfie6CAf3rdy7W5/view?usp=sharing)

I am not really sure why the alarm count went down from 2 to 1 (previously each node had 2 of the same alarms), but we can discuss this and why it is happening after the unplug days.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -1,7 +1,5 @@
 package com.sequenceiq.cloudbreak.cloud.aws.connector.resource;
 
-import static com.sequenceiq.cloudbreak.cloud.PlatformParametersConsts.CLOUDWATCH_CREATE_PARAMETER;
-
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -34,47 +32,46 @@ public class AwsCloudWatchService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsCloudWatchService.class);
 
-    @Value("${freeipa.aws.cloudwatch.suffix:-Status-Check-Failed-System}")
+    @Value("${aws.cloudwatch.suffix:-Status-Check-Failed-System}")
     private String alarmSuffix;
 
-    @Value("${freeipa.aws.cloudwatch.period:60}")
+    @Value("${aws.cloudwatch.period:60}")
     private int cloudwatchPeriod;
 
-    @Value("${freeipa.aws.cloudwatch.evaluationPeriods:2}")
+    @Value("${aws.cloudwatch.evaluationPeriods:2}")
     private int cloudwatchEvaluationPeriods;
 
-    @Value("${freeipa.aws.cloudwatch.threshold:1.0}")
+    @Value("${aws.cloudwatch.threshold:1.0}")
     private double cloudwatchThreshhold;
 
-    @Value("${freeipa.aws.cloudwatch.max-batchsize:100}")
+    @Value("${aws.cloudwatch.max-batchsize:100}")
     private int maxBatchsize;
 
     @Inject
     private AwsClient awsClient;
 
-    public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances, CloudStack stack, String regionName, AwsCredentialView credentialView) {
-        if (isCloudwatchEnabled(stack)) {
-            instances.stream().forEach(instance -> {
-                try {
-                    PutMetricAlarmRequest metricAlarmRequest = new PutMetricAlarmRequest();
-                    metricAlarmRequest.setAlarmActions(Arrays.asList("arn:aws:automate:" + regionName + ":ec2:recover"));
-                    metricAlarmRequest.setAlarmName(instance.getInstanceId() + alarmSuffix);
-                    metricAlarmRequest.setMetricName("StatusCheckFailed_System");
-                    metricAlarmRequest.setStatistic("Maximum");
-                    metricAlarmRequest.setNamespace("AWS/EC2");
-                    metricAlarmRequest.setDimensions(Arrays.asList(new Dimension().withName("InstanceId").withValue(instance.getInstanceId())));
-                    metricAlarmRequest.setPeriod(cloudwatchPeriod);
-                    metricAlarmRequest.setEvaluationPeriods(cloudwatchEvaluationPeriods);
-                    metricAlarmRequest.setThreshold(cloudwatchThreshhold);
-                    metricAlarmRequest.setComparisonOperator("GreaterThanOrEqualToThreshold");
-                    AmazonCloudWatchClient amazonCloudWatchClient = awsClient.createCloudWatchClient(credentialView, regionName);
-                    amazonCloudWatchClient.putMetricAlarm(metricAlarmRequest);
-                    LOGGER.debug("Created cloudwatch alarm for instanceId {}.", instance.getInstanceId());
-                } catch (AmazonCloudWatchException acwe) {
-                    LOGGER.error("Unable to create cloudwatch alarm for instanceId {}: {}", instance.getInstanceId(), acwe.getLocalizedMessage());
-                }
-            });
-        }
+    public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances, String regionName, AwsCredentialView credentialView) {
+        AmazonCloudWatchClient amazonCloudWatchClient = awsClient.createCloudWatchClient(credentialView, regionName);
+
+        instances.stream().forEach(instance -> {
+            try {
+                PutMetricAlarmRequest metricAlarmRequest = new PutMetricAlarmRequest();
+                metricAlarmRequest.setAlarmActions(Arrays.asList("arn:aws:automate:" + regionName + ":ec2:recover"));
+                metricAlarmRequest.setAlarmName(instance.getInstanceId() + alarmSuffix);
+                metricAlarmRequest.setMetricName("StatusCheckFailed_System");
+                metricAlarmRequest.setStatistic("Maximum");
+                metricAlarmRequest.setNamespace("AWS/EC2");
+                metricAlarmRequest.setDimensions(Arrays.asList(new Dimension().withName("InstanceId").withValue(instance.getInstanceId())));
+                metricAlarmRequest.setPeriod(cloudwatchPeriod);
+                metricAlarmRequest.setEvaluationPeriods(cloudwatchEvaluationPeriods);
+                metricAlarmRequest.setThreshold(cloudwatchThreshhold);
+                metricAlarmRequest.setComparisonOperator("GreaterThanOrEqualToThreshold");
+                amazonCloudWatchClient.putMetricAlarm(metricAlarmRequest);
+                LOGGER.debug("Created cloudwatch alarm for instanceId {}.", instance.getInstanceId());
+            } catch (AmazonCloudWatchException acwe) {
+                LOGGER.error("Unable to create cloudwatch alarm for instanceId {}: {}", instance.getInstanceId(), acwe.getLocalizedMessage());
+            }
+        });
     }
 
     public void deleteCloudWatchAlarmsForSystemFailures(CloudStack stack, String regionName, AwsCredentialView credentialView) {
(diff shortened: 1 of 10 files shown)


keyki:
Have we checked what are the permissions of the AWS credential we require? When we enabled the same for FreeIPA we had lots of issues people not able to delete the clusters.

kmanamcheri:
@sxxgrc Looks like you have to rebase. Please rebase and update the review.

sxxgrc:
Rebased, still have some errors though. Trying to look into them at the moment.

sxxgrc:
Okay the errors I got were due to the Launch and Repair tests running into null pointer exceptions due to them using fake region names and fake credentials. The Cloudwatch Service class grabs a CloudWatch client which requires a valid region and credential for that to work.
I decided to add a quick null check in the Cloudwatch service class as I did not want to change the way the test AWS values are set up.

sxxgrc:
> Have we checked what are the permissions of the AWS credential we require? When we enabled the same for FreeIPA we had lots of issues people not able to delete the clusters.

I'm not exactly sure the answer to this. From what I can see, the few places in the AwsCloudWatchService class where a credential is mentioned, is for the creation of the AmazonCloudwatchClient object, however this seems to just be for AWS credentials (i.e. access key and secret key), which are just passed around through the AWS API. Not sure if that is what was the issue last time.

I'm wondering if I can test this at all locally. I was able to delete the clusters just fine.

sxxgrc:
Fixed the test issues, please feel free to look over it and review what I changed. Let me know if you have any questions or if there is something I should be doing differently.

sxxgrc:
@keyki I talked to Jamison today, who had previously worked on the alarm changes for FreeIPA, and he let me know that there is a JSON policy which customers are expected to provide for us which enables creation, deletion, and describing of the Cloudwatch alarms. So I assume this is generally good and has been set up correctly back then.

jamisonbennett:
@sxxgrc we provide the JSON policy to the customer. See https://github.com/hortonworks/cloudbreak/commit/f39bd44d4f263cfae3c6bbb59bb99a2899f67b43 and https://github.com/hortonworks/cloudbreak/commit/ab618c348493edf68b0f66ad2e1bda152cbca553. It looks like put & delete rights were added at the same time, then list was added.

sxxgrc:
Perfect, thank you for the information!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -34,47 +32,48 @@
...
-                    AmazonCloudWatchClient amazonCloudWatchClient = awsClient.createCloudWatchClient(credentialView, regionName);
-                    amazonCloudWatchClient.putMetricAlarm(metricAlarmRequest);
-                    LOGGER.debug("Created cloudwatch alarm for instanceId {}.", instance.getInstanceId());
-                } catch (AmazonCloudWatchException acwe) {
-                    LOGGER.error("Unable to create cloudwatch alarm for instanceId {}: {}", instance.getInstanceId(), acwe.getLocalizedMessage());
-                }
-            });
-        }
+    public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances, String regionName, AwsCredentialView credentialView) {
+        // Grab the CloudWatch client.
```
lacikaaa, line 54: please remove these comments, they don't have any extra information

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -34,47 +32,46 @@
...
+                metricAlarmRequest.setNamespace("AWS/EC2");
+                metricAlarmRequest.setDimensions(Arrays.asList(new Dimension().withName("InstanceId").withValue(instance.getInstanceId())));
+                metricAlarmRequest.setPeriod(cloudwatchPeriod);
+                metricAlarmRequest.setEvaluationPeriods(cloudwatchEvaluationPeriods);
+                metricAlarmRequest.setThreshold(cloudwatchThreshhold);
+                metricAlarmRequest.setComparisonOperator("GreaterThanOrEqualToThreshold");
+                amazonCloudWatchClient.putMetricAlarm(metricAlarmRequest);
+                LOGGER.debug("Created cloudwatch alarm for instanceId {}.", instance.getInstanceId());
+            } catch (AmazonCloudWatchException acwe) {
+                LOGGER.error("Unable to create cloudwatch alarm for instanceId {}: {}", instance.getInstanceId(), acwe.getLocalizedMessage());
```
doktoric, line 72: I guess we should log the error message somewhere here

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -34,47 +32,46 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsCloudWatchService.class);
 
-    @Value("${freeipa.aws.cloudwatch.suffix:-Status-Check-Failed-System}")
+    @Value("${aws.cloudwatch.suffix:-Status-Check-Failed-System}")
```
doktoric, line 35: do we need any helm chart change because of the renamed property?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -86,20 +83,18 @@ public void deleteCloudWatchAlarmsForSystemFailures(CloudStack stack, String reg
...
-                    .collect(Collectors.toList());
-            List<String> instanceIdsNotInStack = instanceIds.stream()
-                    .filter(instanceId -> !instanceIdsFromStack.contains(instanceId))
-                    .collect(Collectors.toList());
-            if (!instanceIdsNotInStack.isEmpty()) {
-                LOGGER.warn("Instance IDs [{}] are not part of cloud stack {}, these instances may have already been deleted on the cloud provider side.",
-                        instanceIdsFromStack, stack);
-            }
-            deleteCloudWatchAlarmsForSystemFailures(regionName, credentialView, instanceIds);
+        List<String> instanceIdsFromStack = stack.getGroups().stream()
```
doktoric, line 86: I think instancemetadata also contains this information

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -116,16 +111,17 @@ private void deleteCloudWatchAlarmsForSystemFailures(String regionName, AwsCrede
...
             DescribeAlarmsRequest request = new DescribeAlarmsRequest().withAlarmNames(alarmNames).withMaxRecords(maxBatchsize);
-            AmazonCloudWatchClient amazonCloudWatchClient = awsClient.createCloudWatchClient(credentialView, regionName);
             List<String> filteredAlarmNames = amazonCloudWatchClient.describeAlarms(request).getMetricAlarms().stream()
                     .map(MetricAlarm::getAlarmName)
                     .collect(Collectors.toList());
             filteredAlarmNamesStream = Stream.of(filteredAlarmNames);
             LOGGER.debug("Checking cloudwatch alarms [{}] for existence and found [{}]", alarmNames, filteredAlarmNames);
         } catch (AmazonCloudWatchException acwe) {
-            LOGGER.error("Unable to describe cloudwatch alarms falling back to delete all alarms indivdually [{}]: {}", alarmNames, acwe.getLocalizedMessage());
+            LOGGER.error("Unable to describe cloudwatch alarms falling back to delete all alarms individually [{}]: {}", alarmNames, acwe.getLocalizedMessage());
```
doktoric, line 124: I guess we should log the error somewhere here

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/component/AwsLaunchTest.java
```diff
@@ -156,6 +157,9 @@
     @MockBean
     private CustomAmazonWaiterProvider customAmazonWaiterProvider;
 
+    @MockBean
```
doktoric, line 160: this is a mock bean here but there is no assertion or verification on this

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -34,47 +32,46 @@
...
+        AmazonCloudWatchClient amazonCloudWatchClient = awsClient.createCloudWatchClient(credentialView, regionName);
+
+        instances.stream().forEach(instance -> {
+            try {
+                PutMetricAlarmRequest metricAlarmRequest = new PutMetricAlarmRequest();
+                metricAlarmRequest.setAlarmActions(Arrays.asList("arn:aws:automate:" + regionName + ":ec2:recover"));
+                metricAlarmRequest.setAlarmName(instance.getInstanceId() + alarmSuffix);
+                metricAlarmRequest.setMetricName("StatusCheckFailed_System");
+                metricAlarmRequest.setStatistic("Maximum");
+                metricAlarmRequest.setNamespace("AWS/EC2");
```
doktoric, line 63: is it possible to create multiple cloudwatch alarms with one request to aws?

sxxgrc: This does not seem to be exposed in the API unfortunatelyt.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
sxxgrc: Not really sure what the helm chart even is, hopefully someone else can answer this. These properties are not being injected from anywhere though as they are just being set in place so I'm not sure why anything fancy would need to be done with them.

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/component/AwsLaunchTest.java
sxxgrc: Changed things up to mock in a way that makes more sense with what's going on and the original error.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
doktoric: nevere mind I see it

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
doktoric: yes.

sxxgrc: Okay sorry I am just now really looking into this, but I don't see an easy way to grab the instance metadata from the inputs we have as that doesn't seem to be exposed through the CloudStack object.
Also what benefit do you see from getting that information through there instead?

It appears to me that the original purpose for using the instances directly from the CloudStack was to be able to see which instances were still lively vs already deleted, and then only delete the alarms from the lively ones. Does using the metadata also enable this, and if so does it make this process better?

doktoric: yeah I just checked this is not a trivial task

doktoric (approved):

PR #9375, opened 2020-11-04
CB-9614 [thunderhead-mock] Throw exception if ums getAccount called with 'altus'

horadla23:
…instead of crnService.getCurrentAccountId() (which is altus in internal case)

Freeipa sync is running with internal user, so we can not use current accountId because it is 'altus' at this point.

Also this commit will ensure that mock ums getAccount call fails if we are calling it with 'altus' anywhere. This should prevent further issues with 'altus' accountId during getAccount call.


keyki (approved):
I hope no one is using altus as tenant :)

keyki:
I don't think we want to hotfix 2.32 with this change. Can you please move the PR to master? On 2.33 it is also fixed.

horadla23:
> I don't think we want to hotfix 2.32 with this change. Can you please move the PR to master? On 2.33 it is also fixed.

sure

sodre90 (approved):

PR #9385, opened 2020-11-09
CB-7552: IDBroker VM is provisioned with an attached disk that remain…

LinaAtAustin:
This is the initial version of the code change.
a. I have checked on long running environment that the data volume of the IDBroker is not used. Details are in CB-7552
b. I only tested on AWS, not on Azure yet. Waiting on getting access to Azure.

When I tried to remove the data volume at IDBroker by setting the volume count to be 0 or the size to be 0GB, the deployment failed. So I set the data volume at IDBroker to be 1 GB, which is the min volume size allowed at https://github.com/hortonworks/cloudbreak/blob/master/cloud-aws/src/main/resources/definitions/aws-vm.json. The validation code that enforces this is at https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L168 and https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L184

Should I change the min volume size to be 0 GB in aws-vm.json? That may cause validation useless for other scenarios. What is your suggestion?

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
@@ -29,6 +29,7 @@ import com.sequenceiq.cloudbreak.converter.spi.CredentialToExtendedCloudCredenti
 import com.sequenceiq.cloudbreak.domain.Template;
 import com.sequenceiq.cloudbreak.domain.VolumeTemplate;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
 import com.sequenceiq.cloudbreak.dto.credential.Credential;
 import com.sequenceiq.cloudbreak.validation.ValidationResult;
 import com.sequenceiq.cloudbreak.workspace.model.User;
@@ -36,6 +37,7 @@ import com.sequenceiq.common.api.type.CdpResourceType;
 
 @Component
 public class TemplateValidator {
+    public static final String GROUP_NAME_ID_BROKER = "idbroker";
 
     @Inject
     private CloudParameterService cloudParameterService;
@@ -52,9 +54,9 @@ public class TemplateValidator {
     private final Supplier<Map<Platform, PlatformParameters>> platformParameters =
             Suppliers.memoize(() -> cloudParameterService.getPlatformParameters());
 
-    public void validate(Credential credential, Template value, Stack stack,
+    public void validate(Credential credential, InstanceGroup instanceGroup, Stack stack,
         CdpResourceType stackType, Optional<User> user, ValidationResult.ValidationResultBuilder validationBuilder) {
-
+        Template value = instanceGroup.getTemplate();
         CloudVmTypes cloudVmTypes = cloudParameterService.getVmTypesV2(
                 extendedCloudCredentialConverter.convert(credential, user),
                 stack.getRegion(),
@@ -85,13 +87,13 @@ public class TemplateValidator {
                 }
             }
 
-            validateVolumeTemplates(value, vmType, platform, validationBuilder);
+            validateVolumeTemplates(value, vmType, platform, validationBuilder, instanceGroup);
             validateMaximumVolumeSize(value, vmType, validationBuilder);
         }
     }
 
     private void validateVolumeTemplates(Template value, VmType vmType, Platform platform,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         for (VolumeTemplate volumeTemplate : value.getVolumeTemplates()) {
             VolumeParameterType volumeParameterType = null;
             Map<Platform, Map<String, VolumeParameterType>> disks = diskMappings.get();
@@ -104,7 +106,7 @@ public class TemplateValidator {
                 }
             }
 
-            validateVolume(volumeTemplate, vmType, platform, volumeParameterType, validationBuilder);
+            validateVolume(volumeTemplate, vmType, platform, volumeParameterType, validationBuilder, instanceGroup);
         }
     }
 
@@ -127,10 +129,10 @@ public class TemplateValidator {
     }
 
     private void validateVolume(VolumeTemplate value, VmType vmType, Platform platform,
-        VolumeParameterType volumeParameterType, ValidationResult.ValidationResultBuilder validationBuilder) {
+        VolumeParameterType volumeParameterType, ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         validateVolumeType(value, platform, validationBuilder);
-        validateVolumeCount(value, vmType, volumeParameterType, validationBuilder);
-        validateVolumeSize(value, vmType, volumeParameterType, validationBuilder);
+        validateVolumeCount(value, vmType, volumeParameterType, validationBuilder, instanceGroup);
+        validateVolumeSize(value, vmType, volumeParameterType, validationBuilder, instanceGroup);
     }
 
     private void validateMaximumVolumeSize(Template value, VmType vmType, ValidationResult.ValidationResultBuilder validationBuilder) {
@@ -157,14 +159,20 @@ public class TemplateValidator {
         }
     }
 
+    private boolean isIDBrokerInstanceGroup(InstanceGroup instanceGroup) {
+        return GROUP_NAME_ID_BROKER.equalsIgnoreCase(instanceGroup.getGroupName());
+    }
+
     private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
+                // IDBroker does not use data volume, so its volume count should be zero
+                // To be backward-compatible, we only check max limit and allow the min limit to be zero for IDBroker
                 if (value.getVolumeCount() > config.maximumNumber()) {
                     validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
-                } else if (value.getVolumeCount() < config.minimumNumber()) {
+                } else if (!isIDBrokerInstanceGroup(instanceGroup) && value.getVolumeCount() < config.minimumNumber()) {
                     validationBuilder.error(String.format("Min allowed volume count for '%s': %s", vmType.value(), config.minimumNumber()));
                 }
             } else {
@@ -174,13 +182,15 @@ public class TemplateValidator {
     }
 
     private void validateVolumeSize(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
+                // IDBroker does not use data volume, so its volume size should be zero
+                // To be backward-compatible, we only check max limit and allow the min limit to be zero for IDBroker
                 if (value.getVolumeSize() > config.maximumSize()) {
                     validationBuilder.error(String.format("Max allowed volume size for '%s': %s", vmType.value(), config.maximumSize()));
-                } else if (value.getVolumeSize() < config.minimumSize()) {
+                } else if (!isIDBrokerInstanceGroup(instanceGroup) && value.getVolumeSize() < config.minimumSize()) {
                     validationBuilder.error(String.format("Min allowed volume size for '%s': %s", vmType.value(), config.minimumSize()));
                 }
             } else {
--- a/core/src/test/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidatorTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidatorTest.java
@@ -1,0 +1,216 @@
+package com.sequenceiq.cloudbreak.controller.validation.template;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.when;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import com.google.common.collect.Sets;
+import com.sequenceiq.cloudbreak.TestUtil;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status;
+import com.sequenceiq.cloudbreak.cloud.model.CloudVmTypes;
+import com.sequenceiq.cloudbreak.cloud.model.Platform;
+import com.sequenceiq.cloudbreak.cloud.model.PlatformDisks;
+import com.sequenceiq.cloudbreak.cloud.model.VmType;
+import com.sequenceiq.cloudbreak.cloud.model.VmTypeMeta;
+import com.sequenceiq.cloudbreak.cloud.model.VolumeParameterType;
+import com.sequenceiq.cloudbreak.cloud.service.CloudParameterService;
+import com.sequenceiq.cloudbreak.controller.validation.LocationService;
+import com.sequenceiq.cloudbreak.converter.spi.CredentialToExtendedCloudCredentialConverter;
+import com.sequenceiq.cloudbreak.domain.Template;
+import com.sequenceiq.cloudbreak.domain.VolumeTemplate;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.stack.cluster.Cluster;
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
+import com.sequenceiq.cloudbreak.dto.credential.Credential;
+import com.sequenceiq.cloudbreak.validation.ValidationResult;
+import com.sequenceiq.cloudbreak.workspace.model.User;
+import com.sequenceiq.common.api.type.CdpResourceType;
+import com.sequenceiq.common.api.type.InstanceGroupType;
+
+@ExtendWith(MockitoExtension.class)
+public class TemplateValidatorTest {
+
+    private Credential credential;
+
+    private User user;
+
+    private Optional<User> optionalUser;
+
+    private CloudVmTypes cloudVmTypes;
+
+    private PlatformDisks platformDisks;
+
+    @Mock
+    private ValidationResult.ValidationResultBuilder builder;
+
+    @Mock
+    private CredentialToExtendedCloudCredentialConverter extendedCloudCredentialConverter;
+
+    @Mock
+    private CloudParameterService cloudParameterService;
+
+    @Mock
+    private LocationService locationService;
+
+    @InjectMocks
+    private InstanceGroup instanceGroup;
+
+    private Stack stack;
+
+    @InjectMocks
+    private TemplateValidator underTest = new TemplateValidator();
+
+    @BeforeEach
+    public void setUp() {
+        credential = TestUtil.awsCredential();
+        user = new User();
+        optionalUser = Optional.of(user);
+        MockitoAnnotations.initMocks(this);
+        stack = TestUtil.stack(Status.AVAILABLE, credential);
+        Cluster cluster = TestUtil.cluster();
+        stack.setCluster(cluster);
+
+        String location = "fake location";
+        VmTypeMeta.VmTypeMetaBuilder vmMetaBuilder = VmTypeMeta.VmTypeMetaBuilder.builder()
+                .withCpuAndMemory(Integer.valueOf(8), Float.valueOf(15))
+                .withPrice(0.42)
+                .withVolumeEncryptionSupport(true)
+                .withSsdConfig(1, 17592,
+                        1, 24);
+        VmType c3VmType = VmType.vmTypeWithMeta("c3.2xlarge", vmMetaBuilder.create(), false);
+        Map<String, Set<VmType>> machines = new HashMap<>();
+        machines.put(location, Collections.singleton(c3VmType));
+        cloudVmTypes = new CloudVmTypes(machines, new HashMap<>());
+        when(cloudParameterService.getVmTypesV2(
+                isNull(), anyString(), isNull(), any(CdpResourceType.class), any(HashMap.class))).thenReturn(cloudVmTypes);
+
+        Platform platform = Platform.platform("AWS");
+        Map<Platform, Map<String, VolumeParameterType>> diskMappings = new HashMap<>();
+        Map<String, VolumeParameterType> diskTypeMap = new HashMap<>();
+        diskTypeMap.put("standard", VolumeParameterType.SSD);
+        diskMappings.put(platform, diskTypeMap);
+
+        platformDisks = new PlatformDisks(new HashMap<>(), new HashMap<>(), diskMappings, new HashMap<>());
+        when(cloudParameterService.getDiskTypes()).thenReturn(platformDisks);
+
+        when(locationService.location(anyString(), isNull())).thenReturn(location);
+    }
+
+    @Test
+    public void validateIDBrokerDataVolumeZeroCountZeroSize() {
+        instanceGroup = createInstanceGroup(0, 0, true);
+        underTest.validate(credential, instanceGroup, stack, CdpResourceType.DATALAKE, optionalUser, builder);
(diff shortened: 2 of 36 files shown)


datalake/src/main/resources/duties/7.2.7/aws/light_duty.json
```diff
@@ -15,7 +15,7 @@
         "attachedVolumes": [
           {
             "count": 1,
-            "size": 100,
+            "size": 1,
```
keyki, line 18: This doesn't seem right. 1 GB disk?

keyki:
We should change the validation if it doesn't allow volume count as 0, but we shouldn't set the disk size to 1 GB.

doktoric:
Also we should check every configprovider how they can handle this like https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/solr/SolrVolumeConfigProvider.java#L25

lajosrodek:
@doktoric I generally agree but the change here supposedly only affects the IDBroker node. That hostgroup has a single attached role (`KNOX` / `IDBROKER`), so none of the present VolumeConfigProviders shall be broken. 😃

LinaAtAustin:
@keyki "We should change the validation if it doesn't allow volume count as 0, but we shouldn't set the disk size to 1 GB."
Right now, the min volume size is based on EC2 type, not depend on the role. That is why if I change the min volume size to be 0 GB for all roles, it may break other scenarios. 

Do you want to add the support to configure min volume count based on both EC2 type and role type? It would be a much bigger change.

Another question:

Why IDBroker has to be on its own host? If IDBroker is on the same host as the other roles on the master node, then we don't even need this change. 
1) If we can move IDBroker to the master node, we can remove the IDBroker node.
2) If we cannot move IDBroker to the master node due to resource limitation (such as CPU and memory), we can move other roles to that IDBroker node and reduce the size of master node to reduce the overall cost. If this approach works, we don't need to remove the data volume on the IDBroker node since other roles need to store data there.

keyki:
IDBroker is separated on purpose. There is an attached instance profile with which you can assume roles. The SSH is also disabled for this node with workload users. Otherwise, anyone would be able to assume roles if they SSH. It cannot live on any other instances that are used by users. 

I'm not sure I understand the volume count question. Having 0 attached volumes cannot depend on the instance type.

LinaAtAustin:
@keyki the min volume size allowed at https://github.com/hortonworks/cloudbreak/blob/master/cloud-aws/src/main/resources/definitions/aws-vm.json, which defines the min volume count based on EC2 type. 

The validation code that enforces this is at https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L168 and https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L184

keyki:
> @keyki the min volume size allowed at https://github.com/hortonworks/cloudbreak/blob/master/cloud-aws/src/main/resources/definitions/aws-vm.json, which defines the min volume count based on EC2 type.
> 
> The validation code that enforces this is at https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L168 and https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java#L184

We could just change the validation to make an exception in case it's an idbroker node?

kmanamcheri:
> We should change the validation if it doesn't allow volume count as 0, but we shouldn't set the disk size to 1 GB.

+1 It seems like an odd pattern to set a 1GB disk for no reason. We should fix the validation logic and the code to get the IDBroker nodes to a state which we want it to be in -- one attached disk of 50GB.

LinaAtAustin:
@keyki I am not done yet. Can you take a look at TemplateValidator.java and see if you like this approach? I am not sure using instance group name is the best way to tell it is IDBroker. Thanks!

doktoric:
> @keyki I am not done yet. Can you take a look at TemplateValidator.java and see if you like this approach? I am not sure using instance group name is the best way to tell it is IDBroker. Thanks!

can we write a unit test for it ? Also if we do the validation can we say now that count = 0?

LinaAtAustin:
@doktoric I can add unit test for volume validator, and set the condition to allow IDBroker count to be 0.

datalake/src/main/resources/duties/7.2.7/aws/medium_duty_ha.json
```diff
@@ -54,8 +54,8 @@
         "instanceType": "t3.medium",
         "attachedVolumes": [
           {
-            "count": 1,
-            "size": 100,
+            "count": 0,
+            "size": 0,
```
kmanamcheri, line 58: If we set "count" to 0, do we also have to set "size" to 0. That's a little confusing. We should remove the "size" field from this JSON if count is 0.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,13 +163,22 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
                 if (value.getVolumeCount() > config.maximumNumber()) {
                     validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
+                } else if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeCount() != 0)  {
+                        // IDBroker does not use data volume, so its volume count should be zero
+                        validationBuilder.error(String.format("IDBroker volume count: %s is not zero", value.getVolumeCount()));
```
kmanamcheri, line 180: Update the error message to state that IDBroker does not use data volume and so you are expecting it to be zero.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,13 +163,22 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
         }
     }
 
+    private boolean isIDBrokerInstanceGroup(InstanceGroup instanceGroup) {
+        return groupNameIDBroker.equalsIgnoreCase(instanceGroup.getGroupName());
```
kmanamcheri, line 167: nit: why use a constant and then a function, and then another helper function. Why can't we just use the string to compare directly?

Also, static strings should be upper case. 

```
IDBROKER_GROUP_NAME.equalsIgnoreCase(instanceGroup.getGroupName());
```

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -29,13 +29,15 @@
...
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
 import com.sequenceiq.cloudbreak.dto.credential.Credential;
 import com.sequenceiq.cloudbreak.validation.ValidationResult;
 import com.sequenceiq.cloudbreak.workspace.model.User;
 import com.sequenceiq.common.api.type.CdpResourceType;
 
 @Component
 public class TemplateValidator {
+    private static String groupNameIDBroker = "idbroker";
```
kmanamcheri, line 40: This can be `final` and by convention `static final` variables are upper cased.

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html

>If your variable stores a constant value, such as static final int NUM_GEARS = 6, the convention changes slightly, capitalizing every letter and separating subsequent words with the underscore character. By convention, the underscore character is never used elsewhere.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,13 +159,24 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
                 if (value.getVolumeCount() > config.maximumNumber()) {
                     validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
+                } else if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeCount() != 0)  {
+                        // IDBroker does not use data volume, so its volume count should be zero
+                        validationBuilder.error(
+                                String.format("IDBroker volume count: %s is not zero. It does't use data volume, and its count must be zero.",
```
kmanamcheri, line 177: You have a typo in `doesn't`

I prefer not using contractions in error logs. Use the full form `does not`

kmanamcheri (commented):
Minor type. Other than that LGTG, but I would like to get someone from CB team to approve it before merging.

LinaAtAustin:
image
DataLake provision failed for runtime 7.2.6 with the current strict check that requires IDBroker data volume count to be zero. I will update the code to loose up the validation. So we allow the IDBroker data volume count and size to be zero, but does not fail for the templates that set IDBroker data volume count and size to be larger than 0.

LinaAtAustin:
@keyki @doktoric @lajosrodek can you review my code and see if we can commit? The latest code change allows the IDBroker data column count to be zero. And also does not fail the template that configures IDBroker data column to be backward compatible. So we can support old template or existing customized template, at the same time, we can save cost for new environments. Thanks!

LinaAtAustin:
@lajosrodek The latest approach I am taking is to set the IDBroker volume count and size to be zero. However, it does not work to do so at the TemplateValidator as it is just a copy of the original template. 

1) Are you OK about this approach? This approach does not cause regression with old template, and also does not create unused data volume at IDBroker.
2) Where is the code that load the template (such as https://github.com/hortonworks/cloudbreak/blob/master/datalake/src/main/resources/duties/7.2.7/aws/medium_duty_ha.json#L57)? So I can set the IDBroker volume count and size to be zero there. Then the new values will be used when creating DataLake stack.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,12 +159,23 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
+
     private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
-                if (value.getVolumeCount() > config.maximumNumber()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeCount() > 0) {
```
doktoric, line 172: why we are doing this? we dont want to set it if the user specify us. QA team can test any scenario.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -174,11 +187,18 @@ private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumePara
...
 
     private void validateVolumeSize(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
-                if (value.getVolumeSize() > config.maximumSize()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeSize() > 0) {
```
doktoric, line 195: I dont think we should do this

LinaAtAustin:
@doktoric I have tried several approaches for this jira. Can you specify which approach I should take (approach 2) or 4)), so we can finish this code change soon? Thanks!

1) When the IDBroker data volume count is not zero, we fail the validation. I have done the code change and tested it. And this approach causes regression for older template or customized template. I don't think we should take this approach.
2) When the IDBroker data volume count is not zero, we allow this in the validation. In the new template, we set the count to be 0. This approach will allow old template to work and create data volume for IDBroker, i.e., old behavior. No data volume is created with the new template. This approach does not cause regression, and still achieve cost saving with new template.
3) When validating the volume template, set the volume count to be zero. This is the approach you see right now. My test shows it does not work. I don't think we should take this approach.
4) When loading template, set the IDBroker data volume count to be zero, i.e., ignore the user configuration for this field. Then the following processing will be as if the user configures the data volume count to be zero. This approach will not cause regression and will have cost saving. Can you tell me where is the code that I can set the count to be zero if you prefer this approach?

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
kmanamcheri: @doktoric I think we never want to attach a data volume for the IDBroker node. I don't think QA team should or want to test this. This is applicable ONLY for IDBroker nodes. The reason is that IDBroker should not need a data volume.

LinaAtAustin:
@doktoric @lajosrodek @keyki Can you review the code change. I am using the approach 4): set the IDBroker volume count and size to be zero when loading the template. 

I have tested and the IDBroker Node was created successfully for 7.2.6. 

image

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,12 +159,21 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
             if (config != null) {
-                if (value.getVolumeCount() > config.maximumNumber()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeCount() != 0) {
+                        // IDBroker does not use data volume, so its volume count should be zero
+                        validationBuilder.error(String.format("IDBroker volume count: %s is not zero", value.getVolumeCount()));
+                    }
+                } else if (value.getVolumeCount() > config.maximumNumber()) {
                     validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
                 } else if (value.getVolumeCount() < config.minimumNumber()) {
```
lajosrodek, line 178: Naive question: Is there possibly any VM type that has a positive minimum size? If so, the change above will cause conflicts if the custom DL is launched with the IDBroker using such an instance type.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -174,11 +185,16 @@ private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumePara
...
             if (config != null) {
-                if (value.getVolumeSize() > config.maximumSize()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeSize() != 0) {
+                        // IDBroker does not use data volume, so its volume size should be zero
+                        validationBuilder.error(String.format("IDBroker volume size: %s is not zero", value.getVolumeSize()));
+                    }
+                } else if (value.getVolumeSize() > config.maximumSize()) {
                     validationBuilder.error(String.format("Max allowed volume size for '%s': %s", vmType.value(), config.maximumSize()));
                 } else if (value.getVolumeSize() < config.minimumSize()) {
```
lajosrodek, line 199: Like above, I wonder if forcing the size & count to zero causes any troubles for custom DL clusters.

core/src/test/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidatorTest.java
```diff
@@ -0,0 +1,217 @@
...
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.when;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import org.junit.Assert;
```
lajosrodek, line 14: The usage of JUnit 5 would be preferable; it is high time to get rid of JUnit 4. For assertions, please consider using AssertJ `assertThat()` etc. as its fluent API is more expressive. (Neither of these are mandatory or enforced anywhere, but it would be nice to use them in new tests at least.)

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -288,9 +292,12 @@ private StackV4Request getStackRequest(SdxClusterRequest sdxClusterRequest, Stac
                 throw new BadRequestException("Can't find template for cloudplatform: " + cloudPlatform + ", shape: " + sdxClusterRequest.getClusterShape() +
                         ", runtime version: " + runtimeVersion);
             }
+            updateStackV4RequestWithIDBrokerVolume(stackRequest);
             stackRequest.getCluster().setRangerRazEnabled(sdxClusterRequest.isEnableRangerRaz());
             return stackRequest;
         } else {
+            updateStackV4RequestWithIDBrokerVolume(internalStackV4Request);
```
lajosrodek, line 299: I wonder if anyone relying on custom DL templates would be pissed about this change if they expected the volume to be present (for whatever reason). I know we are already committing such a change by overwriting the RAZ flag just a few lines below. 😄

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -299,6 +306,26 @@ private StackV4Request getStackRequest(SdxClusterRequest sdxClusterRequest, Stac
...
+            }
+
+            InstanceTemplateV4Request templateV4Request = instanceGroupV4Request.getTemplate();
+            if (templateV4Request == null || templateV4Request.getAttachedVolumes() == null) {
+                continue;
+            }
+
+            for (VolumeV4Request attachedVolumeRequest : templateV4Request.getAttachedVolumes()) {
+                // IDBroker does not use data volume, so its volume count should be zero
+                // set the data volume count and size to be zero to be backward-compatable and avoid creating volume that is not used
```
lajosrodek, line 322: Nit: compatable -> compatible.

lajosrodek (changes requested):
LGTM, just minor comments / questions.

doktoric:
> @doktoric I have tried several approaches for this jira. Can you specify which approach I should take (approach 2) or 4)), so we can finish this code change soon? Thanks!
> 
> 1. When the IDBroker data volume count is not zero, we fail the validation. I have done the code change and tested it. And this approach causes regression for older template or customized template. I don't think we should take this approach.
> 2. When the IDBroker data volume count is not zero, we allow this in the validation. In the new template, we set the count to be 0. This approach will allow old template to work and create data volume for IDBroker, i.e., old behavior. No data volume is created with the new template. This approach does not cause regression, and still achieve cost saving with new template.
> 3. When validating the volume template, set the volume count to be zero. This is the approach you see right now. My test shows it does not work. I don't think we should take this approach.
> 4. When loading template, set the IDBroker data volume count to be zero, i.e., ignore the user configuration for this field. Then the following processing will be as if the user configures the data volume count to be zero. This approach will not cause regression and will have cost saving. Can you tell me where is the code that I can set the count to be zero if you prefer this approach?

I vote for 2 (we can also modify this in the old templates if it works).

doktoric (changes requested):
I just request one last change which mentioned in comments

LinaAtAustin:
@doktoric @lajosrodek I have made code change for option 2) and all tests past. Can you review and merge the code? Thanks!

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,12 +159,22 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
-                if (value.getVolumeCount() > config.maximumNumber()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
+                    if (value.getVolumeCount() > config.maximumNumber()) {
+                        // IDBroker does not use data volume, so its volume count should be zero
+                        // To be backward-compatible, we only check max limit and allow the min limit to be zero for IDBroker
+                        validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
+                    }
+                } else if (value.getVolumeCount() > config.maximumNumber()) {
                     validationBuilder.error(String.format("Max allowed volume count for '%s': %s", vmType.value(), config.maximumNumber()));
                 } else if (value.getVolumeCount() < config.minimumNumber()) {
```
lajosrodek, line 179: Insert `!isIDBrokerInstanceGroup(instanceGroup) && ` at the beginning of condition.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -157,12 +159,22 @@ private void validateVolumeType(VolumeTemplate value, Platform platform, Validat
...
+    }
+
     private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
-                if (value.getVolumeCount() > config.maximumNumber()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
```
lajosrodek, line 171: Please do not duplicate code. There is no need for this `isIDBrokerInstanceGroup(instanceGroup)` branch. Instead, the original validation for the max limit can be reused without changes, and the extra check for the IDBroker instance group name can be added to the min limit validation below. The comment can be also moved in front of the whole `if` block, or possibly inside the min limit validation branch (whichever you prefer).

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -174,11 +186,17 @@ private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumePara
     }
 
     private void validateVolumeSize(VolumeTemplate value, VmType vmType, VolumeParameterType volumeParameterType,
-        ValidationResult.ValidationResultBuilder validationBuilder) {
+        ValidationResult.ValidationResultBuilder validationBuilder, InstanceGroup instanceGroup) {
         if (vmType != null && needToCheckVolume(volumeParameterType, value.getVolumeCount())) {
             VolumeParameterConfig config = vmType.getVolumeParameterbyVolumeParameterType(volumeParameterType);
             if (config != null) {
-                if (value.getVolumeSize() > config.maximumSize()) {
+                if (isIDBrokerInstanceGroup(instanceGroup)) {
```
lajosrodek, line 193: Like above, please remove the new `isIDBrokerInstanceGroup(instanceGroup)` branch and reuse the original max limit validation without any changes.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -174,11 +186,17 @@ private void validateVolumeCount(VolumeTemplate value, VmType vmType, VolumePara
...
+                    if (value.getVolumeSize() > config.maximumSize()) {
+                        // IDBroker does not use data volume, so its volume size should be zero
+                        // To be backward-compatible, we only check max limit and allow the min limit to be zero for IDBroker
+                        validationBuilder.error(String.format("Max allowed volume size for '%s': %s", vmType.value(), config.maximumSize()));
+                    }
+                } else if (value.getVolumeSize() > config.maximumSize()) {
                     validationBuilder.error(String.format("Max allowed volume size for '%s': %s", vmType.value(), config.maximumSize()));
                 } else if (value.getVolumeSize() < config.minimumSize()) {
```
lajosrodek, line 201: Insert `!isIDBrokerInstanceGroup(instanceGroup) && ` at the beginning of condition.

core/src/test/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidatorTest.java
```diff
@@ -0,0 +1,216 @@
...
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.when;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import org.junit.Before;
```
lajosrodek, line 15: Please use JUnit 5 (Jupiter API) for new test classes.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -92,6 +92,8 @@
 @Service
 public class SdxService implements ResourceIdProvider, ResourceBasedCrnProvider {
 
+    private static final String GROUP_NAME_ID_BROKER = "idbroker";
```
lajosrodek, line 95: Please remove, this is not needed anymore.

lajosrodek (changes requested):
Looks fine but needs some minor fixes.

doktoric (changes requested):
please fix @lajosrodek  comments and it can be merged

lajosrodek (approved):
LGTM, nice job! 😃

doktoric (approved):

PR #9388, opened 2020-11-09
CB-9534 Refactoring GrpcUmsClient so it can have proper caching. 

lnardai:
- moved out business logic from GrpcUms client so everything can be handled by EntitlementService
- Refactored UmsAccountAuthorizationService and UmsResourceAuthorizationService to use GrpcUmsClient respectively
- Added OwnerAssignmentService to honor EntitlementServiceUsage
- added warning log message in case we are trying to call UMS with actual altus account id
(we should checki kibana for results)
- Junit4-junit5 refactors

Realums test result:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/1974/


lnardai:
We should discuss tomorrow morning if the caching keys are correct or not. I've tried to be minimalistic and leave out the actorCrn because in most cases the question is raised by CB so it should be internal actor. 
I'm also fine with doing this in 2 phase, enforce the internal actor in such calls then minimize the cache keys.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -403,32 +398,61 @@ public Account getAccountDetails(String actorCrn, String accountId, Optional<Str
...
             UmsClient client = makeClient(channelWrapper.getChannel(), actorCrn);
             return client.listAssigmentsOfUser(requestId.orElse(UUID.randomUUID().toString()), userCrn);
         }
     }
 
-    //"#value.concat(#fieldId).concat(#projectId)"
-//    @Cacheable(cacheNames = "umsUserRightsCache", key = "{ #actorCrn, #userCrn, #right, #resource }")
-    @Cacheable(cacheNames = "umsUserRightsCache", key = "{ #userCrn, #right, #resource }")
-    public boolean checkRight(String actorCrn, String userCrn, String right, String resource, Optional<String> requestId) {
+    @Cacheable(cacheNames = "umsUserHasRightsForResourceCache", key = "{ #userCrn, #right, #resource }")
```
horadla23, line 409: Is it intentional that you want to use same cache as used for `hasRights `?

lnardai: is there a meaningful difference on the ask towards ums from a caching point of view?

horadla23: i guess not, only the name of the cache deceived me, but it's not important

horadla23 (approved):

foldik (approved):
lgtm

horadla23 (approved):

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -379,24 +374,53 @@ public Account getAccountDetails(String actorCrn, String accountId, Optional<Str
...
-                        ThreadBasedUserCrnProvider.getAccountId());
-                return true;
-            } else {
-                // if legacy authz then we will check permission on account level
-                resource = null;
-            }
+        return makeCheckRightCall(actorCrn, userCrn, right, resource, requestId);
+    }
+
+    @Cacheable(cacheNames = "umsUserHasRightsForResourceCache", key = "{ ##actorCrn, userCrn, #right, #resource }")
```
horadla23, line 386: i haven't noticed it, but int test brought out this double ##
```
"statusReason":"Error message: \"Internal server error: Expression [{ ##actorCrn, userCrn, #right, #resource }] @3: EL1043E: Unexpected token. Expected 'identifier' but was 'hash(#)'\""
```

lnardai: Thank you!

PR #9391, opened 2020-11-09
Cb 9647

doktoric:

### DIFF
--- a/core/src/main/resources/schema/app/20201027161305_CB-9127_optimize_cluster_template_load_time.sql
+++ b/core/src/main/resources/schema/app/20201027161305_CB-9127_optimize_cluster_template_load_time.sql
@@ -5,7 +5,9 @@ DELETE FROM clustertemplate WHERE status='DEFAULT';
 ALTER TABLE clustertemplate ADD COLUMN IF NOT EXISTS clouderaRuntimeVersion VARCHAR(255);
 
 ALTER TABLE instancegroup ADD COLUMN IF NOT EXISTS initialNodeCount INTEGER;
+ALTER TABLE instancegroup ALTER COLUMN initialNodeCount SET DEFAULT 0;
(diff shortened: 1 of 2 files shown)


biharitomi (approved):
LGTM, and saw it's working on Richard's machine 😄

core/src/main/resources/schema/app/20201027161305_CB-9127_optimize_cluster_template_load_time.sql
```diff
@@ -4,8 +4,9 @@
 DELETE FROM clustertemplate WHERE status='DEFAULT';
 ALTER TABLE clustertemplate ADD COLUMN IF NOT EXISTS clouderaRuntimeVersion VARCHAR(255);
 
-ALTER TABLE instancegroup ADD COLUMN IF NOT EXISTS initialNodeCount INTEGER;
+ALTER TABLE instancegroup ADD COLUMN IF NOT EXISTS initialNodeCount INTEGER DEFAULT 0;
```
lacikaaa, line 7: you have to set default value in a separate step to avoid table lock when adding a column:
see: https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/schema/app/20200406095012_CB-6385_add_version_to_stack.sql

lacikaaa (changes requested):

core/src/main/resources/schema/app/20201027161305_CB-9127_optimize_cluster_template_load_time.sql
lacikaaa: https://dba.stackexchange.com/questions/211221/add-column-with-default-in-postgresql-without-table-level-lock

doktoric: done

lacikaaa (approved):

PR #9394, opened 2020-11-10
CB-9661 & CB-9663: Fix FreeIPA health checks so that they do not ...

jamisonbennett:
... get stuck in unreachable/unhealthy state

    Fix the FreeIPA health checks so that they run in direct mode on
    instances that have been marked as unhealthy or unreachable.
    7f073c40338aa752cc4af8fe6d210f2a89fd8a7b introduced a regression where
    the instances status of unhealthy and unreachable was stored to the
    database. This caused the health checks to be bypassed for direct mode
    for these instances (they called isAvailable()). This resulted in a
    persistent failure until the instance was repaired. Now, only failed,
    stopped, deleted, and terminated instances use cached instance status
    for the health checks.
    
    Fix the FreeIPA health checks so that the instance status doesn't get
    stuck in an unreachable or unhealthy state.
    7f073c40338aa752cc4af8fe6d210f2a89fd8a7b introduced a regression where
    the instance status was not updated in the case where all instances
    reported a healthy status. It was because the StackStatusCheckerJob
    bypassed the ProviderChecker if the entire cluster reported it was
    available.
    
    This was tested with a local deployment of cloudbreak. The legacy
    health check was tested by first deploying a cluster using an old
    version of cloudbreak.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/entity/InstanceMetaData.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/entity/InstanceMetaData.java
@@ -142,24 +142,8 @@ public class InstanceMetaData {
         this.terminationDate = terminationDate;
     }
 
-    public boolean isCreated() {
-        return InstanceStatus.CREATED.equals(instanceStatus);
-    }
-
     public boolean isAvailable() {
         return instanceStatus.isAvailable();
-    }
-
-    public boolean isFailed() {
-        return InstanceStatus.FAILED.equals(instanceStatus);
-    }
-
-    public boolean isDecommissioned() {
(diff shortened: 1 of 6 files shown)


jamisonbennett:
I made a [comment](https://github.com/hortonworks/cloudbreak/pull/9278#pullrequestreview-515753262) on the code review which contained the regression that the health check classes are getting very tightly coupled and complicated and we should consider refactoring. This bug fix makes the coupling stronger. I would like the refactoring to happen but I want this change to modify as little code as possible as this is the second patch for health checks on the 2.33 RC branch. @christmasferret lets refactor the master branch soon and as you mentioned more logging is needed in this area of the code.

lacikaaa (approved):

PR #9400, opened 2020-11-10
CB-9684. GCP cloud storage logging: attach service account to VM inst…

oleewere:
…ances

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
@@ -9,6 +9,7 @@ import static java.util.Collections.singletonList;
 
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -124,6 +128,14 @@
                 projectId, location.getAvailabilityZone().value(), template.getFlavor()));
         instance.setDescription(description());
         instance.setName(buildableResource.get(0).getName());
+        Optional<CloudFileSystemView> cloudFileSystemView = group.getIdentity();
```
doktoric, line 131: you need the CloudIdentityType = LOG right ?

oleewere: we should set idbroker one as well 

(also we can set both service-accounts for gcp in that case, but probably should not be different against other cloud providers)

doktoric (approved):

PR #9401, opened 2020-11-10
CB-9692 getAllRecipes throws NPE if recipes field is empty in request object

sodre90:


horadla23 (approved):

keyki (changes requested):
Can you please add some unit tests?

sodre90:
sure, I created a unit test for this scenario

core-api/src/test/java/com/sequenceiq/distrox/api/v1/distrox/model/DistroXV1RequestTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+class DistroXV1RequestTest {
+
+    @Test
+    public void getAllRecipesIfRecipeNamesAreNullTest() {
+        DistroXV1Request distroXV1Request = new DistroXV1Request();
+        InstanceGroupV1Request instanceGroupV1Request = new InstanceGroupV1Request();
+        instanceGroupV1Request.setRecipeNames(null);
+        distroXV1Request.setInstanceGroups(Collections.singleton(instanceGroupV1Request));
+        distroXV1Request.getAllRecipes();
```
keyki, line 17: Shouldn't we verify that the result is an empty set?

sodre90: We can verify, I just wanted to create as minimal unit test as I can to test this use case :D anyway, assert is added

keyki (approved):

PR #9403, opened 2020-11-11
CB-9670 [GCP] Propagate the GCS identity parameters everywhere

cegganesh84:
1. In the last PR #9402 missed one location which was identified through private stack testing.
2. Will test this also in a private stack.
3. These are necessary conversions so would not mind sending a separate PR if this is not enough.

./gradlew test


doktoric (approved):

cegganesh84:
Thanks @doktoric , there were some issue with grpc artefact download, so retriggered the failed builds.

PR #9404, opened 2020-11-11
CB-9675 GCP : Add missing GCP cluster definitions for Datalake creati…

cegganesh84:
…on with 7.2.6 and 7.2.7

1. Without this we are not able to bring up GCP cluster with cloud storage.
2. I guess there is no IT test for the number of cluster definitions, only I see it for cluster templates.


doktoric:
I just wanted to raise the same PR now :D thanks @cegganesh84

doktoric (approved):

PR #9409, opened 2020-11-11
CB-9657 DataLake runtime version is null

horadla23:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
@@ -7,6 +7,7 @@ import java.util.Optional;
 
 import javax.inject.Inject;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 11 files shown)


horadla23:
still need to test it e2e, but ready for review

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
```diff
@@ -32,23 +39,64 @@
...
+                    compareRuntimeVersions(requestedStackVersion, StringUtils.substringBefore(cdh.get().getVersion(), "-"));
+                } else {
+                    throw new NotFoundException(String.format("Cannot found CDH details about related datalake stack in CB, name: %s",
+                            relatedDatalakeStack.get().getName()));
+                }
+            } else {
+                throw new NotFoundException(String.format("Cannot found related dalatake stack in CB for environment CRN %s " +
+                        "or the datalake isn't available yet.", environmentCrn));
+            }
+        } catch (Exception e) {
```
gregito, line 82: two things here. you're catching exceptions that have been thrown earlier by yourself which can be replaced with another flow control approach. maybe it's just me but I usually hate these exception driven development kind of approach 🙂 because it is not just unnecessary to invoke any kind of exception but it can be unnecessarily resource-intensive to do so.
on the other hand, you're about to catch an Exception which is "too wide" to catch. it catches everything but the catch block doesn't ensure that the original problem/exception won't happen again. surely there are many cases when multiple kinds of exception should be handle in the same way but IMO one should be aware of the original exception not just to be able to handle it properly but to make a more meaningful log entry for that scenario

gregito (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/stack/StackRuntimeVersionValidator.java
gregito: as I can remember we had such a checkstyle entry to prevent one to catch Exception. certainly, there are cases where catching Exception is inevitable, but in this case probably not. correct me if I'm wrong 🙂

horadla23: ok, i can change the try block logic do not throw those exceptions, thats fine.
regarding too wide catch, the original issue is not happening because of the code in the try block, its completely new logic to spare the unnecessary call in sdx service. i do not want different error handling here, only one thing: checking dl in cb, if any error, then fall back to existing logic (calling sdx service). i see no point for various complex error handling for a singe version validator

horadla23: > doesn't ensure that the original problem/exception won't happen again
most of the cases how can we ensure if:
- datalake is not available
- datalake is no present
- CDH version is not present

I think there is no way to do it, these are user errors, we can handle these with one way: log it and fall back to sdx call

or what would you do in these cases?

what different error handling should be here also?

in general speaking I agree with you, but I cannot see what more we can do here, please give me specific advises, if you have it regarding this case

gregito: thanks!
yeah, I'm thinking the same way. a single version validator shouldn't contain such complex error handling as that one. just the necessary but without sacrificing the foreseeing of the chances of malfunctions.

horadla23: i would keep that exception catch, there is no checkstyle rule against it. here i do no trust logic regarding clustercomponent enough to leave that catch :)

horadla23: i also removed unnecessary throws

bbihari (approved):

gregito (approved):

PR #9411, opened 2020-11-11
CB-9624 validate nodecount of instancegroup with gateway type

sodre90:


horadla23 (approved):

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceGroupV1ToInstanceGroupV4Converter.java
```diff
@@ -101,7 +105,7 @@ private SecurityGroupV4Request createSecurityGroupFromEnvironment(InstanceGroupT
     private List<String> getPorts(InstanceGroupType type) {
         List<String> ret = new ArrayList<>();
         ret.add("22");
-        if (type == InstanceGroupType.GATEWAY) {
+        if (InstanceGroupType.GATEWAY.equals(type)) {
```
doktoric, line 108: I think InstanceGroupType.GATEWAY == type  is equivalent with InstanceGroupType.GATEWAY.equals(type) and as I seen we are mostly using == in case of enum comparison

PR #9421, opened 2020-11-12
CB-9670 fix variable names that comes from fluent/settings.sls in case of GCS related fluentd output config

biharitomi:


doktoric (approved):

cegganesh84 (approved):
LGTM. Thanks!

doktoric (approved):

PR #9422, opened 2020-11-12
CB-8553 Fixed active flow gauge metric by providing object reference to be based on

daszabo:


flow/src/main/java/com/sequenceiq/flow/core/FlowRegister.java
```diff
@@ -22,13 +23,17 @@
     @Inject
     private MetricService metricService;
 
-    private final Map<String, Pair<Flow, String>> runningFlows = new ConcurrentHashMap<>();
+    private Map<String, Pair<Flow, String>> runningFlows = new ConcurrentHashMap<>();
```
attilapalfi92, line 26: Could be assigned null here since this ConcurrentHashMap will be thrown to trash in the @PostConstruct method.

Ideal solution would be initialisation in the constructor and not using PostConstruct, but such things cannot be suggested in this project :D

biharitomi (approved):
LGTM

lacikaaa:
retest this please

PR #9426, opened 2020-11-13
CB-9480 Authorize POST list endpoints

foldik:
Real UMS test:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/2021/


horadla23:
before any review please test this with real-ums-test job

foldik:
Real UMS test:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/2010/

authorization-common/src/main/java/com/sequenceiq/authorization/service/model/And.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.sequenceiq.authorization.resource.AuthorizationResourceAction;
+
+public class And implements Authorization {
```
horadla23, line 16: Maybe you should consider to rename these classes. They make sense for devs who is in the development of this feature, but other devs won't understand it at first. These names `And`, `Or`, etc. are very general. The same applies for `Authorization` interface.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/projection/StackCrnView.java
```diff
@@ -0,0 +1,8 @@
+package com.sequenceiq.cloudbreak.domain.projection;
+
+public interface StackCrnView {
```
horadla23, line 3: I am not sure if this is needed, we have StackView and StackApiView also.

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -198,6 +199,12 @@
     @Query("SELECT s FROM Stack s WHERE s.resourceCrn = :crn")
     Optional<Stack> findByResourceCrn(@Param("crn") String crn);
 
+    @Query("SELECT s.environmentCrn FROM Stack s WHERE s.resourceCrn = :crn")
+    Optional<String> findEnvCrnByResourceCrn(@Param("crn") String crn);
+
+    @Query("SELECT s.resourceCrn as resourceCrn, s.environmentCrn as environmentCrn FROM Stack s WHERE s.resourceCrn IN (:crns)")
+    List<StackCrnView> findAllByResourceCrn(@Param("crns") Set<String> crn);
```
horadla23, line 206: same here, maybe we should move this into StackViewRepository

authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceBasedCrnProvider.java
```diff
@@ -25,5 +27,9 @@ default String getResourceCrnByResourceName(String resourceName) {
         return Optional.empty();
     }
 
+    default Map<String, Optional<String>> getWithEnvironmentCrnsByResourceCrns(Collection<String> resourceCrns) {
```
horadla23, line 30: 'with' not needed in the method name i guess

horadla23 (changes requested):

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/projection/StackCrnView.java
foldik: The intention was to return the least amount of data from the database. Those views contain way more than we need here.

authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceBasedCrnProvider.java
foldik: thanks

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/projection/StackCrnView.java
horadla23: It is clear, I am just not sure if there is significant difference between performances of querying these views. If it is, then it is reasonable.

foldik: I think this interface can be easily refactored, deleted if we would need more data for this feature.

horadla23: alright

authorization-common/src/main/java/com/sequenceiq/authorization/service/model/And.java
foldik: Okay I will rename those to `AuthorizationRule`, `AllMatch` and `AnyMatch`, although I would not give them thirty letters long names just to make it clear (actually more frustrating to read) every time, in every cases that this is about authorization.

lnardai (approved):

horadla23 (approved):

PR #9427, opened 2020-11-13
CB-9501 Redbeams logging: env crn was not filled in on mow-dev.

gergopapi2:
It was hard to find log messages cross-service when looking up by env-crn. This commit adds env-crn to MDC context in controller when it is present in the request body but not in request parameters.
Also, redbeams create parameters received toStrings to be logged properly.
Unrelated change: one-line log message to track azure image copy.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/metrics/CloudbreakMetricService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/metrics/CloudbreakMetricService.java
@@ -3,6 +3,8 @@ package com.sequenceiq.cloudbreak.service.metrics;
 import java.util.ArrayList;
 import java.util.Optional;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/metrics/CloudbreakMetricService.java
```diff
@@ -77,6 +81,7 @@ public void recordImageCopyTime(Stack stack, long startMillis) {
                 MetricTag.REGION.name(), Optional.ofNullable(stack.getRegion()).orElse("NA")};
 
         long millispassed = System.currentTimeMillis() - startMillis;
+        LOGGER.debug("Image copy duration report: {} ms", System.currentTimeMillis() - startMillis);
```
keyki, line 84: The above line is exactly the same calculation.

gergopapi2: Thank you (move error ...).

keyki (approved):

PR #9429, opened 2020-11-13
CB-9748: Use resource role for database backup-restore operations.

kkalvagadda1:
Currently, database backup-restore operations are using account roles instead of resource level roles. This should be changed.


horadla23 (approved):
please squash your commits into one and we can merge this

PR #9434, opened 2020-11-16
CB-9554 recognize if salt highstate fails, because it is is running with another PID

sodre90:


flow/src/main/java/com/sequenceiq/flow/service/flowlog/FlowLogDBService.java
```diff
@@ -267,7 +268,11 @@ public FlowLog getLastFlowLogByResourceCrnOrName(String resource) {
 
     public List<FlowLog> getFlowLogsByFlowIdsCreatedDesc(Set<String> flowIds) {
         LOGGER.info("Getting flow logs by these flow ids: {}", Joiner.on(",").join(flowIds));
-        return flowLogRepository.findAllByFlowIdsCreatedDesc(flowIds);
+        if (flowIds.size() > 0) {
```
lacikaaa, line 271: `isEmpty()` please

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
```diff
@@ -99,7 +99,18 @@ public static ApplyFullResponse showState(SaltConnector sc, String state) {
 
     public static Multimap<String, Map<String, String>> jidInfo(SaltConnector sc, String jid, Target<String> target, StateType stateType) {
         if (StateType.HIGH.equals(stateType)) {
-            return highStateJidInfo(sc, jid);
+            try {
+                return highStateJidInfo(sc, jid);
+            } catch (SaltExecutionWentWrongException e) {
+                String errorMessage = e.getMessage();
+                if (errorMessage != null && errorMessage.contains("The function \"state.highstate\" is running as PID")) {
+                    String runningHighStateJid = errorMessage.replaceAll(".*The function \"state\\.highstate\" is running as PID.*with jid ", "");
```
lacikaaa, line 107: this seems to be ambiguous to me. I think it would be better to use pattern matching here to get the number, which should be a number if I remember it well.
Also there is no unit tests for this.
this branch the if could be moved to a method called sth like `extractRunningHighstateJid` or similar

lacikaaa (changes requested):

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
```diff
@@ -99,7 +99,18 @@ public static ApplyFullResponse showState(SaltConnector sc, String state) {
 
     public static Multimap<String, Map<String, String>> jidInfo(SaltConnector sc, String jid, Target<String> target, StateType stateType) {
         if (StateType.HIGH.equals(stateType)) {
-            return highStateJidInfo(sc, jid);
+            try {
```
lnardai, line 102: Unit test is missing for this logic.

lacikaaa (approved):

PR #9440, opened 2020-11-16
CB-9783 update SDX HA to final shape.

orngejaket:
For 7.2.7+, SDX HA now has 2 gateway nodes and additional 3 core node that primarily
runs Hadoop services that are dependencies of other groups.  As a result there is
reshuffling of where services are installed.  This also includes renaming master
to main.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
@@ -11,33 +11,40 @@
         "cardinality": 2,
         "refName": "master",
         "roleConfigGroupsRefNames": [
-          "hdfs-DATANODE-BASE",
-          "hdfs-JOURNALNODE-BASE",
-          "hdfs-FAILOVERCONTROLLER-BASE",
-          "hdfs-NAMENODE-BASE",
           "hive-HIVEMETASTORE-BASE",
-          "zookeeper-SERVER-BASE",
-          "kafka-GATEWAY-BASE",
-          "kafka-KAFKA_BROKER-BASE",
-          "solr-SOLR_SERVER-BASE",
           "ranger-RANGER_ADMIN-BASE",
           "hdfs-GATEWAY-BASE",
+          "hive-GATEWAY-BASE",
+          "hbase-GATEWAY-BASE",
+          "solr-GATEWAY-BASE",
+          "kafka-GATEWAY-BASE",
+          "hdfs-NAMENODE-BASE",
+          "hdfs-FAILOVERCONTROLLER-BASE"
+        ]
+      },
+      {
+        "cardinality": 2,
+        "refName": "gateway",
+        "roleConfigGroupsRefNames": [
+          "atlas-ATLAS_SERVER-BASE",
+          "ranger-RANGER_ADMIN-BASE",
+          "zookeeper-SERVER-BASE",
+          "knox-KNOX_GATEWAY-BASE",
+          "solr-GATEWAY-BASE",
+          "hdfs-GATEWAY-BASE",
+          "kafka-GATEWAY-BASE",
           "hive-GATEWAY-BASE",
           "hbase-GATEWAY-BASE"
         ]
       },
       {
-        "cardinality": 1,
-        "refName": "gateway",
+        "cardinality": 3,
+        "refName": "core",
         "roleConfigGroupsRefNames": [
           "hdfs-DATANODE-BASE",
           "hdfs-JOURNALNODE-BASE",
-          "ranger-RANGER_ADMIN-BASE",
-          "ranger-RANGER_USERSYNC-BASE",
-          "ranger-RANGER_TAGSYNC-BASE",
-          "zookeeper-SERVER-BASE",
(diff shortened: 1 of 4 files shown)


sxxgrc (approved):
Looks good to me.

hreeve-cloudera (approved):

datalake/src/main/resources/duties/7.2.7/aws/medium_duty_ha.json
```diff
@@ -44,7 +84,7 @@
         }
       },
       "nodeCount": 1,
-      "type": "GATEWAY",
+      "type": "CORE",
```
kmanamcheri, line 87: We have "core" of "CORE" type and "main"/"auxiliary" of "CORE" type. Very confusing :)

kmanamcheri (approved):

core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
```diff
@@ -361,39 +377,39 @@
...
-            "value": "1"
+            "value": "2"
           },
           {
             "name": "min.insync.replicas",
-            "value": "1"
+            "value": "2"
           },
           {
             "name": "unclean.leader.election.enable",
```
gergowilder, line 411: unclean.leader.election.enable defaults to false - you might not need to specify this parameter explicitly.

You can also consider omitting the config override for all the configs that now are set to the default values. Explicitly setting these is also perfectly fine.

doktoric:
we should add the blueprint to the application yaml file https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/application.yml#L390

datalake/src/main/resources/duties/7.2.7/aws/medium_duty_ha.json
orngejaket: Well, we started with gateway as gateway.  Might as well continue the confusion.

core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
orngejaket: Ok.

orngejaket: I removed them but default.replication.factor and min.insync.replicas weren't being set to the right value, so i've set it to default.replication.factor=3 and min.insync.replicas=2.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazBaseConfigProvider.java
```diff
@@ -30,7 +30,7 @@
             Set<HostgroupView> hostgroupViews = source.getHostgroupViews();
 
             return hostgroupViews.stream()
-                    .filter(hg -> hg.getName().toLowerCase().equals("master"))
+                    .filter(hg -> hg.getName().toLowerCase().equals("master") || hg.getName().toLowerCase().equals("main"))
```
kmanamcheri, line 33: Hmm.. Didn't think about this.. Did you run a search-all on all code for the string `"master"`. Hopefully this doesn't break things arbitrarily.

orngejaket: It deploys fine.  But I'll look further into if anything uses that name.

orngejaket: There's 3196 occurrences of "master" in CB.

kmanamcheri: > There's 3196 occurrences of "master" in CB.
:/

I am wondering if we should split this into two tasks
1. Changing the shape
2. Changing the name "master" to "main"

That way, if something does cause an issue with hardcoded string "master", we can look into that separately and we would not need to rollback the whole shape change if it comes down to that. What do you think?

core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
gergowilder: LGTM now.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazBaseConfigProvider.java
orngejaket: I'm going to do the rename of master->main in another commit so if we have to change it back, we can just revert that commit instead of the whole commit.  There's a bunch of user text and integration tests that use the name master.  Maybe we should just name gateway master (since it runs CM anyway).

PR #9449, opened 2020-11-17
CB-9807 gateway_master_secret is visible in plain text in generated D…

doktoric:
…H cluster template. Now we anonymize in json and in simple commands as well.


biharitomi (approved):

bbihari (approved):
LGTM

PR #9460, opened 2020-11-17
CB-9849 GCP : For FreeIPA use the GCP shared project id to query the …

cegganesh84:
…networks

1. If a user gives a shared project id and a vpc in it, then the reverse DNS zone calculations in FreeIPA is not done correctly.
2. To fix this we need to look at the network and subnet in the shared project.
3. This is fixed everywhere except at one place here.
4. The ultimate issue will be that DH can not talk to IDBroker, it will receive 401 due to lack of reverse DNS.
5. CB-9180 will take care of bailing out if we can not calculate reverse DNS zones.

./gradlew build

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
@@ -46,6 +46,7 @@ import com.google.api.services.compute.model.Subnetwork;
 import com.google.api.services.compute.model.SubnetworkList;
 import com.google.common.base.Strings;
 import com.sequenceiq.cloudbreak.cloud.PlatformResources;
+import com.sequenceiq.cloudbreak.cloud.exception.CloudConnectorException;
 import com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil;
 import com.sequenceiq.cloudbreak.cloud.model.AvailabilityZone;
 import com.sequenceiq.cloudbreak.cloud.model.CloudAccessConfigs;
@@ -71,8 +72,8 @@ import com.sequenceiq.cloudbreak.cloud.model.VmTypeMeta;
 import com.sequenceiq.cloudbreak.cloud.model.VmTypeMeta.VmTypeMetaBuilder;
 import com.sequenceiq.cloudbreak.cloud.model.nosql.CloudNoSqlTables;
 import com.sequenceiq.cloudbreak.cloud.model.resourcegroup.CloudResourceGroups;
-import com.sequenceiq.cloudbreak.service.CloudbreakResourceReaderService;
 import com.sequenceiq.cloudbreak.common.json.JsonUtil;
+import com.sequenceiq.cloudbreak.service.CloudbreakResourceReaderService;
 
 @Service
 public class GcpPlatformResources implements PlatformResources {
@@ -129,18 +130,18 @@ public class GcpPlatformResources implements PlatformResources {
         Map<String, Set<CloudNetwork>> result = new HashMap<>();
 
         String networkId = null;
-        String subnetId = null;
+        List<String> subnetIds = new ArrayList<>();
         String sharedProjectId = null;
         if (filters != null) {
             networkId = filters.getOrDefault("networkId", null);
-            subnetId = filters.getOrDefault("subnetId", null);
+            subnetIds = getSubnetIds(filters);
             sharedProjectId = filters.getOrDefault("sharedProjectId", null);
         }
 
-        LOGGER.debug("Get subnets with filter values, networkId : {}, subnetId : {}", networkId, subnetId);
+        LOGGER.debug("Get subnets with filter values, networkId : {}, subnetId : {}", networkId, subnetIds);
         Set<CloudNetwork> cloudNetworks = new HashSet<>();
         NetworkList networkList = getNetworkList(compute, projectId, networkId, sharedProjectId);
-        SubnetworkList subnetworkList = getSubnetworkList(region, compute, projectId, subnetId, sharedProjectId);
+        SubnetworkList subnetworkList = getSubnetworkList(region, compute, projectId, subnetIds, sharedProjectId);
 
         // GCP VPCs are global. Subnets have a global scope in region. So picking the first availability zone in the region for subnet.
         String zone = compute.regions().get(projectId, region.value())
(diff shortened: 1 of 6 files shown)


cegganesh84:
Tested in the private stack that the reverse DNS is actually populated for shared VPC with this fix https://github.com/hortonworks/cloudbreak/pull/9460

```

Received Cloud networks for region [us-west2]: [CloudNetwork{name='ent-private', id='5436519676448672560', subnets=[CloudSubnet{id='7405557895840205592', name='private-us-west2', availabilityZone='us-west2-c', cidr='10.114.128.0/20', privateSubnet=false, mapPublicIpOnLaunch=true, igwAvailable=true, parameters={}}], properties={IPv4Range=, creationTimestamp=2020-10-15T13:49:35.261-07:00, description=, gatewayIPv4=}}]

Subnets for reverse zone calculation : {7405557895840205592=10.114.128.0/20}

Reverse zones : 114.10.in-addr.arpa.

$ ipa dnszone-find --raw
  idnsname: 114.10.in-addr.arpa.
  idnszoneactive: TRUE
  idnssoamname: ipaserver0.gcp-rdns.r6zv-kv2u.dev.cldr.work.
  idnssoarname: hostmaster.gcp-rdns.r6zv-kv2u.dev.cldr.work.
  idnssoaserial: 1605655771
  idnssoarefresh: 3600
  idnssoaretry: 900
  idnssoaexpire: 1209600
  idnssoaminimum: 3600
  idnsallowquery: any;
  idnsallowtransfer: none;

  idnsname: gcp-rdns.r6zv-kv2u.dev.cldr.work.
  idnszoneactive: TRUE
  idnssoamname: ipaserver0.gcp-rdns.r6zv-kv2u.dev.cldr.work.
  idnssoarname: hostmaster.gcp-rdns.r6zv-kv2u.dev.cldr.work.
  idnssoaserial: 1605655771
  idnssoarefresh: 3600
  idnssoaretry: 900
  idnssoaexpire: 1209600
  idnssoaminimum: 3600
  idnsallowquery: any;
  idnsallowtransfer: none;
----------------------------
Number of entries returned 2
----------------------------

```

doktoric:
I just did some modification on the implementation. Probably in this way this is more maintainable.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
```diff
@@ -129,18 +129,26 @@ public CloudNetworks networks(CloudCredential cloudCredential, Region region, Ma
         Map<String, Set<CloudNetwork>> result = new HashMap<>();
 
         String networkId = null;
-        String subnetId = null;
+        List<String> subnetIds = null;
```
lacikaaa, line 132: `null` collection is not ideal, please init an empty one

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
```diff
@@ -129,18 +129,26 @@ public CloudNetworks networks(CloudCredential cloudCredential, Region region, Ma
...
-            subnetId = filters.getOrDefault("subnetId", null);
+            String subnetId = filters.getOrDefault("subnetId", null);
+            if (!Strings.isNullOrEmpty(subnetId)) {
+                subnetIds = List.of(subnetId);
+            } else {
+                String subnetIdsString = filters.getOrDefault("subnetIds", null);
+                if (!Strings.isNullOrEmpty(subnetIdsString)) {
+                    subnetIds = List.of(subnetIdsString.split(","));
+                }
+            }
```
lacikaaa, line 144: this should be moved to a separate method

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
```diff
@@ -184,18 +192,25 @@ public CloudNetworks networks(CloudCredential cloudCredential, Region region, Ma
         return new CloudNetworks(result);
     }
 
-    public SubnetworkList getSubnetworkList(Region region, Compute compute, String projectId, String subnetId, String sharedProjectId) throws IOException {
+    public SubnetworkList getSubnetworkList(Region region, Compute compute, String projectId, List<String> subnetIds,
```
lacikaaa, line 195: this could be a private method and we could skip null check on `subnetIds` if the my previous comment is addressed

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
```diff
@@ -184,18 +192,25 @@ public CloudNetworks networks(CloudCredential cloudCredential, Region region, Ma
...
-                subnetworkList = new SubnetworkList()
-                        .setItems(Collections.singletonList(compute.subnetworks().get(projectId, region.value(), subnetId).execute()));
+                subnetworkList = new SubnetworkList().setItems(new ArrayList<>());
+                for (String subnetId : subnetIds) {
+                    Subnetwork subnetwork = compute.subnetworks().get(projectId, region.value(), subnetId).execute();
+                    subnetworkList.getItems().add(subnetwork);
+                }
             }
         }
         return subnetworkList;
```
lacikaaa, line 216: this seems a bit complicated and could be simplified. Like instead of query every subnet one by one we could still use the list then filter the result.
```
    private SubnetworkList getSubnetworkList(Region region, Compute compute, String projectId, List<String> subnetIds,
        String sharedProjectId) throws IOException {
        String projectIdToQuery = StringUtils.isNotBlank(sharedProjectId) ? sharedProjectId : projectId;
        SubnetworkList subnetworkList = compute.subnetworks().list(projectIdToQuery, region.value()).execute();
        if (subnetIds.isEmpty()) {
            return subnetworkList;
        } else {
            List<Subnetwork> filteredSubnets = subnetworkList.getItems().stream()
                    .filter(subnetwork -> subnetIds.contains(subnetwork.getName()))
            return new SubnetworkList().setItems(filteredSubnets);
        }
    }

File: freeipa/src/main/java/com/sequenceiq/freeipa/service/filter/GcpNetworkFilterProvider.java [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+@Service
+public class GcpNetworkFilterProvider implements NetworkFilterProvider {
+
+    @Override
+    public Map<String, String> provide(Network network, String networkId, Collection<String> subnetIds) {
+        Map<String, String> filter = new HashMap<>();
+        if (network.getAttributes() != null && network.getAttributes().getMap() != null) {
+            Map<String, Object> attributes = network.getAttributes().getMap();
+            String sharedProjectId = (String) attributes.get("sharedProjectId");
+            filter.put("sharedProjectId", sharedProjectId);
```
lacikaaa: this constant at least should be refactored into a field to avoid duplication here

freeipa/src/main/java/com/sequenceiq/freeipa/service/filter/GcpNetworkFilterProvider.java
```diff
@@ -0,0 +1,42 @@
...
+        return CloudPlatform.GCP;
+    }
+
+    private void buildSubnetIdFilter(Collection<String> subnetIds, Map<String, String> filter) {
+        if (subnetIds != null && !subnetIds.isEmpty()) {
+            if (subnetIds.size() > 1) {
+                filter.put("subnetIds", String.join(",", subnetIds));
+            } else {
+                filter.put("subnetId", subnetIds.stream().findFirst().get());
+            }
```
lacikaaa, line 42: I don't think we need this `if`, just use `subnetIds`

freeipa/src/main/java/com/sequenceiq/freeipa/service/filter/GcpNetworkFilterProvider.java
```diff
@@ -0,0 +1,45 @@
...
+        buildSubnetIdFilter(subnetIds, filter);
+        return filter;
+    }
+
+    @Override
+    public CloudPlatform cloudPlatform() {
+        return CloudPlatform.GCP;
+    }
+
+    private void buildSubnetIdFilter(Collection<String> subnetIds, Map<String, String> filter) {
```
lacikaaa, line 36: it would be much nicer if it would return with like a map and we could use `filter.putAll` where it is called instead of modifying the map passed via a parameter. It could return with empty map if not applicable

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/NetworkService.java
```diff
@@ -49,8 +55,13 @@
     public Map<String, String> getFilteredSubnetWithCidr(String environmentCrn, Stack stack, String networkId, Collection<String> subnetIds) {
         Credential credential = credentialService.getCredentialByEnvCrn(environmentCrn);
         ExtendedCloudCredential cloudCredential = extendedCloudCredentialConverter.convert(credential);
+        Map<String, String> filter = new HashMap<>();
+        NetworkFilterProvider networkFilterProvider = networkFilterProviderMap.get(CloudPlatform.valueOf(stack.getCloudPlatform()));
+        if (networkFilterProvider != null) {
+            filter = networkFilterProvider.provide(stack.getNetwork(), networkId, subnetIds);
+        }
```
lacikaaa, line 62: could be moved to a separate method, like `createFilter`

lacikaaa (changes requested):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
doktoric: unfortunately it is not the same here? They are using the same object but in case of list most of the field is null

lacikaaa: yeah, right, instead of creating a new, the old one's items should be overridden, that should do the trick. but the old implementation did the same, creating a new `SubnetworkList` object and adding subnetoworks to it's items.

lacikaaa (approved):
LGTM, only need unit tests

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
doktoric: We just discussed on zoom

doktoric:
@cegganesh84 can you please help with unit tests?

cegganesh84:
@doktoric @lacikaaa @biharitomi I have added unit test for the new class. I could not find existing test classes for other areas of modification. Could this be merged, please? We want to unblock QE.

biharitomi (approved):

PR #9461, opened 2020-11-18
CB-6964: Wrong freeIpa status after instance is deleted on provider side

christmasferret:
During a freeipa health check, it first update instances' status and then the freeipa status. In a particular case, if the instance is deleted on provider side, the freeipa will first updates instance status to deleted. But if the environment runtime breaks ad restarts, the healthcheck will skip and won't update the freeipa status because all instances are already deleted. Therefore, after environment recovers, freeipa shows wrong stopped status. This fix will correct the freeipa status as deleted.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
@@ -118,6 +118,10 @@ public class StackStatusCheckerJob extends StatusCheckerJob {
                                 LOGGER.debug("results is empty, skip update");
                             }
                         }
+                    } else if (alreadyDeletedCount > 0) {
+                        SyncResult syncResult =  new SyncResult("FreeIpa is " + DetailedStackStatus.DELETED_ON_PROVIDER_SIDE,
+                                DetailedStackStatus.DELETED_ON_PROVIDER_SIDE, null);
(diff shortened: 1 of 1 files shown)


christmasferret:
before the fix,
Screen Shot 2020-11-17 at 4 13 12 PM

after the fix,

Screen Shot 2020-11-17 at 6 11 40 PM
Screen Shot 2020-11-17 at 6 11 28 PM

keyki (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/sync/StackStatusCheckerJob.java
```diff
@@ -118,6 +118,10 @@ public void syncAStack(Stack stack) {
                                 LOGGER.debug("results is empty, skip update");
                             }
                         }
+                    } else if (alreadyDeletedCount > 0) {
```
sodre90, line 121: Just a minor thing but can you please add a log line for this branch?

sodre90 (approved):

lacikaaa:
Is this issue present on 2.33?

keyki:
> Is this issue present on 2.33?

Good question, probably it is as the code is exactly the same. We can move this commit back to 2.33 if @christmasferret  confirms as well.

christmasferret:
confirmed on 2.33.0-b102 that it also has this issue.

keyki:
Thanks, moved it to 2.33.

PR #9462, opened 2020-11-18
CDPCP-3350. Provide user sync with an uncached hasRights implementation

handavid:
We discovered through a failing systest that caching has been added to the hasRights call. User sync needs uncached hasRights. I tossed this change together (tested locally) because we need this fixed immediately.


aarman-cloudera (approved):

keyki:
I see the on 2.33 the same cache is there. Shouldn't we fix it on that branch?

lacikaaa (approved):

horadla23:
I agree with @keyki, this should be moved into CB-2.33.0 branch

lnardai:
I would argue this is a release blocker for 2.33 since another sync will fix the issue and the error windows is quite small.

lnardai (approved):

lnardai:
I would second my original statement, this needs to be in 2.33 since it was introduced not that long ago.
CDPCP-3230 ~ Oct/27/20

handavid:
@lnardai I retargeted the change to the CB-2.33.0 branch

horadla23 (approved):

keyki (approved):

PR #9469, opened 2020-11-18
DISTX-539 Drop container size configs from templates

amitagarwal06:
[DISTX-539](https://jira.cloudera.com/browse/DISTX-539)

Fix for OPSAPS-56088 now auto tunes lot of machine dependent configs. Hence there is no need to provide static values to some of these configs in cluster templates. Also value of some of the configs are sub-optimal.
Thus removing following configs from 7.2.2, 7.2.6, 7.2.7 DE templates:
 hive.tez.container.size
 mapreduce.map.memory.mb
 mapreduce.reduce.memory.mb

Testing done:
Have created a DE cluster using a modified 7.2.6 template that includes above mentioned config changes. Cluster creation was successful on AWS using m5d.2xl instance types.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.2/cdp-data-engineering-spark3.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.2/cdp-data-engineering-spark3.bp
@@ -148,15 +148,6 @@
             "refName": "yarn-GATEWAY-BASE",
             "roleType": "GATEWAY",
             "base": true,
-            "configs": [
-              {
-                "name": "mapreduce_map_memory_mb",
-                "value": 4096
-              },
-              {
-                "name": "mapreduce_reduce_memory_mb",
-                "value": 4096
-              },
               {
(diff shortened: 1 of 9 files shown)


core/src/main/resources/defaults/blueprints/7.2.2/cdp-data-engineering-spark3.bp
```diff
@@ -148,15 +148,6 @@
             "refName": "yarn-GATEWAY-BASE",
             "roleType": "GATEWAY",
             "base": true,
-            "configs": [
```
keyki, line 151: `configs` is removed as well

amitagarwal06: Thanks Krisztian for catching this. I have fixed this.

amitagarwal06:
@sidseth @shwethags Pls approve the PR.

shwethags (approved):

PR #9472, opened 2020-11-18
CB-9878 Rename master to main for DL HA.

orngejaket:

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.7/cdp-sdx-medium-ha.bp
@@ -9,7 +9,7 @@
     "hostTemplates": [
       {
         "cardinality": 2,
-        "refName": "master",
+        "refName": "main",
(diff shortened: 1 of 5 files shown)


kmanamcheri (commented):
You had mentioned that there are lots of occurrences of the string "master". Should we search for comments etc and update them (at least in the data lake service areas)?

orngejaket:
Yea there's 3000+ occurrences of the word master in the code base.  So it's unlikely for me to find them all.  Some are part of error messages, others are part of the lightduty int tests.  So there's no telling what affects this could have, however since ha is new, it's shouldn't break anything.  I've also been able to create a datahub and run queries against it with no issue.

doktoric:
I suggest to add main to aws-vm-recommendation.json and azure-vm-recommendation.json

lacikaaa (changes requested):
I can't find any info regarding the motivation and benefit for the renaming neither in the commit message nor in the jira description. Could you add some details if possible to both of them?

doktoric:
do we still need this?

PR #9474, opened 2020-11-18
CDPSDX-1981: missing DNS a record in ipa during datalake provisioning 

christmasferret:
During datalake provisioning, an intermittent error of "does not have corresponding DNS A/AAAA record" happens. Currently we don't have logging for dnsrecord add operation. Only a timestamp exists in the log to show the dns add operation is done. This fix will log the added DNS record and the success return information in the add operation. This will help to confirm and debug the missing DNS a record issue.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
@@ -65,7 +65,7 @@ removing_dns_entries:
 
 add_dns_record:
   cmd.run:
-    - name: /opt/salt/scripts/add_dns_record.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/dnsrecord-add-executed
+    - name: /opt/salt/scripts/add_dns_record.sh 2>&1 | tee -a /var/log/dnsrecord-add.log && test ${PIPESTATUS[0]} -eq 0 && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/dnsrecord-add-executed
     - runas: root
     - failhard: True
     - unless: test -f /var/log/dnsrecord-add-executed


christmasferret:
@jamisonbennett @lacikaaa   Since you guys are experts, see if we can add some logging here to debug CDPSDX-1981


The dns record was missing as below.
Screen Shot 2020-11-18 at 4 52 03 PM


Now we only log  timestamp in /var/log/dnsrecord-add-executed  file

Screen Shot 2020-11-18 at 4 40 17 PM

With this fix, we will be able to see in  /var/log/dnsrecord-add-executed  file   if dns-record add operation is successful or not. 

Screen Shot 2020-11-18 at 4 51 26 PM

lacikaaa:
you can see the output of the script in `/var/log/salt/minion`.
to ensure the log is present in minion log and in the custom file, you could use `tee`, eg:
`/opt/salt/scripts/create-user-facing-cert.sh 2>&1 | tee -a /var/log/generate-user-facing-cert.log && exit ${PIPESTATUS[0]}`

lacikaaa (changes requested):
don't log into the executed file as the state won't run again in case of failure. If you want to log this script into a file, you should use a new one

christmasferret:
> you can see the output of the script in `/var/log/salt/minion`.
> to ensure the log is present in minion log and in the custom file, you could use `tee`, eg:
> `/opt/salt/scripts/create-user-facing-cert.sh 2>&1 | tee -a /var/log/generate-user-facing-cert.log && exit ${PIPESTATUS[0]}`


Thanks for the suggestion! I am adding below. **test ${PIPESTATUS[0]} -eq 0** is needed to carry the condition to the next echo date command. If I use **exit ${PIPESTATUS[0]}**, no matter PIPESTATUS[0] is 0 or 1, echo date will not execute. I tested this on freeipa instance. 

/opt/salt/scripts/add_dns_record.sh 2>&1 | tee -a /var/log/dnsrecord-add-history ; test ${PIPESTATUS[0]} -eq 0 && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/dnsrecord-add-executed

christmasferret:
> don't log into the executed file as the state won't run again in case of failure. If you want to log this script into a file, you should use a new one

Thanks! Indeed I missed this. It cannot log to the same execution log. I logged to /var/log/dnsrecord-add-history log so we can see the details in history log.

orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
```diff
@@ -65,7 +65,7 @@ removing_dns_entries:
 
 add_dns_record:
   cmd.run:
-    - name: /opt/salt/scripts/add_dns_record.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/dnsrecord-add-executed
+    - name: /opt/salt/scripts/add_dns_record.sh 2>&1 | tee -a /var/log/dnsrecord-add-history ; test ${PIPESTATUS[0]} -eq 0 && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/dnsrecord-add-executed
```
lacikaaa, line 68: - is there a reason you used `;` instead of `&&`?
- I think `dnsrecord-add.log` would be better name, and more consistent with the other salt states, as we use the `.log` for logs and the `executed` ending for indicating that the script has run

christmasferret: `;` and && are same here. I changed it to && now. I feel ; would make the one line only separated by a single &&. Therefore, it is clear that first block "/opt/salt/scripts/add_dns_record.sh 2>&1 | tee -a /var/log/dnsrecord-add-history ; test ${PIPESTATUS[0]} -eq 0 " was a whole section. But this is more of a style preference. 

Thanks! I changed the log name to "dnsrecord-add.log"

lacikaaa (approved):

PR #9476, opened 2020-11-18
CB-9852: Redact FreeIPA admin password

jamisonbennett:
Redact the FreeIPA admin password.


lacikaaa (approved):
LGTM, only thing I'm missing is the unit tests which covers the pattern changes

PR #9482, opened 2020-11-19
CB-9902. Increase the default node count for the compute group to 1

sidseth:
for the DE cluster shapes.
This is a workaround for OPSAPS-54928, where starting a compute group
with 0 nodes causes upscaled nodes belonging to this group to be
misconfigured, and not consume available resources (8GB for NMs).


cegganesh84 (approved):
LGTM

smaniraju (approved):

PR #9483, opened 2020-11-20
CB-7210: Ensure that RAZ role is included when RAZ is requested, and isn't when it is not.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-7210

Specifically this should make it so that RAZ datalakes cannot be created if they don't include the rangerraz IDBMMS role. Similarly, if they do include the role and are note trying to create a RAZ datalake, should return an error. 

Note that while we definitely want the first of these changes, we are not sure if the second is ideal. I hope these changes facilitate a discussion on this.

I also included some quick tests that kind of test this, please let me know if I need to expand on them in any way.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
@@ -247,6 +247,7 @@ public class StackRequestManifester {
                 try {
                     // Must pass the internal actor here as this operation is internal-use only; requests with other actors will be always rejected.
                     mappingsConfig = idbmmsClient.getMappingsConfig(INTERNAL_ACTOR_CRN, environmentCrn, Optional.empty());
+                    validateMappingsConfig(mappingsConfig, stackRequest);
                 } catch (IdbmmsOperationException e) {
                     throw new BadRequestException(String.format("Unable to get mappings: %s", e.getMessage()), e);
                 }
@@ -266,6 +267,16 @@ public class StackRequestManifester {
             // getAccountMapping() != null is possible only in case of SdxInternalClusterRequest, in which case the user-given values will be honored.
             LOGGER.info("{} for stack {} in environment {}.", cloudStorage == null ? "Cloud storage is disabled" : "Applying user-provided mappings",
                     stackName, environmentCrn);
+        }
+    }
+
+    void validateMappingsConfig(MappingsConfig mappingsConfig, StackV4Request stackRequest) {
+        // Validate RAZ if enabled, making sure that the RAZ mapping exists when it is required.
+        if (stackRequest.getCluster().isRangerRazEnabled()) {
+            if (!mappingsConfig.getActorMappings().containsKey("rangerraz")) {
+                LOGGER.error("IDBMMS mappings must contain the RAZ role if RAZ is to be created!");
+                throw new BadRequestException("IDBMMS mappings must contain the RAZ role if RAZ is to be created!");
+            }
         }
     }
 
--- a/datalake/src/test/java/com/sequenceiq/datalake/service/sdx/StackRequestManifesterTest.java
+++ b/datalake/src/test/java/com/sequenceiq/datalake/service/sdx/StackRequestManifesterTest.java
@@ -217,6 +217,36 @@ public class StackRequestManifesterTest {
     }
 
     @Test
+    void testRAZSetupCloudStorageAccountMappingsWithRAZMapping() {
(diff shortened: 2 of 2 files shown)


sxxgrc:
Please feel free to add other reviewers that may understand the code I'm changing more. Thanks!

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -269,6 +270,27 @@ void setupCloudStorageAccountMapping(StackV4Request stackRequest, String environ
...
+        } else {
+            if (mappingsConfig.getActorMappings().containsKey("rangerraz")) {
+                errorMessage = "IDBMMS mappings cannot contain the RAZ role if RAZ is not enabled for the cluster!";
+            }
+        }
+
+        // Throw error if validation failed.
+        if (errorMessage != null) {
+            LOGGER.error(errorMessage);
+            throw new IdbmmsOperationException(errorMessage);
```
lajosrodek, line 290: Minor: I would rather directly throw `BadRequestException` since the latter more appropriately expresses that the problem is due to a conflict of cluster settings vs IDBroker mappings. `IdbmmsOperationException` is supposed to  denote problems specific to IDBMMS, e.g. due to missing mappings, communication error or some other internal service error.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -269,6 +270,27 @@ void setupCloudStorageAccountMapping(StackV4Request stackRequest, String environ
...
+    void validateMappingsConfig(MappingsConfig mappingsConfig, StackV4Request stackRequest) {
+        String errorMessage = null;
+
+        // Validate RAZ if enabled, making sure that the RAZ mapping exists only when it is required.
+        if (stackRequest.getCluster().isRangerRazEnabled()) {
+            if (!mappingsConfig.getActorMappings().containsKey("rangerraz")) {
+                errorMessage = "IDBMMS mappings must contain the RAZ role if RAZ is to be created!";
+            }
+        } else {
+            if (mappingsConfig.getActorMappings().containsKey("rangerraz")) {
```
lajosrodek, line 282: I see the general intent of this check, but IDBMMS always includes mappings for other service principals even if they will never be used by any clusters. So I see no real benefit of prohibiting this mapping rule if no RAZ is to be installed. (No other users are supposed to be able to ever impersonate themselves as `rangerraz`.)

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
orngejaket: Badrequest does sound better.

orngejaket (changes requested):

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
gitmfox: +1 to @lajosrodek's comment. I was the one who initially suggested that we might want to disallow setting the RAZ role when creating non-RAZ datalakes, but Lajos makes a good point here, which hadn't occurred to me.

The rangerraz role is a more powerful role than other roles, but, as Lajos points out, our security model assumes that users are not able to impersonate other users.

lajosrodek (approved):
LGTM

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/StackRequestManifester.java
```diff
@@ -269,6 +270,23 @@ void setupCloudStorageAccountMapping(StackV4Request stackRequest, String environ
         }
     }
 
+    void validateMappingsConfig(MappingsConfig mappingsConfig, StackV4Request stackRequest) {
+        String errorMessage = null;
+
+        // Validate RAZ if enabled, making sure that the RAZ mapping exists when it is required.
+        if (stackRequest.getCluster().isRangerRazEnabled()) {
+            if (!mappingsConfig.getActorMappings().containsKey("rangerraz")) {
+                errorMessage = "IDBMMS mappings must contain the RAZ role if RAZ is to be created!";
```
orngejaket, line 279: Just throw the exception here.  Remove the errorMessage variable.

orngejaket (changes requested):

orngejaket (approved):

PR #9487, opened 2020-11-20
CB-9917 [GCP]Datalake creation failed with freeipa backup failure. Fr…

doktoric:
…eeipa backup now configured.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
@@ -6,6 +6,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.common.api.cloudstorage.old.AdlsGen2CloudStorageV1Parameters;
+import com.sequenceiq.common.api.cloudstorage.old.GcsCloudStorageV1Parameters;
 import com.sequenceiq.common.api.cloudstorage.old.S3CloudStorageV1Parameters;
 import com.sequenceiq.common.api.telemetry.request.LoggingRequest;
 import com.sequenceiq.common.api.telemetry.request.TelemetryRequest;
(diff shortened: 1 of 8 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/templates/freeipa_backup.conf.j2
```diff
@@ -2,5 +2,6 @@
 backup_location={{freeipa.backup_location}}
 backup_platform={{freeipa.backup_platform}}
 azure_instance_msi={{freeipa.azure_instance_msi}}
+gcp_service_account={{freeipa.gcp_service_account}}
```
oleewere, line 5: is this a valid option for the backup, or just pre-filled and will be used somehow later?

freeipa/src/main/java/com/sequenceiq/freeipa/converter/backup/BackupConverter.java
```diff
@@ -53,6 +54,11 @@ private void decorateBackupFromLoggingRequest(Backup backup, LoggingRequest logg
                 adlsGen2Params.setSecure(adlsGen2FromRequest.isSecure());
                 adlsGen2Params.setManagedIdentity(adlsGen2FromRequest.getManagedIdentity());
                 backup.setAdlsGen2(adlsGen2Params);
+            } else if (loggingRequest.getGcs() != null) {
+                GcsCloudStorageV1Parameters gcsParams = new GcsCloudStorageV1Parameters();
+                GcsCloudStorageV1Parameters gcsFromRequest = loggingRequest.getGcs();
+                gcsParams.setServiceAccountEmail(gcsFromRequest.getServiceAccountEmail());
+                backup.setGcs(gcsFromRequest);
```
lajosrodek, line 61: I think the argument should be `gcsParams` instead to be consistent with S3 / ADLSg2 above.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/templates/freeipa_backup.conf.j2
doktoric: I guess so because the azure instance msi is also presented here

oleewere (approved):
lgtm

PR #9496, opened 2020-11-23
CB-9443 adding freeipa health check polling to freeipa start process

afarsang:

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/start/OneFreeIpaReachableAttempt.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/start/OneFreeIpaReachableAttempt.java
@@ -1,0 +1,59 @@
+package com.sequenceiq.freeipa.flow.stack.start;
+
+import static com.sequenceiq.cloudbreak.util.Benchmark.checkedMeasure;
+
+import java.util.Optional;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.dyngr.core.AttemptMaker;
+import com.dyngr.core.AttemptResult;
+import com.dyngr.core.AttemptResults;
+import com.sequenceiq.cloudbreak.cloud.scheduler.PollGroup;
+import com.sequenceiq.cloudbreak.cloud.store.InMemoryStateStore;
+import com.sequenceiq.freeipa.client.model.RPCResponse;
+import com.sequenceiq.freeipa.entity.InstanceMetaData;
+import com.sequenceiq.freeipa.entity.Stack;
+import com.sequenceiq.freeipa.service.stack.FreeIpaHealthDetailsService;
+
+public class OneFreeIpaReachableAttempt implements AttemptMaker<Void> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(OneFreeIpaReachableAttempt.class);
+
+    private final FreeIpaHealthDetailsService freeIpaHealthDetailsService;
+
+    private final Stack stack;
+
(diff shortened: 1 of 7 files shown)


jamisonbennett (changes requested):
Thanks for catching this case. I like the idea of running a health check on the startup and I think there is a cleaner way to do it with less code.

@christmasferret added the health check to the end of several repair flows so that the cluster woudln't report available unless everything was actually working correctly. Before this change, the end of the flow caused the status to be available regardless of the actually health check.

See https://github.com/hortonworks/cloudbreak/blob/53fe5bb630050b7f51094d7ae9a607c5fc834cf6/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/action/ChangePrimaryGatewayActions.java#L158 which calls the health check on the last step of the last flow in the flow chain. This calls the same logic as periodic quartz health check job which updates the stack status.

The health checks are already rather complicated with some tightly coupled classes, I would rather avoid duplication of logic.

For the case of a startup, I am not sure if we will need a delay before calling a health check. I know that FreeIPA starts up pretty quickly, but I don't know if it is that instantaneous or not. You might want to test this manually and comment on it in the PR.

afarsang:
@jamisonbennett thanks for the review and your comment.
So the problem was when environment was started the Freeipa starting took several minutes  in Google Cloud (in my local machine it was more than one minute on Google Cloud). It caused problem when CB reported Freeipa available.
My idea was don't let the flow finish until at least one Freeipa service is working. 
What you suggested it is like finish the flow but report back that Freeipa is unhealthy and let the Autosync later set its status accordingly. Did I get it well?
If so I am not sure it will solve the original problem (If datalake reaches to FreeIPA for something during that window, it might fail initially)

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/start/OneFreeIpaReachableAttempt.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public OneFreeIpaReachableAttempt(FreeIpaHealthDetailsService freeIpaHealthDetailsService, Stack stack, Set<InstanceMetaData> instanceMetaDataSet) {
+        this.freeIpaHealthDetailsService = freeIpaHealthDetailsService;
+        this.stack = stack;
+        this.instanceMetaDataSet = instanceMetaDataSet;
+    }
+
+    @Override
+    public AttemptResult<Void> process() {
+        if (stack == null || instanceMetaDataSet == null || instanceMetaDataSet.isEmpty()) {
```
topolyai5, line 38: You can check this in the constructor or in the factory.

topolyai5 (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/start/OneFreeIpaReachableAttempt.java
afarsang: OK, thanks

topolyai5 (approved):

jamisonbennett:
Thanks for the explanation. I agree with you.

I think that the FreeIPA reboot repair will also have the same problem with the GCP health check. See https://github.com/hortonworks/cloudbreak/blob/9bfe8dac9471fecaaf30520051e8ae88a9b985b3/freeipa/src/main/java/com/sequenceiq/freeipa/flow/instance/reboot/RebootService.java#L48. Can you update this to your new service rather than calling the health checks.

The rebuild repair should be OK since it makes sure the instances have started.

jamisonbennett (approved):
Thanks for fixing the health checks when FreeIPA is starting up.

PR #9497, opened 2020-11-23
CB-9925 HDFS service health warnings

orngejaket:
This removes the setting of dfs_journalnode_edits_dir from the medium
duty blueprints along with changing the HdfsVolumeConfigProvider to
be more in line with the other configs.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.2/cdp-sdx-medium-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.2/cdp-sdx-medium-ha.bp
@@ -314,12 +314,6 @@
             },
             {
               "base": true,
-              "configs": [
-                {
-                  "name": "dfs_journalnode_edits_dir",
-                  "value": "/dfs/jn"
-                }
-              ],
               "refName": "hdfs-JOURNALNODE-BASE",
               "roleType": "JOURNALNODE"
             }
--- a/core/src/main/resources/defaults/blueprints/7.2.6/cdp-sdx-medium-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.6/cdp-sdx-medium-ha.bp
@@ -314,12 +314,6 @@
             },
             {
               "base": true,
-              "configs": [
-                {
-                  "name": "dfs_journalnode_edits_dir",
-                  "value": "/dfs/jn"
-                }
-              ],
               "refName": "hdfs-JOURNALNODE-BASE",
               "roleType": "JOURNALNODE"
             }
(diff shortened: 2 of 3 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hdfs/HdfsVolumeConfigProvider.java
```diff
@@ -40,7 +39,7 @@
             case HdfsRoles.JOURNALNODE:
                 return List.of(
                         config("dfs_journalnode_edits_dir",
-                                buildSingleVolumePath(volumeCount, "journalnode"))
+                                buildVolumePathStringZeroVolumeHandled(volumeCount, "journalnode"))
```
keyki, line 42: Are you sure it accepts multiple volumes?

adoroszlai: I think this is the HDFS code where the property is used, seems to confirm your findings:

https://github.com/apache/hadoop/blob/c4ba0ab7dfcfe76f928d06fcb29ca0cc4d778737/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/qjournal/server/JournalNode.java#L178-L182
https://github.com/apache/hadoop/blob/c4ba0ab7dfcfe76f928d06fcb29ca0cc4d778737/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/qjournal/server/JournalNode.java#L335-L340

sodonnel: Yea based on the code @adoroszlai shared and also the docs in hdfs-default.xml:

```
<property>
  <name>dfs.journalnode.edits.dir</name>
  <value>/tmp/hadoop/dfs/journalnode/</value>
  <description>
    The directory where the journal edit files are stored.
  </description>
</property>
```

The journal edits dir can only have a single value. This makes it different from the NN data directory where the namenode can have a comma separated list of directories.

orngejaket: And fixed.

keyki (approved):

oleewere (approved):

PR #9499, opened 2020-11-23
CB-9927: Restore flow is in stuck state when there is a failure in performing database restore.

kkalvagadda1:
There are changes needed in both backup-restore flows to handle failures properly. There were issues in flow configurations causing the flows to be stuck in a pending state.

Below were the issues.
1. When the flows enter the DATALAKE_DATABASE_RESTORE_COULD_NOT_START_STATE and DATALAKE_DATABASE_BACKUP_COULD_NOT_START_STATE states, flows do not wait for the completion of the datalake backup.
2. When the flows enter FullBackupInProgress and FullRestoreInProgress states, database operation status was always set to SUCCESS. That need not be true.
3. Corrected the bean DATALAKE_DATABASE_RESTORE_FINISHED_STATE to DATALAKE_RESTORE_FINISHED_STATE.


sodre90 (approved):

oleewere (approved):
lgtm

PR #9503, opened 2020-11-24
CB-9957 in case of DH cluster we can provision external databases. Th…

doktoric:
…e problem was that somehow the external db reference was null. We should investigate further how it caused but this fix is basically checking the existence of the external rds crn before initiate the start/stop/delete.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.dyngr.Polling;
+import com.google.common.base.Strings;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.database.DatabaseAvailabilityType;
 import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.Cluster;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -87,20 +92,32 @@ public void startDatabase(Cluster cluster, DatabaseAvailabilityType externalData
...
+                waitAndGetDatabase(cluster, databaseCrn, DatabaseOperation.START, false);
+            } else {
+                LOGGER.warn("[INVESTIGATE] The external database type was {} but there was no crn", externalDatabase);
+            }
         } catch (NotFoundException notFoundException) {
             LOGGER.info("Database server not found on redbeams side {}", databaseCrn);
         }
     }
 
+    public boolean externalDatabaseReferenceExist(String databaseCrn) {
```
bergerdenes, line 106: please set as private

bergerdenes (approved):
lgtm

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -99,13 +108,21 @@ public void stopDatabase(Cluster cluster, DatabaseAvailabilityType externalDatab
                 externalDatabase.name(), environment.getName(), cluster.getName());
         String databaseCrn = cluster.getDatabaseServerCrn();
         try {
-            redbeamsClient.stopByCrn(databaseCrn);
-            waitAndGetDatabase(cluster, databaseCrn, DatabaseOperation.STOP, false);
+            if (externalDatabaseReferenceExist(databaseCrn)) {
+                redbeamsClient.stopByCrn(databaseCrn);
+                waitAndGetDatabase(cluster, databaseCrn, DatabaseOperation.STOP, false);
+            } else {
+                LOGGER.warn("[INVESTIGATE] The external database type was {} but there was no crn", externalDatabase);
```
keyki, line 115: Not sure if the MDCcontext is correct at these places, but if not then probably it would be better to include more information in the logs, like data hub name or crn.

doktoric: it must be correct. we checked the logs and the information is there. I just want to highlight how many times it happened.

PR #9504, opened 2020-11-24
CB-9646 [AuthZ] Improve error pop up message: display resource name i…

bbihari:
…nstead of crn

To help users understand what is happening on the UI we added the resource name to the authorization error message where it is possible and won't consume too much resources.

real ums tests: http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/2246/

### DIFF
--- a/authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceAuthorizationService.java
+++ b/authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceAuthorizationService.java
@@ -4,10 +4,13 @@ import static java.util.stream.Collectors.toList;
 
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
@@ -52,6 +55,9 @@ public class ResourceAuthorizationService {
     @Inject
     private List<AuthorizationFactory<? extends Annotation>> authorizationFactories;
 
+    @Inject
+    private ResourceNameFactoryService resourceNameFactoryService;
+
     public void authorize(String userCrn, ProceedingJoinPoint proceedingJoinPoint, MethodSignature methodSignature, Optional<String> requestId) {
         boolean authzEntitled = isAuthorizationEntitlementRegistered();
         Function<AuthorizationResourceAction, String> rightMapper = umsRightProvider.getRightMapper(authzEntitled);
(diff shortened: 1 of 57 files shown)


authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceAuthorizationService.java
```diff
@@ -64,7 +68,7 @@ public void authorize(String userCrn, ProceedingJoinPoint proceedingJoinPoint, M
                     LOGGER.debug("Resource authorization failed: {}", failedAuthorization.toString(rightMapper));
                 }
                 if (authzEntitled) {
-                    throw new AccessDeniedException(failedAuthorization.getAsFailureMessage(rightMapper));
+                    throw new AccessDeniedException(failedAuthorization.getAsFailureMessage(rightMapper, nameMapper));
```
foldik, line 71: cc: @lnardai , @sodre90 

This structure can be quite inefficient when we have a bunch of resources and the user doesn't have right on most of them (delete 100 datahubs and the user doesn't have rights on them maps to 101 different database queries, 1 for the env and 100 for the datahubs). Basically the underlying behaviour is like you do this:

```
for (String crn : crns) {
  String name = db.query("SELECT name FROM xy WHERE crn = :crn")
}
```

instead of `SELECT name FROM xy WHERE crn IN (:crns)` with all the .

I would restructure the logic based on the followings:

- collect the crns from the failedAuthorization to a `Set<String>`,
- sort them into a `Map<ResourceType, Set<String>>`, basically crns by resource types,
- change the `String ResourceNameProvider.getNameByCrn(String)` to `Map<String, String> ResourceNameProvider.getNamesByCrns(Set<String>)` and use the `IN` sql stament in the underlying implementations (most of the time this is easily achiavable except `EnvironmentClientService`),
- pass the result `Map<String, String>` name-crn map to the `AuthorizationRule`.

I checked the code, the above solution wouldn't complicate the code at all and we wouldn't have worse and worse performance with increasing request size. I can even give a few ours to implement this and compare the solution if you wan't it.

foldik (changes requested):
Can have inefficient behaviour when the user doesn't have rights on a lot of resources. It might worth to refactor it a little bit. I provided instructions for the solution.

authorization-common/src/main/java/com/sequenceiq/authorization/service/ResourceAuthorizationService.java
bbihari: 1) I don't think that this would cause performance issues. The names only collected in the case of failed authorizations (in most cases there is only _one_ crn)
2) the code would be more complicated, and different form the already implemented logic (e.g. rightMapper)

authorization-common/src/test/java/com/sequenceiq/authorization/service/ResourceAuthorizationServiceTest.java
```diff
@@ -82,6 +85,7 @@ public void setUp() {
         when(authorizationFactory1.supportedAnnotation()).thenReturn(CheckPermissionByResourceCrn.class);
         when(authorizationFactory2.supportedAnnotation()).thenReturn(CheckPermissionByResourceName.class);
         lenient().when(umsRightProvider.getRightMapper(true)).thenReturn(AuthorizationResourceAction::getRight);
+        //when(resourceNameFactoryService.getNames(any())).thenReturn(Map.of("crn", empty(), "crn1", empty(), "crn2", empty()));
```
foldik, line 88: Do we need this?

foldik (approved):
lgtm

foldik:
Please link the real ums test.

authorization-common/src/test/java/com/sequenceiq/authorization/service/ResourceAuthorizationServiceTest.java
bbihari: nope, thanks finding it :)

PR #9506, opened 2020-11-24
CB-9761 Do not check upgrade matrix for CM version

Bajzathd:
CDH and CM versions may be different from now on, so the upgrade matrix should only be used for CDH versions.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.service.upgrade;
 
+import static com.sequenceiq.cloudbreak.service.upgrade.image.ClusterUpgradeImageFilter.CM_PACKAGE_KEY;
+
 import java.util.Optional;
 
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
```diff
@@ -23,12 +23,17 @@
     private ComponentVersionComparator componentVersionComparator;
 
     public boolean permitCmAndStackUpgrade(ImageFilterParams imageFilterParams, Image image, String versionKey, String buildNumberKey) {
+        return permitCmAndStackUpgrade(imageFilterParams, image, versionKey, buildNumberKey, imageFilterParams.isCheckUpgradeMatrix());
```
keyki, line 26: Wouldn't it be better to set the imageFilterParams  checkUpgradeMatrix to false in case of CM instead of creating a different method where you can override it. It will be difficult to maintain later.

Bajzathd: I agree that it may not be the best solution, but overriding the imageFilterParams for one call does not seems better. How about separating upgradePermissionProvider.permitCmAndStackUpgrade to 2 different methods? And in the cm method we do not care about the matrix.

keyki: Yeah, that seems better.

keyki (approved):

lacikaaa (approved):

PR #9509, opened 2020-11-24
CB-9548 enable firewall rules in case google cloud which are comming …

doktoric:
…from the shared/host project or the user create environment without any firewall rule. Tested scenarios are 1. the vpc information coming from the base project without firewall information. 2. the vpc is shared vpc and the firewall rule coming from the host project. 3. the vpc is shared vpc and the firewall rule coming from the base project.

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.cloud.gcp;
 
+import static com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil.SHARED_PROJECT_ID;
 import static com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil.getMissingServiceAccountKeyError;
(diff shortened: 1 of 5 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpPlatformResources.java
```diff
@@ -262,6 +263,23 @@ public CloudSecurityGroups securityGroups(CloudCredential cloudCredential, Regio
...
+                try {
+                    FirewallList sharedProjectFirewalls = compute.firewalls().list(sharedProjectId).execute();
+                    for (Firewall firewall : sharedProjectFirewalls.getItems()) {
+                        Map<String, Object> properties = new HashMap<>();
+                        properties.put("network", getNetworkName(firewall));
+                        CloudSecurityGroup cloudSecurityGroup = new CloudSecurityGroup(firewall.getName(), firewall.getName(), properties);
+                        result.computeIfAbsent(region.value(), k -> new HashSet<>()).add(cloudSecurityGroup);
+                    }
+                } catch (Exception ex) {
+                    LOGGER.warn(String.format("We can not read the host project which id %s", sharedProjectId));
```
cegganesh84, line 278: nit : which -> with

"We can not read the host project with id"

cegganesh84 (approved):
LGTM from GCP perspective.

PR #9513, opened 2020-11-25
CB-9879 Allow option to not start datahub cluster automatically when …

doktoric:
…environment starts up

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentStartDto.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentStartDto.java
@@ -1,0 +1,85 @@
+package com.sequenceiq.environment.environment.dto;
+
+import com.sequenceiq.cloudbreak.common.event.Payload;
+import com.sequenceiq.common.api.type.DataHubStartAction;
+
+public class EnvironmentStartDto implements Payload {
+
+    private Long id;
+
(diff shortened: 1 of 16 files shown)


doktoric:
retest this please

topolyai5 (changes requested):
I have a suggestion. If we plan to add the `Recover to before stop` state, much better to use an enum with 2 different states. We can extend the num easily

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentStartDto.java
```diff
@@ -0,0 +1,85 @@
+package com.sequenceiq.environment.environment.dto;
```
topolyai5, line 1: is this a good package? I think it does not relate to telemetruy

common-model/src/main/java/com/sequenceiq/common/api/telemetry/model/DataHubStartAction.java
```diff
@@ -0,0 +1,6 @@
+package com.sequenceiq.common.api.telemetry.model;
```
topolyai5, line 1: is this a good package? I think it does not relate to telemetruy

doktoric:
restest this please

topolyai5 (approved):

doktoric:
retest this please

PR #9518, opened 2020-11-25
CB-9580 Salt changes to configure the embedded Postgres DB to use the preserved disk

schfeca75:
**CB-9580 Salt changes to configure the embedded Postgres DB to use the preserved disk**

- introduce new entitlement: CDP_EMBEDDED_DATABASE_ON_ATTACHED_DISK. If it is enabled, the embedded postgres sql server will be setup to use an attached disk to store its data otherwise the default setup will be used.
- salt pillar config changes for postgres embedded db: the sqlserver's directlory, the log directory and the entitlement value will be set in pillar
- salt state changes: setup embedded postgres sql server with the provided pillar parameters.

Hint: These changes only setup postgres sql server for the '/hadoopfs/fs1' attached disk if the entitlement is enabled. The future plan is to setup a separated attached disk for the postgres sql server.

**Enable gateway repair if the embedded database data is stored on attached disk**

- Extend Cluster with new field: embeddedDatabaseOnAttachedDisk to persist where the embedded database stores data
- Gateway repair will be enabled if embedded database stores its data on attached disk
- If entitlement will change it won't affect the existing clusters' behaviour.


mock-thunderhead/src/main/resources/application.yml
```diff
@@ -33,3 +33,4 @@ auth:
     differentdatahubversionthandatalake.enabled: true
     database.wire.encryption.enable: true
     datahub.runtime.upgrade.enable: true
+    embedded.database.on.attached.disk.enable: true
```
keyki, line 36: My recommendation is to make it false by default. We'll enable it once we have a reserved disk for it.

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
```diff
@@ -24,14 +26,29 @@ init-services-db-remote:
 
 init-db-with-utf8:
   cmd.run:
-    - name: rm -rf /var/lib/pgsql/data && runuser -l postgres sh -c 'initdb --locale=en_US.UTF-8 /var/lib/pgsql/data > /var/lib/pgsql/initdb.log' && rm /var/log/pgsql_listen_address_configured
+    - name: rm -rf {{ postgres_data_directory }} && runuser -l postgres sh -c 'initdb --locale=en_US.UTF-8 {{ postgres_data_directory }} > /var/lib/pgsql/initdb.log' && rm /var/log/pgsql_listen_address_configured
     - unless: grep -q UTF-8 /var/lib/pgsql/initdb.log
 
+{%- if postgres_data_on_attached_disk %}
```
keyki, line 32: Will it work for None? None is considered false?

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
```diff
@@ -24,14 +26,29 @@ init-services-db-remote:
...
+
+change-db-location:
```
keyki, line 34: Take a look at https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/freeipa-salt/salt/freeipa/services.sls on how you can figure out service unit files.

keyki:
@schfeca75 please write a proper commit message explaining why the entitlement is needed and what this change is supposed to fix/improve.

mock-thunderhead/src/main/resources/application.yml
lacikaaa: even for the mock?

keyki: mock means everyone who is developing CB. If you say it's fine we can try with true (at least people are going to test it without knowing it :) )

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
lacikaaa: in theory this is always set if no remote db is configured. But would be safer to include `None`

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
```diff
@@ -24,14 +26,29 @@ init-services-db-remote:
...
+  file.replace:
+    - name: /usr/lib/systemd/system/postgresql-10.service
+    - pattern: "Environment=PGDATA=.*"
+    - repl: Environment=PGDATA=/hadoopfs/fs1/pgsql/data
```
keyki, line 38: On Ycloud we don't have any disks at all. In that case our generation will result in /hadoopfs/root1 instance of /hadoopfs/fs1 so assuming 1 might not going to work.

lacikaaa: this repl value is wrong, the unless checks for the directory calculated on java side while this is hard coded: `{{ postgres_data_directory }}`

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/PostgresConfigService.java
```diff
@@ -56,10 +64,25 @@ public void decorateServicePillarWithPostgresIfNeeded(Map<String, SaltPillarProp
...
             postgresConfig.put("configure_remote_db", "true");
+        } else {
+            collectEmbeddedDatabaseConfigs(stack, postgresConfig);
         }
         rdsConfigProviderFactory.getAllSupportedRdsConfigProviders().forEach(provider ->
                 postgresConfig.putAll(provider.createServicePillarConfigMapIfNeeded(stack, cluster)));
         return postgresConfig;
     }
 
+    private void collectEmbeddedDatabaseConfigs(Stack stack, Map<String, Object> postgresConfig) {
```
lacikaaa, line 75: I made some suggestion on slack for refactoring this method

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
```diff
@@ -1,6 +1,8 @@
 {%- from 'metadata/settings.sls' import metadata with context %}
 
 {% set configure_remote_db = salt['pillar.get']('postgres:configure_remote_db', 'None') %}
+{% set postgres_data_directory = salt['pillar.get']('postgres:postgres_data_directory') %}
+{% set postgres_data_on_attached_disk = salt['pillar.get']('postgres:postgres_data_on_attached_disk', 'None') %}
```
lacikaaa, line 5: default should be false and not `None`, shouldn't it?

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/PostgresConfigService.java
```diff
@@ -56,10 +64,25 @@ public void decorateServicePillarWithPostgresIfNeeded(Map<String, SaltPillarProp
...
+        String databaseDirectory = "/var/lib/pgsql/data";
+        boolean databaseDirectoryOnAttachedDisk = false;
+        if (entitlementService.embeddedDatabaseOnAttachedDiskEnabled(INTERNAL_ACTOR_CRN, ThreadBasedUserCrnProvider.getAccountId())) {
```
lacikaaa, line 78: to reflect on this: https://github.com/hortonworks/cloudbreak/pull/9518/files#r530931639
I think this condition should be false on YCloud and only enable this for real cloud providers

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
keyki: Yes, that would answer my next question in the if condition.

mock-thunderhead/src/main/resources/application.yml
schfeca75: Default value has set to false

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
schfeca75: Using 'False; as default value in the beginning of the sls file

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
schfeca75: It was just a copy-paste error.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/PostgresConfigService.java
schfeca75: Refactored based on what we discussed on slack

orchestrator-salt/src/main/resources/salt/salt/postgresql/init.sls
schfeca75: Default value has set to 'False'

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/EmbeddedDatabaseConfigProvider.java
```diff
@@ -0,0 +1,68 @@
...
+                    POSTGRES_LOG_DIRECTORY_KEY, POSTGRES_DEFAULT_LOG_DIRECTORY,
+                    POSTGRES_DATA_ON_ATTACHED_DISK_KEY, false);
+        }
+        LOGGER.debug("Embedded Postgres sql server pillar parameters: {}", result);
+        return result;
+    }
+
+    private Map<String, Object> createEmbeddedDbOnAttachedDiskConfig(Stack stack) {
+        Template template = stack.getPrimaryGatewayInstance().getInstanceGroup().getTemplate();
+        int volumeCount = template == null ? 1 : template.getVolumeTemplates().stream()
```
keyki, line 61: On YCloud it is going to be 0, right? And will be resolved as /hadoopfs/root1 ?

keyki: I checked and it is correct, but fails anyway since the /hadoopfs/root1 folder does not exist.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/PostgresConfigService.java
schfeca75: Embedded database on attached disk will only be enabled next to the entitlement, if there are attached disks.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/container/postgres/EmbeddedDatabaseConfigProvider.java
schfeca75: Embedded database on attached disk will only be enabled next to the entitlement, if there are attached disks.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/EmbeddedDatabaseService.java
```diff
@@ -1,26 +1,48 @@
...
             return new EmbeddedDatabaseInfo(volumeCount > 0, volumeCount);
         } else {
             return new EmbeddedDatabaseInfo(false, 0);
         }
     }
+
+    public EmbeddedDatabaseInfo getEmbeddedDatabaseInfo(Stack stack) {
+        if (stack.getCluster().getEmbeddedDatabaseOnAttachedDisk()) {
+            int volumeCount = calculateVolumeCountOnGatewayGroup(stack);
+            return new EmbeddedDatabaseInfo(volumeCount > 0, volumeCount);
```
keyki, line 35: This could be simplified by passing the volume count only and calculate the true/false in the constructor. This would reduce the code duplication as well.

core/src/main/resources/schema/app/20201130155647_CB-9580_new_field_on_cluster:_embedded_db_is_configured_on_attached_disk_or_not.sql
```diff
@@ -0,0 +1,9 @@
+-- // CB-9580 new field on cluster: embedded db is configured on attached disk or not
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE cluster ADD COLUMN IF NOT EXISTS embeddeddatabaseonattacheddisk boolean NOT NULL DEFAULT FALSE;
```
keyki, line 4: This is the way to add a new column with a default value: https://dba.stackexchange.com/questions/60965/does-postgresql-optimize-adding-columns-with-non-null-defaults/216153#216153

keyki (approved):

PR #9532, opened 2020-11-26
CB-9762 Scheduled cleanup of CLOSE_WAIT autossh connections

sodre90:
There may be multiple customers affected by the file descriptor leak.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/ccm/scripts/socket-wait-cleanup.sh
+++ b/freeipa/src/main/resources/freeipa-salt/salt/ccm/scripts/socket-wait-cleanup.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+: "${SOCKET_WAIT_LIMIT:=1000}"
+socket_wait_count=$(netstat -nat -p | grep ':9443\|:443' | grep CLOSE_WAIT | wc -l)
+if [[ $socket_wait_count -ge $SOCKET_WAIT_LIMIT ]]; then
+  echo "$(date '+%d/%m/%Y %H:%M:%S') - socket wait count ($socket_wait_count) is higher than the limit ($SOCKET_WAIT_LIMIT)" |& tee -a /var/log/socket_wait_cleanup.log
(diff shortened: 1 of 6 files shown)


freeipa/src/main/resources/freeipa-salt/salt/ccm/scripts/socket-wait-cleanup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+: "${SOCKET_WAIT_LIMIT:=1000}"
+socket_wait_count=$(netstat -nat | egrep 'CLOSE_WAIT' | wc -l)
```
oleewere, line 4: what if an another service will do this (does not close connections properly), is this will restart CCM service?
can we check that as well these sockets are all CCM related in the script (you should be able to match process id against the open connections)?

oleewere (changes requested):

orchestrator-salt/src/main/resources/salt/salt/top.sls
```diff
@@ -11,6 +11,7 @@ base:
     - metering
     - ntp
     - postgresql.root-certs
+    - ccm
```
lacikaaa, line 14: I think we should add the script only if CCM is enabled, so the state should be role dependent. Also only gateway nodes configured for CCM if I remember well, so we should not add this to every node

lacikaaa (approved):

freeipa/src/main/resources/freeipa-salt/salt/ccm/scripts/socket-wait-cleanup.sh
sodre90: PR was updated, please check. Relevant part:
socket_wait_count=$(netstat -nat -p | grep ':9443\|:443' | grep CLOSE_WAIT | wc -l)

biharitomi (approved):

pdarvasi (approved):

oleewere (approved):

PR #9533, opened 2020-11-27
CB-9623 Add WorkloadAnalitycs + Log collection features

akanto:


oleewere:
how will this overlap with AccountTelemetry settings?

akanto:
If the account-level settings are reflected in the environment level settings then they will be logged, otherwise we won't log them

oleewere (approved):

PR #9544, opened 2020-12-01
CB-10055 Fix base image validation: should check the image itself first in case of image id is available in the request.

pkedvessy:


doktoric:
can we create a unit test for the jira ticket problem ? I would like to make sure that this will not be broken in the future.

pkedvessy:
@doktoric Unfortunately the jira case can not be tested directly. I have added a new test what covers the jira case indireclty.

doktoric (approved):

PR #9545, opened 2020-12-01
CB-10029 Azure managed image creation error can potentially lock user…

gergopapi2:
… out from using its single RG or subscription

The problem: when a user wants to use a new image, it needs to be copied to its storage account and then a managed image is created from it. The first step in image creation is to persist an image resource to database with status REQUESTED. If an exception comes during image creation then a polling is done. If after the polling the image is still not present then the exception is re-thrown and the user is notified.
However, unbeknownst to the developers, the poller throws an exception on timeout - the code that would set image status to FAILED is not run any more. An image in REQUESTED state would then block the usage of that image forever, it is a trap state.

Solution: the poller after image creation does not throw an exception on any error but lets the calling code decide what to do.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
@@ -1,6 +1,7 @@
 package com.sequenceiq.cloudbreak.cloud.azure.image;
 
 import java.util.Optional;
+import java.util.concurrent.TimeoutException;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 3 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/image/AzureImageService.java
```diff
@@ -70,14 +72,18 @@ private AzureImage createImageAndNotify(AuthenticatedContext ac, VirtualMachineC
...
+            AzureClient client, AzureManagedImageCreationCheckerContext checkerContext, CloudException originalException) {
         Optional<VirtualMachineCustomImage> customImage;
-        azureManagedImageCreationPoller.startPolling(ac, checkerContext);
+        Optional<Throwable> errorResult = azureManagedImageCreationPoller.startPolling(ac, checkerContext, false);
         customImage = findImage(azureImageInfo, client);
         if (customImage.isEmpty()) {
-            LOGGER.error("Failed to create custom image.", e);
             updateImageStatus(ac, azureImageInfo.getImageNameWithRegion(), azureImageInfo.getImageId(), CommonStatus.FAILED);
-            throw new CloudConnectorException(e);
+            if (errorResult.isPresent() && !(errorResult.get() instanceof TimeoutException)) {
```
pdarvasi, line 81: is there a particular reason why you decided not to catch TimeoutException in a try... catch here?

tiborpopovics (approved):

pdarvasi (approved):
Thanks, this looks the cleanest to me!

PR #9546, opened 2020-12-01
CB-10063 Added validation for CloudStorage Base location. 

lnardai:
- Now request which contain whitespaces will throw BadRequestException.

I'm 90% sure buckets and azure storage accounts can't have whitespaces in them.

Note: turns out S3 folder can have whitespaces

### DIFF
--- a/core/src/test/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroxUpgradeServiceTest.java
+++ b/core/src/test/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroxUpgradeServiceTest.java
@@ -32,8 +32,8 @@ import com.sequenceiq.cloudbreak.auth.JsonCMLicense;
 import com.sequenceiq.cloudbreak.auth.PaywallAccessChecker;
 import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
 import com.sequenceiq.cloudbreak.auth.altus.GrpcUmsClient;
-import com.sequenceiq.cloudbreak.core.flow2.service.ReactorFlowManager;
 import com.sequenceiq.cloudbreak.common.exception.BadRequestException;
+import com.sequenceiq.cloudbreak.core.flow2.service.ReactorFlowManager;
 import com.sequenceiq.cloudbreak.service.StackCommonService;
 import com.sequenceiq.cloudbreak.service.image.ImageChangeDto;
 import com.sequenceiq.cloudbreak.service.stack.StackService;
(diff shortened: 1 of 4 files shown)


lajosrodek:
@lnardai Correct.
* S3 buckets: "Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-)." See https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules .
* Azure storage accounts: "Storage account names must be between 3 and 24 characters in length and may contain numbers and lowercase letters only." See https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview?toc=/azure/storage/blobs/toc.json#naming-storage-accounts .

lajosrodek:
More fun:
* S3 object keys: "The name for a key is a sequence of Unicode characters whose UTF-8 encoding is at most 1,024 bytes long. ... You can use any UTF-8 character in an object key name." See https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys .
* Azure Storage Service:
  * Containers: "Container names must start or end with a letter or number, and can contain only letters, numbers, and the dash (-) character. ... All letters in a container name must be lowercase." See https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-and-Referencing-Containers--Blobs--and-Metadata#container-names .
  * ADLSg2 file systems: "The filesystem identifier. The value must start and end with a letter or number and must contain only letters, numbers, and the dash (-) character. ... All letters must be lowercase. The value must have between 3 and 63 characters." So essentially the same as for containers. See https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/filesystem/create .
  * Blobs: "A blob name can contain any combination of characters. ... A blob name must be at least one character long and cannot be more than 1,024 characters long, for blobs in Azure Storage." See https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-and-Referencing-Containers--Blobs--and-Metadata#blob-names .
  * ADLSg2 paths (directories / files): No explicit info; probably same as for blob names.

lnardai:
@lajosrodek 
Thank you for the confirmation.
Although what you've written is correct, I've just checked on AWS and sadly it's possible to create Folder with whitespaces, for e.g.: `s3://cb-group/lnardai/lnardai example/`
And since we are asking for a full path I can't reject all strings that contain a whitespace, instead I went with stripping the string to avoid the errors mentioned in the ticket

horadla23 (approved):

PR #9551, opened 2020-12-01
CFM-1555 We need to enable file provider, since otherwise it won’t write out that block in the resulting file. Also without this change, NiFi and NiFi Registry failed to start

vbrodetskyi:
After previous changes committed in scope of CFM-1555, we need to set "xml.authorizers.userGroupProvider.file-user-group-provider.enabled"="true" for the file provider, since otherwise it won’t write out that block in the resulting file. Also without this change, NiFi and NiFi Registry throws such exception: 
"AuthorizerCreationException: Unable to locate the Configurable User Group Provider: file-user-group-provider"

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.7/cdp-flow-management-small.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.7/cdp-flow-management-small.bp
@@ -50,7 +50,7 @@
             "configs": [
               {
                 "name": "xml.authorizers.userGroupProvider.file-user-group-provider.enabled",
-                "value": "false"
+                "value": "true"
               },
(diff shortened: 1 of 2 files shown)


lajosrodek (approved):
LGTM

doktoric:
@vbrodetskyi we are trying to improve our commit messages so I would like to ask you to add more details into the commit message.
Why we need to change?
What is affected and how?

bbende (approved):
+1

vbrodetskyi:
@lajosrodek @doktoric if you have a chance, could you please merge these changes?

doktoric (approved):

PR #9552, opened 2020-12-01
CB-10087 - InstanceGroup conversion does not copy the original type

gregito:
CB-10087 - InstanceGroup conversion does not copy the original type, so added this operation alongside with the associated test case


bergerdenes (approved):
lgtm

PR #9554, opened 2020-12-02
CB-10085 adding AccountIdAwareResource as the base class of AccountAw…

doktoric:
…areResource. Secret aspect will be able to read the account information from any object with this change. Currently secret ascpect try to use AccountIdAwareResource which iswas removed eas part of the structured event effort. Vault pathes will be fixed and we will not point to undefined path. As a followup I will remove the 'undefined' logic from secret aspect because it is risky and hiding potential problems.


biharitomi (approved):
LGTM

PR #9556, opened 2020-12-02
CB-10085

doktoric:

### DIFF
--- a/freeipa/src/main/resources/schema/app/20201203102940_CB-10140_add_accountId_to_freeipa_SaltSecurityConfig.sql
+++ b/freeipa/src/main/resources/schema/app/20201203102940_CB-10140_add_accountId_to_freeipa_SaltSecurityConfig.sql
@@ -1,0 +1,12 @@
+-- // CB-10140 add accountId to freeipa SaltSecurityConfig
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE saltsecurityconfig ADD COLUMN IF NOT EXISTS accountid VARCHAR(255);
+ALTER TABLE saltsecurityconfig ALTER COLUMN accountid SET DEFAULT NULL;
+UPDATE saltsecurityconfig SET accountid = (SELECT accountid FROM securityconfig WHERE securityconfig.saltsecurityconfig_id = saltsecurityconfig.id) WHERE accountid IS NULL;
+
+-- //@UNDO
+-- SQL to undo the change goes here.
+
+ALTER TABLE saltsecurityconfig DROP COLUMN IF EXISTS accountid;
+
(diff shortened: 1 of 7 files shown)


biharitomi (approved):

secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/SecretAspectService.java
```diff
@@ -52,7 +52,7 @@ public Object proceedSave(ProceedingJoinPoint proceedingJoinPoint) {
                     }
                 }
             } catch (IllegalArgumentException e) {
-                LOGGER.error("Given entity isn't instance of TenantAwareResource. Secret is not deleted!", e);
+                LOGGER.error("Given entity isn't instance of AccountIdAwareResource. Secret is not deleted!", e);
```
lajosrodek, line 55: Nit: Perhaps replace the hardwired class name with `AccountIdAwareResource.class.getSimpleName()` like a few lines below in `findAccountId()`.

secret-engine/src/main/java/com/sequenceiq/cloudbreak/service/secret/service/SecretAspectService.java
```diff
@@ -92,7 +92,7 @@ public Object proceedDelete(ProceedingJoinPoint proceedingJoinPoint) {
...
```
lajosrodek, line 95: Same here.

keyki (approved):

freeipa/src/main/resources/schema/app/20201203102940_CB-10140_add_accountId_to_freeipa_SaltSecurityConfig.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-10140 add accountId to freeipa SaltSecurityConfig
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE saltsecurityconfig ADD COLUMN IF NOT EXISTS accountid VARCHAR(255) DEFAULT 'undefined';
```
keyki, line 4: Default value shouldn't be added like this. It will block the whole table. 
See https://dba.stackexchange.com/questions/60965/does-postgresql-optimize-adding-columns-with-non-null-defaults/216153#216153
Example: https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/schema/app/20201130155647_CB-9580_new_field_on_cluster:_embedded_db_is_configured_on_attached_disk_or_not.sql

Also I think we should set it to null.

doktoric: done

keyki (approved):

PR #9560, opened 2020-12-02
CB-10021 fixing nodeCount in user saved cluster definitions, previous…

afarsang:
…ly it was always 0


doktoric (approved):
thanks for doing and thanks for the test

PR #9562, opened 2020-12-02
CB-10072 Basic CRN resource type validation

horadla23:
- this commit ensures that resources can be created with predefined resourcetype/servictype pair, also some parts of CRN should not be configurable (like partition, region).
- checking for resourcetype/servicetype pair has been added to ValidCrn annotation to avoid situations like calling environment related endpoint with datalake crn.
- introduced new ValidCrnCollection annotation for crn collections in APIs
- adding more restrictive ValidCrn/ValidCrncollection to some environment APIs calls and freeipa request objects
 


horadla23:
@doktoric please add someone from your team to check environment related code if you cannot check this
@keyki i would like to ask you for the same regarding freeipa code

auth-internal/src/main/java/com/sequenceiq/cloudbreak/auth/security/internal/InternalCrnModifier.java
```diff
@@ -63,12 +63,7 @@ public Object changeInternalCrn(ProceedingJoinPoint proceedingJoinPoint) {
 
     private String getAccountIdModifiedCrn(String userCrnString, String accountId) {
         Crn userCrn = Crn.fromString(userCrnString);
-        Crn newUserCrn = Crn.builder()
-                .setService(userCrn.getService())
-                .setAccountId(accountId)
-                .setResourceType(userCrn.getResourceType())
-                .setResource(userCrn.getResource())
-                .build();
+        Crn newUserCrn = Crn.copyWithDifferentAccountId(userCrn, accountId);
```
lnardai, line 66: looks dangerous :)

horadla23: what do you mean by that? it does the same as the current code

lnardai: Yes, I know, but that logic was private to that InternalCrnModifier implementation, while the static Crn is used widely in the code. I'm just contemplating wether it's worth to expose that.

lnardai:
Can you add a simple mock test which has at least 2 CRN's that are invalid and assert the error message return on the API?

auth-internal/src/main/java/com/sequenceiq/cloudbreak/auth/security/internal/InternalCrnModifier.java
horadla23: setAccountId in Crn.builder() was not private, I mean right now everyone can change accountId If he/she want to do so. one solution can be to make the method package private or something like that but InternalCrnModifier and Crn placed in different module. We can move one of them but it can cause other issues.

horadla23:
> Can you add a simple mock test which has at least 2 CRN's that are invalid and assert the error message return on the API?
> Just to make sure no one breaks the nice error message functionality.

sure

also will enhance error message as discussed

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/CrnResourceDescriptor.java
```diff
@@ -0,0 +1,57 @@
...
+    ROLE(Crn.ResourceType.ROLE, Crn.Service.IAM),
+    RESOURCE_ROLE(Crn.ResourceType.RESOURCE_ROLE, Crn.Service.IAM),
+    ALERT(Crn.ResourceType.DATAHUB_AUTOSCALE_CONFIG, Crn.Service.AUTOSCALE),
+    DATABASE(Crn.ResourceType.DATABASE, Crn.Service.REDBEAMS),
+    DATABASE_SERVER(Crn.ResourceType.DATABASE_SERVER, Crn.Service.REDBEAMS),
+    USER(Crn.ResourceType.USER, Crn.Service.IAM),
+    PUBLIC_KEY(Crn.ResourceType.PUBLIC_KEY, Crn.Service.IAM),
+    MACHINE_USER(Crn.ResourceType.MACHINE_USER, Crn.Service.IAM),
+    GROUP(Crn.ResourceType.GROUP, Crn.Service.IAM),
+    POLICY(Crn.ResourceType.POLICY, Crn.Service.IAM);
```
lacikaaa, line 30: just a stupid idea: I think it would be easier to read through the list if it would be grouped by services and the items would be alphabetically ordered inside the groups.
More nice way would be if the items would be in separate file by service, but that maybe would cause more complexity than benefit.

common/src/main/java/com/sequenceiq/cloudbreak/validation/CrnCollectionValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private CrnResourceDescriptor[] resourceDescriptors;
+
+    @Override
+    public void initialize(ValidCrnCollection constraintAnnotation) {
+        resourceDescriptors = constraintAnnotation.resource();
+    }
+
+    @Override
+    public boolean isValid(Collection<String> req, ConstraintValidatorContext constraintValidatorContext) {
```
lacikaaa, line 27: could you break this into smaller methods?

common/src/main/java/com/sequenceiq/cloudbreak/validation/CrnValidator.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.validation.ConstraintValidatorContext;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
+
+import com.google.common.base.Joiner;
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.altus.CrnResourceDescriptor;
+
+public class CrnValidator implements ConstraintValidator<ValidCrn, String> {
```
lacikaaa, line 17: I feel like there is a lot of code duplication here with `CrnCollectionValidator`. Could you move the common parts into a common class?

horadla23: yes, makes sense

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/CrnResourceDescriptor.java
horadla23: yes, grouping seems worth to implement, separate files can be problematic indeed, since services are referring to other services' crns

lacikaaa (approved):
LGTM, I checked only the modifications

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -138,7 +141,7 @@ DatabaseServerV4Response deleteByName(
     @ApiOperation(value = DatabaseServerOpDescription.DELETE_MULTIPLE_BY_CRN, notes = DatabaseServerNotes.DELETE_MULTIPLE_BY_CRN,
             consumes = MediaType.APPLICATION_JSON, nickname = "deleteMultipleDatabaseServersByCrn")
     DatabaseServerV4Responses deleteMultiple(
-            @ApiParam(DatabaseServerParamDescriptions.CRNS) Set<@ValidCrn String> crns,
+            @ApiParam(DatabaseServerParamDescriptions.CRNS) @ValidCrn(resource = DATABASE_SERVER) Set<String> crns,
```
gergopapi2, line 144: Question: does the Crn validation work with a collection of Crns as well?

horadla23: have you checked the change? anyway there is a new mock test for this to make sure

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/EnvironmentTest.java
```diff
@@ -91,6 +94,38 @@ public void testDeleteEnvironment(MockedTestContext testContext) {
...
+                        RunningParameter.expectedMessage(".*Crn provided: " +
+                                "crn:cdp:datalake:us-west-1:acc:datalake:dl has invalid resource type or service type. " +
+                                "Accepted service type / resource type pairs: [(]environments,environment[)].*")
+                                .withKey("EnvironmentGetByCrnAction"))
+                .when(environmentTestClient.deleteMultipleByCrns(invalidCrn, otherInvalidCrn))
+                .expect(BadRequestException.class,
+                        RunningParameter.expectedMessage(".*Crns provided: \\[crn:cdp:datalake:us-west-1:acc:datalake:dl," +
+                                "crn:cdp:datahub:us-west-1:acc:cluster:dh\\] have invalid resource type or service type. " +
+                                "Accepted service type / resource type pairs: [(]environments,environment[)].*")
+                                .withKey("EnvironmentDeleteMultipleByCrnsAction"))
```
horadla23, line 125: @gergopapi2

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
horadla23: sorry, so yes, it is also working with collection of crns

lnardai (approved):

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
```diff
@@ -141,7 +143,7 @@ public EnvironmentCrnResponse getCrnByName(@ResourceName String environmentName)
 
     @Override
     @CheckPermissionByResourceCrn(action = AuthorizationResourceAction.DESCRIBE_ENVIRONMENT)
-    public DetailedEnvironmentResponse getByCrn(@ResourceCrn @TenantAwareParam String crn) {
+    public DetailedEnvironmentResponse getByCrn(@ValidCrn(resource = CrnResourceDescriptor.ENVIRONMENT) @ResourceCrn @TenantAwareParam String crn) {
```
doktoric, line 146: static import ? we used it in EnvironmentEndpoint.java

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
```diff
@@ -591,32 +600,45 @@ public static Builder builder() {
...
-        public Builder setAccountId(String accountId) {
-            this.accountId = checkNotNull(accountId);
+        public Builder setResource(String resource) {
+            this.resource = checkNotNull(resource);
             return this;
         }
 
-        public Builder setResourceType(ResourceType resourceType) {
-            this.resourceType = checkNotNull(resourceType);
+        public Crn build(CrnResourceDescriptor resourceDescriptor) {
```
topolyai5, line 622: is the resourceDescriptor required or why do you add to the param list?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/client/EnvironmentTestClient.java
```diff
@@ -36,6 +38,10 @@
         return new EnvironmentGetAction();
     }
 
+    public Action<EnvironmentTestDto, EnvironmentClient> describeByCrn() {
```
topolyai5, line 41: could you please change for this: `EnvironmentInternalGetAction` or remove one of them?

topolyai5 (changes requested):

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
horadla23: yes, required

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/client/EnvironmentTestClient.java
horadla23: why internal? both of them are worth to test and both of them can be used by anyone

horadla23: ohh, i see now, but my method is not using internalcrn, and that's the real use case in my test

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
topolyai5: As I know, almost all of the fields are required in the crn. Is it a good pattern to add the required fields as a parameter to the build() method?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/client/EnvironmentTestClient.java
topolyai5: Yes, you are right. Ok, it will be ok

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crn.java
horadla23: all fields are required, but some of them have 1 value always, some of them changed to package private in this change and alo crnresourcedescriptor is not part of the crn, so it is not a required field, but helps to fill in some fields of the crn, so it i decided to add to build() as parameter

horadla23: i can also add a new constructor for builder if it is better, instead of adding a parameter to build(), what do you think @topolyai5 ?

topolyai5 (approved):

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
horadla23: because there were conflicts between static imports (CrnResourceDescriptor.ENVIRONMENT vs CredentialType.ENVIRONMENT) i decided to not use static import in case of CrnResourceDescriptor

PR #9565, opened 2020-12-02
CB-10113 Add EnvironmentRequested and EnvironmentStateChanged events …

akanto:
…to usage.proto in order to track how many environments have been created by our customers. We are tracking major features of the environments like regions, availability zones, network settings etc


cegganesh84 (approved):
LGTM

PR #9572, opened 2020-12-04
CB-9123 URL configurable in the integration test. Supported: how many…

topolyai5:
Evaluate the response by URI an http method if it appears in the pre-defined response list. The order of the responses matter with the same path and method. (Check the unit tests)
Evaluate the times:
 * 0: will be permanently in the list, the next ones in the list with the same path and method will be never evaluate
 * 1...n: evaluate n times. If reach the n-th, the response will be removed from the list of path.

Evaluate the response or exception
 * Only if the the status code is 200 return with the response.
 * If the status code is null, throw a NotFound status.


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/mock/answer/DefaultResponseConfigure.java
```diff
@@ -0,0 +1,118 @@
...
+    public Method getMethod() {
+        return method;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public T clearDefinedResponses() {
+//        mock.getDynamicRouteStack().clear(method.getHttpMethod(), path);
```
attilapalfi92, line 41: Why so many commented code lines?

topolyai5: this functionality is disabled temporary. I split the epic to multiple parts but I don't want to remove the old codes. I will re-enable these lines if I can.

biharitomi (approved):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/mock/endpoint/ClouderaManagerEndpoints.java
```diff
@@ -1,157 +1,201 @@
...
+
+    public Commands<T> commands() {
+        return (Commands<T>) EndpointProxyFactory.create(Commands.class, testDto, executeQueryToMockInfrastructure);
+    }
+
+    public ClouderaManagerHosts<T> clouderaManagerHosts() {
+        return (ClouderaManagerHosts<T>) EndpointProxyFactory.create(ClouderaManagerHosts.class, testDto, executeQueryToMockInfrastructure);
     }
 
     @SparkUri(url = API_ROOT + "/cm/importClusterTemplate", requestType = ApiClusterTemplate.class)
```
afarsang, line 51: If sparkservice eliminated we might consider to rename this annotation,or even yield this from a swagger doc (if I recall correctly mocks are generated based on swagger)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/EnvironmentEditTest.java
```diff
@@ -50,39 +51,37 @@ protected void setupTest(TestContext testContext) {
...
     public void authenticationEditWhenSetExistingKeyAndDeleteManagedSuccessfully(MockedTestContext testContext) {
+        String randomPublicKeyId = UUID.randomUUID().toString();
         testContext
-                .given(HttpMock.class).whenRequested(SpiEndpoints.RegisterPublicKey.class).post()
-                .thenReturn((s, uriParameters) -> "")
-                .whenRequested(SpiEndpoints.GetPublicKey.class).get()
-                .pathVariable("publicKeyId", "id")
-                .thenReturn((s, uriParameters) -> true)
+                .given(HttpMock.class)
+                .mockSpi().getPublicKey().get()
```
afarsang, line 57: Honestly I am missing the original keywords here (given when then) a bit.
would it make sense call this as 'whenMockSpiRequestedWith()' or something similar?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/EnvironmentEditTest.java
```diff
@@ -50,39 +51,37 @@ protected void setupTest(TestContext testContext) {
...
                     }
                     if (publicKey != null) {
                         throw new TestFailException("The auth public key should be null");
                     }
                     return t;
                 })
-                .given(HttpMock.class).whenRequested(SpiEndpoints.UnregisterPublicKey.class).post().verify(CheckCount.times(1))
-                .given(HttpMock.class).whenRequested(SpiEndpoints.RegisterPublicKey.class).post().verify(CheckCount.times(0))
+                .mockSpi().unregisterPublicKey().post().times(1).verify()
+                .mockSpi().unregisterPublicKey().post().times(0).verify()
```
afarsang, line 84: I guess this line wanted to be 'registerPublikKey()' verification

afarsang (changes requested):

PR #9575, opened 2020-12-04
CB-5472 remove unused parcel files before upgrade

keyki:
Unfortunately, the parcel files are large and they are even
replicated across multiple folders. Most of these folders
are used by intermediate processes like the distribution, but
they are not used after that point. Before the next upgrade we
need to clean up these folders, otherwise we run out of free space
on the root disk. Eventually, the usage of the folders should be
fixed in CM and these are just workarounds.


pdarvasi (approved):

keyki:
I swear it was green. I just changed the commit message and added some comments to the salt states :)

PR #9577, opened 2020-12-04
CB-10157 Store the original tags in our services and transform GCP la…

biharitomi:
…bels only in cloud-gcp module. This is necessary because we stored the label transformed tags previously. This way we can get the original tags in the pillar which are fundamental from telemtry perspective

@oleewere the pillars will look like this:
`cat /srv/pillar/tags/init.sls
#!json
{
	"tags": {
		"Cloudera-Creator-Resource-Name": "crn:cdp:iam:us-west-1:cloudera:user:tbihari",
		"Cloudera-Environment-Resource-Name": "crn:cdp:environments:us-west-1:cloudera:environment:7d322e9f-5085-4154-80f1-effafa52ac01",
		"Cloudera-Resource-Name": "crn:cdp:datahub:us-west-1:cloudera:cluster:d5cc4415-7a30-4760-aa23-d9981c28c47d",
		"Cloudera-Service-Type": "DATAHUB",
		"Owner": "tbihari@ums.mock",
		"creation-timestamp": "1607345447",
		"owner": "tbihari@ums.mock"
	}
}`

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/util/GcpLabelUtil.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/util/GcpLabelUtil.java
@@ -1,0 +1,50 @@
+package com.sequenceiq.cloudbreak.cloud.gcp.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+
+public final class GcpLabelUtil {
+    private static final Logger LOGGER = LoggerFactory.getLogger(GcpLabelUtil.class);
+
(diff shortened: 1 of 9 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/util/GcpLabelUtil.java
```diff
@@ -0,0 +1,44 @@
...
+        if (Crn.isCrn(value)) {
+            try {
+                Crn crn = Crn.fromString(value);
+                value = crn == null ? value : crn.getResource();
+            } catch (Exception e) {
+                LOGGER.debug("Ignoring CRN ({}) parse error during tag value generation : {}", value, e.getMessage());
+            }
+        }
+        String sanitized = value.split("@")[0].toLowerCase().replaceAll("[^\\w]", "-");
+        return StringUtils.right(sanitized, GCP_MAX_TAG_LEN);
```
doktoric, line 42: I think it would be beneficial to log the generated value

oleewere (approved):
lgtm, thanks

cegganesh84 (commented):
Nice, thanks for doing this.

PR #9579, opened 2020-12-04
CB-10171: Add templates with bursty EC2 types

LinaAtAustin:
The new burst templates are copied from original templates of the same version, then change the EC2 type from m5 family to the corresponding t3a family to save cost.

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterShape.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterShape.java
@@ -3,5 +3,7 @@ package com.sequenceiq.sdx.api.model;
 public enum SdxClusterShape {
     CUSTOM,
     LIGHT_DUTY,
-    MEDIUM_DUTY_HA
+    LIGHT_DUTY_BURST,
+    MEDIUM_DUTY_HA,
+    MEDIUM_DUTY_HA_BURST
 }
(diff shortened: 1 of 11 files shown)


kmanamcheri:
How about implementing an "import" functionality in the template. Right now, we have so many copies of the templates which are 90% similar! Could we implement this in a way like we can say

```
{
  "cluster": {
    "blueprintName": "CDP 1.2 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas",
    "validateBlueprint": false
  },
  "customDomain": {
    "domainName": "cloudera.site",
    "hostgroupNameAsHostname": true
  },
  "instanceGroups": [
    {{ import idbroker from a different template file }},
    {
      "name": "master",
      "template": {
        "instanceType": "t3a.2xlarge",
        "attachedVolumes": [
          {
            "count": 1,
            "size": 250,
            "type": "standard"
          }
        ],
        "rootVolume": {
          "size": 100
        }
      },
      "nodeCount": 1,
      "type": "GATEWAY",
      "recoveryMode": "MANUAL",
      "recipeNames": []
    }
  ]
}
```

kmanamcheri:
Would it be simpler to roll this only for one version (7.2.7 maybe?), so we can test it clearly and gather information about if this is working effectively.

LinaAtAustin:
Although the contents are very similar, getting configuration in the template directly allows the flexibility of changing individual template in the future. When we introduce more logic to change from input, it is much harder for the user to know what they get and it is also harder to support.

LinaAtAustin:
"Would it be simpler to roll this only for one version (7.2.7 maybe?), so we can test it clearly and gather information about if this is working effectively."
I don't see big benefits of holding up on other runtimes. If the behavior of UI is to list all available templates, then the user is free to choose for each environment creation. If an EC2 type does not work for a given runtime (which I think it is unlikely as we have changed SDX setup script for more than a month), the user can use the original template and is not blocked. 

Updating templates for all runtimes allow us to discover issues sooner.

doktoric:
do we still need this?

PR #9581, opened 2020-12-04
CB-10160 - Increase socket timeout and update maxFailoverAttempts=0 for NiFi service

moresandeep:
_This is for a hotfix issue CB-10160_
NiFi team ran into issues with a customer with large uploads. This issue is identified as a critical fix and needs to land as quick as possible (hotfix).

The changes proposed in this PR are:

increase the socket timeout for Nifi
set retry and failover counts to zero
**Testing**
The patch was tested on a cluster running NiFi

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -318,13 +318,13 @@
              {% if 'NIFI' in exposed and 'NIFI_NODE' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>NIFI</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enabled=true;maxFailoverAttempts=0;failoverSleep=1000</value>
              </param>
              {%- endif %}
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -636,6 +636,14 @@
     <service>
         <role>NIFI</role>
         <param name="useTwoWaySsl" value="true" />
+        <param>
+            <name>httpclient.connectionTimeout</name>
+            <value>2m</value>
```
akanto, line 641: Don't we have a 60 sec timeout on nginx anyway?

keyki: I think @biharitomi increased it recently, although I'm not sure on which branch.

biharitomi: I think @akanto is right, I only increased the cluster proxy client and NGiNX timeouts.

biharitomi: But we CB communicate with CM directly through NGiNX on port 9443 and the Knox timeout should not be increased for the deploy client config calls at least.

PR #9563, opened 2020-12-02
CB-9525 DB SSL root cert file location customization

lajosrodek:
* Extract DB SSL root cert file name `/hadoopfs/fs1/database-cacerts/certs.pem` as the new property `cb.externaldatabase.ssl.rootcerts.path` in `core/src/main/resources/application.yml`
* Remove all other former occurrences of `/hadoopfs/fs1/database-cacerts/certs.pem` and let them be populated from the above application property (for Java) and the new pillar property `postgres_root_certs:ssl_certs_file_path` (for Salt)
  * Note: `RedbeamsDbCertificateProvider.getSslCertsFilePath()` depends on neither the `Stack` nor the `Cluster` at the moment. Conditional calculation of the file path may be added later as needed.
* Introduce a new Python dict for DB SSL settings; see `orchestrator-salt/src/main/resources/salt/salt/postgresql/settings.sls`
* Add new unit tests and adapt / migrate existing ones to JUnit 5 & AspectJ
  * Use `org.junit.jupiter.api.Assertions.assertThrows()` instead of `org.assertj.core.api.Assertions.assertThatCode()` for better readability.

PR #9587, opened 2020-12-05
CB-9526 Use SSL with cert&host verification for Salt external DB create

lajosrodek:
* When the external DB server has been provisioned with SSL enforcement (#9215), DB create operations shall always mandate SSL with DB server cert and host name verification.
* Depends on #9563.


lajosrodek:
Base branch setting `CB-9525_DB_root_cert_file_location_customization` is temporary; it will have to be changed to `master` once #9563 gets merged in. The `DO_NOT_MERGE` label can be only removed if the `master` rebase has been done.

doktoric (approved):
seems ok

biharitomi (approved):
LGTM

PR #9588, opened 2020-12-05
Add comment related to cdpcp-2612

zankhna-randeri:
We started considering few group names to be reserved on CDP UMS-service. Those reserved group names also contain IPA protected groups taken from this file. Therefore, whenever IPA protected groups are updated, we need to update reserved group names in UMS as well. This commit adds a comment to track that.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaChecks.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaChecks.java
@@ -4,15 +4,15 @@ import java.util.List;
 import java.util.function.Supplier;
 
 public class FreeIpaChecks {
-    /* Users in this list should not be added/deleted. */
+    /* Users in this list should not be added to or deleted from FreeIPA. */
     public static final List<String> IPA_PROTECTED_USERS = List.of("admin");
 
-    /* Groups in this list should not be added/deleted. These groups are considered to be reserved in CDP-UMS.
-    Therefore,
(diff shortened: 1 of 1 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaChecks.java
```diff
@@ -7,7 +7,9 @@
     /* Users in this list should not be added/deleted. */
     public static final List<String> IPA_PROTECTED_USERS = List.of("admin");
 
-    /* Groups in this list should not be added/deleted. */
+    /* Groups in this list should not be added/deleted. These groups are considered to be reserved in CDP-UMS.
```
handavid, line 10: This comment is now ambiguous because we are now also referring to the UMS (instead of just FreeIPA). Can you update these comments to say that "Groups in this list should not be added to or deleted from FreeIPA".
Also change the above comment to "Users in this list should not be added to or deleted from FreeIPA"
Also change the below comment to "Group membership in this groups should not be explicitly changed in FreeIPA"

zankhna-randeri: Done.

handavid:
@zankhna-randeri can you rebase on master and squash? I suspect that the test failures due to unrelated changes. also, this repo is set up for "rebase and merge", not "squash and merge" (I think).

handavid (approved):

PR #9593, opened 2020-12-07
CB-9943: Extend the Data Hub CDH runtime upgrade process with upgrading the non-CDH services to a new version what we burned into the new image.

tiborpopovics:
In this commit we've introduced some changes in connection with Data Hub upgrade:
- CSD downloading: To upgrade a non-CDH service may require the same version of the CSD what the new version is. By default the old CSD files are already present in the image, therefore we need to download the new ones. This process is handled by a shell script that we add to the host at the beginning of the upgrade process.
  In the case of Data Lake upgrade, we don't need to download the new CSD files because we're not upgrading the non-CDH services.

- Upgrading the non-CDH services (Spark, Nifi, etc...) has a different flow than the regular CDH runtime upgrade. We need to download, distribute, and activate the new parcels, then restart the affected services. There is no upgrade command to call.
  After the non-CDH services have been upgraded the flow continues with upgrading the CDH runtime. This whole process does not affect to the Data Lake upgrade flow. We just upgrade only the CDH version like before.

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
@@ -5,11 +5,11 @@ import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUD
 import static com.sequenceiq.cloudbreak.event.ResourceEvent.CLUSTER_CM_CLUSTER_SERVICES_STARTED;
 import static com.sequenceiq.cloudbreak.event.ResourceEvent.CLUSTER_CM_CLUSTER_SERVICES_STARTING;
 import static com.sequenceiq.cloudbreak.polling.PollingResult.isExited;
-import static com.sequenceiq.cloudbreak.polling.PollingResult.isTimeout;
 import static com.sequenceiq.cloudbreak.util.Benchmark.checkedMeasure;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -35,11 +35,8 @@ import com.cloudera.api.swagger.ParcelsResourceApi;
 import com.cloudera.api.swagger.ServicesResourceApi;
 import com.cloudera.api.swagger.client.ApiClient;
 import com.cloudera.api.swagger.client.ApiException;
-import com.cloudera.api.swagger.model.ApiCdhUpgradeArgs;
 import com.cloudera.api.swagger.model.ApiCommand;
 import com.cloudera.api.swagger.model.ApiCommandList;
-import com.cloudera.api.swagger.model.ApiConfig;
-import com.cloudera.api.swagger.model.ApiConfigList;
 import com.cloudera.api.swagger.model.ApiConfigStalenessStatus;
 import com.cloudera.api.swagger.model.ApiHost;
 import com.cloudera.api.swagger.model.ApiHostRef;
@@ -58,9 +55,10 @@ import com.sequenceiq.cloudbreak.cluster.service.ClusterClientInitException;
 import com.sequenceiq.cloudbreak.cm.client.ClouderaManagerApiClientProvider;
 import com.sequenceiq.cloudbreak.cm.client.ClouderaManagerClientInitException;
 import com.sequenceiq.cloudbreak.cm.client.retry.ClouderaManagerApiFactory;
-import com.sequenceiq.cloudbreak.cm.model.ParcelResource;
 import com.sequenceiq.cloudbreak.cm.model.ParcelStatus;
 import com.sequenceiq.cloudbreak.cm.polling.ClouderaManagerPollingServiceProvider;
+import com.sequenceiq.cloudbreak.cm.polling.PollingResultErrorHandler;
+import com.sequenceiq.cloudbreak.common.type.ComponentType;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.Cluster;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.ClusterComponent;
@@ -106,7 +104,16 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
     private ClouderaManagerConfigService configService;
 
     @Inject
-    private ClouderaManagerParcelService clouderaManagerParcelService;
+    private ClouderaManagerParcelDecommissionService clouderaManagerParcelDecommissionService;
+
+    @Inject
+    private ClouderaManagerParcelManagementService clouderaManagerParcelManagementService;
+
+    @Inject
+    private ClouderaManagerUpgradeService clouderaManagerUpgradeService;
+
+    @Inject
+    private PollingResultErrorHandler pollingResultErrorHandler;
 
     private final Stack stack;
 
@@ -142,11 +149,11 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
                 List<ApiHost> hosts = clouderaManagerApiFactory.getHostsResourceApi(apiClient).readHosts(null, null, SUMMARY).getItems();
                 ApiHostRefList body = createUpscaledHostRefList(upscaleHostNames, hosts);
                 clustersResourceApi.addHosts(clusterName, body);
-                activateParcel(clustersResourceApi);
+                activateParcels(clustersResourceApi);
                 applyHostGroupRolesOnUpscaledHosts(body, hostGroup.getName());
             } else {
                 redistributeParcelsForRecovery();
-                activateParcel(clustersResourceApi);
+                activateParcels(clustersResourceApi);
                 clouderaManagerRoleRefreshService.refreshClusterRoles(apiClient, stack);
             }
             return instanceMetaDatas.stream().map(InstanceMetaData::getDiscoveryFQDN).collect(Collectors.toList());
@@ -157,7 +164,7 @@ public class ClouderaManagerModificationService implements ClusterModificationSe
     }
 
     @Override
-    public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
+    public void cleanupCluster(Telemetry telemetry) {
         if (telemetry != null && telemetry.getWorkloadAnalytics() != null) {
             if (StackType.DATALAKE.equals(stack.getType())) {
                 LOGGER.info("Stack type is datalake, no need for WA cleanup");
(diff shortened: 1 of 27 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
+            LOGGER.info("Could not upgrade Cloudera Runtime services", e);
+            throw new ClouderaManagerOperationFailedException(e.getMessage(), e);
+        }
+    }
+
+    private void checkParcelApiAvailability() throws CloudbreakException {
+        clouderaManagerParcelManagementService.checkParcelApiAvailability(stack, apiClient);
+    }
+
+    private void setParcelRepo(Set<ClouderaManagerProduct> products, ClouderaManagerResourceApi clouderaManagerResourceApi) throws ApiException {
```
keyki, line 257: We should call the method argument to services as well. We're mixing the product and services variables and it's hard to track.

keyki:
Please add a proper commit message. This is a really complex change that requires a thorough explanation of what is changing and why.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
             MgmtServiceResourceApi mgmtServiceResourceApi = clouderaManagerApiFactory.getMgmtServiceResourceApi(apiClient);
             ClouderaManagerResourceApi clouderaManagerResourceApi = clouderaManagerApiFactory.getClouderaManagerResourceApi(apiClient);
 
-            startClouderaManager(stack, apiClient);
+            startClouderaManager();
             checkParcelApiAvailability();
-            setParcelRepo(stackProductParcel, clouderaManagerResourceApi);
+
+            Set<ClouderaManagerProduct> services = getServices(components);
+            ClouderaManagerProduct cdhService = getCdhService(services);
```
keyki, line 186: This line can be moved to the else branch as it's only used there.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
+                .collect(Collectors.toSet());
+    }
+
+    private void upgradeNonCdhServices(Set<ClouderaManagerProduct> services, String cdhServiceName, ParcelResourceApi parcelResourceApi)
+            throws CloudbreakException, ApiException {
+        Set<ClouderaManagerProduct> nonCdhServices = getNonCdhServices(services, cdhServiceName);
+        installServices(nonCdhServices, parcelResourceApi);
+    }
+
+    private void installServices(Set<ClouderaManagerProduct> products, ParcelResourceApi parcelResourceApi) throws ApiException, CloudbreakException {
```
keyki, line 225: We should figure out a better method name. Maybe `downloadAndActivateParcels` ?

keyki, line 225: Similarly use services instead of products.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
+        downloadParcel(products, parcelResourceApi);
+        distributeParcel(products, parcelResourceApi);
+        activateParcel(products, parcelResourceApi);
+    }
+
+    private ClouderaManagerProduct getCdhService(Set<ClouderaManagerProduct> products) {
```
keyki, line 231: services

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
+        return products.stream()
+                .filter(product -> product.getName().equals(com.sequenceiq.cloudbreak.cloud.model.component.StackType.CDH.name()))
                 .findFirst()
                 .orElseThrow(() -> new NotFoundException("Runtime component not found!"));
     }
 
+    private Set<ClouderaManagerProduct> getNonCdhServices(Set<ClouderaManagerProduct> products, String cdhProductName) {
```
keyki, line 238: services

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
keyki, line 257: services

orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/csd-downloader.sh
```diff
@@ -3,11 +3,14 @@
 set -ex
 
 mkdir -p /opt/cloudera/csd
+rm -f /opt/cloudera/csd/*
 cd /opt/cloudera/csd
 
 {% if salt['pillar.get']('cloudera-manager:csd-urls') %}
 {% for url in salt['pillar.get']('cloudera-manager:csd-urls') %}
+{% if salt['pillar.get']('cloudera-manager:csd-urls') %}
```
keyki, line 11: What is this extra if condition for? There should be a test -f here to avoid downloading the csd with every highstate.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
             ParcelResourceApi parcelResourceApi = clouderaManagerApiFactory.getParcelResourceApi(apiClient);
             MgmtServiceResourceApi mgmtServiceResourceApi = clouderaManagerApiFactory.getMgmtServiceResourceApi(apiClient);
             ClouderaManagerResourceApi clouderaManagerResourceApi = clouderaManagerApiFactory.getClouderaManagerResourceApi(apiClient);
 
-            startClouderaManager(stack, apiClient);
+            startClouderaManager();
             checkParcelApiAvailability();
-            setParcelRepo(stackProductParcel, clouderaManagerResourceApi);
+
+            Set<ClouderaManagerProduct> services = getServices(components);
```
keyki, line 185: What does this list contain on a Data Lake?

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,114 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
             refreshParcelRepos(clouderaManagerResourceApi);
-            downloadParcel(stackProductVersion, parcelResourceApi, product);
-            distributeParcel(stackProductVersion, parcelResourceApi, product);
             if (patchUpgrade) {
-                activateParcel(stackProductVersion, parcelResourceApi, product);
+                installServices(services, parcelResourceApi);
                 restartServices(clustersResourceApi);
             } else {
-                callUpgradeCdhCommand(stackProductVersion, clustersResourceApi);
+                upgradeNonCdhServices(services, cdhService.getName(), parcelResourceApi);
```
keyki, line 193: We need to skip this part for a DL completely.

orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/csd-downloader.sh
tiborpopovics: Sorry, this was for testing

orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/csd-downloader.sh
```diff
@@ -3,6 +3,7 @@
 set -ex
 
 mkdir -p /opt/cloudera/csd
+rm -f /opt/cloudera/csd/*
 cd /opt/cloudera/csd
 
 {% if salt['pillar.get']('cloudera-manager:csd-urls') %}
```
keyki, line 9: There should be a test -f within the loop to skip the download if we already have it.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerParcelManagementService.java
```diff
@@ -0,0 +1,114 @@
...
+        try {
+            ApiCommand apiCommand = clouderaManagerResourceApi.refreshParcelRepos();
+            clouderaManagerPollingServiceProvider.startPollingCmParcelRepositoryRefresh(stack, apiClient, apiCommand.getId());
+        } catch (ApiException e) {
+            LOGGER.info("Unable to refresh parcel repo", e);
+            throw new ClouderaManagerOperationFailedException(e.getMessage(), e);
+        }
+    }
+
+    void downloadParcel(Set<ClouderaManagerProduct> products, ParcelResourceApi parcelResourceApi, Stack stack, ApiClient apiClient)
```
keyki, line 60: We should use services instead of products in these methods.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
keyki, line 185: Unfortunately, using the components is not going to work now. In the case of Data Lake, it will try to re-activate all parcels that are on the images and that doesn't work.

keyki: Discussed offline.

keyki:
To summarize the known issues:
- [x] ALL parcels are re-activated during upgrade (even for Data Lake). 
- [x] Non-cdh parcel upgrade is invoked even when there are no non-CDH based parcels
- [x] Undistributing the old version of the parcels after the upgrade does not work , because we're not checking the version as we used before
- [x] CSH-downloader script downloads the jar files with every highstate.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,125 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
-                .filter(clusterComponent -> clusterComponent.getName().equals(com.sequenceiq.cloudbreak.cloud.model.component.StackType.CDH.name()))
+                .filter(clusterComponent -> ComponentType.CDH_PRODUCT_DETAILS.equals(clusterComponent.getComponentType()))
+                .map(this::getClouderaManagerProduct)
+                .collect(Collectors.toSet());
+    }
+
+    private void upgradeNonCdhProducts(Set<ClouderaManagerProduct> services, String cdhServiceName, ParcelResourceApi parcelResourceApi)
+            throws CloudbreakException, ApiException {
+        Set<ClouderaManagerProduct> nonCdhServices = getNonCdhProducts(services, cdhServiceName);
+        if (!nonCdhServices.isEmpty()) {
```
keyki, line 221: Can you extend these with log messages? An else branch can be introduced to log that we're skipping non-cdh based activaten. On the other case we can log what are the parcel names that we are going to activate.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -170,47 +174,125 @@ public void cleanupCluster(Telemetry telemetry) throws CloudbreakException {
...
+
+    private Set<ClouderaManagerProduct> getProducts(Set<ClusterComponent> components) {
         return components.stream()
-                .filter(clusterComponent -> clusterComponent.getName().equals(com.sequenceiq.cloudbreak.cloud.model.component.StackType.CDH.name()))
+                .filter(clusterComponent -> ComponentType.CDH_PRODUCT_DETAILS.equals(clusterComponent.getComponentType()))
+                .map(this::getClouderaManagerProduct)
+                .collect(Collectors.toSet());
+    }
+
+    private void upgradeNonCdhProducts(Set<ClouderaManagerProduct> services, String cdhServiceName, ParcelResourceApi parcelResourceApi)
```
keyki, line 218: services -> products

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterManagerUpgradeService.java
```diff
@@ -98,12 +99,19 @@ private void startClusterServices(Stack stack) throws CloudbreakException {
...
-        Optional<String> license = clusterHostServiceRunner.decoratePillarWithClouderaManagerLicense(cluster.getStack().getId(), servicePillar);
+        Optional<String> license = clusterHostServiceRunner.decoratePillarWithClouderaManagerLicense(stackId, servicePillar);
         clusterHostServiceRunner.decoratePillarWithClouderaManagerRepo(clouderaManagerRepo, servicePillar, license);
         clusterHostServiceRunner.decoratePillarWithClouderaManagerSettings(servicePillar, clouderaManagerRepo);
+        decorateWorkloadClusterPillarWithCsdDownloader(stackType, cluster, servicePillar);
         return new SaltConfig(servicePillar);
     }
+
+    private void decorateWorkloadClusterPillarWithCsdDownloader(StackType stackType, Cluster cluster, Map<String, SaltPillarProperties> servicePillar) {
+        if (StackType.WORKLOAD.equals(stackType)) {
```
keyki, line 113: Can you please add a log message that we're skipping the CSD update because it's a Data Lake

orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/csd-downloader.sh
```diff
@@ -6,11 +6,22 @@ mkdir -p /opt/cloudera/csd
...
+csdUrls=({%- for url in salt['pillar.get']('cloudera-manager:csd-urls') -%}
+{{ url + " " }}
+{%- endfor %})
+
+for url in ${csdUrls[@]}
+do
+  fileName=$(basename $url)
+  if test -f $fileName
+  then
+    echo $fileName "already exists" >> /var/csd_downloaded
```
keyki, line 18: This should go to /var/log

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterManagerUpgradeService.java
```diff
@@ -98,12 +99,21 @@ private void startClusterServices(Stack stack) throws CloudbreakException {
...
```
keyki, line 113: Can you please cover this part with a unit test to make sure we don't update the CSDs in the case of Data Lake. If we break this then the complete Data Lake upgrade will be broken as the CSDs are not downloadable outside of Cloudera.

keyki: Also, this could be improved that only update the CSDs if needed and those CSDs that are needed. But this does not need to be part of this PR.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerParcelManagementService.java
```diff
@@ -0,0 +1,115 @@
...
+        for (ClouderaManagerProduct product : products) {
+            LOGGER.info("Downloading {} parcel.", product.getName());
+            ApiCommand apiCommand = parcelResourceApi.startDownloadCommand(stack.getName(), product.getName(), product.getVersion());
+            PollingResult pollingResult = clouderaManagerPollingServiceProvider.startPollingCdpRuntimeParcelDownload(
+                    stack, apiClient, apiCommand.getId(), new ParcelResource(stack.getName(), product.getName(), product.getVersion()));
+            if (isExited(pollingResult)) {
+                throw new CancellationException("Cluster was terminated while waiting for CDP Runtime Parcel to be downloaded");
+            } else if (isTimeout(pollingResult)) {
+                throw new CloudbreakException("Timeout during the updated CDP Runtime Parcel download.");
+            }
```
lacikaaa, line 72: this part is duplicated a few times, message maybe different a bit, but that could be a parameter

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerUpgradeService.java
```diff
@@ -0,0 +1,72 @@
...
+                ApiCdhUpgradeArgs upgradeArgs = new ApiCdhUpgradeArgs();
+                upgradeArgs.setCdhParcelVersion(stackProductVersion);
+                upgradeCommand = clustersResourceApi.upgradeCdhCommand(stack.getName(), upgradeArgs);
+            }
+            PollingResult pollingResult = clouderaManagerPollingServiceProvider.startPollingCdpRuntimeUpgrade(stack, apiClient, upgradeCommand.getId());
+            if (isExited(pollingResult)) {
+                throw new CancellationException("Cluster was terminated while waiting for CDP Runtime to be upgraded");
+            } else if (isTimeout(pollingResult)) {
+                throw new CloudbreakException("Timeout during CDP Runtime upgrade.");
+            }
```
lacikaaa, line 55: it's kinda the same here, duplicating the same 5 lines, maybe it should be moved some util like thing or a component

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerUpgradeService.java
```diff
@@ -0,0 +1,72 @@
...
+        } catch (ApiException ex) {
+            if (ex.getResponseBody().contains("Cannot upgrade because the version is already CDH")) {
```
lacikaaa, line 57: null check for response body?

orchestrator-salt/src/main/resources/salt/salt/cloudera/csd/csd-downloader.sh
```diff
@@ -6,11 +6,22 @@ mkdir -p /opt/cloudera/csd
...
-{% for url in salt['pillar.get']('cloudera-manager:csd-urls') %}
-curl -L -O -R {{ url }}
-{% endfor %}
+csdUrls=({%- for url in salt['pillar.get']('cloudera-manager:csd-urls') -%}
+{{ url + " " }}
+{%- endfor %})
+
+for url in ${csdUrls[@]}
+do
+  fileName=$(basename $url)
```
lacikaaa, line 15: I would add a log here with url and filename to see what's going on. also I would add a datetime to logs. it usually helps during investigation to see what happened when

lacikaaa: might use `tee` to have the logs in minion log and the `csd_downloaded` too

lacikaaa (approved):

keyki (approved):

PR #9594, opened 2020-12-07
CB-10066 New API methods to add and delete DNS A and CNAME records

lacikaaa:
@ianbuss Please check if this API fits your use-case

@jamisonbennett could you check this? I'm interested in if I'm missing something or you would do differently


freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/doc/DnsModelDescription.java
```diff
@@ -0,0 +1,15 @@
...
+
+public final class DnsModelDescription {
+    public static final String DNS_RECORD = "DNS record name without the domain. eg. 'ipaserver' from 'ipaserver.clouder.site'";
+
+    public static final String DNS_ZONE = "It's the domain. Like if your FQDN is ipaserver.clouder.site, it's 'cloudera.site'. "
+            + "'168.192.in-addr.arpa' for a reverse record like '5.1.168.192.in-addr.arpa'";
+
+    public static final String CNAME_TARGET_FQDN = "The fully qualified domain name of the host the CNAME should point to.";
+
+    public static final String CREATE_REVERSE = "Tries to create a reverse point for the recorc (PTR). Only if reverse zone already exists";
```
lajosrodek, line 11: Typo: `recorc` -> `record`.

lacikaaa: thanks, I fixed locally. For now I won't push it.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/doc/DnsModelDescription.java
```diff
@@ -0,0 +1,15 @@
+package com.sequenceiq.freeipa.api.v1.dns.doc;
+
+public final class DnsModelDescription {
+    public static final String DNS_RECORD = "DNS record name without the domain. eg. 'ipaserver' from 'ipaserver.clouder.site'";
+
+    public static final String DNS_ZONE = "It's the domain. Like if your FQDN is ipaserver.clouder.site, it's 'cloudera.site'. "
+            + "'168.192.in-addr.arpa' for a reverse record like '5.1.168.192.in-addr.arpa'";
+
+    public static final String CNAME_TARGET_FQDN = "The fully qualified domain name of the host the CNAME should point to.";
```
jamisonbennett, line 9: Are the zones expected to have a trailing "."? Technically this is what an FQDN means but I think this should be called out if it is required since its a common misconception.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/DnsV1Endpoint.java
```diff
@@ -75,4 +77,34 @@ void deleteDnsZoneBySubnetId(@QueryParam("environment") @NotEmpty String environ
...
+    void addDnsCnameRecord(@Valid AddDnsCnameRecordRequest request);
+
+    @DELETE
+    @Path("record/cname")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = DnsOperationDescriptions.DELETE_DNS_CNAME_RECORD, produces = MediaType.APPLICATION_JSON, notes = FreeIpaNotes.FREEIPA_NOTES,
+            nickname = "deleteDnsCnameRecordV1")
+    void deleteDnsCnameRecord(@QueryParam("environment") @NotEmpty String environmentCrn, @QueryParam("dnsZone") @NotEmpty String dnsZone,
+            @QueryParam("record") @NotEmpty String record);
 }
```
jamisonbennett, line 110: Should we also add an API to query the a/cname records?

freeipa/src/main/java/com/sequenceiq/freeipa/controller/DnsV1Controller.java
```diff
@@ -126,4 +128,44 @@ public void deleteDnsRecordsByFqdn(@ResourceCrn @NotEmpty String environmentCrn,
...
+
+    @Override
+    @CheckPermissionByResourceCrn(action = EDIT_ENVIRONMENT)
+    @Retryable(value = RetryableFreeIpaClientException.class,
+            maxAttemptsExpression = RetryableFreeIpaClientException.MAX_RETRIES_EXPRESSION,
+            backoff = @Backoff(delayExpression = RetryableFreeIpaClientException.DELAY_EXPRESSION,
+                    multiplierExpression = RetryableFreeIpaClientException.MULTIPLIER_EXPRESSION))
+    public void deleteDnsCnameRecord(@ResourceCrn @NotEmpty String environmentCrn, @NotEmpty String dnsZone, @NotEmpty String record) {
+
+    }
```
jamisonbennett, line 170: It looks like this commit just adds the API and not the implementation. Is that intended?

lacikaaa: yes, implementation is in progress

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/doc/DnsModelDescription.java
lacikaaa: I will test it with and without trailing `.` and if it's necessary I think I will add it on the backend side. But I have managed to create an A record with ipa cli without adding the `.` after the zone name or adding one before. (checked the sent request and the cli hasn't added the dot so I think we are good without it)

freeipa/src/main/java/com/sequenceiq/freeipa/controller/DnsV1Controller.java
keyki: This is just for reviewing the API for the different teams.

hreeve-cloudera (commented):
I'll defer to Jamison, but this looks good to me.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/DnsV1Endpoint.java
lacikaaa: we might add later. it wasn't requested, so I won't implement it in this PR

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/model/AddDnsCnameRecordRequest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Pattern(regexp = DNS_CNAME_PATTERN, message = DNS_CNAME_MSG)
+    private String cname;
+
+    @ApiModelProperty(DnsModelDescription.DNS_ZONE)
+    @Pattern(regexp = DNS_ZONE_PATTERN, message = DNS_ZONE_MSG)
+    private String dnsZone;
+
+    @NotEmpty
+    @ApiModelProperty(value = DnsModelDescription.CNAME_TARGET_FQDN, required = true)
+    @Pattern(regexp = "^[a-zA-Z0-9]+[a-zA-Z0-9-\\.]*[a-zA-Z0-9]+$",
```
jamisonbennett, line 40: I think you might want the following regex: "^[a-zA-Z0-9]+[a-zA-Z0-9.-]*[a-zA-Z0-9.]+$"

1. The dash after the 9 is invalid. Put it at the front/end of the bracketed section so that it doesn't form a range.

2. Assuming the "-" is removed, this regex allows "\\" in FQDNs:
```
echo www\\google.com | egrep "^[a-zA-Z0-9]+[a-zA-Z0-9\\.]*[a-zA-Z0-9]+$"
```

3. nit: This regex technically doesn't allow a trailing "." and that technically is required to be a part of an FQDN (but it is very rare that that is actually used).

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/model/DnsRecordRegexpPatterns.java
```diff
@@ -0,0 +1,19 @@
+package com.sequenceiq.freeipa.api.v1.dns.model;
+
+public final class DnsRecordRegexpPatterns {
+
+    public static final String DNS_ZONE_PATTERN = "^[a-zA-Z0-9-\\.]+$";
```
jamisonbennett, line 5: I think this might be the same rules as above: "^[a-zA-Z0-9]+[a-zA-Z0-9.-]*[a-zA-Z0-9.]+$"

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/model/DnsRecordRegexpPatterns.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public static final String DNS_ZONE_MSG = "DNS zone must be valid. It can conatin alphanumeric characters, dash and dot.";
+
+    public static final String DNS_CNAME_PATTERN = "^(\\*\\.)?[a-zA-Z0-9]+[a-zA-Z0-9-\\.]*[a-zA-Z0-9]+$";
+
+    public static final String DNS_CNAME_MSG = "CNAME must be valid. Might start with '*.' and can conatin alphanumeric characters, dash and dot.";
+
+    public static final String DNS_HOSTNAME_PATTERN = "^[a-zA-Z0-9]+[a-zA-Z0-9-\\.]*[a-zA-Z0-9]+$";
```
jamisonbennett, line 13: Same as above: "^[a-zA-Z0-9]+[a-zA-Z0-9.-]*[a-zA-Z0-9.]+$"

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
+            validateExistingARecordMatchesRequested(request.getIp(), dnsRecord.get());
+        }
+    }
+
+    private String calculateZone(String zoneFromRequest, FreeIpaAndClient freeIpaAndClient) throws FreeIpaClientException {
+        Optional<String> optionalZone = Optional.ofNullable(zoneFromRequest);
+        String zone = optionalZone.orElse(freeIpaAndClient.getFreeIpa().getDomain());
+        boolean providedZoneDifferentThanDefault = !freeIpaAndClient.getFreeIpa().getDomain().equals(StringUtils.removeEnd(zone, "."));
+        if (optionalZone.isPresent() && providedZoneDifferentThanDefault) {
+            validateZoneExists(optionalZone.get(), freeIpaAndClient.getClient());
```
pdarvasi, line 125: a log line would be nice here

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
+            String msg = String.format("Zone [%s] doesn't exists", zone);
+            LOGGER.info(msg);
+            throw new BadRequestException(msg);
+        }
+    }
+
+    private void validateExistingARecordMatchesRequested(String ip, DnsRecord record) {
+        LOGGER.debug("Validating already existing record: {}", record);
+        if (!record.isARecord()) {
+            LOGGER.info("Record already exists and it's not an A record");
```
pdarvasi, line 146: 1 string would be enough instead of 2

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
+            throw new DnsRecordConflictException("Record already exists and it's not an A record");
+        } else if (!record.getArecord().contains(ip)) {
+            LOGGER.info("Record already exists and the IP doesn't match");
```
pdarvasi, line 149: 1 string would be enough instead of 2

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
+            createDnsCnameRecord(freeIpaAndClient.getClient(), zone, request.getCname(), request.getTargetFqdn());
+        } else {
+            validateExistingCnameRecordMatches(dnsRecord.get(), request.getTargetFqdn());
+        }
+    }
+
+    private void validateExistingCnameRecordMatches(DnsRecord record, String targetFqdn) {
+        LOGGER.debug("Validating already existing record: {}", record);
+        if (!record.isCnameRecord()) {
+            LOGGER.info("Record already exists and it's not a CNAME record");
```
pdarvasi, line 182: 1 string would be enough instead of 2

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
+            throw new DnsRecordConflictException("Record already exists and it's not a CNAME record");
+        } else if (!record.getCnamerecord().contains(targetFqdn)) {
+            LOGGER.info("Record already exists and the target doesn't match");
```
pdarvasi, line 185: 1 string would be enough instead of 2

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -55,23 +70,142 @@ private void deleteRegularRecords(FreeIpaClient freeIpaClient, DnsZone dnsZone,
...
             throws FreeIpaClientException {
         Set<DnsRecord> srvRecordsToDelete = allDnsRecordsInZone.stream()
-                .filter(record -> fqdns.stream().anyMatch(fqdn -> record.isHostRelatedSrvRecord(fqdn)))
+                .filter(record -> fqdns.stream().anyMatch(record::isHostRelatedSrvRecord))
                 .collect(Collectors.toSet());
         for (DnsRecord dnsRecord : srvRecordsToDelete) {
             for (String fqdn : fqdns) {
                 List<String> srvRecords = dnsRecord.getHostRelatedSrvRecords(fqdn);
                 if (!srvRecords.isEmpty()) {
                     LOGGER.info("Delete DNS SRV record [{}] for [{}] in zone [{}]", dnsRecord.getIdnsname(), fqdn, dnsZone);
```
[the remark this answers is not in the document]
lacikaaa: haven't modified that part, won't touch it now

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
```diff
@@ -35,16 +48,18 @@
...
+                    multiplierExpression = RetryableFreeIpaClientException.MULTIPLIER_EXPRESSION))
+    @Measure(DnsRecordService.class)
+    public void deleteDnsRecordByFqdn(String environmentCrn, String accountId, List<String> fqdns) throws FreeIpaClientException {
+        FreeIpaAndClient freeIpaAndClient = createFreeIpaAndClient(environmentCrn, accountId);
+        for (DnsZone dnsZone : freeIpaAndClient.getClient().findAllDnsZone()) {
             LOGGER.debug("Looking for records in zone [{}]", dnsZone.getIdnsname());
-            Set<DnsRecord> allDnsRecordsInZone = freeIpaClient.findAllDnsRecordInZone(dnsZone.getIdnsname());
-            deleteRegularRecords(freeIpaClient, dnsZone, allDnsRecordsInZone, fqdns, freeIpa.getDomain());
-            deleteSrvRecords(freeIpaClient, dnsZone, allDnsRecordsInZone, fqdns);
+            Set<DnsRecord> allDnsRecordsInZone = freeIpaAndClient.getClient().findAllDnsRecordInZone(dnsZone.getIdnsname());
```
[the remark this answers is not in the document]
lacikaaa: it's just a pojo get, shouldn't matter

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
lacikaaa: could you elaborate?

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/dns/model/AddDnsCnameRecordRequest.java
lacikaaa: 1. I'm writing unit tests, and it seems valid

2. `\\` is necessary. first `\` is for escaping the dot, as by default dot matches anything, the second one is necessary for java to escape `\`

3. For now I will leave it this way, we can weaken the regexp later if necessary

lacikaaa: I have added unit tests to validate the regex. Could you check?

jamisonbennett (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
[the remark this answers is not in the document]
pdarvasi: it just hurts my ocd

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/dns/DnsRecordService.java
pdarvasi: You have the same text here and in next line:"Record already exists and the IP doesn't match". It would be more convenient to have only 1.

lacikaaa: on purpose. I didn't want to add multiple extra lines just to ensure the log and exception has the same message

PR #9595, opened 2020-12-07
CB-10157 Store the original tags in our services and transform GCP la…

biharitomi:
…bels only in cloud-gcp module. This is necessary because we stored the label transformed tags previously. This way we can get the original tags in the pillar which are fundamental from telemtry perspective

This is the cherry pick of https://github.com/hortonworks/cloudbreak/pull/9577 as it was intended to be part of 2.35 line.


cegganesh84 (approved):
LGTM

horadla23 (approved):
LGTM

PR #9596, opened 2020-12-07
CDPCP-2991. Remove entitlement check for user sync performance improv…

handavid:
…ement

This commit releases the user sync performance improvement by removing the
entitlement check. Behavior is now as follows:

All full user syncs will attempt to use the UMS getUserSyncStateModel RPC.
Failures using this RPC will cause user sync to default back to the regular
user sync code path.

Partial user syncs will still use the regular user sync code path.


lacikaaa (approved):
Has this entitlement been granted to any customer on prod to test in real scenario?

PR #9597, opened 2020-12-08
CB-10173 CLI flag for Public Gateway Endpoint

enis:
This adds two new parameters to the EnvironmentRequest for the feature called "semi-private
networks" also known as "public endpoint access gateway". Details of the feature can be
found in CRB-1318.

Since all of the network related and enableTunnel related parameters are at the environment
level (rather than per-datalake or per-datahub) I added these parameters at the environment-api
rather than datalake-api. Also logically it seemed appropriate to bundle these together in
the EnvironmentNetworkRequest since all network related configuration is there.

Testing done:
 - ./gradlew build


enis:
This PR probably needs more work to add converters, etc. Just putting out there to see whether the API change here makes sense.

hreeve-cloudera:
@enis I think you've got it right, and these are where we'll ultimately have to make the changes. I'll start poking around and see what updates we'll need to make to persisted the new fields to the database, and propagate the information to the appropriate places.

orngejaket:
Currently, this looks fine.

enis:
@hreeve-cloudera has incorporated these changes in #9620. Let me close this one.

PR #9601, opened 2020-12-08
CB-9886 Nodes not killed on Cloud Provider if a scale operation fails

sodre90:


horadla23 (approved):
LGTM, target is still 2.35?

sodre90:
> LGTM, target is still 2.35?

Moved to 2.36, thanks

PR #9604, opened 2020-12-08
CB-10239. Increase the default node count for the compute group to 1 …

sidseth:
…for the DE cluster shapes for 7.2.1 templates

Merged the original PR with a bad commit message (https://github.com/hortonworks/cloudbreak/pull/9599). This is just to fix the commit message, and link to a new jira instead of 9902 since that went into 2.34.


cegganesh84 (approved):
LGTM

sidseth:
OK. That didn't work either. This got merged in with the original commit message.

PR #9609, opened 2020-12-09
CB-10217 Add support for specifying CM hosts' roleRefNames in blueprint

cegganesh84:
1. At the moment the Hue does not point to the current namenode in an HA cluster after failovers.
2. The fix is to use HTTPFS.
3. Adding this role requires support for specifying roleRefNames in the hosts' section of the CM template.
4. Added support for the same and tested via unit tests and end to end tests.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
@@ -468,9 +468,27 @@ public class CmTemplateProcessor implements BlueprintTextProcessor {
     }
 
     public void addHosts(Map<String, List<Map<String, String>>> hostGroupMappings) {
-        hostGroupMappings.forEach((hostGroup, hostAttributes) -> hostAttributes.forEach(
-                attr -> cmTemplate.getInstantiator().addHostsItem(new ApiClusterTemplateHostInfo().hostName(attr.get("fqdn")).hostTemplateRefName(hostGroup))
-        ));
+        ApiClusterTemplateInstantiator instantiator = cmTemplate.getInstantiator();
+        Map<String, List<String>> roleRefNameMap = null;
+        List<ApiClusterTemplateHostInfo> hosts = instantiator.getHosts();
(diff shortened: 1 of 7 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -468,9 +468,27 @@ public ApiClusterTemplate getTemplate() {
     }
 
     public void addHosts(Map<String, List<Map<String, String>>> hostGroupMappings) {
-        hostGroupMappings.forEach((hostGroup, hostAttributes) -> hostAttributes.forEach(
-                attr -> cmTemplate.getInstantiator().addHostsItem(new ApiClusterTemplateHostInfo().hostName(attr.get("fqdn")).hostTemplateRefName(hostGroup))
-        ));
+        ApiClusterTemplateInstantiator instantiator = cmTemplate.getInstantiator();
+        Map<String, List<String>> roleRefNameMap = null;
```
lajosrodek, line 472: Nit: Why not call this directly `finalRoleRefNameMap`? I do not see the justification for having `roleRefNameMap` as it is only used to initialize `finalRoleRefNameMap`.

cegganesh84: That can be done by re-arranging the code.

Here I can not perform the initialization without this check `hosts != null && !hosts.isEmpty()`. IF I have to then I need to move the null check to the fetch function. I can not actually clear the hosts' entry in a function named fetch. So need to find a new name for the function as well.

The code was structured the other way, but only to realize I can not use the reference inside a lambda unless it is a final.

cegganesh84:
This is a bad feature. It can not be done. roleRefNames can be only configured by CM auto-configuration.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
lajosrodek: Ah OK, got it. Makes sense. 😃

PR #9612, opened 2020-12-09
CB-10239:Increase the default node count for the compute group to 1 for the DE cluster shapes for 7.2.1 template

satyabollineni:
…E cluster shapes for 7.2.1 templates"

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.1/aws/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.1/aws/dataengineering-ha.json
@@ -32,7 +32,7 @@
       "recipeNames": []
     },
       {
-        "nodeCount": 0,
+        "nodeCount": 1,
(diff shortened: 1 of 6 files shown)


akanto:
Is this only for 7.2.1? Do 7.2.2 and 7.2.6 clusters suffering from the same issue?

sidseth:
> Is this only for 7.2.1? Do 7.2.2 and 7.2.6 clusters suffering from the same issue?

The changes for 7.2.2, 7.2.6 and 7.2.7 went in via https://github.com/hortonworks/cloudbreak/commit/650f350eb7914054117843679b3d9f77e816b858
(At the time the plan was to have 7.2.2 as the min version. That changed later to include 7.2.1 - hence this PR).

Question from the jira remains. Can this go into 2.3.5 (the earlier changes is already in 2.3.4), or does it wait till 2.3.6?

akanto (approved):

PR #9618, opened 2020-12-09
CB-10255. Ability to use DataBus CNAME instead of the sigma DataBus e…

oleewere:
…ndpoint.

details:
add new entitlement to let customers to use databus cname endpoint instead of the original one

the value would be: CDP_USE_DATABUS_CNAME_ENDPOINT


lnardai (approved):

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
```diff
@@ -0,0 +1,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DataBusEndpointProvider {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DataBusEndpointProvider.class);
+
+    private static final String CNAME_ENDPOINT = "https://dbusapi.us-west-1.altus.cloudera.com";
```
horadla23, line 12: Are we sure this URL won't change? If not, then maybe it is worth to make a new config property for this in application.yml

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/service/DiagnosticsTriggerService.java
```diff
@@ -71,9 +81,12 @@ public FlowIdentifier startDiagnosticsCollection(BaseDiagnosticsCollectionReques
                 && StringUtils.isNotBlank(stack.getCluster().getBlueprint().getStackVersion())) {
             clusterVersion = stack.getCluster().getBlueprint().getStackVersion();
         }
+        String accountId = Crn.fromString(stack.getResourceCrn()).getAccountId();
+        boolean useDbusCnameEndpoint = entitlementService.useDataBusCNameEndpointEnabled(INTERNAL_ACTOR_CRN, accountId);
+        String databusEndpoint = dataBusEndpointProvider.getDataBusEndpoint(telemetry.getDatabusEndpoint(), useDbusCnameEndpoint);
```
horadla23, line 86: This logic is very similar to the logic in TelemetryDecorator and in ClouderaManagerMgmtTelemetryService, might worth to make an own method for it somewhere

oleewere: common has no auth-connector dependency, first i tried to put these kind of code together in auth-connector but @lnardai did not want that (I understand) - so i cannot include Entitlementservice in DataBusEndpointProvider

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
oleewere: does not seemed to be changed ever. I can do a follow-up change if that will be needed. although if it changes it should be changed everywhere so actually it sounds more like a constant still

horadla23: ok, then it is fine for me

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/service/DiagnosticsTriggerService.java
horadla23: ohh i see, makes sense

horadla23 (approved):

PR #9620, opened 2020-12-09
CB-10242 API changes to support semi-private networks

hreeve-cloudera:
Adds new optional fields 'usePublicEndpointAccessGateway' and
'endpointGatewaySubnetIds' to the EnvironmentNetworkV1Request API object. These
fields are persisted to the EnvironmentNetwork and NetworkDto BaseNetwork
objects. Adds new columns for these fields to the environment_network table
and saves them to the database.

Tested with unit tests, and by manually setting the new fields in the IDE and
verifying they were converted/persisted in the database correctly.

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
@@ -14,6 +14,14 @@ public class EnvironmentModelDescription {
     public static final String NETWORKCIDRS = "The network cidrs for the configured vpc";
     public static final String PRIVATE_SUBNET_CREATION = "A flag to enable or disable the private subnet creation.";
     public static final String SERVICE_ENDPOINT_CREATION = "A flag to enable or disable the service endpoint creation.";
+    public static final String PUBLIC_ENDPOINT_ACCESS_GATEWAY = "A flag to enable the Public Endpoint Access Gateway, which provides public UI/API " +
+        "access to private data lakes and data hubs.";
+    public static final String ENDPOINT_ACCESS_GATEWAY_SUBNET_IDS = "Subnet ids for the Public Endpoint Access Gateway. If provided, these " +
+        "are the subnets that will be used to create a public Knox endpoint for out-of-network UI/API access. If not provided, public subnets will be " +
+        "selected from the subnet list provided for environment creation. (Optional)";
+    public static final String ENDPOINT_ACCESS_GATEWAY_SUBNET_METAS = "Subnet metadata for the Public Endpoint Access Gateway. If provided, these " +
+        "are the subnets that will be used to create a public Knox endpoint for out-of-network UI/API access. If not provided, public subnets will be " +
+        "selected from the subnet list provided for environment creation. (Optional)";
     public static final String OUTBOUND_INTERNET_TRAFFIC = "A flag to enable or disable the outbound internet traffic from the instances.";
     public static final String AWS_SPECIFIC_PARAMETERS = "Subnet ids of the specified networks";
     public static final String AZURE_SPECIFIC_PARAMETERS = "Subnet ids of the specified networks";
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/base/EnvironmentNetworkBase.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/base/EnvironmentNetworkBase.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.sequenceiq.cloudbreak.validation.SubnetType;
 import com.sequenceiq.cloudbreak.validation.ValidSubnet;
 import com.sequenceiq.common.api.type.OutboundInternetTraffic;
+import com.sequenceiq.common.api.type.PublicEndpointAccessGateway;
 import com.sequenceiq.common.api.type.ServiceEndpointCreation;
 import com.sequenceiq.environment.api.doc.environment.EnvironmentModelDescription;
 import com.sequenceiq.environment.api.v1.environment.model.EnvironmentNetworkAwsParams;
@@ -41,6 +42,12 @@ public abstract class EnvironmentNetworkBase {
 
     @ApiModelProperty(EnvironmentModelDescription.SERVICE_ENDPOINT_CREATION)
     private ServiceEndpointCreation serviceEndpointCreation = ServiceEndpointCreation.DISABLED;
+
+    @ApiModelProperty(EnvironmentModelDescription.PUBLIC_ENDPOINT_ACCESS_GATEWAY)
+    private PublicEndpointAccessGateway publicEndpointAccessGateway = PublicEndpointAccessGateway.DISABLED;
+
+    @ApiModelProperty(value = EnvironmentModelDescription.ENDPOINT_ACCESS_GATEWAY_SUBNET_IDS)
+    private Set<String> endpointGatewaySubnetIds = Set.of();
 
     @ApiModelProperty(EnvironmentModelDescription.OUTBOUND_INTERNET_TRAFFIC)
     private OutboundInternetTraffic outboundInternetTraffic = OutboundInternetTraffic.ENABLED;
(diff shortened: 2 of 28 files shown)


jamisonbennett:
Thanks @hreeve-cloudera. I gave the PR a quick look. This is in an area that I am less familiar with, so I can only say that I don't see any obvious problems and someone else should give the PR a full review.

Thanks for adding several tests.

enis (approved):
lgtm fwiw. 

I have noticed that this PR does not contain: 
 - Using the provided flag / subnets in the LB creation
 - logic to auto-fill the subnet to use if none has been provided. 

Is the plan to add these as a follow up? I think it makes sense to merge this in since it is contained to the API related changes and persisting of those.

hreeve-cloudera:
> lgtm fwiw.
> 
> I have noticed that this PR does not contain:
> 
>     * Using the provided flag / subnets in the LB creation
> 
>     * logic to auto-fill the subnet to use if none has been provided.
> 
> 
> Is the plan to add these as a follow up? I think it makes sense to merge this in since it is contained to the API related changes and persisting of those.

@enis Yes, I was going to add that logic as part of https://jira.cloudera.com/browse/CB-10095, which is partially done and just had placeholders for these API changes. I wanted to get this in first, so that neither of us will be blocked on it.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/base/EnvironmentNetworkBase.java
```diff
@@ -42,6 +42,12 @@
     @ApiModelProperty(EnvironmentModelDescription.SERVICE_ENDPOINT_CREATION)
     private ServiceEndpointCreation serviceEndpointCreation = ServiceEndpointCreation.DISABLED;
 
+    @ApiModelProperty(EnvironmentModelDescription.PUBLIC_ENDPOINT_ACCESS_GATEWAY)
+    private Boolean usePublicEndpointAccessGateway = false;
```
akanto, line 46: Usually, we prefer enums instead of booleans, it makes it easier to extend it the future.

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
```diff
@@ -14,6 +14,9 @@
     public static final String NETWORKCIDRS = "The network cidrs for the configured vpc";
     public static final String PRIVATE_SUBNET_CREATION = "A flag to enable or disable the private subnet creation.";
     public static final String SERVICE_ENDPOINT_CREATION = "A flag to enable or disable the service endpoint creation.";
+    public static final String PUBLIC_ENDPOINT_ACCESS_GATEWAY = "A flag to select a public or private endpoint access gateway.";
+    public static final String ENDPOINT_ACCESS_GATEWAY_SUBNET_IDS = "Subnet ids for the endpoint access gateway. (Optional)";
```
akanto, line 18: Can you add more description here, please? What is the purpose of the subnetids?

environment/src/main/resources/schema/app/20201208195427_CB-10242_API_changes_for_semi-private_networks.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-10242 API changes for semi-private networks
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE environment_network ADD COLUMN IF NOT EXISTS usepublicendpointaccessgateway BOOLEAN NOT NULL DEFAULT false;
```
doktoric, line 4: this fields can not be not null because in case of a production roll back we need to use the modified scheme with the previous CB version

doktoric: Default value shouldn't be added like this. It will block the whole table.
See https://dba.stackexchange.com/questions/60965/does-postgresql-optimize-adding-columns-with-non-null-defaults/216153#216153
Example: https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/resources/schema/app/20201130155647_CB-9580_new_field_on_cluster:_embedded_db_is_configured_on_attached_disk_or_not.sql

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/NetworkRequestToDtoConverter.java
```diff
@@ -110,4 +115,8 @@ private ServiceEndpointCreation getServiceEndpointCreation(EnvironmentNetworkReq
     private OutboundInternetTraffic getOutboundInternetTraffic(EnvironmentNetworkRequest network) {
         return Optional.ofNullable(network.getOutboundInternetTraffic()).orElse(OutboundInternetTraffic.ENABLED);
     }
+
+    private Boolean getUsePublicEndpointAccessGateway(EnvironmentNetworkRequest network) {
```
doktoric, line 119: use enum please

environment/src/main/java/com/sequenceiq/environment/network/dao/domain/BaseNetwork.java
```diff
@@ -79,6 +79,12 @@
     @Convert(converter = ServiceEndpointCreationConverter.class)
     private ServiceEndpointCreation serviceEndpointCreation;
 
+    private boolean usePublicEndpointAccessGateway;
+
+    @Convert(converter = JsonToString.class)
+    @Column(columnDefinition = "TEXT", nullable = false)
```
doktoric, line 85: we can not set nullable = false here

environment/src/main/java/com/sequenceiq/environment/network/dao/domain/BaseNetwork.java
```diff
@@ -169,6 +175,23 @@ public void setSubnetMetas(Map<String, CloudSubnet> subnetMetas) {
         });
     }
 
+    public boolean isUsePublicEndpointAccessGateway() {
+        return usePublicEndpointAccessGateway;
+    }
+
+    public void setUsePublicEndpointAccessGateway(boolean usePublicEndpointAccessGateway) {
```
doktoric, line 182: enum please

environment/src/main/java/com/sequenceiq/environment/network/dao/domain/BaseNetwork.java
```diff
@@ -169,6 +175,23 @@ public void setSubnetMetas(Map<String, CloudSubnet> subnetMetas) {
...
+        this.usePublicEndpointAccessGateway = usePublicEndpointAccessGateway;
+    }
+
+    public void setEndpointGatewaySubnetMetas(Map<String, CloudSubnet> endpointGatewaySubnetMetas) {
```
doktoric, line 186: enum please

doktoric:
I noticed that the API response changed:
- please modify the e2e test and EnvironmentNetworkTestAssertion to check the response
- probably we should include this response into the public API
- probably we should add this somehow to the UI as well

hreeve-cloudera:
> lgtm fwiw.
> 
> I have noticed that this PR does not contain:
> 
>     * Using the provided flag / subnets in the LB creation
> 
>     * logic to auto-fill the subnet to use if none has been provided.
> 
> 
> Is the plan to add these as a follow up? I think it makes sense to merge this in since it is contained to the API related changes and persisting of those.

@enis Yes, I was going to add that logic as part of https://jira.cloudera.com/browse/CB-10095, which is partially done and just had placeholders for these API changes. I wanted to get this in first, so that neither of us will be blocked on it.

> I noticed that the API response changed:
> 
>     * please modify the e2e test and EnvironmentNetworkTestAssertion to check the response
> 
>     * probably we should include this response into the public API
> 
>     * probably we should add this somehow to the UI as well

Thanks @doktoric. I have some follow up question:

1. I'll update EnvironmentNetworkTestAssertion. For the e2e test, where exactly do I make those changes? I haven't modified the e2e tests before.
2. Is the public API the thunderhead API? @enis is making the Thunderhead API changes in https://jira.cloudera.com/browse/CDPCP-3435. If there's another place I need to make corresponding changes please let me know.
3. These options will be exposed in the UI during environment creation in https://jira.cloudera.com/browse/CB-10093. Is that what you mean, or do you mean we should add the updated response to the UI?

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentModelDescription.java
hreeve-cloudera: Added more detail. Please check out the updated descriptions and let me know if they're sufficient.

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/NetworkRequestToDtoConverter.java
hreeve-cloudera: I've created an enum `PublicEndpointAccessGateway` with values `ENABLED` and `DISABLED`, and used it in all the objects that have `usePublicEndpointAccessGateway`.

hreeve-cloudera:
@doktoric I've added a new e2e test, `EnvironmentWithPublicEndpointAccessGatewayTests`, and have added validation logic for the new fields to `EnvironmentNetworkTestAssertion`.

doktoric (approved):

PR #9621, opened 2020-12-10
CB-9462 [E2E] Enable Private Link creation in CB E2E test

gergopapi2:
This commit enabled to test postgres via private endpoint on azure as well. It can be turned on for AWS and AZURE via environment variable 'integrationtest.privateendpoint.enabled'. For other platforms it is disabled for now.

For AZURE it also needs to be used together with single rg option. If it is a multiple rg deployment, then private endpoints are automatically turned off in int-test.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/PrivateEndpointTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/PrivateEndpointTest.java
@@ -1,0 +1,13 @@
+package com.sequenceiq.it.cloudbreak;
+
+public class PrivateEndpointTest extends GherkinTest {
+
+    public static final String PRIVATE_ENDPOINT_USAGE = "PRIVATE_ENDPOINT_USAGE";
+
+    public static final String PRIVATE_ENDPOINT_ENABLED = "enabled";
+
+    public static final String PRIVATE_ENDPOINT_DISABLED = "disabled";
+
+    private PrivateEndpointTest() {
+    }
+}
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/ResourceGroupTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/ResourceGroupTest.java
@@ -1,14 +1,21 @@
 package com.sequenceiq.it.cloudbreak;
 
-public class ResourceGroupTest {
+public class ResourceGroupTest extends GherkinTest {
     public static final String AZURE_RESOURCE_GROUP_USAGE = "AZURE_RESOURCE_GROUP_USAGE";
 
     public static final String AZURE_RESOURCE_GROUP_USAGE_SINGLE = "SINGLE";
+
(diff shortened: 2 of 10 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/PrivateEndpointTest.java
```diff
@@ -0,0 +1,13 @@
+package com.sequenceiq.it.cloudbreak;
+
+public class PrivateEndpointTest {
```
aszegedi, line 3: Usually we `extends GherkinTest` at this level of the project. If you do not judge, this is not worth it, we can leave this at it is.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/ResourceGroupTest.java
```diff
@@ -5,10 +5,17 @@
 
     public static final String AZURE_RESOURCE_GROUP_USAGE_SINGLE = "SINGLE";
 
+    public static final String AZURE_RESOURCE_GROUP_USAGE_SINGLE_DEDICATED = "SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT";
+
     public static final String AZURE_RESOURCE_GROUP_USAGE_MULTIPLE = "MULTIPLE";
 
     public static final String AZURE_RESOURCE_GROUP_NAME = "AZURE_RESOURCE_GROUP_NAME";
 
     private ResourceGroupTest() {
```
aszegedi, line 14: Usually we `extends GherkinTest` at this level of the project. If you do not judge, this is not worth it, we can leave this at it is.

integration-test/src/main/resources/application.yml
```diff
@@ -69,6 +69,8 @@ integrationtest:
   upgrade:
     currentRuntimeVersion: 7.2.0
     targetRuntimeVersion: 7.2.6
+  privateendpoint:
+    enabled: false
```
aszegedi, line 73: If `privateEndpoint` won't contain any further parameters, you can set this directly:
```
privateEndpoint: false
```

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/PrivateEndpointTest.java
gergopapi2: I see. What is the benefit of extending the GherkinTest class? I see that the methods of the GherkinTest class are used solely in RepositoryConfigTests.

integration-test/src/main/resources/application.yml
gergopapi2: I changed it to privateEndpointEnabled, would that be ok?

aszegedi (approved):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/PrivateEndpointTest.java
gergopapi2: Ok, you convinced me via chat, I change it. It helps the organization.

PR #9623, opened 2020-12-10
CB-10086 Azure disk can be grown to 32767 Gb now. By default we will …

doktoric:
…also restrict the UI for this new number. Unfortunately this big disk can not be cached in case of unmanaged disks so the cacheing now is None because wee are getting this error unless: Only Disk CachingType 'None' is supported for disk with size greater than 4095 GB


biharitomi (approved):

pdarvasi (approved):

pdarvasi:
@doktoric I've retriggered the swagger test

PR #9625, opened 2020-12-10
CB-10278 Permission check should run only before cloudbreak controlle…

horadla23:
…r methods, not before and after every controller method (like spring controllers)


horadla23:
http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/2268/

foldik (approved):
lgtm

PR #9626, opened 2020-12-10
CB-9711 - OS upgrade for datahubs should work

pdarvasi:
Data Hub OS upgrades are supported via cli with " --lock-components --replace-vms ENABLED"


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/UpgradeDistroxFlowEventChainFactory.java
```diff
@@ -14,30 +21,58 @@
...
         if (event.isReplaceVms()) {
-            // TODO
-            LOGGER.warn("Replace VMs for DistroX is not implemented!");
+            Map<String, List<String>> nodeMap = getNodeMap(event);
+            chain.add(new ClusterRepairTriggerEvent(FlowChainTriggers.CLUSTER_REPAIR_TRIGGER_EVENT, event.getResourceId(), nodeMap, false, true));
         }
         return chain;
     }
+
+    private Map<String, List<String>> getNodeMap(DistroxUpgradeTriggerEvent event) {
```
tiborpopovics, line 62: Can you please rename this method to something that is describing the return value? For example `getInstancesByHostGroup` or sometging like that.

keyki (approved):

tiborpopovics (approved):

PR #9627, opened 2020-12-10
CB-10284 Update Azure coordinates got from the az client

daszabo:


doktoric:
I think this should be 2.35 or master

PR #9629, opened 2020-12-11
CB-10290: Fix the FreeIPA log message for the old instance's status

jamisonbennett:
Fix the FreeIPA log message for the old instance's status.

This was manually tested with a local deployment of cloudbreak.


lacikaaa (approved):

christmasferret:
@topolyai5   Just FYI the original commit was 17074f15753983d284dc9fb425ccf03098c9c157

PR #9630, opened 2020-12-11
CB-10291 & CB-10292: Fix the FreeIPA health check API

jamisonbennett:
Fix FreeIPA health summary status on the health API. The summary
correctly reports unreachable, stopped, available, and unhealthy now.

Fix the FreeIPA health check API's instance status when the health
check fails.

Populate the missing instance ID for the FreeIPA health check.

Tested with unit tests and also with a local instance of cloudbreak.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
@@ -1,9 +1,12 @@
 package com.sequenceiq.freeipa.service.stack;
 
+import static java.util.function.Predicate.not;
+
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaHealthDetailsService.java
```diff
@@ -109,17 +111,29 @@ private void addUnreachableResponse(InstanceMetaData instance, HealthDetailsFree
...
+                .filter(not(InstanceMetaData::isTerminated))
+                .map(InstanceMetaData::getInstanceId)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+        List<InstanceStatus> nonTerminatedStatuses = response.getNodeHealthDetails().stream()
+                .filter(nodeHealthDetails -> notTermiatedStackInstanceIds.contains(nodeHealthDetails.getInstanceId()))
+                .map(NodeHealthDetails::getStatus)
+                .collect(Collectors.toList());
+        if (!nonTerminatedStatuses.isEmpty() &&
+                areAllStatusTheSame(nonTerminatedStatuses) && !hasMissingStatus(nonTerminatedStatuses, notTermiatedStackInstanceIds)) {
```
lacikaaa, line 133: we have 3 checks here, either of them fails we will set the status to unhealthy. Will it be obvious from the logs while we set it to unhealthy or shall we add some extra log here?
I'm just worried we will have to dig deep why it's unhealthy, but maybe it's covered by other logs.

lacikaaa (approved):

PR #9632, opened 2020-12-11
CB-10299 modify BaseImageValidatorE2ETest case not to config explicitnew network for SDX internal calls

biharitomi:
This is fixing the base image one for now.
The prewarmed based test works on my machine for GCP. So this is the only fix that is needed.


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/stack/StackTestDtoBase.java
```diff
@@ -263,6 +263,11 @@ private InstanceGroupV4Request getInstanceGroupV4Request(String key) {
         return this;
     }
 
+    public StackTestDtoBase<T> withNetworkOfEnvironment() {
```
pkedvessy, line 266: Can we have only one method for setting the network to null?
withEmtyNetwork does exactly the same

pkedvessy (changes requested):

pkedvessy (approved):

PR #9634, opened 2020-12-11
CB-9972 If freeipa is in unreachable state, badgateway http status code will return instead of internal server error in the freeipa endpoint responses

schfeca75:
Whenever the FreeIPA instance is not reachable and we know it's not reachable because the status indicates it, we're going to return 502 Bad Gateway.

According to the official explanation of the status code:
The server was acting as a gateway or proxy and received an invalid response from the upstream server.

If we don't know that the IPA server is not reachable, because we think it's available, that will remain an Internal Server Error.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/mapper/FreeipaClientExceptionWrapperMapper.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/mapper/FreeipaClientExceptionWrapperMapper.java
@@ -1,0 +1,27 @@
+package com.sequenceiq.freeipa.controller.mapper;
+
+import javax.ws.rs.core.Response.Status;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.common.exception.ExceptionResponse;
+import com.sequenceiq.freeipa.client.FreeIpaClientExceptionWrapper;
+
+@Component
+public class FreeipaClientExceptionWrapperMapper extends BaseExceptionMapper<FreeIpaClientExceptionWrapper> {
+
+    @Override
+    protected Object getEntity(FreeIpaClientExceptionWrapper exception) {
+        return new ExceptionResponse("Error during interaction with FreeIPA: " + exception.getWrappedException().getMessage());
+    }
+
+    @Override
+    Status getResponseStatus() {
+        return Status.INTERNAL_SERVER_ERROR;
+    }
+
+    @Override
+    Class<FreeIpaClientExceptionWrapper> getExceptionType() {
+        return FreeIpaClientExceptionWrapper.class;
+    }
+}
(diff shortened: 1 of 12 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
```diff
@@ -228,10 +229,10 @@ private FreeIpaClientBuilder getFreeIpaClientBuilderForDirectMode(Stack stack, I
         return new FreeIpaClientBuilder(ADMIN_USER, freeIpa.getAdminPassword(), httpClientConfig, gatewayPort, instanceMetaData.getDiscoveryFQDN(), tracer);
     }
 
-    private FreeIpaClientException createFreeIpaStateIsInvalidException(Status stackStatus) {
+    private InvalidFreeIpaStateException createFreeIpaStateIsInvalidException(Status stackStatus) {
         String message = String.format("Couldn't build FreeIPA client. Because FreeIPA is in invalid state: '%s'", stackStatus);
         LOGGER.warn(message);
-        return new FreeIpaClientException(message, new FreeIpaHostNotAvailableException(message));
+        return new InvalidFreeIpaStateException(message, new FreeIpaHostNotAvailableException(message));
```
jamisonbennett, line 235: 1. I am pretty sure this doesn't interfere with the retries, so it should be good. i.e.
```
@Retryable(value = RetryableFreeIpaClientException.class,
            maxAttemptsExpression = RetryableFreeIpaClientException.MAX_RETRIES_EXPRESSION,
            backoff = @Backoff(delayExpression = RetryableFreeIpaClientException.DELAY_EXPRESSION,
                    multiplierExpression = RetryableFreeIpaClientException.MULTIPLIER_EXPRESSION))
```
should work since this isn't one of the cases where FreeIpaClientExceptions can be converted/wrapped with RetryableFreeIpaClientException.


2. Did you consider line 130? I think this might also need to have to switch to the InvalidFreeIpaStateException.
```
                return client.orElseThrow(() -> createFreeIpaUnableToBuildClient(new FreeIpaHostNotAvailableException("No FreeIPA client was available")));
```

freeipa/src/main/java/com/sequenceiq/freeipa/controller/mapper/FreeipaClientExceptionWrapperMapper.java
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.ws.rs.core.Response.Status;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.common.exception.ExceptionResponse;
+import com.sequenceiq.freeipa.client.FreeIpaClientExceptionWrapper;
+
+@Component
+public class FreeipaClientExceptionWrapperMapper extends BaseExceptionMapper<FreeIpaClientExceptionWrapper> {
```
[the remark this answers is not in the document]
lacikaaa: we have `FreeipaClientExceptionMapper` which handles both `FreeIpaClientExcpetion/RetryableFreeIpaClientException`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/FreeIpaClientFactory.java
lacikaaa: 2. for now we would like to handle those cases separately where the status of the FreeIPA is already known to be wrong. I don't think I have seen this issue yet. Still, I removed the callt ot `createFreeIpaUnableToBuildClient` as there is a catch reight after which already calls that

lacikaaa:
@jamisonbennett  as both @keyki and @schfeca75 are on PTO I had to take ownership of this PR. Could you do the review for it? thanks

jamisonbennett (approved):

PR #9642, opened 2020-12-11
CB-10320: Do not enable HBase on S3 for 7.2.6 when Raz is enabled

LinaAtAustin:
RAZ is enabled on 7.2.6, but the code change for HBase feature supporting RAZ is only on 7.2.7. Therefore, only enable HBase cloud storage feature on 7.2.7+.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CMRepositoryVersionUtil.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CMRepositoryVersionUtil.java
@@ -32,6 +32,8 @@ public class CMRepositoryVersionUtil {
 
     public static final Versioned CLOUDERAMANAGER_VERSION_7_2_6 = () -> "7.2.6";
 
+    public static final Versioned CLOUDERAMANAGER_VERSION_7_2_7 = () -> "7.2.7";
+
     public static final Versioned CFM_VERSION_2_0_0_0 = () -> "2.0.0.0";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CMRepositoryVersionUtil.class);
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
@@ -1,6 +1,6 @@
 package com.sequenceiq.cloudbreak.cmtemplate.configproviders.hbase;
 
-import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_2_2;
+import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_2_7;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.isVersionNewerOrEqualThanLimited;
 import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.config;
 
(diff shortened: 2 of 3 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java
```diff
@@ -54,13 +54,13 @@ public boolean isConfigurationNeeded(CmTemplateProcessor cmTemplateProcessor, Te
...
         String cdhVersion = getCdhVersion(source);
-        boolean is722OrNewer = isVersionNewerOrEqualThanLimited(cdhVersion, CLOUDERAMANAGER_VERSION_7_2_2);
+        boolean is727OrNewer = isVersionNewerOrEqualThanLimited(cdhVersion, CLOUDERAMANAGER_VERSION_7_2_7);
         String accountId = ThreadBasedUserCrnProvider.getAccountId();
         boolean sdxHbaseCloudStorageEnabled =
                 entitlementService.sdxHbaseCloudStorageEnabled(accountId);
         return source.getFileSystemConfigurationView().isPresent()
                 && cmTemplateProcessor.isRoleTypePresentInService(getServiceType(), getRoleTypes())
-                && (!datalakeCluster || (is722OrNewer && sdxHbaseCloudStorageEnabled));
+                && (!datalakeCluster || (is727OrNewer && sdxHbaseCloudStorageEnabled));
```
lajosrodek, line 63: Nit: What about UT? If there is no associated test class, it might make sense to add one now.

LinaAtAustin: UT is updated.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CMRepositoryVersionUtil.java
```diff
@@ -32,6 +32,8 @@
 
     public static final Versioned CLOUDERAMANAGER_VERSION_7_2_6 = () -> "7.2.6";
 
+    public static final Versioned CLOUDERAMANAGER_VERSION_7_2_7 = () -> "7.2.7";
```
kmanamcheri, line 35: mmm so we are switching to Runtime and CM versions being different starting from 7.2.7. I

The question now becomes if you should check for the CM version or Cloudera Runtime (CDH) version. If it is Cloudera Runtime version, then we should update the check on runtime (instead of CM).

If this should be a CM check, then there is no CM 7.2.7. After CM 7.2.6, it becomes CM 7.3.0. You should follow up with [CRB-1455](https://jira.cloudera.com/browse/CRB-1455) to understand more about how to check runtime version vs. CM version.

kmanamcheri (dismissed):

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CMRepositoryVersionUtil.java
LinaAtAustin: This is not a problem
1) I get the runtime version from blueprint as input of the check, not CM version. See https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java#L66
2) Although the constant used is defined in the CM utility, we are comparing the stack runtime version against a constant. So the check is still valid even when CM version is different from runtime version. What matters is the meaning of the input variable. see https://github.com/hortonworks/cloudbreak/blob/master/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hbase/HbaseCloudStorageServiceConfigProvider.java#L57

kmanamcheri: Ah that makes sense. Thanks for the information @LinaAtAustin 

since we are checking for the runtime version and not the CM version, lets change the constant name to "CLOUDERA_RUNTIE_VERSION_7_2_7", so its not confusing for future maintainers.

bergerdenes (approved):
lgtm

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CMRepositoryVersionUtil.java
bergerdenes: @kmanamcheri  please open a backburner ticket for this.

kmanamcheri: ok done https://jira.cloudera.com/browse/CB-10353

PR #9648, opened 2020-12-12
CB-10324: Add entitlement for EFS in CloudBreak

LinaAtAustin:
Add entitlement CDP_DATA_LAKE_AWS_EFS in CloudBreak, so we can hide the EFS related code before it is fully functional.

This is blocked by https://github.infra.cloudera.com/thunderhead/thunderhead/pull/3031, which should be committed first, then commit this change.


mock-thunderhead/src/main/resources/application.yml
```diff
@@ -29,6 +29,7 @@ auth:
     mediumdutysdx.enable: true
     upgrade.internalrepo.enable: true
     hbase.cloudstorage.enable: true
+    datalake.efs.enable: true
```
doktoric, line 32: this means that locally we will always use EFS. This is what you want to achieve here ?

LinaAtAustin: I can turn it off, and only enable in my local dev.

bergerdenes:
Do you have a counterpart entitlement in the thunderhead as well?

LinaAtAustin:
Yes. It is in https://jira.cloudera.com/browse/ALTCORE-3497, already committed. Its code review is at https://github.infra.cloudera.com/thunderhead/thunderhead/pull/3031

bergerdenes (commented):
lgtm

bergerdenes (approved):

PR #9650, opened 2020-12-14
CB-9744 Export Ranger on gateway for SDX HA.

orngejaket:
This exports just the Ranger nodes for Knox that are on the gateway nodes.
This also includes a change to Stack to persist the SDX shape and carry it
across to the Salt config.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -601,7 +601,8 @@ public class ClusterHostServiceRunner {
         ExposedService rangerService = exposedServiceCollector.getRangerService();
         List<String> rangerLocations = serviceLocations.get(rangerService.getServiceName());
         if (!CollectionUtils.isEmpty(rangerLocations)) {
-            serviceLocations.put(rangerService.getServiceName(), getSingleRangerFqdn(gatewayConfig.getHostname(), rangerLocations));
+            List<String> rangerGatewayHosts = getRangerFqdn(cluster, gatewayConfig.getHostname(), rangerLocations);
+            serviceLocations.put(rangerService.getServiceName(), rangerGatewayHosts);
         }
         serviceLocations.put(exposedServiceCollector.getClouderaManagerService().getServiceName(), asList(gatewayConfig.getHostname()));
         gateway.put("location", serviceLocations);
@@ -654,7 +655,14 @@ public class ClusterHostServiceRunner {
         return componentLocation;
     }
 
-    private List<String> getSingleRangerFqdn(String primaryGatewayFqdn, List<String> rangerLocations) {
+    private List<String> getRangerFqdn(Cluster cluster, String primaryGatewayFqdn, List<String> rangerLocations) {
+        if (rangerLocations.size() > 1) {
+            String gatewayGroupName = cluster.getStack().getPrimaryGatewayInstance().getInstanceGroup().getGroupName();
+            List<String> hosts = rangerLocations.stream()
+                    .filter(s -> s.contains(gatewayGroupName))
+                    .collect(Collectors.toList());
+            return hosts;
+        }
         return rangerLocations.contains(primaryGatewayFqdn) ? asList(primaryGatewayFqdn) : asList(rangerLocations.iterator().next());
     }
 


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -600,8 +601,13 @@ private void saveGatewayPillar(GatewayConfig gatewayConfig, Cluster cluster,
 
         ExposedService rangerService = exposedServiceCollector.getRangerService();
         List<String> rangerLocations = serviceLocations.get(rangerService.getServiceName());
+
         if (!CollectionUtils.isEmpty(rangerLocations)) {
-            serviceLocations.put(rangerService.getServiceName(), getSingleRangerFqdn(gatewayConfig.getHostname(), rangerLocations));
+            if (SdxClusterShape.LIGHT_DUTY.equals(cluster.getStack().getSdxClusterShape())) {
+                serviceLocations.put(rangerService.getServiceName(), getSingleRangerFqdn(gatewayConfig.getHostname(), rangerLocations));
+            } else if (SdxClusterShape.MEDIUM_DUTY_HA.equals(cluster.getStack().getSdxClusterShape())) {
+                serviceLocations.put(rangerService.getServiceName(), rangerLocations.stream().filter(s -> s.contains("gateway")).collect(Collectors.toList()));
```
lacikaaa, line 609: this doesn't seems to be a good long term solution. I would suggest to fetch the gateways' hostnames/fqdns and filter in them instead of the magic `gateway` string

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/SdxClusterShape.java
```diff
@@ -0,0 +1,8 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.common;
+
+public enum SdxClusterShape {
```
lacikaaa, line 3: as this must match with the other `SdxClusterShape` it must be ensured via unit tests

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -600,8 +601,13 @@ private void saveGatewayPillar(GatewayConfig gatewayConfig, Cluster cluster,
 
         ExposedService rangerService = exposedServiceCollector.getRangerService();
         List<String> rangerLocations = serviceLocations.get(rangerService.getServiceName());
+
         if (!CollectionUtils.isEmpty(rangerLocations)) {
-            serviceLocations.put(rangerService.getServiceName(), getSingleRangerFqdn(gatewayConfig.getHostname(), rangerLocations));
+            if (SdxClusterShape.LIGHT_DUTY.equals(cluster.getStack().getSdxClusterShape())) {
+                serviceLocations.put(rangerService.getServiceName(), getSingleRangerFqdn(gatewayConfig.getHostname(), rangerLocations));
+            } else if (SdxClusterShape.MEDIUM_DUTY_HA.equals(cluster.getStack().getSdxClusterShape())) {
```
lacikaaa, line 608: is it possible to have multiple gateway node for `LIGHT_DUTY`? I mean do we need this `if-else` at all? couldn't be it like all gateway node has ranger and `LIGHT_DUTY` has only one GW while `MEDIUM` has multiple and the result would be the same. In this case we won't need the cluster shape on CB side

lacikaaa (changes requested):

horadla23 (commented):
during sdx creeation, cb double checks sdx in datalake service in StackService (`sdxClientService.getByCrn(externalCrn.get());`) so I would not store clustershape twice, but get it from response and pass  to the flow (i guess `saveGatewayPillar` method called from a flow step).
also it does look odd in case of datahub cluster.
since it doesn't change after creation, it's not a big deal, but if it is possible, i would avoid to store same data in two services.
what do you think about this?

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
orngejaket: No, medium has multiple nodes called gateway, light duty has none.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
orngejaket: The gateways are called gateway in Medium duty.  But I can change it to that instead.

core-api/src/main/java/com/sequenceiq/cloudbreak/doc/ModelDescriptions.java
```diff
@@ -241,6 +241,7 @@ private ModelDescriptions() {
         public static final String TUNNEL = "Configuration that the connection going directly or with cluster proxy or with ccm and cluster proxy.";
         public static final String FLOW_ID = "Flow identifier for the current stack creation. Only returned during the stack create request/response.";
         public static final String EXTERNAL_DATABASE = "External database parameters for the stack.";
+        public static final String SDX_CLUSTER_SHAPE = "SDX Cluster Shape";
```
hreeve-cloudera, line 244: This is a very basic description. Can you add more information, like explaining what a cluster shape is?

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
hreeve-cloudera: I think you can filter on `stack.getPrimaryGatewayInstance().getInstanceGroup().getGroupName()`. For medium duty the group name should be "gateway", and for light duty it would be "master".

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
lacikaaa: but in case of `LIGHT_DUTY` the master is `GATEWAY` type, while in case of `MEDIUM_DUTY_HA` not the master is the gateway. We could depend on that information and select those FQDNs. `Stack` has a method: `getGatewayInstanceMetadata` which could be used to get all the gateway instances.

orngejaket:
I've refactored and greatly simplified this based on @hreeve-cloudera's feedback.

lacikaaa (approved):
LGTM, @horadla23 please check

lacikaaa:
retest this please

horadla23 (approved):

lacikaaa (commented):
Looks like it's breaking int tests, most likely `String gatewayGroupName = cluster.getStack().getPrimaryGatewayInstance().getInstanceGroup().getGroupName();` is causing a null pointer

PR #9651, opened 2020-12-14
CB-9745 Enabling Knox sticky sessions for ui and api requests.

orngejaket:
This also enables round robining to all the hosts in the topology,
and should not affect Knox services that do not have this enabled.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -258,11 +258,11 @@
              {% if 'RANGER' in exposed and 'RANGER_ADMIN' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>RANGER</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enableStickySession=false;noFallback=false;enableLoadBalancing=true</value>
              </param>
              <param>
                  <name>RANGERUI</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enableStickySession=true;noFallback=true;enableLoadBalancing=true</value>
              </param>
              {%- endif %}
              {% if 'YARNUIV2' in exposed and 'RESOURCEMANAGER' in salt['pillar.get']('gateway:location') -%}
@@ -278,7 +278,7 @@
              {% if 'SOLR' in exposed and 'SOLR_SERVER' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>SOLR</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enableStickySession=true;noFallback=true;enableLoadBalancing=true</value>
              </param>
              {%- endif %}
              {% if 'JOBHISTORYUI' in exposed and 'JOBHISTORY' in salt['pillar.get']('gateway:location') -%}
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology_api.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology_api.xml.j2
@@ -165,7 +165,7 @@
             {% if 'HIVE' in exposed and 'HIVESERVER2' in salt['pillar.get']('gateway:location') -%}
             <param>
                 <name>HIVE</name>
-                <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                <value>enableStickySession=true;noFallback=true;enableLoadBalancing=true</value>
             </param>
             {%- endif %}
             {% if 'LIVYSERVER_API' in exposed and 'LIVY_SERVER' in salt['pillar.get']('gateway:location') -%}
@@ -195,7 +195,7 @@
             {% if 'RANGER' in exposed and 'RANGER_ADMIN' in salt['pillar.get']('gateway:location') -%}
             <param>
                 <name>RANGER</name>
-                <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                <value>enableStickySession=false;noFallback=false;enableLoadBalancing=true</value>
             </param>
             {%- endif %}
             {% if 'YARNUIV2' in exposed and 'RESOURCEMANAGER' in salt['pillar.get']('gateway:location') -%}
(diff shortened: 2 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -198,21 +198,21 @@
              {% if 'ATLAS' in exposed and 'ATLAS_SERVER' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>ATLAS</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000;enableStickySessions=true</value>
```
kmanamcheri, line 201: Is this enabled for all runtime versions? Is there a way to enable this only for 7.2.7 and above?

kmanamcheri (commented):
I don't see any particular change which enables Knox round-robin load balancing. Is that automatically enabled if you set `enableStickySessions=true` ?

Are there any other side effects of `enableStickySessions=true`? Was there a reason we didn't enable that before? Is this a new property?

moresandeep:
> I don't see any particular change which enables Knox round-robin load balancing. Is that automatically enabled if you set `enableStickySessions=true` ?
> 
> Are there any other side effects of `enableStickySessions=true`? Was there a reason we didn't enable that before? Is this a new property?

Yes this is a new feature https://knox.apache.org/books/knox-1-5-0/user-guide.html#Default+Service+HA+support
By default with `enableStickySessions=true` roundrobin is turned on you can explicitly turn it off though with `enableLoadBalancing` property.

moresandeep:
@orngejaket was this feature tested? I would like this to be tested before we turn it on.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -198,21 +198,21 @@
...
              </param>
              <param>
                  <name>CM-UI</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000;enableStickySessions=true</value>
              </param>
              {%- endif %}
              {% if 'HIVE' in exposed and 'HIVESERVER2' in salt['pillar.get']('gateway:location') -%}
```
moresandeep, line 218: Why are we not this for HIVE? Hive would be one to benefit the most.

orngejaket: Hive is connected to by the compute engines, so it shouldn't need this?  Or should it?  I've only done this for the ui/api services.

Also it's possible that the datahub would use those configurations as well, I'm not sure if they should be supported?

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
```diff
@@ -198,21 +198,21 @@
              {% if 'ATLAS' in exposed and 'ATLAS_SERVER' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>ATLAS</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000;enableStickySessions=true</value>
              </param>
              <param>
                  <name>ATLAS-API</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000;enableStickySessions=true</value>
```
moresandeep, line 205: This would break Atlas, we should not be using Knox HA for atlas since it is active-passive. adding stickysession would break it badly.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
moresandeep: The original bug was opened for Hive and I tested this, like dev tested it using beeline for hive.

kmanamcheri: @orngejaket We want to change only for Datalake right? Is the current code change applicable for Datalake or Datahubs as well?

@moresandeep I understand that the original one was opened for Hive, but we'd like to use it for Atlas/Ranger and the other components in the datalake.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
orngejaket: All run times, knox should ignore it if it's not supported.  There's no way to make it 7.2.7+ only.

kmanamcheri:
> > I don't see any particular change which enables Knox round-robin load balancing. Is that automatically enabled if you set `enableStickySessions=true` ?
> > Are there any other side effects of `enableStickySessions=true`? Was there a reason we didn't enable that before? Is this a new property?
> 
> Yes this is a new feature https://knox.apache.org/books/knox-1-5-0/user-guide.html#Default+Service+HA+support
> By default with `enableStickySessions=true` roundrobin is turned on you can explicitly turn it off though with `enableLoadBalancing` property.

@moresandeep Can we only enable the load balancing without the sticky sessions?

moresandeep:
> > > I don't see any particular change which enables Knox round-robin load balancing. Is that automatically enabled if you set `enableStickySessions=true` ?
> > > Are there any other side effects of `enableStickySessions=true`? Was there a reason we didn't enable that before? Is this a new property?
> > 
> > 
> > Yes this is a new feature https://knox.apache.org/books/knox-1-5-0/user-guide.html#Default+Service+HA+support
> > By default with `enableStickySessions=true` roundrobin is turned on you can explicitly turn it off though with `enableLoadBalancing` property.
> 
> @moresandeep Can we only enable the load balancing without the sticky sessions?
Sure, we can using `enableLoadBalancing` flag

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
moresandeep: We should be careful with this, there are some edge cases that could cause issues with retries. 
Also, Atlas is active/passive so HA will not work there.

orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
orngejaket: Ok so no changes to Atlas then?

orngejaket:
I've updated the review based on QA feedback and suggested settings for the services.

moresandeep (approved):
Looks good.

hreeve-cloudera (approved):

PR #9657, opened 2020-12-15
CB-9926: Add retries to the legacy health check's ping

jamisonbennett:
Add retries to the legacy health check's ping.


jamisonbennett:
@lacikaaa I marked this DO_NOT_MERGE. I would like to run some more manual tests since I am changing `legacyParseMessages()`. I want to check a few failure cases to ensure correctness before this gets merged. But I thought it is best to start a review. I don't really like the retry helper class, do you have better suggestions for a better way?

lacikaaa:
yes, that doesn't seems the best way. I think you have to move out that part of the code to a separate class. Otherwise you can selfinject, but that's not better.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaInstanceHealthDetailsService.java
```diff
@@ -0,0 +1,161 @@
...
+        nodeResponse.setStatus(rpcResponse.getResult() ? InstanceStatus.CREATED : InstanceStatus.UNHEALTHY);
+        nodeResponse.setName(instanceMetaData.getDiscoveryFQDN());
+        nodeResponse.setInstanceId(instanceMetaData.getInstanceId());
+        boolean found = false;
+        for (RPCMessage message : rpcResponse.getMessages()) {
+            Matcher nodeMatcher = NEW_NODE_PATTERN.matcher(message.getMessage());
+            if (nodeMatcher.find()) {
+                found = true;
+            }
+            if (!found) {
```
christmasferret, line 133: _found_ variable is not necessary.  use  (!nodeMatcher.find()) should be good enough.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/FreeIpaInstanceHealthDetailsService.java
```diff
@@ -0,0 +1,161 @@
...
+                // appears in the preceding message.
+                if (EXTERNAL_COMMAND_OUTPUT.equals(message.getName())) {
+                    Matcher matcher = RESULT_PATTERN.matcher(message.getMessage());
+                    if (matcher.find()) {
+                        if (!STATUS_OK.equals(matcher.group(STATUS_GROUP))) {
+                            nodeResponse.addIssue(precedingMessage);
+                        }
+                    }
+                    precedingMessage = message.getMessage();
+                }
```
christmasferret, line 146: precedingMessage = message.getMessage(); will have no effect here as mutating string will not change the string in nodeResponse above. Could append the message before addIssue()

jamisonbennett:
I had to drop the retires from 3 to 2 for the health check so that it didn't exceed the RPC timeout. I will remove the DO_NOT_MERGE tag now.

freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -169,7 +166,7 @@ public FreeIpaClient build(boolean withPing) throws URISyntaxException, IOExcept
                 additionalHeaders.forEach(request::addHeader);
                 additionalHeadersStickySessionFirstRpc.forEach(request::addHeader);
                 try (CloseableHttpResponse response = client.execute(request)) {
-                    if (UNAVIALLBE_PING_HTTP_RESPONSES.contains(response.getStatusLine().getStatusCode())) {
+                    if (isUnreachableHttpStatus(response.getStatusLine().getStatusCode())) {
```
christmasferret, line 169: It is strange that the catch clause after this is LOGGER.debug. Might be LOGGER.error()

catch (Exception e) {
                String msg = "Ping failed";
                LOGGER.debug(msg, e);

jamisonbennett: I will break this part out into a separate PR.

lacikaaa (approved):

jamisonbennett:
When I force-pushed the CB-9926 branch from af76d4b to 3957e5f, this closed the PR by accident. I was trying to rebase to the lastest.... I will open a new PR.

PR #9661, opened 2020-12-16
CB-10299 limit the length of the GCP buckets as it could not be longer than 63 characters

biharitomi:


bergerdenes (approved):
lgtm

cegganesh84:
Nice catch @biharitomi, thanks for fixing this!

PR #9663, opened 2020-12-16
CB-10351: Improve exception logging when aws api throws exception during Rds creation

tiborpopovics:
Currently when the stackDescribe API call failing the code is ignoring the AmazonServiceException. Therefore we don't what was the error.
In this fix I've added exception logging and I extended our CloudConnectorException with the original exception as a root cause.


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsRdsLaunchService.java
```diff
@@ -88,8 +87,9 @@
         } catch (AmazonServiceException ignored) {
```
topolyai5, line 87: it is not ignored anymore. Could you please rename it?

topolyai5 (approved):

lacikaaa (approved):

PR #9668, opened 2020-12-16
CB-8937: ipa-custodia and ipa-dnskeysyncd restart after 60 seconds

christmasferret:
ipa-custodia and ipa-dnskeysyncd now will automatically restart after 60 seconds with failure exit code. We change it to 3 seconds and always to be consistent with all other services.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/services.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/services.sls
@@ -1,3 +1,31 @@
+ipadnskeysyncdRestart:
+   file.replace:
+    - name: /usr/lib/systemd/system/ipa-dnskeysyncd.service
+    - pattern: '^Restart=.*'
+    - repl: 'Restart=always'
+    - backup: False
+
+ipadnskeysyncdRestartSec:
+   file.replace:
+     - name: /usr/lib/systemd/system/ipa-dnskeysyncd.service
+     - pattern: '^RestartSec=.*'
+     - repl: 'RestartSec=3'
+     - backup: False
+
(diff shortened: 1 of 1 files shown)


christmasferret:
Screen Shot 2020-12-16 at 4 18 58 PM
Screen Shot 2020-12-16 at 4 19 14 PM


Tested with kill and restart.

lacikaaa (commented):
Is there a reason you haven't extended the list in the pillar instead of adding new states?
https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/freeipa-salt/pillar/freeipa/services.sls

christmasferret:
> Is there a reason you haven't extended the list in the pillar instead of adding new states?
> https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/freeipa-salt/pillar/freeipa/services.sls

@lacikaaa Now services salt are adding 
**Restart=always
RestartSec=3**
after the line **[Service]**

It is creation not replacement. 

So if I add them to services, it will become settings as below

[Unit]
Description=IPA Custodia Service

[Service]
**Restart=always
RestartSec=3**
Type=notify
ExecStart=/usr/libexec/ipa/ipa-custodia /etc/ipa/custodia/custodia.conf
PrivateTmp=yes
**Restart=on-failure
RestartSec=60s**

[Install]
WantedBy=multi-user.target


Here, I am doing a replacement of settings in ipa-custodiaj and ipa-dnskeysyncd. 

Besides, there is already customized setting for dirSrv in the same file already. It is OK to introduce a customized setting for ipa services.

lacikaaa (approved):

PR #9671, opened 2020-12-17
CB-10406. Diagnostics: Support destination for DL/DH fails with clust…

oleewere:
…er not found error.

details:
- this happens if the cluster id is not the same as the stack id


horadla23 (approved):
LGTM

PR #9675, opened 2020-12-21
CB-10095 Subnet selection for public gateway endpoint

hreeve-cloudera:
Adds new logic to create a public load balancer to act as a public gateway to
the Knox endpoints in an otherwise private system. Logic will be triggered if
the "publicEndpointAccessGateway" flag in the environment request is set to
ENABLED, and the tenant has the new CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY
entitlement.

A list of subnets to use for the public LB can optionally be provided as part
of the environment creation request. If provided, any private subnets in the
list will be discarded, and a public subnet in the same AZ as the selected stack
subnet will be used. If the endpoint subnets aren't explicitly provided, the
subnets provided for environment creation are used instead. If public subnets in
the appropriate availability zones are not provided, creation will fail
immediately.

Tested with unit tests and by running CB locally with hardcoded API fields to
verify cloud resources were created and configured appropriately.


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/EnvironmentBaseNetworkConverter.java
```diff
@@ -32,34 +34,37 @@ public Network convertToLegacyNetwork(EnvironmentNetworkResponse source, String
...
+                subnetsToParse = source.getSubnetMetas();
+            } else {
+                subnetsToParse = source.getGatewayEndpointSubnetMetas();
+            }
+            Map<String, CloudSubnet> publicSubnetMetas = subnetsToParse.entrySet().stream()
+                .filter(entry -> !entry.getValue().isPrivateSubnet())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+            Optional<CloudSubnet> endpointGatewayCloudSubnet = chooseSubnet(null,
+                publicSubnetMetas, selectedAZ, false);
```
enis, line 56: Should we still use the `selectedAZ` here? The selectedAZ can be coming from the network request, but the user maybe explicitly specifying the subnet to use for the endpoint gateway anyways if provided. I was thinking that the provided getGatewayEndpointSubnetMetas will always take precedence.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkView.java
```diff
@@ -55,6 +57,23 @@ public boolean isSubnetList() {
         return isSubnetList() ? List.of(getExistingSubnet().split(",")) : (isExistingSubnet() ? List.of(getExistingSubnet()) : List.of());
     }
 
+    private String getEndpointGatewaySubnet() {
+        return network.getStringParameter(ENDPOINT_GATEWAY_SUBNET_ID);
+    }
+
+    public boolean isEndpointGatewaySubnetSet() {
```
enis, line 64: Seeing the name of this function together with the one below, I was just confused why we are having a check for both a `Set` (as in HashSet) and a `List` check. It turns out it is a check for existence. 

Maybe we can merge the List vs non-list cases to simplify. If there is only 1 subnet, then it is a list of 1.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/NetworkCreationHandler.java
```diff
@@ -114,6 +126,50 @@ private void setChildEnvironmentNetworkIfItHasParentWithTheSameCloudProvider(Env
...
+            if (subnetMetas != null && !subnetMetas.isEmpty()) {
+                if (endpointGatewaySubnetMetas != null && !endpointGatewaySubnetMetas.isEmpty()) {
+                    LOGGER.debug("Verifying that provided endpoint gateway subnets share availability zones with provided environment subnets.");
+                    Set<String> subnetAZs = subnetMetas.values().stream()
+                        .map(CloudSubnet::getAvailabilityZone)
+                        .collect(Collectors.toSet());
+                    Set<String> endpointGatewaySubnetAZs = endpointGatewaySubnetMetas.values().stream()
+                        .map(CloudSubnet::getAvailabilityZone)
+                        .collect(Collectors.toSet());
+                    if (!subnetAZs.equals(endpointGatewaySubnetAZs)) {
```
enis, line 152: Should this check be a subset check rather than equals() check? We can have env with 3 subnets, but endpoint gateway with 1 subnet, no?

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/NetworkCreationHandler.java
```diff
@@ -114,6 +126,50 @@ private void setChildEnvironmentNetworkIfItHasParentWithTheSameCloudProvider(Env
...
+                    LOGGER.debug("Verifying that public subnets in availability zones that match the private subnets were provided.");
+                    Set<String> privateAZs = subnetMetas.values().stream()
+                        .filter(CloudSubnet::isPrivateSubnet)
+                        .map(CloudSubnet::getAvailabilityZone)
+                        .collect(Collectors.toSet());
+                    Set<String> publicAZs = subnetMetas.values().stream()
+                        .filter(subnet -> !subnet.isPrivateSubnet())
+                        .map(CloudSubnet::getAvailabilityZone)
+                        .collect(Collectors.toSet());
+                    if (!privateAZs.equals(publicAZs)) {
```
enis, line 165: same here.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkView.java
hreeve-cloudera: I can rename the function to make it more clear. I don't really want to change the methods themselves because they were written following the same pattern as is used for the environment subnet list.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/EnvironmentBaseNetworkConverter.java
hreeve-cloudera: I'd argue against the getGatewayEndpointSubnetMetas taking precedence, since their use case is very limited vs. the private subnets which are used more widely. But we do need to include the selected AZ, because the private and public subnets needs to be in the same AZ for the load balancer to work.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/NetworkCreationHandler.java
hreeve-cloudera: No, we can't. The load balancer needs subnets in the same AZs as all the EC2 instances it's routing to. So if our private subnets are in az-1, az-2, and az-3, and our public subnets are in az-1 and az-2, then if an EC2 instance is created in az-3 it's completely inaccessible to the load balancer. In order to prevent that possibility, we require the users to provide both private and public subnets in all the same AZs, which is what this check is doing here.

enis: My understanding is that the initial set of subnets given at the time of the env creation can contain more subnets than actually gets used for DL/DH instances (like subnets for dwx, liftie, etc). In DH you can select which subnet to use, but apparently not for DL? 

Maybe this is fine, as long as somebody from CB who knows the subnet selection logic takes a look at it.

hreeve-cloudera: That's true. Not every subnet in the lists being processed here will be used. Which is really why I have this logic. At this stage I don't know which of the subnets will be selected, so to make sure there won't be any hiccups in the eventual subnet selection logic I just make sure that we have subnets with matching AZs here and fail immediately if not.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
```diff
@@ -39,9 +41,16 @@
 
     private String endpoint;
 
+    /**
+     * @deprecated Use {@link #targetGroupSet} instead.
+     */
+    @Deprecated
     @OneToMany(mappedBy = "loadBalancer", cascade = CascadeType.REMOVE, orphanRemoval = true)
     private Set<TargetGroup> targetGroups = new HashSet<>();
```
lacikaaa, line 49: can't we just remove these fields from the entity and leave them in the database. As you wrote somewhere these are not really used in any environment yet. If we can't remove them please create a followup jira to remove them in the next release, if you haven't created one already. thanks

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/EnvironmentBaseNetworkConverter.java
```diff
@@ -32,34 +34,37 @@ public Network convertToLegacyNetwork(EnvironmentNetworkResponse source, String
...
+            if (endpointGatewayCloudSubnet.isEmpty()) {
+                throw new BadRequestException("Could not find public subnet in availability zone: " + availabilityZone);
+            }
+            LOGGER.debug("Chosen endpoint gateway subnet: {}", endpointGatewayCloudSubnet.get());
+            attributes.put("endpointGatewaySubnetId", endpointGatewayCloudSubnet.get().getId());
+        }
```
lacikaaa, line 62: could you move it to a new method called sth like chooseSubnetForEndpointGateway?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -349,39 +350,79 @@ private void validateStackAuthentication(StackV4Request source) {
         return convertedSet;
     }
 
-    private Set<LoadBalancer> createLoadBalancers(StackV4Request source, Stack stack) {
+    private Set<LoadBalancer> createLoadBalancers(StackV4Request source, Stack stack, DetailedEnvironmentResponse environment) {
```
lacikaaa, line 353: I think it would be useful to add some extra log lines, as there are a several if and it would help us investigate later

lacikaaa (changes requested):

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
bergerdenes: because of continuous deployment, we need to keep backward and forward compatibility
so a follow-up jira is indeed needed to remove these.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
```diff
@@ -256,32 +254,59 @@ private void updateCloudformationWithLoadBalancers(AuthenticatedContext ac, Clou
         }
     }
 
-    private AwsLoadBalancerScheme determinePublicVsPrivateSchema(AwsNetworkView awsNetworkView, AmazonEC2Client amazonEC2Client) {
-        String subnetId = awsNetworkView.getExistingSubnet();
-        String vpcId = awsNetworkView.getExistingVpc();
+    private void convertLoadBalancer(CloudLoadBalancer cloudLoadBalancer, List<CloudResource> instances,
```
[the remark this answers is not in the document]
hreeve-cloudera: It converting from `CloudLoadBalancer` to `AwsLoadBalancer`. So in that sense it was always creating an instance of `AwsLoadBalancer` and setting its properties based on the properties of `CloudLoadBalancer`. That piece of the logic hasn't really changed. I will update it to return the `AwsLoadBalancer` and add it to the list in the parent method instead of here.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/loadbalancer/AwsTargetGroup.java
```diff
@@ -1,28 +1,24 @@
...
     private String arn;
 
-    public AwsTargetGroup(int port, AwsLoadBalancerScheme scheme, int order, List<String> instanceIds) {
+    public AwsTargetGroup(AwsLoadBalancerScheme scheme, int port) {
         this.port = port;
-        this.order = order;
-        this.instanceIds = instanceIds;
-        name = getTargetGroupName(port, scheme);
+        this.name = getTargetGroupName(port, scheme);
+        this.instanceIds = new HashSet<>();
```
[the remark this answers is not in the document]
hreeve-cloudera: By field level, do you mean:
```
private final Set<String> instanceIds = new HashSet<>();
```

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
hreeve-cloudera: Right. Originally I had deleted them, but I had a talk with Arpad about the compatibility requirements and that led me to adding them back and marking them deprecated. I'll open another JIRA to clean these methods up once this PR is in.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
hreeve-cloudera: Please review this method again and let me know if the new log statements cover everything you think we'll need.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkView.java
hreeve-cloudera: Changed this method name to `containsEndpointGatewaySubnet`.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -305,28 +307,26 @@
...
+          "TargetType" : "instance",
+          <#if existingVPC>
+          "VpcId" : { "Ref" : "VPCId" }
+          <#else>
+          "VpcId" : { "Ref" : "VPC" }
+          </#if>
+          <#if listener.targetGroup.instanceIds?size != 0 >
+          ,"Targets" : [
+            <#list listener.targetGroup.instanceIds as i>
+              { "Id" : "${i}" }<#if (i_index + 1) != listener.targetGroup.instanceIds?size>,</#if>
```
bergerdenes, line 324: <#sep>,

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetaDataCollectorTest.java
```diff
@@ -323,4 +389,18 @@ private AuthenticatedContext authenticatedContext() {
         authenticatedContext.putParameter(AmazonEC2Client.class, amazonEC2Client);
         return authenticatedContext;
     }
+
+    private void setupMethodsForLoadBalancer() {
+        LoadBalancer internalLoadBalancer = new LoadBalancer()
+            .withDNSName("internal-lb.aws.dns")
```
bergerdenes, line 395: minor: use string constants instead and reuse them in assertions

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
```diff
@@ -183,6 +184,10 @@ public boolean datalakeLoadBalancerEnabled(String accountId) {
         return isEntitlementRegistered(accountId, CDP_DATA_LAKE_LOAD_BALANCER);
     }
 
+    public boolean publicEndpointAccessGatewayEnabled(String accountId) {
```
bergerdenes, line 187: observation: the entitlement flag is not used extensively, that is, a lot of changed functionality cannot be switched off as-is

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/view/AwsNetworkViewTest.java
```diff
@@ -121,4 +122,20 @@ public void testMultipleSubnetCidrEmpty() {
         assertTrue(underTest.getExistingVpcCidrs().contains("1.1.1.1"));
     }
 
+    @Test
+    public void testSingleEndpointGatewaySubnet() {
+        when(network.getStringParameter(ENDPOINT_GATEWAY_SUBNET_ID)).thenReturn("subnet-123");
```
bergerdenes, line 127: minor: please use constants

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/NetworkCreationHandlerTest.java
```diff
@@ -0,0 +1,241 @@
...
+        Map<String, CloudSubnet> subnets = createDefaultPrivateSubnets();
+        subnets.putAll(createDefaultPublicSubnets());
+
+        when(environmentService.findEnvironmentById(any())).thenReturn(environmentOptional);
+        when(cloudNetworkService.retrieveSubnetMetadata(any(EnvironmentDto.class), any())).thenReturn(subnets);
+        when(cloudNetworkService.retrieveEndpointGatewaySubnetMetadata(any(EnvironmentDto.class), any())).thenReturn(Map.of());
+        when(environmentResourceService.createAndSetNetwork(any(), any(), any(), any(), any())).thenReturn(network);
+
+        underTest.accept(environmentDtoEvent);
+    }
```
bergerdenes, line 141: missing assertions

environment/src/test/java/com/sequenceiq/environment/network/service/SubnetIdProviderTest.java
```diff
@@ -161,6 +163,82 @@ public void shouldReturnNullInCaseOfNonSupportedCloudPlatform() {
...
+        NetworkDto networkDto = NetworkDto.builder()
+            .withSubnetMetas(subnets)
+            .withCbSubnets(subnets)
+            .withEndpointGatewaySubnetMetas(publicSubnets)
+            .build();
+
+        String actual = underTest.provideEndpointGateway(networkDto, CloudPlatform.AWS, "id-2");
+
+        Assertions.assertNotNull(actual);
+    }
```
bergerdenes, line 185: assert on actual values?
same for all tests below

bergerdenes (changes requested):

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
```diff
@@ -94,14 +103,34 @@ public void setEndpoint(String endpoint) {
...
+    /**
+     * @deprecated Use {@link #setTargetGroupSet()} instead.
+     */
+    @Deprecated
     public void setTargetGroups(Set<TargetGroup> targetGroups) {
         this.targetGroups = targetGroups;
     }
 
+    public Set<TargetGroup> getTargetGroupSet() {
+        return targetGroupSet;
```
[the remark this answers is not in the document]
hreeve-cloudera: I've handled this by setting both the new and the old values in the new setter methods. For example:
```
    public void setTargetGroupSet(Set<TargetGroup> targetGroups) {
        this.targetGroupSet = targetGroups;
        this.targetGroups = targetGroups;
    }
```
I think that'll work, but let me know if that's incorrect.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
hreeve-cloudera: I've added the entitlement check anywhere we're setting the endpoint gateway related fields. If the original environment request has those fields set but the entitlement is disabled, then it will unset them before processing the request. I also added a lot of unit tests to verify the behavior when the entitlement is disabled and enabled.

core/src/main/resources/schema/app/20201210210553_CB-10095_Add_multiple_load_balancers_to_target_group.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE IF NOT EXISTS targetgroup_loadbalancer (
+    targetgroupset_id     bigint NOT NULL,
+    loadbalancerset_id    bigint NOT NULL
+);
+
+ALTER TABLE ONLY targetgroup_loadbalancer ADD CONSTRAINT targetgroup_loadbalancer_pkey PRIMARY KEY (targetgroupset_id, loadbalancerset_id);
+ALTER TABLE ONLY targetgroup_loadbalancer ADD CONSTRAINT fk_targetgroup_loadbalancer_targetgroup_id FOREIGN KEY (targetgroupset_id) REFERENCES targetgroup(id);
+ALTER TABLE ONLY targetgroup_loadbalancer ADD CONSTRAINT fk_targetgroup_loadbalancer_loadbalancer_id FOREIGN KEY (loadbalancerset_id) REFERENCES loadbalancer(id);
+
+INSERT INTO targetgroup_loadbalancer (SELECT id, loadbalancer_id FROM targetgroup);
```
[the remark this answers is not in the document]
bergerdenes: We need to do them in separate release versions (like if you this PR will be incorporated into 2.37 then the second phase is 2.38 and the column removal is 2.39). 
As far as I know, for zero downtime migration, this is the only way.

bergerdenes (approved):
LGTM

core/src/main/resources/schema/app/20201210210553_CB-10095_Add_multiple_load_balancers_to_target_group.sql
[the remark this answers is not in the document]
hreeve-cloudera: Noted. I'll create the JIRAs and target them for the appropriate releases.

lacikaaa (approved):

lacikaaa:
please be sure you have only **one commit** on this branch

hreeve-cloudera:
> please be sure you have only **one commit** on this branch

@lacikaaa I only have one commit for the change. The other two commits are from rebases. I don't see those commit hashes in the git log and am not sure how or if I can squash them into one commit.

lacikaaa:
I see 3 commits in the PR, 2 of them merging master to your branch. maybe you merged master to your branch via github. We usually rebase our branches locally and force push:
- `git fetch`
- `git rebase origin/master`
- `git push -f origin <branch>`
This way you should have an up-to-date branch without conflict. If you have conflicts you have to resolve them during the rebase.

hreeve-cloudera:
@lacikaaa That is generally what I do. I think what might have happened with those other two commits is that I had smalll merge conflicts, and clicked the button in the PR to resolve it via the github UI instead of doing it locally. I now know not to do that.

Let me try rebasing locally and doing another force push, and see if that clears up those other two commits.

PR #9680, opened 2020-12-22
CB-10411: Add FreeIPA backup debug to logs

jamisonbennett:
Add FreeIPA backup debug to salt output and /var/log/ipabackup.log.
Include redaction of the x-amz-security-token.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/backups.sls
@@ -13,6 +13,14 @@
     - group: root
     - mode: 640
 
+/usr/local/bin/backup-log-filter.sh:
+  file.managed:
+    - makedirs: True
+    - user: root
+    - group: root
+    - mode: 700
+    - source: salt://freeipa/scripts/backup-log-filter.sh
+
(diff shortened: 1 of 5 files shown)


jamisonbennett:
I added lot rotation in case the hourly backups fail to upload the backup to the storage location. In that case, the logs will grow since there will be more backups that are attempted to be synced every hour until it is successful. It appears that the logs may grow pretty quickly during failure cases since every hour it will attempt to sync all backups that haven't been successfully uploaded. A provision failure due to permissions issues resulted in 16 MB of logs (with the provision retries there were about a dozen full backups that were attempted to be uploaded about dozen times).

lacikaaa (approved):
LGTM, only the space seems a bit much for the log. KDC log already can take 12GB, if both take it we would might need instances with bigger root disks.(I think it's 50GB right now)

pdarvasi:
> LGTM, only the space seems a bit much for the log. KDC log already can take 12GB, if both take it we would might need instances with bigger root disks.(I think it's 50GB right now)

We might consider decreasing the "rotate 12" to a lower value

pdarvasi (approved):

jamisonbennett:
I decreased the log retention to less than 1 GB total.

PR #9682, opened 2020-12-23
CB-10440: Fix target for cname records to allow external references

jamisonbennett:
The FreeIPA management service API for adding cname DNS records did
not support external FQDNs for the target. This was fixed.

This was tested by looking up the DNS record in a local deployment of
cloudbreak.


hreeve-cloudera (approved):

jamisonbennett:
@lacikaaa can you please also confirm that you agree with this change as it does change the behavior of the API?

We have not released since the original cname API was added so we don't really need to worry about backward compatibility outside of mow-dev and right now the only consumer is @hreeve-cloudera for registering the load balancers. We need external domain names for the load balancers.

lacikaaa (approved):
good catch! thanks!

PR #9683, opened 2020-12-23
CB-9746: Adding LoadBalancer FQDN endpoints to the datalake services.

sxxgrc:
This PR addresses the following Jiras:
https://jira.cloudera.com/browse/CB-9746
https://jira.cloudera.com/browse/CB-10123

Specifically, I am attempting to change the endpoints assigned to all of the Datalake services from being constructed with a Knox URL to being constructed with the LoadBalancer URL.

The end-goal is for this change to be reflected across both the CDP UI and any other way of attempting to access the individual service endpoints

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -6,6 +6,7 @@ import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUD
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.isVersionNewerOrEqualThanLimited;
 import static com.sequenceiq.cloudbreak.core.bootstrap.service.ClusterDeletionBasedExitCriteriaModel.clusterDeletionBasedModel;
 import static java.util.Collections.singletonMap;
+import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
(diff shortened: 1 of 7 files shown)


core/src/main/resources/schema/app/20201221232206_CB-9746:_Use_new_loadbalancer_endpoints_for_DL_services.sql
```diff
@@ -0,0 +1,10 @@
+-- // CB-9746: Use new loadbalancer endpoints for DL services
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE loadbalancer ADD COLUMN IF NOT EXISTS fqdn character varying(255) DEFAULT 'No Info';
+UPDATE loadbalancer SET fqdn = 'No Info' WHERE fqdn IS NULL;
```
hreeve-cloudera, line 5: I don't think we should have the default be a string value. You're not checking the value of the load balancer FQDN name anywhere that I saw, so it seems like the code could see the string "No Info" and think that's a perfectly valid domain. I don't think we need a default value, since if it isn't set it should just be null, and then the statements that check if it's null will correctly evaluate to false.

mock-thunderhead/src/main/resources/application.yml
```diff
@@ -34,4 +34,4 @@ auth:
     database.wire.encryption.enable: true
     datahub.runtime.upgrade.enable: true
     embedded.database.on.attached.disk.enable: true
-    datalake.loadbalancer.enable: false
+    datalake.loadbalancer.enable: true
```
hreeve-cloudera, line 37: I've been asked by the CB team to not enable this by default, since it only really works if you're using mow-dev UMS and most people won't be.

hreeve-cloudera (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -629,7 +630,19 @@ private void addGatewayUserFacingCertAndFqdn(GatewayConfig gatewayConfig, Cluste
             gateway.put("userfacingkey", cluster.getStack().getSecurityConfig().getUserFacingKey());
             gateway.put("userfacingcert", cluster.getStack().getSecurityConfig().getUserFacingCert());
         }
-        String fqdn = cluster.getFqdn();
+
+        String fqdn = null;
+        if (cluster.getStack().getLoadBalancers() != null && cluster.getStack().getLoadBalancers().size() > 0) {
+            fqdn = (cluster.getStack().getLoadBalancers().stream()
+                    .filter(lb -> lb.getType() == LoadBalancerType.PUBLIC).findAny()
+                    .orElse(cluster.getStack().getLoadBalancers().iterator().next()))
```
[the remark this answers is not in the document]
orngejaket: LoadBalanceryType.PUBLIC.equals(lb.getType())

orngejaket (changes requested):
Similar to hreeve's comments.

hreeve-cloudera (approved):

orngejaket (approved):

PR #9687, opened 2021-01-04
CB-10447 DisableCheckPermissions isn't handled correctly in Permissio…

horadla23:
…nCheckService

### DIFF
--- a/authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
+++ b/authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.authorization.service;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.lang.annotation.Annotation;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


horadla23:
we can merge this after 2.36 branch cutting if that is better

authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
```diff
@@ -91,6 +92,16 @@ public Object hasPermission(ProceedingJoinPoint proceedingJoinPoint) {
         return commonPermissionCheckingUtils.proceed(proceedingJoinPoint, methodSignature, startTime);
     }
 
+    private Optional<String> getInitiatorUserCrn(ProceedingJoinPoint proceedingJoinPoint, MethodSignature methodSignature) {
```
lturcsanyi, line 95: The method name could be confusing, because it persists the initiator user as well.

lturcsanyi (approved):

authorization-common/src/main/java/com/sequenceiq/authorization/service/PermissionCheckService.java
horadla23: right, i'll rename it to make it easier to understand

PR #9690, opened 2021-01-04
CB-10394 CB-10392 added proxy and network information to central EDH …

doktoric:
…metrics. We are using CDPNetworkDetails now as other places using it as well. Added unit test as well to cover use cases.

### DIFF
--- a/usage-collection/src/main/proto/usage.proto
+++ b/usage-collection/src/main/proto/usage.proto
@@ -980,8 +980,8 @@ message CDPEnvironmentDetails {
     int32 numberOfAvailabilityZones = 3;
     // Number of availability zones used in the environment
     string availabilityZones = 4;
-    // Tunnel type of the environment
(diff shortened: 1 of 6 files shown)


usage-collection/src/main/proto/usage.proto
```diff
@@ -980,8 +980,8 @@ message CDPEnvironmentDetails {
     int32 numberOfAvailabilityZones = 3;
     // Number of availability zones used in the environment
     string availabilityZones = 4;
-    // Tunnel type of the environment
-    CDPEnvironmentsEnvironmentTunnel.Value tunnelType = 5;
+    // Network metadata
+    CDPNetworkDetails networkDetails = 5;
```
akanto, line 984: This looks like an incompatible change, I think you cannot submit this into thunderhead repo anymore, and this will break EDH as well.

PR #9693, opened 2021-01-04
CB-10098 Register internal load balancer DNS with FreeIPA

hreeve-cloudera:
Adds a new step to cluster creation to bootstrap the load balancer's domain
name in FreeIPA. If more than one load balancer is present, then the logic will
prefer the private load balancer to register, and will fall back to the public
load balancer if that's all that's available.

Includes changes to make all load balancer endpoints match, and updates to the
Route 53 LB domain registration. The Route 53 registration is an inverse of the
FreeIPA, where it prefer the public load balancer if it exists and uses the
private load balancer as a fallback.

Tested with unit tests and manually by creating a data lake and looking up the
LB entry in FreeIPA.


core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+        }
+    }
+
+    private String getDomainFromStack(Stack stack) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        DetailedEnvironmentResponse environment = environmentClientService.getByCrn(stack.getEnvironmentCrn());
+        return getDomainNameProvider().getDomainName(environment.getName(), getWorkloadSubdomain(userCrn));
+    }
```
lacikaaa, line 104: I'm not sure if you really need this. If this domain is the same which is handled by FreeIPA you could leave that field blank and FreeIPA will take care of it.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationActions.java
```diff
@@ -153,11 +154,27 @@ protected Selectable createRequest(StackContext context) {
         };
     }
 
+    @Bean(name = "BOOTSTRAPPING_FREEIPA_ENDPOINT_STATE")
+    public Action<?, ?> bootStrappingPrivateEndpointAction() {
+        return new AbstractStackCreationAction<>(BootstrapPublicEndpointSuccess.class) {
+            @Override
+            protected void doExecute(StackContext context, BootstrapPublicEndpointSuccess payload, Map<Object, Object> variables) {
+                clusterCreationService.bootstrapPrivateEndpoints(context.getStack());
```
lacikaaa, line 162: if an exception is thrown here it is not handled

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/GatewayPublicEndpointManagementService.java
```diff
@@ -276,4 +287,16 @@ private String getEndpointNameForStack(Stack stack) {
...
+        Optional<LoadBalancer> loadBalancerOptional = loadBalancers.stream()
+            .filter(lb -> lb.getType() == LoadBalancerType.PUBLIC)
+            .findFirst();
+        if (loadBalancerOptional.isEmpty()) {
+            loadBalancerOptional = loadBalancers.stream()
+                .filter(lb -> lb.getType() == LoadBalancerType.PRIVATE)
+                .findFirst();
+        }
+        return loadBalancerOptional;
+    }
```
lacikaaa, line 301: this method looks like a duplication except the order is changed. Could you generalize this?

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+    private Optional<LoadBalancer> selectLoadBalancer(Set<LoadBalancer> loadBalancers) {
+        Optional<LoadBalancer> loadBalancerOptional = loadBalancers.stream()
+            .filter(lb -> lb.getType() == LoadBalancerType.PRIVATE)
+            .findFirst();
+        if (loadBalancerOptional.isEmpty()) {
+            loadBalancerOptional = loadBalancers.stream()
+                .filter(lb -> lb.getType() == LoadBalancerType.PUBLIC)
+                .findFirst();
+        }
+        return loadBalancerOptional;
```
lacikaaa, line 115: result should be logged

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+                    request.setTargetFqdn(StringUtils.appendIfMissing(loadBalancer.getDns(), DOMAIN_PART_DELIMITER));
+                    request.setEnvironmentCrn(stack.getEnvironmentCrn());
+                    dnsV1Endpoint.addDnsCnameRecord(request);
+                } else if (StringUtils.isNotEmpty(loadBalancer.getIp())) {
+                    AddDnsARecordRequest request = new AddDnsARecordRequest();
+                    request.setHostname(endpoint);
+                    request.setDnsZone(domain);
+                    request.setIp(loadBalancer.getIp());
+                    request.setEnvironmentCrn(stack.getEnvironmentCrn());
+                    dnsV1Endpoint.addDnsARecord(request);
```
lacikaaa, line 67: I think we should log the requests

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                LoadBalancer loadBalancer = loadBalancerOptional.get();
+                String endpoint = loadBalancer.getEndpoint();
+                String domain = getDomainFromStack(stack);
+
+                if (StringUtils.isNotEmpty(loadBalancer.getDns())) {
+                    dnsV1Endpoint.deleteDnsCnameRecord(stack.getEnvironmentCrn(), domain, endpoint);
+                } else if (StringUtils.isNotEmpty(loadBalancer.getIp())) {
+                    dnsV1Endpoint.deleteDnsARecord(stack.getEnvironmentCrn(), domain, endpoint);
```
lacikaaa, line 92: I think it would be nice to add a log line so we would know from the logs which one was called

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+import com.sequenceiq.cloudbreak.service.environment.EnvironmentClientService;
+import com.sequenceiq.cloudbreak.service.stack.LoadBalancerPersistenceService;
+import com.sequenceiq.common.api.type.LoadBalancerType;
+import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
+import com.sequenceiq.freeipa.api.v1.dns.DnsV1Endpoint;
+import com.sequenceiq.freeipa.api.v1.dns.model.AddDnsARecordRequest;
+import com.sequenceiq.freeipa.api.v1.dns.model.AddDnsCnameRecordRequest;
+
+@Service
+public class FreeIPAEndpointManagementService extends BasePublicEndpointManagementService {
```
lacikaaa, line 25: `DnsV1Endpoint` expects environment edit right from the user right now. If a datahub admin or similar can launch cluster who doesn't have env edit right I think these calls would fail.
I think a good solution would be to make these calls as internal actor.
cc @horadla23 @keyki @jamisonbennett

lacikaaa (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
horadla23: precisely, please use `ThreadBasedUserCrnProvider.doAsInternActor(() -> method)`

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -59,14 +56,11 @@ private boolean isKnoxGatewayDefinedInServices(Set<ServiceComponent> serviceComp
             .anyMatch(serviceComponent -> KnoxRoles.KNOX_GATEWAY.equals(serviceComponent.getComponent()));
     }
 
-    public String generateLoadBalancerEndpoint(Stack stack, LoadBalancerType type) {
+    public String generateLoadBalancerEndpoint(Stack stack) {
         StringBuilder name = new StringBuilder()
             .append(stack.getName())
             .append('-')
             .append(ENDPOINT_SUFFIX);
-        if (LoadBalancerType.PUBLIC.equals(type)) {
```
doktoric, line 67: unit test would be nice here

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private String getDomainFromStack(Stack stack) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        DetailedEnvironmentResponse environment = environmentClientService.getByCrn(stack.getEnvironmentCrn());
+        return getDomainNameProvider().getDomainName(environment.getName(), getWorkloadSubdomain(userCrn));
+    }
+
+    private Optional<LoadBalancer> selectLoadBalancer(Set<LoadBalancer> loadBalancers) {
+        Optional<LoadBalancer> loadBalancerOptional = loadBalancers.stream()
+            .filter(lb -> lb.getType() == LoadBalancerType.PRIVATE)
```
doktoric, line 108: we are using here lb.getType() == LoadBalancerType.PRIVATE other places using equals in your PR. I think it would be nice to use the same mechanism.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private String getDomainFromStack(Stack stack) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        DetailedEnvironmentResponse environment = environmentClientService.getByCrn(stack.getEnvironmentCrn());
+        return getDomainNameProvider().getDomainName(environment.getName(), getWorkloadSubdomain(userCrn));
+    }
+
+    private Optional<LoadBalancer> selectLoadBalancer(Set<LoadBalancer> loadBalancers) {
+        Optional<LoadBalancer> loadBalancerOptional = loadBalancers.stream()
```
doktoric, line 107: privateLoadBalancerOptional ?

doktoric: I would actually do a java service which can select a private or a public loadbalancer based on the parameter

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/GatewayPublicEndpointManagementService.java
```diff
@@ -276,4 +287,16 @@ private String getEndpointNameForStack(Stack stack) {
         return loadBalancerPersistenceService.findByStackId(stack.getId()).stream()
             .map(LoadBalancer::getEndpoint).collect(Collectors.toSet());
     }
+
+    private Optional<LoadBalancer> selectLoadBalancer(Set<LoadBalancer> loadBalancers) {
+        Optional<LoadBalancer> loadBalancerOptional = loadBalancers.stream()
```
doktoric, line 292: dulplicated logic

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/GatewayPublicEndpointManagementService.java
```diff
@@ -172,7 +178,12 @@ public void deleteLoadBalancerDnsEntry(Stack stack, String environmentName) {
             environmentName = environment.getName();
         }
 
-        for (LoadBalancer loadBalancer : loadBalancers) {
+        Optional<LoadBalancer> loadBalancerOptional = selectLoadBalancer(loadBalancers);
```
doktoric, line 181: this logic seems to be duplicated. I think it would be beneficial the refactor it

core/src/test/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+    public void testRegisterLoadBalancerWithDNS() {
+        LoadBalancer loadBalancer = new LoadBalancer();
+        loadBalancer.setType(LoadBalancerType.PUBLIC);
+        loadBalancer.setDns(LB_DNS);
+        loadBalancer.setEndpoint(LB_ENDPOINT);
+        when(loadBalancerPersistenceService.findByStackId(any())).thenReturn(Set.of(loadBalancer));
+
+        ThreadBasedUserCrnProvider.doAs(TEST_USER_CRN, () -> underTest.registerLoadBalancerDomainWithFreeIPA(stack));
+
+        ArgumentCaptor<AddDnsCnameRecordRequest> requestCaptor = ArgumentCaptor.forClass(AddDnsCnameRecordRequest.class);
```
doktoric, line 102: this really jut a little thing but which one is more correct AddDnsCnameRecordRequest vs AddDns**C**NameRecordRequest 
@lacikaaa

lacikaaa: CNAME is the original format and I follow the guideline when all the letters are capital, in the variable/class etc only the first one is uppercase. eg:
FreeIPA->FreeIpa
DNS->Dns
IPAddress->IpAddress

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/GatewayPublicEndpointManagementService.java
hreeve-cloudera: Per Mihaly's comment below, I've removed this and the other `selectLoadBalancer` method and replaced them with a more generalized method in `LoadBalancerConfigService`.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: This method has been moved to `LoadBalancerConfigService`.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: I'm also not sure I need it either. But I don't think it hurts to have it and I'd rather err on the side of making sure it's defined explicitly instead of hoping it's the same domain as in FreeIPA.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/GatewayPublicEndpointManagementService.java
hreeve-cloudera: Moved to `LoadBalancerConfigService`.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationActions.java
hreeve-cloudera: I've added a try/catch to the methods being called in `FreeIPAEndpointManagementService`, since what's being called here is just a series of wrappers that ends up in `FreeIPAEndpointManagementService`.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: I've put these wrappers in `ClusterPublicEndpointManagementService`, where it's calling this service. For example:
````
ThreadBasedUserCrnProvider.doAsInternalActor(() ->
  freeIPAEndpointManagementService.deleteLoadBalancerDomainFromFreeIPA(stack));
````

File: core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java (hunk 8 above)
Comment (hreeve-cloudera) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java:67]: I've created a unit test class for this service, and added test cases for some of the other logic. I should have had a unit test for the previous version of the method. But now that it's just a string appender I'm not sure what would be tested here. The previous logic wasn't actually used anywhere, since we don't actually use public load balancers anywhere. (Public load balancers are being added in https://github.com/hortonworks/cloudbreak/pull/9675, although the enums have been renamed there.)

File: core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java [hunk 14]
```diff
@@ -0,0 +1,121 @@
...
+    @Inject
+    private EnvironmentClientService environmentClientService;
+
+    @Inject
+    private LoadBalancerPersistenceService loadBalancerPersistenceService;
+
+    @Inject
+    private LoadBalancerConfigService loadBalancerConfigService;
+
+    public void registerLoadBalancerDomainWithFreeIPA(Stack stack) {
```
bergerdenes, line 44: this method is too complex

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,121 @@
...
+                        LOGGER.error("Unable to register load balancer with FreeIPA", e);
+                    }
+                } else {
+                    LOGGER.error("Unable to determine load balancer type. Load balancer will not be registered with FreeIPA.");
+                }
+            }
+        }
+    }
+
+    public void deleteLoadBalancerDomainFromFreeIPA(Stack stack) {
```
bergerdenes, line 86: this method is too complex

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: Refactored.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: Refactored.

doktoric (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
lacikaaa: I'm not sure this is a good idea to change those instead of the calls which are really meant to be internal. What do you think @horadla23 ?

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
lacikaaa: then I think you should invest more time into this: what kind of domain could be returned by this code, can it be different than the domain in freeipa? can it be different than `Stack` entity domain field? If we can't tell why we need this and how it should work, I won't expect us to understand it a few month later.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
horadla23: in long term i have to agree with @lacikaaa . if someone changes service logic and wants to get the initiator user crn for example, then it will return with false result. the best is to change only the endpoint calls to internal

horadla23: also please check if DnsV1Controller is prepared for internal calls based on the readme I provided (meaning that the proper parameter is annotated to be able to get current account id instead of static account id of internal actor). For me it seems it is not, and the DnsV1Endpoint call will fail some accountid issue

lacikaaa: I will do the freeipa part in `DnsV1Controller` to work with internal actor asap

lacikaaa: on PR: https://github.com/hortonworks/cloudbreak/pull/9740

hreeve-cloudera: @lacikaaa @horadla23 I've read the linked document. I'm afraid I couldn't quite follow it though, and am not sure what I should be doing.

The registration method is only called by `StackV4Controller`, and the delete method looks like it could be triggered by either `StackV4Controller` or `DistroXV1Controller`. Registration is triggered by a `post` or `postInternal` call, and deletion is triggered by the various `delete*` method calls. I see in `StackV4Controller` that the methods that will trigger my code all have either a `@AccountId` or `@InitiatorUserCrn` annotation. This isn't true for the `DistroXV1Controller` calls. So I'm not sure exactly what changes I need to make.

Mihaly, you said:

> I'm not sure this is a good idea to change those instead of the calls which are really meant to be internal.

Which "calls which are really meant to be internal" are you referring to exactly? Do you mean the calls directly in the `FreeIPAEndpointManagementService`? Would it be more correct to move the `doAsInternalActor` to be inside those methods, instead of around the code making the method call?

lacikaaa: yes, you should add doAsInternal around `dnsV1Endpoint` eg:
```
ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsCnameRecord(request));

File: core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java (hunk 1 above)
Comment (hreeve-cloudera) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java]: I talked to Jamison about how the DNS zones work. I've removed this method and am not passing the dnsZone in the requests anymore. I ran a test and everything is still being registered correctly without explicitly setting the dnsZone.

File: core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java [hunk 16]
```diff
@@ -0,0 +1,128 @@
...
+    private void sendAddDnsARecordRequest(Stack stack, LoadBalancer loadBalancer) {
+        String ip = loadBalancer.getIp();
+        String endpoint = loadBalancer.getEndpoint();
+
+        AddDnsARecordRequest request = new AddDnsARecordRequest();
+        request.setHostname(endpoint);
+        request.setIp(ip);
+        request.setEnvironmentCrn(stack.getEnvironmentCrn());
+        LOGGER.debug("Registering load balancer with target IP {} in FreeIPA with A record {}", ip, endpoint);
+        ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsARecord(request));
```
lacikaaa, line 84: this should call `addDnsARecordInternal`

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -0,0 +1,128 @@
...
+    private void sendAddDnsCnameRecordRequest(Stack stack, LoadBalancer loadBalancer) {
+        String targetFQDN = StringUtils.appendIfMissing(loadBalancer.getDns(), DOMAIN_PART_DELIMITER);
+        String endpoint = loadBalancer.getEndpoint();
+
+        AddDnsCnameRecordRequest request = new AddDnsCnameRecordRequest();
+        request.setCname(endpoint);
+        request.setTargetFqdn(targetFQDN);
+        request.setEnvironmentCrn(stack.getEnvironmentCrn());
+        LOGGER.debug("Registering load balancer with target FQDN {} in FreeIPA with CNAME {}", targetFQDN, endpoint);
+        ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsCnameRecord(request));
```
lacikaaa, line 72: this should call `addDnsCnameRecordInternal`

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: So should the full call look like:
```
ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsCnameRecordInternal(ThreadBasedUserCrnProvider.getAccountId(), request));
```

lacikaaa: yes, except you switched the method names. it should be for the A record

hreeve-cloudera: Right, sorry, copy/pasted from the wrong line. Should be updated now.

lacikaaa (approved):
LGTM
please test with the latest changes before we merge

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
horadla23: sorry, i haven't noticed it earlier, the call should look like this:
```
ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsARecordInternal(accountId, request));
```

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
horadla23:
```
ThreadBasedUserCrnProvider.doAsInternalActor(() -> dnsV1Endpoint.addDnsCnameRecordInternal(accountId, request));
```

horadla23 (changes requested):
added comments to existing conversations about accountId usage

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: Done. I'll run another test to make sure it works before merging it in, but the code changes have been pushed for review.

horadla23 (approved):

PR #9694, opened 2021-01-05
CB-10096 Expand Knox load balancer to data hubs

enis:
Added a new entitlement for the public / private load balancer to the datahubs. This is separate
entitlement than the datalake one to err on the consevative side.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
@@ -380,15 +380,22 @@ public class StackV4RequestToStackConverter extends AbstractConversionServiceAwa
     }
 
     private boolean shouldCreateInternalKnoxLoadBalancer(StackType type, EnvironmentNetworkResponse network) {
-        // TODO expand this to data hubs
+        return shouldCreateInternalKnoxLoadBalancerForDatalake(type, network) ||
+            shouldCreateInternalKnoxLoadBalancerForDatahub(type, network);
+    }
+
+    private boolean shouldCreateInternalKnoxLoadBalancerForDatalake(StackType type, EnvironmentNetworkResponse network) {
         return StackType.DATALAKE.equals(type) &&
             (entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId())
             || isEndpointGatewayEnabled(network));
     }
 
+    private boolean shouldCreateInternalKnoxLoadBalancerForDatahub(StackType type, EnvironmentNetworkResponse network) {
+        return  StackType.WORKLOAD.equals(type) && isEndpointGatewayEnabled(network);
+    }
+
     private boolean shouldCreateExternalKnoxLoadBalancer(StackType type, EnvironmentNetworkResponse network) {
-        // TODO expand this to data hubs
-        return StackType.DATALAKE.equals(type) && isEndpointGatewayEnabled(network);
+        return (StackType.DATALAKE.equals(type) || StackType.WORKLOAD.equals(type)) && isEndpointGatewayEnabled(network);
     }
 
     private boolean isEndpointGatewayEnabled(EnvironmentNetworkResponse network) {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -38,7 +38,7 @@ public class LoadBalancerConfigService {
     private LoadBalancerPersistenceService loadBalancerPersistenceService;
 
     public Set<String> getKnoxGatewayGroups(Stack stack) {
-        LOGGER.info("Fetching list of instance groups with Knox gateway installed");
+        LOGGER.debug("Fetching list of instance groups with Knox gateway installed");
         Set<String> groupNames = new HashSet<>();
         Cluster cluster = stack.getCluster();
         if (cluster != null) {
@@ -48,7 +48,7 @@ public class LoadBalancerConfigService {
         }
 
         if (groupNames.isEmpty()) {
-            LOGGER.info("Knox gateway is not explicitly defined; searching for CM gateway hosts");
+            LOGGER.debug("Knox gateway is not explicitly defined; searching for CM gateway hosts");
             groupNames = stack.getInstanceGroups().stream()
                 .filter(i -> InstanceGroupType.isGateway(i.getInstanceGroupType()))
                 .map(InstanceGroup::getGroupName)


keyki (approved):

lacikaaa (changes requested):

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/model/Entitlement.java
```diff
@@ -36,6 +36,7 @@
     CDP_LIST_FILTERING,
     CDP_EMBEDDED_DATABASE_ON_ATTACHED_DISK,
     CDP_DATA_LAKE_LOAD_BALANCER,
+    CDP_DATA_HUB_LOAD_BALANCER,
```
hreeve-cloudera, line 39: Here's a question: How many entitlements do we want to have? I created the data lake specific LB entitlement because there are two uses cases for the LB: data lake HA and semi-private networks. And I wanted to be able to enable one or the other. But for the data hub I don't think we currently have a second use case for it. We'll only want to enable it if we're using semi-private networks. So should we have a data hub specific entitlement to be consistent with the data lake specific entitlement? Or would it be sufficient to just check for the semi-private network specific entitlement, which I added as part of https://github.com/hortonworks/cloudbreak/pull/9675? I could see the argument being made either way. Just wanted to bring it up.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -349,27 +349,34 @@ private void validateStackAuthentication(StackV4Request source) {
         return convertedSet;
     }
 
+    private boolean shouldCreateLoadBalancers(StackV4Request source) {
```
hreeve-cloudera, line 352: I did something similar here in https://github.com/hortonworks/cloudbreak/pull/9675. It's actually probably going to be a bit of a pain to consolidate this change and that one because they both mess with this part of the code a lot.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -349,27 +349,34 @@ private void validateStackAuthentication(StackV4Request source) {
...
         Set<LoadBalancer> loadBalancers = new HashSet<>();
         Set<TargetGroup> targetGroups = new HashSet<>();
-        // TODO expand this to data hubs
-        if (StackType.DATALAKE.equals(source.getType()) &&
-            entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId())) {
-            LOGGER.info("Setting up load balancers for stack {}", source.getName());
-            Set<String> knoxGatewayGroupNames = loadBalancerConfigService.getKnoxGatewayGroups(stack);
-            Set<InstanceGroup> knoxGatewayGroups = stack.getInstanceGroups().stream()
+        if (!shouldCreateLoadBalancers(source)) {
+            return loadBalancers;
```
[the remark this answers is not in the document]
enis: What you suggest is an anti-pattern. Elsewhere we typically follow early-return pattern (https://medium.com/swlh/return-early-pattern-3d18a41bba8) which has the advantage of simpler indentation, and simpler mental model to follow because the function reads as: 
 - if not enabled return early
 - rest of the logic.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/model/Entitlement.java
enis: Thanks for bringing it up. Now we have 3 entitlements it seems. I had started this patch before looking at #9675. The reason DH is different than DL is just in case we have a bug for DHs but not for DLs. This is unlikely, but I wanted to be cautious. Let me rebase this on top of #9675, and get rid of this entitlement in favor of CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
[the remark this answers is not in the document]
lacikaaa: then:
```
if (!shouldCreate) {
  return Sets.of()
} else {
  initialize vars
  logic
  return loadBalancers
}
```
it's more readable. I look at the code, and I see there are 2 output by indentation, without really reading any code looking if there is any return.

biharitomi (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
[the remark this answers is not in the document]
enis: Ended up rebasing on top of #9675, which did not end up requiring this part of the change. So this discussion is obsolete now.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/model/Entitlement.java
enis: Actually ended up rebasing on top of #9675 and getting rid of this entitlement, because having 3 is a bit too much. 
Did you add CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY to the thunderhead code base yet?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -349,27 +349,34 @@ private void validateStackAuthentication(StackV4Request source) {
         return convertedSet;
     }
 
+    private boolean shouldCreateLoadBalancers(StackV4Request source) {
+        return (StackType.DATALAKE.equals(source.getType()) &&
```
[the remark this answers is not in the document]
enis: These entitlement checks now look a bit different since now the patch is sitting on top of #9675.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
hreeve-cloudera: Yes, here: https://jira.cloudera.com/browse/CDPSDX-2728

common-model/src/main/java/com/sequenceiq/common/api/type/LoadBalancerType.java
```diff
@@ -1,6 +1,6 @@
 package com.sequenceiq.common.api.type;
 
 public enum LoadBalancerType {
-    PUBLIC,
-    PRIVATE
+    ENDPOINT_ACCESS_GATEWAY,
```
biharitomi, line 4: The same here, do we really need this refactor?
On the other hand for me PUBLIC/PRIVATE is more clear and more fit into CB's domain.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -24,12 +29,41 @@
...
             return null;
         }
+        return selectSubnet(network, tunnel, cloudPlatform, network.getCbSubnetValues(), null);
+    }
+
+    public String provideEndpointGateway(NetworkDto network, CloudPlatform cloudPlatform, String baseSubnetId) {
+        if (baseSubnetId == null || network == null ||
+                ((network.getEndpointGatewaySubnetMetas() == null || network.getEndpointGatewaySubnetMetas().isEmpty())
+                && (network.getCbSubnets() == null || network.getCbSubnets().isEmpty()))) {
+            LOGGER.debug("Check failed, returning null");
```
biharitomi, line 44: Can we explain why the check failed, it would result in an easier root cause analyses from the logs??

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
```diff
@@ -24,12 +29,41 @@
...
+            return null;
+        }
+        // Use tunnel type DIRECT so that any logic that checks if CCM/private subnets are required evaluates to false
```
biharitomi, line 47: CCM is a parameter on environment request, is there any validation that this feature and CCM could not be enabled at the same time??
If not and a customer has enabled load balancer feature then the CCM option doesn't make sense on the wizards.

biharitomi (changes requested):

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
hreeve-cloudera: @biharitomi the comments on this class belong in my PR. Enis just got them from rebasing on an old version of that PR.

@enis there have been significant changes since you last rebased. You should rebase again.

hreeve-cloudera: Per the comment, I don't think there's any reason we can't enable CCM and this feature at the same time. That's really the whole point of the feature; the customer can have a fully private environment that's only accessible inside their VPC, but they want public UI and API endpoints so that the UI and API can be called by jobs outside of their VPC. In that case we still need a way to tell CB that the environment itself is private and should use CCM.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
hreeve-cloudera: I'll update this in PR https://github.com/hortonworks/cloudbreak/pull/9675/

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/loadbalancer/AwsLoadBalancerScheme.java
```diff
@@ -1,16 +1,23 @@
 package com.sequenceiq.cloudbreak.cloud.aws.loadbalancer;
 
 public enum AwsLoadBalancerScheme {
-    PUBLIC("internet-facing"),
-    PRIVATE("internal");
+    INTERNET_FACING("internet-facing", "External"),
```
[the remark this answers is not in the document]
hreeve-cloudera: This resource isn't persisted. I decided to do the name refactor to match what's actually used by AWS, which makes things on the AWS side more clear. The reason it was PUBLIC/PRIVATE before is so that I could convert directly between this and `LoadBalanceType`, but in https://github.com/hortonworks/cloudbreak/pull/9693 I added a converter for those enums so copying the names directly is no longer required.

common-model/src/main/java/com/sequenceiq/common/api/type/LoadBalancerType.java
hreeve-cloudera: This change has been reverted in https://github.com/hortonworks/cloudbreak/pull/9693.

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
hreeve-cloudera: Actually I checked and the changes in this class have all been removed from my PR. The logic to select the endpoint gateway subnet is instead being done in `EnvironmentBaseNetworkConverter`. Because the endpoint gateway subnet must be in the same AZ as the stack subnet, we need to wait until after that subnet selection is final before we can select the endpoint gateway subnet. So this logic was happening too early and was removed.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -380,15 +380,13 @@ private void validateStackAuthentication(StackV4Request source) {
     }
 
     private boolean shouldCreateInternalKnoxLoadBalancer(StackType type, EnvironmentNetworkResponse network) {
-        // TODO expand this to data hubs
-        return StackType.DATALAKE.equals(type) &&
+        return (StackType.DATALAKE.equals(type) || StackType.WORKLOAD.equals(type)) &&
             (entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId())
```
hreeve-cloudera, line 384: `entitlementService.datalakeLoadBalancerEnabled()` should only be considered if the stack is of type `DATALAKE`. We don't want the data hub logic to trigger except for semi-private networks. So it'd be something like
```
return (StackType.DATALAKE.equals(type) && entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId()))
            || ((StackType.DATALAKE.equals(type) || StackType.WORKLOAD.equals(type)) && isEndpointGatewayEnabled(network));
```

hreeve-cloudera (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
enis: This logic was something like that before one of the rebases. Let me see do one more attempt.

hreeve-cloudera (approved):

environment/src/main/java/com/sequenceiq/environment/network/service/SubnetIdProvider.java
enis: Indeed these are due to the fact that the branch that this PR was based on got changed. Please take another look since it is a much smaller patch.

enis:
@biharitomi please take another look since some of the changes where coming from a different base for the PR.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
```diff
@@ -380,15 +380,22 @@ private void validateStackAuthentication(StackV4Request source) {
...
+    }
+
+    private boolean shouldCreateInternalKnoxLoadBalancerForDatalake(StackType type, EnvironmentNetworkResponse network) {
         return StackType.DATALAKE.equals(type) &&
             (entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId())
             || isEndpointGatewayEnabled(network));
     }
 
+    private boolean shouldCreateInternalKnoxLoadBalancerForDatahub(StackType type, EnvironmentNetworkResponse network) {
+        return  StackType.WORKLOAD.equals(type) && isEndpointGatewayEnabled(network);
```
biharitomi, line 394: Shouldn't this also consider an entitlement for {{WORKLOAD}} or that entitlement has already been eliminated?

biharitomi (changes requested):

biharitomi (approved):

lacikaaa (approved):

biharitomi:
@enis could you please squash these changes into one commit?
I think it should not be 2 commits.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/StackV4RequestToStackConverter.java
hreeve-cloudera: Enis and I had a talk about that. The reason there's a DATALAKE specific entitlement is because the load balancer is used for both the public endpoint gateway logic, and for data lake HA. So I wanted a way to enable or disable one of those features at a time. But for data hubs, the load balancer is only used in the public endpoint gateway feature. So we decided to just use that entitlement to gate the data hub load balancer.

biharitomi: Thanks, I tracked it down from the code and approved the PR, only comment is to squash the changes into one commit.

enis:
bq.  I think it should not be 2 commits.
We can squash and merge in github, no? That is the standard practice normally. But let me push a squashed one in case.

enis:
@biharitomi, @lacikaaa just pushed a squashed commit. Do you mind merging.

hreeve-cloudera:
@enis It needs to finish up the tests before it can be merged in. I can do it when it's ready though.

PR #9699, opened 2021-01-05
Validate if the non-CDH parcels and CSDs are uploaded to archive duing Data Hub upgrade

tiborpopovics:


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/CsdLocationFilter.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CsdLocationFilter.class);
+
+    @Override
+    public boolean filterImage(Image image, Image currentImage, StackType stackType) {
+        if (StackType.WORKLOAD.equals(stackType)) {
+            if (image == null || image.getPreWarmCsd() == null || image.getPreWarmCsd().isEmpty()) {
+                LOGGER.debug("Image or CSD parcels are not present. Image: {}", image);
+                return false;
+            } else {
+                List<String> csdList = image.getPreWarmCsd();
```
keyki, line 24: This only check the CSDs, what about the pre-warm parcels?

tiborpopovics: I've created a new validator for pre-warm percels. Please check it.

PR #9704, opened 2021-01-05
CB-10455: Fix logic which treated unreachable FreeIPA as reachable

jamisonbennett:
Fixed incorrect logic which treated FreeIPA is being reachable when it
was unreachable and cluster proxy was used. In this case cluster proxy
returns 502 which was incorrectly treated as a successful connection.

Although it would become unreachable for other reasons (e.g. the 3
minute health check), the health API would report incorrect results
for up to 3 minutes. There were also misleading logs which stated
FreeIPA was reachable.

This was manually tested with a local cloudbreak cluster.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
@@ -13,7 +13,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.net.ssl.HostnameVerifier;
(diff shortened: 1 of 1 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientBuilder.java
```diff
@@ -169,7 +166,7 @@ public FreeIpaClient build(boolean withPing) throws URISyntaxException, IOExcept
                 additionalHeaders.forEach(request::addHeader);
                 additionalHeadersStickySessionFirstRpc.forEach(request::addHeader);
                 try (CloseableHttpResponse response = client.execute(request)) {
-                    if (UNAVIALLBE_PING_HTTP_RESPONSES.contains(response.getStatusLine().getStatusCode())) {
+                    if (isUnreachableHttpStatus(response.getStatusLine().getStatusCode())) {
                         throw new HttpException("Ping failed with http status code " + response.getStatusLine().getStatusCode());
```
christmasferret, line 170: Do you want to change LOGGER.debug(msg, e); to LOGGER.error(msg, e);  to help kibana search?


            } catch (Exception e) {
                String msg = "Ping failed";
                LOGGER.debug(msg, e);
                throw new RetryableFreeIpaClientException(msg, new FreeIpaHostNotAvailableException(msg, e));

jamisonbennett: This error is rethrown a few times but it does get logged at the higher level with the error level.

jamisonbennett: This is rethrown a few times and then it gets logged at the higher level at error level.

PR #9705, opened 2021-01-05
CB-10143 User sync takes a significant amount of time

horadla23:
- this commit introduce batch calls between FMS - Freeipa in order to speed up usersync duration and decrease calls to Freeipa

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
@@ -1,15 +1,19 @@
 package com.sequenceiq.freeipa.service.freeipa;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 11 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -39,24 +43,18 @@ public void setWorkloadCredential(FreeIpaClient freeIpaClient, String username,
...
-            } catch (FreeIpaClientException e) {
-                recordWarning(entry.getKey(), e, warnings);
-                if (e.isClientUnusable()) {
-                    LOGGER.warn("Client is not usable for further usage");
-                    throw e;
-                }
-            }
+            String username = entry.getKey();
+            WorkloadCredential workloadCredential = entry.getValue();
+            String ansEncodedKrbPrincipalKey = KrbKeySetEncoder.getASNEncodedKrbPrincipalKey(workloadCredential.getKeys());
```
lajosrodek, line 51: Typo: "ans" -> "asn".

horadla23: nice catch, thanks!

PR #9706, opened 2021-01-05
CB-10363 Highstate must always correct the userid / group id

lacikaaa:
It's possible during upgrade different images has cloudera-scm user and group with different UID/GID.
Cloudera Manager server can use 4 directories. With this change we ensure before starting the server these directories has the right owner. As a side effect we create these directories even if CM doesn't need some of them.
The list of the directories is coming from JAVA side where we create the configuration for CM itself, this way it's ensured we'll set the owners if a new directory introduced.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -54,6 +54,7 @@ import com.sequenceiq.cloudbreak.cloud.model.ClouderaManagerRepo;
 import com.sequenceiq.cloudbreak.cloud.model.StackTags;
 import com.sequenceiq.cloudbreak.cloud.scheduler.CancellationException;
 import com.sequenceiq.cloudbreak.cluster.api.ClusterPreCreationApi;
+import com.sequenceiq.cloudbreak.cluster.model.ServiceLocationMap;
 import com.sequenceiq.cloudbreak.cluster.service.ClusterComponentConfigProvider;
 import com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil;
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -493,15 +494,19 @@ public void decoratePillarWithClouderaManagerCsds(Cluster cluster, Map<String, S
...
-                        "deterministic_uid_gid", isVersionNewerOrEqualThanLimited(cmVersion, CLOUDERAMANAGER_VERSION_7_2_1),
-                        "cloudera_scm_sudo_access", CMRepositoryVersionUtil.isSudoAccessNeededForHostCertRotation(clouderaManagerRepo))))));
+                singletonMap("cloudera-manager", Map.of(
+                        "settings", Map.of(
+                                "heartbeat_interval", cmHeartbeatInterval,
+                                "missed_heartbeat_interval", cmMissedHeartbeatInterval,
+                                "set_cdp_env", isVersionNewerOrEqualThanLimited(cmVersion, CLOUDERAMANAGER_VERSION_7_0_2),
+                                "deterministic_uid_gid", isVersionNewerOrEqualThanLimited(cmVersion, CLOUDERAMANAGER_VERSION_7_2_1),
+                                "cloudera_scm_sudo_access", CMRepositoryVersionUtil.isSudoAccessNeededForHostCertRotation(clouderaManagerRepo)),
+                        "directories", serviceLocations.getAllVolumePath()))));
```
keyki, line 509: Maybe we could call it mgmt_service_directories instead?

lacikaaa: sure, I will rename it

PR #9707, opened 2021-01-05
CB-10451: Perform salt update before performing datalake backup.

kkalvagadda1:
Create a flow chain using the flow for updating salt and also flow for database backup.

This flow chain should first trigger the flow to update the salt pillar and sates in the nodes before triggering the flow for database backup.

This is needed because data lake nodes setup using runtime 7.1.0 and 7.2.0 might not have the necessary salt pillar and salt states needed to perform a database backup.

Tested the changes by triggering the database backup using CBD. I made sure that salt update of triggered before database backup. Additionally, i deleted some of the salt scripts needed for database backup and made sure that database backup was successful and the salt scripts are restored.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/BackupDatalakeDatabaseFlowEventChainFactory.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/BackupDatalakeDatabaseFlowEventChainFactory.java
@@ -1,0 +1,30 @@
+package com.sequenceiq.cloudbreak.core.flow2.chain;
+
+import com.sequenceiq.cloudbreak.common.event.Selectable;
+import com.sequenceiq.cloudbreak.core.flow2.cluster.salt.update.SaltUpdateEvent;
+import com.sequenceiq.cloudbreak.core.flow2.event.DatabaseBackupTriggerEvent;
+import com.sequenceiq.cloudbreak.reactor.api.event.StackEvent;
+import com.sequenceiq.flow.core.chain.FlowEventChainFactory;
+import org.springframework.stereotype.Component;
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/chain/BackupDatalakeDatabaseFlowEventChainFactory.java
```diff
@@ -0,0 +1,31 @@
...
+public class BackupDatalakeDatabaseFlowEventChainFactory implements FlowEventChainFactory<DatabaseBackupTriggerEvent> {
+    @Override
+    public String initEvent() {
+        return FlowChainTriggers.DATALAKE_DATABASE_BACKUP_CHAIN_TRIGGER_EVENT;
+    }
+
+    @Override
+    public Queue<Selectable> createFlowTriggerEventQueue(DatabaseBackupTriggerEvent event) {
+        Queue<Selectable> chain = new ConcurrentLinkedQueue<>();
+        String selector = DATABASE_BACKUP_EVENT.event();
```
lajosrodek, line 25: Nit: What is the benefit of having an explicit variable instead of just inlining the value in line 27 below?

kkalvagadda1: There is no benefit. I can make a change inlining the value.

kkalvagadda1: addressed.

lacikaaa (approved):

PR #9713, opened 2021-01-06
CB-10444 - fixed error message for wrong osType when validation a …

attilapalfi92:
… HDP repository json in maintenance mode. fix the uploading / saving mechanism of that json (StackEndpoint / put)

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
@@ -34,6 +34,7 @@ import javax.inject.Inject;
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -117,6 +118,8 @@ import com.sequenceiq.cloudbreak.service.rdsconfig.RdsConfigService;
 import com.sequenceiq.cloudbreak.service.sharedservice.SharedServiceConfigProvider;
 import com.sequenceiq.cloudbreak.service.stack.StackService;
 import com.sequenceiq.cloudbreak.util.JsonUtil;
+
+import groovyx.net.http.HttpResponseException;
 
 @Service
 public class ClusterService {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -371,12 +374,24 @@ public String getStackRepositoryJson(Long stackId) {
...
+            String stackRepositoryJson = ambariClient.getLatestStackRepositoryAsJson(cluster.getName(), osType, stackRepoId);
+            if (stackRepositoryJson == null) {
+                throw new BadRequestException(String.format("Stack Repository response coming from Ambari server was null "
+                        + "for cluster '%s' and repo url '%s'.", cluster.getName(), stackRepoId));
+            }
+            return stackRepositoryJson;
+        } catch (Exception e) {
+            // this instanceof is necessary because the groovy client hides somehow the HttpResponseException
+            // see https://jira.cloudera.com/browse/CB-10444 for more information
+            if (e instanceof HttpResponseException) {
```
pdarvasi, line 387: so you mean HttpResponseException cannot be caught separately?

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
```diff
@@ -371,12 +374,24 @@ public String getStackRepositoryJson(Long stackId) {
...
+                HttpResponseException cast = HttpResponseException.class.cast(e);
+                if (cast.getResponse().getStatus() == HttpStatus.SC_NOT_FOUND) {
+                    throw new BadRequestException("The specified osType is wrong. " + cast.getMessage());
```
pdarvasi, line 390: pls. add osType value to the logline, too

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
attilapalfi92: java compiler doens't let you catch a checked exception if its not listed in the method signature. thats the point of doing this weird implementation.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ClusterService.java
attilapalfi92: ok

sodre90 (approved):

pdarvasi (approved):

PR #9723, opened 2021-01-06
CB-10512 SDX HA provisioning fails with lazyload exception.

orngejaket:
This fixes an issue where the instance groups are not fully loaded when trying to
access them.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -59,6 +59,7 @@ import com.sequenceiq.cloudbreak.cluster.model.ServiceLocationMap;
 import com.sequenceiq.cloudbreak.cluster.service.ClusterComponentConfigProvider;
 import com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil;
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
+import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
 import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.container.postgres.PostgresConfigService;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.host.decorator.HostAttributeDecorator;
@@ -72,10 +73,10 @@ import com.sequenceiq.cloudbreak.domain.stack.cluster.gateway.ExposedServices;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.gateway.Gateway;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.gateway.GatewayTopology;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.host.HostGroup;
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
 import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceMetaData;
 import com.sequenceiq.cloudbreak.dto.KerberosConfig;
 import com.sequenceiq.cloudbreak.dto.LdapView;
-import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
 import com.sequenceiq.cloudbreak.kerberos.KerberosConfigService;
 import com.sequenceiq.cloudbreak.ldap.LdapConfigService;
 import com.sequenceiq.cloudbreak.orchestrator.exception.CloudbreakOrchestratorCancelledException;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -662,7 +667,10 @@ private void addGatewayUserFacingCertAndFqdn(GatewayConfig gatewayConfig, Cluste
 
     private List<String> getRangerFqdn(Cluster cluster, String primaryGatewayFqdn, List<String> rangerLocations) {
         if (rangerLocations.size() > 1) {
-            String gatewayGroupName = cluster.getStack().getPrimaryGatewayInstance().getInstanceGroup().getGroupName();
+            Stack stack = cluster.getStack();
+            Set<InstanceGroup> groups = instanceGroupService.findByStackId(stack.getId());
+            stack.setInstanceGroups(groups);
```
hreeve-cloudera, line 672: I don't like that we're modifying the Stack object here. I know you're just setting it with the instance groups that are part of the stack, but the Cluster and Stack objects are pass around a lot and I don't think we should modify them unless we have to.

There's a method in `InstanceMetaDataService` called `getPrimaryGatewayInstanceMetadata()`, which wraps to the SQL query to fetch that object. I think you should add a similar query to the `InstanceGroupService`/`InstanceGroupRepository` classes and just call that directly. Like:
```
InstanceGroup instanceGroup = instanceGroupService.getPrimaryGatewayInstanceGroup(cluster.getStack.getId());
if (instanceGroup != null) {
    String gatewayGroupName = instanceGroup.getGroupName();
    ....
```

hreeve-cloudera (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
orngejaket: The stack object's getPrimaryGatwayinstance method does a lot of filtering of the instance groups and instance group metadata.  There's a lot of logic i don't want to reproduce here, so setting it gets it for free.  I can unset it afterwards if needed.

hreeve-cloudera: The `stack.getPrimaryGatewayInstance()` is functionally just going through all the instance metadata looking for an instance group of type `GATEWAY_PRIMARY`. The `InstanceMetaDataService.getPrimaryGatewayInstanceMetadata()` does that with a single SQL query, and it looks like it actively loads the instance group in the repository class:
```
    @EntityGraph(value = "InstanceMetaData.instanceGroup", type = EntityGraph.EntityGraphType.LOAD)
    @Query("SELECT i FROM InstanceMetaData i WHERE i.instanceMetadataType = 'GATEWAY_PRIMARY' AND i.instanceStatus <> 'TERMINATED' "
            + "AND i.instanceGroup.stack.id= :stackId")
    Optional<InstanceMetaData> getPrimaryGatewayInstanceMetadata(@Param("stackId") Long stackId);
```
So I don't think you even need to modify the `InstanceGroupService`. I'm pretty sure you can just do this:
```
Optional<InstanceMetadata> instanceMetadata = instanceMetaDataRepository.getPrimaryGatewayInstanceMetadata(cluster.getStack().getId());
if (instanceMetata.isPresent) {
    String gatewayGroupName = instanceMetadata.get().getInstanceGroup().getGroupName();
    ...
}
```
And then you wouldn't be introducing new eager loading in `InstanceGroup` or modifying the `Stack` object.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -662,7 +667,9 @@ private void addGatewayUserFacingCertAndFqdn(GatewayConfig gatewayConfig, Cluste
 
     private List<String> getRangerFqdn(Cluster cluster, String primaryGatewayFqdn, List<String> rangerLocations) {
         if (rangerLocations.size() > 1) {
-            String gatewayGroupName = cluster.getStack().getPrimaryGatewayInstance().getInstanceGroup().getGroupName();
+            // SDX HA has multiple ranger instances in different groups, in Knox we only want to expose the ones on the gateway.
+            InstanceGroup gatewayInstanceGroup =  instanceGroupService.getPrimaryGatewayInstanceGroupByStackId(cluster.getStack().getId());
+            String gatewayGroupName = gatewayInstanceGroup.getGroupName();
```
hreeve-cloudera, line 672: This will throw an NPE is `gatewayInstanceGroup` is null;

orngejaket: Changed to be optional and throw an error that the instance group is not found for that stack.

hreeve-cloudera (approved):

PR #9724, opened 2021-01-07
CB-10452: Database backup should support hdfs location as well.

kkalvagadda1:
As runtime 7.1.0 and 7.2.0 do not support backup to object stores, we should support backup to HDFS location as well.
This is needed to support backup for data lake that are using runtime 7.1.0 and 7.20.


hreeve-cloudera (approved):
LGTM

keyki (approved):

PR #9734, opened 2021-01-08
CB-10213: Create EFS file system 1

LinaAtAustin:
This is ready to commit as the first step. I will have additional commits to fully support EFS because I don't want to make the update too big to review. 

The EFS design document is at https://docs.google.com/document/d/1FKxH8iZuJbDaZcktcjvoOOKT2ERQrbvFfPqQTGJcvRk/edit?usp=sharing. The "Implementation Details" section discusses approaches.

Tests done:
1) All unit tests past
2) Run local CB, the Data Lake was created successfully with an EFS instance. I have added hard-coded EFS configuration in AwsLaunchService to trigger the EFS creation. I have verified in AWS console that EFS was created successfully. I will add more updates to add file system field in CloudStack, create mount target, mount EC2 to EFS, and provide cloud API for such configuration.

In the next commit, I will update Data Lake request to provide such configuration to Cloud Break AWS.


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/LifeCycleState.java
```diff
@@ -0,0 +1,58 @@
...
+            throw new IllegalArgumentException("Value cannot be null or empty!");
+        }
+
+        for (LifeCycleState enumEntry : LifeCycleState.values()) {
+            if (enumEntry.toString().equals(value)) {
+                return enumEntry;
+            }
+        }
+
+        throw new IllegalArgumentException("Cannot create enum from " + value + " value!");
```
hreeve-cloudera, line 42: You can replace pretty all the functionality of this method with:
```
if (Strings.isNullOrEmpty(value)) {
    throw new IllegalArgumentException("Value cannot be null or empty!");
}
return LifeCycleState.valueOf(value.toUpperCase());
```

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
```diff
@@ -207,6 +208,21 @@ private ModelContext buildDefaultModelContext(AuthenticatedContext ac, CloudStac
             .withPrefixListIds(getPrefixListIds(amazonEC2Client, regionName, network.getOutboundInternetTraffic()))
             .withEncryptedAMIByGroupName(encryptedImageCopyService.createEncryptedImages(ac, stack, resourceNotifier));
 
+        AwsEfsFileSystem efsFileSystem = null;
+
+        // TODO: S3 uses this field. Need to add another field in CloudStack for EFS.
+        // Ideally, use a list of SpiFileSystem
+        if (stack.getFileSystem().isPresent()) {
```
hreeve-cloudera, line 215: Why don't you combine this and the if statement below? So if this is true, `modelContext.withEnableEfs(true);`, etc., and if not `modelContext.withEnableEfs(false);`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
```diff
@@ -0,0 +1,188 @@
...
+    public static final String KEY_THROUGHPUT_MODE = "throughputMode";
+
+    // Allowed values: DISABLED | DISABLING | ENABLED | ENABLING
+    private static final String BACKUP_POLICY_STATUS_DEFAULT = "DISABLED";
+
+    // Allowed values: generalPurpose | maxIO
+    private static final String PERFORMANCE_MODE_DEFAULT = "generalPurpose";
+
+    // Allowed values: bursting | provisioned
+    private static final String THROUGHPUT_MODE_BURSTING = "bursting";
```
hreeve-cloudera, line 40: Is there a reason these were done as strings and not enums?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/AwsEfsResourceBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+        // TODO: We should create EFS using CloudFormation. And may create EFS related resources here.
+        List<CloudResource> computeResources = context.getComputeResources(privateId);
+        Optional<CloudResource> inputEfsSet = computeResources.stream()
+                .filter(resource -> ResourceType.AWS_EFS.equals(resource.getType()))
+                .findFirst();
+
+        if (inputEfsSet.isPresent()) {
+            return List.of(inputEfsSet.get());
+        } else {
+            return Collections.<CloudResource>emptyList();
```
hreeve-cloudera, line 79: You can also just do `List.of()` here to return an empty list.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -495,8 +495,51 @@
...
+        <#if efsFileSystem.throughputMode == "provisioned">
+          "ProvisionedThroughputInMibps": "${efsFileSystem.provisionedThroughputInMibps}",
+        </#if>
+        "ThroughputMode": "${efsFileSystem.throughputMode}"
+      }
+    }
+    </#if>
   }
-  <#if mapPublicIpOnLaunch || (enableInstanceProfile && !existingRole)>
+  <#if mapPublicIpOnLaunch || (enableInstanceProfile && !existingRole) || !existingVPC || !existingSubnet || enableEfs>
```
hreeve-cloudera, line 542: What's the rationale for adding the `!existingVPC || !existingSubnet` checks here?

Based on my talks with Krisztian `existingVPC` and `existingSubnet` are always true, because network creation happens in an earlier part of the code. So by the time the code gets here there is already an existing network, and those will always evaluate to false anyway.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -515,15 +558,24 @@
...
   <#if !existingVPC>
-    ,"CreatedVpc": {
+    <#if mapPublicIpOnLaunch || (enableInstanceProfile && !existingRole)>,</#if>
+    "CreatedVpc": {
         "Value" : { "Ref" : "VPC" }
     }
   </#if>
   <#if !existingSubnet>
-    ,"CreatedSubnet": {
+    <#if mapPublicIpOnLaunch || (enableInstanceProfile && !existingRole) || !existingVPC>,</#if>
```
hreeve-cloudera, line 567: Yeah, I can't quite follow the changes you made to these if statements around the network settings. Can you talk through them?

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/component/AwsLaunchTest.java
```diff
@@ -176,10 +199,18 @@ public void launchStack() throws Exception {
         setupDescribeSubnetResponse();
         setupDescribePrefixListsResponse();
 
+        setupCreateFileSystem();
+        setupDescribeFileSystems();
+        setupDescribeMountTargets();
+        setupDeleteMountTarget();
+        setupDeleteFileSystem();
+
```
hreeve-cloudera, line 207: Let's create a new test instead of changing the old one. Leave the old one as-is for regression testing.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/LifeCycleState.java
LinaAtAustin: Thanks! it should be LifeCycleState.valueOf(value.toLowerCase()).

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
LinaAtAustin: this part is not final. I have inserted testing code to fake EFS request and test. It was removed in this code review. The official way is to add EFS request from Cloud API, then we can extract it here. I am making changes to remove the TODO here as well.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
LinaAtAustin: this is used directly in CloudFormation template. I can try if it can handle enum correctly

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/LifeCycleState.java
hreeve-cloudera: Your enums are in upper case though, right? `valueOf` converts a string to an enum, if the string matches the enum name. It's not related to the "value" you're manually assigning to the enums. In fact, I don't think you need to explicitly assign a "value" at all.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: In some testing cases, it may not be true.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
LinaAtAustin: I have used enum now.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: the condition is when to add "," here. If there is no more element, adding "," will result in invalid template. If there is element, not adding "," will result in invalid template. 

This change makes sure that the template is always valid in different test scenarios. 

I have encountered failure, so I have to make such change to get all tests pass.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudEfsAttributes.java
```diff
@@ -0,0 +1,241 @@
...
+    public String getName() {
+        if (tags == null) {
+            return EFS_TAGKEY_NAME_VALUE_DEFAULT;
+        }
+
+        return tags.getOrDefault(EFS_TAGKEY_NAME, EFS_TAGKEY_NAME_VALUE_DEFAULT);
+    }
+
+    public void setName(String efsName) {
+        if (tags == null) {
```
topolyai5, line 78: This is not thread-safe, could you please init at construct time?

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudEfsAttributes.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        public Builder withDeleteOnTermination(Boolean deleteOnTermination) {
+            this.deleteOnTermination = deleteOnTermination;
+            return this;
+        }
+
+        public Builder withName(String efsName) {
+            if (tags == null) {
+                tags = new HashMap<>();
```
topolyai5, line 194: This is not thread-safe, could you please init at construct time?

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/filesystem/efs/ThroughputMode.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Create instance from input string
+     *
+     * @param value the input in String to be converted to the enum
+     * @return ThroughputMode corresponding state to the input value
+     * @throws IllegalArgumentException If the specified value does not map to one of the known values in this enum.
+     */
+    public static ThroughputMode fromValue(String value) {
+        if (StringUtils.isEmpty(value)) {
+            throw new IllegalArgumentException("Value cannot be null or empty!");
```
topolyai5, line 24: Could you please write a test for this case?

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudEfsAttributes.java
LinaAtAustin: fixed

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudEfsAttributes.java
LinaAtAustin: fixed

topolyai5 (approved):

PR #9738, opened 2021-01-08
CB-10250 mock test for user sync

sodre90:

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/FreeIpaSyncTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/FreeIpaSyncTest.java
@@ -1,0 +1,56 @@
+package com.sequenceiq.it.cloudbreak.testcase.mock;
+
+import javax.inject.Inject;
+
+import org.testng.annotations.Test;
+
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status;
+import com.sequenceiq.freeipa.api.v1.freeipa.user.model.UserSyncState;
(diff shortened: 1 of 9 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/FreeIpaSyncTest.java
```diff
@@ -0,0 +1,52 @@
...
+            when = "calling a freeipe start",
+            then = "freeipa sould be available")
+    public void testSyncFreeIpa(MockedTestContext testContext) {
+        getFreeIpaHealthCheckHandler().setHealthy();
+        testContext
+                .given(FreeIpaTestDto.class).withCatalog(getImageCatalogMockServerSetup().getFreeIpaImageCatalogUrl())
+                .when(freeIpaTestClient.create())
+                .await(Status.AVAILABLE);
+        testContext.given(FreeIpaUserSyncTestDto.class)
+                .await(UserSyncState.SYNC_FAILED);
```
horadla23, line 43: does this mean it will fail everytime in case of mock?

sodre90: yes, now the first usersync in install phase fails. I don't have time to debug why it happens. Then the normal usersync completes without failure.

horadla23 (changes requested):
I almost forgot, the point was in the task to make an internal usersync call but i cannot find which part will make it possible. I know it will not work e2e, but somehow we need to make sure to not broke usersync call with internal actor.

horadla23 (approved):

PR #9739, opened 2021-01-08
CB-10529 fixing the bucket to be able to figure out what will be the …

doktoric:
…name {account-id}-{project-id}. Also modified the bucket location to be the environment region. The last copy method modifed to rewrite which is recommended for google in case of cross region copy


biharitomi (approved):

cegganesh84 (approved):
This is fantastic! Thanks Richard.

PR #9741, opened 2021-01-08
CB-10552 batched query of service account emails in GCP related object storage validator, as only the first 20 emails were considered by the previous logic

biharitomi:


cegganesh84 (approved):
LGTM

aszegedi (approved):

PR #9742, opened 2021-01-08
CB-8677 Fix FreeIPA HA in GCP

cegganesh84:
1. If FreeIPA is installed in multiple hosts then the `hostname -i` gives the same IP address many times in a space-separated form.
2. It somehow happens only in GCP, and without the MULTI_IP_SUBNET guest OS feature.
3. This makes the ipa-client-install fail.
4. The fix here removes all but one IP address from `hostname -i` output.
5. Tested the script manually with 1, 2, and 3 IP addresses.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
@@ -3,7 +3,12 @@
 set -e
 
 FQDN=$(hostname -f)
-IPADDR=$(hostname -i)
+# Get the ipaddresses of the host
+IPADDRS=$(hostname -i)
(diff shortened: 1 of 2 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_install.sh
```diff
@@ -3,7 +3,8 @@
 set -e
 
 FQDN=$(hostname -f)
-IPADDR=$(hostname -i)
+IPADDRS=$(hostname -i)
+IPADDR="${IPADDRS%% *}"
```
lacikaaa, line 7: could you add some comments how this expression works? also some log would be nice to make it easier to investigate if something goes sideways

doktoric (approved):

lacikaaa (approved):

PR #9744, opened 2021-01-10
CB-10359 - increase cluster template related converter test coverage vol.1

gregito:
Increasing line and decision test coverage for:
- ClusterTemplateV4RequestToClusterTemplateConverter
- DistroXAuthenticationToStaAuthenticationConverter
- DistroXClusterToClusterConverter
- DistroXDatabaseRequestToStackDatabaseRequestConverter
- DistroXImageToImageSettingsConverter
- InstanceGroupV1ToInstanceGroupV4Converter
- DistroXParameterConverter
- StackV4RequestToStackConverter
- DistroXV1RequestToStackV4RequestConverter
- ClusterTemplateV4RequestToClusterTemplateConverter
- NetworkV1ToNetworkV4Converter
- TelemetryConverter

### DIFF
--- a/core/src/test/java/com/sequenceiq/cloudbreak/service/publicendpoint/dns/BaseDnsEntryServiceTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/service/publicendpoint/dns/BaseDnsEntryServiceTest.java
@@ -11,7 +11,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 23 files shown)


core/src/test/java/com/sequenceiq/cloudbreak/service/publicendpoint/dns/BaseDnsEntryServiceTest.java
```diff
@@ -127,7 +129,7 @@ void testCreateOrUpdate() {
         when(dnsManagementService.createOrUpdateDnsEntryWithIp(USER_CRN, ACCOUNT_ID, FQDN_2, ENVIRONMENT_NAME, false, List.of(PUBLIC_IP_2))).thenReturn(true);
```
afarsang, line 129: I think would be wise to use any() instead of USER_CRN
IT can happen when  getUser() != USER_CRN

afarsang: or getting the actual getUser() and use it instead of USER_CRN

afarsang (changes requested):

core/src/test/java/com/sequenceiq/cloudbreak/service/publicendpoint/dns/BaseDnsEntryServiceTest.java
gregito: Abandoning the usage of the USER_CRN field is not necessary since we're executing the method under test with this:

`ThreadBasedUserCrnProvider.doAs(USER_CRN, () -> underTest.createOrUpdate(stack));`

so we're expecting the usage of this field on purpose

afarsang (approved):

PR #9747, opened 2021-01-11
CB-10561. Fix Salt error due to missing pillar (hostattrs) after down…

sidseth:
…grade from 2.36

to 2.35.


sidseth:
Haven't tested the part about hostattrs | length. It fails without that.

sidseth:
Updated to trigger the precommits. Will post an update when done testing locally.

keyki (approved):

PR #9754, opened 2021-01-12
Add more backup logs to CB 2.35

jamisonbennett:
CB-10411: Add FreeIPA backup debug to logs
    
    Add FreeIPA backup debug to salt output and /var/log/ipabackup.log.
    Include redaction of the x-amz-security-token.
    Added log rotation for /var/log/ipabackup.log.
    
    CB-10441: Add missing log statements for FreeIPA create request
    
    Add missing log statements for FreeIPA create request.

### DIFF
--- a/common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/old/GcsCloudStorageV1Parameters.java
+++ b/common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/old/GcsCloudStorageV1Parameters.java
@@ -50,4 +50,10 @@ public class GcsCloudStorageV1Parameters implements FileSystemAwareCloudStorage 
         return Objects.hash(serviceAccountEmail);
     }
 
+    @Override
+    public String toString() {
+        return "GcsCloudStorageV1Parameters{" +
+                "serviceAccountEmail='" + serviceAccountEmail + '\'' +
+                '}';
+    }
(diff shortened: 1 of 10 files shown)


common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/old/GcsCloudStorageV1Parameters.java
```diff
@@ -50,4 +50,10 @@ public int hashCode() {
         return Objects.hash(serviceAccountEmail);
     }
 
+    @Override
+    public String toString() {
+        return "GcsCloudStorageV1Parameters{" +
+                "serviceAccountEmail='" + serviceAccountEmail + '\'' +
```
keyki, line 56: E-mail might be sensitive?

lacikaaa: this is already in master and maybe 2.36

keyki:
@doktoric  since both of these commits are present in 2.36 I don't think we should do another hotfix to delay 2.36 promotion.

jamisonbennett:
Closing this in favor of getting 2.36 without the delay of an additional hotfix for 2.35. These changes are in 2.36.

PR #9757, opened 2021-01-12
CB-10592 Trim user name from leading and trailing spaces before sync

lacikaaa:
If user first or last name starts or ends with spaces the sync for that user would fail on FreeIPA side.
To avoid such issues we have to trim these spaces during user sync.
Workloadusername also stripped from spaces and now checked against if it's blank isntead of empty.

Please wait for @handavid approval before merging

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/FmsUserConverter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/FmsUserConverter.java
@@ -1,13 +1,13 @@
 package com.sequenceiq.freeipa.service.freeipa.user.conversion;
 
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Strings;
-import com.sequenceiq.freeipa.service.freeipa.user.model.FmsUser;
+import static com.google.common.base.Preconditions.checkArgument;
+
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 
-import static com.google.common.base.Preconditions.checkArgument;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto;
+import com.google.common.annotations.VisibleForTesting;
+import com.sequenceiq.freeipa.service.freeipa.user.model.FmsUser;
 
 @Component
 public class FmsUserConverter {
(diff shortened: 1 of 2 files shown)


keyki (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/FmsUserConverter.java
```diff
@@ -36,11 +36,11 @@ public FmsUser toFmsUser(
     }
 
     private FmsUser createFmsUser(String workloadUsername, String firstName, String lastName) {
-        checkArgument(!Strings.isNullOrEmpty(workloadUsername));
+        checkArgument(StringUtils.isNotBlank(workloadUsername));
         FmsUser fmsUser = new FmsUser();
-        fmsUser.withName(workloadUsername);
-        fmsUser.withFirstName(StringUtils.defaultIfBlank(firstName, NONE_STRING));
-        fmsUser.withLastName(StringUtils.defaultIfBlank(lastName, NONE_STRING));
+        fmsUser.withName(StringUtils.strip(workloadUsername));
```
handavid, line 41: This isn't needed since the workload username is generated internally and is guaranteed to contain only alphanumeric, '-' and '_'. I would prefer that the workload user name is not modified by user sync.

see https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/docs/cdp/workload_usernames.md for details on how workload usernames are generated.

handavid (approved):
looks good. thanks!
I have one comment about trimming the workload username

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/FmsUserConverter.java
lacikaaa: in that case it won't make any modification, so I see no harm done here by stripping it from spaces. But if you think it can cause any issue later I will revert that part

handavid: Please revert this part.
Semantically, we want the workload username in the control plane to match exactly with the username in the workloads. It is better to pass the field through directly without any transformations, even if the transformations are not expected to make any modifications.

lacikaaa: sure, done. please check again

handavid (approved):
lgtm

PR #9763, opened 2021-01-12
CB-10536. Add internal list APIs to environments and FMS

handavid:
This commit supports the use case where an internal service needs to
list environments or list FreeIPA stacks. This follows the pattern
of the FMS describeInternal API where a parallel API is added that
includes the accountId as a query parameter. Authorization for these
APIs is limited to the internal actor only through the @InternalOnly
annotation.

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
@@ -19,6 +19,7 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
 import com.sequenceiq.cloudbreak.auth.altus.CrnResourceDescriptor;
+import com.sequenceiq.cloudbreak.auth.security.internal.AccountId;
 import com.sequenceiq.cloudbreak.jerseyclient.RetryAndMetrics;
 import com.sequenceiq.cloudbreak.validation.ValidCrn;
 import com.sequenceiq.common.api.telemetry.request.FeaturesRequest;
@@ -92,6 +93,13 @@ public interface EnvironmentEndpoint {
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = EnvironmentOpDescription.LIST, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES, nickname = "listEnvironmentV1")
     SimpleEnvironmentResponses list();
+
+    @GET
+    @Path("internal")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.INTERNAL_LIST, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
+            nickname = "internalListEnvironmentV1")
+    SimpleEnvironmentResponses listInternal(@QueryParam("accountId") @AccountId String accountId);
 
     @PUT
     @Path("/name/{name}/change_credential")
--- a/environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
@@ -13,6 +13,7 @@ import javax.transaction.Transactional.TxType;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
+import com.sequenceiq.cloudbreak.auth.security.internal.AccountId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Controller;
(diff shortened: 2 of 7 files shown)


lacikaaa (approved):

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
```diff
@@ -80,6 +80,13 @@
             nickname = "listFreeIpaClustersByAccountV1")
     List<ListFreeIpaResponse> list();
 
+    @GET
+    @Path("internal/list")
```
doktoric, line 84: why we need /list here ?

handavid: there is already a GET internal endpoint to describeInternal. adding /list here matches the non-internal endpoints.
/ --> describe
internal --> describeInternal
/list --> list
internal/list --> listInternal

horadla23: in this case it seems different endpoint is needed for internal calls, since original one uses accountid of the user (thus wonn't work in case of internal actor) and we should not break the current list method to add a new parameter and make it mandatory in case of internal call, so it is ok in this case

fyi, if a list method of a given api has a crn or accountid parameter, then different method not needed, just need to use one of annotations mentioned in authorization-common/readme.md regarding internal calls

also fyi, authorization of lists is currently disabled, later list will be filtered baseed on user and we need to prepare filtering logic to skip filtering and return full list based on the account id

freeipa/src/main/java/com/sequenceiq/freeipa/controller/FreeIpaV1Controller.java
```diff
@@ -160,6 +160,12 @@ public DescribeFreeIpaResponse describeInternal(@ResourceCrn String environmentC
         return freeIpaListService.list(accountId);
```
horadla23, line 160: you can use `listInternal` method here to ensure if both list logic give the same result

horadla23 (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/v1/EnvironmentController.java
```diff
@@ -233,6 +234,15 @@ public SimpleEnvironmentResponses list() {
         return new SimpleEnvironmentResponses(responses);
     }
 
+    @Override
+    @InternalOnly
+    public SimpleEnvironmentResponses listInternal(@AccountId String accountId) {
+        List<EnvironmentDto> environmentDtos = environmentService.listByAccountId(accountId);
+        List<SimpleEnvironmentResponse> responses = environmentDtos.stream().map(environmentResponseConverter::dtoToSimpleResponse)
+                .collect(Collectors.toList());
+        return new SimpleEnvironmentResponses(responses);
```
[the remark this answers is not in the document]
handavid: done

freeipa/src/main/java/com/sequenceiq/freeipa/controller/FreeIpaV1Controller.java
handavid: done.
I was worried that this may interact with the @InternalOnly annotation, but it doesn't. I'm guessing that the annotations are used to create a proxy when called by outside classes but won't affect calls from within the controller class.

horadla23: yes, spring won't proxy calls within class, thus aspect cannot catch the call and authorization won't happen once again.

horadla23 (approved):

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
```diff
@@ -93,6 +94,13 @@ SimpleEnvironmentResponses deleteMultipleByNames(Set<String> names,
     @ApiOperation(value = EnvironmentOpDescription.LIST, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES, nickname = "listEnvironmentV1")
     SimpleEnvironmentResponses list();
 
+    @GET
+    @Path("internal")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.INTERNAL_LIST, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
+            nickname = "internalListEnvironmentV1")
+    SimpleEnvironmentResponses listInternal(@QueryParam("accountId") @AccountId String accountId);
```
handavid, line 102: @horadla23 do we need to annotate these query parameters @NotEmpty?

horadla23: no, but you need to add '@AccountId' here too since it is a validation annotation

horadla23: oh sorry it is already there, then this should be ok

handavid: ok. thanks! I didn't know if @accountId implied @notempty.

horadla23: https://github.com/hortonworks/cloudbreak/blob/master/auth-internal-api/src/main/java/com/sequenceiq/cloudbreak/auth/security/internal/AccountIdValidator.java validation checks also if string is blank or not so it should be enough

handavid:
the last push just fixed the JIRA in the commit message (from CDPCP to CB)

PR #9767, opened 2021-01-13
CB-10624 we don't handle properly the unauthorized errors while we tr…

topolyai5:
…y to fetch the access configs or encryption keys


doktoric (approved):

biharitomi (approved):
LGTM

PR #9771, opened 2021-01-13
CB-10241 -Data Hub page should show the runtime version instead of th…

pdarvasi:
…e blueprint version

This commit indtoduces stack.stackVersion field to be able to have proper version in Data Hub list view - instead of the current solution of including the Bluperint.stackVersion.
stackVersion is updated in the last step of DH upgrade, after the successful runtime upgrade.


core/src/main/resources/schema/app/20210112140444_CB-10241-Data_Hub_page_should_show_the_runtime_version_instead_of_the_blueprint_version.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-10241-Data Hub page should show the runtime version instead of the blueprint version
+-- Migration SQL that makes the change goes here.
+ALTER TABLE stack DROP COLUMN IF EXISTS stackversion;
+ALTER TABLE stack ADD COLUMN stackversion VARCHAR (255);
```
lajosrodek, line 4: Can `stackversion` already exist with a different definition, or rather just want to make this migration idempotent? If the latter, `ADD COLUMN IF NOT EXISTS` would be better so that no `DROP COLUMN` is needed.

pdarvasi: You are right, corrected!

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -256,7 +256,11 @@
             + "u.userName as username, "
             + "u.userCrn as usercrn, "
             + "c.certExpirationState as certExpirationState "
-            + "FROM Stack s LEFT JOIN s.cluster c LEFT JOIN c.blueprint b LEFT JOIN s.stackStatus ss LEFT JOIN s.creator u "
+            + "FROM Stack s "
+            + "LEFT JOIN s.cluster c "
```
doktoric, line 260: this will cause heavy fetches from the database

doktoric:
I think this is already presented in the stack.component table or not?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
pdarvasi: I have not modified the from clause, just made it more readable

pdarvasi:
> I think this is already presented in the stack.component table or not?

Yes, but it is a part of a json fragment and the component table is not joined in the stack-view query. This PR handles the version the same way as it is in DL service.

keyki (approved):

PR #9772, opened 2021-01-13
CB-10419: [CCMv2] New route for knox from nginx on port 9443

jszeles:
CcmV2 agent currently forwards requests only to nginx on port 9443.

Control Plane requests to cluster's /resourcemanager and similar api need to be routed to knox for features like DH autoscaling to succeed.

Hence nginx on port 9443 should be configured with a new rule which routes all requests with path prefix "/knox/<clustername>" to knox running on the node. The re-route rule to knox should not include path prefix "/knox" and should only include /<clustername>.


jszeles:
retest this please

jszeles:
retest this please

TheTinkerDad (approved):

PR #9777, opened 2021-01-13
CB-9483 Connect timeout on FreeIPA client won't translate to Forbidde…

lacikaaa:
…n exception

In this commit:
- [FMS] LDAP and Kerberos endpoint calls won't return Forbidden anymore when FreeIPA is unreachable
- [CB] Kerberos config fetching code won't handle forbidden differently than other exceptions. Only not found will be handled as if there is no FreeIPA present
- [CB] Calls FMS as internal actor to get Kerberos and LDAP configuration so it won't be tied to user role
- [FMS] For future use describe methods for LDAP and Kerberos endpoints are ready for internal calls
- [CB] After Kerberos validation state failure notification will display the exception message too instead of the simple `Kerberos validation failed.`

Tested locally with SDX creation:
- with unreachable FreeIPA while it is AVAILABLE in FMS
- with unreachable FreeIPA when it's UNREACHABLE in FMS
- with working FreeIPA

### DIFF
--- a/common/src/main/resources/messages/messages.properties
+++ b/common/src/main/resources/messages/messages.properties
@@ -151,7 +151,7 @@ cluster.externaldatabase.stop.commenced=External database stop commenced.
 cluster.externaldatabase.stop.failed=External database stop failed.
 cluster.externaldatabase.stop.finished=External database stop finished.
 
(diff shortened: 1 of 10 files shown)


horadla23 (approved):
lgtm

jamisonbennett (approved):

PR #9778, opened 2021-01-13
CB-10265: Only update the required CSDs with Data Hub upgrade

tiborpopovics:
When we send an upgrade request it's important to use only the required CSD files and pre warmed parcels to filter the images.
The second part of this change is during the upgrade we need only download the required CSD files.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ImageFilterParamsFactory.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ImageFilterParamsFactory.java
@@ -1,0 +1,54 @@
+package com.sequenceiq.cloudbreak.service.upgrade;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.model.ClouderaManagerProduct;
+import com.sequenceiq.cloudbreak.cloud.model.catalog.Image;
+import com.sequenceiq.cloudbreak.cluster.service.ClouderaManagerProductsProvider;
+import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.stack.cluster.ClusterComponent;
+import com.sequenceiq.cloudbreak.service.parcel.ParcelService;
+import com.sequenceiq.cloudbreak.service.upgrade.image.ImageFilterParams;
+
+@Component
+class ImageFilterParamsFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ImageFilterParamsFactory.class);
+
+    @Inject
+    private ParcelService parcelService;
+
+    @Inject
+    private ClouderaManagerProductsProvider clouderaManagerProductsProvider;
+
+    ImageFilterParams create(Image image, boolean lockComponents, Stack stack) {
+        return new ImageFilterParams(image, lockComponents, getStackRelatedParcels(stack), stack.getType());
+    }
+
+    private Map<String, String> getStackRelatedParcels(Stack stack) {
+        Set<ClusterComponent> componentsByBlueprint = parcelService.getParcelComponentsByBlueprint(stack);
+        if (stack.isDatalake()) {
+            ClouderaManagerProduct stackProduct = getCdhProduct(componentsByBlueprint);
+            LOGGER.debug("For datalake clusters only the CDH parcel is related in CM: {}", stackProduct);
+            return Map.of(stackProduct.getName(), stackProduct.getVersion());
+        } else {
+            Set<ClouderaManagerProduct> products = clouderaManagerProductsProvider.getProducts(componentsByBlueprint);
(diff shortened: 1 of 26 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ImageFilterParamsFactory.java
```diff
@@ -0,0 +1,54 @@
...
+    private ParcelService parcelService;
+
+    @Inject
+    private ClouderaManagerProductsProvider clouderaManagerProductsProvider;
+
+    ImageFilterParams create(Image image, boolean lockComponents, Stack stack) {
+        return new ImageFilterParams(image, lockComponents, getActivatedParcels(stack), stack.getType());
+    }
+
+    private Map<String, String> getActivatedParcels(Stack stack) {
```
keyki, line 37: Probably we shouldn't call it getActivatedParcels anymore, because we're not checking the CM anymore. Perhaps something like `getStackRelatedParcels`.

keyki: In the `ImageFilterParams` we should rename the `activatedParcels` field as well.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ImageFilterParamsFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        Set<ClusterComponent> componentsByBlueprint = parcelService.getParcelComponentsByBlueprint(stack);
+        if (stack.isDatalake()) {
+            ClouderaManagerProduct stackProduct = getCdhProduct(componentsByBlueprint);
+            LOGGER.debug("For datalake clusters only the CDH parcel is activated in CM: {}", stackProduct);
+            return Map.of(stackProduct.getName(), stackProduct.getVersion());
+        } else {
+            Set<ClouderaManagerProduct> products = clouderaManagerProductsProvider.getProducts(componentsByBlueprint);
+            LOGGER.debug("The activated parcels for the datahub cluster: {}", products);
```
keyki, line 45: The 2 log message should be changed to say these are the parcels related to a particular stack rather than saying they're activated.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/CsdLocationFilter.java
```diff
@@ -1,33 +1,64 @@
...
             }
         } else {
-            LOGGER.debug("Skip filtering because the stack type is {}", stackType);
+            LOGGER.debug("Skip filtering because the stack type is {}", imageFilterParams.getStackType());
             return false;
         }
     }
+
+    private List<String> getPreWarmCsdList(Image image, ImageFilterParams imageFilterParams) {
+        Set<String> activatedParcelNames = getParcelNames(imageFilterParams);
```
keyki, line 42: Let's `activatedParcelNames` to related or relevant parcel names.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/CsdLocationFilter.java
```diff
@@ -1,33 +1,64 @@
...
+        List<String> imagePreWarmCsd = image.getPreWarmCsd();
+        return CollectionUtils.isEmpty(activatedParcelNames) ? imagePreWarmCsd
```
keyki, line 44: So this means if we can't find any related parcel we're going to check all CSDs?

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/CsdLocationFilter.java
```diff
@@ -1,33 +1,64 @@
...
+                : filterPreWarmCsdListByActivatedParcels(activatedParcelNames, imagePreWarmCsd);
+    }
+
+    private List<String> filterPreWarmCsdListByActivatedParcels(Set<String> activatedParcelNames, List<String> imagePreWarmCsd) {
```
keyki, line 48: Renaming suggestion to accommodate previous renames: filterPreWarmCsdListByRelatedParcels

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/PreWarmParcelLocationFilter.java
```diff
@@ -32,17 +36,31 @@ public boolean filterImage(Image image, Image currentImage, StackType stackType)
...
         return Optional.ofNullable(image)
                 .map(Image::getPreWarmParcels).orElse(Collections.emptyList())
                 .stream()
+                .filter(filterByActivatedParcels(activatedParcels))
                 .flatMap(List::stream)
                 .filter(parcel -> StringUtils.hasText(parcel) && parcel.startsWith(URL_PREFIX))
                 .collect(Collectors.toList());
     }
+
+    private Predicate<List<String>> filterByActivatedParcels(Map<String, String> activatedParcels) {
```
keyki, line 54: Ame renaming.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/CsdLocationFilter.java
tiborpopovics: Exactly!

PR #9781, opened 2021-01-14
CB-10598 Azure network resource group is not deleted when "Create new…

gergopapi2:
… RG" is used

When you create a new Azure environment using the create RG option (multiple RGs) the control plane creates 4 RGs. One of them is only for the VNet and when you delete the environment it should delete the network RG as well. The problem is that currently only the single RG name is fetched during deletion.


pdarvasi (approved):

pdarvasi:
@gergopapi2 LGTM, I've re-triggered the failing IT.

PR #9785, opened 2021-01-14
CB-10536. Add internal API to get usersync status

handavid:
This is similar to my previous pull request. I identified one more API that we'll need to call internally.

An internal usersync status API is added to the FMS
to allow internal actors to retrieve status of usersync
operations.


horadla23 (approved):

handavid:
run tests

handavid:
Where are my manners. I forgot to say "please!"

PR #9786, opened 2021-01-14
CB-10143 User sync takes a significant amount of time

horadla23:
- this commit introduce batch calls between FMS - Freeipa in order to speed up usersync duration and decrease calls to Freeipa
- if fullsync finishes with warning and part of users and groups were synced successfully, we try to sync missing ones once again

- this change also contains a smaller refactor to decrease size of FreeipaClient class and make fms - freeipa operations easier to handle (for now only usersync related operations)
- added integration test to check duration and result of usersync
- added unit tests in order to increase coverage
- added entitlement for batch call, if it works in long term, we can remove it and also logic for individual calls during usersync

you can read more details on this in the jira ticket, but regarding performance improvement, this change can decrease the duration of initial usersync on mow-dev from ~1hour to ~15minutes based on measurement performed in the last couple of weeks


horadla23:
i"m still testing this, but in the meantime it is ready for review

horadla23:
still need to test with FreeipaHA

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -187,10 +194,8 @@ public Host addHost(String fqdn) throws FreeIpaClientException {
     }
 
     public User deleteUser(String userUid) throws FreeIpaClientException {
-        FreeIpaChecks.checkUserNotProtected(userUid, () -> String.format("User '%s' is protected and cannot be deleted from FreeIPA", userUid));
-        List<Object> flags = List.of(userUid);
-        Map<String, Object> params = Map.of();
-        return (User) invoke("user_del", flags, params, User.class).getResult();
+        return UserRemoveOperation.get(userUid).invoke(this).orElseThrow(() ->
```
jamisonbennett, line 197: I think there are several places that might be looking for `FreeIpaErrorCodes.NOT_FOUND` in the original exception and I am concerned that this might not be handled correctly in the bulk case. I don't have something specific to point to but I wanted to check if this case was considered.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -200,16 +205,8 @@ public Role deleteRole(String roleName) throws FreeIpaClientException {
...
-        List<Object> flags = List.of(user);
-        Map<String, Object> params = Map.of(
-                "givenname", firstName,
-                "sn", lastName,
-                "loginshell", "/bin/bash",
-                "random", true,
-                "setattr", "krbPasswordExpiration=" + MAX_PASSWORD_EXPIRATION_DATETIME
-        );
-        return (User) invoke("user_add", flags, params, User.class).getResult();
+        return UserAddOperation.get(user, firstName, lastName).invoke(this).orElseThrow(() ->
```
jamisonbennett, line 208: Same comment as above but this time the code looks for `FreeIpaErrorCodes.DUPLICATE_ENTRY` in the original exception.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
```diff
@@ -60,6 +61,10 @@ public boolean isEntitledFor(String accountId, Entitlement entitledFor) {
         return isEntitlementRegistered(accountId, entitledFor);
     }
 
+    public boolean isFmsToFreeipaBatchCallEnabled(String accountId) {
```
jamisonbennett, line 64: I think you might need to update FreeIPA to 4.6.7 and then check the version of CB which is deployed so it doesn't contain an old CB version. https://www.freeipa.org/page/Releases/4.6.7 fixes CVE-2019-10195 for logging sensitive information in the batch APIs. You may need to check if this type of issue will expose the hashes or other sensitive information. CB might have some filters to protect against it. But please verify this.

jamisonbennett (changes requested):
Thanks for adding the unit tests, I had a few things which I wanted to have checked but I think we might need to upgrade FreeIPA to 4.6.7 or later.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
horadla23: @handavid already mentioned to take care about logging, please check `UserModOperation` and `SetWlCredentialOperation`, I changed `toString` in `getParams` to prevent app to log sensitive data. Updating FreeIpa would require more work, additional testing, so I would do that in a different task.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: yes, right, i'll introduce a method to get acceptable error codes and handle it in batch call, in case of remove it will be not_found

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: yes, same here, in case of useradd, duplicate_entry error should be acceptable

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: added it, anyway i tested batch call with useradd operation for user who is existed in freeipa alrady, it seemed batch call handled it, and doesn't even thrown a duplicateeentry error

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: same here, added it, anyway i tested batch call with userdel operation for user who is not existed in freeipa already, it seemed batch call handled it, and doesn't even thrown a notfound error

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
jamisonbennett: @horadla23 Can you please also check the logs on disk on FreeIPA? I see that the FMS will mask out the sensitive data. We have had a few cases logging sensitive information to disk on FreeIPA and I would rather not reintroduce something which was already fixed. I just want to make sure the password hashes don't end up in the logs.

horadla23: sure, will check

horadla23: @jamisonbennett do you know about any specific log which should be checked?
i checked:
`/var/log/ipaclient-install.log`
`/var/log/ipaserver-install.log`
`/var/log/ipa/ipactl.log`
`/var/log/kadmind.log`
`/var/log/krb5kdc.log`
also checked similar logs in cloudstorage
i haven't found any sensitivity info, also no sign about batch calls in these logs

lacikaaa: `/var/log/httpd/*` is the important one

jamisonbennett: @horadla23 that is a pretty good list. The only other logs I would check would be:
`/var/log/httpd/*`

horadla23: yeah, i checked access_log, but missed error_log, strange but error_log contains entries about. batch call and sensitive info is hidden in the log, example:
```
[Wed Jan 20 15:45:29.934583 2021] [:error] [pid 19564] ipa: INFO: admin@ADAMHORV.XCU2-8Y8X.WL.CLOUDERA.SITE: batch: user_mod(u'fakemockuser7', ipasshpubkey=(), setattr=(u'cdpHashedPassword=[FILTERED]', u'cdpUnencryptedKrbPrincipalKey=[FILTERED]', u'krbPasswordExpiration=20220119154400Z')): SUCCESS
[Wed Jan 20 15:45:29.958331 2021] [:error] [pid 19564] ipa: INFO: admin@ADAMHORV.XCU2-8Y8X.WL.CLOUDERA.SITE: batch: user_mod(u'fakemockuser74', ipasshpubkey=(), setattr=(u'cdpHashedPassword=[FILTERED]', u'cdpUnencryptedKrbPrincipalKey=[FILTERED]', u'krbPasswordExpiration=20220119154401Z')): SUCCESS
[Wed Jan 20 15:45:29.981802 2021] [:error] [pid 19564] ipa: INFO: admin@ADAMHORV.XCU2-8Y8X.WL.CLOUDERA.SITE: batch: user_mod(u'fakemockuser8', ipasshpubkey=(), setattr=(u'cdpHashedPassword=[FILTERED]', u'cdpUnencryptedKrbPrincipalKey=[FILTERED]', u'krbPasswordExpiration=20220119154400Z')): SUCCESS
[Wed Jan 20 15:45:30.005462 2021] [:error] [pid 19564] ipa: INFO: admin@ADAMHORV.XCU2-8Y8X.WL.CLOUDERA.SITE: batch: user_mod(u'fakemockuser71', ipasshpubkey=(), setattr=(u'cdpHashedPassword=[FILTERED]', u'cdpUnencryptedKrbPrincipalKey=[FILTERED]', u'krbPasswordExpiration=20220119154401Z')): SUCCESS
```

from usersync related operations' parameters (user_add_ user_del, group_add, group_del, group_add_member, group_remove_member, user_mod) only user_mod's parameters contains sensitive information, so these look good also

horadla23: @lacikaaa @jamisonbennett do you know if there are older versions of FreeIpa which should be checked?

jamisonbennett: Cluster provisioned prior to 2.33 will not have https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/httpd-log-filter.sh. This is not specific to the bulk API change, so I don't see this as making anything worse.

jamisonbennett (commented):
I think it looks good, but I wanted to make sure at least someone else agrees with my thoughts about this not making anything worse with respect to what is logged in clusters that are provisioned prior to 2.33.

horadla23:
> I think it looks good, but I wanted to make sure at least someone else agrees with my thoughts about this not making anything worse with respect to what is logged in clusters that are provisioned prior to 2.33.

Of course, I'll search for at least one more reviewer. Btw this will be behind an entitlement first, so we can easily switch this on and off.

horadla23:
@lacikaaa @jamisonbennett @lnardai @sodre90 if you have anything more to add this, please comment it, otherwise pls approve it.
If someone else should look at this, please add him/her to the reviewers as well.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/GroupAddMemberOperation.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private BiConsumer<String, String> warnings;
+
+    private GroupAddMemberOperation(String group, Collection<String> users, BiConsumer<String, String> warnings) {
+        this.group = group;
+        this.users = users;
+        this.warnings = warnings;
+    }
+
+    public static GroupAddMemberOperation get(String group, Collection<String> users, BiConsumer<String, String> warnings) {
```
lacikaaa, line 34: I don't understand why do you use a static method to create an instance instead of the constructor. I don't see any additional value and also `get` is a bit misleading, `create` would be more meaningful in my opinion.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/GroupAddOperation.java
```diff
@@ -0,0 +1,61 @@
...
+    public String getMethodName() {
+        return "group_add";
+    }
+
+    @Override
+    protected List<Object> getFlags() {
+        return List.of(group);
+    }
+
+    public Optional<Group> invoke(FreeIpaClient freeIpaClient) throws FreeIpaClientException {
```
lacikaaa, line 43: shouldn't these invoke methods have `Override` annotation?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -704,8 +639,28 @@ public TopologySegment deleteTopologySegment(String topologySuffixCn, TopologySe
...
+
+    public void callBatch(BiConsumer<String, String> warnings, List<Object> operations, Integer partitionSize,
+            Set<FreeIpaErrorCodes> acceptableErrorCodes) throws FreeIpaClientException {
+        List<List<Object>> partitions = Lists.partition(operations, partitionSize);
+        for (List<Object> operationsPartition : partitions) {
+            BatchOperation.get(operationsPartition, warnings, acceptableErrorCodes).invoke(this);
+        }
+    }
+
+    public static void checkIfClientStillUsable(FreeIpaClientException e) throws FreeIpaClientException {
```
lacikaaa, line 655: why static?

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
```diff
@@ -704,8 +639,28 @@ public TopologySegment deleteTopologySegment(String topologySuffixCn, TopologySe
...
+        if (e.isClientUnusable()) {
+            LOGGER.warn("Client is not usable for further usage");
```
lacikaaa, line 657: nit: sounds a bit strange reading this. I think sth like `Client is not usable anymore` should be enough, or similar

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -72,6 +80,8 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UserSyncService.class);
 
+    private static final Integer BATCH_ADD_REMOVE_MEMBER_PARTITION_SIZE = 50;
```
lacikaaa, line 83: shouldn't this be configurable via environment?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -358,14 +349,65 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
...
             }
 
+            if (fullSync && !warnings.isEmpty() && fmsToFreeipaBatchCallEnabled) {
+                UsersStateDifference usersStateDifferenceAfterSync = compareUmsAndFreeIpa(umsUsersState, fullSync, freeIpaClient);
+                if (usersStateDifferenceChanged(usersStateDifferenceBeforeSync, usersStateDifferenceAfterSync)) {
+                    LOGGER.info(String.format("Sync was partially successful for %s, thus we are trying it once again", stack.getResourceCrn()));
+                    warnings.clear();
+                    applyDifference(umsUsersState, environmentCrn, warnings, usersStateDifferenceAfterSync, freeIpaClient, fmsToFreeipaBatchCallEnabled);
+                }
+            }
```
lacikaaa, line 359: would be nice to move into a method and give it some nice name. eg. `retrySyncIfBatchCallHasWarnings`
I had to look at the log line to understand why we need this

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/AbstractFreeipaOperation.java
```diff
@@ -0,0 +1,56 @@
...
+import com.sequenceiq.freeipa.client.FreeIpaClientException;
+import com.sequenceiq.freeipa.client.model.RPCResponse;
+
+public abstract class AbstractFreeipaOperation<R> {
+
+    public abstract String getMethodName();
+
+    public abstract Optional<R> invoke(FreeIpaClient freeipaClient) throws FreeIpaClientException;
+
+    public void appendOperationToBatchCall(List<Object> operations) {
```
lacikaaa, line 19: this is a bit odd for me. I would expect that this could return the result map, and the caller would add the Map to the list/collection

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -413,181 +455,151 @@ UsersState getIpaUserState(FreeIpaClient freeIpaClient, UmsUsersState umsUsersSt
...
-                    checkIfClientStillUsable(e);
+                    LOGGER.error("Failed to delete {}", user, e);
+                    warnings.accept(user, "Failed to remove user:" + e.getMessage());
+                    freeIpaClient.checkIfClientStillUsable(e);
                 }
             }
         }
+        if (fmsToFreeipaBatchCallEnabled) {
+            freeIpaClient.callBatch(warnings, operations, Set.of(FreeIpaErrorCodes.NOT_FOUND));
+        }
```
lacikaaa, line 550: I see these 3 lines multiple times and also the `handleOperation` has the same if condition. It would be possible to collect the operations at first, and if it's batch then call the batch API else invoke them 1-by-1. In this case handleOperation won't invoke it, just collect them, and there could be an invoke/call part with the if condition.

lacikaaa (changes requested):
thanks for the changes! this will be very helpful.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/GroupAddMemberOperation.java
horadla23: i like to use static method for initialization to increase readability, not a big deal. yes, `create` makes more sense, i'll change these

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/GroupAddOperation.java
horadla23: yes, it should, i do not know why those are missing, will fix them

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: logic depends only on the parameter, not on the freeipaclient object. but it seems, FreeIpaClientExceptionUtil has the same logic, so i'll remove this

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: makes sense, will change it

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
horadla23: i was not sure about this, i am afraid if someone uses too high value, it will cause issues and it requires lot of testing also. but if you think it is worth to make it configurable, i'll make that change

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
horadla23: sure, i'll do it

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
lacikaaa: I can't say I have seen we changed something like this in prod as a hotfix, but if we would like to run it with different numbers in dev environments to test it, then we could apply it easily without affecting prod. So yes, I think it worth moving out.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
lacikaaa: but I think it's harder to mock, so I'm against using static. Also in this case it wouldn't make any difference.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
horadla23: unfortunately, there is two types of callBatch, one with default partition size and one for custom partition size, also some operation have one foreach loop, some has two foreach loop, so each operation differ a little bit. 
so in case of your suggestion, collection part of the logic should be somehow generalized which is quite hard in case of these operations, since for loops are quite custom, but if there is a way, let's discuss it in private.

horadla23: ok, for second thought, i see your point, yes, i think it will work

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java
horadla23: that's a valid point, i'll remove static, and keep this method because it has logging

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -23,12 +27,9 @@
     public void setWorkloadCredential(FreeIpaClient freeIpaClient, String username, WorkloadCredential workloadCredential)
             throws IOException, FreeIpaClientException {
         LOGGER.debug("Setting workload credentials for user '{}'", username);
-
         try {
-            String ansEncodedKrbPrincipalKey = KrbKeySetEncoder.getASNEncodedKrbPrincipalKey(workloadCredential.getKeys());
-            freeIpaClient.userSetWorkloadCredentials(username,
-                    workloadCredential.getHashedPassword(), ansEncodedKrbPrincipalKey, workloadCredential.getExpirationDate(),
-                    workloadCredential.getSshPublicKeys().stream().map(UserManagementProto.SshPublicKey::getPublicKey).collect(Collectors.toList()));
+            String expiration = freeIpaClient.formatDate(workloadCredential.getExpirationDate());
```
handavid, line 31: you can move the date formatting to `getOperation`. This will remove the repeated code on line 49 and simplify the method signature for `getOperation`.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -358,14 +349,65 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
                 LOGGER.debug("Finished {}.", LogEvent.SYNC_CLOUD_IDENTITIES);
             }
 
+            if (fullSync && !warnings.isEmpty() && fmsToFreeipaBatchCallEnabled) {
+                UsersStateDifference usersStateDifferenceAfterSync = compareUmsAndFreeIpa(umsUsersState, fullSync, freeIpaClient);
+                if (usersStateDifferenceChanged(usersStateDifferenceBeforeSync, usersStateDifferenceAfterSync)) {
+                    LOGGER.info(String.format("Sync was partially successful for %s, thus we are trying it once again", stack.getResourceCrn()));
+                    warnings.clear();
```
handavid, line 356: I know that we aren't really doing anything with the warnings yet, but I fear that we will lose information by clearing them here. We will return only those warnings from the retry. E.g., If the retry fails with some exception before adding a user, then the warnings won't indicate that the user wasn't added.

horadla23: sure, makes sense to keep warningsfrom original sync

horadla23:
@lacikaaa @handavid please check new changes based on your suggestions

lacikaaa (approved):
LGTM, please wait for @handavid approval too

jamisonbennett (approved):

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/BatchOperation.java
```diff
@@ -0,0 +1,65 @@
...
+        this.acceptableErrorCodes = acceptableErrorCodes;
+    }
+
+    public static BatchOperation create(List<Object> operations, BiConsumer<String, String> warnings, Set<FreeIpaErrorCodes> acceptableErrorCodes) {
+        return new BatchOperation(operations, warnings, acceptableErrorCodes);
+    }
+
+    @Override
+    public String getOperationName() {
+        return METHOD_NAME_BATCH;
```
handavid, line 40: is there a reason that BatchOperation uses a constant here? The other operations return a String directly.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -358,14 +351,70 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
                 LOGGER.debug("Finished {}.", LogEvent.SYNC_CLOUD_IDENTITIES);
             }
 
+            retrySyncIfBatchCallHasWarnings(stack, umsUsersState, fullSync, warnings, freeIpaClient,
```
handavid, line 354: this should probably move above the cloud identity sync since the retry does not affect cloud identities.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -358,14 +351,70 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
...
 
-    private SyncStatusDetail internalSynchronizeStackForDeleteUser(Stack stack, String deletedWorkloadUser) {
+    private void retrySyncIfBatchCallHasWarnings(Stack stack, UmsUsersState umsUsersState, boolean fullSync, Multimap<String, String> warnings,
+            FreeIpaClient freeIpaClient, boolean fmsToFreeipaBatchCallEnabled, UsersStateDifference usersStateDifferenceBeforeSync)
+            throws FreeIpaClientException, IOException {
+        if (fullSync && !warnings.isEmpty() && fmsToFreeipaBatchCallEnabled) {
+            UsersStateDifference usersStateDifferenceAfterSync = compareUmsAndFreeIpa(umsUsersState, fullSync, freeIpaClient);
+            if (usersStateDifferenceChanged(usersStateDifferenceBeforeSync, usersStateDifferenceAfterSync)) {
+                LOGGER.info(String.format("Sync was partially successful for %s, thus we are trying it once again", stack.getResourceCrn()));
+                applyDifference(umsUsersState, stack.getEnvironmentCrn(), warnings, usersStateDifferenceAfterSync, freeIpaClient, fmsToFreeipaBatchCallEnabled);
```
handavid, line 370: I thought a little more about the warnings. The retry is used to resolve the warnings so it doesn't make sense to return all the original warnings. What do you think of this?
```
Multimap<String, String> retryWarnings = ArrayListMultimap.create();
try {
    applyDifference(umsUsersState, stack.getEnvironmentCrn(), retryWarnings, usersStateDifferenceAfterSync, freeIpaClient, fmsToFreeipaBatchCallEnabled);
    warnings.clear();
} finally {
    warnings.putAll(retryWarnings);
}
```
The original warnings are combined with the retry warnings used if applyDifference raises an exception while only the retry warnings are used if applyDifference completes.

handavid (changes requested):
I left a few minor comments comment. Otherwise LGTM.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
horadla23: i'll put this into the retry method

handavid (approved):

PR #9787, opened 2021-01-14
CB-10617 Semi-private network: Datahub creation fails with Unable to configure load balancer

hreeve-cloudera:
Adds new logic to the NetworkV1ToNetworkV4Converter to detect the public endpoint gateway
is enabled for the environment, and if so selects and adds the subnet for the endpoint to
the AWS network object. The logic for selecting a subnet has been moved out of
EnvironmentBaseNetworkConverter where it originally lived and into a separate SubnetSelector.
Support for other cloud providers will be added as this feature is extended to new cloud
providers.

Tested with unit tests and by creating data hub and verifying LB creation was successful.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/loadbalancer/EndpointGatewayNetworkValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/loadbalancer/EndpointGatewayNetworkValidator.java
@@ -1,0 +1,73 @@
+package com.sequenceiq.cloudbreak.controller.validation.loadbalancer;
+
+import java.util.Optional;
+
+import javax.inject.Inject;
+
+import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.google.common.base.Strings;
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
+import com.sequenceiq.cloudbreak.converter.v4.environment.network.SubnetSelector;
+import com.sequenceiq.cloudbreak.validation.ValidationResult;
+import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
+import com.sequenceiq.cloudbreak.validation.Validator;
+import com.sequenceiq.common.api.type.PublicEndpointAccessGateway;
+import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentNetworkResponse;
+
+@Component
+public class EndpointGatewayNetworkValidator implements Validator<Pair<String, EnvironmentNetworkResponse>> {
+
+    static final String NO_BASE_SUBNET = "No cluster subnet id specified. Endpoint gateway subnet cannot be determined.";
+
+    static final String NO_BASE_SUBNET_META = "Could not determine availability zone of cluster subnet %s";
+
+    static final String NO_USABLE_SUBNET_IN_ENDPOINT_GATEWAY = "Unable to find public subnet in availability zone %s in " +
+        "provided endoint gateway subnet list.";
+
+    static final String NO_USABLE_SUBNET_IN_CLUSTER = "Unable to find public subnet in availability zone %s in cluster subnets.";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EndpointGatewayNetworkValidator.class);
+
+    @Inject
+    private SubnetSelector subnetSelector;
+
+    @Override
+    public ValidationResult validate(Pair<String, EnvironmentNetworkResponse> subject) {
+        String baseSubnetId = subject.getLeft();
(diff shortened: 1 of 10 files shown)


biharitomi (changes requested):

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/NetworkV1ToNetworkV4Converter.java
```diff
@@ -157,8 +170,18 @@ private AwsNetworkV4Parameters convertToAwsStackRequest(Pair<AwsNetworkV1Paramet
...
+                response.setSubnetId(value.getPreferedSubnetId());
+            }
+
+            if (PublicEndpointAccessGateway.ENABLED.equals(value.getPublicEndpointAccessGateway()) &&
+                    entitlementService.publicEndpointAccessGatewayEnabled(ThreadBasedUserCrnProvider.getAccountId())) {
+                Optional<CloudSubnet> endpointGatewaySubnet = subnetSelector.chooseSubnetForEndpointGateway(value, response.getSubnetId());
+                if (endpointGatewaySubnet.isPresent()) {
+                    response.setEndpointGatewaySubnetId(endpointGatewaySubnet.get().getId());
+                } else {
+                    throw new BadRequestException("Unable to identify public subnet to use for endpoint gateway.");
```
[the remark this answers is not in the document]
hreeve-cloudera: Thanks for letting me know about that! I think that's going to come in handy.

I've added a `EndpointGatewayNetworkValidator` and new unit tests for it to test each possible failure scenario.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/loadbalancer/EndpointGatewayNetworkValidator.java
```diff
@@ -0,0 +1,73 @@
...
+                        resultBuilder.error(String.format(NO_BASE_SUBNET_META, baseSubnetId));
+                    } else {
+                        if (!MapUtils.isEmpty(network.getGatewayEndpointSubnetMetas())) {
+                            LOGGER.debug("Attempting to validate endpoint gateway subnet using provided endpoint subnets.");
+                            Optional<CloudSubnet> endpointGatewaySubnet = subnetSelector.chooseSubnetForEndpointGateway(network, baseSubnetId);
+                            resultBuilder.ifError(endpointGatewaySubnet::isEmpty,
+                                String.format(NO_USABLE_SUBNET_IN_ENDPOINT_GATEWAY, baseSubnet.get().getAvailabilityZone()));
+                        } else {
+                            LOGGER.debug("Attempting to validate endpoint gateway subnet using cluster subnets.");
+                            Optional<CloudSubnet> endpointGatewaySubnet = subnetSelector.chooseSubnetForEndpointGateway(network, baseSubnetId);
```
enis, line 63: This statement is exactly the same one as the one in the `if` condition above. Seems fishy.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/NetworkV1ToNetworkV4Converter.java
```diff
@@ -157,8 +175,23 @@ private AwsNetworkV4Parameters convertToAwsStackRequest(Pair<AwsNetworkV1Paramet
...
+            if (entitlementService.publicEndpointAccessGatewayEnabled(ThreadBasedUserCrnProvider.getAccountId())) {
+                ValidationResult validationResult = endpointGatewayNetworkValidator.validate(new ImmutablePair<>(response.getSubnetId(), value));
+                if (validationResult.getState() == ValidationResult.State.ERROR || validationResult.hasError()) {
+                    throw new BadRequestException("Endpoint gateway subnet validation failed: " + validationResult.getFormattedErrors());
+                }
+
+                if (PublicEndpointAccessGateway.ENABLED.equals(value.getPublicEndpointAccessGateway())) {
+                    Optional<CloudSubnet> endpointGatewaySubnet = subnetSelector.chooseSubnetForEndpointGateway(value, response.getSubnetId());
+                    if (endpointGatewaySubnet.isPresent()) {
+                        // This should always be true because the endpointGatewayNetworkValidator verifies this optional is set
```
enis, line 191: Maybe throw an UnexpectedException() or something similar. Or maybe add an `assert false` statement in else?

enis:
I also gave this one a test together with #9803, and I was able to create a DH with 2 private subnets, 2 public subnets, and the LB binding to the public subnet.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/loadbalancer/EndpointGatewayNetworkValidator.java
hreeve-cloudera: The only reason it's broken out like this is so that when I set the error message, I can be very specific about which fields weren't set. I've refactored it a bit so that we only need to have the `subnetSelector.chooseSubnetForEndpointGateway` call once.

biharitomi (approved):
LGTM

PR #9789, opened 2021-01-15
CB-10668: Fix retries when ClusterProxyException is thrown

jamisonbennett:
The FreeIPA management service was fix so that when a
ClusterProxyExcpetion is thrown, the HTTP response status code can be
decoded and it can be converted into a retryable exception.

This was tested on a local deployment of cloudbreak.

### DIFF
--- a/cluster-proxy/src/main/java/com/sequenceiq/cloudbreak/clusterproxy/ClusterProxyException.java
+++ b/cluster-proxy/src/main/java/com/sequenceiq/cloudbreak/clusterproxy/ClusterProxyException.java
@@ -1,11 +1,22 @@
 package com.sequenceiq.cloudbreak.clusterproxy;
 
+import java.util.Optional;
+
 public class ClusterProxyException extends RuntimeException {
-    public ClusterProxyException(String message) {
+
+    private final Optional<ClusterProxyError> clusterProxyError;
+
+    public ClusterProxyException(String message, Optional<ClusterProxyError> clusterProxyError) {
         super(message);
+        this.clusterProxyError = clusterProxyError;
     }
 
     public ClusterProxyException(String message, Throwable cause) {
         super(message, cause);
+        clusterProxyError = Optional.empty();
+    }
+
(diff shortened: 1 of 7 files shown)


freeipa-client/build.gradle
```diff
@@ -26,6 +26,8 @@ dependencies {
   compile            group: 'org.glassfish.jersey.core',        name: 'jersey-common',               version: jerseyCoreVersion
   testImplementation group: 'org.mockito',                      name: 'mockito-core',                version: mockitoVersion
   testImplementation group: 'org.hamcrest',                     name: 'hamcrest',                    version: hamcrestVersion
+  compile            group: 'org.springframework.boot',         name: 'spring-boot-starter-web',     version: springBootVersion
```
lacikaaa, line 29: why do we need this dependency in the client?

cluster-proxy/src/main/java/com/sequenceiq/cloudbreak/clusterproxy/ClusterProxyException.java
```diff
@@ -1,11 +1,22 @@
...
 
+import java.util.Optional;
+
 public class ClusterProxyException extends RuntimeException {
-    public ClusterProxyException(String message) {
-        super(message);
+
+    private final Optional<ClusterProxyError> clusterProxyError;
+
+    public ClusterProxyException(ClusterProxyError clusterProxyError) {
```
lacikaaa, line 9: when `ClusterProxyException` construction happens `ClusterProxyError` is already an optional. Shall this constructor accept the optional form so the rewrapping won't be necessary?

freeipa-client/build.gradle
jamisonbennett: `freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClient.java` is the first file in this module that depends on cluster proxy module and the cluster proxy module uses spring `@Component` annotations.

lacikaaa (approved):

PR #9798, opened 2021-01-15
CB-10673: [CCMv2] Salt Pillars should be initialized

jszeles:
Pillars are extended with ccmv2 related value.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -48,6 +48,7 @@ import com.sequenceiq.cloudbreak.api.service.ExposedServiceCollector;
 import com.sequenceiq.cloudbreak.auth.CMLicenseParser;
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -582,6 +587,10 @@ private void saveGatewayPillar(GatewayConfig gatewayConfig, Cluster cluster,
         gateway.put("password", cluster.getPassword());
         gateway.put("enable_knox_ranger_authorizer", enableKnoxRangerAuthorizer);
 
+        String userCrn = stackService.get(stack.getId()).getCreator().getUserCrn();
+        final boolean ccmV2Enabled = entitlementService.ccmV2Enabled(Crn.safeFromString(userCrn).getAccountId());
```
lacikaaa, line 591: this must come from the tunnel and not from the entitlement. also final is not necessary

lacikaaa (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
TheTinkerDad: Fixed, thank you!

TheTinkerDad:
Opening a new PR as GitHub fails to pick up and amended commit and the review still points to the old commit.

PR #9799, opened 2021-01-15
CB-10358: Changes to user-data-helper script to support one way TLS between inverting proxy agent and nginx

jszeles:
We need to turn off ssl_verify_client in the ssl configuration, and should use localhost:9443.


orchestrator-salt/src/main/resources/salt/salt/nginx/conf/ssl.conf
```diff
@@ -1,6 +1,10 @@
 #curl --verbose --key ./key.pem --cert ./cert.pem -k --user "user:password" -H "Accept: application/json" https://104.155.27.67:9443/saltboot/health
 server {
-    listen       9443;
+    {%- if salt['pillar.get']('gateway:enable_ccmv2') %}
```
lacikaaa, line 3: please add default value, `false`. the same for the other condition

lacikaaa (changes requested):

lacikaaa (approved):

PR #9803, opened 2021-01-15
CB-10670 Datalake creation semi-private network creation fails with public subnets

hreeve-cloudera:
Adds new logic to load balancer creation to check if private subnets are available.
If not, then a public LB will be created instead of a private LB. The LB creation
logic was also moved from StackV4RequestToStackConverter to LoadBalancerConfigService,
and has been refactored to add more log statements about what conditions are being
evaluated, and to simplify the checks needed. New unit tests were added for the
new conditional checks.

Tested via unit tests, and by Ganesh Kumar Sankaran via applying the patch and
creating a data hub with semi-private networks enabled.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -13,6 +13,10 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.google.common.base.Preconditions;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.common.StackType;
+import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
+import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
+import com.sequenceiq.cloudbreak.cloud.model.CloudSubnet;
 import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateProcessor;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.knox.KnoxRoles;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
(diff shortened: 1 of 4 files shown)


cegganesh84:
> Tested via unit tests, and by Ganesh Kumar Sankaran via applying the patch and
creating a data hub with semi-private networks enabled.

Yes, I created a datalake with it and worked with public subnets which is the original issue. For datahub, I am hitting CB-10617 :-)

enis (commented):
Checked the main logic for entitlement checks and internal / external LB decisions. Seems pretty correct to me.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -115,4 +126,103 @@ public String getLoadBalancerUserFacingFQDN(Long stackId) {
...
+
+            // TODO CB-9368 - create target group for CM instances
+        }
+
+        return loadBalancers;
+    }
+
+    private boolean isLoadBalancerEnabled(DetailedEnvironmentResponse environment) {
+        return environment != null &&
+            (entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId()) ||
```
biharitomi, line 162: This part of the condition should check also the stack type, shouldn't it?
If datalakeLoadBalancerEnabled returns with true then a DH cluster evaluates to true and will have load balancer configured.

biharitomi (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: Thanks for that catch! I've changed the entitlement checks to be dependent on the stack type, and have also added a unit test case for a datahub stack with only the datalakeLoadBalancerEnabled entitlement, to make sure we don't create load balancers in that case.

biharitomi (approved):

enis (approved):

PR #9810, opened 2021-01-18
DISTX-559 DH Autoscaling Periscope Node Allocation.

smaniraju:
Periscope Node has a heart beat service which updates heartbeat in DB every 30 seconds indicating its aliveness.
Also there is a LeaderElectionService which polls active nodes which have updated heart beat in last 60 seconds.
Under certain scenarios when DB is slow or Node is slow and a single Leader heartBeat is delayed, Leader assumes it is not leader even though it is still leader and stops the process of allocating new DH Clusters to one of the periscope node.

Hence code is updated so that allocating new DH clusters is stopped only if active nodes are not empty.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
@@ -112,8 +112,8 @@ public class LeaderElectionService {
                         try {
                             long limit = clock.getCurrentTimeMillis() - heartbeatThresholdRate;
                             List<PeriscopeNode> activeNodes = periscopeNodeRepository.findAllByLastUpdatedIsGreaterThan(limit);
-                            reallocateOrphanClusters(activeNodes);
                             if (!activeNodes.isEmpty()) {
+                                reallocateOrphanClusters(activeNodes);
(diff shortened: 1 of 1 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/service/ha/LeaderElectionService.java
```diff
@@ -112,8 +112,8 @@ public void run() {
                         try {
                             long limit = clock.getCurrentTimeMillis() - heartbeatThresholdRate;
                             List<PeriscopeNode> activeNodes = periscopeNodeRepository.findAllByLastUpdatedIsGreaterThan(limit);
-                            reallocateOrphanClusters(activeNodes);
                             if (!activeNodes.isEmpty()) {
+                                reallocateOrphanClusters(activeNodes);
```
topolyai5, line 116: what will happen if the active nodes are empty?

smaniraju: When active nodes are empty, which is possible when there is a delayed heart beat, there is no action to be performed. The next timer cycle would continue processing the allocation when nodes update their heartbeat but the timer itself is not stopped which happens now.

topolyai5 (approved):

PR #9811, opened 2021-01-18
CB-10704 transform the CCM connectivity error message to an understan…

keyki:
…dable by the end user

It is a common case when customers launch clusters with CCM enabled and it fails due to different
networking issues. The error message that we prompt back to the user is not really understandable
by them. If developers are looking at the error message they immediatelly know that the customer
has connectivity issues, but it is not trivial to the customers. If we would provide a more clear
message to them where the issue is they might can debug and fix the error themselves.

A typical error message look like this:
com.sequenceiq.environment.exception.FreeIpaOperationFailedException: FreeIpa creation operation failed. FreeIpa creation failed. Status: 'CREATE_FAILED' statusReason: 'Status: 502 Bad Gateway Response: {"status":502,"code":"cluster-proxy.ccm.endpoint-unavailable","message":"Unable to get endpoint from CCM for key (i-nstanceid, GATEWAY)"}'

Not to mention that we also expose Cloudera internal details. This error message is changed as part of this PR to:
The Control Plane was not able to establish the connection with the CM server instance. This means that the reverse SSH tunnel (autossh process) running on the instance could not connect to the Cloudera server. Please check your connection and proxy settings and make sure the instance can reach *.ccm.cdp.cloudera.com'

As part of this PR, the Salt bootstrap phase is also fixed in FreeIPA to make the flow step fail upon
failure and do not ignore it.

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/message/StackStatusMessageTransformator.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/message/StackStatusMessageTransformator.java
@@ -1,0 +1,47 @@
+package com.sequenceiq.cloudbreak.message;
+
+import static org.apache.commons.lang3.StringUtils.isNoneBlank;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.annotation.PostConstruct;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+import org.yaml.snakeyaml.Yaml;
+
+import com.sequenceiq.cloudbreak.util.FileReaderUtils;
+
+@Component
+public class StackStatusMessageTransformator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StackStatusMessageTransformator.class);
+
+    private Map<String, String> statusMessagePatterns;
+
+    @PostConstruct
+    public void init() {
+        try {
+            String file = FileReaderUtils.readFileFromClasspath("messages/stack-status-messages.yaml");
+            statusMessagePatterns = new Yaml().load(file);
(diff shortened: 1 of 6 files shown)


common/src/main/resources/messages/stack-status-messages.yaml
```diff
@@ -0,0 +1 @@
+'cluster-proxy.ccm.endpoint-unavailable': 'The Control Plane was not able to establish the connection with the CM server instance. This means that the reverse SSH tunnel (autossh process) running on the instance could not connect to the Cloudera server. Please check your connection and proxy settings and make sure the instance can reach *.ccm.cdp.cloudera.com'
```
lacikaaa, line 1: I think it can be misleading to call it `CM server instance`, especially this is used for FreeIPA only

common/src/main/java/com/sequenceiq/cloudbreak/message/StackStatusMessageTransformator.java
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            String file = FileReaderUtils.readFileFromClasspath("messages/stack-status-messages.yaml");
+            statusMessagePatterns = new Yaml().load(file);
+            LOGGER.info("Status messages loaded for stack: {}", statusMessagePatterns);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't load stack status messsages", e);
+        }
+    }
+
+    public String transformMessage(String rawMessage) {
```
lacikaaa, line 36: a log line would be nice if the message is transformed with the original one. just in case

lacikaaa (changes requested):

common/src/main/resources/messages/stack-status-messages.yaml
keyki: Good point.

lacikaaa (approved):

PR #9812, opened 2021-01-18
CB-10597 adding new exception mapper to handle AwsRegionSelection failure, and prompt new error message when CredentialValidation is failing

afarsang:


environment/src/main/java/com/sequenceiq/environment/credential/service/ServiceProviderCredentialAdapter.java
```diff
@@ -43,6 +43,9 @@
 @Component
 public class ServiceProviderCredentialAdapter {
 
+    public static final String FAILED_CREDETIAL_VERIFICATION_MESSAGE = "Failed to verify the credential "
+            + "(try later if policies and roles are newly created or modified): ";
```
gregito, line 47: Maybe we could be a little bit more precise about how much should one wait until retry their request. "later" can be a couple of mins or a day.. I mean only if we can be more precise with a rough estimation like "a couple of minutes later" or something like this. What do you think?

afarsang: honestly don't know. I've got many similar messages without further indicator about the time frame, so I thought it will be alright here as well

environment/src/main/java/com/sequenceiq/environment/exception/mapper/AwsDefaultRegionSelectionExceptionMapper.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.aws.AwsDefaultRegionSelectionFailed;
+
+@Component
+public class AwsDefaultRegionSelectionExceptionMapper extends BaseExceptionMapper<AwsDefaultRegionSelectionFailed> {
+
+    @Override
+    Response.Status getResponseStatus(AwsDefaultRegionSelectionFailed exception) {
+        return Response.Status.FORBIDDEN;
```
topolyai5, line 14: what is the reason to FORBIDDEN?

biharitomi: because with the specified credentials CB is unable to verify regions as it gets Forbidden back from AWS

biharitomi (approved):
LGTM and thanks!

environment/src/main/java/com/sequenceiq/environment/exception/mapper/AwsDefaultRegionSelectionExceptionMapper.java
topolyai5: I didn't see anywhere in the code this but ok for me

topolyai5 (approved):

PR #9813, opened 2021-01-18
CB-10510. Create new cluster service: CDP nodestatus monitor (on mast…

oleewere:
…er node)

details:
- create a new service on the salt master nodes as cdp-nodestatus-monitor
- add network connectivity check agaainst databus for salt highstate (disabled by default)
- cdp-nodestatus-monitor should run around every hour and refresh the states based on cdp-doctor outputs
- expose nodestatus app through nginx (both for DL/DH and FreeIPA clusters)

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/TelemetryClusterDetails.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/TelemetryClusterDetails.java
@@ -25,6 +25,10 @@ public class TelemetryClusterDetails {
 
     private final String version;
 
+    private final String databusEndpoint;
+
+    private final boolean databusEndpointValidation;
+
     private TelemetryClusterDetails(Builder builder) {
         this.name = builder.name;
         this.type = builder.type;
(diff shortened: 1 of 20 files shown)


oleewere:
functional testing:
start a freeipa/DL without the changes
on DL:
- apply all the changes on the nodes (without updating pillars) -> run highstate ->cdp-nodestatus-monitor has started properly, nginx also has started:
```bash
Succeeded: 135 (changed=5)
Failed:      0
--------------
Total states run:     135
Total run time:     6.572 s
```

telemetry.cdpTelemetryVersion was moved out, but it is gathered from the `grains.roles`, so moving the version calculation does not have any effect (it's not in pillars)
- if databusEndpoint or validation is missing -> no databus validation is running (also check cdpTelemetryVersion, so that won't happen on old images)
- server username and password for cdp-nodestatus monitor is not set yet on the Java side (will be done with the additional java code changes that will include a client which will talk with the cdp-nodestatus monitor server), based on the code, all of this can work without the nodestatus pillars

keyki (approved):

PR #9815, opened 2021-01-18
CB-10251 Rename SparkUri to MockUri. Annotation processing improved, now the uri can be generated from interface name.

topolyai5:
Pro:
You can easily decide what is the right URL that you want to mock. CM has a lot of `hosts`, `commands` URL. 
Less string duplication, the name is generated from the interface name, e.g:
```
ClouderaManagerEndpoints.CmV31Api.ClustersByClusterName.Hosts.class ->
/{mockUuid}/api/v31/clusters/{clusterName}/hosts
```

Cons:
At the beginning harder to read or add a new endpoint but we will edit rarely.  


doktoric (approved):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/mock/MockUriNameParser.java
```diff
@@ -0,0 +1,102 @@
+package com.sequenceiq.it.cloudbreak.dto.mock;
+
+import java.beans.Introspector;
+import java.lang.reflect.Method;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
```
bergerdenes, line 7: I think we don't need this.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/mock/MockUriNameParser.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+            if (StringUtils.isNotBlank(sb) && !uri.startsWith("/")) {
+                sb.append("/");
+            }
+            sb.append(uri);
+            outerClassName = getInnerClassName(outerClassName, className);
+        }
+        if (StringUtils.isEmpty(sb)) {
+            throw new IllegalArgumentException(
+                    clazz.getName() + " SparkUri annotation url value is empty"
```
bergerdenes, line 43: annotation name is MockUri from now on

bergerdenes (changes requested):
2 minor things

bergerdenes (approved):

PR #9818, opened 2021-01-19
CB-10712. Increase timeout for CM client even further to mitigate dep…

oleewere:
…loy client issues

details:
- increase CM client proxy timeout to 4 min
- write/read/connect timeout should be larger than that value, so set those to 245 seconds

### DIFF
--- a/autoscale/src/main/resources/application.yml
+++ b/autoscale/src/main/resources/application.yml
@@ -60,10 +60,10 @@ cb:
     contextPath: "/cb"
   cm:
     client:
-      cluster.proxy.timeout: 120000
-      connect.timeout.seconds: 125
-      read.timeout.seconds: 125
-      write.timeout.seconds: 125
+      cluster.proxy.timeout: 240000
+      connect.timeout.seconds: 245
+      read.timeout.seconds: 245
+      write.timeout.seconds: 245
 
 rest:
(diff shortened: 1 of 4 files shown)


lacikaaa (approved):

biharitomi (approved):

biharitomi (changes requested):
The CCM related clients also need to be revised.

biharitomi (approved):

biharitomi:
I approved the PR, but afaik the timeout needs to be increased in cluster-proxy its self cc: @cegganesh84 
Based on Cluster-Proxy related follow up within: https://jira.cloudera.com/browse/ENGESC-5493

doktoric:
I think the base branch should be 2.36 right ?

oleewere:
@doktoric , not sure, for a use case like that we can recommend this: CB-10725 (see in internal jira)

pdarvasi (approved):

pdarvasi:
@akanto What's your opinion, should we merge this one?

bergerdenes (approved):

akanto:
This is an acceptable temporary workaround. It does not solve the real problem, but it increases the timeout over the 3.5 minutes that can be found in CM.

PR #9826, opened 2021-01-19
CB-10735 Semi-private network entitlement check incorrect

hreeve-cloudera:
Adds a check in EnvironmentBaseNetworkConverter to only trigger the semi-private
network logic if the feature is enabled for the environment.

Tested with a new unit test.


enis (approved):
LGTM.

biharitomi (approved):

PR #9827, opened 2021-01-20
CB-10699 Datahub HA template fails with missing ngninx ssl conf

orngejaket:
This splits the gateway or knox logic into their own and sets the
nginx.init on the gateway instance instead of the knox instance as
DH HA has knox deployed on non gateway nodes.


keyki (approved):

cegganesh84:
Thanks @orngejaket and @keyki

PR #9829, opened 2021-01-20
CB-8677 Fix FreeIPA HA in GCP (take 3)

cegganesh84:
1. FreeIPA does not work with link-local IP addresses in 4.6.6 https://lists.fedorahosted.org/archives/list/freeipa-users@lists.fedorahosted.org/thread/U2UKIJ424CVAITZDHP736AYNP7VPKAAH/
2. It is hard to use 4.7.0 with Centos 7. For two reasons, 4.7.0 is a bad version so there is no repository support and also it is backward incompatible with other packages installed in the system.
3. We can not move to Centos 8, where 4.8.0 is supported, as it is getting deprecated sooner than Centos 7.
4. So reverting to a bug for GCP when there is a failover scenario we can not resolve internet hostnames until the replica is rebooted.
5. This seems to be a much better compromise than not supporting HA at all for freeipa.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
@@ -13,7 +13,6 @@ echo "The first ipaddress of the host is $IPADDR"
 if [ ! -f /etc/resolv.conf.orig ]; then
   cp /etc/resolv.conf /etc/resolv.conf.orig
 fi
-FORWARDERS=$(grep -Ev '^#|^;' /etc/resolv.conf.orig | grep nameserver | awk '{print "--forwarder " $2}');
 
 install -m644 /etc/resolv.conf.install /etc/resolv.conf
 
@@ -32,6 +31,15 @@ ipa-client-install \
   --force-join \
   --ssh-trust-dns \
   --no-ntp
+
+FORWARDERS=$(grep -Ev '^#|^;' /etc/resolv.conf.orig | grep nameserver | awk '{print "--forwarder " $2}')
+PRIMARY_IPA=$(grep -Ev '^#|^;' /etc/resolv.conf | grep nameserver | awk '{print $2}')
+
+if [[ "${FORWARDERS}" == *" 169.254."* ]]; then
+  echo "IPA does not work with link-local IP addresses, so not using it as the forwarder"
+  FORWARDERS="--forwarder $PRIMARY_IPA --auto-forwarders "
+  cp /etc/resolv.conf.orig /etc/resolv.conf
+fi
 
 ipa-replica-install \
           --setup-ca \


lacikaaa:
so what would be set as forwarder in case of `--auto-forwarders`?

cegganesh84:
> so what would be set as forwarder in case of --auto-forwarders?

There are no arguments for `--auto-forwarders`. It will use the primary replica, which will not work in the failover scenario, unless the system/network is restarted. Basically reverts what was done for CB-8329, but only for GCP.

lacikaaa:
it doesn't look like a long term solution. shall we merge this to unblock some test and create a followup to work on the final one?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
```diff
@@ -13,7 +13,14 @@ echo "The first ipaddress of the host is $IPADDR"
 if [ ! -f /etc/resolv.conf.orig ]; then
   cp /etc/resolv.conf /etc/resolv.conf.orig
 fi
-FORWARDERS=$(grep -Ev '^#|^;' /etc/resolv.conf.orig | grep nameserver | awk '{print "--forwarder " $2}');
+NAMESERVER=$(grep -Ev '^#|^;' /etc/resolv.conf.orig | grep nameserver | awk '{print $2}')
+
+if [[ "${NAMESERVER%.*.*}" == "169.254" ]]; then
+  echo "IPA does not work with link-local IP addresses"
+  FORWARDERS="--auto-forwarders"
```
jamisonbennett, line 20: This basically means that GCP will not be HA. It will only be able to resolve external DNS queries via the original primary gateway. If that node is stopped, experiencing an outage, or has been repaired, then it will fail. In the case of rebuilding the primary gateway that failure will be permanent.

I don't know what we do in this case, but this will result in the customer being unable to access external data sources after a repair.

Is it possible to configure the GCP DNS servers to be allocated an IP address that is not link-local?

jamisonbennett (changes requested):

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
lacikaaa: I'm worried about bringing patches to freeipa and maintaining our own distribution in the end :)

cegganesh84: All great points. How about I set the forwarder to `8.8.8.8` in the replica.

Few things,

1. The experiences for GCP are planned for H2 2021.
2. The datahub and datalake nodes do not need FreeIPA to resolve external DNS addresses as they have unbound.
3. Patching is a good option, but as @lacikaaa mentioned there might be maintenance concerns. Also it will take time to do it cleanly for an extended period of maintenance.

@lacikaaa @jamisonbennett please let me know what you think.

lacikaaa: @cegganesh84 so the primary instance creation succeeds and it has the right DNS server configured. Is this statement right?
If it's right, could we configure the forwarder for the replicas after the installation?

cegganesh84: Good question. No, the primary has auto-forwarder only. Which it will point to the global DNS via the /etc/resolv.conf. This is the problem only for the replicas. Replica's /etc/resolv.conf should point to the primary except for failover scenarios. @jamisonbennett please correct me here if I am making too many bad assumptions.

jamisonbennett: > This is the problem only for the replicas.
Yes

> Replica's /etc/resolv.conf should point to the primary except for failover scenarios.
No, the replicas should always reference the upstream DNS server. The primary instance and replicas should all identical DNS records since that data is stored in LDAP and replicated across all the nodes. When DNS resolution is external to FreeIPA (e.g. www.google.com), then we want to resolve using the upstream DNS servers.

> How about I set the forwarder to 8.8.8.8 in the replica.
While this would resolve public entries, I fear it may miss DNS records for other nodes in the VPC that are outside of FreeIPA's zone. It may also miss other entries.

cegganesh84: > The primary instance and replicas should all identical DNS records since that data is stored in LDAP and replicated across all the nodes.

How about setting the forwarder to 127.0.0.1. That should point to /etc/resolv.conf anyways.

cegganesh84: > How about setting the forwarder to 127.0.0.1. That should point to /etc/resolv.conf anyways.

That would mean the ability to run unbound in a FreeIPA node as well, which is not possible today.

> The primary instance and replicas should all identical DNS records since that data is stored in LDAP and replicated across all the nodes.

@jamisonbennett I will go back to this. What purpose does /etc/resolv.conf serve in FreeIPA replica? In other words how --auto-forwarders work differently in primary vs replica?

cegganesh84: https://www.systutorials.com/docs/linux/man/1-ipa-replica-install/
```
--forwarder=IP_ADDRESS
Add a DNS forwarder to the DNS configuration. You can use this option multiple times to specify more forwarders, but at least one must be provided, unless the --no-forwarders option is specified.
--no-forwarders
Do not add any DNS forwarders. Root DNS servers will be used instead.
--auto-forwarders
Add DNS forwarders configured in /etc/resolv.conf to the list of forwarders used by IPA DNS.
```

@jamisonbennett It looks like we want the behavior of `--auto-forwarders` only in the replica too. So there is no need for explicitly setting to the one in /etc/resolv.conf.orig, we could simply swap the files in the replica for GCP.

cegganesh84: I think the confusion was because of how the DNS is bootstrapped CB-8271.

Because of which the primary's /etc/resolv.conf is pointing to the cloud provider DNS and the replica's /etc/resolv.conf is pointing to the primary's ip-address.

lacikaaa: you can't as the replica must find the primary during installation, and that works only if the primary is set as DNS server

jamisonbennett: I think we have the following options:
1. Revert CB-8271 and use `--auto-forwarders`
CB-8271 allowed every instance of FreeIPA to no longer have a /etc/hosts which pointed to every other instance of FreeIPA. It was a much cleaner bootstrapping, especially when you consider a cluster that had many repairs. There are some potential problems with having lots of old entries in /etc/hosts, if we are somehow still referencing the hostname somewhere it will attempt to go to an IP address that may be from a different cluster of FreeIPA (e.g. if the IP was reused) rather than getting a clear error that the host does not exist. We may need to be more vigilant in cleaning up /etc/hosts.

> you can't as the replica must find the primary during installation, and that works only if the primary is set as DNS server

I think it may actually be possible if:
* all the FreeIPA servers setup /etc/hosts with every FreeIPA servers IP address during provisioning and repair
* replicas don't join the cluster first but the are just installed as a part of the cluster
* we add some cleanup for all FreeIPA instances /etc/hosts after repair

But this still has a downside. CB-8356 becomes less possible. I don't know how we will be able to update the DNS forwarders after installation to 169.x.x.x IP addresses unless FreeIPA is patched to support it.

2. Patch FreeIPA via salt to work with the GCP IP addresses for DNS forwarding.
It looked like FreeIPA possibly fixed this for newer versions.

jamisonbennett: I think the option to revert CB-8271 and use `--auto-forwarders` is the quickest solution but patching FreeIPA to work with the GCP IP addresses is a better technical (assuming the patch actually works cleanly against our version of FreeIPA, I didn't test it). I know there has been concerns about patching FreeIPA, so I am good with either option.

cegganesh84: Thanks for the input folks. I took a slightly different approach. Wanted to communicate in terms of code this time. Please review the latest patch.

cegganesh84: I actually tested this latest change and it worked.

The replica install succeeded.

The forwarders are used correctly `2021-01-29T06:20:05Z DEBUG will use DNS forwarders: [CheckedIPAddressLoopback('10.114.16.44'), '169.254.169.254']`

It automagically changes the resolv.conf to point to itself after that,

```
2021-01-29T06:28:44Z DEBUG   [8/8]: changing resolv.conf to point to ourselves
2021-01-29T06:28:44Z DEBUG Backing up system configuration file '/etc/resolv.conf'

```

Screen Shot 2021-01-28 at 10 36 12 PM

lacikaaa:
thanks, it looks better to me.
But one thing is missed, this won't handle if there are multiple nameservers in the resolve.conf.orig

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
```diff
@@ -44,6 +54,7 @@ ipa-replica-install \
           --mkhomedir \
           --ip-address "$IPADDR" \
           $FORWARDERS \
+          --auto-forwarders \
```
jamisonbennett, line 57: I see this as a problem.

Scenario:

Initial config:
> ipaserver0 (primary) = 10.0.0.1 with forwarders 169.254.0.1 
ipaserver1 (replica) = 10.0.0.2 with forwarders 169.254.0.1 & 10.0.0.1 

Repair ipaserver0:
> ipaserver1 (replica) = 10.0.0.2 with forwarders 169.254.0.1 & 10.0.0.1 
ipaserver2 (replica promoted to new primary) = 10.0.0.3 with forwarders 169.254.0.1 & 10.0.0.2 

A new cluster is provisioned and 10.0.0.1 is allocated to a new FreeIPA cluster. DNS query goes to 10.0.0.3, it gets forwarded to 10.0.0.2, it gets forwarded to 10.0.0.1, and 10.0.0.1 responds with information from a cluster that is supposed to be isolated from this cluster.

1. Is there anything that can be done to remove the forwarder after `ipa-replica-install`?
2. If not, could we at least limit `--auto-forwarders` to only being used for the "192.254.x.x" case above?

lacikaaa: in theory `ipa dnsconfig-mod` could be used to modify forwarders, but I never tested

cegganesh84: Thanks for the feedback. We could do option 2. "least limit --auto-forwarders to only being used for the "169.x.x" case above"

cegganesh84:
> thanks, it looks better to me.
> But one thing is missed, this won't handle if there are multiple nameservers in the resolve.conf.orig

Good feedback. I checked all the three cloud providers there are no multiple nameservers.

jamisonbennett (approved):

lacikaaa:
> Good feedback. I checked all the three cloud providers there are no multiple nameservers.

that doesn't matter, as it is the default setting. customers can define their own nameservers, and they can have multiple, and we should use all of them. eg. check cloudera VPCs

cegganesh84:
> that doesn't matter, as it is the default setting. customers can define their own nameservers, and they can have multiple, and we should use all of them. eg. check cloudera VPCs

Ok, at least it is a new feature. I did not change the logic of getting nameserver from `/etc/resolv.conf.orig`. It was there already. All I added was new logic of getting nameserver from only `/etc/resolv.conf`, which can be only one.

cegganesh84:
@lacikaaa Could this be merged, please? QE is blocked testing this feature.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
```diff
@@ -33,6 +32,17 @@ ipa-client-install \
   --ssh-trust-dns \
   --no-ntp
 
+CLOUD_NAMESERVER=$(grep -Ev '^#|^;' /etc/resolv.conf.orig | grep nameserver | awk '{print $2}')
```
lacikaaa, line 35: this is breaking existing functionality:
```
# cat test
nameserver 1.2.3.4
nameserver 5.4.3.2
```
old output:
```
# grep -Ev '^#|^;' test | grep nameserver | awk '{print "--forwarder " $2}'
--forwarder 1.2.3.4
--forwarder 5.4.3.2
```
new output:
```
grep -Ev '^#|^;' test | grep nameserver | awk '{print $2}'
1.2.3.4
5.4.3.2
```

so the new one won't handle when there is multiple nameserver on the network

lacikaaa (changes requested):

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_replica_install.sh
cegganesh84: Got it. Will fix it. I missed the awk change.

cegganesh84: Sorry about this, resolved the issue now.

lacikaaa (approved):

PR #9830, opened 2021-01-20
CB-10756 - fix for NPE when one wants to obtain the value of a tag for an empty SdxClusterDetailResponse

gregito:
fix for NPE when one wants to obtain the value of a tag for an empty SdxClusterDetailResponse

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterDetailResponse.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterDetailResponse.java
@@ -33,7 +33,8 @@ public class SdxClusterDetailResponse extends SdxClusterResponse implements Tagg
 
     @Override
     public String getTagValue(String key) {
-        return Optional.ofNullable(stackV4Response.getTags())
+        return Optional.ofNullable(stackV4Response)
+                .map(stack -> stack.getTags())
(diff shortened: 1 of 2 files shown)


datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterDetailResponse.java
```diff
@@ -33,7 +33,8 @@ public void setStackV4Response(StackV4Response stackV4Response) {
 
     @Override
     public String getTagValue(String key) {
-        return Optional.ofNullable(stackV4Response.getTags())
+        return Optional.ofNullable(stackV4Response)
+                .map(stack -> stack.getTags())
```
topolyai5, line 37: looks like, `getTags()` can be null

topolyai5 (changes requested):

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterDetailResponse.java
gregito: yeah, that's why mapping it. if it is null then the lambda evaluates it as an empty result hence the - expected - null shall return

topolyai5 (approved):

PR #9833, opened 2021-01-20
CB-10747 remove the CM license log message when we decorate the Salt …

keyki:
…pillars

When we started to integrate with archive.cloudera.com we introduced a message
that would show the used CM license so we can check it is properly used. This
log message is no necessary and it exposes the customers' licenses as well.


tiborpopovics (approved):

lacikaaa:
would it make sense to log some part of the licence? like the header with the id and the expiration date

keyki:
> would it make sense to log some part of the licence? like the header with the id and the expiration date

Even the header is sensitive, so we would need to filter from there as well.

PR #9836, opened 2021-01-20
CB-10788. Enhance flow service to be able to query flow state by flow type.

oleewere:
detaails:
- Long term: provide different endpoints for diagnostics to follow progress of flows without providing the ids
- at first step, that can be done to query flows by type (only for flow logs for now, can be done for flow chains later)


foldik (approved):
LGTM

PR #9837, opened 2021-01-20
CB-10077 Adds support to handle image catalogs without versions block.

pkedvessy:


core/src/main/java/com/sequenceiq/cloudbreak/service/image/VersionBasedImageProvider.java
```diff
@@ -0,0 +1,120 @@
+package com.sequenceiq.cloudbreak.service.image;
```
pkedvessy, line 1: This code is not a new one, it is coming from the ImageCatalogService.

core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceDefaultNotFoundTest.java
```diff
@@ -61,10 +64,29 @@
...
+
+    @InjectMocks
+    private VersionBasedImageCatalogService versionBasedImageCatalogService;
+
     @Before
     public void beforeTest() {
         ReflectionTestUtils.setField(underTest, "cbVersion", "5.0.0");
         ReflectionTestUtils.setField(underTest, "defaultCatalogUrl", "");
+        ReflectionTestUtils.setField(underTest, "imageCatalogServiceProxy", imageCatalogServiceProxy);
+        ReflectionTestUtils.setField(imageCatalogServiceProxy, "advertisedImageCatalogService", advertisedImageCatalogService);
```
Bajzathd, line 87: @InjectMocks and @Mock annotations should take care of these

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
```diff
@@ -121,10 +115,7 @@
     private TransactionService transactionService;
 
     @Inject
-    private PrefixMatcherService prefixMatcherService;
-
-    @Inject
-    private LatestDefaultImageUuidProvider latestDefaultImageUuidProvider;
+    private ImageCatalogServiceProxy imageCatalogServiceProxy;
```
Bajzathd, line 118: ImageCatalogServiceProxy injected into ImageCatalogService, sounds like it should be backwards. I think the version and advertised services are the strategies, and ImageCatalogServiceProxy is more like ImageCatalogHandler? Naming things is really hard :smile:

core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceDefaultNotFoundTest.java
pkedvessy: It is true, when you would inject mocks but in this case I'm "injecting" real implementations with mocked dependencies. This way we can make sure the refactored implementation is in line with the original one.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
pkedvessy: Yes it is :-). We use similar naming convention in CB for the same pattern. In my point of view it is not the case where we should use "handler" naming. I think the best would be to rename the original ImageCatalogService to ImageService but it might mess veterans up :-D

Bajzathd (approved):

PR #9840, opened 2021-01-21
CB-10775 Missing LB metdata causing Datalake stack creation time out

hreeve-cloudera:
Improved the logic to set the load balancer type so that it's based on the "subnetId"
network parameter being passed in for stack creation. If "subnetId" is not set, the
load balancer object will not be created. Additionally, error in load balancer metadata
collection now results in a log message instead of throwing an exception.

Tested with new unit tests and by creating a data lake and data hub and verifying the
expected network values were all set, and the LB created was successfully and metadata
collected.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -2,12 +2,14 @@ package com.sequenceiq.cloudbreak.service;
 
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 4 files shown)


keyki:
Can you please send the PR to 2.37?

bergerdenes:
Please add more logging.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -157,6 +176,29 @@ public String getLoadBalancerUserFacingFQDN(Long stackId) {
...
+
+    private boolean isNetworkUsingPublicSubnet(Network network, EnvironmentNetworkResponse envNetwork) {
+        return isSelectedSubnetAvailableAndRequestedType(network, envNetwork, false);
+    }
+
+    private boolean isSelectedSubnetAvailableAndRequestedType(Network network, EnvironmentNetworkResponse envNetwork, boolean isPrivate) {
+        if (network != null) {
+            Json attributes = network.getAttributes();
+            Map<String, Object> params = attributes == null ? Collections.emptyMap() : attributes.getMap();
+            String subnetId = String.valueOf(params.get(SUBNET_ID));
```
bergerdenes, line 191: this might cause an NPE if params map does not have SUBNET_ID

bergerdenes (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: It won't cause an NPE. It will however convert a null value to the string "null", which isn't what I want to happen. I've changed this line to do a null check and added new unit tests to test that everything works when `network`, `network.getAttributes()`, and `params.get(SUBNET_ID)` are each null.

hreeve-cloudera:
> Please add more logging.

I've added more logging to the changed parts of the code. I have CB-10773 marked to do a in-depth examination of the logging around load balancer creation, and to add a lot more there. That's separate from this change because I wanted to get this change in to unblock people. But if there's a specific scenario you want logged more let me know.

hreeve-cloudera:
> Can you please send the PR to 2.37?

Done.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
bergerdenes: no, actually String.valueOf(null) is indeed NPE

bergerdenes (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: The code of `String.valueOf` is:
```
    public static String valueOf(Object obj) {
        return obj == null ? "null" : obj.toString();
    }
```
That doesn't look like it'd throw an NPE to me?

PR #9850, opened 2021-01-21
CB-10794 Datalake creation fails on mow-dev with unable to configure load balancer

hreeve-cloudera:
The error was happening because AwsLaunchService was looking for the public subet
in the "endpointGatewaySubnetId" settings, and wasn't using the "subnetId" setting
for the public load balancer even when that's what should have been used. I added
logic to fall back to "subnetId" if "endpointGatewaySubnetId" isn't set. Also added
an additional check to make sure the load balancer conversion was successful, and
added more logging around the creation.

Tested with a new unit test class.


keyki:
Shouldn't this go to 2.37 as well?

hreeve-cloudera:
> Shouldn't this go to 2.37 as well?

@keyki Sorry, it is in 2.37. I just forgot to change the base branch of the PR.

biharitomi (approved):

PR #9851, opened 2021-01-21
CB-10246 Refactor integration test project for real UMS tests

aszegedi:
Real UMS default account and environment should be able to parameterise via environment or application parameters.

Along with this change additional refactor steps had need to be done:
- Eliminate FreeIpaUserSyncTestDto wait for UserSyncState, improve setOperationId and wait for syncOperationStatus (getLastUserSyncOperationId)
- Make CloudbreakUserCache non-static
- Introduce CloudbrakActor for Actor interface
- Fix SetActingUser for UMS tests and CleanUp
- Fix DoAsAdmin RunningParamerer
- Make AuthorizationTestUtil non-static
- Fix CloudbreakUserCache init
- Fix acting user delegation in UMS tests and remove Admin from Distrox Refresh
- Introduce InitializedKeys to CloudbreakUserCache 
- Fix RealUmsUserCache init for Legacy account 
- Fix LEGACY_ACC_ENV_ADMIN user to legacy auth tests 
- Fix getAdminAccessKeyByAccountId 
- Introduce initMicroserviceClientsForUMSAccountAdmin


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/freeipa/FreeIpaGetLastSyncStatusAction.java
```diff
@@ -0,0 +1,32 @@
...
+    @Override
+    protected FreeIpaUserSyncTestDto freeIpaAction(TestContext testContext, FreeIpaUserSyncTestDto testDto, FreeIpaClient client) throws Exception {
+        Log.when(LOGGER, format(" Environment Crn: [%s], freeIpa Crn: %s", testDto.getEnvironmentCrn(), testDto.getRequest().getEnvironments()));
+        Log.whenJson(LOGGER, format(" FreeIPA get last sync status request: %n"), testDto.getRequest());
+        SyncOperationStatus syncOperationStatus = client.getFreeIpaClient()
+                .getUserV1Endpoint()
+                .getLastSyncOperationStatus(testDto.getEnvironmentCrn());
+        testDto.setOperationId(syncOperationStatus.getOperationId());
+        LOGGER.info("Last sync is in state: [{}], last sync operation: [{}]", syncOperationStatus.getStatus(),
+                OBJECT_MAPPER.writeValueAsString(syncOperationStatus));
```
horadla23, line 27: why do we need ObjectMapper here to log an object?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,158 @@
...
+                for (int j = 0; j < jsonarray1.length(); j++) {
+                    accountId = Crn.fromString(((JSONObject) jsonarray1.get(j)).getString("crn")).getAccountId();
+                    String displayName = ((JSONObject) jsonarray1.get(j)).getString("displayName");
+                    String desc = ((JSONObject) jsonarray1.get(j)).getString("desc");
+                    String crn = ((JSONObject) jsonarray1.get(j)).getString("crn");
+                    String accessKey = ((JSONObject) jsonarray1.get(j)).getString("accessKey");
+                    String secretKey = ((JSONObject) jsonarray1.get(j)).getString("secretKey");
+                    boolean admin = Boolean.parseBoolean(((JSONObject) jsonarray1.get(j)).getString("admin"));
+                    cloudbreakUsers.add(new CloudbreakUser(accessKey, secretKey, displayName, crn, desc, admin));
+                }
```
horadla23, line 113: to be honest, i am not a huge fan of extracting fields from jsonarray manually, we should create a model which has same structure and parse it, like before

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,158 @@
...
+                    .orElseThrow(() -> new TestFailException(String.format("There is no real UMS admin in account: %s", accountId)));
+            LOGGER.info(" Real UMS account admin has been found in repository:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
+                    adminUser.getDisplayName(), adminUser.getCrn(), adminUser.getAccessKey(), adminUser.getSecretKey(), adminUser.getAdmin());
+            return adminUser.getAccessKey();
+        } catch (Exception e) {
+            throw new TestFailException(String.format("Cannot get the real UMS admin in account: %s, because of: %s", accountId, e.getMessage()), e);
+        }
+    }
+
+    public CloudbreakUser getAdminByAccountId(String accountId) {
```
horadla23, line 141: this and previous method has almost same logic, you should create a common method for them

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/ums/UmsTestDto.java
```diff
@@ -23,8 +26,12 @@
 
     private static final String ENV_USER_CRN = "crn:altus:iam:us-west-1:altus:resourceRole:EnvironmentUser";
 
+    private static final String ENV_CREATE_CRN = "crn:altus:iam:us-west-1:altus:role:EnvironmentCreator";
```
horadla23, line 29: please rename it to `ENV_CREATOR_CRN`

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/manowar/EnvironmentLegacyAuthzGetTest.java
```diff
@@ -0,0 +1,72 @@
...
+import com.sequenceiq.it.cloudbreak.client.FreeIpaTestClient;
+import com.sequenceiq.it.cloudbreak.context.Description;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.environment.EnvironmentTestDto;
+import com.sequenceiq.it.cloudbreak.dto.freeipa.FreeIpaTestDto;
+import com.sequenceiq.it.cloudbreak.dto.telemetry.TelemetryTestDto;
+import com.sequenceiq.it.cloudbreak.testcase.authorization.AuthUserKeys;
+import com.sequenceiq.it.cloudbreak.testcase.e2e.AbstractE2ETest;
+
+public class EnvironmentLegacyAuthzGetTest extends AbstractE2ETest {
```
horadla23, line 21: this test logic almost the same as in `EnvironmentLegacyAuthzGetTest ` in `authorization` package, you should make here a common part also to remove duplications

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/manowar/UsersAndGroupsEnvironmentTests.java
```diff
@@ -0,0 +1,134 @@
...
+import com.sequenceiq.it.cloudbreak.dto.freeipa.FreeIpaUserSyncTestDto;
+import com.sequenceiq.it.cloudbreak.dto.sdx.SdxTestDto;
+import com.sequenceiq.it.cloudbreak.dto.telemetry.TelemetryTestDto;
+import com.sequenceiq.it.cloudbreak.dto.ums.UmsTestDto;
+import com.sequenceiq.it.cloudbreak.testcase.authorization.AuthUserKeys;
+import com.sequenceiq.it.cloudbreak.testcase.e2e.AbstractE2ETest;
+import com.sequenceiq.it.cloudbreak.util.spot.UseSpotInstances;
+import com.sequenceiq.sdx.api.model.SdxClusterStatusResponse;
+
+public class UsersAndGroupsEnvironmentTests extends AbstractE2ETest {
```
horadla23, line 29: lt's discuss this, since there is sync test already, merged yesteerday, and also some methods maybe not needed here

horadla23 (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakActor.java
```diff
@@ -0,0 +1,66 @@
...
+        return new CloudbreakUser(accessKey, secretKey, username + " at tenant " + tenantName);
+    }
+
+    @Override
+    public CloudbreakUser useRealUmsUser(String key) {
+        LOGGER.info("Getting the requested real UMS user by key:: {}", key);
+        return getByDisplayName(key);
+    }
+
+    public CloudbreakUser useRealUmsUser(String key, String environmentKey, String accountKey) {
```
lnardai, line 55: If we are initialising the cache with a concrete accountId, why do we need this accountKey method?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,152 @@
...
+    public CloudbreakUser getByDisplayName(String name) {
+        return getByDisplayName(name, commonCloudProperties.getUser().getRealUms().getEnvironmentKey(),
+                commonCloudProperties.getUser().getRealUms().getAccountKey());
     }
 
-    public CloudbreakUser getByName(String name) {
-        if (usersByAccount == null) {
-            initUsers();
+    public CloudbreakUser getByDisplayName(String name, String environmentKey, String accountKey) {
+        setRealUmsUserEnvironment(environmentKey);
```
lnardai, line 55: Why do we need to set a `Compomnent`'s field here? Is this even thread safe? 
Also it's not a really good pattern to change a @Value annotation on the go. Can't this initialization be done in a PostConstruct? and only be done once? Since we agreed we only run tests in a single account.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,152 @@
...
+
+        if (usersByAccount == null || !initializedKeys.containsKey(environmentKey) || !initializedKeys.containsValue(accountKey)) {
+            initUsers(environmentKey, accountKey);
         }
-        return usersByAccount.values().stream().flatMap(Collection::stream)
-                .filter(u -> u.getDisplayName().equals(name)).findFirst().get();
+        CloudbreakUser user = usersByAccount.values().stream().flatMap(Collection::stream)
+                .filter(u -> u.getDisplayName().equals(name)).findFirst()
+                .orElseThrow(() -> new TestFailException(String.format("There is no real UMS user with name: %s in account: %s", name, accountKey)));
+        LOGGER.info(" Real UMS user has been found in repository:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
```
lnardai, line 64: This "repository" is misleading for me, somewhat implies there is a JPA repository in the picture.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,152 @@
...
+
+    public void setRealUmsUserEnvironment(String realUmsUserEnvironmentKey) {
+        this.realUmsUserEnvironment = realUmsUserEnvironmentKey;
+    }
+
+    public void setInitializedKeys(Map<String, String> keys) {
+        this.initializedKeys = keys;
+    }
+
+    public void initUsers(String environmentKey, String accountKey) {
```
lnardai, line 97: What is the difference between environmentKey and accountKey?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/EnvironmentLegacyAuthzGetTest.java
```diff
@@ -29,11 +31,17 @@
...
+
     @Override
     protected void setupTest(TestContext testContext) {
-        useRealUmsUser(testContext, AuthUserKeys.LEGACY_NON_POWER);
-        useRealUmsUser(testContext, AuthUserKeys.LEGACY_POWER);
-        useRealUmsUser(testContext, AuthUserKeys.LEGACY_ACC_ENV_ADMIN);
+        useRealUmsUser(testContext, AuthUserKeys.LEGACY_NON_POWER, REAL_UMS_ENVIRONMENT_KEY, REAL_UMS_ACCOUNT_KEY);
+        useRealUmsUser(testContext, AuthUserKeys.LEGACY_POWER, REAL_UMS_ENVIRONMENT_KEY, REAL_UMS_ACCOUNT_KEY);
+        useRealUmsUser(testContext, AuthUserKeys.LEGACY_ACC_ENV_ADMIN, REAL_UMS_ENVIRONMENT_KEY, REAL_UMS_ACCOUNT_KEY);
+        initializeDefaultBlueprints(testContext);
```
lnardai, line 42: Why are these lines needed in the testcase?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakActor.java
aszegedi: We need this for favour of compatibility with old legacy test cases, while we are running these in same build with real UMS test cases. We are using other test account for Legacy. So we need to initialise the user cache for these separately during build time:
https://github.com/hortonworks/cloudbreak/blob/4edce862479c90528bdcdde52c3075c552e13edf/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/EnvironmentLegacyAuthzGetTest.java#L39-L41

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/EnvironmentLegacyAuthzGetTest.java
aszegedi: Originally these are: https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/EnvironmentLegacyAuthzGetTest.java#L34-L36
even all the further actions done by `LEGACY_ACC_ENV_ADMIN`:
- `credentialTestClient.create()`
- `environmentTestClient.create()`
- `freeIpaTestClient.create()`
as you can see.

So this is actually an environment creation test case with `LEGACY_ACC_ENV_ADMIN` user. So I've introduce the related steps in the test setup and, because of the setup part can be done in the `setupTest` with init blueprints and create credential and image catalog with this user...etc.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
lnardai: Would it be feasible to make this to an enum? That way if people see the possible values, like dev,int,stage,prod might get the idea on what this parameter means.

Or the other idea is to change it to deployment?

aszegedi: The new `resources/ums-users/api-credentials.json` structure: 
```
{
  "dev": {
    "default": [...],
    "legacy": [...]
  },
  "int": {},
  "stage": {}
}
```
here the `environmentKey` can be `dev, int, stage` and `accountKey` can be `default, legacy`.

aszegedi: Yes, we can change `environmentKey` to ENUM or rename to `deploymentKey`.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
aszegedi: No worries, I can change that part of the message if it is strange for you.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
aszegedi: Actually we run tests in 2 separate accounts in parallel at [cloudbreak-real-ums-test](http://ci-cloudbreak.eng.hortonworks.com/job/cloudbreak-real-ums-test/) and also at [integration-test-pr-real-ums-test
](http://ci-cloudbreak.eng.hortonworks.com/job/integration-test-pr-real-ums-test/).
However very useful if we can use environment variables or application parameters to set up default values for these tests.
Furthermore as far as I know Legacy tests wont be with us after a while and we can step forward to single account for single run concept.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -1,66 +1,152 @@
...
+        LOGGER.info("Real UMS users are initializing by environment: {} and account: {}. Repository is present at: {} path", environmentKey, accountKey,
+                userConfigPath);
         try {
-            this.usersByAccount = JsonUtil.readValue(
-                    FileReaderUtils.readFileFromClasspathQuietly(userConfigPath), new TypeReference<>() {
-                    });
-        } catch (IOException e) {
-            throw new RuntimeException(String.format("Can't read file: %s It's possible you did run make fetch-secrets", userConfigPath));
+            fetchedUserRepository = JsonUtil.readValue(
+                    FileReaderUtils.readFileFromClasspathQuietly(userConfigPath), new TypeReference<Map<String, Map<String, List<CloudbreakUser>>>>() { });
```
horadla23, line 108: i would read json file during a PostConstruct method, and we need only a method without parameters which gives back users based on the environment and account which should be one during the test

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/EnvironmentLegacyAuthzGetTest.java
horadla23: let's discuss this, i am not remembering fully what we have discussed regarding this last time

integration-test/src/main/resources/application.yml
```diff
@@ -36,6 +36,9 @@ integrationtest:
   redbeams:
     server: http://localhost
   user:
+    realUms:
+      accountKey: default
+      environmentKey: dev
```
horadla23, line 41: these are not user related informations, i would move them at least one level higher, also these structure would be simpler:
```
ums:
  account: default
  environment: dev
```
or something like that

horadla23 (changes requested):
let's discuss the CloudbreakUserCache part, it seems it is prepared for the case when we are running it agains multiple environment and account, but I think we need to run a test suite only against one environment and account at a time

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
aszegedi: Sounds good! Separate test jobs are sound feasible and more maintainable for me as well. I buy the idea!

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/freeipa/FreeIpaUserSyncAssertion.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private FreeIpaUserSyncAssertion() {
+    }
+
+    public static Assertion<FreeIpaUserSyncTestDto, FreeIpaClient> userSyncTakesLessThanGivenDuration(long givenDuration) {
+        return (testContext, freeIpaUserSyncTestDto, freeIpaClient) -> {
+            long duration = freeIpaUserSyncTestDto.getResponse().getEndTime() - freeIpaUserSyncTestDto.getResponse().getStartTime();
+
+            LOGGER.info(String.format("FreeIpa last user sync duration took %d sec", TimeUnit.MILLISECONDS.toSeconds(duration)));
+            assertThat(duration)
```
topolyai5, line 29: As I know, the exception is handled by our homework. The assert will throw an Error. We were an issue with this, the framework didn't handle the error correctly. The test finished there and validate nothing

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/freeipa/FreeIpaUserSyncAssertion.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public static Assertion<FreeIpaUserSyncTestDto, FreeIpaClient> userSyncDoneWithNoFailures() {
+        return (testContext, freeIpaUserSyncTestDto, freeIpaClient) -> {
+            List<FailureDetails> failures = freeIpaUserSyncTestDto.getResponse().getFailure();
+
+            if (failures != null && !failures.isEmpty()) {
+                LOGGER.info(String.format("Freeipa last user sync had the following issues: %s", Joiner.on(",").join(failures)));
+            }
+            assertThat(failures)
```
topolyai5, line 44: same

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/RunningParameter.java
```diff
@@ -27,17 +37,25 @@
...
 
-    public Actor getWho() {
-        if (doAsAdmin) {
-            if (CloudbreakUserCache.getInstance().isInitialized()) {
-                return Actor.useRealUmsUser(AuthUserKeys.ACCOUNT_ADMIN);
+    @Inject
+    private CloudbreakActor cloudbreakActor;
+
+    public CloudbreakUser getWho() {
+        if (Optional.ofNullable(doAsAdmin).orElse(false)) {
```
topolyai5, line 44: `doAsAdmin` cannot be null

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/RunningParameter.java
```diff
@@ -27,17 +37,25 @@
...
+            try {
+                if (Optional.ofNullable(cloudbreakActor.isInitialized()).orElse(false)) {
```
topolyai5, line 46: `cloudbreakActor.isInitialized` cannot be null

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java
```diff
@@ -82,18 +87,21 @@ public void testCreateEnvironmentWithDh(TestContext testContext) {
                 .await(EnvironmentStatus.AVAILABLE)
                 // testing unauthorized calls for environment
                 .whenException(environmentTestClient.describe(), ForbiddenException.class, expectedMessage("Doesn't have 'environments/describeEnvironment'" +
-                        " right on 'environment' " + environmentShortPattern(testContext)).withWho(Actor.useRealUmsUser(AuthUserKeys.ENV_CREATOR_B)))
+                        " right on 'environment' " + environmentShortPattern(testContext)).withWho(cloudbreakActor.useRealUmsUser(AuthUserKeys.ENV_CREATOR_B)))
                 .whenException(environmentTestClient.describe(), ForbiddenException.class, expectedMessage("Doesn't have 'environments/describeEnvironment'" +
-                        " right on 'environment' " + environmentShortPattern(testContext)).withWho(Actor.useRealUmsUser(AuthUserKeys.ENV_CREATOR_B)))
+                        " right on 'environment' " + environmentShortPattern(testContext)).withWho(cloudbreakActor.useRealUmsUser(AuthUserKeys.ZERO_RIGHTS)))
```
topolyai5, line 92: did you change the rights intentionally?

topolyai5 (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java
aszegedi: Yes. Unfortunately this was changed mistakenly by my previous [1c635b55a7b86ab312166a8d14e9fe25a29f76c9](https://github.com/hortonworks/cloudbreak/commit/1c635b55a7b86ab312166a8d14e9fe25a29f76c9#diff-a9ac0b63b70ca4d2b930b820ebe436a37685f6c9ba84f5cdf4b2c16690ac334f) commit. 

Original values are:
https://github.com/hortonworks/cloudbreak/blob/2a14ab75d9ebefc7727b96bd24917ad9210679fc/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java#L88-L92

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/freeipa/FreeIpaUserSyncAssertion.java
aszegedi: As far as I know: https://github.com/hortonworks/cloudbreak/blob/29c1eb460e221512c125cb9b6e09205ca7117046/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java#L318-L332
part should handle the failed assertions in the `then` steps.

topolyai5 (approved):
LGTM

horadla23 (approved):

PR #9853, opened 2021-01-24
CB-10813. Handle/interrupt CM deployClientConfig timeout during upsca…

oleewere:
…le/downscale

details:
Until deployClientConfig call on CM side is not async (only partly, that means it starts running the deployClientConfig commands but the API call itself is not returning and times out...it's mostly around 3.5 min but can take 12 mins etc.)
Implementation:
- Interrupt deployClientConfig http call on CB side
- we need to know the commandId -> next step is try out listing the active commands in CM
- if active commands will be empty (could happen if it finishes around that 5 sec), try to use /cmf/commands/commandTable call (based on the UI), that contains all commands, which returns a JSON (by setting the right cookie from the list active commands api command)
- set the timeout for the interrupt -> 120 sec, as ~130 sec is the timeout for the cluster proxy, so do the interrupt before that times out (that one is the lowest at the moment)
- also handle invalid or not found responses + added UTs

### DIFF
--- a/autoscale/src/main/resources/application.yml
+++ b/autoscale/src/main/resources/application.yml
@@ -64,6 +64,8 @@ cb:
       connect.timeout.seconds: 245
       read.timeout.seconds: 245
       write.timeout.seconds: 245
+      commands:
+        deployClientConfig.interrupt.timeout.seconds: 120
 
 rest:
   debug: false
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
@@ -1,0 +1,261 @@
+package com.sequenceiq.cloudbreak.cm;
+
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+import com.cloudera.api.swagger.ClustersResourceApi;
+import com.cloudera.api.swagger.client.ApiClient;
+import com.cloudera.api.swagger.client.ApiException;
+import com.cloudera.api.swagger.client.ApiResponse;
+import com.cloudera.api.swagger.client.Pair;
+import com.cloudera.api.swagger.model.ApiCommand;
+import com.cloudera.api.swagger.model.ApiCommandList;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
+import com.sequenceiq.cloudbreak.cm.model.CommandResource;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.service.CloudbreakException;
+import com.squareup.okhttp.HttpUrl;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.Response;
+import com.squareup.okhttp.ResponseBody;
+
+@Service
+public class ClouderaManagerDeployClientConfigProvider {
+
+    private static final Logger LOGGER =
+            LoggerFactory.getLogger(ClouderaManagerDeployClientConfigProvider.class);
+
+    private static final String CMF_COMMANDS_COMMAND_TABLE_URI_PATH = "/cmf/commands/commandTable";
+
+    private static final String DEPLOY_CLIENT_CONFIG_COMMAND_NAME = "DeployClusterClientConfig";
+
+    private static final int INTERVAL_MINUTES = 30;
+
+    private static final int ERROR_CODES_FROM = 400;
+
+    private final Integer interruptTimeoutSeconds;
+
+    public ClouderaManagerDeployClientConfigProvider(
+            @Value("${cb.cm.client.commands.deployClientConfig.interrupt.timeout.seconds:}") Integer interruptTimeoutSeconds) {
+        this.interruptTimeoutSeconds = interruptTimeoutSeconds;
+    }
+
+    /**
+     * Obtain deploy cluster client config command ID with different strategies:
+     * - first run deployClusterClientConfig command on CM, if it returns successfully,
+     *   get the command ID of the command.
+     * - if deployClusterClientConfig CM command times out, use listActiveCommands against
+     *   CM API and find the command ID from that response.
+     * - if listActiveCommands is empty (in case of the command finishes during the
+     *   deployClusterClientConfig timeout), use /cmf/commands/commandTable call (with the
+     *   response cookie of listActiveCommands call)
+     */
+    public BigDecimal deployClientConfigAndGetCommandId(
+            ClustersResourceApi api, Stack stack) throws CloudbreakException, ApiException {
+        ExecutorService executor = createExecutor();
+        Future<BigDecimal> future = executor.submit(() -> {
+            ApiCommand deployCommand = api.deployClientConfig(stack.getName());
+            return deployCommand.getId();
+        });
+        try {
+            return future.get(interruptTimeoutSeconds, TimeUnit.SECONDS);
+        } catch (TimeoutException timeoutException) {
+            LOGGER.debug("Deploy client config command took too much time. Start command ID "
+                    + "query by listing active commands");
+            ApiResponse<ApiCommandList> commandListResponse =
+                    api.listActiveCommandsWithHttpInfo(stack.getName(), null);
+            BigDecimal commandIdFromActiveCommands = getCommandIdFromActiveCommands(commandListResponse);
+            if (commandIdFromActiveCommands != null) {
+                return commandIdFromActiveCommands;
+            }
+            LOGGER.debug("The last deploy client config command could not be found  "
+                    + "by listing active commands. Trying {} call", CMF_COMMANDS_COMMAND_TABLE_URI_PATH);
+            BigDecimal commandIdFromCommandsTable = getCommandIdFromCommandsTable(api, commandListResponse.getHeaders());
+            if (commandIdFromCommandsTable != null) {
+                return commandIdFromCommandsTable;
+            } else {
+                throw new CloudbreakException(
+                        String.format("Obtaining Cloudera Manager Deploy config command ID was not possible neither by"
+                                + " listing Cloudera Manager commands nor using commandTable [stack: %s]", stack.getName()));
+            }
+        } catch (ExecutionException ee) {
+            throw new ApiException(ee.getCause());
+        } catch (InterruptedException e) {
+            throw new CloudbreakException(
+                    "Obtaining Cloudera Manager Deploy config command ID interrupted", e);
+        } finally {
+            executor.shutdown();
+        }
+    }
+
+    @VisibleForTesting
+    BigDecimal getCommandIdFromActiveCommands(ApiResponse<ApiCommandList> response) {
+        BigDecimal result = null;
+        if (response.getStatusCode() >= ERROR_CODES_FROM) {
+            return result;
+        }
+        ApiCommandList commandList = response.getData();
+        List<ApiCommand> commands = commandList.getItems();
+        long currentStartTime = Long.MIN_VALUE;
+        for (ApiCommand command : commands) {
+            if (command.getParent() == null && command.getStartTime() != null
+                    && DEPLOY_CLIENT_CONFIG_COMMAND_NAME.equals(command.getName())) {
+                long startTime = new DateTime(command.getStartTime())
+                        .toDate().toInstant().toEpochMilli();
+                if (startTime > currentStartTime) {
+                    LOGGER.debug("Found latest DeployClusterClientConfig command " +
+                            "with [command_id: {}, starTime: {}] by active commands]",
+                            command.getId(), command.getStartTime());
+                    currentStartTime = startTime;
+                    result = command.getId();
+                }
+            }
+        }
+        return result;
+    }
+
+    @VisibleForTesting
+    BigDecimal getCommandIdFromCommandsTable(ClustersResourceApi api,
+            Map<String, List<String>> headers) throws CloudbreakException, ApiException {
+        try {
+            OkHttpClient httpClient = api.getApiClient().getHttpClient();
+            Request request = createRequest(api.getApiClient(), headers);
+            Response response = httpClient
+                    .newCall(request)
+                    .execute();
+            if (response.code() >= ERROR_CODES_FROM) {
+                LOGGER.debug("{} request against Cloudera Manager API returned with status code: {}, response: {}",
+                        CMF_COMMANDS_COMMAND_TABLE_URI_PATH, response.toString(), response.code());
+                throw new CloudbreakException(
+                        String.format("%s request against CM API returned with status code: %d",
+                                CMF_COMMANDS_COMMAND_TABLE_URI_PATH, response.code()));
+            }
+            return getCommandIdFromCMResponse(response);
+        } catch (IOException e) {
+            throw new CloudbreakException(
+                    String.format("DeployClusterClientConfig - error during processing %s CM request",
+                            CMF_COMMANDS_COMMAND_TABLE_URI_PATH), e);
+        }
+    }
+
+    private BigDecimal getCommandIdFromCMResponse(Response response) throws IOException {
+        BigDecimal result = null;
+        try (ResponseBody responseBody = response.body()) {
+            String responseBodyStr = responseBody.string();
+            LOGGER.debug("Start processing commandTable response: {}", responseBodyStr);
+            List<CommandResource> commandList =
+                    new ObjectMapper().readValue(responseBodyStr, new TypeReference<>() {
+                    });
+            LOGGER.debug("Processing commandTable were successful.");
+            long currentStartTime = Long.MIN_VALUE;
+            for (CommandResource command : commandList) {
+                if (DEPLOY_CLIENT_CONFIG_COMMAND_NAME.equals(command.getName())) {
+                    LOGGER.debug("Found DeployClusterClientConfig command based on "
+                                    + "{} response [command_id: {}]", CMF_COMMANDS_COMMAND_TABLE_URI_PATH,
+                            command.getId());
+                    long startTime = command.getStart();
+                    if (startTime > currentStartTime) {
+                        LOGGER.debug("Found latest DeployClusterClientConfig command " +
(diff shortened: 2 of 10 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,240 @@
...
+    private static final int ERROR_CODES_FROM = 400;
+
+    private final Integer interruptTimeoutSeconds;
+
+    public ClouderaManagerDeployClientConfigProvider(
+            @Value("${cb.cm.client.commands.deployClientConfig.interrupt.timeout.seconds:}") Integer interruptTimeoutSeconds) {
+        this.interruptTimeoutSeconds = interruptTimeoutSeconds;
+    }
+
+    public BigDecimal calculateDeployClientConfigCommandId(
```
keyki, line 62: calculate is a bit misleading, I would rename it to something like deployClientConfigAndGetCommandId

oleewere: can be getCommandId as well as that's clear that is for deployClientConfig from the name (of course it does not tell it uses 3 different way to do it)

autoscale/src/main/resources/application.yml
```diff
@@ -64,6 +64,8 @@ cb:
       connect.timeout.seconds: 245
       read.timeout.seconds: 245
       write.timeout.seconds: 245
+      commands:
+        deployClientConfig.interrupt.timeout.seconds: 120
```
doktoric, line 68: I think we dont use camel cases here

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+                return commandIdFromActiveCommands;
+            }
+            LOGGER.debug("The last deploy client config command could not be found  "
+                    + "by listing active commands. Trying {} call", CMF_COMMANDS_COMMAND_TABLE_URI_PATH);
+            BigDecimal commandIdFromCommandsTable = getCommandIdFromCommandsTable(api, headers);
+            if (commandIdFromCommandsTable != null) {
+                return commandIdFromCommandsTable;
+            } else {
+                throw new CloudbreakException(
+                        String.format("Obtaining CM Deploy config command ID was not possible both by"
```
doktoric, line 99: Cm -> Cloudera Manager

doktoric: I think it would be nice to show a more user friendly message

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+                                + " listing CM commands or using commandTable [stack: %s]", stack.getName()));
+            }
+        } catch (ExecutionException ee) {
+            throw new ApiException(ee.getCause());
+        } catch (InterruptedException e) {
+            throw new CloudbreakException(
+                    "Obtaining CM Deploy config command ID interrupted", e);
```
doktoric, line 106: Cm -> Cloudera Manager

autoscale/src/main/resources/application.yml
oleewere: there are many camel case usages in the config, also i would not separate the command name

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+            return deployCommand.getId();
+        });
+        try {
+            return future.get(interruptTimeoutSeconds, TimeUnit.SECONDS);
+        } catch (TimeoutException timeoutException) {
+            LOGGER.debug("Deploy client config command took too much time. Start command ID "
+                    + "query by listing active commands");
+            ApiResponse<ApiCommandList> commandListResponse =
+                    api.listActiveCommandsWithHttpInfo(stack.getName(), null);
+            Map<String, List<String>> headers = commandListResponse.getHeaders();
```
pdarvasi, line 87: nit: this is needed only 7 lines below, if it returns before, then it was not needed at all

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+                    "Obtaining Cloudera Manager Deploy config command ID interrupted", e);
+        } finally {
+            executor.shutdown();
+        }
+    }
+
+    @VisibleForTesting
+    BigDecimal getCommandIdFromActiveCommands(ApiResponse<ApiCommandList> response) {
+        BigDecimal result = null;
+        if (response.getStatusCode() >= ERROR_CODES_FROM) {
```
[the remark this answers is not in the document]
oleewere: things e.g. redirect can be valid

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+        List<Pair> queryParams = new ArrayList<>();
+        Map<String, String> headerParams = new HashMap<>();
+        String[] localVarAccepts = new String[]{"application/json"};
+        String localVarAccept = apiClient.selectHeaderAccept(localVarAccepts);
+        if (localVarAccept != null) {
+            headerParams.put("Accept", localVarAccept);
+        }
+        String[] localVarContentTypes = new String[0];
+        String localVarContentType = apiClient.selectHeaderContentType(localVarContentTypes);
+        headerParams.put("Content-Type", localVarContentType);
```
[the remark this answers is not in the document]
oleewere: read the comment on the method, that's technically the same mechanism that is used in ApiClient. so im using the same initialization how it would be look like if ApiClient would have a valid api call for getting the recent commands

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+        long startTime = new DateTime()
+                .minusMinutes(INTERVAL_MINUTES)
+                .toDate().toInstant().toEpochMilli();
+        long endTime = new DateTime()
+                .plusMinutes(INTERVAL_MINUTES)
+                .toDate().toInstant().toEpochMilli();
+        queryParams.add(new Pair("startTime", Long.toString(startTime)));
+        queryParams.add(new Pair("endTime", Long.toString(endTime)));
+        if (headers.containsKey("Set-Cookie")) {
+            headerParams.put("Cookie", headers.get("Set-Cookie").get(0));
```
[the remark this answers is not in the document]
oleewere: it puts the response Set-Cookie header to the request Cookie header, so it's not 100% a copy as the header name is different. also im enhancing query params as well.. Of course  I can remove "custom" from the method name and rename it to something else

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,261 @@
...
+     * Extends request that is created based on the ApiClient endpoint implementations
+     * with default query params + Cookie header from listActiveCommands call.
+     * Example query: /cmf/commands/commandTable?limit=501&startTime=1234&endTime1235
+     */
+    private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair> queryParams, Map<String, String> headerParams) {
+        queryParams.add(new Pair("limit", "501"));
+        long startTime = new DateTime()
+                .minusMinutes(INTERVAL_MINUTES)
+                .toDate().toInstant().toEpochMilli();
+        long endTime = new DateTime()
```
bergerdenes, line 245: I think we do not need to go ahead of time, especially not 30 minutes

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,261 @@
...
+
+    /**
+     * Extends request that is created based on the ApiClient endpoint implementations
+     * with default query params + Cookie header from listActiveCommands call.
+     * Example query: /cmf/commands/commandTable?limit=501&startTime=1234&endTime1235
+     */
+    private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair> queryParams, Map<String, String> headerParams) {
+        queryParams.add(new Pair("limit", "501"));
+        long startTime = new DateTime()
+                .minusMinutes(INTERVAL_MINUTES)
```
bergerdenes, line 243: I feel 30 minutes is a bit much anyway.

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    private ClouderaManagerDeployClientConfigProvider underTest;
+
+    private Stack stack;
+
+    private final Map<String, List<String>> emptyHeaders = new HashMap<>();
+
+    @BeforeEach
+    public void setUp() {
+        underTest = Mockito.spy(new ClouderaManagerDeployClientConfigProvider(120));
```
bergerdenes, line 80: extract 120 as constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    @Test
+    public void testGetDeployClientConfigCommandIdNullAnswers()
+            throws ApiException, CloudbreakException, InterruptedException,
+            ExecutionException, TimeoutException {
+        // GIVEN
+        ApiResponse<ApiCommandList> response = createApiCommandListResponse();
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 115: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+            ExecutionException, TimeoutException, IOException {
+        // GIVEN
+        Map<String, List<String>> headers = new HashMap<>();
+        List<String> headerList = new ArrayList<>();
+        headerList.add("mycookie");
+        headers.put("Set-Cookie", headerList);
+        Request request = createDefaultRequest();
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 177: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+        verify(executorService, times(1)).shutdown();
+    }
+
+    @Test
+    public void testGetDeployClientConfigCommandIdByListCommandsApiException()
+            throws ApiException, InterruptedException, ExecutionException, TimeoutException {
+        // GIVEN
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 153: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+        verify(executorService, times(1)).shutdown();
+    }
+
+    @Test
+    public void testGetDeployClientConfigCommandIdByListCommands()
+            throws ApiException, CloudbreakException, InterruptedException, ExecutionException, TimeoutException {
+        // GIVEN
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 137: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    @Test
+    public void testGetDeployClientConfigCommandIdByCommandTableNotFound()
+            throws ApiException, InterruptedException, ExecutionException,
+            TimeoutException, IOException {
+        // GIVEN
+        Request request = createDefaultRequest();
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 228: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+        CloudbreakException exception = assertThrows(CloudbreakException.class,
+                () -> underTest.deployClientConfigAndGetCommandId(clustersResourceApi, stack));
+        // THEN
+        assertTrue(exception.getMessage().contains("error during processing"));
+        verify(executorService, times(1)).shutdown();
+    }
+
+    private ApiCommand createCommand() {
+        ApiCommand command = new ApiCommand();
+        command.setId(new BigDecimal(1L));
```
bergerdenes, line 274: use a more random value for easy differentiation

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    @Test
+    public void testGetDeployClientConfigCommandIdByCommandTableInvalidResponse()
+            throws ApiException, InterruptedException, ExecutionException,
+            TimeoutException, IOException {
+        // GIVEN
+        Request request = createDefaultRequest();
+        doReturn(executorService).when(underTest).createExecutor();
+        given(executorService.submit(any(Callable.class))).willReturn(future);
+        given(future.get(120, TimeUnit.SECONDS)).willThrow(new TimeoutException());
```
bergerdenes, line 254: use constant

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
```diff
@@ -0,0 +1,329 @@
...
+        given(clustersResourceApi.listActiveCommandsWithHttpInfo(anyString(), isNull()))
+                .willReturn(createApiCommandListResponse(404));
```
bergerdenes, line 256: consider extracting HTTP codes to constants

bergerdenes (changes requested):

cluster-cm/src/test/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProviderTest.java
oleewere: ok, make sense, changed this value to 7, also changed the responses from listActiveCommands to 4,5,6 (and keep the responses for /cmf/commands/commandTable to 1,2,3) - this way it's more clear which operation returned the value

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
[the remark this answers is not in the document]
oleewere: seems like in okhttp, follow. redirect is set by default so it should be handled

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -0,0 +1,253 @@
...
+    @VisibleForTesting
+    BigDecimal getCommandIdFromCommandsTable(ClustersResourceApi api,
+            Map<String, List<String>> headers) throws CloudbreakException, ApiException {
+        try {
+            OkHttpClient httpClient = api.getApiClient().getHttpClient();
+            Request request = createRequest(api.getApiClient(), headers);
+            Response response = httpClient
+                    .newCall(request)
+                    .execute();
+            if (response.code() >= ERROR_CODES_FROM) {
```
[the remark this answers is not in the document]
oleewere: resolving this as well as the above one

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
oleewere: seems a lot but as commandTable is the 3rd api call, in a bad case it's possible the commands before that can take 8mins, maybe even the date on the CB host machine can be shifted, a 1 hour range for querying safer than like 30 minutes (note that commandTable does not contain every subcommand only the larger ones)

pdarvasi (approved):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
oleewere: will check -1 value is valid or not (as i remember that is used by the UI)

bergerdenes (approved):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
oleewere: will do a follow up change regarding the intervals

PR #9857, opened 2021-01-25
CB-10740 LB doesn't failover when Knox instance is stopped

hreeve-cloudera:
Replaced the single port being used for load balancer config with TargetGroupPortPair.
This contains a port to direct traffic to, and a port to use for health checks.
Previous LB logic was using health the of nginx port to determine Knox instance health.
The new logic allows different ports to be used for traffic forwarding and health
checks. For the Knox LB, traffic will be forwarded to the nginx port, but health checks
will be done on the Knox port. So if Knox goes down, the LB will be able to detect it
and fail over to the second node.

Tested via unit tests and by creating a data lake and verifying LB configuration and
failover logic.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPair.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPair.java
@@ -1,0 +1,56 @@
+package com.sequenceiq.cloudbreak.cloud.model;
+
+import java.util.Objects;
+
+/**
+ * A joining of ports related to target groups. Target groups are groups of instances that a load balancer routes traffic
+ * to. The instances in the target group have a port on which they receive traffic from the load balancer (trafficPort),
+ * and a port on which the load balancer runs its health checks (healthCheckPort). If the health check responses from
+ * the healthCheckPort ever go into an unhealthy state, the load balancer will stop forwarding traffic to the trafficPort
+ * for the unhealthy instance.
+ */
+public class TargetGroupPortPair {
+
+    private final int trafficPort;
+
+    private final int healthCheckPort;
+
+    public TargetGroupPortPair(int trafficPort, int healthCheckPort) {
+        this.trafficPort = trafficPort;
+        this.healthCheckPort = healthCheckPort;
+    }
+
+    public Integer getTrafficPort() {
+        return trafficPort;
+    }
+
+    public Integer getHealthCheckPort() {
+        return healthCheckPort;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        TargetGroupPortPair portPair = (TargetGroupPortPair) o;
+        return trafficPort == portPair.trafficPort &&
+            healthCheckPort == portPair.healthCheckPort;
+    }
+
+    @Override
+    public int hashCode() {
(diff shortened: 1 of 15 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPair.java
```diff
@@ -0,0 +1,40 @@
+package com.sequenceiq.cloudbreak.cloud.model;
+
+import java.util.Objects;
+
+public class TargetGroupPortPair {
```
brycederriso, line 5: **praise:** Thank you for making this a real class!

And, I like that it's immutable!

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPair.java
```diff
@@ -0,0 +1,40 @@
...
+            return false;
+        }
+        TargetGroupPortPair that = (TargetGroupPortPair) o;
+        return trafficPort == that.trafficPort && healthCheckPort == that.healthCheckPort;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(trafficPort, healthCheckPort);
+    }
```
brycederriso, line 39: **suggestion:** Consider overriding `toString()`.

cloud-api/src/test/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPairTest.java
```diff
@@ -0,0 +1,42 @@
+package com.sequenceiq.cloudbreak.cloud.model;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.junit.Test;
+
+public class TargetGroupPortPairTest {
```
brycederriso, line 10: **praise:** Hooray for unit tests! Thank you.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPair.java
brycederriso, line 5: **suggestion:** I know that we're not huge fans of comments around these parts, but I think it would be good to mention in a Java doc where we're getting the target group from, and that this has to do with Cloud load balancers in particular.

brycederriso (approved):
Looks good to me, I have a suggestion for java docs, but I know that's controversial.

cloud-api/src/test/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPairTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private static final int TRAFFIC_PORT = 443;
+
+    private static final int HEALTH_PORT = 8443;
+
+    private TargetGroupPortPair underTest;
```
orngejaket, line 16: Is this used?

orngejaket (approved):

cloud-api/src/test/java/com/sequenceiq/cloudbreak/cloud/model/TargetGroupPortPairTest.java
hreeve-cloudera: Oh gosh. I'm so used to that unit test pattern I just created that object automatically. No, it's not. I'll remove it.

PR #9871, opened 2021-01-26
CB-10423 7.1.0 -> 7.2.6 upgrade fix: if the restartNeeded flag is set  but salt-bootstrap does not support this flag, then we change the salt master ip address to an another one, so minions will be restarted

sodre90:
Restart needed flag was introduced for the following use case:

In a repair scenario where salt master has the same IP address as before, the minions on other instances are not restarted by salt-bootstrap.
Salt minion does not try to communicate with master by itself for a long time (30mins or sthg like this).
So when we are waiting for the minions' keys on salt-master then timeout can happen if all the minions haven't tried to communicate with master during this time frame.

Timeout error message in this scenario: There are missing nodes from salt network response.
Solution was to introduce a restartNeeded flag in salt-bootstrap and we restart salt minions every time we do a salt-master replacement. (it used to be restarted only in case when salt-master IP address changed)

In case of older salt-bootsrap ( pre 0.13.4 ) the restartNeeded flag is not interpreted. So in an upgrade scenario (it has a repair flow in itself) where the upgrade starts from a 7.1.0 image, the minions are not restarted so the above mentioned timeout could happen.

It is a dirty fix but if the restartNeeded flag is set but salt-bootstrap does not support this flag, then we change the salt master ip address to an another one (1.2.3.4), so salt-minion will be restarted by older salt-bootstrap also.

The original IP will be set in the next iteration because restartNeeded flag will be false.

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
@@ -61,7 +61,7 @@ public class SaltBootstrap implements OrchestratorBootstrap {
         if (!targets.isEmpty()) {
             LOGGER.debug("Missing targets for SaltBootstrap: {}", targets);
 
-            SaltAction saltAction = createBootstrap(params.isRestartNeeded());
+            SaltAction saltAction = createBootstrap(params.isRestartNeededFlagSupported(), params.isRestartNeeded());
             GenericResponses responses = sc.action(saltAction);
 
             Set<Node> failedTargets = new HashSet<>();
@@ -113,7 +113,7 @@ public class SaltBootstrap implements OrchestratorBootstrap {
                 : new MinionAcceptor(sc, saltAction.getMinions(), new AcceptAllFpMatcher(), new DummyFingerprintCollector());
     }
 
-    private SaltAction createBootstrap(boolean restartNeeded) {
+    private SaltAction createBootstrap(boolean restartNeededFlagSupported, boolean restartNeeded) {
         SaltAction saltAction = new SaltAction(SaltActionType.RUN);
         if (params.getCloud() != null) {
             saltAction.setCloud(new Cloud(params.getCloud()));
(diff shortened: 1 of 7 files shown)


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
```diff
@@ -136,23 +136,45 @@ private SaltAction createBootstrap(boolean restartNeeded) {
...
+               itself) where the upgrade starts from a 7.1.0 image, the minions are not restarted so the above mentioned timeout could happen.
+
+               It is a dirty fix but if the restartNeeded flag is set but salt-bootstrap does not support this flag, then we change the
+               salt master ip address to an another one (1.2.3.4), so salt-minion will be restarted by older salt-bootstrap also.
+
+               The original IP will be set in the next iteration because restartNeeded flag will be false. */
+            minion.setServers(Collections.singletonList("1.2.3.4"));
+        } else {
+            minion.setServers(getGatewayPrivateIps());
+        }
```
lacikaaa, line 177: just for readability's sake: could you refactor it sth like:
```
minion.setServers(calculateServers(restartNeededFlagSupported, restartNeeded));
....
/*
all the comment, maybe in proper javadoc format
*/
private List<String> calculateServers(boolean restartNeededFlagSupported, boolean restartNeeded) {
	if (!restartNeededFlagSupported && restartNeeded) {
	....
	}
}
```

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
```diff
@@ -136,29 +136,61 @@ private SaltAction createBootstrap(boolean restartNeeded) {
...
+     *
+     *  The original IP will be set in the next iteration because restartNeeded flag will be false.
+     *
+     * @param restartNeededFlagSupported is restartNeeded flag supported by salt-bootstrap
+     * @param restartNeeded restart minions
+     * @return salt master(s) ip adress(es)
+     */
+    private List<String> calculateServerIps(boolean restartNeededFlagSupported, boolean restartNeeded) {
+        if (!restartNeededFlagSupported && restartNeeded) {
+            return Collections.singletonList("1.2.3.4");
```
keyki, line 188: Maybe we should use a loopback IP because 1.2.3.4 is a valid external IP. https://who.is/whois-ip/ip-address/1.2.3.4.

sodre90: true, I changed it to 127.0.0.1

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
sodre90: done

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
```diff
@@ -136,29 +136,61 @@ private SaltAction createBootstrap(boolean restartNeeded) {
...
+     *
+     *     Timeout error message in this scenario: There are missing nodes from salt network response.
+     *     Solution was to introduce a restartNeeded flag in salt-bootstrap and we restart salt minions every time we do a salt-master replacement.
+     *     (it used to be restarted only in case when salt-master IP address changed)
+     *
+     *  In case of older salt-bootsrap ( pre 0.13.4 ) the restartNeeded flag is not interpreted. So in an upgrade scenario (it has a repair flow in
+     *  itself) where the upgrade starts from a 7.1.0 image, the minions are not restarted so the above mentioned timeout could happen.
+     *
+     *  It is a dirty fix but if the restartNeeded flag is set but salt-bootstrap does not support this flag, then we change the
+     *  salt master ip address to an another one (1.2.3.4), so salt-minion will be restarted by older salt-bootstrap also.
```
lacikaaa, line 178: and the comment is not irght about the IP :D

lacikaaa (approved):

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/SaltBootstrap.java
sodre90: ahh.. thanks :D

PR #9876, opened 2021-01-26
CB-10849: Improve error message for manual FreeIPA Azure backup

jamisonbennett:
Improve the error message for manual FreeIPA Azure backup. When a
user runs the Azure FreeIPA backup inside a sudo environment which
doesn't replace the keyctl session then it fails due to access
insufficient access.

This was tested with a local deployment on Azure.


jasonwmcswain:
Looks good @jamisonbennett.  With these changes for Azure, do we need to make improvements for AWS and GCP?

jamisonbennett:
@jasonwmcswain Only Azure CLI uses the keyctl service to access the keyrings. No changes are needed for the other cloud providers.

lacikaaa (approved):

PR #9883, opened 2021-01-27
CB-10815 Refactor SaltStack sync all call to async

lacikaaa:
`saltutil.sync_all` call was a sync call which doesn't cause any issue if the size is under 200 nodes.
Over 200 nodes it takes too much time and caused timeout. Moving it to async solved the issue. It's handled similarly to highstate call.
Also measurements has been added to salt calls, so we will be able to check which salt calls are taking long time.

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/target/HostList.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/target/HostList.java
@@ -19,4 +19,11 @@ public class HostList implements Target<String> {
     public String getType() {
         return "list";
     }
+
+    @Override
+    public String toString() {
+        return "HostList{" +
+                "targets=" + targets == null ? "null" : String.join(", ", targets) +
+                '}';
+    }
 }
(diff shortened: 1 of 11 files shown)


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/target/HostList.java
```diff
@@ -19,4 +19,11 @@ public String getTarget() {
     public String getType() {
         return "list";
     }
+
+    @Override
+    public String toString() {
+        return "HostList{" +
+                "targets=" + targets +
```
biharitomi, line 26: Please join the name of the targets as the logged value could be only a reference based on the implementation of the Collection.

biharitomi (changes requested):
Otherwise, LGTM

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/target/HostList.java
lacikaaa: fixed, please check again

PR #9888, opened 2021-01-27
CB-10813. Fix: in case of cluster-proxy, the base context is missing …

oleewere:
…from the request url

see CB-10879 as well

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
@@ -29,7 +29,6 @@ import com.cloudera.api.swagger.model.ApiCommandList;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Splitter;
 import com.sequenceiq.cloudbreak.cm.model.CommandResource;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.service.CloudbreakException;
@@ -45,13 +44,17 @@ public class ClouderaManagerDeployClientConfigProvider {
     private static final Logger LOGGER =
             LoggerFactory.getLogger(ClouderaManagerDeployClientConfigProvider.class);
 
-    private static final String CMF_COMMANDS_COMMAND_TABLE_URI_PATH = "/cmf/commands/commandTable";
+    private static final String CMF_COMMANDS_ACTIVE_COMMAND_TABLE_URI_PATH = "/cmf/commands/activeCommandTable";
+
+    private static final String CMF_COMMANDS_RECENT_COMMAND_TABLE_URI_PATH = "/cmf/commands/commandTable";
 
     private static final String DEPLOY_CLIENT_CONFIG_COMMAND_NAME = "DeployClusterClientConfig";
 
     private static final int INTERVAL_MINUTES = 30;
 
     private static final int ERROR_CODES_FROM = 400;
+
+    private static final int INVALID_INDEX = -1;
 
     private final Integer interruptTimeoutSeconds;
 
@@ -67,8 +70,8 @@ public class ClouderaManagerDeployClientConfigProvider {
      * - if deployClusterClientConfig CM command times out, use listActiveCommands against
      *   CM API and find the command ID from that response.
      * - if listActiveCommands is empty (in case of the command finishes during the
-     *   deployClusterClientConfig timeout), use /cmf/commands/commandTable call (with the
-     *   response cookie of listActiveCommands call)
+     *   deployClusterClientConfig timeout), use /cmf/commands/activeCommandTable call, if that won't work either,
+     *   use /cmf/commands/commandTable call (both with the response cookie of listActiveCommands call)
      */
     public BigDecimal deployClientConfigAndGetCommandId(
             ClustersResourceApi api, Stack stack) throws CloudbreakException, ApiException {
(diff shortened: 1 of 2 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -258,4 +255,19 @@ private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair>
...
+
+    @VisibleForTesting
+    void addPathSegmentsFromDefaultUrl(
+            HttpUrl defaultHttpUrl, HttpUrl.Builder httpUrlBuilder) {
+        List<String> defaultPathSegments = defaultHttpUrl.pathSegments();
+        int versionNumberIndex = VERSION_NUMBER_INITIAL_VALUE;
+        for (int i = 0; i < defaultPathSegments.size(); i++) {
+            String pathSegment = defaultPathSegments.get(i);
+            if ("api".equals(pathSegment)) {
+                versionNumberIndex = i + 1;
```
lacikaaa, line 267: maybe you code use `indexOf` to find the `api`

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -258,4 +253,17 @@ private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair>
     ExecutorService createExecutor() {
         return Executors.newSingleThreadExecutor();
     }
+
+    @VisibleForTesting
+    void addPathSegmentsFromDefaultUrl(
+            HttpUrl defaultHttpUrl, HttpUrl.Builder httpUrlBuilder) {
+        List<String> defaultPathSegments = defaultHttpUrl.pathSegments();
+        int apiIndex = defaultPathSegments.indexOf("api");
```
lacikaaa, line 261: it might be an edge case or never should happen but you should handle if this returns `-1`, so it's not found. just in case. 
also some log would be nice here to see what's going on, what we modifying and what's the result

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -258,4 +253,17 @@ private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair>
     ExecutorService createExecutor() {
         return Executors.newSingleThreadExecutor();
     }
+
+    @VisibleForTesting
```
lacikaaa, line 257: a good sign this class already needs refactoring

oleewere: agreed, but i think i won't do that now:)

oleewere: the whole code should be removed anyway if the CM command will become async

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -258,4 +265,21 @@ private void enhanceRequestParams(Map<String, List<String>> headers, List<Pair>
...
+    void addPathSegmentsFromDefaultUrl(
+            HttpUrl defaultHttpUrl, HttpUrl.Builder httpUrlBuilder) {
+        List<String> defaultPathSegments = defaultHttpUrl.pathSegments();
+        int apiIndex = defaultPathSegments.indexOf("api");
+        int versionNumberIndex =  apiIndex == INVALID_INDEX ? INVALID_INDEX : apiIndex + 1;
+        for (int i = 0; i < defaultPathSegments.size(); i++) {
+            String pathSegment = defaultPathSegments.get(i);
+            if (i != apiIndex && i != versionNumberIndex) {
```
lacikaaa, line 277: although this code works well, and I think we might go with it, this solution is not self explaining when we look at the code.
like you have to understand that if apiIndex took the value of `-1` then `versionNumberIndex` would take the same value and this if condition would be always true. I think the code should be more self explaining

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDeployClientConfigProvider.java
```diff
@@ -135,30 +146,30 @@ BigDecimal getCommandIdFromActiveCommands(ApiResponse<ApiCommandList> response)
...
-                        CMF_COMMANDS_COMMAND_TABLE_URI_PATH, response.toString(), response.code());
+                        path, response.toString(), response.code());
                 throw new CloudbreakException(
                         String.format("%s request against CM API returned with status code: %d",
-                                CMF_COMMANDS_COMMAND_TABLE_URI_PATH, response.code()));
+                                path, response.code()));
             }
-            return getCommandIdFromCMResponse(response);
+            return getCommandIdFromCMResponse(response, path);
         } catch (IOException e) {
```
lacikaaa, line 165: if there is any exception is received which is not `IOException` during `CMF_COMMANDS_ACTIVE_COMMAND_TABLE_URI_PATH` call, we won't try it with `CMF_COMMANDS_RECENT_COMMAND_TABLE_URI_PATH`. Is it expected to behave like this?

oleewere: yes, if the call is broken (bad response etc.) I would expect the other one as well... also if let's say activeCommandTable fails, that does not mean that deployClusterClientConfig command will be found in commandTable

lacikaaa (approved):

lacikaaa:
please ask someone who has merge rights for 2.36 to merge this pr

PR #9889, opened 2021-01-27
CB-9882 Create StructuresSyncEvent quartz job to periodically scrape …

nadamdb:
…info about clusters.

The quartz job runs every hour (configurable) and creates a StructuredSyncEvent about clusters present even in stopped state so we can get information about both running an stopped clusters. At startup a job is registered to every currently present cluster and registers a job for every newly created cluster. A second commit will contain the protobuf logging of StructuredSyncEvents.


core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/job/StructuredSynchronizerJob.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    protected Object getMdcContextObject() {
+        return stackService.getById(getStackId());
+    }
+
+    @Override
+    protected void executeTracedJob(JobExecutionContext context) throws JobExecutionException {
+        if (flowLogService.isOtherFlowRunning(getStackId())) {
```
sodre90, line 52: Why do we need this? Shouldn't we collect data if something is happening with the cluster?

nadamdb: The goal is to periodically have info about clusters that are "not doing anything". Right now we can only have telemetry data in EDH about clusters that had something happened to them, like a stop/upgrade/start/etc event, but not about clusters that are "just running", so we cannot answer from this data that how many clusters are running/stopped/existing at a given moment.

sodre90: But shouldn't we collect clusters which stuck in a flow? Theoretically, it shouldn't happen, but it happened many times.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationActions.java
```diff
@@ -270,6 +275,7 @@ protected Selectable createRequest(ClusterViewContext context) {
             protected void doExecute(ClusterViewContext context, ClusterProxyGatewayRegistrationSuccess payload, Map<Object, Object> variables) {
                 clusterCreationService.clusterInstallationFinished(context.getStack());
                 jobService.schedule(context.getStackId(), StackJobAdapter.class);
+                syncJobService.schedule(context.getStackId(), StructuredSynchronizerJobAdapter.class);
```
sodre90, line 278: I can't see where is the unschedule for these kind of jobs. Shouldn't we unschedule when customer terminte a cluster?

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/job/StructuredSynchronizerJob.java
```diff
@@ -0,0 +1,76 @@
...
+            LOGGER.debug("StructuredSynchronizerJob cannot run, because flow is running for stack: {}", getStackId());
+            return;
+        }
+        try {
+            StructuredSyncEvent structuredEvent = structuredSyncEventFactory.createStructuredSyncEvent(getStackId());
+            LOGGER.debug(structuredEvent.getStatus());
```
sodre90, line 58: Status is not a big help in debugging, it would be better to log stack name, status etc.  Some more words could be helpful for debugging also, like "Collected data to send: {}", structuredEvent

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/StructuredSyncEventFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    private NodeConfig nodeConfig;
+
+    @Value("${info.app.version:}")
+    private String serviceVersion;
+
+    public StructuredSyncEvent createStructuredSyncEvent(Long resourceId) {
+        return createStructuredSyncEvent(resourceId, null);
+    }
+
+    public StructuredSyncEvent createStructuredSyncEvent(Long resourceId, Exception exception) {
```
sodre90, line 46: are we use the method with exception parameter? if not then we can remove it and then we can remove the if statement below

core/src/main/java/com/sequenceiq/cloudbreak/structuredevent/job/StructuredSynchronizerJobAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.quartz.Job;
+import org.springframework.context.ApplicationContext;
+import org.springframework.data.repository.CrudRepository;
+
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.quartz.model.JobResourceAdapter;
+import com.sequenceiq.cloudbreak.repository.StackRepository;
+
+public class StructuredSynchronizerJobAdapter extends JobResourceAdapter<Stack> {
```
bbihari, line 11: Do we need the full Stack here? Maybe using StackView or similar would consume less resources.
Btw, have we got any estimate about the load this new scheduled task would generate?

bbihari (approved):

PR #9890, opened 2021-01-28
CB-10839 Cannot open hue UI when Public Endpoint Access Gateway is enabled

hreeve-cloudera:
Adds the load balancer FQDN to the list of Hue Knox proxies if the load balancer exists.
Tested with unit tests.


enis (approved):
lgtm.

kmanamcheri (approved):
lgtm.

lacikaaa:
the fix version in jira is 2.37
Are you sure it's needed only in master?

keyki:
> the fix version in jira is 2.37
> Are you sure it's needed only in master?

There is an ongoing slack conversation about whether we should put in the 2.37 release. It depends on a particular customer since it's a TP feature. However, since we're doing a hotfix again in 2.36 there is still time to merge to 2.37 if we want to.

keyki (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -105,19 +105,21 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
-                .findAny()
-                .orElse(loadBalancers.iterator().next());
-            if (StringUtils.isNotEmpty(preferredLB.getFqdn())) {
-                return preferredLB.getFqdn();
-            } else {
-                return loadBalancers.stream()
-                    .map(LoadBalancer::getFqdn)
-                    .filter(StringUtils::isNotEmpty)
-                    .findFirst().orElse(null);
+        if (stackId != null) {
```
lacikaaa, line 108: when can this be null? at first glance this doesn't seem like a valid scenario

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -105,19 +105,21 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
+            Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stackId);
+            if (!loadBalancers.isEmpty()) {
+                LoadBalancer preferredLB = loadBalancers.stream()
+                    .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
+                    .findAny()
```
lacikaaa, line 113: why don't you filter out LBs without FQDN here and you won't need the next if-else

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -105,19 +105,21 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
-            if (StringUtils.isNotEmpty(preferredLB.getFqdn())) {
-                return preferredLB.getFqdn();
-            } else {
-                return loadBalancers.stream()
-                    .map(LoadBalancer::getFqdn)
-                    .filter(StringUtils::isNotEmpty)
-                    .findFirst().orElse(null);
+        if (stackId != null) {
+            Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stackId);
+            if (!loadBalancers.isEmpty()) {
```
lacikaaa, line 110: this shouldn't be necessary as you can initialize a stream on an empty set

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: I was hitting an NPE in a unit test and added this check while I was tracking down the source. It was something else, but I left it in because I figured it couldn't hurt. But if it's not a valid scenario I'll put it back.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -105,23 +105,15 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
-                    .findFirst().orElse(null);
-            }
-        }
+        Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stackId).stream()
+            .filter(lb -> StringUtils.isNotEmpty(lb.getFqdn()))
+            .collect(Collectors.toSet());
+        LoadBalancer loadBalancer = loadBalancers.stream()
+            .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
+            .findAny()
+            .orElse(loadBalancers.isEmpty() ? null : loadBalancers.iterator().next());
```
lacikaaa, line 114: you might use `orElseGet` in this case, like:
```
        LoadBalancer loadBalancer = loadBalancers.stream()
            .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
            .findAny()
            .orElseGet(() -> loadBalancers.stream().findAny().orElse(null));
```
it would prevent java from evaluating the code even when it's not necessary

lacikaaa (approved):

PR #9891, opened 2021-01-28
CB-10656: Limit which type of Data Hubs can be upgraded

tiborpopovics:
In this commit, I've introduced a new condition to limit which type of DataHubs can be upgraded.

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -446,7 +446,6 @@ cb:
   custom.user.data: |
     touch /tmp/custom-user-data
     date >> /tmp/custom-user-data
-
   mail:
     smtp:
       auth: true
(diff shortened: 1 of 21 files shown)


core/src/main/resources/application.yml
```diff
@@ -446,6 +446,7 @@ cb:
   custom.user.data: |
     touch /tmp/custom-user-data
     date >> /tmp/custom-user-data
+  upgrade.enablecustomtemplate: true
```
keyki, line 449: Can you please remove this option anyways? Let's allow all custom blueprints. I'm not sure what the end behavior will be with custom blueprints, but I don't think a deployment wide option will help.

keyki (changes requested):

keyki (approved):

PR #9901, opened 2021-01-28
CB-10867 While we try to validate the Azure credential, we get a Runt…

topolyai5:
…imeException. However, the AuthenticatedException can be found in the tree of causes. Now we search recursively the cause exception. Rules:

* if cause is null, get the closest mapper for the exception
if cannot find a mapper or the mapper is the default, we search the next cause mapper
any other case, use the found mapper

With this test, we can test the real mapper finder

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
+++ b/environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
@@ -4,6 +4,8 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.Providers;
 
+import org.apache.commons.lang3.tuple.Pair;
+
 public abstract class SearchCauseExceptionMapper<T extends Throwable> extends BaseExceptionMapper<T> {
 
     @Context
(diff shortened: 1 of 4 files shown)


doktoric:
2.37 pls

topolyai5:
> 2.37 pls

done

environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
```diff
@@ -12,13 +14,29 @@
...
             return defaultResponse;
         }
-        BaseExceptionMapper exceptionMapper = (BaseExceptionMapper<? extends Throwable>) providers.getExceptionMapper(exception.getCause().getClass());
-        if (exceptionMapper == null) {
+        Pair<BaseExceptionMapper, ? extends Throwable> pair = searchRecursively(exception);
+        if (pair.getKey() == null || pair.getKey() == this) {
             return defaultResponse;
         }
-        return exceptionMapper.getResponseStatus(exception.getCause());
+        return pair.getKey().getResponseStatus(pair.getValue());
```
biharitomi, line 24: Can we use one return statement here? That would me more readable.

environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
```diff
@@ -12,13 +14,29 @@
...
+    private Pair<BaseExceptionMapper, Throwable> searchRecursively(Throwable exception) {
+        if (exception.getCause() == null) {
+            BaseExceptionMapper exceptionMapper = getExceptionMapper(exception);
+            return Pair.of(exceptionMapper, exception);
+        }
+        BaseExceptionMapper<? extends Throwable> exceptionMapper = getExceptionMapper(exception.getCause());
+        if (exceptionMapper == null || exceptionMapper instanceof DefaultExceptionMapper) {
+            return searchRecursively(exception.getCause());
+        }
+        return Pair.of(exceptionMapper, exception.getCause());
```
biharitomi, line 36: The same for this method too.

environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
```diff
@@ -12,13 +14,29 @@
     @Override
     Response.Status getResponseStatus(T exception) {
         Response.Status defaultResponse = Response.Status.BAD_REQUEST;
-        if (exception.getCause() != null && Exception.class == exception.getCause().getClass()) {
+        if (exception == null || exception.getCause() != null && Exception.class == exception.getCause().getClass()) {
```
biharitomi, line 17: It will return with default response, but doesn't this hide root causes where the root cause is encapsulated in an "Exception" class??

biharitomi (changes requested):

environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
topolyai5: Do you want to merge some return statements or I should hold the result in a variable and return with them in one place?

keyki:
Is this a regression? If not I wouldn't put in the 2.37 branch for now.

environment/src/main/java/com/sequenceiq/environment/exception/mapper/SearchCauseExceptionMapper.java
biharitomi: one global variable for the return value, of course the recursive call could remain as it isnow

biharitomi (approved):
LGTM

PR #9903, opened 2021-01-28
CB-10875 Redbeams termination flow fix

schfeca75:
Handle the case when the dbstack is already deleted from the database, but the termination flow has not finished.
After redbeams process restart the flow will continue its execution:
* The termination flow will handle when dbstack is not found in database during flowcontext creation.
* The termination finsished and failed action will handle if dbstack is missing from the context.

### DIFF
--- a/redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
+++ b/redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
@@ -25,6 +25,7 @@ import com.sequenceiq.redbeams.converter.cloud.CredentialToCloudCredentialConver
 import com.sequenceiq.redbeams.converter.spi.DBStackToDatabaseStackConverter;
 import com.sequenceiq.redbeams.domain.stack.DBStack;
 import com.sequenceiq.redbeams.dto.Credential;
+import com.sequenceiq.redbeams.exception.RedbeamsException;
 import com.sequenceiq.redbeams.flow.redbeams.common.RedbeamsContext;
 import com.sequenceiq.redbeams.flow.redbeams.common.RedbeamsEvent;
 import com.sequenceiq.redbeams.flow.redbeams.termination.event.terminate.TerminateDatabaseServerFailed;
@@ -48,8 +49,15 @@ public abstract class AbstractRedbeamsTerminationAction<P extends RedbeamsEvent>
     @Inject
     private DBStackToDatabaseStackConverter databaseStackConverter;
 
+    private boolean failOnMissingDBStack;
+
     protected AbstractRedbeamsTerminationAction(Class<P> payloadClass) {
+        this(payloadClass, true);
+    }
+
+    protected AbstractRedbeamsTerminationAction(Class<P> payloadClass, boolean failOnMissingDBStack) {
         super(payloadClass);
+        this.failOnMissingDBStack = failOnMissingDBStack;
     }
 
     @PostConstruct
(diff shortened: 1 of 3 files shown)


redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/action/RedbeamsTerminationActions.java
```diff
@@ -113,18 +113,22 @@ protected void prepareExecution(DeregisterDatabaseServerSuccess payload, Map<Obj
...
             @Override
-            protected Selectable createRequest(RedbeamsContext context) {
+            protected void doExecute(RedbeamsContext context, DeregisterDatabaseServerSuccess payload, Map<Object, Object> variables) throws Exception {
                 // Delete the DB stack here instead of deregistration so that we can keep track of its status
                 // through the termination
-                metricService.incrementMetricCounter(MetricType.DB_TERMINATION_FINISHED, Optional.of(context.getDBStack()));
-                dbStackService.delete(context.getDBStack().getId());
-
-                return new RedbeamsEvent(REDBEAMS_TERMINATION_FINISHED_EVENT.name(), 0L);
+                DBStack dbStack = context.getDBStack();
```
gergopapi2, line 119: I would make an optional from the DBStack right here and use them in the following two statements.

redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
```diff
@@ -64,29 +64,30 @@ protected void doExecute(RedbeamsContext context, P payload, Map<Object, Object>
...
+                    dbStack.getPlatformVariant(), location, userName, accountId);
             try {
                 Credential credential = credentialService.getCredentialByEnvCrn(dbStack.getEnvironmentId());
                 cloudCredential = credentialConverter.convert(credential);
             } catch (Exception ex) {
                 LOGGER.warn("Could not detect credential for environment: {}", dbStack.getEnvironmentId());
             }
             databaseStack = databaseStackConverter.convert(dbStack);
+        } else {
+            LOGGER.warn("DBStack for {} id is not found in the database, it is possible only if the redbeams process was killed during the execution" +
```
gergopapi2, line 89: I would reformulate, because currently we think it is only possible under one condition but we might prove wrong. 
It is possible only => it seems to be only possible.

redbeams/src/main/java/com/sequenceiq/redbeams/flow/redbeams/termination/AbstractRedbeamsTerminationAction.java
```diff
@@ -64,29 +64,30 @@ protected void doExecute(RedbeamsContext context, P payload, Map<Object, Object>
...
         DBStack dbStack = null;
-        dbStack = optionalDBStack.get();
-        MDCBuilder.buildMdcContext(dbStack);
-        Location location = location(region(dbStack.getRegion()), availabilityZone(dbStack.getAvailabilityZone()));
-        String userName = dbStack.getOwnerCrn().getUserId();
-        String accountId = dbStack.getOwnerCrn().getAccountId();
-        cloudContext = new CloudContext(dbStack.getId(), dbStack.getName(), dbStack.getResourceCrn().toString(), dbStack.getCloudPlatform(),
-                dbStack.getPlatformVariant(), location, userName, accountId);
-
         if (optionalDBStack.isPresent()) {
```
gergopapi2, line 73: In other termination flow steps if DBStack is not present then it will cause NPE. I think it might make sense to propagate the Optional to those places as well and fail with a custom exception when there is no DBStack.

gergopapi2 (changes requested):

keyki (approved):

PR #9904, opened 2021-01-28
CB-10817 - Azure datahub scaling from 100 to 200 nodes fails with wei…

pdarvasi:
…rd error

This PR solves 2 issues, for the use case of recovering from failed upscale:

- CloudInstance inherits cloudResource.getParameters() containing RG name in single RG mode. This solves that the VM-s were fetched from the default, incorrect RG

` INFO  c.m.a.m.c.VirtualMachines listByResourceGroup] <-- 404  https://management.azure.com/subscriptions/3ddda1c7-d1f5-4e7b-ac81-0523f483b3b3/resourceGroups/gk1prod-ha98430/providers/Microsoft.Compute/virtualMachines?api-version=2020-06-01 (136 ms, 107-byte body)`

- We do not query the VM for a given volume set if does not have the VM's name stored in cloudResource.getInstanceId(). This prevents like the following. 
I was not able to reproduce the cause why instanceId is missing for these volume sets, but those should be deleted independently

` doRollbackAndDecreaseNodeCount:148 DEBUG c.s.c.c.t.c.CloudFailureHandler  Resource can not be deleted. Reason: java.lang.IllegalArgumentException: Parameter vmName is required and cannot be null.`

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
@@ -33,7 +33,6 @@ import com.microsoft.azure.management.compute.Disk;
 import com.microsoft.azure.management.compute.DiskSkuTypes;
 import com.microsoft.azure.management.compute.DiskStorageAccountTypes;
 import com.microsoft.azure.management.compute.OperatingSystemStateTypes;
-import com.microsoft.azure.management.compute.PowerState;
 import com.microsoft.azure.management.compute.VirtualMachine;
 import com.microsoft.azure.management.compute.VirtualMachineCustomImage;
 import com.microsoft.azure.management.compute.VirtualMachineDataDisk;
@@ -468,7 +467,7 @@ public class AzureClient {
         return handleAuthException(() -> azure.virtualMachines().listByResourceGroup(resourceGroup));
     }
 
-    public VirtualMachine getVirtualMachine(String resourceGroup, String vmName) {
+    public VirtualMachine getVirtualMachineByResourceGroup(String resourceGroup, String vmName) {
         return handleAuthException(() -> azure.virtualMachines().getByResourceGroup(resourceGroup, vmName));
     }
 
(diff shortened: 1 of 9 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/upscale/AzureUpscaleService.java
```diff
@@ -111,6 +110,14 @@
         }
     }
 
+    private List<CloudInstance> getNewInstances(List<CloudResource> newInstances) {
```
doktoric, line 113: It seems this part was not unit tested. can you please add a unit test for this method?

pdarvasi: I agree that it would be useful but I'd prefer doing it in a follow-up as I hardly modified the behaviour and it might cover a Linkedin error. Do you agree?

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -160,6 +159,7 @@ public ResourceGroup createResourceGroup(String name, String region, Map<String,
     public Deployment createTemplateDeployment(String resourceGroupName, String deploymentName, String templateContent, String parameterContent) {
         return handleAuthException(() -> {
             try {
+                LOGGER.debug("");
```
bergerdenes, line 162: I guess you meant to actually print out something

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/resource/AzureVolumeResourceBuilder.java
```diff
@@ -220,17 +221,21 @@ public CloudResource delete(AzureContext context, AuthenticatedContext auth, Clo
 
         for (CloudResourceStatus cloudResourceStatus : removableDisks) {
             CloudResource cloudResource = cloudResourceStatus.getCloudResource();
-            String instanceName = cloudResource.getInstanceId();
             List<String> volumeIds = getVolumeSetAttributes(cloudResource).getVolumes().stream()
                     .map(VolumeSetAttributes.Volume::getId)
                     .collect(Collectors.toList());
             for (String volumeId : volumeIds) {
                 if (ResourceStatus.ATTACHED.equals(cloudResourceStatus.getStatus())) {
+                    String instanceName = cloudResource.getInstanceId();
```
bergerdenes, line 229: This method is too complex. Please create a follow-up task to refactor.

bergerdenes (changes requested):
2 minor points

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/resource/AzureVolumeResourceBuilder.java
pdarvasi: Thx, pls. re-check

bergerdenes (approved):
LGTM

PR #9906, opened 2021-01-29
CB-10921. Make sure CB using the proper deployClusterClientConfig com…

oleewere:
…mand ID during upscale/downscale.

details:
- precheck commands before the first deployClusterClientConfig
- use polling for getting the newest deployClusterClientConfig
- use the prechecked command id against the newest command id - to make sure CB wont use the first one instead of the newest one
- do the polling against only the successful deployClusterClientConfig commands, as for failures, CM starts a new command by itself, so CB can continously check against the the commands


oleewere:
closing it for now, opening to 2.37

PR #9910, opened 2021-01-29
CB-10844 CM launched on Gateway in Flow Management Light/Heavy duty t…

vbrodetskyi:
…emplates. Changed 'gateway' node name to 'management', according to customer request.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.8/cdp-flow-management-small.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.8/cdp-flow-management-small.bp
@@ -290,7 +290,7 @@
         ]
       },
       {
-        "refName": "gateway",
+        "refName": "management",
         "cardinality": 1,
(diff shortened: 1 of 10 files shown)


lajosrodek (approved):
LGTM

vbrodetskyi:
@lajosrodek @doktoric should i also merge this change to some other branches or master will be enough?

lajosrodek:
@vbrodetskyi `master` (v2.38 at the moment) is sufficient if you are fine with having it only from CB 2.38 onwards. Otherwise, it would have to be pushed directly into some release branch and then rolled out as a Prod Hotfix. 😃 

Holding back with merging until someone from CB can also approve.

vbrodetskyi:
@lajosrodek thank you for clarification, according to fix version in jira https://jira.cloudera.com/browse/CB-10844, it should be i 2.38, so looks like master branch should be enough.

doktoric (approved):

PR #9912, opened 2021-01-29
CB-10911 E2E test cases fails, created resources are persisted for cleanup job in case of timeout

biharitomi:
CB-10911  - In case of E2E test case timeout the created resources are persisted into a JSON file by a new TestNG listener. Increase the timeout for Environment stop/start test case. In case of timeout the test case should be marked failed.
Also some log enhancements has been added to the changes.


aszegedi (approved):

keyki:
Is this issue valid on 2.37 as well? If it is we can merge it to 2.37, because it does not change the production code.

aszegedi:
Yes. 2.37 is also affected by this `feature`.

biharitomi:
Retest this please.

Issue #499, opened 2015-03-06
Documentation for configuration of cloudbreak for OpenStack


saurabhmishra:

doktoric:
Hi @saurabhmishra,
You can deploy our full stack with this repository https://github.com/sequenceiq/cloudbreak-deployment . You can find some useful information in the Readme files.
If you have any question then feel free to ask.
Br, 
@doktoric

PR #9913, opened 2021-01-29
CB-10214: Create mount targets of the EFS instance

LinaAtAustin:
Code:
Add code to create mount target with its security group.
Add more unit tests

Test:
I have manually tested by enabling the entitlement CDP_DATA_LAKE_AWS_EFS and add EFS parameters in Cluster request. In local deployment, the Data Lake was created successfully, and in AWS UI, the mount target security group and mount target got created correctly.


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverter.java
```diff
@@ -103,6 +103,11 @@ public ClusterV4Response convert(Cluster source) {
         return clusterResponse;
     }
 
+    // used for test
+    public void setCloudStorageConverter(CloudStorageConverter cloudStorageConverter) {
```
topolyai5, line 107: in this case, please use with package protected

core/src/test/java/com/sequenceiq/cloudbreak/converter/stack/cluster/ClusterToClusterV4ResponseConverterTest.java
```diff
@@ -44,13 +54,17 @@
...
+import com.sequenceiq.common.model.FileSystemType;
 
 @RunWith(MockitoJUnitRunner.class)
 public class ClusterToClusterV4ResponseConverterTest extends AbstractEntityConverterTest<Cluster> {
 
     @InjectMocks
     private ClusterToClusterV4ResponseConverter underTest;
 
+    @InjectMocks
+    private CloudStorageConverter cloudStorageConverter = new CloudStorageConverter();
```
topolyai5, line 66: If I understand correctly, do you want to use the real functionality for this class in the tests?

topolyai5 (changes requested):
Could you please extend the commit message?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverter.java
LinaAtAustin: done

core/src/test/java/com/sequenceiq/cloudbreak/converter/stack/cluster/ClusterToClusterV4ResponseConverterTest.java
LinaAtAustin: The test of real implementation of CloudStorageConverter is in CloudStorageConverterTest. I don't need to use the real functionality of CloudStorageConverter here. So I have updated the test to mock the implementation of CloudStorageConverter.

LinaAtAustin:
"extend the commit message" added more info in commit message.

topolyai5 (approved):
LGTM

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
```diff
@@ -433,7 +433,7 @@ private S3Identity identityRequestToS3(StorageIdentityBase storageIdentityReques
 
     private EfsIdentity identityRequestToEfs(StorageIdentityBase storageIdentityRequest) {
         EfsIdentity efsIdentity = new EfsIdentity();
-        efsIdentity.setInstanceProfile(storageIdentityRequest.getS3().getInstanceProfile());
+        efsIdentity.setInstanceProfile(storageIdentityRequest.getEfs().getInstanceProfile());
```
lajosrodek, line 436: Out of curiosity, what is the purpose of this EC2 instance profile? Where is it used?

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -495,6 +495,25 @@
...
+        <#else>
+          "VpcId" : { "Ref" : "VPC" },
+        </#if>
+        "GroupDescription": "Security group for EFS mount target",
+        "SecurityGroupIngress": [
+          {
+            "IpProtocol": "tcp",
+            "FromPort": "2049",
+            "ToPort": "2049",
+            "CidrIp": "0.0.0.0/0"
```
lajosrodek, line 512: I suppose this SG is to control which clients may be able to connect to the mount target (https://docs.aws.amazon.com/efs/latest/ug/accessing-fs-create-security-groups.html). If so, why not restrict it to the CIDR of the subnet?

lajosrodek (changes requested):
Looks okay, I had just a few questions.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
LinaAtAustin: This will be used for access control of the EFS.

lajosrodek: I surely know what an instance profile is generally meant to provide. 😃 So, let me rephrase: FreeIPA, DL and DH EC2 instances always have either of two EC2 instance profiles attached (exactly one, as it is not possible to attach multiple ones): the Log profile/role or the IDBroker Assumer profile/role. So I am wondering where the access rights for EFS come into play. Are you planning to extend these roles with EFS permissions (presumably the Log one, as I guess EFS will not be attached to IDBroker nodes)?

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: This is AWS document regarding mount target:
A mount target provides an IP address for an NFSv4 endpoint at which you can mount an Amazon EFS file system. You mount your file system using its Domain Name Service (DNS) name, which resolves to the IP address of the EFS mount target in the same Availability Zone as your EC2 instance. You can create one mount target in each Availability Zone in an AWS Region. If there are multiple subnets in an Availability Zone in your VPC, you create a mount target in one of the subnets. Then all EC2 instances in that Availability Zone share that mount target. https://docs.aws.amazon.com/efs/latest/ug/how-it-works.html

I have talked with Haley, and she mentioned that we only support a single subnet within an availability zone. So I will make the change to use the existing subnet IP ranger to be more secure. Later on, if we support multiple subnet within an availability zone, we need to expend this range.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
lajosrodek: So access to EFS file system instances seems to be [governed by file system policies](https://docs.aws.amazon.com/efs/latest/ug/iam-access-control-nfs-efs.html). I am not sure if any additional permissions are needed for the EC2 instance itself. Have you possibly verified this?

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: I ended up using the VPC subnets as ingress. So it will allow access within the VPC. In this way, we don't need to change later, and still secure. For each sub-folder at the EFS, we have additional mechanism, such as access point or security group, to control the fine-grained access.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -495,6 +495,27 @@
...
+    "MountTargetSecurityGroup": {
+      "Type": "AWS::EC2::SecurityGroup",
+      "Properties": {
+        <#if existingVPC>
+          "VpcId" : { "Ref" : "VPCId" },
+        <#else>
+          "VpcId" : { "Ref" : "VPC" },
+        </#if>
+        "GroupDescription": "Security group for EFS mount target",
+        "SecurityGroupIngress": [
```
doktoric, line 507: this can easily hit customer issues when we havea limited policy

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
doktoric: this is definitely an additional thing to support this solution. Even the current roles are hard to understand and it seems now we need one more.

LinaAtAustin: I am not planning to attach EFS to IDBroker nodes. 
Since EC2 instance can only attach one instance profile, I will need to extend those roles with EFS permissions.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: The allowed IP address should cover all IP addresses in VPC, and right now, only EC2 from the Data Lake should access them. Do you see what kind of access could fail?

What is your suggestion?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
LinaAtAustin: I have discussed with Lajos in slack. I will handle this in another jira. So close this discussion for now.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: From Lajos "the "pre-existing SG" case is what is problematic here; we shall not attempt creating a new SG for the EFS mount target if there are existing SGs choosen by the customer."

I discussed with Lajos offline and realize Richard concerns about creating security group if user provides security group. Updated the CloudFormation template.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +536,28 @@
         </#if>
         "ThroughputMode": "${efsFileSystem.throughputMode}"
       }
+    },
+    "MountTarget": {
```
doktoric, line 540: should we put here enableEfs as well ?

LinaAtAustin: Yes. It should be inside "enableEfs", and it is already inside. See line #499 "<#if enableEfs>", and line #562 "</#if>". If it is not inside, the integration tests won't pass.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
lajosrodek: As discussed offline, it is unclear what kind of permissions are needed for EC2 instances in order to utilize EFS mounts. IDBroker and RAZ are out of the game, as the mounting logic must not depend on either Hadoop or any specific user. More elaboration and experimentation is needed to clarify these, but that may be covered in a subsequent PR.

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverterTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
```
lajosrodek, line 11: Please do not use JUnit 4 for new tests; JUnit 5 is preferable.

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
```diff
@@ -1,4 +1,4 @@
-package com.sequenceiq.cloudbreak.converter.stack.cluster;
+package com.sequenceiq.cloudbreak.converter.v4.stacks.cluster;
 
 import static org.junit.Assert.assertEquals;
```
lajosrodek, line 3: Could you please migrate the class to JUnit 5?

lajosrodek (changes requested):

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -469,6 +469,9 @@
 		  </#list>
 		  <#if vpcSubnet?has_content && vpcSubnet?size != 0>
             <#list vpcSubnet as s>
+                <#if enableEfs>
```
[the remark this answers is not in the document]
LinaAtAustin: I have removed the line #472. 

I have manually created EFS and then mount EC2 instance there. Actually, it is better to have the same SG used by EC2 as the SG associated with the EFS mount target to avoid any permission issue or mount problem.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +536,28 @@
...
+        "FileSystemId": { "Ref": "ElasticFileSystem" },
+        <#if !existingSubnet>
+            "SubnetId": { "Ref": "PublicSubnet" },
+        <#else>
+            "SubnetId": { "Ref": "SubnetId" },
+        </#if>
+        <#if instanceGroups?size != 0>
+            "SecurityGroups" : [
+            <#list instanceGroups as group>
+              <#list group.cloudSecurityIds as cloudSecurityId>
```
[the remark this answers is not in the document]
LinaAtAustin: done.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +536,28 @@
...
+    "MountTarget": {
+      "Type": "AWS::EFS::MountTarget",
+      "Properties": {
+        "FileSystemId": { "Ref": "ElasticFileSystem" },
+        <#if !existingSubnet>
+            "SubnetId": { "Ref": "PublicSubnet" },
+        <#else>
+            "SubnetId": { "Ref": "SubnetId" },
+        </#if>
+        <#if instanceGroups?size != 0>
```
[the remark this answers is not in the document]
LinaAtAustin: done

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +536,28 @@
...
+            "SecurityGroups" : [
+            <#list instanceGroups as group>
+              <#list group.cloudSecurityIds as cloudSecurityId>
+                "${cloudSecurityId}"<#if cloudSecurityId_has_next || group_has_next>,</#if>
+              </#list>
```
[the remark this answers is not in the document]
LinaAtAustin: done

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
```diff
@@ -187,14 +187,14 @@ public AwsEfsParameters fileSystemToEfsParameters(FileSystem fileSystem) {
         }
 
         efsParameters.setName(fileSystem.getName());
```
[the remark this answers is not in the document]
LinaAtAustin: removed

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
```diff
@@ -178,7 +178,7 @@ public AwsEfsParameters fileSystemToEfsParameters(FileSystem fileSystem) {
...
 
-        Map<String, String> configurations = null;
+        Map<String, Object> configurations = null;
         try {
             configurations = JsonUtil.readValue(fileSystem.getConfigurations().getValue(), Map.class);
         } catch (IOException ex) {
```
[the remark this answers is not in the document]
LinaAtAustin: In other places when reading json failed, it throws exception. Since EFS is not essential for stack, I decide to log error message, and return null. So do not create EFS instance. When we move some services to EFS, I will change it to fail the stack creation.

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverterTest.java
LinaAtAustin: done

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +536,28 @@
...
+            </#list>
+            ]
+        <#else>
+            "SecurityGroups" : [ { "Ref" : "ClusterNodeSecurityGroup${group.groupName?replace('_', '')}" } ]
+        </#if>
```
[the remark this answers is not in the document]
LinaAtAustin: fixed

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
LinaAtAustin: done

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +533,28 @@
...
+    "MountTarget": {
+      "Type": "AWS::EFS::MountTarget",
+      "Properties": {
+        "FileSystemId": { "Ref": "ElasticFileSystem" },
+        <#if !existingSubnet>
+            "SubnetId": { "Ref": "PublicSubnet" },
+        <#else>
+            "SubnetId": { "Ref": "SubnetId" },
+        </#if>
+        "SecurityGroups" : [
```
lajosrodek, line 546: Putting back my earlier comment that disappeared due to code changes:

According to the [respective AWS CF doc page](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-mounttarget.html), `SecurityGroups` may contain

> Up to five VPC security group IDs

so this limit (if it is a hard one, which would be worth double-checking) may be easily exceeded when having many instancegroups and multiple SGs per instancegroup. I fear it will not be trivial to come up with a perfect solution here.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -533,6 +533,28 @@
...
+          <#list instanceGroups as group>
+            <#if group.cloudSecurityIds?size != 0>
```
lajosrodek, line 548: I am not sure if it is a valid scenario, but imagine that some instancegroup has prescribed SGs (say `sg-1` and `sg-2`), while some other instancegroup does not have a predefined SG (and thus gets an SG created above like `ClusterNodeSecurityGroupNotExisting`). Then this logic will generate a JSON fragment as below:

```
"SecurityGroups" : [
"sg-1",
"sg-2",
{ "Ref" : "ClusterNodeSecurityGroupNotExisting" }
]
```
So the list will be heterogeneous, some items being simple strings, while others being objects (with a single `"Ref"` property). If this is indeed a valid scenario, I wonder if CF accepts such a JSON.

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
lajosrodek: Sorry, this does not seem to be done. It is still based on JUnit 4.

lajosrodek (changes requested):
Some questions and an UT request.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
lajosrodek: @doktoric, @topolyai5   Is it possible to encounter the scenario described above?

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
LinaAtAustin: I added new file with test in CloudStorageConverterTest, and it is JUnit 5 now.

I added new tests in the existing file ClusterToClusterV4ResponseConverterTest, which uses Junit 4. Do you want to change this to JUnit 5 as well? that would requires me to change existing tests. I can do it.

doktoric:
based on what I see my question is that. Did you tried to provision a cluster with this ? I am asking because based on @lajosrodek the CF template generation can cause issues.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
lajosrodek: After talking to Richard, it seems likely that custom DLs do permit such situations. So even it does not affect customers, this internal scenario still needs to be covered.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: I have tries on AWS console for two scenarios, the observations are
1) When adding security groups ingress_SG as ingress of another security group target_SG (this is the original implementation to create a target_SG for the mount target, then adding the ingress_SG of the EC2 instances as ingress of the target_SG), there is hard limit of only 5 ingress_SG can be added. 
Security Group ingress does not allow more than 5 SG
Finds the same Security Group when ingress has less than 5 Security Groups
2) When adding security groups ingress_SG to the mount target of EFS directly (this is the update implementation based on Richard suggestion. We don't create a target_SG for the mount target. Instead, we use the ingress_SG of the EC2 instances directly), there is also a hard limit of only 5 ingress_SG can be added. 
EFS mount target does not allow more than 5 SG

My conclusion is that both approaches have hard-limit of 5 SG can be used for access control for EFS. I can run a local deployment and verify using more than 5 ingress_SG will fail the EFS deployment. 

This indicates when it is possible for customers to specify more than 5 SGs and we want to guarantee the permission works, we need to look for other approaches. One approach I have identified is to use access point instead of mount group. I am experimenting this approach right now.

LinaAtAustin:
Richard, I have not created a cluster with the scenario described by Lajos mentioned that has hybrid SG configuration, i.e., some instance groups have SG specified, but some other other instance groups do not specify SG and requires SG to be created. 
The main issue I am working on is the hard limit of 5 SG for EFS access. If customers can create more than 5 SGs, we need to look for another approach to make sure all deployment scenarios work.

LinaAtAustin:
I have done the following tests
1) added more testing cases in CloudFormationTemplateBuilderTest. Some instance group has existing security group ID, and some group does not. The generated CF is valid json string. This test does not test if CF accepts generated string.
      "MountTarget":{
         "Type":"AWS::EFS::MountTarget",
         "Properties":{
            "FileSystemId":{
               "Ref":"ElasticFileSystem"
            },
            "SubnetId":{
               "Ref":"SubnetId"
            },
            "SecurityGroups":[
               {
                  "Ref":"ClusterNodeSecurityGroupgateway"
               },
               "sg-02fe9cf52ed86b28b",
               "sg-03610a4f8454e526f",
               "sg-03617a9585e72f75a",
               "sg-04ad3ca3786d34436",
               "sg-071f64ce36f37f14d"
            ]
         }
      }

2) I have tried to created such scenario in local deployment. However, the UI does not allow me to configure security group in some instance group and let other instance group to create new security group. 
Also, only one SG can be configured per instance group on UI. To have more than 5 SG, I have to add code in "https://github.com/hortonworks/cloudbreak/blob/master/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SecurityAccessManifester.java#L36" to add more SG in SecurityGroupV4Request. And the provision failed. I checked AWS console, and there is no EFS created. However, I cannot find error message with the reason of failure.

Local deployment provision failed

lajosrodek:
@LinaAtAustin Re "Some instance group has existing security group ID, and some group does not."
1. Could you please check if CF actually accepts (and correctly interprets) the template when the MountTarget `"SecurityGroups"` block contains this mixed setup (i.e. plain string items as well as `{ "Ref" : "ClusterNodeSecurityGroup... }` ones)?
2. The UI certainly does not permit this, nor does the public CDP CLI. With the internal CLI build, however, you can provision custom DLs that have whatever combinatination of SGs desired. I will send you the details offline. (This implies that while customers can never encounter this, CLDR eng / QE / field etc. folks may still run into such setups.)

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
LinaAtAustin: done

LinaAtAustin:
EFS mount target created with both existing and newly created security groups

I have two instance groups: one with existing security group, and another does not have, so cloud formation will create one for it. I have being able to create EFS that contains both security groups using cdp cli at local deployment. So the mixture of existing security group with newly created security group does not cause issue. 

"MountTarget":{
"Type":"AWS::EFS::MountTarget",
"Properties":{
"FileSystemId":{
"Ref":"ElasticFileSystem"
},
"SubnetId":{
"Ref":"SubnetId"
},
"SecurityGroups":[
{
"Ref":"ClusterNodeSecurityGroupgateway"
},
"sg-03610a4f8454e526f"
]
}
}

I will test with more than 5 security groups next.

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
LinaAtAustin: I have confirmed that CloudFormation does NOT allow more than 5 security groups associated with EFS.
I have an instance group with 5 existing security groups, and another instance group with newly created security group. they will all be associated with the mount target of the EFS. And the provision of data lake failed.

"DATA LAKE STATUS
 Provisioning Failed
REASON
Data Lake creation failed: CloudFormation stack lina-0215-2-dl-184 creation failed. Resource never entered the desired state as it failed. The following resource(s) failed to create: [AmbariNodesmaster, ElasticFileSystem, ClusterNodeSecurityGroupidbroker]. AmbariNodesmaster: The security group 'sg-02fe9cf52ed86b28b' does not exist (Service: AmazonAutoScaling; Status Code: 400; Error Code: ValidationError; Request ID: 4990c62f-c269-483b-abee-00e3778d192e; Proxy: null), ClusterNodeSecurityGroupidbroker: Resource creation cancelled, ElasticFileSystem: Resource creation cancelled"

From AWS console, EFS was created, but it has no mount target because a mount target does not allow more than 5 security groups.

LinaAtAustin: EFS mount target not created due to more than 5 security groups

LinaAtAustin:
@lajosrodek if we have to allow the mount target to support more than 5 security groups, then I will remove the code for creating mount target in aws-cf-stack.ftl, and just commit the code that has bug fix and added testing code. Then I will start another review for using access point to mount EC2 instance.

1. I have created EFS in CloudFormation. I will create EFS access point in Cloud Formation. And I can get it done quickly. Then I can have EFS ID and access point ID. This is the CloudFormation change for access point https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-accesspoint.html, and I found the 
2. Pass data to Salt, then in Salt, mount EC2 instance to EFS access point "$ mount -t efs -o tls,accesspoint=fsap-12345678 fs-12345678: /localmountpoint". https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#mount-with-access-point
3. remount EFS when EC2 instance restarts. "Update the EC2 /etc/fstab file with an entry for the EFS file system." see section "Using /etc/fstab to mount automatically" in https://docs.aws.amazon.com/efs/latest/ug/mount-fs-auto-mount-onreboot.html. I plan not to use security group, instead, use access point. So "In Step 6: Configure Security Group, set Assign a security group to Select an existing security group. " does not apply. This is some discussion on the benefits of access point over mount target https://www.reddit.com/r/aws/comments/kwibz8/efs_access_point_vs_mount_target/

lajosrodek:
@LinaAtAustin That Reddit thread (which cannot be really considered an official AWS documentation) has only a single reply, and even that implies that EFS Access Points are not -- and cannot be -- restricted to a particular VPC. Mount targets have a private IP as mentioned there, whereas Access Points do not (they have no associated dedicated DNS entry at all). As discussed in Slack, I strongly feel that Access Points are not appropriate for us due to their limited security restrictions. (They only support IAM policy conditions, but that is really cumbersome to employ in our use case.)

cloud-aws/src/main/resources/templates/aws-cf-stack.ftl
lajosrodek: Thanks for checking! 👍

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/ClusterToClusterV4ResponseConverterTest.java
lajosrodek: Super, thanks! 👍

LinaAtAustin:
@lajosrodek I can make changes to the security groups for mount target. Is the following approach acceptable?
1) take the first security group from each instance group. 
2) If not reaching 5 after removing duplicates, get the second security group from each instance group. 


What if there are more than 5 instance groups? then we need to make sure there are no more than 5 unique security groups used in all instance groups.

LinaAtAustin:
@doktoric @topolyai5 I have discussed with Lajos offline and the suggested approach is
1. I add a new field in DL template for EFS configuration. One of the EFS configuration is the list of instance groups whose security groups will be used on EFS mount target. And I add a validation to check how many SG total for those allowed instance groups. If exceeding 5, then fail validation. This can ensure the provision of CloudFormation will not have issue due to mount target. This work will be done in another jira CB-11225
2. From that EFS configuration parameter, I only use the SG of those instance groups to attach to the mount target. This update will be done in this jira. 

Can you let me know if you agree with this approach? Thanks!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
```diff
@@ -53,9 +55,17 @@
...
+    private List<String> associatedInstanceGroupNames;
+
     public AwsEfsFileSystem(String backupPolicyStatus, boolean encrypted, String fileSystemPolicy, Map<String, String> fileSystemTags, String kmsKeyId,
-            List<String> lifeCyclePolicies, String performanceMode, Double provisionedThroughputInMibps, String throughputMode) {
-        this.backupPolicyStatus = backupPolicyStatus != null ? backupPolicyStatus : BACKUP_POLICY_STATUS_DEFAULT;
+            List<String> lifeCyclePolicies, String performanceMode, Double provisionedThroughputInMibps, String throughputMode,
+            List<String> associatedInstanceGroupNames) {
+
+        if (associatedInstanceGroupNames == null || associatedInstanceGroupNames.size() == 0) {
+            throw new InvalidInputException("EFS associated instance group names are not set");
```
lajosrodek, line 65: I do not think it makes sense to throw an AWS exception in CB. Please use something more standard like `IllegalArgumentException` or whatever more CB specific.

Apart from the exception type, we shall have some proper validation already at the controller/converter/service level so that only sane values are passed down to `cloud-aws`.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
```diff
@@ -173,4 +183,30 @@ public String getThroughputMode() {
...
+            backupPolicyStatus = BACKUP_POLICY_STATUS_DEFAULT;
+        }
+
+        if (performanceMode == null) {
+            performanceMode = PERFORMANCE_MODE_DEFAULT;
+        }
+
+        if (throughputMode == null) {
+            throughputMode = THROUGHPUT_MODE_DEFAULT;
+        }
```
lajosrodek, line 206: What is the benefit for doing this instead of the former default fallback init in the constructor?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
```diff
@@ -53,9 +55,17 @@
 
     private String throughputMode;
 
+    private List<String> associatedInstanceGroupNames;
+
     public AwsEfsFileSystem(String backupPolicyStatus, boolean encrypted, String fileSystemPolicy, Map<String, String> fileSystemTags, String kmsKeyId,
```
lajosrodek, line 60: Not a big deal but this constructor is taking 9 arguments. Perhaps it would make sense to use the Builder pattern instead to have a cleaner code.

common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/AwsEfsParameters.java
```diff
@@ -34,6 +34,10 @@
 
     private String throughputMode;
 
+    // EFS only accepts the security groups of the listed instance groups and
+    // only attach to the EC2 instances in those instance groups
```
lajosrodek, line 38: Change "attach" -> "attaches" or "gets attached".

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
```diff
@@ -139,6 +143,7 @@ public FileSystem requestToAdditionalFileSystem(CloudStorageBase cloudStorageReq
         configurations.put(CloudEfsConfiguration.KEY_PERFORMANCE_MODE, efsParameters.getPerformanceMode());
         configurations.put(CloudEfsConfiguration.KEY_PROVISIONED_THROUGHPUT_INMIBPS, efsParameters.getProvisionedThroughputInMibps());
         configurations.put(CloudEfsConfiguration.KEY_THROUGHPUT_MODE, efsParameters.getThroughputMode());
+        configurations.put(CloudEfsConfiguration.KEY_ASSO_INSTANCE_GROUP_NAMES, efsParameters.getAssociatedInstanceGroupNames());
```
lajosrodek, line 146: Where is `getAssociatedInstanceGroupNames()` gets validated? I have seen a null/emptiness check in `AwsEfsFileSystem`, but it is probably too late to do that there. In particular, we shall check for the total number of SGs involved in the specified instancegroups, and freak out if the count exceeds the limit of 5.

lajosrodek (changes requested):
Thanks Lina. Some further comments:

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
doktoric: +1

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
doktoric: KEY_ASSO_INSTANCE_GROUP_NAMES -> KEY_ASSOCIATED_INSTANCE_GROUP_NAMES

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
LinaAtAustin: I was planning to do validation in the next jira CB-11225 to keep this one small. I will add validation in this one then.

Changed the exception type.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
LinaAtAustin: done

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/cluster/CloudStorageConverter.java
LinaAtAustin: I have added the validation in com.sequenceiq.common.api.cloudstorage.validation.CloudStorageValidator.
changed the key name to KEY_ASSOCIATED_INSTANCE_GROUP_NAMES

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/efs/AwsEfsFileSystem.java
LinaAtAustin: check style does not allow that. The function is too complicated. NPath Complexity exceeds the limit. I have to break it into another function to lower the complexity of the constructor function

common-model/src/main/java/com/sequenceiq/common/api/cloudstorage/AwsEfsParameters.java
LinaAtAustin: done

doktoric:
I guess we can close this. reopen if required

PR #9915, opened 2021-01-29
CB-10521 Refactor expected exception

aszegedi:
Actually in mock and authorization tests are lot of validation steps for an expected error/exception. The stack-traces of these exceptions are logged as ERROR in the related suite log. These additional errors are confusing when the test are failing and we are looking for the root cause of the issue.

So we should not log the stack-trace of the expected exceptions.

I've created new `when` test step as `whenException` where a predefined exception is handled by default. This step is failing only in case of any other kind of error/exception. Here we can catch this with the well known `TestFailException` then add to the `TestContext.exceptionMap`.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
@@ -18,13 +18,13 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
-import org.springframework.util.StringUtils;
 import org.testng.ITestResult;
 import org.testng.Reporter;
 
@@ -186,6 +186,46 @@ public abstract class TestContext implements ApplicationContextAware {
         return instanceAwait;
     }
 
+    public <E extends Exception, T extends CloudbreakTestDto, U extends MicroserviceClient> T whenException(Class<T> entityClass,
+            Class<? extends MicroserviceClient> clientClass, Action<T, U> action, Class<E> expectedException) {
+        return whenException(entityClass, clientClass, action, expectedException, emptyRunningParameter());
+    }
+
+    public <E extends Exception, T extends CloudbreakTestDto, U extends MicroserviceClient> T whenException(Class<T> entityClass,
+            Class<? extends MicroserviceClient> clientClass, Action<T, U> action, Class<E> expectedException, RunningParameter runningParameter) {
+        return whenException(getEntityFromEntityClass(entityClass, runningParameter), clientClass, action, expectedException, runningParameter);
+    }
+
+    public <E extends Exception, T extends CloudbreakTestDto, U extends MicroserviceClient> T whenException(T entity,
+            Class<? extends MicroserviceClient> clientClass, Action<T, U> action, Class<E> expectedException) {
+        return whenException(entity, clientClass, action, expectedException, emptyRunningParameter());
+    }
+
+    public <E extends Exception, T extends CloudbreakTestDto, U extends MicroserviceClient> T whenException(T entity,
+            Class<? extends MicroserviceClient> clientClass, Action<T, U> action, Class<E> expectedException, RunningParameter runningParameter) {
+        checkShutdown();
+        String key = runningParameter.getKey();
+        if (StringUtils.isBlank(key)) {
+            key = action.getClass().getSimpleName();
+        }
+        if (!getExceptionMap().isEmpty() && runningParameter.isSkipOnFail()) {
+            LOGGER.info("Should be skipped because of previous error. when [{}]", key);
+            return entity;
+        }
+
+        CloudbreakUser who = getWho(runningParameter);
+
+        LOGGER.info("when exception {} action on {} by {}, name: {}", key, entity, who, entity.getName());
+        Log.whenException(LOGGER, action.getClass().getSimpleName() + " action on " + entity + " by " + who);
+
+        try {
+            return doAction(entity, clientClass, action, who.getAccessKey());
+        } catch (Exception e) {
+            exceptionValidation(expectedException, e, key, runningParameter, "whenException");
+        }
+        return entity;
+    }
+
     public <T extends CloudbreakTestDto, U extends MicroserviceClient> T when(Class<T> entityClass, Class<? extends MicroserviceClient> clientClass,
             Action<T, U> action) {
         return when(entityClass, clientClass, action, emptyRunningParameter());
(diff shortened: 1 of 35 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -186,6 +186,64 @@ public InstanceAwait getInstanceAwait() {
...
+        LOGGER.info("when {} action on {} by {}, name: {}", key, entity, who, entity.getName());
+        Log.when(LOGGER, action.getClass().getSimpleName() + " action on " + entity + " by " + who);
+
+        try {
+            return doAction(entity, clientClass, action, who.getAccessKey());
+        } catch (Exception e) {
+            if (!e.getClass().equals(expectedException)) {
+                String message = String.format("Expected exception (%s) does not match with the actual exception (%s).",
+                        expectedException, e.getClass());
+                getExceptionMap().put("when", new TestFailException(message));
```
topolyai5, line 227: Mybe do you want to write the `whenException` in the log?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -186,6 +186,64 @@ public InstanceAwait getInstanceAwait() {
...
+
+        CloudbreakUser who = getWho(runningParameter);
+
+        LOGGER.info("when {} action on {} by {}, name: {}", key, entity, who, entity.getName());
+        Log.when(LOGGER, action.getClass().getSimpleName() + " action on " + entity + " by " + who);
+
+        try {
+            return doAction(entity, clientClass, action, who.getAccessKey());
+        } catch (Exception e) {
+            if (!e.getClass().equals(expectedException)) {
```
topolyai5, line 224: Could you please refactor into a new method and reuse here and in `expect` as well?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/ClusterCreationTest.java
```diff
@@ -29,9 +29,10 @@ public void createClusterWithoutNetworkAndPlacementShouldBadRequestException(Moc
                 .given(StackTestDto.class)
                 .withEmptyNetwork()
                 .withEmptyPlacement()
-                .when(StackTestAction::create, key("error"))
-                .expect(BadRequestException.class, key("error")
-                        .withExpectedMessage("Network must be specified!"))
+                .whenException(StackTestAction::create, BadRequestException.class,
+                        key("error")
```
topolyai5, line 33: I think, in this case, we don't need the key. But it won't cause any error.

topolyai5 (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
aszegedi: This is intentional. If the actual and expected exception don't met, we should see a `TestFailException` and put this to the `ExceptionMap` for later validation. Along with this we will see `the message` in the console log and the formatted HTML log as well.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/ClusterCreationTest.java
aszegedi: This will be the `key` at `LOGGER.info("when {} action on {} by {}, name: {}", key, entity, who, entity.getName());`. So we should change this key to a more meaningful one, for example: **badRequest** or something similar.

Thank you for pointing this out!

topolyai5: Yes, you are right. Thanks.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
topolyai5: I can easier to find the method by action. If I read the `whenException`, I know where should I search in the code. If I read `when` probably I will search the when methods

aszegedi: That's the point! You are right, I've overlooked your comment in the code.

topolyai5 (approved):
LGTM

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -204,10 +244,9 @@ public InstanceAwait getInstanceAwait() {
             RunningParameter runningParameter) {
         checkShutdown();
         String key = runningParameter.getKey();
-        if (StringUtils.isEmpty(key)) {
+        if (key == null || key.isEmpty()) {
```
lnardai, line 247: ` StringUtils.isBlank(null)      = true
 StringUtils.isBlank("")        = true  
 StringUtils.isBlank(" ")       = true  
 StringUtils.isBlank("bob")     = false  
 StringUtils.isBlank("  bob  ") = false`

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CreateDhWithDatahubCreator.java
```diff
@@ -34,6 +34,11 @@
 import com.sequenceiq.it.cloudbreak.util.AuthorizationTestUtil;
 
 public class CreateDhWithDatahubCreator extends AbstractIntegrationTest {
+
+    private static Actor envcreatorb = Actor.useRealUmsUser(AuthUserKeys.ENV_CREATOR_B);
+
+    private static Actor zero = Actor.useRealUmsUser(AuthUserKeys.ZERO_RIGHTS);
```
lnardai, line 40: `zero` might be not that intuitive from the testContext. Can we change this to zeroRightsUser?
Also it looks strange to move these out to Class level. is this a pattern elsewhere?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CredentialCreateTest.java
```diff
@@ -20,6 +20,8 @@
 
 public class CredentialCreateTest extends AbstractIntegrationTest {
 
+    private static final Actor ENVCREATORB = Actor.useRealUmsUser(AuthUserKeys.ENV_CREATOR_B);
```
lnardai, line 23: This in not consistent with `CreateDhWithDatahubCreator`, in the other test it was not Static final Actor. I'm not a fan of these Class level extractions.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/CredentialCreateTest.java
```diff
@@ -75,11 +80,11 @@ public void testCreateCredentialWithZeroRoles(TestContext testContext) {
...
                 .given(CredentialTestDto.class)
-                .when(credentialTestClient.create(), RunningParameter.key("Unauthorized"))
-                .expect(ForbiddenException.class,
-                        RunningParameter.expectedMessage("You have no right to perform " +
-                                "environments/createCredential in account 460c0d8f-ae8e-4dce-9cd7-2351762eb9ac")
-                                .withKey("Unauthorized"))
+                .whenException(credentialTestClient.create(), ForbiddenException.class,
+                        RunningParameter
+                                .expectedMessage("You have no right to perform environments/createCredential in account 460c0d8f-ae8e-4dce-9cd7-2351762eb9ac")
+                                .withKey("Unauthorized")
```
lnardai, line 86: A general comment, is it still needed to pack these behind keys? since with this new statement we are checking if the exception happened or not right away. Can we omit these keys with this implementation?

lnardai (approved):
Overall looks good, left some minor comments.
 I think it's a great improvement on readability and the test logs are much easier to comprehend.

PR #9919, opened 2021-02-01
CB-10864 Expose additional rights

agov:
Added some missing account level rights to RightV4 to be used on UI


authorization-common-api/src/main/java/com/sequenceiq/authorization/info/model/RightV4.java
```diff
@@ -12,6 +12,12 @@
 public enum RightV4 {
     // account level
     ENV_CREATE(AuthorizationResourceAction.CREATE_ENVIRONMENT),
+    LIST_ASSIGNED_ROLES(AuthorizationResourceAction.LIST_ASSIGNED_ROLES),
```
horadla23, line 15: is this an account level right?

agov: @horadla23 Yes, it's assigned with IamUser

horadla23 (approved):

PR #9922, opened 2021-02-02
CB-10980 - Validate FreeIPA backup permission after after deployment

jamisonbennett:
Before installing FreeIPA, perform a test to ensure the instance has
write access to the object store for backup files. This improves the
delay before failues are reported. Previously it would take more than
40 minutes, now an error is reported within 5 minutes.

This was validated by timing how long it took to fail when deployed
using a local instance of cloudbreak and deployming on AWS.


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -141,6 +141,9 @@
     @Value("${cb.max.salt.database.dr.retry.onerror:5}")
     private int maxDatabaseDrRetryOnError;
 
+    @Value("${cb.max.salt.validation.retry:3}")
+    private int maxValidationRetry;
```
oleewere, line 145: i think this should be a bit more specific as that is only used for storage validation

oleewere (changes requested):

jamisonbennett:
It looks like the FreeIpaInstallService was refactored recently. I will need to update this PR to match the current master.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaCloudStorageValidationService.java
```diff
@@ -0,0 +1,80 @@
...
+        return allNodes;
+    }
+
+    private void validateNodeIsPresent(Set<Node> allNodes) throws CloudbreakOrchestratorFailedException {
+        if (allNodes.isEmpty()) {
+            String errorMessage = "There are no nodes to validate FreeIPA cloud storage.";
+            LOGGER.error(errorMessage);
+            throw new CloudbreakOrchestratorFailedException(errorMessage);
+        }
+    }
```
lacikaaa, line 79: this part looks like duplication from `FreeIpaInstallService`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaCloudStorageValidationService.java
```diff
@@ -0,0 +1,80 @@
...
+    @Inject
+    private FreeIpaSaltConfigService freeIpaSaltConfigService;
+
+    public void validate(Long stackId) throws CloudbreakOrchestratorException {
+        Stack stack = stackService.getByIdWithListsInTransaction(stackId);
+        Set<InstanceMetaData> instanceMetaDatas = stack.getNotDeletedInstanceMetaDataSet();
+        List<GatewayConfig> gatewayConfigs = gatewayConfigService.getGatewayConfigs(stack, instanceMetaDatas);
+        Set<Node> allNodes = mapInstancesToNodes(instanceMetaDatas);
+
+        createAndPushPillarsToNodes(stackId, stack, gatewayConfigs, allNodes);
```
lacikaaa, line 49: just wondering if we could avoid pushing the same pillars twice. maybe a new flowstep would be necessary. I'm concerned because creating the pillars includes multiple interservice calls also, so it could take a bit of time.

akanto:
Thanks for doing this! This is the most frequent error message what customers are receiving today:



com.sequenceiq.environment.exception.FreeIpaOperationFailedException: FreeIpa creation operation failed. FreeIpa creation failed. Status: 'CREATE_FAILED' statusReason: 'Failed: Orchestrator component went failed in 28 min(s), message: Target: [ipaserver0.jxxss.dsdsd.cloudera.site] Node: ipaserver0.jose-cdp.ztoh-5shg.cloudera.site Error(s): Failed to upload freeipa backup to object storage, please check the assigned permissions to the instance'


My ask is that can we rephrase this error message as well, by providing more detailed information:
- what instance profile/msi has been used what think is not good
- what is the location where we tied to upload, but we had no permission
- what is the UI / CLI input parameter what they shall check

Probably we can also eliminate the com.sequenceiq.environment.exception.FreeIpaOperationFailedException from the beginning.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/flow/FreeIpaCloudStorageValidationService.java
jamisonbennett: Good point. I cleaned up a few other places that had similar logic.

jamisonbennett:
@akanto the error message now looks like this:
```
com.sequenceiq.environment.exception.FreeIpaOperationFailedException: FreeIpa creation operation failed. FreeIpa creation failed. Status: 'CREATE_FAILED' statusReason: '1. Validating FreeIPA cloud storage permission for backup failed. The instance profile arn:aws:iam::XXXXXXXX:instance-profile/mow-dev did not have permission to write to s3://jamison-cdp1/jamison-bennett-env2/logs/cluster-backups/freeipa/jamison-bennett-env2-15644-freeipa_56693871-0a6b-4c91-ad7c-2412c7ae7245. If provisioning was done using the UI, then verify the log's instance profile and logs location base when provisioning in the UI. If provisioning was done using the CLI, then verify the JSON which was provided when creating the environment. Specifically verify the logStorage.instanceProfile and logStorage.storageLocationBase. Refer to Cloudera documentation at https://docs.cloudera.com/management-console/cloud/environments/topics/mc-idbroker-minimum-setup.html for the required rights.'
```

jamisonbennett:
@oleewere I cleaned up several places that had similar code for converting `Set<InstanceMetaData>` into `Set<Node>`. One place was the diagnostics. There are 2 minor changes in the code which I think would have actually been a problem before but I wanted you to check this.
1. The diagnostics conversion for public IP addresses now uses the "wrapper" which falls back to the private IP if the public IP is not set.
2. There is a new check to ensure that the list is not empty.

oleewere (approved):
lgtm

jamisonbennett:
closing in favor of merging it into 2.38 https://github.com/hortonworks/cloudbreak/pull/9969

PR #9926, opened 2021-02-02
CB-10708 Addig infrastructure-mock.p12 cert to java cacerts during startup for IT. 

lnardai:
- Bumped jersey core version to use same as in Spring modules.
- need to revert to secure(false) once it passed for local execution


topolyai5 (approved):
LGTM

biharitomi (approved):
LGTM

PR #9928, opened 2021-02-02
CB-10984 anonymize the salt-boot password when logging the user-data …

keyki:
…script

Anonymization was not called when we logged the generated user-data script.
As part of this PR both the CB and FreeIPA user-data will be anonymized from
sensitive data. I also noticed that the regex matcher doesn't work when
the password is in between quotas so I have fixed that issue as well. The
user-data will be logged in the following way:

export CLOUD_PLATFORM="AZURE"
export START_LABEL=98
export PLATFORM_DISK_PREFIX=sd
export LAZY_FORMAT_DISK_LIMIT=12
export IS_GATEWAY=true
export TMP_SSH_KEY="****"
export SSH_USER=cloudbreak
export SALT_BOOT_PASSWORD=****
export SALT_BOOT_SIGN_KEY=****
export CB_CERT=cert
export IS_PROXY_ENABLED=true
export PROXY_HOST=proxy.host
export PROXY_PORT=1234
export PROXY_USER="user"
export PROXY_PASSWORD="****"
export IS_CCM_ENABLED=false
export IS_CCM_V2_ENABLED=false

### DIFF
--- a/common/src/test/java/com/sequenceiq/cloudbreak/common/anonymizer/AnonymizerUtilTest.java
+++ b/common/src/test/java/com/sequenceiq/cloudbreak/common/anonymizer/AnonymizerUtilTest.java
@@ -29,6 +29,84 @@ public class AnonymizerUtilTest {
     }
 
     @Test
+    public void testWithBothSimpleAndDoubleQuotes() {
+        String testData = ""
+                + "export test1_password=pwd\n"
+                + "export test1_pass=pwd\n"
+                + "export test1_secret=pwd\n"
+                + "export test1_key=pwd\n"
+                + "export test1_credential=pwd\n"
+
+                + "export test2_password:pwd\n"
+                + "export test2_pass:pwd\n"
+                + "export test2_secret:pwd\n"
+                + "export test2_key:pwd\n"
+                + "export test1_credential:pwd\n"
+
+                + "export test3_password:'pwd'\n"
+                + "export test3_pass:'pwd'\n"
+                + "export test3_secret:'pwd'\n"
+                + "export test3_key:'pwd'\n"
+                + "export test3_credential:'pwd'\n"
+
+                + "export test4_password='pwd'\n"
+                + "export test4_pass='pwd'\n"
+                + "export test4_secret='pwd'\n"
+                + "export test4_key='pwd'\n"
+                + "export test4_credential='pwd'\n"
+
+                + "export test5_password=\"pwd\"\n"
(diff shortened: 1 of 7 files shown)


common/src/test/java/com/sequenceiq/cloudbreak/common/anonymizer/AnonymizerUtilTest.java
```diff
@@ -28,6 +28,72 @@ public void testCapitalHidePasswordSingleQuote() {
...
+                + "export test4_secret=\"pwd\"\n"
+                + "export test4_key=\"pwd\"\n"
+                + "export test4_credential=\"pwd\"\n"
+
+                + "export \"test5_password\":\"pwd\"\n"
+                + "export \"test5_pass\":\"pwd\"\n"
+                + "export \"test5_secret\":\"pwd\"\n"
+                + "export \"test5_key\":\"pwd\"\n"
+                + "export \"test5_credential\":\"pwd\"\n";
+
```
jamisonbennett, line 63: Can you please add a test for `password=\'pwd\'` `pass=\'pwd\'` etc.? I think validation is missing for all the cases with an equals symbol and with single quotes.

keyki: Sure!

keyki: test3 is with single quotes but with `:`. I can add tests with `=` and single quote.

jamisonbennett (approved):

core/src/main/resources/logback.xml
```diff
@@ -6,7 +6,7 @@
     <logger name="org.springframework.statemachine.support" level="ERROR"/>
     <logger name="com.sequenceiq.ambari.client.AmbariClientUtils" level="ERROR"/>
     <logger name="org.apache.kafka.clients.producer" level="ERROR"/>
-    <logger name="com.amazonaws.request" level="DEBUG"/>
+    <!-- <logger name="com.amazonaws.request" level="DEBUG"/> -->
```
lacikaaa, line 9: do we want to keep this? if yes we could create a new appender with anonymization enabled on that log

keyki: I don't want to do that. We send a lot of requests to AWS and it has a serious performance drawback to anonymize all the requests.

PR #9932, opened 2021-02-02
CB-10988 Disable default AvailabilitySet assignment

bergerdenes:
If the VMs are placed into AvailabilitySet the maximum host count is 200
which is not adequate for some customers.


pdarvasi (approved):

pdarvasi:
@bergerdenes @doktoric Does this PR mean we have an agreement to turn off the default AS assignment?

PR #9933, opened 2021-02-02
CDPCP-3685 Update usermanagement.proto and references to account pass…

gitmfox:
…word policy

This commit updates usermanagement.proto to be in sync with the
version UMS is running in mow-dev. The motivation for the update
is to bring in changes needed to optimize password hash update
during user sync.

Since the last time the proto was updated, a compatibility-breaking
change was made to the Account message: the 'passwordPolicy' field
was replaced by 'globalPasswordPolicy' and 'machineUserPasswordPolicy'.
The 'passwordPolicy' field number was reused for 'globalPasswordPolicy',
which means that for the last two months or so, Cloudbreak was getting
the global password policy when it accessed 'passwordPolicy'. If any
customers set a machine user-specific policy, it was ignored.

This commit updates the references to the password policy fields,
and implements the semantics of the machine user policy: if present,
it is used for machine users; otherwise, the global policy is used,
if present.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
@@ -125,7 +125,7 @@ public class PasswordService {
         try {
             String userId = getUserIdFromUserCrn(actorCrn, userCrn);
 
-            Optional<Instant> expirationInstant = calculateExpirationTime(actorCrn, accountId);
+            Optional<Instant> expirationInstant = calculateExpirationTime(userCrn, accountId);
 
             List<SetPasswordRequest> requests = new ArrayList<>();
             for (Stack stack : stacks) {
@@ -157,11 +157,12 @@ public class PasswordService {
     }
 
     @VisibleForTesting
-    Optional<Instant> calculateExpirationTime(String actorCrn, String accountId) {
+    Optional<Instant> calculateExpirationTime(String userCrn, String accountId) {
         LOGGER.debug("calculating expiration time for password in account {}", accountId);
         UserManagementProto.Account account = umsClient.getAccountDetails(INTERNAL_ACTOR_CRN, accountId, MDCUtils.getRequestId());
-        if (account.hasPasswordPolicy()) {
-            long maxLifetime = account.getPasswordPolicy().getWorkloadPasswordMaxLifetime();
+        Optional<UserManagementProto.WorkloadPasswordPolicy> passwordPolicy = getPasswordPolicyForUser(account, userCrn);
+        if (passwordPolicy.isPresent()) {
+            long maxLifetime = passwordPolicy.get().getWorkloadPasswordMaxLifetime();
             if (maxLifetime != 0L) {
                 LOGGER.debug("Calculating password expiration by adding lifetime '{}' to current time", maxLifetime);
                 return Optional.of(clock.getCurrentInstant().plusMillis(maxLifetime));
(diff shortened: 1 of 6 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -175,6 +176,21 @@ private void internalSetPasswords(String operationId, String accountId, String a
...
+        switch (crn.getResourceType()) {
+        case USER:
+            return account.hasGlobalPasswordPolicy() ? Optional.of(account.getGlobalPasswordPolicy()) : Optional.empty();
+        case MACHINE_USER:
+            if (account.hasMachineUserPasswordPolicy()) {
+                return Optional.of(account.getMachineUserPasswordPolicy());
+            }
+            return account.hasGlobalPasswordPolicy() ? Optional.of(account.getGlobalPasswordPolicy()) : Optional.empty();
+        default:
+            throw new IllegalArgumentException(String.format("ActorCrn %s is not of resource type USER or MACHINE_USER", userCrn));
```
handavid, line 190: nit: UserCrn instead of ActorCrn

handavid (approved):
In practice, there wasn't any incorrect behavior exposed. The set password api prefers to use the hashed password and expiration from the UMS if FreeIPA supports hashed passwords. It only falls back on calculating the expiration for old old FreeIPAs, which we are confident don't actually exist.

gitmfox:
> In practice, there wasn't any incorrect behavior exposed. The set password api prefers to use the hashed password and expiration from the UMS if FreeIPA supports hashed passwords. It only falls back on calculating the expiration for old old FreeIPAs, which we are confident don't actually exist.

I'm seeing something different in my testing, which leads me to believe that there was some user-visible incorrect behavior. Here's what I'm doing:

1. Use 'cdp iam set-workload-password-policy' to set both a global and machine user-specific password policy, where the two policies have different password lifetimes.
2. Use 'cdp environments set-password' to set a password for a machine user.
3. Use 'ipa user-show --all' on the IPA node to check the password expiration date that was set.

Without this change, the expiration seen in step 3 corresponds to the global password policy, not the machine user-specific policy. With this change, the expiration corresponds to the machine user policy, as it should.

The code in PasswordService that this PR changes is the code that is reading the password expiration from UMS. Before this change, it was always using the global policy, even if a machine user policy was set.

gitmfox:
I see from some more testing that the incorrect behavior only affects the (deprecated) 'environments set-password' API, not 'iam set-workload-password'. @handavid, is that what you were referring to?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
gitmfox: Fixed.

handavid:
> > In practice, there wasn't any incorrect behavior exposed. The set password api prefers to use the hashed password and expiration from the UMS if FreeIPA supports hashed passwords. It only falls back on calculating the expiration for old old FreeIPAs, which we are confident don't actually exist.
> 
> I'm seeing something different in my testing, which leads me to believe that there was some user-visible incorrect behavior. Here's what I'm doing:
> 
> 1. Use 'cdp iam set-workload-password-policy' to set both a global and machine user-specific password policy, where the two policies have different password lifetimes.
> 2. Use 'cdp environments set-password' to set a password for a machine user.
> 3. Use 'ipa user-show --all' on the IPA node to check the password expiration date that was set.
> 
> Without this change, the expiration seen in step 3 corresponds to the global password policy, not the machine user-specific policy. With this change, the expiration corresponds to the machine user policy, as it should.
> 
> The code in PasswordService that this PR changes is the code that is reading the password expiration from UMS. Before this change, it was always using the global policy, even if a machine user policy was set.

Did the machine user you tested have a hashed password? If you called `environments set-password` w/o first calling `iam set-workload-password` then the PasswordHandler will use the wrong expiration. After a machine user has a hashed password in the control plane, the PasswordHandler should use that and the correct expiration regardless of what password was sent in the environments set-password request.

handavid:
retest this please

gitmfox:
run tests

gitmfox:
David:

> Did the machine user you tested have a hashed password? If you called `environments set-password` w/o first calling `iam set-workload-password` then the PasswordHandler will use the wrong expiration. After a machine user has a hashed password in the control plane, the PasswordHandler should use that and the correct expiration regardless of what password was sent in the environments set-password request.

That explains it -- I did indeed call 'environments set-password' without first calling 'iam set-workload-password.' Thanks for the clarification.

gitmfox:
retest this please

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/PasswordService.java
```diff
@@ -175,6 +176,21 @@ private void internalSetPasswords(String operationId, String accountId, String a
         }
     }
 
+    private Optional<UserManagementProto.WorkloadPasswordPolicy> getPasswordPolicyForUser(UserManagementProto.Account account, String userCrn) {
+        Crn crn = Crn.safeFromString(userCrn);
+        switch (crn.getResourceType()) {
```
lacikaaa, line 181: could you fix this switch-case indentation?

gitmfox: > could you fix this switch-case indentation?

Fixed.

lacikaaa:
we use rebase & merge to merge prs. please squash your commits into one

gitmfox:
> we use rebase & merge to merge prs. please squash your commits into one

Squashed.

gitmfox:
@lacikaaa I think this PR may have gotten into a strange state when I squashed the commits. I'm closing it, and will create a new one.

PR #9938, opened 2021-02-03
CB-10839 Cannot open hue UI when Public Endpoint Access Gateway is enabled

hreeve-cloudera:
The original fix for this bug revealed an older bug in the Hue config logic that had been missed
until now. This fixes that logic, by adding a template variable for knox_proxyhosts, and by
correctly setting the template config "hue-knox_proxyhosts" setting, instead of the invalid
"knox_proxyhosts" setting.

Tested with unit tests and in local CB environment.


bergerdenes (approved):

lajosrodek:
Nice work!

cegganesh84:
Thanks for the new unit test!

PR #9940, opened 2021-02-03
CB-10612 API changes for In place enablement of semi-private networks

hreeve-cloudera:
The API changes that will be used for the new in-place upgrade for the new load balancer
and endpoint gateway features. The logic will be triggered via an API call to the
environment service. The environment service will then call the internal stack API
to update the laod balancer for each applicable stack.

As part of the request, the endpoint gateway can be enabled, and public subnets to use for the
gateway provided. Load balancers will be created based on the current entitlements (including
creating data lake load balancers if that entitlement is enabled), and whether or not the endpoint
gateway is enabled for the environment.

This change adds the new API calls and the services they trigger, and stubs out the logic that will
be implemented in subsequent PRs. This change will be broken into three PRs:
- API changes (this one)
- Environment service changes
- Stack service changes

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.ClusterOpDescr
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.ClusterOpDescription.SET_MAINTENANCE_MODE_BY_NAME;
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.ClusterOpDescription.UPDATE_PILLAR_CONFIG;
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.ClusterOpDescription.UPDATE_SALT;
+import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.StackOpDescription.UPDATE_LOAD_BALANCERS;
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.StackOpDescription.CHECK_FOR_UPGRADE_CLUSTER_IN_WORKSPACE;
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.StackOpDescription.CHECK_IMAGE_IN_WORKSPACE;
 import static com.sequenceiq.cloudbreak.doc.OperationDescriptions.StackOpDescription.CHECK_IMAGE_IN_WORKSPACE_INTERNAL;
@@ -429,5 +430,12 @@ public interface StackV4Endpoint {
     @ApiOperation(value = OperationDescriptions.StackOpDescription.RENEW_CERTIFICATE, produces = MediaType.APPLICATION_JSON,
             notes = Notes.RENEW_CERTIFICATE_NOTES, nickname = "renewStackCertificate")
     FlowIdentifier renewCertificate(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
+            @QueryParam("initiatorUserCrn") String initiatorUserCrn);
+
+    @PUT
+    @Path("internal/{name}/update_load_balancers")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = UPDATE_LOAD_BALANCERS, nickname = "updateLoadBalancersInternal")
+    FlowIdentifier updateLoadBalancersInternal(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
             @QueryParam("initiatorUserCrn") String initiatorUserCrn);
 }
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/doc/OperationDescriptions.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/doc/OperationDescriptions.java
@@ -59,6 +59,8 @@ public class OperationDescriptions {
         public static final String DATABASE_RESTORE_INTERNAL = "Performs a restore of the database from a provided location, internal only";
         public static final String ROTATE_CERTIFICATES = "Rotates the certificates of the cluster";
         public static final String RENEW_CERTIFICATE = "Trigger a certificate renewal on the desired cluster which is identified via stack's name";
+        public static final String UPDATE_LOAD_BALANCERS = "Updates an existing cluster with load balancers, including adding the Endpoint Gateway " +
+            "if it's enabled.";
     }
 
     public static class ClusterOpDescription {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerUpdateService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerUpdateService.java
@@ -1,0 +1,15 @@
+package com.sequenceiq.cloudbreak.service;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.dto.NameOrCrn;
+
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.flow.api.model.FlowIdentifier;
+
+@Service
+public class LoadBalancerUpdateService {
+
+    public FlowIdentifier updateLoadBalancers(NameOrCrn nameOrCrn, Long workspaceId) {
+        return null;
+    }
+}
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
@@ -50,6 +50,7 @@ import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.domain.view.StackApiView;
 import com.sequenceiq.cloudbreak.logger.MDCBuilder;
 import com.sequenceiq.cloudbreak.retry.RetryableFlow;
+import com.sequenceiq.cloudbreak.service.LoadBalancerUpdateService;
 import com.sequenceiq.cloudbreak.service.ClusterCommonService;
 import com.sequenceiq.cloudbreak.service.DatabaseBackupRestoreService;
 import com.sequenceiq.cloudbreak.service.StackCommonService;
@@ -113,6 +114,9 @@ public class StackOperations implements ResourceBasedCrnProvider {
 
     @Inject
     private ClusterDBValidationService clusterDBValidationService;
+
+    @Inject
+    private LoadBalancerUpdateService loadBalancerUpdateService;
 
     public StackViewV4Responses listByEnvironmentName(Long workspaceId, String environmentName, List<StackType> stackTypes) {
         Set<StackViewV4Response> stackViewResponses;
@@ -391,5 +395,10 @@ public class StackOperations implements ResourceBasedCrnProvider {
             CertificatesRotationV4Request certificatesRotationV4Request) {
         LOGGER.debug("Starting cluster autotls certificates rotation: " + nameOrCrn);
         return clusterCommonService.rotateAutoTlsCertificates(nameOrCrn, workspaceId, certificatesRotationV4Request);
+    }
+
+    public FlowIdentifier updateLoadBalancers(@NotNull NameOrCrn nameOrCrn, Long workspaceId) {
+        LOGGER.debug("Creating load balancers for stack: " + nameOrCrn);
+        return loadBalancerUpdateService.updateLoadBalancers(nameOrCrn, workspaceId);
     }
 }
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
@@ -25,6 +25,10 @@ public class EnvironmentOpDescription {
     public static final String GET_CRN_BY_NAME = "Get the crn of an environment by name.";
     public static final String GET_NAME_BY_CRN = "Get the name of an environment by crn.";
     public static final String UPDATE_CONFIG_BY_CRN = "Update the configuration for all stacks in the Environment by the Environment CRN";
+    public static final String UPDATE_LOAD_BALANCERS_BY_ENV_NAME = "Updates all cluster in an environment with load balancers, including adding the " +
+        "Endpoint Access Gateway if it's enabled. Environment is specified by name.";
+    public static final String UPDATE_LOAD_BALANCERS_BY_ENV_CRN = "Updates all cluster in an environment with load balancers, including adding the " +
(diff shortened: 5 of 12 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerUpdateService.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.flow.api.model.FlowIdentifier;
+
+@Service
+public class LoadBalancerUpdateService {
+
+    public FlowIdentifier updateLoadBalancers(NameOrCrn nameOrCrn, Long workspaceId) {
+        return null;
```
doktoric, line 13: why is this null?

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
```diff
@@ -227,4 +228,19 @@ void postStartByCrn(@ValidCrn(resource = CrnResourceDescriptor.ENVIRONMENT) @Pat
     @ApiOperation(value = EnvironmentOpDescription.UPDATE_CONFIG_BY_CRN, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
         nickname = "updateConfigsInEnvironmentByCrnV1")
     void updateConfigsInEnvironmentByCrn(@ValidCrn(resource = CrnResourceDescriptor.ENVIRONMENT) @PathParam("crn") String crn);
+
+    @POST
+    @Path("/name/{name}/update_load_balancers")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = EnvironmentOpDescription.UPDATE_LOAD_BALANCERS_BY_ENV_NAME, produces = MediaType.APPLICATION_JSON, notes = ENVIRONMENT_NOTES,
+        nickname = "updateEnvironmentLoadBalancersByNameV11")
+    void updateEnvironmentLoadBalancersByName(@PathParam("name") String envName, @NotNull EnvironmentLoadBalancerUpdateRequest request);
```
doktoric, line 237: why we dont using a put endpoint for this?

doktoric: I mean the current environment service PUT endpoint

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentLoadBalancerService.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.environment.dto.EnvironmentLoadBalancerDto;
+
+@Service
+public class EnvironmentLoadBalancerService {
+
+    public void updateLoadBalancerInEnvironmentAndStacks(EnvironmentDto environmentDto, EnvironmentLoadBalancerDto environmentLbDto) {
+        // TODO
```
doktoric, line 12: something missing here ?

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerUpdateService.java
hreeve-cloudera: Because this PR is intended just to have the API changes and to stub out the services they call. The actual implementation of those services are in PRs 9963 and 9964. I broke the change into three PRs because it's a large change, and I thought that would be easier for people to review.

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentLoadBalancerService.java
hreeve-cloudera: No. This PR is just to stub these things out, and they're filled out in https://github.com/hortonworks/cloudbreak/pull/9963.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerUpdateService.java
hreeve-cloudera: Updated the description to make this more clear.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/endpoint/EnvironmentEndpoint.java
hreeve-cloudera: I wasn't familiar with the PUT method, but after reading about it it does make more sense. I'll update the API call.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
```diff
@@ -379,4 +383,9 @@ public CertificatesRotationV4Response rotateAutoTlsCertificates(@NotNull NameOrC
         LOGGER.debug("Starting cluster autotls certificates rotation: " + nameOrCrn);
         return clusterCommonService.rotateAutoTlsCertificates(nameOrCrn, workspaceId, certificatesRotationV4Request);
     }
+
+    public FlowIdentifier updateLoadBalancers(@NotNull NameOrCrn nameOrCrn, Long workspaceId) {
```
enis, line 387: Is there any reason why this is not part of `editByName(EnvironmentEditRequest)` rather than a different method altogether? Seems like the `EnvironmentEditRequest` already takes the `EnvironmentNetworkRequest` that contains info about access gateway related params.

enis (commented):
You have used the access gateway in the CB API as well before. Is there any reason why these are "update load balancer" rather than "update access gateway" type of naming?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
enis: If so most of the classes in this PR may not be needed.

hreeve-cloudera: As far as I can tell, the `editByName` logic just updates the environment as its stored in the database. It doesn't actually do things like change the created cloud resources to reflect the updated environment option, or create new resources. That's what I need to do. So I decided to create my own API call that would do a similar database update for the network, but then would also actually make the Stack service calls to create the load balancer and do all the configuration that needs to be done.

hreeve-cloudera:
> You have used the access gateway in the CB API as well before. Is there any reason why these are "update load balancer" rather than "update access gateway" type of naming?

Originally, it was `add_endpoint_gateway` instead of `update_load_balancer`. But after talking to Karthik, I decided to generalize it a bit so that if we ever have need to add load balancers to an existing environment, we won't need to make a new API call. I figured if we wanted to make it clearer to the end user, we could do that on the thunderhead side. Like have a thunderhead API call that's "--add-endpoint-gateway" and it would just always call the CB `update_load_balancer` endpoint with the `publicEndpointAccessGateway=ENABLED` in the request.

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentLoadBalancerDto.java
```diff
@@ -0,0 +1,122 @@
+package com.sequenceiq.environment.environment.dto;
+
+import java.util.Set;
+
+import com.sequenceiq.cloudbreak.common.event.Payload;
+import com.sequenceiq.common.api.type.PublicEndpointAccessGateway;
+import com.sequenceiq.environment.environment.domain.Environment;
+
+public class EnvironmentLoadBalancerDto implements Payload {
```
brycederriso, line 9: **question:** It looks to me like this `Dto` is being used as a value object that is passed between Java/Spring services.

I'm used to `Dto`s as wire-level objects at some point -- will the DTO be used that way in the future?

This also assume that DTO actually stands for Data Transfer Object.

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -326,4 +328,11 @@ public EnvironmentFeatures convertToEnvironmentTelemetryFeatures(FeaturesRequest
         features.setClusterLogsCollection(featuresRequest.getClusterLogsCollection());
         return features;
     }
+
+    public EnvironmentLoadBalancerDto initLoadBalancerDto(EnvironmentLoadBalancerUpdateRequest request) {
```
brycederriso, line 332: **suggestion:** You might consider a `.from()` or `.toDto()` static method on the `EnvironmentLoadBalancerDto` or `EnvironmentLoadBalancerUpdateRequest` respectively.

sxxgrc (approved):
Looks good to me!

brycederriso (approved):
Changes look good to me. I have a question and a suggestion, but they're minor.

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
hreeve-cloudera: I generally try to avoid static methods in Cloudbreak. I know they're used some places in the code, but generally I find the code easier to work with when the methods aren't static.

core-api/src/main/java/com/sequenceiq/cloudbreak/doc/OperationDescriptions.java
```diff
@@ -59,6 +59,8 @@
         public static final String DATABASE_RESTORE_INTERNAL = "Performs a restore of the database from a provided location, internal only";
         public static final String ROTATE_CERTIFICATES = "Rotates the certificates of the cluster";
         public static final String RENEW_CERTIFICATE = "Trigger a certificate renewal on the desired cluster which is identified via stack's name";
+        public static final String UPDATE_LOAD_BALANCERS = "Updates an existing cluster with load balancers, including adding the Endpoint Gateway " +
+            "if it's enabled.";
```
orngejaket, line 63: I think this line can be removed.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
```diff
@@ -430,4 +431,11 @@ CertificatesRotationV4Response rotateAutoTlsCertificates(@PathParam("workspaceId
             notes = Notes.RENEW_CERTIFICATE_NOTES, nickname = "renewStackCertificate")
     FlowIdentifier renewCertificate(@PathParam("workspaceId") Long workspaceId, @PathParam("name") String name,
             @QueryParam("initiatorUserCrn") String initiatorUserCrn);
+
+    @PUT
+    @Path("internal/{name}/update_load_balancers")
```
orngejaket, line 436: Does this also update loadbalancers that we use for datalake ha?  If not maybe rename it to "update_public_load_balaner"

orngejaket: And corresponding new methods and classes.

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
```diff
@@ -25,6 +25,10 @@
     public static final String GET_CRN_BY_NAME = "Get the crn of an environment by name.";
     public static final String GET_NAME_BY_CRN = "Get the name of an environment by crn.";
     public static final String UPDATE_CONFIG_BY_CRN = "Update the configuration for all stacks in the Environment by the Environment CRN";
+    public static final String UPDATE_LOAD_BALANCERS_BY_ENV_NAME = "Updates all cluster in an environment with load balancers, including adding the " +
+        "Endpoint Access Gateway if it's enabled. Environment is specified by name.";
```
orngejaket, line 29: Reword a bit.
Something like: "Updates all clusters in an env with load balancers, including adding the Endpoint Access Gateway by the Environment name."

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
```diff
@@ -25,6 +25,10 @@
...
+    public static final String UPDATE_LOAD_BALANCERS_BY_ENV_CRN = "Updates all cluster in an environment with load balancers, including adding the " +
+        "Endpoint Access Gateway if it's enabled. Environment is specified by CRN.";
```
orngejaket, line 31: Similar comment tp above.

orngejaket (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentLoadBalancerDto.java
hreeve-cloudera: I've changed this object so that it no longer has the `Environment` member. I think that was the only thing that was potentially non serializable, and we have way to look it up later in the part of the code that actually needs it.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/StackV4Endpoint.java
hreeve-cloudera: Yes, it will update all load balancers depending on what entitlements you have. I deliberately designed it to be broader than just the public LB, so that if we ever had need of that functionality we wouldn't have to re-code it. I realize the title is misleading in that regard.

environment-api/src/main/java/com/sequenceiq/environment/api/doc/environment/EnvironmentOpDescription.java
hreeve-cloudera: Technically that's grammatically incorrect. The "by the Environment name" clause is in the wrong place because it looks like it applies to the Endpoint Gateway and not the environment. And I don't think we get enough out of shortening "environment" to make it worth it. I believe this is customer facing information. Unless you have a really strong opinion I'd prefer to leave it as-is.

core-api/src/main/java/com/sequenceiq/cloudbreak/doc/OperationDescriptions.java
hreeve-cloudera: It's used here: https://github.com/hortonworks/cloudbreak/pull/9940/files#diff-25a6dcfa1c517acc3b181311af4acdadb603914209d2ea7414441cdea1dee574R438

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
hreeve-cloudera: I say that and realize `builder()` is a static method :). Regardless, I understand your point, but prefer to follow existing patterns unless it's a chore to.

orngejaket (approved):

PR #9942, opened 2021-02-03
CB-11007 CCM Enabled Icon is not show in UI for CCMV2 Environments.

smaniraju:
1. UI and CDP CLI consider an environment is CCM Enabled only when TunnelValue is CCM.
2. Since Environment is CCM enabled even when CCMV2 is used internally, Environment Api Tunnel response needs to be initialized as CCM. (Customer Facing CDP CLI only has enableTunnel(False\True), internal ccmv1, ccmv2 protocols are not exposed to customers.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/credential/v1/converter/TunnelConverter.java
+++ b/environment/src/main/java/com/sequenceiq/environment/credential/v1/converter/TunnelConverter.java
@@ -17,4 +17,8 @@ public class TunnelConverter {
         }
         return tunnelRequest;
     }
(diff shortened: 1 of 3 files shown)


lacikaaa:
this doesn't look like a good idea. I mean we are saying about an env it's using CCMv1 when it's actually using v2. This should be a UI change

smaniraju:
> this doesn't look like a good idea. I mean we are saying about an env it's using CCMv1 when it's actually using v2. This should be a UI change

Ok.

PR #9943, opened 2021-02-03
CB-9193. Eliminate class java.lang.String cannot be cast to class jav…

oleewere:
…a.util.Map (java.lang.String and java.util.Map are in module java.base of loader 'bootstrap') from Salt response.

details:
- Use JsonNode objects in order to get rid of casts and instanceof operations + class cast errors
    - add more unit tests, especially invalid json values - those that causes the runtime errors
    - CB could hide cast and nullpointers in case of simple states - so keeping this by catching runtime execptions
    - change isOldSalt method to hasDataFieldInResult - as it's not related with older salt responses, many responses still contains data field in the responses

started with the tests first in order to achieve/reproduce the existing behaviours


oleewere:
keeping in draft to make sure, behaviour is fine for timeouts as well

oleewere:
opening for review, few notes:
- there are some difference between state apply and highstate, many runtime exception is caught, so im doing the same (seemed like nullpointer exceptions + cast exceptions are also handled) -> so i kept the same behaviour
- in the validation part, you can see what kind of errors we had before just with cast errors or nullpointers, so throwing readable errors for those
- also question is there any specific kind of use case that i should cover during testing (i mean negative test cases, not like upgrade/scale flow etc.)
(1 use case that was not covered  here - and because of that it's not covered here as well - if your nodes are totally empty  you got a jid argument error from salt)

lnardai (approved):
LGTM!

oleewere:
just putting back to draft as the PR is old so I will retest the use cases (as the conflict is resolved)

sodre90 (approved):

PR #9944, opened 2021-02-03
CB-10938 Decode authorization errors

Bajzathd:
When a user does not have the permission to perform the requested action AWS api may return an encoded authorization error, that can be decoded with another api call (that needs the sts:DecodeAuthorizationMessage permission). In case an encoded authorization message is detected, we will try to decode it automatically and return a human readable authorization error message. To achieve this all api calls had to be moved to their own client services, and they had to be wrapped with a proxy that can handle errors.


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/mapper/SdkClientExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
...
+    @Inject
+    private AwsEncodedAuthorizationFailureMessageDecoder awsEncodedAuthorizationFailureMessageDecoder;
+
+    public RuntimeException map(AwsCredentialView awsCredentialView, SdkClientException e) {
+        String message = awsEncodedAuthorizationFailureMessageDecoder.decodeAuthorizationFailureMessageIfNeeded(awsCredentialView, e.getMessage());
+        if (!message.equals(e.getMessage())) {
+            return new CloudConnectorException(message, e);
+        }
+        if (message.contains("Rate exceeded") || message.contains("Request limit exceeded")) {
+            return new ActionFailedException(e.getMessage());
```
lajosrodek, line 25: Nit: At this point `message.equals(e.getMessage()) == true`, so `e.getMessage()` can be simply replaced with `message`.

Bajzathd: Fixed, thanks!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/client/AmazonEc2RetryClient.java
```diff
@@ -1,52 +1,184 @@
...
-        return retry.testWith2SecDelayMax15Times(() -> mapThrottlingError(() -> client.describeVolumes(request)));
+        return retry.testWith2SecDelayMax15Times(() -> client.describeVolumes(request));
     }
 
     public AttachVolumeResult attachVolume(AttachVolumeRequest request) {
-        return retry.testWith2SecDelayMax15Times(() -> mapThrottlingError(() -> client.attachVolume(request)));
+        return retry.testWith2SecDelayMax15Times(() -> client.attachVolume(request));
+    }
+
+    public DescribeRegionsResult describeRegions(DescribeRegionsRequest describeRegionsRequest) {
```
topolyai5, line 92: these methods are not retryable. Are they should retryable or maybe create another class for these methods

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsEncodedAuthorizationFailureMessageDecoder.java
```diff
@@ -0,0 +1,63 @@
...
+            } catch (AWSSecurityTokenServiceException e) {
+                if ("AccessDenied".equals(e.getErrorCode())) {
+                    result = message.replaceAll(matcher.group(0), "(Please add sts:DecodeAuthorizationMessage right to your IAM policy to get more details.)");
+                } else {
+                    LOGGER.error("Failed to decode authorization failure message", e);
+                }
+            } catch (Exception e) {
+                LOGGER.error("Failed to decode authorization failure message", e);
+            }
+        }
```
topolyai5, line 46: please add an else branch, if find is false

topolyai5 (changes requested):
Please check all `@VisibleForTesting`. Either public or `@VisibleForTesting`

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/mapper/SdkClientExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
...
+import com.sequenceiq.cloudbreak.cloud.exception.CloudConnectorException;
+import com.sequenceiq.cloudbreak.service.Retry.ActionFailedException;
+
+@Component
+public class SdkClientExceptionMapper {
+
+    @Inject
+    private AwsEncodedAuthorizationFailureMessageDecoder awsEncodedAuthorizationFailureMessageDecoder;
+
+    public RuntimeException map(AwsCredentialView awsCredentialView, String region, SdkClientException e) {
```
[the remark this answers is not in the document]
Bajzathd: I am only using already existing exception types, and in the service layer, so it should not cause any troubles

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/client/AmazonEc2RetryClient.java
Bajzathd: I would prefer to rename the service, removing the retry from the name as it does not really matter from the point of the using class

config/checkstyle/checkstyle.xml
```diff
@@ -119,6 +119,10 @@
         <module name="SuperClone" />
         <module name="MutableException" />
         <module name="AvoidStarImport" />
+        <module name="ImportControl">
+            <property name="file" value="${config_loc}/import-control.xml"/>
+            <property name="path" value="cloud-aws/src/main/java.*$"/>
```
topolyai5, line 124: what do this?

Bajzathd: It is used to check that no service uses aws api without wrapping it in a client class - so that the error message mapping will be applied to every call

topolyai5 (approved):

topolyai5:
I have tried it locally but looks like this does not work existed error message. Is it a big change if you add to old messages?

pkedvessy (approved):

lajosrodek:
Thanks, nice work! 👍

PR #9947, opened 2021-02-03
CB-11009. Add disk and memory data to cdp doctor report.

oleewere:

### DIFF
--- a/orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cdp-doctor-commands.yaml.j2
+++ b/orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cdp-doctor-commands.yaml.j2
@@ -6,6 +6,14 @@ commands:
       report: false
     - command: netstat -tlpn
       output: /tmp/doctor_netstat.txt
+    - command: free -h
+      output: /tmp/doctor_free.txt
+    - command: ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%mem | head
+      output: /tmp/doctor_top_proc.txt
+    - command: df -h
+      output: /tmp/doctor_disk_free.txt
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/cdp-doctor-commands.yaml.j2
```diff
@@ -6,6 +6,14 @@ commands:
       report: false
     - command: netstat -tlpn
       output: /tmp/doctor_netstat.txt
+    - command: free -h
+      output: /tmp/doctor_free.txt
+    - command: ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%mem | head
+      output: /tmp/doctor_top_proc.txt
+    - command: df -h
+      output: /tmp/doctor_disk_full.txt
```
pdarvasi, line 14: why is the name dist_full.txt? df is short for disk free

oleewere: it's just accidental...so i have created one check for only the **logs** and one for the **full** filesystem based on the command name. can be changed if you think disk free as name is better (to be alias of the command)

oleewere: whatever, fixed it to not confuse anyone by the naming

attilapalfi92 (approved):

biharitomi:
@oleewere to check volume related problems could you please add the `lsblk` command output to collected results?

oleewere:
@biharitomi ..if you write it 1 minute earlier, almost:) ... i can create a follow up PR

biharitomi:
Thanks @oleewere , that will be nice to have that way too.

PR #9948, opened 2021-02-03
CB-10851 Indicate for SDX if CM is unreachable

sodre90:
now CB syncer checks if cm is running, if not, then exception is catched in CM SDX, syncer will sync instances from provider side, and if provider side is reachable, nothing will happen thus user won't get the information about that CM is unreachable from CDP (from example in case of CCM clusters)

in this case, every dh creation will fail while getting sdx context from CM of SDX, but if SDX CM is unreachable from CDP, any dh creation attempt is unnecessary


core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/StackSyncService.java
```diff
@@ -226,11 +233,21 @@ private void syncDeletedInstance(Stack stack, Map<InstanceSyncState, Integer> in
         }
     }
 
-    private void handleSyncResult(Stack stack, Map<InstanceSyncState, Integer> instanceStateCounts, boolean stackStatusUpdateEnabled) {
+    private void handleSyncResult(Stack stack, Map<InstanceSyncState, Integer> instanceStateCounts, boolean stackStatusUpdateEnabled, boolean cmServerRunning) {
```
lnardai, line 236: Getting quite a lot of control booleans. We could extract this to a SyncConfig class or something similar.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/status/SdxStatusService.java
```diff
@@ -50,22 +51,29 @@
...
         setStatusForDatalake(status, statusReason, sdxCluster);
-        sdxNotificationService.send(status.getDefaultResourceEvent(), sdxCluster);
+        sdxNotificationService.send(status.getDefaultResourceEvent(), Collections.singleton(sdxCluster.getClusterName()), sdxCluster);
     }
 
     public SdxCluster setStatusForDatalakeAndNotify(DatalakeStatusEnum status, String statusReason, Long datalakeId) {
         return sdxClusterRepository.findById(datalakeId).map(cluster -> {
             setStatusForDatalake(status, statusReason, cluster);
-            sdxNotificationService.send(status.getDefaultResourceEvent(), cluster);
+            sdxNotificationService.send(status.getDefaultResourceEvent(), Collections.singleton(cluster.getClusterName()), cluster);
```
lnardai, line 60: Isn't this information is inside the cluster object?

lnardai (approved):

PR #9949, opened 2021-02-03
CB-10992 LIVY_FOR_SPARK3_API mapping to ui topology was wrong. REcrea…

doktoric:
…ted the block. We investigated with @biharitomi and @bergerdenes, @keyki it seems some kind of wrong character was there so just deletet and created the target again.


bergerdenes (approved):
(not so) funny.

PR #9950, opened 2021-02-03
CB-10991 Allow to edit on Azure and GCP.

topolyai5:

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/dto/AuthenticationDtoConverter.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/dto/AuthenticationDtoConverter.java
@@ -21,12 +21,9 @@ public class AuthenticationDtoConverter {
 
     public EnvironmentAuthentication dtoToAuthentication(AuthenticationDto authenticationDto) {
         EnvironmentAuthentication environmentAuthentication = new EnvironmentAuthentication();
-        if (StringUtils.isNotEmpty(authenticationDto.getPublicKey())) {
-            ValidationResult validationResult = publicKeyValidator.validatePublicKey(authenticationDto.getPublicKey());
-            if (!validationResult.hasError()) {
-                List<String> parts = Arrays.asList(StringUtils.split(authenticationDto.getPublicKey(), " "));
-                environmentAuthentication.setPublicKey(String.format("%s %s %s", parts.get(0), parts.get(1), authenticationDto.getLoginUserName()));
-            }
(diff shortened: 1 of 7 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/dto/AuthenticationDtoConverter.java
```diff
@@ -21,12 +21,9 @@ public AuthenticationDtoConverter(PublicKeyValidator publicKeyValidator) {
 
     public EnvironmentAuthentication dtoToAuthentication(AuthenticationDto authenticationDto) {
         EnvironmentAuthentication environmentAuthentication = new EnvironmentAuthentication();
-        if (StringUtils.isNotEmpty(authenticationDto.getPublicKey())) {
-            ValidationResult validationResult = publicKeyValidator.validatePublicKey(authenticationDto.getPublicKey());
-            if (!validationResult.hasError()) {
-                List<String> parts = Arrays.asList(StringUtils.split(authenticationDto.getPublicKey(), " "));
-                environmentAuthentication.setPublicKey(String.format("%s %s %s", parts.get(0), parts.get(1), authenticationDto.getLoginUserName()));
-            }
+        if (isValidSshKey(authenticationDto.getPublicKey())) {
```
doktoric, line 24: dont we need to check first that the public key exists?

environment/src/main/resources/application.yml
```diff
@@ -101,6 +101,16 @@ environment:
   enabledplatforms: AWS,AZURE,YARN,GCP
   enabledParentPlatforms: AWS,MOCK
   enabledChildPlatforms: YARN,MOCK
+  existing:
+    ssh:
+      key:
+        update:
+          support: AWS,MOCK
```
doktoric, line 108: ycloud ?

doktoric: what happens with if we add a new provider but that is not here

topolyai5: We won't update the raw public key or public key id.

environment/src/main/java/com/sequenceiq/environment/environment/dto/AuthenticationDtoConverter.java
topolyai5: it is the raw public key. We want to validate the format here. However, I moved the conditions onto a method. The conditions are the same than previously.

doktoric (approved):

PR #9957, opened 2021-02-03
CB-11020. Metering heartbeat versioning is broken.

oleewere:

it only uses version 1 or 2, not any newer one, version 3 is already used from 2020 dec 16, that means newer images uses the older metering streams instead of the new ones


vaibhavgandhi12:
LGTM! 🚢 should be granted by CB team.

vaibhavgandhi12 (approved):
Forgot to approve in the previous comment. Please wait for final approval from CB devs.

oleewere:
the change is straightforward as that is copied from fluent/settings.sls or telemetry/settings.sls, so merging this

PR #9958, opened 2021-02-03
CB-11025 - [Knox] Increase replayBufferSize value from 64 to 128 for Ranger service

moresandeep:
This change increases the replayBufferSize value from 64 to 128 for Ranger service. This change affects topologies cdp-proxy.xml and cdp-proxy-api.xml.


moresandeep:
Thanks @keyki !

PR #9960, opened 2021-02-03
SMON issues

doktoric:

### DIFF
--- a/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/config/SMONMemoryWorkaroundService.java
+++ b/cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/config/SMONMemoryWorkaroundService.java
@@ -11,7 +11,16 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.common.StackType;
 @Service
 public class SMONMemoryWorkaroundService {
 
-    private static final long GB_IN_BYTE = 1073741824;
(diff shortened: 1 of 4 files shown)


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/config/SMONMemoryWorkaroundService.java
```diff
@@ -48,47 +57,78 @@
...
+        if (numberOfNodes <= smonSmallClusterMaxSize) {
+            minMemoryInGig = 1;
+        } else if (numberOfNodes <= smonMediumClusterMaxSize) {
+            minMemoryInGig = 2;
+        } else if (numberOfNodes <= smonLargeClusterMaxSize) {
+            minMemoryInGig = 7;
+        } else {
+            minMemoryInGig = 11;
+        }
+        return getAtLeastMinMemoryInGig(memoryInGig, minMemoryInGig);
```
bergerdenes, line 108: Math.max(memoryInGig, minMemoryInGig);

bergerdenes (approved):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/config/SMONMemoryWorkaroundService.java
doktoric: thanks

PR #9968, opened 2021-02-04
CB-11017 In case of GCP, If we don't send the volume type, we set HDD…

topolyai5:
… as default but this type doesn't exist and this leads to an NPE. Validation added and HDD added to the diskTypes list


topolyai5:
If someine try to create a GCP datahub without volume type, we convert it to HDD. But this value doesn't exist and we should add the HDD to the mapping list.

biharitomi (approved):

PR #9972, opened 2021-02-05
CDPCP-3685 Update usermanagement.proto and references to account pass…

gitmfox:
…word policy

This commit updates usermanagement.proto to be in sync with the
version UMS is running in mow-dev. The motivation for the update
is to bring in changes needed to optimize password hash update
during user sync.

Since the last time the proto was updated, a compatibility-breaking
change was made to the Account message: the 'passwordPolicy' field
was replaced by 'globalPasswordPolicy' and 'machineUserPasswordPolicy'.
The 'passwordPolicy' field number was reused for 'globalPasswordPolicy',
which means that for the last two months or so, Cloudbreak was getting
the global password policy when it accessed 'passwordPolicy'. If any
customers set a machine user-specific policy, it was ignored.

This commit updates the references to the password policy fields,
and implements the semantics of the machine user policy: if present,
it is used for machine users; otherwise, the global policy is used,
if present.

### DIFF
--- a/auth-connector/src/generated/main/grpc/com/cloudera/thunderhead/service/usermanagement/UserManagementGrpc.java
+++ b/auth-connector/src/generated/main/grpc/com/cloudera/thunderhead/service/usermanagement/UserManagementGrpc.java
@@ -2881,6 +2881,43 @@ public final class UserManagementGrpc {
      return getProcessWorkloadSSOAuthnReqMethod;
   }
   @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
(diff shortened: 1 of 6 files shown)


gitmfox:
@lacikaaa, @handavid: this is the same change as https://github.com/hortonworks/cloudbreak/pull/9933 (including review feedback), which you both have already reviewed and approved. I closed the earlier PR because I thought it might have gotten into a strange state when I squashed the commits.

I hope you don't mind giving this one a quick look. Sorry to have created additional work for you.

lacikaaa (approved):

PR #9973, opened 2021-02-05
CB-11040: Fix race condition which intermittently corrupts CS.cfg

jamisonbennett:
When pki-tomcatd@pki-tomcat starts, it always rewrites the config file
/var/lib/pki/pki-tomcat/ca/conf/CS.cfg. It even does this if the file
does not change. There is a delay of about 8 seconds between when the
service command returns and when the file is rewritten. This can cause
the file to become corrupted if any other process modifies the file
during this operation. It is recommended that the CA be shutdown
before editing this file and then the CA can be started again. This
fix was done.

This method was tested with a loop that modified the file many times
between service restarts. Without the fix, the CA failed almost every
time. With this fix, I was unable to get it to fail.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
@@ -10,18 +10,12 @@ update_cnames:
         - file: /opt/salt/scripts/update_cnames.sh
 
 one_week_next_update_grace_period:
-  file.replace:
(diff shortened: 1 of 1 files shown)


lacikaaa:
I think we should move this to 2.38

keyki:
We discussed where we should merge this PR, @jamisonbennett  can you please change the base version to CB-2.38.0?

lacikaaa (approved):

PR #9981, opened 2021-02-05
Sending a more events to EDH

akanto:

CB-11067 Support stop, start, upscale, downscale and upgrade reporting to EDH. There are some incompatible changes in the usage.proto, but since the usage.proto was not promoted to thunderhead yet, this shaould not cause any problem, but it will reduce the maintenance cost in the future. (HEAD -> edh-improve, origin/edh-improve) [Attila Kanto]

CB-11046 Provide more details about which flow step has failed. This is particularly important in case of Datahub and Datalake since they are created in flowchains and it can have multiple exists. [Attila Kanto]

### DIFF
--- a/structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/log/cluster/ClusterRequestedLogger.java
+++ b/structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/log/cluster/ClusterRequestedLogger.java
@@ -3,6 +3,12 @@ package com.sequenceiq.cloudbreak.structuredevent.service.telemetry.log.cluster;
 import static com.cloudera.thunderhead.service.common.usage.UsageProto.CDPClusterStatus.Value.CREATE_FAILED;
 import static com.cloudera.thunderhead.service.common.usage.UsageProto.CDPClusterStatus.Value.CREATE_FINISHED;
 import static com.cloudera.thunderhead.service.common.usage.UsageProto.CDPClusterStatus.Value.CREATE_STARTED;
+import static com.cloudera.thunderhead.service.common.usage.UsageProto.CDPClusterStatus.Value.DOWNSCALE_FAILED;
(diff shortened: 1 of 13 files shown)


tiborm (approved):
Looks good by observation. +1

structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/log/cluster/ClusterRequestedLogger.java
```diff
@@ -26,7 +32,10 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClusterRequestedLogger.class);
 
-    private static final EnumSet<UsageProto.CDPClusterStatus.Value> TRIGGER_CASES = EnumSet.of(CREATE_STARTED, CREATE_FINISHED, CREATE_FAILED);
+    private static final EnumSet<UsageProto.CDPClusterStatus.Value> TRIGGER_CASES = EnumSet.of(CREATE_STARTED, CREATE_FINISHED, CREATE_FAILED,
```
nadamdb, line 35: Are the RESUME/SUSPEND _STARTED/_FINISHED/_FAILED and the UPGRADE/UPSCALE/DOWNSCALE _STARTED  use cases intentionally excluded from the trigger cases?

nadamdb (approved):
Looks good to me too, I only have one question regarding the trigger cases in ClusterRequestedLogger (see in comment). If that behaviour is intentional then +1

oleewere (approved):
LGTM, I can merge it after Adam's question is answered or fixed

structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/log/cluster/ClusterRequestedLogger.java
akanto: It is intentional. I am adding only those events that are mutating the DL/DH values.  The start/stop and upgrade started, scaling started events are changing only the state, therefore they are logged only to the state changed table.

akanto:
Merging, since 3 approve has been received.

PR #9983, opened 2021-02-06
CB-11072: Avoid the need to stop ranger and HMS services before performing restore.

kkalvagadda1:
While performing backup and restore we need to make sure that there are no updates to metadata to make sure that we have a consistent backup. Currently, it is documented but there is no way to enforce it.

This approach will make sure that HMS and Ranger's data is updated while backup-restore operations are performed on data lake.

This change basically does below

1. Limit the number of connections to the database to 0.
2. Close all the active connections to the database.
3. perform the restore operation.
4. Limit the number of connections to the database to -1 (unlimited)

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
@@ -89,6 +89,19 @@ replace_ranger_group_before_import() {
   fi
 }
 
+close_existing_connections() {
+  SERVICE=$1
+  doLog "INFO Closing existing connections to ${SERVICE}"
+  psql --host="$HOST" --port="$PORT" --dbname="postgres" --username="$USERNAME" -c "SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity WHERE pg_stat_activity.datname = '${SERVICE}' AND pid <> pg_backend_pid();" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || errorExit "Unable to close connections to ${SERVICE}"
+}
+
+limit_incomming_connection() {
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -112,7 +129,7 @@ run_restore() {
   then
     STRIPPED_S3_URL=$(echo "$BACKUP_LOCATION" | sed -e 's/[a-fA-F0-9]\{8\}-[a-fA-F0-9]\{4\}-[a-fA-F0-9]\{4\}-[a-fA-F0-9]\{4\}-[a-fA-F0-9]\{12\}_database_backup\/*\*\?$//g')
     doLog "Begining to perform s3guard import for ${STRIPPED_S3_URL}"
-    hadoop s3guard import "$STRIPPED_S3_URL" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || doLog "ERROR Unable to import S3Guard metadata."
+ #   hadoop s3guard import "$STRIPPED_S3_URL" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || doLog "ERROR Unable to import S3Guard metadata."
```
keyki, line 132: Is this line commented out on purpose?

kkalvagadda1: Good catch. Nope. I just commented it as it was taking time in my testing. I will update the patch.

kkalvagadda1: addressed it.

brycederriso (approved):
Changes look good to me.

PR #9984, opened 2021-02-06
CB-10672 Fix FreeIPA HA (repair) in GCP (take 4)

cegganesh84:
1. Found this during FreeIPA HA full repair, although it could have happened on an install too.
2. Sometimes it happens that ipaserver0 was assigned ipaserver1 hostname.
3. This breaks the replica install with this bug https://bugzilla.redhat.com/show_bug.cgi?id=1364139
4. There is no fix for the above bug in a VM install.
5. Nonetheless this is a nasty bug, that needs fixing anyway. Elsewhere in the code, there is an assumption that server 0 gets ipaserver0 hostname.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/InstanceGroupRequestToInstanceGroupConverter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/InstanceGroupRequestToInstanceGroupConverter.java
@@ -1,6 +1,6 @@
 package com.sequenceiq.freeipa.converter.instance;
 
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 import javax.inject.Inject;
@@ -45,7 +45,7 @@ public class InstanceGroupRequestToInstanceGroupConverter {
     }
 
     private void addInstanceMetadatas(InstanceGroupRequest request, InstanceGroup instanceGroup, String hostname, String domain) {
-        Set<InstanceMetaData> instanceMetaDataSet = new HashSet<>();
+        Set<InstanceMetaData> instanceMetaDataSet = new LinkedHashSet<>();
         for (int i = 0; i < request.getNodeCount(); i++) {
             InstanceMetaData instanceMetaData = new InstanceMetaData();
             instanceMetaData.setInstanceGroup(instanceGroup);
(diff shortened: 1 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/InstanceGroupRequestToInstanceGroupConverter.java
```diff
@@ -45,7 +45,7 @@ public InstanceGroup convert(InstanceGroupRequest source, String accountId, Stri
     }
 
     private void addInstanceMetadatas(InstanceGroupRequest request, InstanceGroup instanceGroup, String hostname, String domain) {
-        Set<InstanceMetaData> instanceMetaDataSet = new HashSet<>();
+        Set<InstanceMetaData> instanceMetaDataSet = new LinkedHashSet<>();
```
akanto, line 48: Can you add a unit test, please, that verifies that instances have the right order?

jamisonbennett (commented):
I agree that this is a good change to make. I am thinking this PR might have been posted prior to our discussion we had last night about the HA restore procedure. So these comments are mostly for other people that maintain this in the future.


The restore from backup procedure takes into account lots of corner cases of cleaning up the old host information, so this change probably is a necessity but it does make things more deterministic at initial install so its an improvement. The HA restore form backup procedure clears out all of the old ipaservers then it cycles out all of the fresh VMs for a new set. The recovery still needs to handle all of these corner cases. We may have to restore from a backup from any node. E.g. restore ipaserver0, ipaserver1, ipaserver2, ipaserver3, or even ipaserver25 (extreme case with lots of prior repairs). When these get restored we don't necessarily restore to a new ipaserver with the same number. The restore procedure takes this into account. We also cycle out all the original servers to handle the cases where there are CA/CRL masters that get changed (or duplicated in the case of the CRL master). Even with this change forcing ipaserver0 to be the initial primary gateway, we can't assume that ipaserver0 is always the CA master (I think it will be the CRL master at least until a repair is performed). The CA master can change during any operation that changes the topology (e.g. replica install, restore backup, repair, etc.). These types of behaviors are sometimes not initially obvious but only become a problem after time (e.g. 4 hours for a duplicate CRL master, 1 week for a missing CRL master, months later a repair can fail if we don't cleanup something like ipaserver3/ipaserver25 from the restored backup).

freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/InstanceGroupRequestToInstanceGroupConverter.java
cegganesh84: Sorry about that. Added a unit test.

cegganesh84:
> I agree that this is a good change to make. I am thinking this PR might have been posted prior to our discussion we had last night about the HA restore procedure. So these comments are mostly for other people that maintain this in the future.
> 

@jamisonbennett Just wanted to point one thing that I missed conveying, this is a GCP specific issue. GCP resets the hostname if it is not explicitly during creation. So we use these FQDNs only in GCP and only for FreeIPA nodes. So the ordering of the hostnames is a must-do in GCP for the FreeIPA nodes in HA mode.

I agree with rest of your points in the recovery mode.

jamisonbennett (approved):

jamisonbennett:
@cegganesh84 Thanks for the clarification about this being a GCP specific issue. I didn't realize how GCP works for hostname assignment.

PR #9988, opened 2021-02-07
CB-11006 Fix the disk mounting logic to support Azure VM types withou…

daszabo:
…t resource/temporary disk

The mount and format scripts, in case of Azure, now dynamically determine the root, resource and managed disks.
Also reused the device names field of volumesets for Logical Unit Numbers, needed for the script to
properly pair devices to disks.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/bootstrapnodes/format-and-mount-common.sh
+++ b/orchestrator-salt/src/main/resources/salt/bootstrapnodes/format-and-mount-common.sh
@@ -83,6 +83,8 @@ get_root_disk() {
     root_partition=$(lsblk | grep /$ | cut -f1 -d' ' )
     if [[ $root_partition =~ "nvme" ]]; then
         echo "/dev/$(lsblk | grep /$ | cut -f1 -d' ' | sed 's/p\w//g' | cut -c 3-)"
+    elif [[ $CLOUD_PLATFORM -eq AZURE ]]
+        echo "/dev/$(ls -l /dev/disk/azure/ | awk '{print $9"\t"$11}' | grep -E '(root\s)' | cut -f2 | cut -f3 -d'/')"
     else
         echo "/dev/$(lsblk | grep /$ | cut -f1 -d' ' | sed 's/[0-9]//g' | cut -c 3-)"
     fi
--- a/orchestrator-salt/src/main/resources/salt/salt/disks/mount/scripts/find-device-and-format.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/disks/mount/scripts/find-device-and-format.j2
@@ -10,7 +10,8 @@ LOG_FILE=/var/log/find_device_and_format.log
 VERSION="V1.0"
 
 # INPUT - expected lists
-#   ATTACHED_VOLUME_NAME_LIST - contains a list of volume device names attached to the instance, format: space separated list
+#   ATTACHED_VOLUME_NAME_LIST - contains a list of volume device names attached to the instance, format: space separated list,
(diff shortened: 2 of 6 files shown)


orchestrator-salt/src/main/resources/salt/salt/disks/mount/scripts/find-device-and-format.j2
```diff
@@ -104,6 +104,17 @@ save_env_vars_to_log_file() {
     log $LOG_FILE CLOUD_PLATFORM=$CLOUD_PLATFORM
 }
 
+get_temporary_disk_device_name() {
+  if [[ $CLOUD_PLATFORM -eq AZURE ]]; then
+    temporary_disk="/dev/$(ls -l /dev/disk/azure/ | awk '{print $9"\t"$11}' | grep -E '(resource[^-])' | cut -f2 | cut -f3 -d'/')"
+    log $LOG_FILE found temprary disk: \"$temporary_disk\"
```
lajosrodek, line 110: Typo: temprary -> temporary.

orchestrator-salt/src/main/resources/salt/salt/disks/mount/scripts/find-device-and-format.j2
```diff
@@ -104,6 +104,17 @@ save_env_vars_to_log_file() {
     log $LOG_FILE CLOUD_PLATFORM=$CLOUD_PLATFORM
 }
 
+get_temporary_disk_device_name() {
+  if [[ $CLOUD_PLATFORM -eq AZURE ]]; then
+    temporary_disk="/dev/$(ls -l /dev/disk/azure/ | awk '{print $9"\t"$11}' | grep -E '(resource[^-])' | cut -f2 | cut -f3 -d'/')"
```
gergopapi2, line 109: Question: what happens if there is no temporary disk? I think then the value of temporary_disk would be "/dev".
On AWS the return value is empty if there are no ephemeral disks. Will this difference cause no problems in the mounting? Will the script not want to mount the "/dev"?

gergopapi2 (commented):
A general question at ephemerals, otherwise LGTM.

orchestrator-salt/src/main/resources/salt/bootstrapnodes/format-and-mount-common.sh
```diff
@@ -83,6 +83,8 @@ get_root_disk() {
     root_partition=$(lsblk | grep /$ | cut -f1 -d' ' )
     if [[ $root_partition =~ "nvme" ]]; then
         echo "/dev/$(lsblk | grep /$ | cut -f1 -d' ' | sed 's/p\w//g' | cut -c 3-)"
+    elif [[ $CLOUD_PLATFORM -eq AZURE]]
```
pdarvasi, line 86: AZURE]] --> pls add extra whitespace

pdarvasi (approved):

gergopapi2:
A comment that came up several time and can be addressed as a separate work item, I add it here not to forget: it would be great to have a test that would  test all instance types and just check if the scripts did work for them. This test could be potentially costly so it would be enough to run it once a week or so.

biharitomi:
> A remark that came up several time and can be addressed as a separate work item, I add it here in order to not to forget: it would be great to have a test that would test all instance types and just check if the scripts did work for them. This test could be potentially costly so it would be enough to run it once a week or so.

I think you should create a JIRA item to cover it.

biharitomi (approved):
LGTM

biharitomi (approved):
LGTM

biharitomi (approved):
LGTM

sodre90 (approved):

PR #9989, opened 2021-02-07
CB-10983 Move Experience delete flow step into clusters deletion flow

bergerdenes:


environment/src/test/java/com/sequenceiq/environment/experience/liftie/LiftieConnectorListClustersTest.java
```diff
@@ -70,71 +76,41 @@ void testInvocationBuilderProviderShouldCreateCallForExecutionBasedOnTheWebTarge
...
-        ListClustersResponse result = getUnderTest().listClusters(TEST_ENV_NAME, TEST_TENANT, null, null);
+        IllegalStateException thrown = assertThrows(IllegalStateException.class, () -> getUnderTest().listClusters(TEST_ENV_NAME, TEST_TENANT, null, null));
 
-        assertNotNull(result);
-
-        assertNull(result.getClusters());
-
-        assertNotNull(result.getPage());
-        assertEquals(0, result.getPage().getTotalPages());
+        assertTrue(thrown.getCause() instanceof RuntimeException);
```
gregito, line 104: assertThrows will throw its exception when not the expected type of exception comes at the result of the executed method, therefore here one should make the assertion on the message instead of the actual type.
Also since the "thrown" field is already a subclass of the RuntimeException class this assertion has no point in the first place

gregito: The same goes for the other places

gregito (changes requested):

environment/src/test/java/com/sequenceiq/environment/experience/liftie/LiftieConnectorListClustersTest.java
bergerdenes: Fixed

gregito (approved):

PR #9991, opened 2021-02-08
CB-11078 Environment FAILED_STATUSES set does not contains all the related EnvironmentStatus

aszegedi:
At `com.sequenceiq.it.cloudbreak.testcase.e2e.environment.EnvironmentStopStartTests.testCreateStopStartEnvironment` the stopped environment cannot get into `AVAILABLE` state. However test is waiting for this state even the `START_FREEIPA_FAILED` environment status draws up.

So we need to update the related failed statuses set at `EnvironmentStatus` class.


biharitomi (approved):
LGTM

PR #9995, opened 2021-02-08
Merge 2.37 into 2.38

biharitomi:


keyki:
I don't think we should merge this yet as it breaks the Azure repair tests.

biharitomi:
> I don't think we should merge this yet as it breaks the Azure repair tests.

I think maybe one iteration is affected as the fix is on the way. Right? @sodre90 
@keyki This required almost an hour of merge conflict resolving and multiple colleges have been part of it, so I would not like to re-resolve conflicts with merging real logic.

keyki:
> > I don't think we should merge this yet as it breaks the Azure repair tests.
> 
> I think maybe one iteration is affected as the fix is on the way. Right? @sodre90
> @keyki This required almost an hour of merge conflict resolving and multiple colleges have been part of it, so I would not like to re-resolve conflicts with merging real logic.

Noted. We can merge then.

sodre90 (approved):

PR #10004, opened 2021-02-08
Revert "CB-10578 Populate new targetgroup_loadbalancer table from CB-10095"

hreeve-cloudera:
This reverts commit 1b8ba026e741a6be56692498b48f0e71c5268f5b due to an error in
the sql script.


hreeve-cloudera:
I'm closing this because I realized the error was caused by an invalid database entry that had been made during my testing. This shouldn't impact any other environments.

PR #10008, opened 2021-02-09
CB-8529 Starting Hue fails with longer stackname. Validating blueprin…

doktoric:
…t groups, generated endpoints and stackname. The validation will fails:

1. IF distrox_cluster_name.length > 20 AND cm_Template contains HUE
2. IF distrox_template_hostgroup_name.length > 6 AND hostgroup contains HUE
3. generated_hostname.length > 63 AND hostgroup contains HUE


akanto (approved):

akanto:
Looks good, many thanks.

PR #10012, opened 2021-02-09
CB-11122. Do not check cdpTelemetryVersion (role) in case of updatePa…

oleewere:
…ckage parameter is provided.

details:
if updatePackage is used, cdp-telemetry binari is updated to the latest available version (non-snapshot), if that's the case, makes no sense to not run certain part of the code that are version specific


aakanser (approved):
looks good to me

PR #10017, opened 2021-02-10
CB-10566 Endpoints for collecting used images

Bajzathd:
Cloudbreak and Freeipa endpoints for collecting the images that are in use. These can and will be used for runtime statistics and cleaning up unused images from the catalogs and cloud provider storage to save on cloud costs.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/UtilV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/UtilV4Controller.java
@@ -10,6 +10,7 @@ import org.springframework.stereotype.Controller;
 
 import com.sequenceiq.authorization.annotation.CheckPermissionByAccount;
 import com.sequenceiq.authorization.annotation.DisableCheckPermissions;
+import com.sequenceiq.authorization.annotation.InternalOnly;
 import com.sequenceiq.authorization.resource.AuthorizationResourceAction;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.util.UtilV4Endpoint;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.util.requests.RenewCertificateV4Request;
@@ -21,6 +22,7 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.util.responses.ResourceEventRes
 import com.sequenceiq.cloudbreak.api.endpoint.v4.util.responses.SecurityRulesV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.util.responses.StackMatrixV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.util.responses.SupportedExternalDatabaseServiceEntryV4Response;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.util.responses.UsedImagesListV4Response;
 import com.sequenceiq.cloudbreak.api.util.ConverterUtil;
 import com.sequenceiq.cloudbreak.common.user.CloudbreakUser;
 import com.sequenceiq.cloudbreak.event.ResourceEvent;
@@ -29,6 +31,7 @@ import com.sequenceiq.cloudbreak.service.StackMatrixService;
 import com.sequenceiq.cloudbreak.service.account.PreferencesService;
 import com.sequenceiq.cloudbreak.service.cluster.RepositoryConfigValidationService;
 import com.sequenceiq.cloudbreak.service.filesystem.FileSystemSupportMatrixService;
+import com.sequenceiq.cloudbreak.service.image.UsedImagesProvider;
 import com.sequenceiq.cloudbreak.service.securityrule.SecurityRuleService;
 import com.sequenceiq.cloudbreak.service.stack.flow.StackOperationService;
 import com.sequenceiq.cloudbreak.structuredevent.CloudbreakRestRequestThreadLocalService;
@@ -65,6 +68,9 @@ public class UtilV4Controller extends NotificationController implements UtilV4En
 
     @Inject
     private StackOperationService stackOperationService;
+
+    @Inject
+    private UsedImagesProvider usedImagesProvider;
 
     @Value("${info.app.version:}")
     private String cbVersion;
@@ -129,4 +135,9 @@ public class UtilV4Controller extends NotificationController implements UtilV4En
         return Response.ok().build();
     }
 
+    @Override
+    @InternalOnly
+    public UsedImagesListV4Response usedImages() {
+        return usedImagesProvider.getUsedImages();
+    }
 }
--- a/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
@@ -18,6 +18,7 @@ import com.sequenceiq.authorization.service.list.AuthorizationResource;
 import com.sequenceiq.authorization.service.model.projection.ResourceCrnAndNameView;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.StackType;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status;
+import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.domain.Network;
 import com.sequenceiq.cloudbreak.domain.projection.AutoscaleStack;
 import com.sequenceiq.cloudbreak.domain.projection.StackClusterStatusView;
@@ -106,6 +107,10 @@ public interface StackRepository extends WorkspaceResourceRepository<Stack, Long
     @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig "
             + "WHERE s.terminated = null AND (s.type is not 'TEMPLATE' OR s.type is null)")
     Set<Stack> findAllAliveWithInstanceGroups();
+
+    @Query("SELECT im.image FROM Stack s LEFT JOIN s.instanceGroups ig LEFT JOIN ig.instanceMetaData im "
+            + "WHERE s.terminated = null AND (s.type is not 'TEMPLATE' OR s.type is null) AND im.image is not null")
+    List<Json> findImagesOfAliveStacks();
 
     @Query("SELECT s.id as id, s.name as name, s.stackStatus as status FROM Stack s "
             + "WHERE s.stackStatus.status IN :statuses AND (s.type is not 'TEMPLATE' OR s.type is null)")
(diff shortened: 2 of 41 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/UtilV4Controller.java
```diff
@@ -129,4 +135,9 @@ public Response renewCertificate(RenewCertificateV4Request renewCertificateV4Req
         return Response.ok().build();
     }
 
+    @Override
+    @DisableCheckPermissions
```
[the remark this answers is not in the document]
Bajzathd: I am discussing it with @lnardai on how to implement authorization, and he will review it
It was only opened as a draft PR so that @pkedvessy can help with some issues I have at the moment

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/UsedImagesTest.java
```diff
@@ -0,0 +1,171 @@
...
+        testContext
+                .as(internalActor)
+                .given(UsedImagesTestDto.class)
+                .when(utilTestClient.usedImages())
+                .then((tc, usedImagesTestDto, client) -> {
+                    List<UsedImageStacksV4Response> usedImages = usedImagesTestDto.getResponse().getUsedImages();
+                    UsedImageStacksV4Response usedImageStacksV4Response = usedImages.stream()
+                            .filter(usedImage -> usedImage.getImage().getImageId().contains(SDX_IMAGE_UUID))
+                            .findFirst().orElseThrow(() -> new TestFailException(String.format("SDX image is NOT in use with ID:: %s", SDX_IMAGE_UUID)));
+                    LOGGER.info("Used SDX image ID:: {}", usedImageStacksV4Response.getImage().getImageId());
```
aszegedi, line 125: TestFailException is needed, because of `Assert` just generate an error, these errors won't appear in the test results. So test can show as passed even the related assertions are failing. However test context exception map is gathering all the exception during test and finally these are shown in the test log and test results as well.

aszegedi (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/UsedImagesTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+                .given(EnvironmentTestDto.class)
+                .when(environmentTestClient.delete())
+                .await(EnvironmentStatus.ARCHIVED)
+
+                .then((tc, testDto, client) -> verifySdxImageIsNotUsed(tc, testDto, internalActor))
+                .validate();
+    }
+
+    private <T extends CloudbreakTestDto> T verifyImagesAreUsed(TestContext testContext, T testDto, CloudbreakUser internalActor) {
```
aszegedi, line 120: Some special validation would be useful here for the case: `this can be called only with internal actor only.`. So here would introduce some `whenException` step, where the authorisation error is expected for describe action with non internal user.

aszegedi (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/UsedImagesTest.java
Bajzathd: Great idea, I will add it, thanks!

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/FreeIpaSyncTest.java
```diff
@@ -47,7 +47,7 @@ public void testSyncFreeIpaWithInternalActor(MockedTestContext testContext) {
                 .await(OperationState.COMPLETED)
                 .validate();
 
-        CloudbreakUser internalActor = cloudbreakActor.create(testContext.getActingUserCrn().getAccountId(), "__internal__actor__");
+        CloudbreakUser internalActor = cloudbreakActor.createInternal(testContext.getActingUserCrn().getAccountId());
```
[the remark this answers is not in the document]
Bajzathd: I will rewrite it so that the createInternal simply calls the create with the internal user username

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/UsedImagesTest.java
Bajzathd: I think this should be fixed in the test framework, but this will be fine for now.

aszegedi (approved):

authorization-common-api/src/main/java/com/sequenceiq/authorization/resource/AuthorizationResourceAction.java
```diff
@@ -81,6 +81,7 @@
     POWERUSER_ONLY("cloudbreak/allowPowerUserOnly", null),
     LIST_ASSIGNED_ROLES("iam/listAssignedResourceRoles", null),
     STRUCTURED_EVENTS_READ("structured_events/read", AuthorizationResourceType.STRUCTURED_EVENT),
+    COLLECT_USED_IMAGES("cloudbreakadmin/collectUsedImages", null),
```
horadla23, line 84: do we plan to introduce a specific admin role for this? I am asking this only for curiosity

horadla23 (approved):
LGTM

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakActor.java
```diff
@@ -47,6 +47,11 @@ public CloudbreakUser create(String tenantName, String username) {
         return new CloudbreakUser(accessKey, secretKey, username + " at tenant " + tenantName);
     }
 
+    @Override
+    public CloudbreakUser createInternal(String tenantName) {
```
lnardai, line 51: Internalactor's has nothing to do with tenant. It's always:
crn:altus:iam:us-west-1:altus:user:__internal__actor__

lnardai: What is the reason you need to have internalActor?

horadla23: I wanted to ask the same. You can use original user and use internal client to call this new endpoint internally

Bajzathd: It just remained there from the previous InternalOnly authorization, currently it is only used here: https://github.com/hortonworks/cloudbreak/blob/6c28b0a9a04223dc19567b5f20858a43a0281574/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/FreeIpaSyncTest.java#L50

authorization-common-api/src/main/java/com/sequenceiq/authorization/resource/AuthorizationResourceAction.java
Bajzathd: the plan is to create a machine user with only this permission so we can call these endpoints from a Jenkins job, using the internal cdp cli

horadla23: if this meant to used with internal cdp cli only, then I think cdp cli should use internal actor and we should allow these endpoints only for internal actor, then we do not need to add a new right to UMS, what do you think about this?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -103,7 +103,7 @@
             + "FROM Stack s LEFT JOIN s.cluster c LEFT JOIN s.workspace WHERE s.terminated = null AND (s.type is not 'TEMPLATE' OR s.type is null)")
     List<StackTtlView> findAllAlive();
 
-    @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig "
+    @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig LEFT JOIN FETCH ig.instanceMetaData im "
```
lnardai, line 106: Is it possible to reduce the size of this resultset? This might get huge and we are only using the InstanceGroups from it.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakActor.java
horadla23: I see, it is a completely different test, not related to the new endpoints. Then I can accept this, but anyway I would prefer to using internal client instead of creating an internal cloudbreakuser cc @aszegedi

authorization-common-api/src/main/java/com/sequenceiq/authorization/resource/AuthorizationResourceAction.java
lnardai: It's not that simple, we've had a discussion with UMS. 
This is the way Admin API is meant to be used. You can't authenticate as an internalActor though the coreadmin api.

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
Bajzathd: good catch, we only need the image so I will modify the query

authorization-common-api/src/main/java/com/sequenceiq/authorization/resource/AuthorizationResourceAction.java
lnardai: Interesting. I just realised that this API can be called by any PowerUser, which was not the intention I think.
At least our framework will just ask UMS about this right and PowerUser has it all.

Is there any separate logic that distinguish internal rights?

horadla23: I don't think so. I missed your pervious comment yesterday, so if internalactor usage is not possible, then makes sense to add a new right.

lnardai (changes requested):
Had a discussion with the UMS team.
In short we need to make this API internal. I'll talk with you after standup.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakActor.java
Bajzathd: I removed it completely and used the internal client

authorization-common-api/src/main/java/com/sequenceiq/authorization/resource/AuthorizationResourceAction.java
Bajzathd: This right was removed (it will be checked in the api service), and marked the endpoints InternalOnly

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
Bajzathd: modified both cb and freeipa queries to just return the images instead of the full stack objects

lnardai (approved):

PR #10018, opened 2021-02-10
CB-10953 remove the CDP_EMBEDDED_DATABASE_ON_ATTACHED_DISK entitlement

keyki:
In case an embedded Postgres is used we will attach an extra volume and
we will store the Postgres data on this disk. During repair/upgrade
we preserve this disk hence we can enable repair and upgrade features
for Data Hubs that don't have external DBs.


keyki:
retest this please

schfeca75 (approved):

PR #10021, opened 2021-02-10
CB-11130 - unify error handling approach over the different kind of experiences during remote call execution

gregito:


bergerdenes (approved):
It is very fishy with the executeSilently
Please open a followup jira to remove this method from the CommonExperienceConnectorServiceTest class!

bergerdenes (changes requested):
remove executeSilently

bergerdenes (approved):

PR #10024, opened 2021-02-10
CB-11079 if the vm is started on GCP manually and in the CB the env s…

topolyai5:
…tatus is running, we don't handle correctly the status. We set the status to null and this null lead to a NPE. I have removed the condition.

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
@@ -446,18 +446,19 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
         Compute compute = context.getCompute();
         String instanceId = instance.getInstanceId();
         try {
+            LOGGER.info("Gcp operations are preparing: instanceId: {}, projectId: {}, availabilityZone: {}", instanceId, projectId, availabilityZone);
             Get get = compute.instances().get(projectId, availabilityZone, instanceId);
-            String state = stopRequest ? "RUNNING" : "TERMINATED";
             Instance instanceResponse = get.execute();
-            if (state.equals(instanceResponse.getStatus())) {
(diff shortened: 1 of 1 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -447,17 +447,11 @@ private CloudVmInstanceStatus stopStart(GcpContext context, AuthenticatedContext
...
-            if (state.equals(instanceResponse.getStatus())) {
-                Operation operation = stopRequest ? compute.instances().stop(projectId, availabilityZone, instanceId).setPrettyPrint(true).execute()
-                        : executeStartOperation(projectId, availabilityZone, compute, instanceId, instance.getTemplate(), instanceResponse.getDisks());
-                CloudInstance operationAwareCloudInstance = createOperationAwareCloudInstance(instance, operation);
-                return new CloudVmInstanceStatus(operationAwareCloudInstance, InstanceStatus.IN_PROGRESS);
-            } else {
-                LOGGER.debug("Instance {} is not in {} state - won't start it.", state, instanceId);
-                return null;
-            }
+            Operation operation = stopRequest ? compute.instances().stop(projectId, availabilityZone, instanceId).setPrettyPrint(true).execute()
```
biharitomi, line 451: Can we use If condition here instead of Elvis operator and please add some additional log that identifies which operation should be initiated?

biharitomi (approved):

biharitomi (changes requested):
Please check my previous comment.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
topolyai5: updated

doktoric (approved):

biharitomi (approved):

PR #10028, opened 2021-02-10
CB-11107 - Supporting paywall for existing HDP clusters - add default…

pdarvasi:
… repos behind paywall

- repos in application.yml are moved behind paywall.
- Debian9 repos that are not present anymore are deleted.
- added HDF-3.5


pdarvasi:
retest this please

schfeca75 (approved):

PR #10029, opened 2021-02-10
CB-10796 Disable changing of primary gateway for SDX.

orngejaket:
This disables the change primary gateway flow for the Datalake since CM HA is not yet implemented.


keyki (approved):

sodre90:
Tested both light and medium duty scenario. All was successful.

PR #10030, opened 2021-02-10
CB-11131 Stabilization fixes

bergerdenes:
Rework DWX and Liftie mocks to consistently return only one freshly created
cluster (this code will be removed upon merging to master, now it servers
testing and demonstration purposes only).

Add Experiences scan enabled flag and set it to default so it will not run by
default on local development machines. It is enabled for integration tests.

Move Spring configuration under 'environment' object. Also, externalize
Experiences-related configuration to help dps-k8s configuration.
A overriding config loader is implemented which reads classpath config first
and tries to override it with configuration file from the file system.
Renamed many parameters.


environment/src/main/resources/experiences-config.yml
```diff
@@ -0,0 +1,12 @@
+environment:
+  experience:
+    scan:
+      liftie:
+        workloads: mon-platform
+        api: https://localhost:10090
```
bergerdenes, line 6: maybe `address` instead of `api` ?

gregito: address is better imo

bergerdenes: renamed

gregito:
in general lgtm

gregito (approved):

gregito (approved):

doktoric (approved):

PR #10033, opened 2021-02-12
CB-11144 ScalingRequest.scaleup() will be executed without any error, in case if error has happened on cloudbreak side

schfeca75:
CB-11144 ScalingRequest.scaleup() will be executed without any error, in case if error has happened on cloudbreak side

* Check the putstack rest response from cloudbreak in ScalingRequest.scaleup() and if its statuscode is not in the SUCCESSFUL family, throws an AutoScalingException
* New entry will be created about the error in the Periscope History and notification will be sent


foldik (approved):
lgtm

biharitomi (approved):

PR #10035, opened 2021-02-13
DISTX-546 Implement cloud provider load-balancer support for datahub …

cegganesh84:
…cluster

1. This allows explicitly setting up a load-balancer to any datahub cluster
2. This reuses the load-balancer implementation for semi-private networks
3. When the load-balancer is enabled removed the restriction of one gateway node because this will be used by DE HA cluster and clusters with custom HA definition
4. If the load-balancer is enabled in the environment, then it is not possible to disable the load-balancer for a particular datahub cluster.
5. Modified 7.2.8 only for now to do the testing and getting the review in parallel, while I hash out the backward incompatibility in the endpoint for the released versions.

Tested with 7.2.7 HA cluster definition in the private stack that the load balancer is created. Also tested if it is not in the cluster definition the load balancer is not created

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -152,14 +152,14 @@ public class LoadBalancerConfigService {
     }
 
     public boolean isLoadBalancerCreationConfigured(Stack stack, DetailedEnvironmentResponse environment) {
-        return !setupLoadBalancers(stack, environment, true).isEmpty();
+        return !setupLoadBalancers(stack, environment, true, false).isEmpty();
     }
 
-    public Set<LoadBalancer> createLoadBalancers(Stack stack, DetailedEnvironmentResponse environment) {
-        return setupLoadBalancers(stack, environment, false);
+    public Set<LoadBalancer> createLoadBalancers(Stack stack, DetailedEnvironmentResponse environment, boolean loadBalancerFlagEnabled) {
+        return setupLoadBalancers(stack, environment, false, loadBalancerFlagEnabled);
     }
 
-    public Set<LoadBalancer> setupLoadBalancers(Stack stack, DetailedEnvironmentResponse environment, boolean dryRun) {
+    public Set<LoadBalancer> setupLoadBalancers(Stack stack, DetailedEnvironmentResponse environment, boolean dryRun, boolean loadBalancerFlagEnabled) {
         if (dryRun) {
             LOGGER.info("Checking if load balancers are enabled and configurable for stack {}", stack.getName());
         } else {
@@ -167,8 +167,12 @@ public class LoadBalancerConfigService {
         }
         Set<LoadBalancer> loadBalancers = new HashSet<>();
 
-        if (isLoadBalancerEnabled(stack.getType(), environment) && getSupportedPlatforms().contains(stack.getCloudPlatform())) {
-            LOGGER.debug("Load balancers are enabled for data lake and data hub stacks.");
+        if (getSupportedPlatforms().contains(stack.getCloudPlatform()) && (loadBalancerFlagEnabled || isLoadBalancerEnabled(stack.getType(), environment))) {
+            if (!loadBalancerFlagEnabled) {
+                LOGGER.debug("Load balancers are enabled for data lake and data hub stacks.");
+            } else {
+                LOGGER.debug("Load balancer is explicitly defined for the stack.");
+            }
             Optional<TargetGroup> knoxTargetGroup = setupKnoxTargetGroup(stack, dryRun);
             if (knoxTargetGroup.isPresent()) {
                 if (isNetworkUsingPrivateSubnet(stack.getNetwork(), environment.getNetwork())) {
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
@@ -26,6 +26,7 @@ import com.sequenceiq.cloudbreak.service.datalake.SdxClientService;
 import com.sequenceiq.cloudbreak.service.environment.EnvironmentClientService;
 import com.sequenceiq.common.api.telemetry.request.TelemetryRequest;
 import com.sequenceiq.common.api.telemetry.response.TelemetryResponse;
+import com.sequenceiq.common.api.type.InstanceGroupType;
 import com.sequenceiq.distrox.api.v1.distrox.model.DistroXV1Request;
 import com.sequenceiq.distrox.api.v1.distrox.model.network.NetworkV1Request;
 import com.sequenceiq.distrox.api.v1.distrox.model.tags.TagsV1Request;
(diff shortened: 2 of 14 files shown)


core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -99,9 +100,31 @@ public StackV4Request convert(DistroXV1Request source) {
...
 
+    private void checkMultipleGatewayNodes(DistroXV1Request distroXV1Request) {
+        if (distroXV1Request.getInstanceGroups() == null || distroXV1Request.getInstanceGroups().isEmpty()) {
+            return;
+        }
+        distroXV1Request.getInstanceGroups().forEach(ig -> {
+            if (InstanceGroupType.GATEWAY == ig.getType()) {
+                if (distroXV1Request.isEnableLoadBalancer()) {
+                    if (ig.getNodeCount() < 1) {
+                        throw new BadRequestException("Instance group with GATEWAY type must contain at least 1 node!");
```
bergerdenes, line 116: how is loadbalancing possible with 1 node?

core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
```diff
@@ -11,10 +11,11 @@
     "externalDatabase": {
       "availabilityType": "HA"
     },
+    "enableLoadBalancer": true,
     "instanceGroups": [{
       "nodeCount": 1,
       "name": "manager",
-      "type": "GATEWAY",
+      "type": "CORE",
```
bergerdenes, line 18: are you sure to run with 1 CORE node?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
cegganesh84: @bergerdenes Cloud load-balancing solution is used in multiple contexts in Cloudbreak. It acts as a static endpoint. If you move CM to a different node and create the cluster with the same name, you still get the same endpoint. In that context allowing a single node GATEWAY with load-balancer is a useful feature. 

The other contexts are,

1. Semi-private networks where the load-balancer is hosted in public network/IP, and the VMs are hosted in private network/IP.
2. Loadbalancing with more than one node, to achieve resiliency in case of node failures. Like Hive endpoint.

cegganesh84:
@bergerdenes @hreeve-cloudera @doktoric Could you please kindly take a look at this.

cegganesh84:
A corresponding PR for thunderhead to enable the load-balancer via CLI https://github.infra.cloudera.com/thunderhead/thunderhead/pull/3677

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -146,12 +146,16 @@ public String getLoadBalancerUserFacingFQDN(Long stackId) {
         return loadBalancerOptional;
     }
 
-    public Set<LoadBalancer> createLoadBalancers(Stack stack, DetailedEnvironmentResponse environment) {
+    public Set<LoadBalancer> createLoadBalancers(Stack stack, DetailedEnvironmentResponse environment, boolean loadBalancerDefined) {
```
hreeve-cloudera, line 149: Could we rename `loadBalancerDefined`? I don't think it's very descriptive, because at this point a load balancer isn't defined. Maybe something like `loadBalancerFlagEnabled`?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -7,6 +7,7 @@
 
 import javax.inject.Inject;
 
+import com.sequenceiq.common.api.type.InstanceGroupType;
```
hreeve-cloudera, line 10: nit: This import is in the wrong place. It should be moved down with the rest of the `import com.*` below.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
```diff
@@ -99,9 +100,31 @@ public StackV4Request convert(DistroXV1Request source) {
...
+        checkMultipleGatewayNodes(source);
         return request;
     }
 
+    private void checkMultipleGatewayNodes(DistroXV1Request distroXV1Request) {
+        if (distroXV1Request.getInstanceGroups() == null || distroXV1Request.getInstanceGroups().isEmpty()) {
+            return;
+        }
+        distroXV1Request.getInstanceGroups().forEach(ig -> {
+            if (InstanceGroupType.GATEWAY == ig.getType()) {
```
hreeve-cloudera, line 113: The current load balancer logic already checks for the existence of the Knox service, and will only create load balancers if Knox is defined either explicitly or implicitly. Is there a reason not related to Knox we're checking the instance type here?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceGroupV1ToInstanceGroupV4Converter.java
```diff
@@ -45,9 +44,6 @@
 
     private InstanceGroupV4Request convert(InstanceGroupV1Request source, DetailedEnvironmentResponse environment) {
         InstanceGroupV4Request response = new InstanceGroupV4Request();
-        if (InstanceGroupType.GATEWAY == source.getType() && source.getNodeCount() != 1) {
```
hreeve-cloudera, line 48: Why was this changed?

core/src/test/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverterTest.java
```diff
@@ -17,6 +18,10 @@
 import java.util.Map;
 import java.util.Set;
 
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.RecoveryMode;
```
hreeve-cloudera, line 21: nit: Reorder imports.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/DistroXV1RequestToStackV4RequestConverter.java
cegganesh84: Yes, the datahub has a special check not to allow GATEWAY node type to be defined for more than one node. This change basically enables to relax that check when a load-balancer is defined. This will enable us to have a master group with two nodes as GATEWAY where in the future we will also have CM HA.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceGroupV1ToInstanceGroupV4Converter.java
cegganesh84: This check was moved to a place where we can check if it is a load-balancer enabled cluster or not.

bergerdenes (approved):

cegganesh84:
Thanks for the reviews folks. Could this be merged?

bergerdenes:
Please have at least another pair of eyes check this PR

hreeve-cloudera (approved):

cegganesh84:
There was a merge conflict from another load-balancer-related change. I have rebased and the checks are passing again.

PR #10039, opened 2021-02-15
CB-10620 Connectivity check on FreeIPA nodes

daszabo:


freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -82,6 +92,46 @@ public SyncResult getStatus(Stack stack, Set<InstanceMetaData> checkableInstance
         }
     }
 
+    private void checkNodeHealthReports(Stack stack, Set<InstanceMetaData> checkableInstances) {
+        for (InstanceMetaData instanceMetaData : checkableInstances) {
+            try {
+                LOGGER.debug("Fetching node health reports for instance: {}", instanceMetaData.getInstanceId());
+                RPCResponse<HealthReport> meteringReportRPCResponse = checkedMeasure(() -> freeIpaNodeStatusService.nodeMeteringReport(stack, instanceMetaData),
```
oleewere, line 99: it's true that we have a metering endpoint on every node, but it needs to be used only by datahub

oleewere (changes requested):

freeipa/src/main/resources/freeipa-salt/salt/nginx/conf/ssl-locations.d/nodestatus.conf
```diff
@@ -1,5 +1,5 @@
 location /nodestatus {
-  proxy_pass         https://{{ salt['grains.get']('fqdn') }}:61888;
+  proxy_pass         http://127.0.0.1:61888;
```
jamisonbennett, line 2: Isn't is still supposed to be https?

oleewere: not yet, it will be similar as satboot, basi auth without ssl (actually won't be anything secret in the communication, maybe later ssl can be added if needed)

oleewere (approved):
lgtm

PR #10045, opened 2021-02-15
CB-10498 cloudcontext refactor to make it a builder. now we can make …

doktoric:
…sure that in case on image creation we are always using the account uuid


topolyai5 (approved):
LGTM

PR #10056, opened 2021-02-16
CB-11188 Optionally initialize real UMS users by parameter

aszegedi:
Integration tests' flow polling is failing in case of `ums-users/api-credentials.json` is present at `integration-test` project. This is a special case, when the real UMS user store has been initialised but we do not want to run UMS or AuthZ test suits. However flow polling is happening with admin user. So mistakenly UMS admin has been provided for `getAdminMicroserviceClient`.

We agreed we should introduce the needed validation for `accountId` in `getAdminByAccountId` and an additional parameter for authorisation tests only.

**_NOTES:_**
- `additional parameter for authorization tests only` was implemented different way: Instead of introducing a separate test suite parameter I applied a new `CloudbreakUserCache` class parameter as `useUmsStore`. This new parameter is going to be set when we use `useRealUmsUser` (from `AbstractIntegrationTest`). So on this way we can rest assured the `getAdminMicroserviceClient` and `doAsAdmin` are getting the UMS admin for the used UMS account when and only when we use Real UMS user for tests.
- In case os some Environment service related actions (edit and delete where these are returning with the response) the `testDto.setResponse` step was missing. So I made up the flaw.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
@@ -85,30 +85,39 @@ public class CloudbreakUserCache {
         this.usersByAccount = users;
     }
 
-    public CloudbreakUser getByDisplayName(String name) {
+    public CloudbreakUser getUserByDisplayName(String name) {
         if (MapUtils.isEmpty(usersByAccount)) {
             initRealUmsUserCache();
         }
-        CloudbreakUser user = usersByAccount.values().stream().flatMap(Collection::stream)
-                .filter(u -> u.getDisplayName().equals(name)).findFirst()
-                .orElseThrow(() -> new TestFailException(String.format("There is no real UMS user with::%n name: %s%n deployment::account: %s::%s", name,
-                        realUmsUserDeployment, realUmsUserAccount)));
-        LOGGER.info(" Real UMS user has been found:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
-                user.getDisplayName(), user.getCrn(), user.getAccessKey(), user.getSecretKey(), user.getAdmin());
-        return user;
+        if (isInitialized()) {
+            CloudbreakUser user = usersByAccount.values().stream().flatMap(Collection::stream)
+                    .filter(u -> u.getDisplayName().equals(name)).findFirst()
+                    .orElseThrow(() -> new TestFailException(String.format("There is no real UMS user with::%n name: %s%n deployment: %s%n account: %s%n",
+                            name, realUmsUserDeployment, realUmsUserAccount)));
+            LOGGER.info(" Real UMS user has been found:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
+                    user.getDisplayName(), user.getCrn(), user.getAccessKey(), user.getSecretKey(), user.getAdmin());
+            return user;
+        } else {
+            throw new TestFailException("Cannot get real UMS user by name, because of 'ums-users/api-credentials.json' is not available.");
+        }
     }
 
     public CloudbreakUser getAdminByAccountId(String accountId) {
-        LOGGER.info("Getting the requested real UMS admin by account Id: {}", accountId);
-        try {
-            CloudbreakUser adminUser = usersByAccount.values().stream().flatMap(Collection::stream)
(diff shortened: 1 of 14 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -89,30 +92,49 @@ public CloudbreakUser getByDisplayName(String name) {
...
-            CloudbreakUser adminUser = usersByAccount.values().stream().flatMap(Collection::stream)
-                    .filter(CloudbreakUser::getAdmin).findFirst()
-                    .orElseThrow(() -> new TestFailException(String.format("There is no real UMS admin in account: %s", accountId)));
-            LOGGER.info(" Real UMS account admin has been found:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
-                    adminUser.getDisplayName(), adminUser.getCrn(), adminUser.getAccessKey(), adminUser.getSecretKey(), adminUser.getAdmin());
-            return adminUser;
-        } catch (Exception e) {
-            throw new TestFailException(String.format("Cannot get the real UMS admin in account: %s, because of: %s", accountId, e.getMessage()), e);
+        if (!usersByAccount.containsKey(accountId)) {
+            throw new TestFailException("Account mismatch:: Real UMS '" + accountId + "' account ID is not present! Please make sure: " +
```
horadla23, line 111: I would simplify this message, like this: `accountId .. is missing from ums-users/api-credential.json`

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -89,30 +92,49 @@ public CloudbreakUser getByDisplayName(String name) {
...
             initRealUmsUserCache();
         }
-        CloudbreakUser user = usersByAccount.values().stream().flatMap(Collection::stream)
-                .filter(u -> u.getDisplayName().equals(name)).findFirst()
-                .orElseThrow(() -> new TestFailException(String.format("There is no real UMS user with::%n name: %s%n deployment::account: %s::%s", name,
-                        realUmsUserDeployment, realUmsUserAccount)));
-        LOGGER.info(" Real UMS user has been found:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
-                user.getDisplayName(), user.getCrn(), user.getAccessKey(), user.getSecretKey(), user.getAdmin());
-        return user;
+        if (MapUtils.isNotEmpty(usersByAccount)) {
```
horadla23, line 95: `isInitialized` method can be used here

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -89,30 +92,49 @@ public CloudbreakUser getByDisplayName(String name) {
...
+            CloudbreakUser user = usersByAccount.values().stream().flatMap(Collection::stream)
+                    .filter(u -> u.getDisplayName().equals(name)).findFirst()
+                    .orElseThrow(() -> new TestFailException(String.format("There is no real UMS user with::%n name: %s%n deployment::account: %s::%s", name,
+                            realUmsUserDeployment, realUmsUserAccount)));
+            LOGGER.info(" Real UMS user has been found:: \nDisplay name: {} \nCrn: {} \nAccess key: {} \nSecret key: {} \nAdmin: {} ",
+                    user.getDisplayName(), user.getCrn(), user.getAccessKey(), user.getSecretKey(), user.getAdmin());
+            return user;
+        } else {
+            throw new TestFailException("UMS user store 'ums-users/api-credentials.json' is not available." +
```
horadla23, line 104: the message should be something related to the method, like: "Getting user by name ... failed since ums user store is empty"

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/actor/CloudbreakUserCache.java
```diff
@@ -39,6 +40,8 @@
     @Value("${integrationtest.ums.deploymentKey:}")
     private String realUmsUserDeployment;
 
+    private boolean useUmsStore;
```
horadla23, line 43: i would not store this flag here, this cache is related to users, this flag is a global test config which decides we are using ums or not

horadla23 (changes requested):

horadla23 (approved):
LGTM

PR #10063, opened 2021-02-17
CB-11162 We don't want to accept a whitespace in the storage location…

topolyai5:
… or in the base path. If the whitespace at the end of the path, we will trim first and we will validate after the trim.

@akanto please check the unit tests, if these cases are valid or not

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentValidatorService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentValidatorService.java
@@ -9,6 +9,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.ws.rs.BadRequestException;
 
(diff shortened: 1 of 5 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentValidatorService.java
```diff
@@ -219,4 +221,18 @@ private Boolean singleInstanceIsOnDemandOrSpot(FreeIpaCreationDto freeIpaCreatio
...
+    public ValidationResult validateStorageLocation(String storageLocation) {
+        ValidationResultBuilder resultBuilder = new ValidationResultBuilder();
+        if (storageLocation != null) {
+            Pattern pattern = Pattern.compile(".*\\s.*");
+            Matcher matcher = pattern.matcher(storageLocation.trim());
+            if (matcher.find()) {
+                resultBuilder.error("You have added some whitespace to the storage location: " + storageLocation);
+            }
+        } else {
+            resultBuilder.error("You don't add a storage location, please provide a valid storage location.");
```
doktoric, line 234: You did not specified the storage location, please provide a valid storage location.
Would be nice to see which storage location

topolyai5: which storage location? the regexp or valid chars?

akanto (approved):

PR #10065, opened 2021-02-17
CB-11217 Extracted all platform configs under cdp: namespace. Left th…

lnardai:
…e global supportedPlatforms+experimental platforms and added a separate block for features/platforms configuration. We should be able to clearly see what feature is enabled on what platforms, and this should be decoupled from the fact which platform is enabled at the moment. Default values should only be provided through application.yml configurations, never @Value os PostConstruct annotations.

**Please DO NOT MERGE THIS until we agreed on a format and helm charts are updated properly.**


doktoric:
This probably requires cloudbreak-deployer updates if we have any pointer there for this properties.

lnardai:
retest this!

doktoric:
retest this please

doktoric:
@lnardai do you plan to finish this?

PR #10069, opened 2021-02-17
CB-11221. Diagnostics - Send case number as text.

oleewere:

### DIFF
--- a/orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/filecollector.yaml.j2
+++ b/orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/filecollector.yaml.j2
@@ -90,7 +90,7 @@ collector:
         customer:
             accountId: {{ filecollector.accountId }}
             licenseId:{% if filecollector.issue %}
-        Case_number: {{ filecollector.issue }}{% endif %}
+        Case_number: "{{ filecollector.issue }}"{% endif %}
 {% if destination == "ENG" and fluent.dbusClusterLogsCollection %}
     compress: false
     deleteProcessedTempFilesOneByOne: true


attilapalfi92 (approved):

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/filecollector.yaml.j2
```diff
@@ -90,7 +90,7 @@ collector:
         customer:
             accountId: {{ filecollector.accountId }}
             licenseId:{% if filecollector.issue %}
-        Case_number: {{ filecollector.issue }}{% endif %}
+        Case_number: "{{ filecollector.issue }}"{% endif %}
```
aakanser, line 93: I'm not very familiar with yaml.j2. What would happen if the issue is sent in with quotes? Would this result in `"\"text\""` ?

From the UDX UI, the input will just be a number, so I expect it will result in `"number"`, which is good. I'm just curious about other cases.

aakanser (approved):
looks good to me, just had a single question

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/template/filecollector.yaml.j2
oleewere: this is what we are doing with the output: read the yaml then save this object as a json and put into the bundle.
with yaml the use case could be:
Case_number: 11
or 
Case_number: hello world

the first one will be a number the second one will be a string, adding quotes will result that both will be string anyway

aakanser: sounds good, thank you!

PR #10071, opened 2021-02-17
Flowchain names

akanto:
With this commit the name of the flowchain is passed to flow and also passed to the structured events. The benefit of this is that we can create EDH events based on the FlowChain names which highly simplifies the switch case statements in ClusterUseCaseMapper and EnvironmentUseCaseMapper classes. I have no intention to track the parents of the FlowChains since tracking the child FLowChain provides enough details for us.

The ClusterUseCaseMapper and EnvironmentUseCaseMapper is going to be refactored in the next commit.


oleewere (approved):
lgtm, thanks, that looks useful

PR #10075, opened 2021-02-17
CB-11231 GCP : Adding cluster definitions for spark3 and dataengineer…

cegganesh84:
…ing ha

1. Selected the nodegroups from the template and the instance types from dataengineering definition.
2. This should help finalize the list of templates supported for GCP datahub GA.


doktoric:
seems we can not load several templates. maybe it a parsing issue.

cegganesh84:
Thanks, @doktoric very nice test case, fixed the issue.

doktoric (approved):

PR #10081, opened 2021-02-18
 DWX-6365 Adding ECR repo access to cloudbreak client

clmstr:
A ECRAccessRead policy will enable Cloudbreak client to describe repositories. CDP has a new feature which requires customers to define custom ECR repositories and Cloudbreak client will need to have rights to be able to describe repositories if the custome repo is defined on the same account.


clmstr:
recheck

clmstr:
yes

clmstr:
> Can one of the admins verify this patch?

yes

PR #10086, opened 2021-02-18
CB-11105 remove idbroker from cluster

sodre90:
IDBroker was fetched in every stack get query. (with its relations like
cluster etc -> hibernate created a join query with these relations)

Explanation: It is because Stack <--> Cluster is a OneToOne relation
and Cluster <--> IdBroker is also a OneToOne relation. A OneToOne relation
 can not be lazy in the case if the owner of the relation is the other entity.
In this example the Idbroker entity has the cluster id column.
Hibernate can not guess if Cluster has an IDBroker or not because this
information is in another table, so IDBroker table will be fetched.

With this fix IdBroker entity can be fetched only with it's service.

https://stackoverflow.com/questions/1444227/how-can-i-make-a-jpa-onetoone-relation-lazy

"Unconstrained (nullable) one-to-one association is the only one that
can not be proxied without bytecode instrumentation. The reason for this
is that owner entity MUST know whether association property should
contain a proxy object or NULL and it can't determine that by looking at
its base table's columns due to one-to-one normally being mapped via
shared PK, so it has to be eagerly fetched anyway making proxy
pointless."

### DIFF
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/cluster/Cluster.java
@@ -172,9 +172,6 @@ public class Cluster implements ProvisionEntity, WorkspaceAwareResource {
     @OneToOne(mappedBy = "cluster", cascade = {CascadeType.PERSIST, CascadeType.REMOVE}, orphanRemoval = true)
     private Gateway gateway;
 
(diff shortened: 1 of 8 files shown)


lacikaaa:
could you add more info to the commit message?

akanto (approved):

akanto:
looks good, the reason (mainly performance) should be added to the commit message and after that, we can merge it

sodre90:
Of course, I added detailed info on why it will decrease DB load.

lacikaaa (approved):

PR #10087, opened 2021-02-18
CB-11236 We don't need to stop the CM if it is already stopped or una…

topolyai5:
…vailable any other reasons. Like other CM api calls, we retry 5 times before we say that is not running

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/ClusterStopHandler.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/ClusterStopHandler.java
@@ -2,6 +2,8 @@ package com.sequenceiq.cloudbreak.reactor.handler.cluster;
 
 import javax.inject.Inject;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
@@ -17,6 +19,9 @@ import reactor.bus.EventBus;
 
 @Component
 public class ClusterStopHandler implements EventHandler<ClusterStopRequest> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClusterStopHandler.class);
+
     @Inject
     private StackService stackService;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/ClusterStopHandler.java
```diff
@@ -37,7 +42,12 @@ public void accept(Event<ClusterStopRequest> event) {
         ClusterStopResult result;
         try {
             Stack stack = stackService.getByIdWithListsInTransaction(request.getResourceId());
-            apiConnectors.getConnector(stack).stopCluster(false);
+            boolean clusterManagerRunning = apiConnectors.getConnector(stack).clusterStatusService().isClusterManagerRunning();
```
lacikaaa, line 45: as I see only `ApiException` is handled here. Are we sure we can't receive any other type of exception when CM is not running?

topolyai5: indeed, I'm checking that

topolyai5: I have stopped the CM and working fine, but I didn't check the exception

lacikaaa: have you tested it with clusterproxy/CCM?

topolyai5: Nit yet, I will try to test it

topolyai5: Answer for the first question:
We can receive an ApiException here, the ApiException store the status code like 502

lacikaaa (approved):

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/ClusterStopHandler.java
topolyai5: Working with cluster proxy as well

topolyai5: dh works as well

PR #10088, opened 2021-02-18
CB-10926 Removed 7.2.8 Datamart from gcloud because it does not reall…

doktoric:
…y work on cloud scenarios.


cegganesh84 (approved):
Thank you!

PR #10090, opened 2021-02-19
DISTX-570 Suspend AS for repeated scaling failures of cluster

smaniraju:
AS should be suspended if repeated scaling failures are encountered for the same cluster.
Just marking Cluster Status as suspended is not correct, because Cluster Status would be remarked as "Running" in the next Periscope - CB Status Sync.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.periscope.monitor.handler;
 
+import static com.sequenceiq.periscope.common.MessageCode.AUTOSCALING_SUSPENDED;
+
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
```diff
@@ -42,17 +52,12 @@ public void onApplicationEvent(UpdateFailedEvent event) {
...
             updateFailures.remove(autoscaleClusterId);
-            LOGGER.debug("Suspended cluster monitoring for cluster '{}' due to failing update attempts", cluster.getStackCrn());
-        }
-    }
-
-    private void suspendCluster(Cluster cluster) {
-        if (!cluster.getState().equals(ClusterState.SUSPENDED)) {
-            clusterService.setState(cluster.getId(), ClusterState.SUSPENDED);
+            historyService.createEntry(ScalingStatus.DISABLED, messagesService.getMessage(AUTOSCALING_SUSPENDED), cluster);
+            LOGGER.info("Suspended Autoscaling for cluster '{}' due to repeated failure of scaling attempts", cluster.getStackCrn());
```
sidseth, line 60: Would be useful to include the number of failures, and over what duration of time.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
```diff
@@ -12,21 +14,29 @@
...
     @Inject
     private ClusterService clusterService;
 
+    @Inject
+    private HistoryService historyService;
+
+    @Inject
+    private CloudbreakMessagesService messagesService;
+
     private final Map<Long, Integer> updateFailures = new ConcurrentHashMap<>();
```
sidseth, line 40: Are failures ever reset?, other than at the point when the cluster is being disabled (or Periscope restarts)?

sidseth (commented):
I'm not sure disabling AutoScaling is the best approach (Users would have explicitly enabled it - so coming back to find it disabled would not be the best experience). At what point would this be re-enabled?
An alternate could be something along the lines of not attempting to AutoScale the cluster for a certain time duration? Any other approaches?

smaniraju:
The type of failures handled by UpdateFailedHandler is limited to issues found during gathering of metrics and sending CB Requests. Hence these type of failures will not recover automatically, in production logs I observed same failure for same cluster continuously over multiple days.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
smaniraju: It is reset as soon as AS is suspended.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
smaniraju: Time tracking of failures is not implemented in periscope. That would be a new feature.

smaniraju:
Will impelement a different approach for this.

PR #10094, opened 2021-02-19
CB-11230 remove sync from sdx start and sdx stop flow

horadla23:


sodre90:
please remove SdxSyncHandler

horadla23:
> please remove SdxSyncHandler and related request response

@sodre90 thx, I missed that those are not used anymore, removed

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/SdxStopState.java
```diff
@@ -8,7 +8,6 @@
 public enum SdxStopState implements FlowState {
 
     INIT_STATE,
-    SDX_STOP_SYNC_STATE,
```
akanto, line 11: What happens with the flows that are in this state when this change is rolled out? Suck flow due " java.lang.IllegalArgumentException: No enum const" or similar?

horadla23: Yes, probably that would be the case, I need to check it tomorrow.
Maybe the safest and simplest solution would be to keep SdxSyncHandler, related steps, events and only remove logic from SdxSyncHandler and SdxSyncHandler would give back SdxSyncSuccessEvent everytime, in that case existing flows wouldn't be in stucked state and also future flows wouldn't get into stuck state in case of sync failure, since there would be no failure in sync step.
Or do you know @akanto or @sodre90 or @schfeca75 if there is a safe way to remove a particular step (and related events, handlers, etc.) of a flow without breaking existing ones?

akanto: Can we just skip this step? I mean that from the former flow step if we send such an event which skips this step, then next time (e.g one release later) it would be safe to completely eliminate this step.

sodre90: yes, I think the best option is to create a new step as Adam did (modify the original one)

```
.from(SDX_START_RDS_START_STATE).to(SDX_START_START_STATE)
            .event(SdxStartEvent.SDX_START_RDS_FINISHED_EVENT).defaultFailureEvent()	            .event(SdxStartEvent.SDX_START_RDS_FINISHED_EVENT).defaultFailureEvent()

```

but do not remove this former flow step

```
.from(SDX_START_SYNC_STATE).to(SDX_START_START_STATE)	
            .event(SdxStartEvent.SDX_SYNC_FINISHED_EVENT).noFailureEvent()
```

Also, remove the handler and request-response as Adam did.
But keep the:

```
@Bean(name = "SDX_STOP_SYNC_STATE"
    public Action<?, ?> sdxSync() {
```

It should just do nothing but send a SdxStartStopEvent instead of the handler request and former SdxSyncSuccessEvent which was sent by the removed handler.

In this case, if the flow was at SDX_START_SYNC_STATE then it can be continued and it will send a SdxSyncSuccessEvent in every scenario, so the flow will step to SDX_START_START_STATE.

In the next release we can remove the
```
@Bean(name = "SDX_STOP_SYNC_STATE"
    public Action<?, ?> sdxSync() {
```
and also:
```
.from(SDX_START_SYNC_STATE).to(SDX_START_START_STATE)	
            .event(SdxStartEvent.SDX_SYNC_FINISHED_EVENT).noFailureEvent()
```

horadla23: thx, makes sense, will change PR soon

lnardai (approved):

sodre90 (approved):
LGTM

PR #10099, opened 2021-02-20
CB-11281. Create common flow progress API for env/dl/dh/freeipa/redbeams

oleewere:
details:
- create flow progress hoder for every flow (filled at startup) - handles init/cancelled/finished/failed states with default values
- 2 new api endpoint for flow public -? get all progress state or get only the last (in a more readable format with progression percent and elapsed time)
- small refactor on a flow (flow id was not considered in the method name, so make it more clear by using find first instead of find all)
- removed OrderedFlow interface (to fill progress map for everything on startup)
- no type record yet for the flows (can be useful later with aa human readable name)
- add endpoints for all of these components: environments/redbeams/freeipa/datalake/datahub 
- for datalake api, the response differs, it also contains the flows from core api (as there would be no other way to get those...only by different endpoints, but not sure the user should now about that)


oleewere:
example output:

```json
{
   "flowId":" ...",
   "flowChainId": "...",
   "created" : 123455,
   "elapsedTimeInSec": 25.05,
   "progress": 92,
   "transitions": [
       {
         "state": "WHATEVER_STATE_1",
         "status": "SUCCESS",
         "nextEvent": "MY_EVENT_3",
         "elapsedTimeSec": 13
       },
       {
         "state": "WHATEVER_STATE_2",
         "status": "PENDING",
         "nextEvent": "MY_EVENT_2",
         "elapsedTimeSec": 12.05
       }
   ]
}
```

foldik (approved):
lgtm

akanto (approved):

flow/src/main/java/com/sequenceiq/flow/controller/FlowPublicController.java
```diff
@@ -27,4 +30,16 @@ public FlowCheckResponse hasFlowRunningByChainId(String chainId, String resource
     public FlowCheckResponse hasFlowRunningByFlowId(String flowId, String resourceCrn) {
         return flowService.getFlowState(flowId);
     }
+
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.POWERUSER_ONLY)
```
akanto, line 35: I think it is ok now, but the question is why power user? I guess a proper authz would be desirable.

E.g if I am just a DataHub admin, then my permission should be sufficient to check what is happening with my DH.

oleewere: seems like we would need individual endpoints for ever resource, so probably for now it will be available on flow api as well, but if we would want any UI or resource specific features around it, specific endpoints will need to be implemented

oleewere: ok, aadded new endpoints

PR #10100, opened 2021-02-20
CB-11282 ensure that we have proper tracking in EDH for securityAcces…

akanto:
…sType. Which means that we would like to track whether customers are using CIDR or security groups to allow access their clusters

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/dto/SecurityAccessDto.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/dto/SecurityAccessDto.java
@@ -1,5 +1,9 @@
 package com.sequenceiq.environment.environment.dto;
 
+import org.apache.commons.lang3.StringUtils;
+
+import com.sequenceiq.cloudbreak.validation.SubnetValidator;
+
 public class SecurityAccessDto {
 
     private String securityGroupIdForKnox;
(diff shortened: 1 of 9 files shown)


nadamdb (approved):

environment/src/main/java/com/sequenceiq/environment/environment/dto/SecurityAccessDto.java
```diff
@@ -38,12 +49,22 @@ public static Builder builder() {
 
     public static class Builder {
 
+        private static final String TYPE_CIDR_WIDE_OPEN = "CIDR_WIDE_OPEN";
```
bergerdenes, line 52: smells like it should have been an enum

bergerdenes (changes requested):
please make the new field as an enum

environment/src/main/java/com/sequenceiq/environment/environment/dto/SecurityAccessDto.java
akanto: They are converted to String anyway, I would rather delete the static declarations

akanto: done

bergerdenes (approved):

biharitomi (approved):
LGTM

PR #10101, opened 2021-02-22
DWX-6365 Adding ECR repo access to cloudbreak client 

clmstr:
A ECRAccessRead policy will enable Cloudbreak client to describe repositories. CDP has a new feature which requires customers to define custom ECR repositories and Cloudbreak client will need to have rights to be able to describe repositories if the custome repo is defined on the same account.

### DIFF
--- a/cloud-aws/src/main/resources/definitions/aws-cb-policy.json
+++ b/cloud-aws/src/main/resources/definitions/aws-cb-policy.json
@@ -249,6 +249,18 @@
       ]
     },
     {
+      "Sid": "ECRAccessRead",
+      "Effect": "Allow",
+      "Action": [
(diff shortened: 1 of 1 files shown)


clmstr:
> Can one of the admins verify this patch?

yes

clmstr:
> Can one of the admins verify this patch?

yes

lajosrodek (approved):
LGTM. The policy file size is 5655 bytes including the change, which is still below the 6144-character limit on managed policies [imposed by IAM](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html). (Note: 6144 characters are counted by ignoring white space, so the effective policy size is even smaller.)

doktoric (approved):

PR #10104, opened 2021-02-22
CB-8476 Azure instance deallocation improvement

lturcsanyi:
- Check actual status of instances and deallocate only those that are not STOPPED
 - Collect deallocation results properly
 - Retry deallocation if something went wrong (exception raised)
 - Improve error logging

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureUtils.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureUtils.java
@@ -57,6 +57,7 @@ import com.sequenceiq.common.api.type.ResourceType;
 
 import rx.Completable;
 import rx.Observable;
+import rx.exceptions.CompositeException;
 import rx.schedulers.Schedulers;
 
 @Component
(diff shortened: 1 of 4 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureUtils.java
```diff
@@ -243,23 +247,44 @@ public void validateStorageTypeForGroup(AzureDiskType diskType, String flavor) {
     }
 
     @Retryable(backoff = @Backoff(delay = 1000, multiplier = 2, maxDelay = 10000), maxAttempts = 5)
-    public List<CloudVmInstanceStatus> deallocateInstances(AuthenticatedContext ac, List<CloudInstance> vms) {
+    public List<CloudVmInstanceStatus> deallocateInstances(AuthenticatedContext ac, List<CloudInstance> vms) throws BatchInstanceActionFailedException {
```
akanto, line 250: Please be so kind to split it into smaller methods for testability.

lturcsanyi: Sure, I will do.

sodre90 (approved):

PR #10118, opened 2021-02-23
CB-11300 Put flow in failed state even if consumer is not found for an event

sodre90:


foldik (changes requested):
The change seems to be good, but can you please write a few tests to show those cases that this change protects from?

horadla23 (approved):
lgtm

PR #10120, opened 2021-02-23
CB-11286 revive IT's TagUtil unit tests and modify tag verification to support cases when the resource creator that comes in resource tag response is related to different CRN partition

biharitomi:
To avoid issues with non-real UMS environments when some of the resources are created with resource creator tag that has `cdp` CRN partition even E2E test framework is configured with an access/secret key pair that relates to a CRN with `altus` partition.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
@@ -4,13 +4,14 @@ import static java.lang.String.format;
 
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 import org.springframework.util.CollectionUtils;
-import org.springframework.util.StringUtils;
 import org.testng.asserts.SoftAssert;
 
+import com.google.common.base.Objects;
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.cloud.gcp.util.GcpLabelUtil;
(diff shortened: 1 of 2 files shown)


integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
```diff
@@ -134,31 +142,39 @@ private void validateClouderaCreatorResourceNameTag(TaggedResponse response, Str
         if (StringUtils.isEmpty(tagValue)) {
             tagValue = response.getTagValue(tag.toLowerCase());
         }
-        if (gcpLabelTransformedValue(tagValue, actingUserCrn.toString())) {
+        if (actingUserCrn != null && gcpLabelTransformedValue(tagValue, actingUserCrn.toString())) {
             Log.log(LOGGER, format(" Default tag: [%s] value is: [%s] equals [%s] acting user CRN transformed to a GCP label value! ", tag, tagValue,
                     actingUserCrn));
         } else {
```
Bajzathd, line 148: actingUserCrn can be null in the else and it could case NPE in crnEqualsWithoutConsideringPartition

horadla23: i think it is ensured that actingusercrn won't be null

horadla23 (approved):
LGTM

integration-test/src/main/java/com/sequenceiq/it/util/TagsUtil.java
biharitomi: I fixed it, as fixed many condition with null and empty string check :)

Bajzathd (approved):

aszegedi (approved):

PR #10124, opened 2021-02-24
CB-11302: Allow the ability for customer to choose a previously relea…

LinaAtAustin:
…sed hotfix

-The approach is to add a new input in cdp cli. So customers can specify the image catalog and uuid when creating the data lake
-add a new field in the SdxClusterRequest that contains the image info
-the image info is set to StackV4Request

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ImageCatalogV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ImageCatalogV4Controller.java
@@ -4,6 +4,7 @@ import static com.sequenceiq.authorization.resource.AuthorizationResourceAction.
 import static com.sequenceiq.authorization.resource.AuthorizationVariableType.CRN;
 
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
@@ -38,6 +39,7 @@ import com.sequenceiq.cloudbreak.cloud.model.catalog.Images;
 import com.sequenceiq.cloudbreak.common.type.ResourceEvent;
 import com.sequenceiq.cloudbreak.domain.ImageCatalog;
 import com.sequenceiq.cloudbreak.service.image.ImageCatalogService;
+import com.sequenceiq.cloudbreak.service.image.StatedImage;
 import com.sequenceiq.cloudbreak.workspace.controller.WorkspaceEntityType;
 
 @Controller
@@ -145,6 +147,14 @@ public class ImageCatalogV4Controller extends NotificationController implements 
     @CheckPermissionByResourceName(action = AuthorizationResourceAction.DESCRIBE_IMAGE_CATALOG)
     public ImagesV4Response getImagesByName(Long workspaceId, @ResourceName String name, String stackName, String platform) throws Exception {
         Images images = imageCatalogService.getImagesByCatalogName(workspaceId, name, stackName, platform);
+        return converterUtil.convert(images, ImagesV4Response.class);
+    }
+
+    @Override
+    @CheckPermissionByResourceName(action = AuthorizationResourceAction.DESCRIBE_IMAGE_CATALOG)
+    public ImagesV4Response getImageByCatalogNameAndImageId(Long workspaceId, @ResourceName String name, String imageId) throws Exception {
+        StatedImage statedImage = imageCatalogService.getImageByCatalogName(workspaceId, imageId, name);
+        Images images = new Images(List.of(), List.of(statedImage.getImage()), Set.of());
         return converterUtil.convert(images, ImagesV4Response.class);
     }
 }
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
@@ -33,6 +33,7 @@ import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncStatus;
 import com.sequenceiq.sdx.api.model.SdxClusterDetailResponse;
 import com.sequenceiq.sdx.api.model.SdxClusterRequest;
 import com.sequenceiq.sdx.api.model.SdxClusterResponse;
+import com.sequenceiq.sdx.api.model.SdxCustomClusterRequest;
 import com.sequenceiq.sdx.api.model.SdxDatabaseBackupResponse;
 import com.sequenceiq.sdx.api.model.SdxDatabaseBackupStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxDatabaseRestoreResponse;
@@ -255,4 +256,10 @@ public interface SdxEndpoint {
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = ROTATE_CERTIFICATES, nickname = "rotateAutoTlsCertificatesByCrn")
     FlowIdentifier rotateAutoTlsCertificatesByCrn(@PathParam("crn") String crn, @Valid CertificatesRotationV4Request rotateCertificateRequest);
+
+    @POST
+    @Path("{name}/custom_image")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "create custom SDX cluster", produces = MediaType.APPLICATION_JSON, nickname = "createCustomSdx")
+    SdxClusterResponse create(@PathParam("name") String name, @Valid SdxCustomClusterRequest createSdxClusterRequest);
 }
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequest.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequest.java
@@ -1,49 +1,8 @@
 package com.sequenceiq.sdx.api.model;
 
-import java.util.HashMap;
-import java.util.Map;
-
-import javax.validation.Valid;
-import javax.validation.constraints.NotNull;
-
-import com.sequenceiq.common.api.tag.request.TaggableRequest;
-
-public class SdxClusterRequest implements TaggableRequest {
-
-    @NotNull
-    private String environment;
-
-    @NotNull
-    private SdxClusterShape clusterShape;
+public class SdxClusterRequest extends SdxClusterRequestBase {
 
     private String runtime;
-
-    private SdxCloudStorageRequest cloudStorage;
-
-    private SdxDatabaseRequest externalDatabase;
-
-    @Valid
-    private SdxAwsRequest aws;
-
-    private Map<String, String> tags;
-
-    private boolean enableRangerRaz;
-
-    public String getEnvironment() {
-        return environment;
-    }
-
-    public void setEnvironment(String environment) {
-        this.environment = environment;
-    }
-
-    public SdxClusterShape getClusterShape() {
-        return clusterShape;
-    }
-
-    public void setClusterShape(SdxClusterShape clusterShape) {
-        this.clusterShape = clusterShape;
-    }
 
     public String getRuntime() {
         return runtime;
@@ -53,71 +12,10 @@ public class SdxClusterRequest implements TaggableRequest {
         this.runtime = runtime;
     }
 
-    public void setTags(Map<String, String> tags) {
-        this.tags = tags;
-    }
-
-    public Map<String, String> getTags() {
-        return tags;
-    }
-
-    @Override
-    public void addTag(String key, String value) {
-        initAndGetTags().put(key, value);
-    }
-
-    public Map<String, String> initAndGetTags() {
-        if (tags == null) {
-            tags = new HashMap<>();
-        }
-        return tags;
-    }
-
-    public void addTags(Map<String, String> tags) {
-        initAndGetTags().putAll(tags);
-    }
-
-    public SdxCloudStorageRequest getCloudStorage() {
-        return cloudStorage;
-    }
-
-    public void setCloudStorage(SdxCloudStorageRequest cloudStorage) {
-        this.cloudStorage = cloudStorage;
-    }
-
-    public SdxDatabaseRequest getExternalDatabase() {
-        return externalDatabase;
-    }
-
-    public void setExternalDatabase(SdxDatabaseRequest externalDatabase) {
-        this.externalDatabase = externalDatabase;
-    }
-
-    public SdxAwsRequest getAws() {
-        return aws;
-    }
-
-    public void setAws(SdxAwsRequest aws) {
-        this.aws = aws;
-    }
-
-    public boolean isEnableRangerRaz() {
-        return enableRangerRaz;
-    }
-
-    public void setEnableRangerRaz(boolean enableRangerRaz) {
-        this.enableRangerRaz = enableRangerRaz;
-    }
-
     @Override
     public String toString() {
-        return "SdxClusterRequest{" +
-                "clusterShape=" + clusterShape +
-                ", runtime='" + runtime + '\'' +
-                ", cloudStorage=" + cloudStorage +
-                ", externalDatabase=" + externalDatabase +
-                ", aws=" + aws +
-                ", enableRangerRaz=" + enableRangerRaz +
+        return "SdxClusterRequest{" + "base=" + super.toString() +
+                ", runtime=" + runtime +
                 '}';
     }
 }
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequestBase.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequestBase.java
@@ -1,0 +1,122 @@
+package com.sequenceiq.sdx.api.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+import com.sequenceiq.common.api.tag.request.TaggableRequest;
+
+public class SdxClusterRequestBase implements TaggableRequest {
+
+    @NotNull
+    private String environment;
+
+    @NotNull
+    private SdxClusterShape clusterShape;
+
+    private SdxCloudStorageRequest cloudStorage;
+
+    private SdxDatabaseRequest externalDatabase;
+
+    @Valid
+    private SdxAwsRequest aws;
+
+    private Map<String, String> tags;
+
+    private boolean enableRangerRaz;
+
+    public String getEnvironment() {
+        return environment;
+    }
+
+    public void setEnvironment(String environment) {
+        this.environment = environment;
+    }
+
+    public SdxClusterShape getClusterShape() {
+        return clusterShape;
+    }
+
+    public void setClusterShape(SdxClusterShape clusterShape) {
+        this.clusterShape = clusterShape;
+    }
+
+    public void setTags(Map<String, String> tags) {
+        this.tags = tags;
+    }
+
+    public Map<String, String> getTags() {
+        return tags;
+    }
+
+    @Override
+    public void addTag(String key, String value) {
+        initAndGetTags().put(key, value);
+    }
+
+    public Map<String, String> initAndGetTags() {
+        if (tags == null) {
+            tags = new HashMap<>();
+        }
+        return tags;
+    }
+
+    public void addTags(Map<String, String> tags) {
+        initAndGetTags().putAll(tags);
+    }
+
+    public SdxCloudStorageRequest getCloudStorage() {
+        return cloudStorage;
+    }
+
+    public void setCloudStorage(SdxCloudStorageRequest cloudStorage) {
+        this.cloudStorage = cloudStorage;
+    }
+
+    public SdxDatabaseRequest getExternalDatabase() {
+        return externalDatabase;
+    }
+
+    public void setExternalDatabase(SdxDatabaseRequest externalDatabase) {
+        this.externalDatabase = externalDatabase;
+    }
+
+    public SdxAwsRequest getAws() {
+        return aws;
+    }
+
+    public void setAws(SdxAwsRequest aws) {
+        this.aws = aws;
+    }
+
+    public boolean isEnableRangerRaz() {
+        return enableRangerRaz;
+    }
+
+    public void setEnableRangerRaz(boolean enableRangerRaz) {
+        this.enableRangerRaz = enableRangerRaz;
+    }
+
+    public void copyTo(SdxClusterRequestBase toInstance) {
+        toInstance.setEnvironment(environment);
+        toInstance.setClusterShape(clusterShape);
+        toInstance.setTags(tags);
+        toInstance.setCloudStorage(cloudStorage);
+        toInstance.setExternalDatabase(externalDatabase);
+        toInstance.setAws(aws);
+        toInstance.setEnableRangerRaz(enableRangerRaz);
+    }
+
+    @Override
+    public String toString() {
+        return "SdxClusterRequestBase{" +
+                "clusterShape=" + clusterShape +
+                ", cloudStorage=" + cloudStorage +
+                ", externalDatabase=" + externalDatabase +
+                ", aws=" + aws +
+                ", enableRangerRaz=" + enableRangerRaz +
+                '}';
+    }
+}
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
@@ -1,0 +1,50 @@
+package com.sequenceiq.sdx.api.model;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.image.ImageSettingsV4Request;
+
+public class SdxCustomClusterRequest extends SdxClusterRequestBase {
+    private ImageSettingsV4Request imageSettingsV4Request;
+
+    public ImageSettingsV4Request getImageSettingsV4Request() {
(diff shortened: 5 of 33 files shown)


LinaAtAustin:
I have run the local deployment to test by hardcoding the imagine info the SdxClusterRequest. However, I kept on getting external database deployment issue. Not sure if it is caused by the cost saving script or the image info I used {"catalog":"cloudbreak-default","id":"a0b2b750-9845-42e0-526f-fcba813f33ac"}. Will test again later

LinaAtAustin:
Screen Shot 2021-02-24 at 12 05 49 AM

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequest.java
```diff
@@ -29,6 +30,8 @@
 
     private boolean enableRangerRaz;
 
+    private ImageSettingsV4Request imageSettingsV4Request;
```
sodre90, line 33: Please create a separate endpoint with a request like "CustomSdxClusterRequest". The runtime version is meaningless for this request, so we can create an ancestor class for SdxClusterRequest and CustomSdxClusterRequest without runtime version field. There is some logic for runtime version and it should be reviewed.

akanto: I am also surprised that sdx-api is able reference classes from core-api

akanto:
This pull request requires junit tests.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -309,6 +309,7 @@ private StackV4Request getStackRequest(SdxClusterRequest sdxClusterRequest, Stac
                         ", runtime version: " + runtimeVersion);
             }
             stackRequest.getCluster().setRangerRazEnabled(sdxClusterRequest.isEnableRangerRaz());
+            stackRequest.setImage(sdxClusterRequest.getImageSettingsV4Request());
```
sodre90, line 312: What if somebody uses internal api and sends a stackv4request without image info and would like to use image provided in image field? It should be handled in 'else' branch.

LinaAtAustin: OK. I will handle that scenario as well. Thanks!

sodre90:
> This pull request requires junit tests.

I think e2e mock tests also required. 

This new field has risks. Until now cloudbreak core decided which image will be used to create the datalake stack but it won't be true anymore. We tried not to expose image id to datalake API because it can cause many support cases as customers can specify such dumb things like aws image for azure environment or base image etc. It was allowed only for internal use for QE.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequest.java
LinaAtAustin: introducing a base class, which does not have "runtime" field requires changing a lot of places. So I derive the SdxCustomClusterRequest from SdxClusterRequest and validate that the runtime field and image field cannot be set at the same time.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
LinaAtAustin: Handle that for both cases

LinaAtAustin:
I will create a new entitlement in CB-11344. Once that is committed, I will add entitlement check in SdxService to allow custom request only when the entitlement is on.

I have added several unit tests. Will run end-to-end test tomorrow.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
```diff
@@ -2,7 +2,7 @@
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.StackV4Request;
 
-public class SdxInternalClusterRequest extends SdxClusterRequest {
+public class SdxInternalClusterRequest extends SdxCustomClusterRequest {
```
lajosrodek, line 5: I wonder if it has any benefit (or dare I say, if it makes sense) to allow an extra image UUID override for the internal request. In particular, `SdxInternalClusterRequest` already contains a complete `StackV4Request` that may or may not have an image UUID inside. In any case, I do not think it is a good idea to override the ID inside the `StackV4Request` with the one coming from SdxCustomClusterRequest.

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxCustomController.java
```diff
@@ -0,0 +1,45 @@
...
+    @Inject
+    private SdxMetricService metricService;
+
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.CREATE_DATALAKE)
+    public SdxClusterResponse create(String name, @Valid SdxCustomClusterRequest createSdxClusterRequest) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        Pair<SdxCluster, FlowIdentifier> result = sdxService.createSdx(userCrn, name, createSdxClusterRequest, null);
+        SdxCluster sdxCluster = result.getLeft();
+        metricService.incrementMetricCounter(MetricType.INTERNAL_SDX_REQUESTED, sdxCluster);
```
lajosrodek, line 39: Do not we possibly need some either kind of metric here?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -190,6 +202,32 @@ public SdxCluster getByCrn(String userCrn, String clusterCrn) {
...
     }
 
+    public String getRuntimeVersionFromImageRequest(ImageSettingsV4Request imageSettingsV4Request) {
+        try {
+            LOGGER.info("Calling cloudbreak to get runtime version for the given image catalog {} and image id {}",
+                    imageSettingsV4Request.getCatalog(), imageSettingsV4Request.getId());
+            ImageV4Response imageV4Response =  ThreadBasedUserCrnProvider.doAsInternalActor(
+                    () -> {
+                        try {
+                            return imageCatalogV4Endpoint.getImageByNameAndId(0L, imageSettingsV4Request.getCatalog(), imageSettingsV4Request.getId());
```
lajosrodek, line 212: Not a big deal but could you please extract `0L` as a new constant `WORKSPACE_ID`?

LinaAtAustin:
@sodre90 "This new field has risks. " I have what cloud platform is and I get ImageV4Response using UUID to know the runtime. So I can check if the platform matches (ImageV4Response.imageSetsByProvider contains the mapping from cloud provider to the image sets). If you want to check other items, I can add those validation before creating the stack. In this way, we can achieve both flexibility and avoid wrong configuration as much as possible. 

If customer wants to use an image with Its UUID, I know it is not available when getting its info. So the DL creation can fail fast. customer can use a different image or request to put that image in CDP. Ideally, each customer can have its own image catalog, and Cloudera can charge for maintaining such additional service. The storage cost should be much smaller than what customers want to pay to get this service.

I will add mock tests.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
LinaAtAustin: if ((stackRequest.getImage() == null) && (sdxClusterRequest instanceof SdxCustomClusterRequest)) {
            stackRequest.setImage(((SdxCustomClusterRequest) sdxClusterRequest).getImageSettingsV4Request());
}

You can see the overwrite happens only if the StackV4Request contains no image settings. Are you still concerned? Peter wants to set image when user specifies it in internal request. I am trying to satisfy both opinions

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxCustomController.java
LinaAtAustin: I added a new metric type CUSTOM_SDX_REQUESTED to record the metric

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
LinaAtAustin: done

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
lajosrodek: Yes, I have seen the above logic but I think it shall not touch the image ID in `SdxInternalClusterRequest` whatsoever. If the user sending the req omitted the image ID, he/she probably wants to leave it that way. If he/she wanted to have an image ID explicitly, that could have been included in `SdxInternalClusterRequest.stackV4Request`, without relying on the UUID override of `SdxCustomClusterRequest`.

In short, my feeling is `SdxInternalClusterRequest` shall keep `SdxClusterRequest` as its superclass and not have the extras of `SdxCustomClusterRequest`. The logic you quoted will still work in both cases, but it will override the `null image` only when using `SdxCustomClusterRequest` instead of `SdxInternalClusterRequest`.

@sodre90 @doktoric @lnardai Please do not be shy and share your thoughts. 😃

lajosrodek: Actually, thinking a bit more I would even go further: I think `SdxCustomClusterRequest` should be a type not related to either `SdxClusterRequest` or `SdxInternalClusterRequest`. The cleanest approach would be to make `SdxCustomClusterRequest` be a subclass of `Object`, having all the fields of `SdxClusterRequest` except for the `runtime`. That way the conflict between `runtime` and `SdxCustomClusterRequest.imageSettingsV4Request` could be avoided entirely.

LinaAtAustin: @lajosrodek I prefer to not overwrite `image` in `SdxCustomClusterRequest.` In this way, user has full control of its value. Adding additional logic internally could be confusing to user. @sodre90 Are you OK with this? User who issues internal CLI can setting the image, why not let user get what they set directly?

@lajosrodek In the internal CLI, we have the same issue of conflict between `SdxInternalClusterRequest.StackV4Request.runtime` and `SdxInternalClusterRequest.ImageSettingsV4Request`. 

`The cleanest approach would be to make SdxCustomClusterRequest be a subclass of Object, having all the fields of SdxClusterRequest except for the runtime`
How about the following approach?
1. I define SdxCustomClusterRequest as you mentioned
2. then In SdxCustomController, I create `SdxClusterRequest` whose `runtime` is null and add a new input `ImageSettingsV4Request` for SdxService.createSdx(). In this way, we can avoid the conflict totally, and still minimize the code changes in many places.

lajosrodek: @LinaAtAustin Yes, `SdxInternalClusterRequest` should have also used `Object` as its superclass instead of `SdxClusterRequest`. The `SdxClusterRequest.runtime` is in conflict with `SdxInternalClusterRequest.StackV4Request` because the latter may prescribe a totally different runtime & image combination. We cannot help that here easily.

Instead, what I suggested are
1. Not to make `SdxInternalClusterRequest` be a subclass of `SdxCustomClusterRequest`; and
2. Make `SdxCustomClusterRequest` be a subclass of `Object`.

The controller change (i.e. converting `SdxCustomClusterRequest` into `SdxClusterRequest` and passing the additional `ImageSettingsV4Request` to `SdxService.createSdx()` sounds good. 👍  

Let us wait for some more comments as the rest of the team has not yet expressed its opinions.

core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/ImageCatalogV4Controller.java
```diff
@@ -142,4 +144,11 @@ public ImagesV4Response getImagesByName(Long workspaceId, @ResourceName String n
         Images images = imageCatalogService.getImagesByCatalogName(workspaceId, name, stackName, platform);
         return converterUtil.convert(images, ImagesV4Response.class);
     }
+
+    @Override
+    @CheckPermissionByResourceName(action = AuthorizationResourceAction.DESCRIBE_IMAGE_CATALOG)
+    public ImageV4Response getImageByNameAndId(Long workspaceId, @ResourceName String name, String imageId) throws Exception {
```
sodre90, line 150: can we rename it to getImageByCatalogNameAndImageId to make its name more specific?

LinaAtAustin: sure. It is done

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxInternalClusterRequest.java
sodre90: @LinaAtAustin We would like to avoid any "instanceof". As I see we have too many parameters that can be null, so I understand your point and your code, but as of now, I would recommend if we just add the ImageSettingsV4Request as a new parameter to createSdx method. It is ugly, but it is the easiest way for now and I will create a ticket for us to refactor / rearrange createSdx method.

I think we should create an abstract ancestor class for SdxClusterRequest like SdxClusterRequestBase and put all its parameters from SdxClusterRequest except runtime. SdxClusterRequest and SdxCustomClusterRequest should extend SdxClusterRequestBase. So now we have to add one more parameter to createSdx, the runtime field, and we have to change sdxClusterRequest paramater's class to SdxClusterRequestBase. I admit it is way more ugly now, we added 2 more parameters to createSdx method but on the API side we are fine. Maybe we can create a parameter class with parameters of createSdx but it is not a big improvement :) (userCrn, name, sdxClusterRequestBase, runtime, internalStackV4Request, imageSettingsV4Request).

You can implement your code without instanceOf, but with some null checks (our instanceOf hater soul will be in peace) and we will be fine on API side (no runtime field in custom request). The code is very ugly at the moment (the original one), it will be a little uglier but we can explode this method later, I will create a jira about this because I don't want to push this work to you. We were the ones who created this mess, it is our task to clean it.

I wouldn't touch SdxInternalClusterRequest (I think it shouldn't have runtime anyway, but if we remove it then we will break the API compatibility..)

lajosrodek: Good point, using a common ancestor for `SdxClusterRequest` and `SdxCustomClusterRequest` to extract common fields makes more sense. 😄

LinaAtAustin: I agree. Will update soon

LinaAtAustin:
@lajosrodek @sodre90 Can you take a look at the update? I have changed the API based on our discussion. I have not got time to add mock end-to-end test since I have been working on customer escalation since Friday and over the weekend. Thanks!

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequestBase.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @NotNull
+    private SdxClusterShape clusterShape;
+
+    private SdxCloudStorageRequest cloudStorage;
+
+    private SdxDatabaseRequest externalDatabase;
+
+    @Valid
+    private SdxAwsRequest aws;
```
lajosrodek, line 24: I know this is not part of the PR, but perhaps it made sense to add a proper `toString()` implementation for the classes involved here. `SdxClusterShape` is an enum, so it will work as expected, but the other three types will show the default (and pretty useless) `Object.toString()` output.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
```diff
@@ -0,0 +1,59 @@
...
+    public String getRuntime() {
+        return null;
+    }
+
+    @Override
+    public String toString() {
+        String baseString = super.toString();
+        String additionalString = ", imageSettingsV4Request=" + imageSettingsV4Request +  '}';
+        String ownString = baseString.replace("SdxClusterRequestBase", "SdxCustomClusterRequest");
+        return ownString.replace("}", additionalString);
```
lajosrodek, line 29: Like in `SdxClusterRequest`, I do not think it is a good idea to resort to string-wrestling.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public Pair<SdxClusterRequest, ImageSettingsV4Request> convert() throws IOException {
```
lajosrodek, line 32: Please use a more specific name, such as `asPair()`, `toPair()` or `convertToPair()`.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            @Override
+            public ImageSettingsV4Request getRight() {
+                return getImageSettingsV4Request();
+            }
+
+            @Override
+            public ImageSettingsV4Request setValue(ImageSettingsV4Request value) {
+                return value;
```
lajosrodek, line 55: Or perhaps throw `UnsupportedOperationException`?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -230,13 +277,30 @@ public SdxCluster getByNameInAccount(String userCrn, String name) {
...
+            throw new BadRequestException("SDX customer request is not valid.", ex);
+        }
+    }
+
     public Pair<SdxCluster, FlowIdentifier> createSdx(final String userCrn, final String name, final SdxClusterRequest sdxClusterRequest,
             final StackV4Request internalStackV4Request) {
+        return createSdx(userCrn, name, sdxClusterRequest, internalStackV4Request, null);
+    }
+
+    protected Pair<SdxCluster, FlowIdentifier> createSdx(final String userCrn, final String name, final SdxClusterRequest sdxClusterRequest,
```
lajosrodek, line 295: I feel this should be `private`.

lajosrodek (changes requested):
Thanks, looks alright. Some further minor comments.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequest.java
```diff
@@ -53,71 +12,10 @@ public void setRuntime(String runtime) {
...
-                "clusterShape=" + clusterShape +
-                ", runtime='" + runtime + '\'' +
-                ", cloudStorage=" + cloudStorage +
-                ", externalDatabase=" + externalDatabase +
-                ", aws=" + aws +
-                ", enableRangerRaz=" + enableRangerRaz +
-                '}';
+        String baseString = super.toString();
+        String ownString = baseString.replace("SdxClusterRequestBase", "SdxClusterRequest");
+        return ownString.replace("}", ", runtime='" + runtime + '}');
```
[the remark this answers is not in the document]
LinaAtAustin: that is good point. I will update. Thanks!

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
LinaAtAustin: I will update accordingly.

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
LinaAtAustin: I changed to convertToPair()

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxCustomClusterRequest.java
LinaAtAustin: I changed it to "throw exception "

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
LinaAtAustin: done

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterRequestBase.java
LinaAtAustin: I have created CB-11443 to track that effort. So I can focus on the feature in this PR.

lajosrodek (approved):
LGTM nice work!

LinaAtAustin:
@sodre90 I have added mock test. Can you take a look? Thanks!

LinaAtAustin:
DataLake created with custom image
The image catalog and ID
I enabled the entitlement CDP_DATA_LAKE_CUSTOM_IMAGE, set the catalog to be "cloudbreak-default" and image id to be "b512d50c-44ee-4e87-a5a3-a308ead95086". DL was created successfully using the custom image.

topolyai5 (commented):
Is this feature an internal or external feature?

if internal why don't you use the internal endpoint? if external why don't you integrate to the existing endpoint? Is it really need?

LinaAtAustin:
@topolyai5 
It is for beta cli. So it is not internal. It is hidden from normal customer. The reason for this to be in beta CLI is because we don't want to expose this to all customers, and would like only for PMs to use it for select customers.
The way to do this is to set the x-hidden-retention property in the yaml.. CLI update is tracked in https://jira.cloudera.com/browse/CDPCP-4045

@perdos suggested to use another endpoint as the request contains image settings, to avoid conflicting info, it does not have "runtime" field. This is to prevent mistake that customer could make and cause issue.

topolyai5 (changes requested):
As we talked in Slack channel, could you please move the endpoint operation to the common controller as an operation: /sdx/custom
I suggest, rename it to a more meaningful name, that refers to the operation

LinaAtAustin:
@topolyai5 @sodre90  
Peter, you have mentioned in https://github.com/hortonworks/cloudbreak/pull/10124#discussion_r581762031 and asked me to "Please create a separate endpoint with a request like "CustomSdxClusterRequest". The runtime version is meaningless for this request, so we can create an ancestor class for SdxClusterRequest and CustomSdxClusterRequest without runtime version field. There is some logic for runtime version and it should be reviewed."

Gergely is asking me to not create a separate endpoint, and move it back to the common controller. These are conflicting requests.
Can you discuss and decide what way I should go to finish this code change? I have very limited time before I am away next week.

@topolyai5 how important is it to remove this new endpoint? The concern of moving it to the same common controller is that new we have two different types of requests that contain different information to the same endpoint. It may be easier for customer to make mistake. 

The new feature in this code change is not intend to be open to the most customers as Karthik mentioned in https://jira.cloudera.com/browse/CDPCP-4045

having separate endpoint makes it easy to hide this feature, and only allow limited customer to access it.

LinaAtAustin:
Gergely, I saw you and Peter reached agreement. I am making this change now.

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
```diff
@@ -255,4 +256,10 @@ RangerCloudIdentitySyncStatus getRangerCloudIdentitySyncStatus(@PathParam("envCr
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = ROTATE_CERTIFICATES, nickname = "rotateAutoTlsCertificatesByCrn")
     FlowIdentifier rotateAutoTlsCertificatesByCrn(@PathParam("crn") String crn, @Valid CertificatesRotationV4Request rotateCertificateRequest);
+
+    @POST
+    @Path("/customimage/{name}")
```
topolyai5, line 261: Can you please follow the pattern? we use this pattern: `{name}/operation_name` or `{crn}/operation_name`. Our URL parser won't recognise the pattern.

LinaAtAustin: @topolyai5 Thanks for the comment. Will @Path("{name}/customimage") work? If so, I will make update tomorrow. 

Is this the only change I need to make? If not, can we have 30 minutes zoom session to discuss and get all changes we need at once, then I can work on it on Monday. Thanks!

topolyai5: I mean for this path: `{name}/custom_image`

topolyai5: maybe it can work with crn as well?

LinaAtAustin: @topolyai5 I don't think it work with crn. This is create function. the caller can only provide the DL name. The crn is not assigned yet. To get the details, we can use the crn, and actually reuse the existing function to do so. 

The added new function is only create() with customized image info.

topolyai5: yes, you are right

topolyai5 (approved):

PR #10123, opened 2021-02-23
CB-11307 Introduce 7.2.10 Cloudera Runtime

pkedvessy:
See detailed description in the commit message.

PR #10127, opened 2021-02-24
Intoduce 7.2.10 runtime

Bajzathd:


horadla23:
sorry if i am speaking from outside of the topic, but there is a PR for this already as I see
https://github.com/hortonworks/cloudbreak/pull/10123
cc @pkedvessy

Bajzathd:
@horadla23 this was created to verify that there is a bug that fails the integration tests, this is why it is just a draft. Closing it.

horadla23:
ohh, ok, makes sense, i just checked PRs and saw that there are two PR for this

PR #10130, opened 2021-02-24
CB-11328 Agen token copy fails when a minion is unavailable

lacikaaa:
With this commit if the copy fails we would only log it.
This is necessary because during downscale highstate is invoked thus agent token generation and copy is invoked.
If one of the instances is down, which doesn't have the token the downscale would.
With this change it won't block downscale anymore.
During upscale it shouldn't cause any issues, as agents without token won't start and it means the failure would still happen in the orchestration state a bit later.

The change includes turning off `set -e` during salt copy and evaluate the return code later. In some cases the `false` text is not included in the return value so we have to check the return code too.

Another change is adding the token generation log to the diagnostic bundle as if the output is too big, it won't be in the salt minion log.


oleewere (approved):
lgtm

schfeca75 (approved):

PR #10137, opened 2021-02-24
CB-11008 Tag validation for long and short tags also apply transforme…

doktoric:
…r. In case of GCP this is required.


biharitomi (approved):
LGTM

PR #10145, opened 2021-02-25
CB-8953 - fixed bug when some cloud provider api calls were lonely spans and not included in the flow traces

attilapalfi92:
…
Root cause was that there are 2 executors that we use to call cloud provider apis. These use new threads from their thread pools, therefore the tracing context is lost in them.

Initially I started passing the tracing context in 3 separate executor tasks and when I was there, I've done some refactoring.
Later I realised there is a better solution: passing the tracing context in a taskdecorator. I did that and kept the refactoring of ResourceCreationCallable, ResourceDeletionCallable & ResourceStopStartCallable classes.


bergerdenes (approved):
nit: typo
question: is it necessary to separate ResourceXXXCallable and ResourceXXXCallablePayload? does it have any added value?
apart from these LGTM

PR #10148, opened 2021-02-25
CB-11350 remove those minions which are in denied list and also in unaccepted list at the same time

sodre90:


lacikaaa (approved):

biharitomi (approved):
LGTM

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java
```diff
@@ -36,14 +36,32 @@ public MinionAcceptor(SaltConnector sc, List<Minion> minions, MinionFingerprintM
...
+    private void removeMinionIdsInBothDeniedAndUnacceptedState(List<String> deniedMinions, List<String> unacceptedMinions)
+            throws CloudbreakOrchestratorFailedException {
+        LOGGER.info("Unaccepted minions: {}", unacceptedMinions);
+        LOGGER.info("Denied minions: {}", deniedMinions);
+        List<String> minionIdsInBothDeniedAndUnacceptedState = deniedMinions.stream().distinct()
+                .filter(unacceptedMinions::contains).collect(Collectors.toList());
+        if (!minionIdsInBothDeniedAndUnacceptedState.isEmpty()) {
+            LOGGER.info("There are minions in denied and unaccepted state at the same time, lets remove them: " +
+                    minionIdsInBothDeniedAndUnacceptedState);
+            sc.wheel("key.delete", minionIdsInBothDeniedAndUnacceptedState, Object.class);
```
akanto, line 59: Shouldn't we introduce a method called Wheel.deleteKey?  Similar like as we do in  SaltStates.* Ther reason for that is delete.keys is used in multiple places in the code and we could have common logging  and maybe even error handling.

akanto (approved):

PR #10149, opened 2021-02-25
CB-11070 CDP PC missing firewall for egress/ingress

hreeve-cloudera:
Adds additional pass conditions to the logic that validates the AWS subnets provided for the public endpoint
gateway load balancer. Previously a subnet could only be used if it had an attached internet gateway. Now
it can be used if it passes one of the following three conditions:
  1. The subnet is tagged with "cdp-public-endpoint-gateway-lb" : "override".
  2. The subnet has an internet gateway attached.
  3. The subnet is routed through an AWS network firewall, and that firewall has an attached IGW.

Tested with unit tests and by enabling the feature with subnets with the override tag, and subnets routed
though a public network firewall.


biharitomi (approved):
LGTM

PR #10150, opened 2021-02-25
CB-11356 Handle LB with missing endpoint

hreeve-cloudera:
Adds a new filter for LB endpoint names in GatewayPublicEndpointManagementService that will ignore
any null or empty endpoint values. Also adds a supported platform condition in LoadBalancerConfigService,
to prevent unexpeced results on platforms that haven't been fully implemented yet.

Tested with unit tests.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -1,7 +1,9 @@
 package com.sequenceiq.cloudbreak.service;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -51,6 +53,8 @@
 
     private static final Set<Integer> DEFAULT_KNOX_PORTS = Set.of(443);
 
+    private static final Set<String> SUPPORTED_CLOUD_PLATFORMS = Set.of(AWS);
```
biharitomi, line 56: It could be a configurable value to be more future proof.

biharitomi (changes requested):
Otherwise it LGTM

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: Added `cb.loadBalancer.supportedPlatforms` setting to application.yml.

biharitomi (approved):
LGTM

PR #10151, opened 2021-02-25
CB-11358 Cannot create cluster in any region on UI if "af-south-1" region is enabled on AWS

schfeca75:
CB-11358 Cannot create cluster in any region on UI if "af-south-1" region is enabled on AWS

* Upgrade awssdk library to 1.11.961
* filter out unsupported regions by the awssdk library RegionUtil class in AWSPlatformResources.regions() method


doktoric:
as a followup a unit test would be nice. lgtm

doktoric (approved):

PR #10155, opened 2021-02-25
CB-11344: Add entitlement for customizing data lake image

LinaAtAustin:
-The flag in thunderhead is committed in ALTCORE-3511


jamisonbennett:
LGTM, but I think you might need to wait until the UMS with the entitlement gets promoted. I don't know what would happen if CB were promoted without the UMS change. That is why I am not yet giving the approval. If it defaults to "false" without throwing then we are probably good.

LinaAtAustin:
@jamisonbennett We can wait until tomorrow for CB team takes a look and by then, the entitlements in thunderhead should be there. Thanks!

mock-thunderhead/src/main/resources/application.yml
```diff
@@ -27,6 +27,7 @@ auth:
     upgrade.internalrepo.enable: true
     hbase.cloudstorage.enable: true
     datalake.efs.enable: false
+    datalake.customimage.enable: false
```
doktoric, line 30: in local mode I guess we should enable this. cc @sodre90

lajosrodek (commented):
LGTM. I had better await the decision regarding the MockUMS default, but otherwise this looks okay.

lajosrodek:
@LinaAtAustin @jamisonbennett  From CB's perspective, it is totally okay to merge & deploy this change before the UMS promotion. CB will be able to determine if the entitlement is present or not, that only depends on the UMS `getAccount` method that returns the list of entitlements granted for the queried tenant. So CB will keep seeing that the entitlement is absent until any tenant will have it granted actually.

The only critical place the UMS change promotion is required is granting & revoking entitlements. For instance, I got the errors below when attempting to grant / revoke `CDP_LRODEK_FOOBAR` in mow-dev:
```
...
An error occurred: Entitlement name CDP_LRODEK_FOOBAR is not valid (Status Code: 400; Error Code: INVALID_ARGUMENT; Service: coreadmin; Operation: grantEntitlement; Request ID: c4a64b52-025c-428c-950f-b37793184a2e;)

...
An error occurred: 'CDP_LRODEK_FOOBAR' entitlement does not exist in account '00180000011JIXPAA4' (Status Code: 404; Error Code: NOT_FOUND; Service: coreadmin; Operation: revokeEntitlement; Request ID: 98069fe7-e570-4653-83f5-069496dd4009;)
```
The latter error seems to be directly related to the entitlement grants instead of UMS knowing about this specific entitlement.

mock-thunderhead/src/main/resources/application.yml
LinaAtAustin: Can you confirm should I enable it or not? 

The actual feature is not in committed yet. So this setting does not take effect. Also, this should only affect the mock test, right?

LinaAtAustin:
@lajosrodek  Thanks for your comment. I have tried granting entitlement and it works.

```
lina.li@10326 cbd-local % /Users/lina.li/gerrit/thunderhead/tools/cdpcp-account-helper/build/venv/bin/cdp coreadmin get-account --account-id 9d74eee4-1cad-45d7-b645-7ccf9edbb73d --profile cdphelper_dev_admin_linali | grep "CDP_DATA_LAKE_CUSTOM_IMAGE"                

2021-02-25 15:48:11,155 - MainThread - cdpcli.clidriver - WARNING - You are running a INTERNAL release of the CDP CLI, which has different capabilities from the standard public release. Find the public release at: https://pypi.org/project/cdpcli/
                "entitlementName": "CDP_DATA_LAKE_CUSTOM_IMAGE",
```

kmanamcheri (commented):
LGTM
Ship it after you have addressed the comment about localmode.

mock-thunderhead/src/main/resources/application.yml
sodre90: we can enable it yes

sodre90 (approved):

mock-thunderhead/src/main/resources/application.yml
LinaAtAustin: I will do it in https://github.com/hortonworks/cloudbreak/pull/10124 to enable that.

PR #10158, opened 2021-02-25
CB-11327: Add Atlas keytabs as a fallback for medium duty.

brycederriso:
Quick fix for [CB-11327](https://jira.cloudera.com/browse/CB-11327).

Add an additional keytab to the look up chain.

Tested by running medium duty backup/restore locally.
Verified that I got successes back from the script:
```
+ doLog 'INFO Succesfully restored hive'
...
+ doLog 'INFO Succesfully restored ranger'
...
+ doLog 'INFO Completed restore.'
```

And `/var/log/dl_postgres_restore.log` confirms that restore was successful.

See also, [the second PR](https://github.com/hortonworks/cloudbreak/pull/10157), which is not a hack.


kkalvagadda1 (approved):
This is a workaround. Please create Jira to implement functionality where backup can be performed on any host group.

brycederriso:
[Here's the Jira to clean up.](https://jira.cloudera.com/browse/CB-11382)

PR #10169, opened 2021-02-26
CB-11142 SDX HA Minion not being accepted by master

orngejaket:
When a salt master in SDX is being repaired, it's old minion fingerprint is
still on the other salt master and is in a denied state.  So we delete that
to allow the new minion key to be accepted.

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.orchestrator.salt.poller.join;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


sxxgrc (approved):

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java
```diff
@@ -65,6 +66,13 @@ private void removeMinionIdsInBothDeniedAndUnacceptedState(SaltConnector sc, Lis
             throw new CloudbreakOrchestratorFailedException("There were minions in denied and unaccepted state at the same time: " +
                     minionIdsInBothDeniedAndUnacceptedState);
         }
+
+        ArrayList<String> minionIdsOnlyDenied = new ArrayList<String>(deniedMinions);
+        minionIdsOnlyDenied.removeAll(unacceptedMinions);
+        if (!minionIdsOnlyDenied.isEmpty()) {
+            LOGGER.info("There are minions in denied state, removing: ", minionIdsOnlyDenied);
+            sc.wheel("key.delete", minionIdsOnlyDenied, Object.class);
+        }
```
lacikaaa, line 75: could you move this to a separate method, and I think this should be called outside of `removeMinionIdsInBothDeniedAndUnacceptedState`

I would create a method called `cleanupMinionIds` which would call this new method and `removeMinionIdsInBothDeniedAndUnacceptedState`

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java
```diff
@@ -43,15 +44,21 @@ public void acceptMinions() throws CloudbreakOrchestratorFailedException {
...
             } else {
-                LOGGER.info("No unaccepted minions found on master");
+                LOGGER.info("No unaccepted minions found on master: [{}]", sc.getHostname());
             }
         }
     }
 
+    private void cleanupMinionIds(SaltConnector sc, List<String> deniedMinions, List<String> unacceptedMinions)
+            throws CloudbreakOrchestratorFailedException {
+        removeMinionIdsInBothDeniedAndUnacceptedState(sc, deniedMinions, unacceptedMinions);
```
sodre90, line 58: As I see with this change we remove every denied minions (first step is delete denied and unaccepted, then delete other denied). So I think it would be better to delete every denied minion in one step. Code will be easier to understand and it will do only one request to salt master. You can throw exceptionin this case as it was thrown in unaccepted+denied case, something like this: 

new CloudbreakOrchestratorFailedException("There were minions in denied state: " + deniedMinions);

orngejaket: Why would I need to throw an exception?

Ok, I can build a list of (denied && unaccepted) || denied

orngejaket: So upon further thinking, each salt call is fixing a different problem.  So maybe we should leave as is.  If we run into a scenario where we have both being executed and it doesn't fix the issue, it'd be good to log what it was doing rather than a single large call with less logging.

sodre90: If you remove the dead minions then "target" nodes' keys will not be accepted in this iteration. It will not cause any trouble, but we will do the minion ip addr collection, but it will fail for sure because the former denied minions will be missing. So this was the reason I thrown an exception when I removed minions both in unaccepted and denied state. But I don't push it, with this we can skip an unnecessary call.

sodre90: Ok, we can leave it as is, it is an edge case so we should not remove any minions in normal circumstances.

sodre90 (approved):

PR #10170, opened 2021-02-26
CB-11372: Add Atlas keytabs as a fallback for medium duty.

brycederriso:
Closes [CB-11327](https://jira.cloudera.com/browse/CB-11327).
Look for Atlas keytabs on backup and restore scripts.

This is in addition to the other keytabs that have already been checked.

Tested locally by running backup and restore using a medium duty datalake.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -68,6 +68,7 @@ run_kinit() {
   HDFS_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*hdfs*" | head -n 1)
   HBASE_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -a \( -path "*hbase-REGIONSERVER*" -o -path "*hbase-MASTER*" \) -a -type f | head -n 1)
   SOLR_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*solr-SOLR_SERVER*" | head -n 1)
+  ATLAS_KEYTAB=$(find /run/cloudera-scm-agent/process/ -name "*.keytab" -path "*atlas-ATLAS_SERVER*" | head -n 1)
 
   if kinit_as hdfs "$HDFS_KEYTAB"; then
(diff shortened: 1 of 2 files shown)


lacikaaa (commented):
I'm going to address the part changed:
- I see some code duplication here. why are these not in a common file?
- this doesn't seems like a sustainable thing to do. finding random keytab and trying to kinit with. Isn't there any better method? Why is anything else than hdfs is necessary? it will be a hdfs command which copies the files over, seems odd doing it in the name of atlas

brycederriso:
You're right @lacikaaa, this is not the way we want to do backup/restore moving forward.

We have an [additional ticket](https://jira.cloudera.com/browse/CB-11382) to fix the nodes on which backup restore is run -- essentially we're having to do this because the backup/restore scripts are being run on the Gateway nodes in medium duty, but they should be run on one of the "master" nodes.

Medium duty restore is currently broken, which is a blocker for customers trying to move off of light duty data lakes. Hence, the quick-fix in this PR.

brycederriso:
> I see some code duplication here. why are these not in a common file?

I would argue against extracting the common code just because it's only shared twice, across two 150 line files, and the import facilities in bash are just `sourceing` the script into the global namespace.

I'm not sure it's worth it yet to move the common code out. We can open an improvement Jira to address that refactoring though.

brycederriso:
I think it's worth mentioning again that the context around this is that without restore to medium duty data lakes we can't upgrade customers off of light duty.

We had the medium duty restore working a few months ago, but the CM roles assigned to the gateway nodes have changed which broke things.

We _are_ working on a better solution to this by more precisely targeting the `core` master nodes on the Java `orchestrator-salt` side.

lacikaaa (approved):
I'm fine with letting this in. But I don't think duplicating everything into 2 different script is more maintainable than sourcing a common one with 3-4 functions, as this technology is used by Linux distros eg to manage networks

lnardai (approved):
LGTM!

lacikaaa:
your PR was agains 2.39.0 instead of CB-2.39.0
please reopen to the right one

lnardai:
Although, in retrospect found out the the branch settings are incorrect:
the correct branch is CB-2.39.0.

lnardai:
I've restored the branch, please reopen the PR, and we'll merge it.

brycederriso:
Agh! Sorry, @lnardai @lacikaaa here's the new one.
https://github.com/hortonworks/cloudbreak/pull/10180

PR #10042, opened 2021-02-15
CB-11043 Introduce 7.2.9 Cloudera Runtime

pkedvessy:
See detailed description in the commit message.

PR #10172, opened 2021-02-28
CFM-1739 Fix scaling up for Data Flow Light Duty datahub cluster. Added new hostGroup to cluster template, with nodeCount=0 and only NiFi, to make NiFi scale up possible for Light Duty cluster. Also set SSL/TLS properties in related blueprints to fix issue with TLS/SSL not enabled for new hostGroup.

vbrodetskyi:

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.8/cdp-flow-management-small.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.8/cdp-flow-management-small.bp
@@ -49,6 +49,30 @@
             "base": true,
             "configs": [
               {
+                "name": "ssl_enabled",
+                "value": "true"
+              },
+              {
+                "name": "ssl_server_keystore_location",
+                "value": "{{CM_AUTO_TLS}}"
+              },
+              {
+                "name": "ssl_server_keystore_password",
+                "value": "{{CM_AUTO_TLS}}"
+              },
+              {
+                "name": "ssl_server_keystore_keypassword",
+                "value": "{{CM_AUTO_TLS}}"
+              },
+              {
+                "name": "ssl_client_truststore_location",
+                "value": "{{CM_AUTO_TLS}}"
+              },
+              {
+                "name": "ssl_client_truststore_password",
+                "value": "{{CM_AUTO_TLS}}"
+              },
+              {
                 "name": "xml.authorizers.userGroupProvider.file-user-group-provider.enabled",
                 "value": "true"
               },
@@ -278,6 +302,14 @@
         "roleConfigGroupsRefNames": [
           "nifi-NIFI_NODE-BASE",
           "zookeeper-SERVER-BASE",
+          "core_settings-STORAGEOPERATIONS-BASE"
+        ]
+      },
+      {
+        "refName": "nifi_scaling",
+        "cardinality": 0,
+        "roleConfigGroupsRefNames": [
+          "nifi-NIFI_NODE-BASE",
           "core_settings-STORAGEOPERATIONS-BASE"
         ]
       },
--- a/core/src/main/resources/defaults/clustertemplates/7.2.8/aws/flow-management-small.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.8/aws/flow-management-small.json
@@ -31,6 +31,25 @@
         }
       },
       {
+        "nodeCount": 0,
+        "name": "nifi_scaling",
+        "type": "CORE",
+        "recoveryMode": "MANUAL",
+        "template": {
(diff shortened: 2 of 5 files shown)


lajosrodek (approved):
LGTM

lajosrodek:
@vbrodetskyi  This looks okay, but only affects 7.2.8. There are already templates for 7.2.9 (#10042) and 7.2.10 (#10123). Could you please take care of them as well (perhaps in a separate PR)?

vbrodetskyi:
@lajosrodek thank you for review, i've added changed for 7.2.9 and 7.2.10.

adenes (approved):
Thanks @vbrodetskyi , LGTM

core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
```diff
@@ -48,6 +48,30 @@
             "roleType": "NIFI_NODE",
             "base": true,
             "configs": [
+              {
+                "name": "ssl_enabled",
+                "value": "true"
+              },
+              {
+                "name": "ssl_server_keystore_location",
```
doktoric, line 56: this will only helps us in case of default templates

vbrodetskyi: @doktoric yep, agree, i think that information could be added to documentation and also we should double check CM will resolve related jira to make that workable for non default cases.
@pvillard31 what do you think about that?

doktoric: I think we need an ETA on the CM fix because we dont want to add such thing in the future.

vbrodetskyi: @doktoric agree that it should be fixed by CM for future releases but anyway ssl/tls properties that were added, should not break anything, correct?

pvillard31: Do we have a proper JIRA to track this with the CM team? I think Matt S. said that it would be fixed in the next release of CM but we need confirmation.

What does it mean that it'll only work "in case of default templates"? I would not expect anyone to create a template with an empty group for scaling, no? What should be documented exactly?

vbrodetskyi: @pvillard31 as i understand it should be fixed here https://jira.cloudera.com/browse/OPSAPS-54928, but there is no fix version and any ETA in this jira. About "in case of default templates", as i understand some customers using ability to create their own templates for datahub clusters, so scale up could fail for them, because they don't know about related ssl/tls properties and new hostGroup with cardinality 0 should be added. But if we expect users will not create custom templates, then nothing to document. @doktoric what do you think? Could this patch be merged?

doktoric: OPSAPS-59770 is the ticket what I raised with blocker.
@vbrodetskyi I can merge but honestly I would like to remove this as soon as possible. So I would like to see an ETA on the CM ticket. If we dont get a fix for it then we need to add this configuration to the next versions as well.

vbrodetskyi: @doktoric agree, i will remove this fix from 7.2.9 and 7.2.10. I've added comment about related OPSAPS jira https://jira.cloudera.com/browse/CFM-1739?focusedCommentId=4351063&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4351063.

vbrodetskyi: @doktoric i've updated patch and all checks have passed. If you agree, could you please merge changes? Also could you please clarify when they will be available on mow-dev? Thank you.

doktoric (approved):

doktoric:
please squash into one commit

vbrodetskyi:
> please squash into one commit

@doktoric that could be done by "Squash and merge pull request", isn't it?

lajosrodek:
@vbrodetskyi  "Squash and merge" is disabled in the CB repo on purpose, "Rebase and merge" is the usual variant we pick when merging PRs.

vbrodetskyi:
@lajosrodek got it, thank you. Will do.

vbrodetskyi:
@lajosrodek @doktoric done.

PR #10175, opened 2021-03-01
CB-10621 DataHUB and DataLake node status api

daszabo:


oleewere:
I will close this as the other version of the change has been already merged

PR #10177, opened 2021-03-01
CB-11152: Add encryptionKeyUrl CLI param in Azure environment creation

juhi-09:
CB-11152: Add encryptionKeyUrl CLI param in Azure environment creation
This part of the code stores the encryptionKeyUrl in column 'encryption_keyUrl'.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
@@ -37,6 +37,7 @@ import com.sequenceiq.environment.api.v1.environment.model.request.aws.AwsEnviro
 import com.sequenceiq.environment.api.v1.environment.model.request.aws.AwsFreeIpaParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.aws.S3GuardRequestParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureEnvironmentParameters;
+import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureResourceEncryptionParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureResourceGroup;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.ResourceGroupUsage;
 import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentCrnResponse;
@@ -53,6 +54,7 @@ import com.sequenceiq.environment.environment.dto.LocationDto;
 import com.sequenceiq.environment.environment.dto.SecurityAccessDto;
 import com.sequenceiq.environment.environment.dto.telemetry.EnvironmentFeatures;
 import com.sequenceiq.environment.network.dto.NetworkDto;
+import com.sequenceiq.environment.parameter.dto.AzureResourceEncryptionParametersDto;
 import com.sequenceiq.environment.parameter.dto.ResourceGroupCreation;
 import com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern;
 import com.sequenceiq.environment.parameter.dto.AwsParametersDto;
@@ -208,8 +210,23 @@ public class EnvironmentApiConverter {
                                         || !StringUtils.isEmpty(resourceGroup.getName()))
                                 .map(this::azureResourceGroupToAzureResourceGroupDto)
                                 .orElse(buildDefaultResourceGroupDto())
-                ).build();
+                )
+                .withEncryptionParameters(
+                        Optional.ofNullable(azureEnvironmentParameters)
+                                .map(AzureEnvironmentParameters::getResourceEncryptionParameters)
+                                .filter(resourceEncryptionParameters -> Objects.nonNull(resourceEncryptionParameters.getEncryptionKeyUrl()))
+                                .map(this::azureResourceEncryptionParametersToAzureEncryptionParametersDto)
+                                .orElse(null)
+                )
+                .build();
 
+    }
+
+    private AzureResourceEncryptionParametersDto azureResourceEncryptionParametersToAzureEncryptionParametersDto(
+            AzureResourceEncryptionParameters azureResourceEncryptionParameters) {
+        return AzureResourceEncryptionParametersDto.builder()
+                .withEncryptionKeyUrl(azureResourceEncryptionParameters.getEncryptionKeyUrl())
+                .build();
     }
 
     private AzureResourceGroupDto buildDefaultResourceGroupDto() {
@@ -332,8 +349,8 @@ public class EnvironmentApiConverter {
 
     public EnvironmentLoadBalancerDto initLoadBalancerDto(EnvironmentLoadBalancerUpdateRequest request) {
         return EnvironmentLoadBalancerDto.builder()
-            .withEndpointAccessGateway(request.getPublicEndpointAccessGateway())
-            .withEndpointGatewaySubnetIds(request.getSubnetIds())
-            .build();
+                .withEndpointAccessGateway(request.getPublicEndpointAccessGateway())
+                .withEndpointGatewaySubnetIds(request.getSubnetIds())
+                .build();
     }
 }
--- a/environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentResponseConverter.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentResponseConverter.java
@@ -9,9 +9,11 @@ import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.util.NullUtil;
 import com.sequenceiq.common.api.type.Tunnel;
+import com.sequenceiq.cloudbreak.common.exception.BadRequestException;
 import com.sequenceiq.environment.api.v1.environment.model.request.aws.AwsEnvironmentParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.aws.S3GuardRequestParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureEnvironmentParameters;
+import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureResourceEncryptionParameters;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.AzureResourceGroup;
 import com.sequenceiq.environment.api.v1.environment.model.request.azure.ResourceGroupUsage;
 import com.sequenceiq.environment.api.v1.environment.model.request.gcp.GcpEnvironmentParameters;
@@ -31,11 +33,12 @@ import com.sequenceiq.environment.environment.dto.EnvironmentDto;
 import com.sequenceiq.environment.environment.dto.LocationDto;
 import com.sequenceiq.environment.environment.dto.SecurityAccessDto;
 import com.sequenceiq.environment.network.dto.NetworkDto;
-import com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern;
 import com.sequenceiq.environment.parameter.dto.AwsParametersDto;
 import com.sequenceiq.environment.parameter.dto.AzureParametersDto;
+import com.sequenceiq.environment.parameter.dto.AzureResourceEncryptionParametersDto;
 import com.sequenceiq.environment.parameter.dto.AzureResourceGroupDto;
 import com.sequenceiq.environment.parameter.dto.ParametersDto;
+import com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern;
 import com.sequenceiq.environment.proxy.v1.converter.ProxyConfigToProxyResponseConverter;
 
 @Component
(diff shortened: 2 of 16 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -208,8 +210,23 @@ private AzureParametersDto azureParamsToAzureParametersDto(AzureEnvironmentParam
...
+                                .map(AzureEnvironmentParameters::getResourceEncryptionParameters)
+                                .filter(resourceEncryptionParameters -> Objects.nonNull(resourceEncryptionParameters.getKeyUrl()))
+                                .map(this::azureResourceEncryptionParamtersToAzureEncryptionParametersDto)
+                                .orElse(null)
+                )
+                .build();
+
+    }
 
+    private AzureResourceEncryptionParametersDto azureResourceEncryptionParamtersToAzureEncryptionParametersDto(
```
lajosrodek, line 225: Typo: `Paramters` -> `Parameters`.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
```diff
@@ -25,6 +25,9 @@
     @Convert(converter = ResourceGroupUsagePatternConverter.class)
     private ResourceGroupUsagePattern resourceGroupUsagePattern;
 
+    @Column(name = "encryption_parameter_keyUrl")
+    private String resourceEncryptionParameterKeyUrl;
```
lajosrodek, line 29: What about Vault usage? I though we agreed that encryption key URLs should be also kept there like the DH EBS encryption KMS ARNs. So this probably should be some `Secret<String>` thing or however it is done elsewhere.

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
```diff
@@ -124,11 +127,21 @@ private ValidationResult validateEntitlement(ValidationResultBuilder validationR
...
                 default:
-                    return validationResultBuilder.build();
+                    break;
+            }
+        }
+
+        String keyUrl = azureResourceEncryptionParametersDto.getKeyUrl();
+        if (Objects.nonNull(keyUrl)) {
+            if (!entitlementService.isAzureDiskSSEWithCMKEnabled(accountId)) {
+                LOGGER.info("Invalid request, isAzureDiskSSEWithCMKEnabled entitlement turned off for account {}", accountId);
```
doktoric, line 137: I think we should highlight here the entitlement name

doktoric: btw do we need any additional Azure policy for using this feature?

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureResourceEncryptionParametersDto.java
```diff
@@ -0,0 +1,37 @@
+package com.sequenceiq.environment.parameter.dto;
+
+public class AzureResourceEncryptionParametersDto {
+    private final String keyUrl;
```
lajosrodek, line 4: If we decide in favor of using Vault for this, I wonder if it should already happen in the DTO or only later in the entity. Should be checked what the convention is.

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureResourceEncryptionParametersDto.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AzureResourceEncryptionParametersDto(AzureResourceEncryptionParametersDto.Builder builder) {
```
lajosrodek, line 6: This should be `private`. Only the builder shall have access to the constructor.

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
doktoric: I think we should also validate in the current creation flow that the key is really exist on azure side

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureParametersDto.java
```diff
@@ -4,37 +4,51 @@
 
     private AzureResourceGroupDto azureResourceGroupDto;
 
+    private AzureResourceEncryptionParametersDto azureResourceEncryptionParametersDto;
+
     public AzureParametersDto(Builder builder) {
```
lajosrodek, line 9: I know it is not your change but this constructor should be `private`.

lajosrodek (changes requested):
Some comments.

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
juhi-09: Updated the error to include the Entitlement, `isAzureDiskSSEWithCMKEnabled`.
I don't think additional policies are required, figuring out how to include the key validation, will update either this PR or following one(for CB-11152 only) to have the validation, that's in the to-do list.

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureResourceEncryptionParametersDto.java
juhi-09: Working on this part, will update the PR.

lajosrodek:
@juhi-09 Ah, just came to my mind: Since you added a new field to an entity (`AzureParameters.resourceEncryptionParameterKeyUrl`), you also need a DB schema update. That must be included in this PR, otherwise the first attempt of persisting `AzureParameters` will fail with an SQL error (no such column etc.)

The DB schema change, obviously, depends on the field datatype, so you shall first take care of the `String` to `Secret` switch and then add the DB schema update script with the right column definition.

The Key existence validation inside the Env create flow may come in a separate PR, because that will also need some changes in `cloud-azure`, so is not a 5-minute task. 😄

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
lajosrodek, line 137: No, the entitlement is called `CDP_CB_AZURE_DISK_SSE_WITH_CMK`. Please use that name here and also in the validation error below. Furthermore, this log shall be an `error()`.

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureResourceEncryptionParametersDto.java
lajosrodek: Okay, so let us do the `Secret` conversion only at the entity level.

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentApiConverter.java
```diff
@@ -208,8 +210,23 @@ private AzureParametersDto azureParamsToAzureParametersDto(AzureEnvironmentParam
                                         || !StringUtils.isEmpty(resourceGroup.getName()))
                                 .map(this::azureResourceGroupToAzureResourceGroupDto)
                                 .orElse(buildDefaultResourceGroupDto())
-                ).build();
+                )
```
lajosrodek, line 213: Please add unit tests for the changed parts of `EnvironmentApiConverter`.

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentResponseConverter.java
```diff
@@ -174,14 +176,18 @@ private AwsEnvironmentParameters awsEnvParamsToAwsEnvironmentParams(ParametersDt
     }
 
     private AzureEnvironmentParameters azureEnvParamsToAzureEnvironmentParams(ParametersDto parameters) {
-        return Optional.ofNullable(parameters.getAzureParametersDto())
+        AzureResourceGroupDto resourceGroupDto = Optional.ofNullable(parameters.getAzureParametersDto())
```
lajosrodek, line 179: Please add unit tests for the changed parts of `EnvironmentResponseConverter`.

environment/src/main/java/com/sequenceiq/environment/parameters/v1/converter/AzureEnvironmentParametersConverter.java
```diff
@@ -43,6 +44,10 @@ protected void postConvert(BaseParameters baseParameters, Environment environmen
                 .map(AzureParametersDto::getAzureResourceGroupDto)
                 .map(AzureResourceGroupDto::getResourceGroupUsagePattern)
                 .orElse(null));
+        azureParameters.setResourceEncryptionParameterKeyUrl(azureParametersDto
```
lajosrodek, line 47: Please add unit tests for the changed parts of `AzureEnvironmentParametersConverter`.

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureParametersDto.java
```diff
@@ -4,37 +4,51 @@
 
     private AzureResourceGroupDto azureResourceGroupDto;
 
-    public AzureParametersDto(Builder builder) {
+    private AzureResourceEncryptionParametersDto azureResourceEncryptionParametersDto;
+
+    private AzureParametersDto(Builder builder) {
         azureResourceGroupDto = builder.azureResourceGroupDto;
+        azureResourceEncryptionParametersDto = builder.azureResourceEncryptionParametersDto;
```
lajosrodek, line 11: Please add unit tests for the changed parts of `AzureParametersDto` (at least test the builder & getter).

structuredevent-model/src/main/java/com/sequenceiq/environment/parameter/dto/AzureResourceEncryptionParametersDto.java
```diff
@@ -0,0 +1,37 @@
+package com.sequenceiq.environment.parameter.dto;
+
+public class AzureResourceEncryptionParametersDto {
+    private final String keyUrl;
+
+    private AzureResourceEncryptionParametersDto(AzureResourceEncryptionParametersDto.Builder builder) {
```
lajosrodek, line 6: Please add unit tests for `AzureResourceEncryptionParametersDto` (at least test the builder & getter).

lajosrodek (changes requested):
Some further requests.

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
juhi-09: Shall I consider changing above log level as well for `singleResourceGroupDedicatedStorageAccountEnabled` and `singleResourceGroupDeploymentEnabled`.
Though I think, it is logged as INFO here and later passed as error to be validated by the `AzureParameterValidatorTest.java ` if error is present - no further occurrences of validation are seen.

lajosrodek: Ah, good point. Okay, so we can keep using `info()`, perhaps an `error()` would be an overkill. But I agree that the above two log messages should be corrected to contain the right entitlement name.

lajosrodek (approved):
LGTM, only a small comment on correcting log messages.

environment/src/main/resources/schema/app/20210302145523_CB-11152:_Add_keyUrl_CLI_param_in_Azure_environment_creation.sql
```diff
@@ -0,0 +1,7 @@
+-- // CB-11152: Add keyUrl CLI param in Azure environment creation
+-- Migration SQL that makes the change goes here.
+ALTER TABLE environment_parameters ADD COLUMN IF NOT EXISTS encryption_parameter_keyUrl varchar(255);
+
+-- //@UNDO
+-- SQL to undo the change goes here.
+ALTER TABLE environment_parameters DROP COLUMN IF EXISTS encryption_parameter_keyUrl;
```
lajosrodek, line 7: With the above entity change, `encryption_parameter_keyUrl` shall be replaced with `encryption_keyUrl`. Be sure to first roll back the original SQL script in your local DB before renaming the column here.

lajosrodek (changes requested):
Thanks for the rename. One ask:

environment/src/main/resources/schema/app/20210302145523_CB-11152:_Add_keyUrl_CLI_param_in_Azure_environment_creation.sql
```diff
@@ -0,0 +1,7 @@
+-- // CB-11152: Add keyUrl CLI param in Azure environment creation
+-- Migration SQL that makes the change goes here.
+ALTER TABLE environment_parameters ADD COLUMN IF NOT EXISTS encryption_keyUrl varchar(255);
```
doktoric, line 3: 255 is always enough?

environment/src/main/java/com/sequenceiq/environment/environment/v1/converter/EnvironmentResponseConverter.java
```diff
@@ -219,8 +232,8 @@ private ResourceGroupUsage resourceGroupUsagePatternToResourceGroupUsage(Resourc
                 return ResourceGroupUsage.MULTIPLE;
             case USE_SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT:
                 return ResourceGroupUsage.SINGLE_WITH_DEDICATED_STORAGE_ACCOUNT;
-                default:
-                    throw new RuntimeException("Unknown resource group usage pattern: %s" + resourceGroupUsagePattern);
+            default:
+                throw new RuntimeException("Unknown resource group usage pattern: %s" + resourceGroupUsagePattern);
```
biharitomi, line 236: I know this is not new code, but why is this a Runtime exception?
I think we should use an exception that indicates that the provided value illegal so maybe bad request should be used here if this is only a request time validation or an illegalstate exception otherwise.

environment/src/main/java/com/sequenceiq/environment/parameters/validation/validators/parameter/AzureParameterValidator.java
```diff
@@ -124,11 +127,21 @@ private ValidationResult validateEntitlement(ValidationResultBuilder validationR
...
+            }
+        }
+
+        String encryptionKeyUrl = azureResourceEncryptionParametersDto.getEncryptionKeyUrl();
+        if (Objects.nonNull(encryptionKeyUrl)) {
+            if (!entitlementService.isAzureDiskSSEWithCMKEnabled(accountId)) {
+                LOGGER.info("Invalid request, CDP_CB_AZURE_DISK_SSE_WITH_CMK entitlement turned off for account {}", accountId);
+                return validationResultBuilder.error(
+                        "You specified encryptionKeyUrl to use Server Side Encryption for Azure Managed disks with CMK, "
+                                + "but that feature is currently disabled. Enable 'CDP_CB_AZURE_DISK_SSE_WITH_CMK' to use SSE with CMK.").
```
biharitomi, line 140: I would prefer a more user-friendly error message here, that emphasise that the Server Side Encryption related entitlement is not enabled for your account and please contact with us. I think `CDP_CB_AZURE_DISK_SSE_WITH_CMK` means nothing for a customer as they couldn't set entitlement explicitly AFAIK.

biharitomi (changes requested):

doktoric (approved):

biharitomi (approved):

lajosrodek (approved):
LGTM

PR #10181, opened 2021-03-01
CB-6425 use salt to fetch reachable nodes instead of our instancemeta…

sodre90:
…data states


horadla23 (approved):

akanto:
Looks good, but, a few lines of unit test could not harm

akanto (approved):

PR #10184, opened 2021-03-01
CB-10587: Fix FreeIPA replica install failure so that it cleans up ...

jamisonbennett:
... prior replication agreements

Fix FreeIPA replica install failure so that it cleans up prior
replication agreements. This handles the case where the
`ipa-replica-install` needs to be retried but the replication
agreements to the new node are already stored in the other FreeIPA
nodes.

This was tested by manually causing the replica installation to fail
after it completed the installlation so that the step would be
repeated.


lacikaaa:
could you add some log messages?

lacikaaa (approved):

PR #10185, opened 2021-03-02
Decrease the size of DB for DH in AWS to 100G

satyabollineni:
The size of external DB for all clouds today for Datahub clusters today is 1000GB which is too high and costly for customer since most of that is going in vacuum. In order to arrive at better defaults ,changing the External DB size to 100G. Will create a separate PR for Azure and GCP as well.

Before the change: 

image

Post the change:

image

### DIFF
--- a/core/src/main/resources/externaldatabase/aws/database-template.json
+++ b/core/src/main/resources/externaldatabase/aws/database-template.json
@@ -1,5 +1,5 @@
 {
   "instanceType": "db.m5.large",
   "vendor": "postgres",
-  "volumeSize": 1000
+  "volumeSize": 100
 }


cegganesh84 (approved):
LGTM. Needs 2.39 owner approval. CC: @doktoric

akanto (approved):

doktoric:
why we only do this for AWS?

satyabollineni:
@doktoric This change need to be tested on GCP and AZURE as well but I have to setup my accounts on these clouds. So since Porshe is pushing support team for this change, we decided make a separate commit for AZURE and GCP.

PR #10186, opened 2021-03-02
CB-11433 Cannot propagate the downscale error to the structured event…

topolyai5:
…. We just write the common message to the event:  but we can add the detailed error message from the exception


doktoric (approved):

biharitomi (approved):
LGTM

PR #10189, opened 2021-03-02
CB-11391 Add some information to the flow accept logging. we need to …

topolyai5:
…add more valuable information like the event.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/logger/LogContextAspects.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/logger/LogContextAspects.java
@@ -8,6 +8,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.service.events.CloudbreakCompositeEvent;
+
+import reactor.bus.Event;
+
 @Component("CloudbreakLogContextAspects")
 @Aspect
 public class LogContextAspects {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/logger/LogContextAspects.java
```diff
@@ -20,6 +24,9 @@ public void interceptCloudbreakEventHandlerAcceptMethod() {
     @Before("com.sequenceiq.cloudbreak.logger.LogContextAspects.interceptCloudbreakEventHandlerAcceptMethod()")
     public void buildLogContextForReactorHandler(JoinPoint joinPoint) {
         com.sequenceiq.flow.logger.LogContextAspects.buildLogContextForEventHandler(joinPoint);
-        LOGGER.debug("A Reactor event handler's 'accept' method has been intercepted: {}, MDC logger context is built.", joinPoint.toShortString());
+        Event<CloudbreakCompositeEvent> event = (Event<CloudbreakCompositeEvent>) joinPoint.getArgs()[0];
```
lacikaaa, line 27: isn't it a bit risky to get the first item? can't this throw a nullpointer?
also, are you sure about the cast? shall we check it before?

topolyai5: This method always has an `Event<CloudbreakCompositeEvent>` argument.

doktoric (approved):

PR #10191, opened 2021-03-02
CB-11340 We are always wrapping exceptions in handleException, this m…

akanto:
We are always wrapping exceptions in handleException, this makes error handling harder since we need to deal with one kind of exception. It is better that if the exception is a child of CloudbreakServiceException, then we shall just reuse it and not wrap it


akanto:
Retest it please

lnardai (approved):

lnardai (approved):
LGTM!

PR #10192, opened 2021-03-02
CB-11437 - Fix an issue with Atlas POST requests exceeding buffer size

moresandeep:
This is related to https://jira.cloudera.com/browse/CB-11437
Looks like Atlas POST requests are failing because the buffer size was exceeded. This fix will increase the buffer size and fix the issue. 

This has been tested on a local cluster in mow-int.


moresandeep:
@keyki can I get a review for this, this is a blocker for the upcoming release. 
Thank you!

lacikaaa (approved):

PR #10195, opened 2021-03-02
CB-10962 allow maximum 23 volumes per instance template on all of the cloud platforms

biharitomi:

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/VolumeParameterConfig.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/VolumeParameterConfig.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.cloud.model;
 
+import com.sequenceiq.cloudbreak.validation.VolumeParameterConstants;
+
 public class VolumeParameterConfig {
 
     public static final VolumeParameterConfig EMPTY = new VolumeParameterConfig();
@@ -27,7 +29,7 @@ public class VolumeParameterConfig {
         this.minimumSize = minimumSize;
         this.maximumSize = maximumSize;
         this.minimumNumber = minimumNumber;
-        this.maximumNumber = maximumNumber;
+        this.maximumNumber = calculateSupportedMaximumNumber(maximumNumber);
     }
 
     public VolumeParameterType volumeParameterType() {
@@ -59,5 +61,13 @@ public class VolumeParameterConfig {
                 + ", minimumNumber=" + minimumNumber
                 + ", maximumNumber=" + maximumNumber
                 + '}';
+    }
+
+    private Integer calculateSupportedMaximumNumber(Integer maximumNumber) {
+        Integer result = maximumNumber;
+        if (maximumNumber != null && maximumNumber > VolumeParameterConstants.MAXIMUM_NUMBER_OF_VOLUMES) {
+            result = VolumeParameterConstants.MAXIMUM_NUMBER_OF_VOLUMES;
+        }
+        return result;
     }
 }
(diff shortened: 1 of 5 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/VolumeParameterConfig.java
```diff
@@ -60,4 +62,12 @@ public String toString() {
                 + ", maximumNumber=" + maximumNumber
                 + '}';
     }
+
+    private Integer calculateSupportedMaximumNumber(Integer maximumNumber) {
+        Integer result = maximumNumber;
+        if (maximumNumber > VolumeParameterConstants.MAXIMUM_NUMBER_OF_VOLUMES) {
+            result = VolumeParameterConstants.MAXIMUM_NUMBER_OF_VOLUMES;
+        }
+        return result;
```
lajosrodek, line 71: Perhaps simplify to `return Math.min(VolumeParameterConstants.MAXIMUM_NUMBER_OF_VOLUMES, maximumNumber)`?

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/InstanceTemplateValidator.java
```diff
@@ -1,17 +1,24 @@
...
 import com.sequenceiq.cloudbreak.validation.ValidationResult;
 import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
 import com.sequenceiq.cloudbreak.validation.Validator;
+import com.sequenceiq.cloudbreak.validation.VolumeParameterConstants;
 
 @Component
 public class InstanceTemplateValidator implements Validator<Template> {
 
+    static final String INVALID_NUMBER_OF_VOLUMES_MESSAGE_PATTERN = "Maximum '%s' volumes could be configured for an instance. "
+            + "Please modify the instance template with name: '%s', it has '%s' volumes configured!";
```
lajosrodek, line 20: The first and third `%s` should be replaced with `%d` as those are integers (`long` & `Integer`).

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/VolumeParameterConfig.java
```diff
@@ -60,4 +62,12 @@ public String toString() {
                 + ", maximumNumber=" + maximumNumber
                 + '}';
     }
+
+    private Integer calculateSupportedMaximumNumber(Integer maximumNumber) {
```
lajosrodek, line 66: Can it be `null`? If yes, the `>` comparison will result in an NPE. If not, `int` would be better.

common/src/main/java/com/sequenceiq/cloudbreak/validation/VolumeParameterConstants.java
```diff
@@ -0,0 +1,8 @@
+package com.sequenceiq.cloudbreak.validation;
+
+public final class VolumeParameterConstants {
+    public static final Integer MAXIMUM_NUMBER_OF_VOLUMES = 23;
```
lajosrodek, line 4: Why not use the primitive `int`?

lajosrodek (commented):
My unwanted 2 cents.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/VolumeParameterConfig.java
biharitomi: I can use this instead of the if condition, but I think that's a style related decision and for me the condition emphasise more that the number is overridden and easier to read then a Math expression 😄

doktoric (approved):

topolyai5 (approved):

PR #10197, opened 2021-03-03
CDPCP-3459: Only update workload credentials in IPA when they have ch…

gitmfox:
…anged in UMS

Before this change, user sync would always update a user's workload credentials,
because we had no way of knowing whether or not they had changed.

A previous thunderhead PR added versioning of workload credentials to UMS.
With this PR, the user sync service in FMS now consumes this version, and if
the version of the credentials currently stored in the IPA is not older than
the UMS version, we do not update the credentials for that user.

In order to do this, we must store the workload credentials version in the
user's LDAP record, so that user sync can compare it with the UMS version.
This is accomplished by writing a JSON document containing the version to the
user's 'title' attribute when updating the user's credentials in the IPA.
The version is encapsulated in a protobuf message, which is encoded and
represented as a string-valued property in the JSON document. Please see
the design document attached to CDPCP-3456 for details about the format of
this document, as well as the rationale for using proto and JSON.

User records in existing environments will not have this JSON string in
the 'title' attribute. If the 'title' attribute is not set, or if it is
set to a string that does not conform to the expected JSON format, user
sync assumes that the credentials are out of date, and updates them.
This update sets the 'title' attribute correctly, so that subsequent
user sync operations can perform the version check.

The JSON document also contains the user's CRN. This is not required for
the version check, but will be useful in the future for other purposes,
including addressing some edge cases related to deleting users. The CRN
is included now so that we don't have to update the JSON schema later,
and deal with the backward compatibility issues that would arise.

All of this behavior is enabled by the CDP_USER_SYNC_CREDENTIALS_UPDATE_OPTIMIZATION
entitlement, which is not currently granted to any tenant in any deployment.


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class UserMetadataConverter {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserMetadataConverter.class);
+
+    private static final ObjectReader OBJECT_READER = new ObjectMapper().readerFor(JsonUserMetadata.class)
```
handavid, line 32: look at `com.sequenceiq.cloudbreak.common.json.JsonUtil` and `Json`. You should be able to use those classes rather than your own OBJECT_READER

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FreeIpaUsersState.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class FreeIpaUsersState {
+
+    private final UsersState usersState;
+
+    private final ImmutableMap<String, UserMetadata> userMetadataMap;
```
handavid, line 14: Don't we calculate this when generating the users states from both the UMS and the FreeIPA side? If so then this can be pushed into the UsersState.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,9 +125,42 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
         return usersToAdd;
     }
 
-    public static ImmutableSet<String> calculateUsersToRemove(UmsUsersState umsState, UsersState ipaState) {
+    public static ImmutableSet<String> calculateUsersWithCredentialsToUpdate(UmsUsersState umsState, FreeIpaUsersState ipaState,
```
handavid, line 128: This whole class still assumes that, if the FreeIPA username matches, then it's the same user in the control plane. Do we have another JIRA open for using the crn for these matches instead?

handavid (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +33,82 @@
...
+            LOGGER.debug("Setting workload credentials for user '{}'", username);
+            getOperation(username, userCrn, workloadCredential, credentialsUpdateOptimizationEnabled, freeIpaClient).invoke(freeIpaClient);
+        } else {
+            LOGGER.debug("Not setting workload credentials for user '{}' because credentials are already up to date", username);
+        }
     }
 
-    public void setWorkloadCredentials(boolean fmsToFreeipaBatchCallEnabled, FreeIpaClient freeIpaClient, Map<String, WorkloadCredential> workloadCredentials,
+    public void setWorkloadCredentials(boolean fmsToFreeipaBatchCallEnabled, boolean credentialsUpdateOptimizationEnabled, FreeIpaClient freeIpaClient,
+            Map<String, WorkloadCredential> workloadCredentials, Set<String> usersWithCredentialsToUpdate, Map<String, String> userToCrnMap,
```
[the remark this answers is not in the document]
gitmfox: Yes, the keysets of `workloadCredentials` and `userToCrnMap` are guaranteed to be supersets of `usersWithCredentialsToUpdate`. The `workloadCredentials` and `userToCrnMap` keysets are also guaranteed to be equal to each other, and to the set returned by `UmsUsersState.getUsersState().getUsers()`. All three are populated at the same time in `ActorHandler.handleActor`, and all three contain keys for every user that UMS knows about. This invariant is checked by the *UmsUsersStateProviderTest unit tests.

That said, it makes sense to add a 'checkState' (or the moral equivalent) check when retrieving a user's credentials from `workloadCredentials`. I'll add that.

Yeah, the parameter list getting messy. I'll implement one of the approaches you suggest to clean it up.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -386,38 +394,42 @@ private boolean usersStateDifferenceChanged(UsersStateDifference beforeSync, Use
...
         applyStateDifferenceToIpa(environmentCrn, freeIpaClient, usersStateDifference, warnings::put, fmsToFreeipaBatchCallEnabled);
         LOGGER.debug("Finished {}.", LogEvent.APPLY_DIFFERENCE_TO_IPA);
 
         if (!FreeIpaCapabilities.hasSetPasswordHashSupport(freeIpaClient.getConfig())) {
             LOGGER.debug("IPA doesn't have password hash support, no credentials sync required for env:{}", environmentCrn);
         } else {
-            // Sync credentials for all users and not just diff. At present there is no way to identify that there is a change in password for a user
+            // If the credentials update optimization is enabled, sync credentials for users with updated credentials.
+            // Otherwise, sync credentials for all users.
             LOGGER.debug("Starting {} for {} users ...", LogEvent.SET_WORKLOAD_CREDENTIALS, umsUsersState.getUsersWorkloadCredentialMap().size());
```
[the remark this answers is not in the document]
gitmfox: We're logging this in `UsersStateDifference`, along with the count of users to add, users to delete, etc. Is that sufficient, or should we log them here as well?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Do you feel strongly about this? It's not a lot of code to create the reader/writer objects, and it allows this class to control the configuration of the mappers exactly. The fasterxml doc also explicitly says that ObjectReader/Writer are "fully immutable," while ObjectMapper is said to be "thread-safe" as long as it's configured before use. This made me feel better about using ObjectReader/Writer, although I admin I'm being a little paranoid there. There's also a pattern of creating purpose-built object mappers in the CB code base.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FreeIpaUsersState.java
gitmfox: UserMetadata is a tuple of (CRN, workload credentials version). On the UMS side, it seemed very natural to model the credentials version as a property of WorkloadCredential, which we don't have on the IPA side. The CRN, I think, belongs in FmsUser, although it can't go there right now because we don't always know the CRN on the IPA side. (I filed CDPCP-3994 to track refactoring the CRN into FmsUser.) So I'd propose leaving this as is for now, but revisiting it in the context of CDPCP-3994. What do you think?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Right, in this change I didn't want to introduce any functional differences not related to the password hash optimization -- beyond storing the CRN, which it makes sense to do now to avoid backward compatibility headaches later. 

CDPCP-892 is related, but not exactly what you're talking about. I think there may be a few other related JIRAs floating around. I'll take a look, and file a new one if there isn't an appropriate one already.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FreeIpaUsersState.java
gitmfox: On further thought, it will be a while before the CRN can be moved into FmsUser, and using UserMetadata to store the CRN and workload credential version on both the UMS and IPA sides makes a lot of sense. So I'll go ahead and push this down into UsersState, as you suggested.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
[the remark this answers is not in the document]
gitmfox: I see now; this message should log how many users have updated creds, rather than the total number of users. I'll make that change.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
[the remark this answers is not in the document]
gitmfox: > All three are populated at the same time in `ActorHandler.handleActor`, and all three contain keys for every user that UMS knows about. 

Correction to what I said above: all three contain keys for every user *with access to the environment* that UMS knows about.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +33,82 @@
...
-        LOGGER.debug("Setting workload credentials for user '{}'", username);
-        getOperation(username, workloadCredential, freeIpaClient).invoke(freeIpaClient);
+    @Inject
+    private UserMetadataConverter userMetadataConverter;
+
+    public void setWorkloadCredential(boolean credentialsUpdateOptimizationEnabled, FreeIpaClient freeIpaClient, String username, String userCrn,
+            WorkloadCredential workloadCredential) throws IOException, FreeIpaClientException {
+        boolean updateCredentials = true;
+        if (credentialsUpdateOptimizationEnabled) {
+            Optional<com.sequenceiq.freeipa.client.model.User> ipaUser = freeIpaClient.userFind(username);
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -432,12 +444,12 @@ private SyncStatusDetail internalSynchronizeStackForDeleteUser(Stack stack, Stri
...
-            LOGGER.debug("Finished {}, found {} users and {} groups.", LogEvent.RETRIEVE_PARTIAL_IPA_STATE, ipaUserState.getUsers().size(),
-                    ipaUserState.getGroups().size());
+            FreeIpaUsersState ipaUserState = getIpaStateForUser(freeIpaClient, deletedWorkloadUser);
+            LOGGER.debug("Finished {}, found {} users and {} groups.", LogEvent.RETRIEVE_PARTIAL_IPA_STATE, ipaUserState.getUsersState().getUsers().size(),
+                    ipaUserState.getUsersState().getGroups().size());
 
-            if (!ipaUserState.getUsers().isEmpty()) {
-                ImmutableCollection<String> groupsToRemove = ipaUserState.getGroupMembership().get(deletedWorkloadUser);
+            if (!ipaUserState.getUsersState().getUsers().isEmpty()) {
+                ImmutableCollection<String> groupsToRemove = ipaUserState.getUsersState().getGroupMembership().get(deletedWorkloadUser);
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
handavid: I'm ok with whatever you decide

gitmfox: @handavid, please let me know your thoughts on the above.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/FreeIpaUsersState.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: OK, I'll leave the reader/writer objects in the converter class, then.

gitmfox:
@handavid, I believe all of your feedback has been addressed now.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
```diff
@@ -0,0 +1,100 @@
...
+            .without(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+
+    private static final ObjectWriter OBJECT_WRITER = new ObjectMapper().writerFor(JsonUserMetadata.class);
+
+    @SuppressWarnings("checkstyle:NPathComplexity")
+    public Optional<UserMetadata> toUserMetadata(User user) {
```
handavid, line 38: this method is really hard to read with all the return statements. I just tossed some ideas together here (I don't really like my null checks...), but can you try to clarify this a bit (hopefully to the point you won't need to suppress NPathComplexity)?

```
JsonUserMetadata jsonUserMetadataOptional = getJsonUserMetadata(user);
if (jsonUserMetadata != null &&
    hasValidCrn(jsonUserMetadata) {
  UserManagementProto.WorkloadUserSyncActorMetadata protoMetadata = getProtoMeta(jsonUserMetadata);
  if (protoMetadata != null) {
    return Optional.of(new UserMetadata(jsonUserMetadata.getCrn(), protoMetadata.getWorkloadCredentialsVersion()));
}
return Optional.empty();
```

handavid (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: I find the "early return" style in a case like this more readable than a series of nested if/else blocks. I did rewrite this method to avoid the checkstyle warning, but felt that made it more difficult to follow the logic, and kept it as it is.

However, since the checkstyle warning indicates that the Cloudbreak style convention is to avoid multiple return statements, I'll rewrite it to conform to that.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -386,42 +398,60 @@ private boolean usersStateDifferenceChanged(UsersStateDifference beforeSync, Use
...
-            workloadCredentialService.setWorkloadCredentials(fmsToFreeipaBatchCallEnabled, freeIpaClient,
-                    umsUsersState.getUsersWorkloadCredentialMap(), warnings::put);
+            // If the credentials update optimization is enabled, sync credentials for users with updated credentials.
+            // Otherwise, sync credentials for all users.
+            LOGGER.debug("Starting {} for {} users ...", LogEvent.SET_WORKLOAD_CREDENTIALS, usersStateDifference.getUsersWithCredentialsToUpdate().size());
+            Map<String, Pair<String, WorkloadCredential>> userToCredentialUpdateParamsMap =
+                    usersStateDifference.getUsersWithCredentialsToUpdate().stream().collect(Collectors.toMap(
+                            Function.identity(), username -> getCredentialUpdateParams(username, umsUsersState)));
+            workloadCredentialService.setWorkloadCredentials(fmsToFreeipaBatchCallEnabled, credentialsUpdateOptimizationEnabled, freeIpaClient,
+                    userToCredentialUpdateParamsMap, warnings::put);
```
lacikaaa, line 438: these boolean stuff getting a bit out of hand, and if you need to comment makes me think something is fishy here.
So your comment would suggest that at this point we will branch the logic depending if this feature is enabled and sync all users or only the updated one. But this is not the case. All users will be updated which are in `usersStateDifference.getUsersWithCredentialsToUpdate()`
I think you should simply drop this comment.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -386,42 +398,60 @@ private boolean usersStateDifferenceChanged(UsersStateDifference beforeSync, Use
...
             LOGGER.debug("Finished {}.", LogEvent.SET_WORKLOAD_CREDENTIALS);
         }
     }
 
+    private Pair<String, WorkloadCredential> getCredentialUpdateParams(String username, UmsUsersState umsUsersState) {
+        UserMetadata userMetadata = umsUsersState.getUsersState().getUserMetadataMap().get(username);
+        if (userMetadata == null) {
+            throw new IllegalStateException(String.format("No UserMetadata mapping for user %s", username));
```
lacikaaa, line 446: What happens when this `IllegalStateException` is thrown? Where will it be handled? Won't this stop the whole usersync instead of failing the sync only for one user? Maybe a non runtime exception would be better and force handling these. also some logs would be nice here

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
```diff
@@ -0,0 +1,100 @@
...
+        checkArgument(!Strings.isNullOrEmpty(crn));
+
+        UserManagementProto.WorkloadUserSyncActorMetadata protoMetadata = UserManagementProto.WorkloadUserSyncActorMetadata.newBuilder()
+                .setWorkloadCredentialsVersion(workloadCredentialsVersion)
+                .build();
+        String protoMetadataBase64Encoded = BaseEncoding.base64().encode(protoMetadata.toByteArray());
+
+        JsonUserMetadata jsonUserMetadata = new JsonUserMetadata(crn, protoMetadataBase64Encoded);
+        try {
+            return OBJECT_WRITER.writeValueAsString(jsonUserMetadata);
```
lacikaaa, line 95: This is the point where I would say a comment would be nice to have some details about the decision why we have this complicated logic to store a long. I know there is the jira with the doc, but 2 lines should be sufficient. thanks

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
 
+    public static ImmutableSet<String> calculateUsersWithCredentialsToUpdate(UmsUsersState umsState, UsersState ipaState,
+            boolean credentialsUpdateOptimizationEnabled) {
+        ImmutableSet<String> usersWithCredentialsToUpdate;
+
+        if (!credentialsUpdateOptimizationEnabled) {
+            usersWithCredentialsToUpdate = ImmutableSet.copyOf(umsState.getUsersState().getUsers().stream()
+                    .map(FmsUser::getName)
+                    .filter(username -> !FreeIpaChecks.IPA_PROTECTED_USERS.contains(username))
+                    .collect(Collectors.toSet()));
```
lacikaaa, line 136: move this to a separate method please

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+                        if (ipaUserMetadata == null) {
+                            usersWithCredentialsToUpdateBuilder.add(username);
+                        } else {
+                            WorkloadCredential umsCredential = umsState.getUsersWorkloadCredentialMap().get(username);
+                            if (ipaUserMetadata.getWorkloadCredentialsVersion() < umsCredential.getVersion()) {
+                                usersWithCredentialsToUpdateBuilder.add(username);
+                            }
+                        }
+                    });
+            usersWithCredentialsToUpdate = usersWithCredentialsToUpdateBuilder.build();
```
lacikaaa, line 153: this should be refactored to a separate method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+                        UserMetadata ipaUserMetadata = ipaState.getUserMetadataMap().get(username);
+                        if (ipaUserMetadata == null) {
+                            usersWithCredentialsToUpdateBuilder.add(username);
+                        } else {
+                            WorkloadCredential umsCredential = umsState.getUsersWorkloadCredentialMap().get(username);
+                            if (ipaUserMetadata.getWorkloadCredentialsVersion() < umsCredential.getVersion()) {
+                                usersWithCredentialsToUpdateBuilder.add(username);
+                            }
+                        }
+                    });
```
lacikaaa, line 152: this should be also move to a separate method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
         return usersToAdd;
     }
 
+    public static ImmutableSet<String> calculateUsersWithCredentialsToUpdate(UmsUsersState umsState, UsersState ipaState,
+            boolean credentialsUpdateOptimizationEnabled) {
+        ImmutableSet<String> usersWithCredentialsToUpdate;
```
lacikaaa, line 130: after moving to methods it could be simplified like:
```
ImmutableSet<String> usersWithCredentialsToUpdate = credentialsUpdateOptimizationEnabled ? calcUsersWithUpdate(umsState, ipaState) : allUsers(umsState);

File: freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java [hunk 16]
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+                            if (ipaUserMetadata.getWorkloadCredentialsVersion() < umsCredential.getVersion()) {
+                                usersWithCredentialsToUpdateBuilder.add(username);
+                            }
+                        }
+                    });
+            usersWithCredentialsToUpdate = usersWithCredentialsToUpdateBuilder.build();
+        }
+
+        LOGGER.info("usersWithCredentialsToUpdate size = {}", usersWithCredentialsToUpdate.size());
+        LOGGER.debug("usersWithCredentialsToUpdate = {}", usersWithCredentialsToUpdate);
```
lacikaaa, line 157: I don't think we need 2 log lines here. I mean I don't know if there is any benefit if we see the number of users to be updated during an investigation, but not the actual users. what do you think?

lacikaaa (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Done.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/SetWlCredentialOperation.java
```diff
@@ -38,8 +47,17 @@ public static SetWlCredentialOperation create(String user, String hashedPassword
             attributes.add("cdpHashedPassword=" + hashedPassword);
             attributes.add("cdpUnencryptedKrbPrincipalKey=" + unencryptedKrbPrincipalKey);
             attributes.add("krbPasswordExpiration=" + expiration);
+        }
+
+        // In order for the workload credentials update optimization to work correctly, we must update the title attribute
+        // whenever we set workload credentials in the IPA, whether or not a workload password is set. This is because the
+        // credentials version might have changed; for example, due to the addition or deletion of an SSH public key.
+        // However, when the optimization is disabled, we don't modify the title attribute at all.
+        title.ifPresent(titleAttr -> attributes.add("title=" + titleAttr));
```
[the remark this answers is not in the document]
gitmfox: > why are we using the title? shouldn't we use a custom attribute? are we sure we won't need this field in freeipa in the future?

After some discussion with Kat, we believe that it's OK for us to take over an existing attribute, since we "own" the IPA/LDAP user record -- that is, we have full control over it, and it should not ever be modified by customers. This saves us the work of updating the schema, and of dealing with backward compatibility.

We aren't using 'title' for anything now, and it's hard to imagine that we would ever need to store an actual job title. The other advantage of 'title' is that it's in the default set of LDAP attributes searched by the IPA `user_find` command, which gives us the ability to search for a user by CRN easily.

There is some discussion related to this in this doc: https://docs.google.com/document/d/1jXhCruiM18cvfONoiQUrf37TW_pBhJpc-LweH_a4R1U/edit?usp=sharing

If you still have concerns, we should set up a meeting to discuss this with Kat and the CB and ID mgmt teams.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/operation/SetWlCredentialOperation.java
```diff
@@ -38,8 +47,17 @@ public static SetWlCredentialOperation create(String user, String hashedPassword
...
+        if (!attributes.isEmpty()) {
```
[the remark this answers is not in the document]
gitmfox: > do we really need this if?

I believe so. Currently, if the user does not have a hashed password set, we do not emit a `setattr` command, and we don't want to change the current behavior if the CDP_USER_SYNC_CREDENTIALS_UPDATE_OPTIMIZATION entitlement is not enabled.

Also, I don't know what would happen if we emitted a `setattr` without any attributes to set as parameters -- the IPA might not like that.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: I was following the pattern in this class, which emits two similar log statements (INFO with count, DEBUG with resource names) for all the various change sets it computes (users to add, users to delete, etc.). If we're always running with DEBUG logs enabled, then I guess there's not much point in logging the count. But otherwise, I think it makes sense to avoid bloating the production logs, and I do find it's useful to see the count, at least for debugging.

But I'm fine either way.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +32,67 @@
...
-        } else {
-            for (SetWlCredentialOperation operation : operations) {
-                try {
-                    operation.invoke(freeIpaClient);
-                } catch (FreeIpaClientException e) {
-                    recordWarning(operation.getUser(), e, warnings);
-                    freeIpaClient.checkIfClientStillUsable(e);
+        });
+
+        if (!operations.isEmpty()) {
```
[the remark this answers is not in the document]
gitmfox: Without this check, we execute an empty batch call against the IPA if batch calls are enabled.

I'll move the check to inside the batch call if block.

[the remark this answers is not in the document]
gitmfox: Actually, the check can go around the whole method, saving a little work if there are no credentials to update.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/model/Entitlement.java
```diff
@@ -40,5 +40,6 @@
     CDP_USE_CM_SYNC_COMMAND_POLLER,
     CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY,
     FMS_FREEIPA_BATCH_CALL,
-    CDP_CB_AZURE_DISK_SSE_WITH_CMK;
+    CDP_CB_AZURE_DISK_SSE_WITH_CMK,
+    CDP_USER_SYNC_CREDENTIALS_UPDATE_OPTIMIZATION;
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -56,10 +69,31 @@ public void accept(Event<SetPasswordRequest> setPasswordRequestEvent) {
...
+                        }
+                    }
+                }
+
+                if (updateCredentials) {
+                    workloadCredentialService.setWorkloadCredential(credentialsUpdateOptimizationEnabled, freeIpaClient, request.getUsername(),
+                            request.getUserCrn(), workloadCredential);
+                } else {
+                    LOGGER.debug("Not setting workload credentials for user '{}' because credentials are already up to date", request.getUsername());
+                }
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +32,67 @@
...
+        if (credentialsUpdateOptimizationEnabled) {
+            if (StringUtils.isBlank(crn)) {
+                // The user's CRN is gotten from the UMS, not the IPA, and so should always be present.
+                throw new IllegalStateException(String.format("Missing CRN for user %s ", user));
+            }
+            String userMetadataJson = userMetadataConverter.toUserMetadataJson(crn, workloadCredential.getVersion());
+            return SetWlCredentialOperation.create(user, workloadCredential.getHashedPassword(), asnEncodedKrbPrincipalKey,
+                    sshPublicKeys, expiration, userMetadataJson);
+        } else {
+            // Do not set the title attribute if the credentials update optimization is not enabled.
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +32,67 @@
...
+            WorkloadCredential workloadCredential) throws IOException, FreeIpaClientException {
         LOGGER.debug("Setting workload credentials for user '{}'", username);
-        getOperation(username, workloadCredential, freeIpaClient).invoke(freeIpaClient);
+        getOperation(username, userCrn, workloadCredential, credentialsUpdateOptimizationEnabled, freeIpaClient).invoke(freeIpaClient);
     }
 
-    public void setWorkloadCredentials(boolean fmsToFreeipaBatchCallEnabled, FreeIpaClient freeIpaClient, Map<String, WorkloadCredential> workloadCredentials,
-            BiConsumer<String, String> warnings) throws FreeIpaClientException {
+    public void setWorkloadCredentials(boolean fmsToFreeipaBatchCallEnabled, boolean credentialsUpdateOptimizationEnabled, FreeIpaClient freeIpaClient,
+            Map<String, Pair<String, WorkloadCredential>> userToCredentialUpdateParamsMap, BiConsumer<String, String> warnings)
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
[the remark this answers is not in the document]
gitmfox: Moved the check to go around the whole method.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -361,17 +371,19 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
         }
     }
 
+    @SuppressWarnings("checkstyle:ParameterNumber")
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
gitmfox: Comment dropped.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
gitmfox: This is intended as a precondition check, like requireNonNull/checkArgument/etc. If it fails, it means there is a bug. The reason for throwing IllegalStateException, rather than calling Precondition.checkState, is that I saw that being done elsewhere in the Cloudbreak code, and so thought that was the pattern to follow. 

I changed this to use standard requireNonNull checks, which will hopefully clarify the intent.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Done.

Also, with regard to this comment on ll. 35-38 of UserMetadataConverter:

> please remove @SuppressWarnings and refactor this method. I agree with @handavid , it's hard to follow what happens here

This is done too. (Github wouldn't let me reply directly to the comment.)

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -113,6 +125,39 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+            usersWithCredentialsToUpdate = ImmutableSet.copyOf(umsState.getUsersState().getUsers().stream()
+                    .map(FmsUser::getName)
+                    .filter(username -> !FreeIpaChecks.IPA_PROTECTED_USERS.contains(username))
+                    .collect(Collectors.toSet()));
+        } else {
+            ImmutableSet.Builder<String> usersWithCredentialsToUpdateBuilder = ImmutableSet.builder();
+            umsState.getUsersState().getUsers().stream()
+                    .map(FmsUser::getName)
+                    .filter(username -> !FreeIpaChecks.IPA_PROTECTED_USERS.contains(username))
+                    .forEach(username -> {
```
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -340,15 +345,19 @@ private SyncStatusDetail internalSynchronizeStack(Stack stack, UmsUsersState ums
         Multimap<String, String> warnings = ArrayListMultimap.create();
         try {
             FreeIpaClient freeIpaClient = freeIpaClientFactory.getFreeIpaClientForStack(stack);
-            boolean fmsToFreeipaBatchCallEnabled = entitlementService.isFmsToFreeipaBatchCallEnabled(Crn.fromString(environmentCrn).getAccountId());
-            UsersStateDifference usersStateDifferenceBeforeSync = compareUmsAndFreeIpa(umsUsersState, fullSync, freeIpaClient);
-            applyDifference(umsUsersState, environmentCrn, warnings, usersStateDifferenceBeforeSync, freeIpaClient, fmsToFreeipaBatchCallEnabled);
+            String accountId = Crn.fromString(environmentCrn).getAccountId();
```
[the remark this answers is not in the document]
gitmfox: I'll address in the next PR (adds the force-workload-credentials-update flag), which refactors the code that creates the UserSyncOptions object to a higher spot in the call chain, where the accountId is available as a parameter.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,50 +34,95 @@
...
+                    sshPublicKeys, expiration);
+        }
     }
 
     private void recordWarning(String username, Exception e, BiConsumer<String, String> warnings) {
         LOGGER.warn("Failed to set workload credentials for user '{}'", username, e);
         warnings.accept(username, "Failed to set workload credentials:" + e.getMessage());
     }
+
+    public static class WorkloadCredentialUpdate {
```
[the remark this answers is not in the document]
handavid: for simplicity and it's the style. The only public inner classes we have in this module so far are Builders.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
```diff
@@ -76,4 +92,34 @@ public void accept(Event<SetPasswordRequest> setPasswordRequestEvent) {
...
+            workloadCredentialService.setWorkloadCredential(credentialsUpdateOptimizationEnabled, freeIpaClient,
+                    new WorkloadCredentialUpdate(request.getUsername(), request.getUserCrn(), workloadCredential));
+        } else {
+            LOGGER.debug("Not setting workload credentials for user '{}' because credentials are already up to date", request.getUsername());
+        }
+    }
+
+    private boolean credentialUpdateRequired(boolean credentialsUpdateOptimizationEnabled, String username, FreeIpaClient freeIpaClient,
+            WorkloadCredential workloadCredential) throws FreeIpaClientException {
+        boolean updateCredentials = true;
```
lacikaaa, line 112: could you drop the variable and just use `return true` or `return false`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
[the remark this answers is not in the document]
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
```diff
@@ -0,0 +1,124 @@
...
+        boolean validCrn = false;
+        if (crn.isPresent()) {
+            Crn crnVal = crn.get();
+            if (crnVal.getService() == Crn.Service.IAM &&
+                    (crnVal.getResourceType() == Crn.ResourceType.USER || crnVal.getResourceType() == Crn.ResourceType.MACHINE_USER)) {
+                validCrn = true;
+            } else {
+                LOGGER.warn("CRN {} in metadata for user with workload username {} is not a user or machine user CRN", crnToValidate, username);
+            }
+        }
```
lacikaaa, line 109: I think this way it's a bit more readable and shorter:
```
    private boolean isValidActorCrn(String crnToValidate, String username) {
        try {
            Crn crn = (Crn.safeFromString(crnToValidate));
            if (crn.getService() == Crn.Service.IAM &&
                    (crn.getResourceType() == Crn.ResourceType.USER || crn.getResourceType() == Crn.ResourceType.MACHINE_USER)) {
                return true;
            } else {
                LOGGER.warn("CRN {} in metadata for user with workload username {} is not a user or machine user CRN", crnToValidate, username);
                return false;
            }
        } catch (CrnParseException e) {
            LOGGER.warn("Unable to parse CRN {} in metadata for user with workload username {}", crnToValidate, username, e);
            return false;
        }
    }

File: freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java [hunk 30]
```diff
@@ -0,0 +1,124 @@
...
+    private Optional<UserManagementProto.WorkloadUserSyncActorMetadata> decodeProtoMetadata(String encodedMetadata, String username) {
+        Optional<UserManagementProto.WorkloadUserSyncActorMetadata> protoMetadata = Optional.empty();
+        try {
+            byte[] base64Decoded = BaseEncoding.base64().decode(encodedMetadata);
+            protoMetadata = Optional.of(UserManagementProto.WorkloadUserSyncActorMetadata.parseFrom(base64Decoded));
+        } catch (Exception e) {
+            LOGGER.warn("Unable to decode meta attribute {} for user work workload username {}", encodedMetadata, username, e);
+        }
+        return protoMetadata;
+    }
```
lacikaaa, line 123: similarly to the previous:
```
    private Optional<UserManagementProto.WorkloadUserSyncActorMetadata> decodeProtoMetadata(String encodedMetadata, String username) {
        try {
            byte[] base64Decoded = BaseEncoding.base64().decode(encodedMetadata);
            return Optional.of(UserManagementProto.WorkloadUserSyncActorMetadata.parseFrom(base64Decoded));
        } catch (Exception e) {
            LOGGER.warn("Unable to decode meta attribute {} for user work workload username {}", encodedMetadata, username, e);
            return Optional.empty();
        }
    }

File: freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java [hunk 31]
```diff
@@ -0,0 +1,124 @@
...
+        JsonUserMetadata jsonUserMetadata = new JsonUserMetadata(crn, protoMetadataBase64Encoded);
+        try {
+            return OBJECT_WRITER.writeValueAsString(jsonUserMetadata);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException(String.format("Unable to serialize %s as JSON: ", jsonUserMetadata), e);
+        }
+    }
+
+    private Optional<JsonUserMetadata> getJsonUserMetadata(User user) {
+        Optional<JsonUserMetadata> jsonUserMetadata = Optional.empty();
```
lacikaaa, line 83: please drop the variable, no need for it

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UserSyncOptions.java
```diff
@@ -0,0 +1,28 @@
...
+        return fullSync;
+    }
+
+    public boolean fmsToFreeIpaBatchCallEnabled() {
+        return fmsToFreeIpaBatchCallEnabled;
+    }
+
+    public boolean credentialsUpdateOptimizationEnabled() {
+        return credentialsUpdateOptimizationEnabled;
+    }
```
lacikaaa, line 27: please start the method name with a verb if possible. in case of booleans it's usually `is` eg:
`isFullSync`

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -183,4 +204,28 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+                .filter(username -> !FreeIpaChecks.IPA_PROTECTED_USERS.contains(username) && credentialsAreStale(username, umsState, ipaState))
+                .collect(ImmutableSet.toImmutableSet());
+    }
+
+    private static boolean credentialsAreStale(String username, UmsUsersState umsState, UsersState ipaState) {
+        boolean updateCredentials = true;
+        UserMetadata ipaUserMetadata = ipaState.getUserMetadataMap().get(username);
+        if (ipaUserMetadata != null) {
+            WorkloadCredential umsCredential = umsState.getUsersWorkloadCredentialMap().get(username);
+            updateCredentials = ipaUserMetadata.getWorkloadCredentialsVersion() < umsCredential.getVersion();
```
lacikaaa, line 227: just use return here, no need for the variable. if you put the next line into an else, it would be obvious by looking at the method it has 2 exit point and no need to check where the variable could be modified during execution

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredentialUpdate.java
```diff
@@ -0,0 +1,35 @@
...
+public class WorkloadCredentialUpdate {
+
+    private final String username;
+
+    private final String userCrn;
+
+    private final WorkloadCredential workloadCredential;
+
+    public WorkloadCredentialUpdate(String username, String userCrn, WorkloadCredential workloadCredential) {
+        checkArgument(!StringUtils.isBlank(username), "username must not be blank");
```
lacikaaa, line 17: you could use `isNotBlank` method

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredentialUpdate.java
```diff
@@ -0,0 +1,35 @@
...
+        this.username = username;
+        checkArgument(!StringUtils.isBlank(userCrn), "user CRN must not be blank");
```
lacikaaa, line 19: in my opinion it would be better to have all the validations first then the assignments

mock-thunderhead/src/main/resources/application.yml
```diff
@@ -36,6 +36,7 @@ auth:
     environment.experience.deletion.enable: true
     endpointgateway.enable: false
     cm.sync.command.poller.enable: true
+    user.sync.credentials.update.optimization.enable: false
```
lacikaaa, line 39: we usually enable new feature in the mock by default so developers would test it. Is there any reasons against turning it on?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
```diff
@@ -29,46 +31,62 @@
...
             throws IOException, FreeIpaClientException {
-        LOGGER.debug("Setting workload credentials for user '{}'", username);
-        getOperation(username, workloadCredential, freeIpaClient).invoke(freeIpaClient);
+        LOGGER.debug("Setting workload credentials for user '{}'", update.getUsername());
+        getOperation(update.getUsername(), update.getUserCrn(), update.getWorkloadCredential(), credentialsUpdateOptimizationEnabled, freeIpaClient)
+                .invoke(freeIpaClient);
     }
 
-    public void setWorkloadCredentials(boolean fmsToFreeipaBatchCallEnabled, FreeIpaClient freeIpaClient, Map<String, WorkloadCredential> workloadCredentials,
+    public void setWorkloadCredentials(UserSyncOptions options, FreeIpaClient freeIpaClient, Set<WorkloadCredentialUpdate> updates,
```
lacikaaa, line 44: could you check this refactor. it should do the same but I think it's more readable:
```
    public void setWorkloadCredentials(UserSyncOptions options, FreeIpaClient freeIpaClient, Set<WorkloadCredentialUpdate> updates,
            BiConsumer<String, String> warnings) throws FreeIpaClientException {
        if (!updates.isEmpty()) {
            List<SetWlCredentialOperation> operations = updates.stream()
                    .flatMap(update -> createUpdateOperation(options, freeIpaClient, warnings, update))
            if (options.fmsToFreeIpaBatchCallEnabled()) {
                updateInBatch(freeIpaClient, warnings, operations);
            } else {
                updateOneByOne(freeIpaClient, warnings, operations);
            }
        } else {
            LOGGER.debug("Nothing to update");
        }
    }

    private void updateOneByOne(FreeIpaClient freeIpaClient, BiConsumer<String, String> warnings, List<SetWlCredentialOperation> operations) throws FreeIpaClientException {
        LOGGER.debug("updateOneByOne log");
        for (SetWlCredentialOperation operation : operations) {
            try {
                operation.invoke(freeIpaClient);
            } catch (FreeIpaClientException e) {
                recordWarning(operation.getUser(), e, warnings);
                freeIpaClient.checkIfClientStillUsable(e);
            }
        }
    }

    private void updateInBatch(FreeIpaClient freeIpaClient, BiConsumer<String, String> warnings, List<SetWlCredentialOperation> operations) throws FreeIpaClientException {
        LOGGER.debug("Some log");
        List<Object> batchCallOperations = operations.stream()
                .map(AbstractFreeipaOperation::getOperationParamsForBatchCall)
        String operationName = operations.stream()
                .map(UserModOperation::getOperationName)
                .findFirst().orElse("unknown");
        freeIpaClient.callBatch(warnings, batchCallOperations, batchPartitionSizeProperties.getByOperation(operationName), Set.of());
    }

    private Stream<SetWlCredentialOperation> createUpdateOperation(UserSyncOptions options, FreeIpaClient freeIpaClient, BiConsumer<String, String> warnings, WorkloadCredentialUpdate update) {
        try {
            return Stream.of(getOperation(update.getUsername(), update.getUserCrn(), update.getWorkloadCredential(),
                    options.credentialsUpdateOptimizationEnabled(), freeIpaClient));
        } catch (IOException e) {
            recordWarning(update.getUsername(), e, warnings);
            LOGGER.warn("I would log this here");
            return Stream.empty();
        }
    }

lacikaaa (changes requested):
thanks for the changes. I added a few comments

gitmfox:
@lacikaaa, do you mind if I address your latest round of comments in a separate PR?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/user/handler/SetPasswordHandler.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/conversion/UserMetadataConverter.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UserSyncOptions.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredentialUpdate.java
gitmfox: Done.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/WorkloadCredentialUpdate.java
gitmfox: Done.

mock-thunderhead/src/main/resources/application.yml
gitmfox: Enabled.

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/WorkloadCredentialService.java
gitmfox: Done.

gitmfox:
> @lacikaaa, do you mind if I address your latest round of comments in a separate PR?

@lacikaaa, please disregard the above. I have made the requested changes.

gitmfox:
retest this please

lacikaaa:
retest this please

lacikaaa (approved):

PR #10205, opened 2021-03-08
DISTX-564 DE HA Template has been updated for production workloads.

smaniraju:
1. Manager, Master, Compute, Gateway : Root Volume increased to 150gb, attached volume increased to 150gb ssd
2. Worker : Root Volume increased to 150gb, attached volume increased to 500gb ssd (1.5TB for 3 workers)
3. Manager, Master, Gateway, Worker, Compute upgraded to 64 GB Machines
4. AWS Manager, Worker, Master HostGroups configured with AWS Placement Settings.
5. Azure  Availability Set definition for Compute HostGroup removed since Compute is expected to scale beyond 200.
6. Above changes applied in 7.2.8, 7.2.9 and 7.2.10 templates.

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering-ha.json
@@ -17,12 +17,19 @@
       "type": "GATEWAY",
       "recoveryMode": "MANUAL",
       "template": {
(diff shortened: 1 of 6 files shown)


cegganesh84 (approved):
This looks good to me. Since we are changing the instance type we need to send it out in the release notes. The field uses a cost calculator for default cluster shapes. They will be able to update that based on the release notes.

PR #10135, opened 2021-02-24
CDPD-19363:set tez.runtime.pipelined.sorter.lazy-allocate.memory to true by default in DH

satyabollineni:
tez.runtime.pipelined.sorter.lazy-allocate.memory is set to true by default and the screenshot from tez-site.xml is :
Screenshot 2021-02-23 at 10 13 31 PM
 
and the CM UI having the value in tez gateway configs:

image

PR #10209, opened 2021-03-09
CDPD-19363:set tez.runtime.pipelined.sorter.lazy-allocate.memory to t…

satyabollineni:
tez.runtime.pipelined.sorter.lazy-allocate.memory is set to true by default in 7.2.10 templates, The change was already made for 7.2.8 and 7.2.9 templates are part of this PR: https://github.com/hortonworks/cloudbreak/pull/10135


cegganesh84 (approved):
LGTM

doktoric:
is this only applicable for 7.2.10

satyabollineni:
> is this only applicable for 7.2.10? if customer using a custom template do that needs this property as well ?

This Changes for 7.2.8, 7.2.9 are already made earlier https://github.com/hortonworks/cloudbreak/pull/10135 since 7.2.10 was not on master at that time.

PR #10210, opened 2021-03-09
CB-11428: Download CSD parcels only in case of workload clusters.

tiborpopovics:
Download CSD parcels only in case of workload clusters because DataLake does not require these parcels.


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -501,9 +502,11 @@ public void decoratePillarWithClouderaManagerRepo(ClouderaManagerRepo repo, Map<
         return pillarValues;
     }
 
-    public void decoratePillarWithClouderaManagerCsds(Cluster cluster, Map<String, SaltPillarProperties> servicePillar) {
-        List<ClouderaManagerProduct> product = clusterComponentConfigProvider.getClouderaManagerProductDetails(cluster.getId());
-        addClouderaManagerCsdsToServicePillar(product, servicePillar);
+    private void decoratePillarWithClouderaManagerCsds(Cluster cluster, Map<String, SaltPillarProperties> servicePillar, StackType type) {
+        if (StackType.WORKLOAD.equals(type)) {
```
pdarvasi, line 506: could you pls add 2 log lines indicating the 2 conditions of the "if" clause?

lacikaaa (approved):

PR #10213, opened 2021-03-09
CB-11458 When starting the datahub delete flow, the user is not popul…

topolyai5:
…ated into the security context, trigggering the fallback logic to stack creator. In the flow we are calling authenticated APIs (e.g. get env by crn) that fail to authenticate the deleted user, causing the flow to fail. In this PR I have removed the legacy authenticatedUserService and changed to ThreadBasedUserCrnProvider

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/service/CbEventParameterFactory.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/service/CbEventParameterFactory.java
@@ -2,38 +2,25 @@ package com.sequenceiq.cloudbreak.core.flow2.service;
 
 import java.util.Map;
 
-import javax.inject.Inject;
-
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/service/CbEventParameterFactory.java
```diff
@@ -2,38 +2,25 @@
...
 public class CbEventParameterFactory implements EventParameterFactory {
 
-    @Inject
-    private StackService stackService;
-
-    @Inject
-    private AuthenticatedUserService authenticatedUserService;
-
+    @Override
     public Map<String, Object> createEventParameters(Long stackId) {
```
biharitomi, line 15: We can remove the stackid parameter as we don't using it anymore

biharitomi (approved):

PR #10211, opened 2021-03-09
DISTX-575:Use gp2 storages for AWS

satyabollineni:
The default disk types that are launched with cluster templates are changed to gp2. Screenshot below for the same:
Screenshot 2021-03-09 at 3 53 55 PM

The change is made for 7.2.8,7.2.9.7.2.10 for DE and DE-spark3 templates. I see the changes were already being made for DE-HA templates as part of this PR: https://github.com/hortonworks/cloudbreak/pull/10205/files

PR #10217, opened 2021-03-10
DISTX-575:Use gp2 storages for AWS

satyabollineni:
The default disk types that are launched with cluster templates are changed to gp2. Screenshot below for the same:

image
The change is made for 7.2.8,7.2.9.7.2.10 for DE and DE-spark3 templates. I see the changes were already being made for DE-HA templates as part of this PR: https://github.com/hortonworks/cloudbreak/pull/10205/files


doktoric:
please add a more detailed commit message which describes the problem and the solution

doktoric:
why only these templates changed?

doktoric (approved):

lajosrodek:
Follow-up to #10211.

PR #10219, opened 2021-03-10
CB-10577: Removed medium duty and loadbalancer entitlement checks for AWS.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-10577

This is meant to remove the check for the CDP_MEDIUM_DUTY_SDX and CDP_DATA_LAKE_LOAD_BALANCER entitlements during the datalake provisioning process if the cloud provider is AWS, it still checks for these entitlements otherwise.

Specifically, this will result in all AWS datalake provisions being set up with an accompanying loadbalancer, and globally enable medium duty setups for everyone.

I went through and added some unit tests to make sure this is working as expected.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
@@ -539,9 +539,10 @@ public class SdxService implements ResourceIdProvider, ResourceCrnAndNameProvide
         ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
         if (SdxClusterShape.MEDIUM_DUTY_HA.equals(sdxClusterRequest.getClusterShape())) {
             boolean mediumDutySdxEntitlementEnabled = entitlementService.mediumDutySdxEnabled(Crn.safeFromString(environment.getCreator()).getAccountId());
-            if (!mediumDutySdxEntitlementEnabled) {
-                validationBuilder.error("Provisioning a medium duty data lake cluster is not enabled for this account. " +
-                        "Contact Cloudera support to enable CDP_MEDIUM_DUTY_SDX entitlement for the account.");
+            boolean entitlementRequiredForCloudProvider = isMediumDutyEntitlementRequiredForCloudProvider(environment.getCloudPlatform());
+            if (!mediumDutySdxEntitlementEnabled && entitlementRequiredForCloudProvider) {
+                validationBuilder.error(String.format("Provisioning a medium duty data lake cluster is not enabled for %s. " +
+                        "Contact Cloudera support to enable CDP_MEDIUM_DUTY_SDX entitlement for the account.", environment.getCloudPlatform()));
             }
             if (!isMediumDutySdxSupported(sdxClusterRequest.getRuntime())) {
                 validationBuilder.error("Provisioning a Medium Duty SDX shape is only valid for CM version >= " + MEDIUM_DUTY_REQUIRED_VERSION +
@@ -552,6 +553,10 @@ public class SdxService implements ResourceIdProvider, ResourceCrnAndNameProvide
         if (validationResult.hasError()) {
             throw new BadRequestException(validationResult.getFormattedErrors());
         }
+    }
+
+    private boolean isMediumDutyEntitlementRequiredForCloudProvider(String cloudPlatform) {
+        return !(AWS.equalsIgnoreCase(cloudPlatform));
     }
 
     /**
--- a/datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxServiceTest.java
+++ b/datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxServiceTest.java
@@ -725,16 +725,57 @@ class SdxServiceTest {
     }
 
     @Test
-    void testSdxCreateMediumDutySdxNoEntitlement() throws IOException {
+    void testSdxCreateAzureMediumDutySdxNoEntitlement() throws IOException {
+        final String runtime = "7.2.7";
+        SdxClusterRequest sdxClusterRequest = createSdxClusterRequest(runtime, MEDIUM_DUTY_HA);
+        when(sdxClusterRepository.findByAccountIdAndEnvNameAndDeletedIsNull(anyString(), anyString())).thenReturn(new ArrayList<>());
+        mockEnvironmentCall(sdxClusterRequest, CloudPlatform.AZURE);
+        when(entitlementService.mediumDutySdxEnabled(anyString())).thenReturn(false);
+        BadRequestException badRequestException = assertThrows(BadRequestException.class,
+                () -> underTest.createSdx(USER_CRN, CLUSTER_NAME, sdxClusterRequest, null));
+        assertEquals(String.format("1. Provisioning a medium duty data lake cluster is not enabled for %s. ", CloudPlatform.AZURE.name()) +
+                "Contact Cloudera support to enable CDP_MEDIUM_DUTY_SDX entitlement for the account.", badRequestException.getMessage());
+    }
+
+    @Test
+    void testSdxCreateGcpMediumDutySdxNoEntitlement() throws IOException {
+        final String runtime = "7.2.7";
+        SdxClusterRequest sdxClusterRequest = createSdxClusterRequest(runtime, MEDIUM_DUTY_HA);
+        when(sdxClusterRepository.findByAccountIdAndEnvNameAndDeletedIsNull(anyString(), anyString())).thenReturn(new ArrayList<>());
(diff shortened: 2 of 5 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -539,7 +539,7 @@ private void validateMediumDutySdxEnablement(SdxClusterRequest sdxClusterRequest
         ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
         if (SdxClusterShape.MEDIUM_DUTY_HA.equals(sdxClusterRequest.getClusterShape())) {
             boolean mediumDutySdxEntitlementEnabled = entitlementService.mediumDutySdxEnabled(Crn.safeFromString(environment.getCreator()).getAccountId());
-            if (!mediumDutySdxEntitlementEnabled) {
+            if (!(AWS.equalsIgnoreCase(environment.getCloudPlatform()) || mediumDutySdxEntitlementEnabled)) {
                 validationBuilder.error("Provisioning a medium duty data lake cluster is not enabled for this account. " +
```
kmanamcheri, line 543: Update this to state that "medium duty data lake cluster on **Azure** is not enabled"

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -539,7 +539,7 @@ private void validateMediumDutySdxEnablement(SdxClusterRequest sdxClusterRequest
         ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
         if (SdxClusterShape.MEDIUM_DUTY_HA.equals(sdxClusterRequest.getClusterShape())) {
             boolean mediumDutySdxEntitlementEnabled = entitlementService.mediumDutySdxEnabled(Crn.safeFromString(environment.getCreator()).getAccountId());
-            if (!mediumDutySdxEntitlementEnabled) {
+            if (!(AWS.equalsIgnoreCase(environment.getCloudPlatform()) || mediumDutySdxEntitlementEnabled)) {
```
kmanamcheri, line 542: I prefer not negating under brackets. For example, instead of saying "not (A or B)", you can say "not A and not B". They are logically equivalent. The reason is that the next time someone goes to modify this line, they might miss the brackets and might accidentally break the logic. We want to make it easy for the next person to not make accidental mistakes. 

To make this logic cleaner, do something like this

```
  boolean isAWS = AWS.equalsIgnoreCase(environment.getCloudPlatform());
  if (!isAWS && mediumDutySdxEntitlementEnabled) {
    // print error
  } 
```

The logic reads much more cleaner now: "If its not AWS and you have medium duty enabled, print the error"

datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxServiceTest.java
```diff
@@ -729,14 +729,42 @@ void testSdxCreateMediumDutySdxNoEntitlement() throws IOException {
         final String runtime = "7.2.7";
         SdxClusterRequest sdxClusterRequest = createSdxClusterRequest(runtime, MEDIUM_DUTY_HA);
         when(sdxClusterRepository.findByAccountIdAndEnvNameAndDeletedIsNull(anyString(), anyString())).thenReturn(new ArrayList<>());
-        mockEnvironmentCall(sdxClusterRequest, CloudPlatform.AWS);
+        mockEnvironmentCall(sdxClusterRequest, CloudPlatform.AZURE);
```
kmanamcheri, line 732: Should we also test with `CloudPlatform.GCP` to ensure that the checks are done on GCP. We should make sure that the check-removal is only for AWS and for all other platforms, we will check the entitlement.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
sxxgrc: Wouldn't we want to make this general for any of the cloud platforms? Would it be better to say "...medium duty data lake cluster is not enabled for this cloud platform"?

datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxServiceTest.java
sxxgrc: Ah yeah we can test with it as well.

sxxgrc:
Made the requested changes.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
hreeve-cloudera: Why don't you spit the difference, and do something like:
```
validationBuilder.error(String.format("Provisioning a medium duty data lake cluster is not enabled for %s. " +
        "Contact Cloudera support to enable CDP_MEDIUM_DUTY_SDX entitlement for the account.", environment.getCloudPlatform()));
```

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentLoadBalancerService.java
```diff
@@ -63,4 +63,14 @@ public void updateLoadBalancerInEnvironmentAndStacks(EnvironmentDto environmentD
                     environmentLbDto.getEndpointAccessGateway(), environmentLbDto.getEndpointGatewaySubnetIds(), userCrn);
         }
     }
+
+    private boolean isLoadBalancerEnabledForDatalake(String accountId, String cloudPlatform) {
+        return entitlementService.datalakeLoadBalancerEnabled(accountId) ||
+                isLoadBalancerEntitlementNotRequiredForCloudProvider(cloudPlatform) ||
+                entitlementService.publicEndpointAccessGatewayEnabled(accountId);
```
hreeve-cloudera, line 70: nit: It would be more efficient to have the `isLoadBalancerEntitlementNotRequiredForCloudProvider(cloudPlatform)` line be the first check, because it's a simple string compare and the cheapest operation.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -539,7 +539,8 @@ private void validateMediumDutySdxEnablement(SdxClusterRequest sdxClusterRequest
         ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
         if (SdxClusterShape.MEDIUM_DUTY_HA.equals(sdxClusterRequest.getClusterShape())) {
             boolean mediumDutySdxEntitlementEnabled = entitlementService.mediumDutySdxEnabled(Crn.safeFromString(environment.getCreator()).getAccountId());
-            if (!mediumDutySdxEntitlementEnabled) {
+            boolean entitlementNotRequiredForCloudProvider = isMediumDutyEntitlementNotRequiredForCloudProvider(environment.getCloudPlatform());
+            if (!mediumDutySdxEntitlementEnabled && !entitlementNotRequiredForCloudProvider) {
```
kmanamcheri, line 543: Double negative is a little confusing. Change it to `entitlementRequiredForCloudProvider` and the associated logic above

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
sxxgrc: Sounds good I'll do that.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -539,7 +539,8 @@ private void validateMediumDutySdxEnablement(SdxClusterRequest sdxClusterRequest
         ValidationResultBuilder validationBuilder = new ValidationResultBuilder();
         if (SdxClusterShape.MEDIUM_DUTY_HA.equals(sdxClusterRequest.getClusterShape())) {
             boolean mediumDutySdxEntitlementEnabled = entitlementService.mediumDutySdxEnabled(Crn.safeFromString(environment.getCreator()).getAccountId());
-            if (!mediumDutySdxEntitlementEnabled) {
+            boolean entitlementNotRequiredForCloudProvider = isMediumDutyEntitlementNotRequiredForCloudProvider(environment.getCloudPlatform());
```
kmanamcheri, line 542: Same here. Change the function to "isMediumDutyEntitlementRequiredForCloudProvider".

Prefer using positive verbs to negative verbs.

kmanamcheri (commented):
Rerun you checkstyles before submitting.

sxxgrc:
Made the requested changes and fixed the checkstyle error.

kmanamcheri (approved):

PR #10222, opened 2021-03-10
DISTX-585 Autoscaling trigger failure should be made available in history events.

smaniraju:
Autoscaling trigger failure should be made available in history events.

### DIFF
--- a/autoscale-api/src/main/resources/messages/messages.properties
+++ b/autoscale-api/src/main/resources/messages/messages.properties
@@ -24,3 +24,5 @@ autoscale.validation.time.negative.adjustment=Schedule-Based Autoscaling Node Co
 
 autoscale.notification.history.update=Autoscaling History Updated
 autoscale.notification.config.update=Autoscaling Config Updated
+
+autoscale.trigger.failure=Autoscaling Trigger Failed. Please verify Cluster Connectivity and Environment User(s) Synchronization.
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
@@ -12,10 +12,14 @@ import org.springframework.context.ApplicationListener;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.logger.MDCBuilder;
+import com.sequenceiq.cloudbreak.message.CloudbreakMessagesService;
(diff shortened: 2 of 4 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
```diff
@@ -27,16 +31,20 @@
...
     private final Map<Long, Integer> updateFailures = new ConcurrentHashMap<>();
 
     @Override
     public void onApplicationEvent(UpdateFailedEvent event) {
         long autoscaleClusterId = event.getClusterId();
         Cluster cluster = clusterService.findById(autoscaleClusterId);
-        if (cluster == null) {
-            return;
-        }
         MDCBuilder.buildMdcContext(cluster);
```
smaniraju, line 46: This is a no-op block since findById throws NotFoundException if clusterId is not found.

autoscale-api/src/main/resources/messages/messages.properties
```diff
@@ -24,3 +24,5 @@ autoscale.validation.time.negative.adjustment=Schedule-Based Autoscaling Node Co
 
 autoscale.notification.history.update=Autoscaling History Updated
 autoscale.notification.config.update=Autoscaling Config Updated
+
+autoscale.trigger.failure=Autoscaling Trigger Failed. please verify Cluster Connectivity and Environment User(s) Synchronization.
```
sidseth, line 28: Super nit. s/please/Please

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
sidseth: Ack. Curious what happens in case the Exception is thrown.

sidseth (approved):
Looks good, except for a tiny update to the message.

autoscale-api/src/main/resources/messages/messages.properties
smaniraju: fixed

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/UpdateFailedHandler.java
smaniraju: Exceptions thrown out of onApplicationEvent are ignored.

PR #10229, opened 2021-03-10
CB-11530 Upload freeipa-client dependencies to nexus

kmanamcheri:
Previously, in CDPSDX-2788, the build scripts were updated to upload
freeipa-client as a jar into nexus. However, freeipa-client has two
dependencies: common, and cluster-proxy. We need to upload these
also to nexus so that thunderhead can use the freeipa-client.

Verified that common and cluster-proxy do not have any further
dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -225,7 +225,7 @@ subprojects {
   task('buildInfo', type: BuildInfoTask, dependsOn: processResources)
 
   afterEvaluate { Project project ->
-    if (project.name in ['core', 'autoscale', 'freeipa', 'redbeams', 'environment', 'datalake', 'integration-test'] && project.plugins.hasPlugin('org.springframework.boot')) {
+    if (project.name in ['core', 'autoscale', 'freeipa', 'redbeams', 'environment', 'datalake', 'integration-test', 'common'] && project.plugins.hasPlugin('org.springframework.boot')) {
       buildInfo.configure {
         destination = file("$project.buildDir")
         basename = project.bootJar.baseName
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -11,6 +11,17 @@ repositories {
   maven { url "https://plugins.gradle.org/m2/" }
 }
 
+uploadBootArchives {
+  repositories {
+    mavenDeployer {
+      configuration = configurations.deployerJars
+      repository(url: "$System.env.NEXUS_URL") {
+        authentication(userName: "$System.env.NEXUS_USER", password: "$System.env.NEXUS_PASSWORD")
+      }
+    }
+  }
+}
+
 jar {
   baseName = 'common'
 }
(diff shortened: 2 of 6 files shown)


jamisonbennett:
I would like to see if we can remove the freeipa client dependency on the cluster proxy. I would think it would be best if we could inject an interface that handles this. That way other consumers that don't know anything about cluster proxy don't need to have the dependency.

lacikaaa:
I think both cluster-proxy and common module has spring-boot dependency so you should use the `uploadBootArchives` task to upload them.
Also I don't know how much is used from these modules. If only a few class maybe it worth duplicating them instead of shipping those with freeipa-client

kmanamcheri:
> I think both cluster-proxy and common module has spring-boot dependency so you should use the `uploadBootArchives` task to upload them.
> Also I don't know how much is used from these modules. If only a few class maybe it worth duplicating them instead of shipping those with freeipa-client

Good point. I'll address both yours and Jamison's comment by trying to remove the dependencies by duplicating them or making them a runtime dependency.

kmanamcheri:
@jamisonbennett I removed the cluster proxy dependency by making the exception classes as interfaces and abstract. This is probably ok assuming that `thunderhead` has its own ClusterProxyException and ClusterProxyError classes.

@lacikaaa I could not remove the `common` dependency because it pulls in a bunch of other things (like Tracer, common-logging framework etc). So I think its best to just get the common module as a dependency as well. I modified to use `uploadBootArchives`

jamisonbennett (approved):

mahendra-rk (approved):
LGTM!

build.gradle
```diff
@@ -232,7 +232,7 @@ subprojects {
         buildVersion = project.version
       }
     }
-    if (project.name in ['freeipa-client']) {
+    if (project.name in ['freeipa-client', 'common']) {
```
lacikaaa, line 235: I think `common` should be added to line 228

scripts/build-dev.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 set -x
 
-./gradlew -Penv=jenkins -b build.gradle buildInfo build uploadBootArchives :freeipa-client:uploadArchives -Pversion=$VERSION --parallel --stacktrace -x checkstyleMain -x checkstyleTest -x spotbugsMain -x spotbugsTest
+./gradlew -Penv=jenkins -b build.gradle buildInfo build uploadBootArchives :freeipa-client:uploadArchives :common:uploadBootArchives -Pversion=$VERSION --parallel --stacktrace -x checkstyleMain -x checkstyleTest -x spotbugsMain -x spotbugsTest
```
lacikaaa, line 4: I don't think you have to add this common related task. the same for `build-release.sh`

kmanamcheri: Is it because there is the 'uploadBootArchives' which runs for all projects?

lacikaaa: yes

lacikaaa:
have you managed to test:
- the upload works for common? (eg with using local nexus)
- `ClusterProxyException` handling in the client still works as before?

PR #10230, opened 2021-03-11
CB-10092: Adding an internal loadbalancer to YCloud datalake setups.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-10092

### Status

This has been tested fully, end-to-end, with additional verification that nothing has been broken for AWS/Azure LD/HA setups.

### Main Changes

- A launchLoadBalancer() method was added to the stack creation flow within the launchStack logic
- A new class called YarnLaunchLoadBalancerService was added which handles all of the launch logic
- The gateway nodes created before the launch loadbalancer steps have their IPs retrieved
- A new docker image (which will be hosted where the default YCloud image is hosted) was created and is being pointed to by the LoadBalancer container created, this image will install the haproxy service and point it at the provided servers
- The loadbalancer container's IP is retrieved in the same way as other YCloud container IPs, but we take advantage of the fact that we know what the loadbalancer application name is and use that instead of anything more complex

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/ResourceConnector.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/ResourceConnector.java
@@ -63,6 +63,20 @@ public interface ResourceConnector<R> {
             AdjustmentType adjustmentType, Long threshold) throws Exception;
 
     /**
+     * Updates an existing stack with one or more load balancers, if the load balances do not already exist. This method will initiate the
+     * creation of load balancers on the Cloud platform, and will configure the load balancer routing in accordance with the specified
+     * target group configuration. It returns a list of CloudResourceStatus for any created load balancers.
+     *
+     * @param authenticatedContext the authenticated context which holds the client object
+     * @param stack                contains the full description of infrastructure
+     * @param persistenceNotifier  Cloud platform notifies the Cloudbreak over this interface if a resource is allocated on the Cloud platfrom
+     * @return the status of load balancers allocated on Cloud platform
+     * @throws Exception in case of any error
+     */
+    List<CloudResourceStatus> launchLoadBalancers(AuthenticatedContext authenticatedContext, CloudStack stack, PersistenceNotifier persistenceNotifier)
+            throws Exception;
+
+    /**
      * Launches a database stack on a cloud platform. The stack consists of the following resources:
      * - a single database server instance
      * - depending on the platform, other associated, required resources (e.g., a DB subnet group for RDS)
@@ -248,19 +262,4 @@ public interface ResourceConnector<R> {
      * @throws TemplatingNotSupportedException if templating is not supported by provider
      */
     String getDBStackTemplate() throws TemplatingNotSupportedException;
-
-    /**
-     * Updates an existing stack with one or more load balancers, if the load balances do not already exist. This method will initiate the
-     * creation of load balancers on the Cloud platform, and will configure the load balancer routing in accordance with the specified
-     * target group configuration. It returns a list of CloudResourceStatus for any created load balancers.
-     *
-     * @param authenticatedContext the authenticated context which holds the client object
-     * @param stack                contains the full description of infrastructure
-     * @param persistenceNotifier  Cloud platform notifies the Cloudbreak over this interface if a resource is allocated on the Cloud platfrom
-     * @return the status of load balancers allocated on Cloud platform
-     * @throws Exception in case of any error
-     */
-    List<CloudResourceStatus> updateLoadBalancers(AuthenticatedContext authenticatedContext, CloudStack stack,
-            PersistenceNotifier persistenceNotifier) throws Exception;
-
 }
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
@@ -104,6 +104,11 @@ public class AwsResourceConnector implements ResourceConnector<Object> {
     }
 
     @Override
+    public List<CloudResourceStatus> launchLoadBalancers(AuthenticatedContext authenticatedContext, CloudStack stack, PersistenceNotifier persistenceNotifier) {
+        return awsLoadBalancerLaunchService.updateCloudformationWithLoadBalancers(authenticatedContext, stack, persistenceNotifier);
+    }
+
+    @Override
     public List<CloudResourceStatus> launchDatabaseServer(AuthenticatedContext ac, DatabaseStack stack,
             PersistenceNotifier persistenceNotifier) throws Exception {
         return awsRdsLaunchService.launch(ac, stack, persistenceNotifier);
@@ -200,11 +205,5 @@ public class AwsResourceConnector implements ResourceConnector<Object> {
         } catch (IOException e) {
             throw new CloudConnectorException("can't get freemarker template", e);
         }
-    }
-
-    @Override
-    public List<CloudResourceStatus> updateLoadBalancers(AuthenticatedContext authenticatedContext, CloudStack stack,
-            PersistenceNotifier persistenceNotifier) {
-        return awsLoadBalancerLaunchService.updateCloudformationWithLoadBalancers(authenticatedContext, stack, persistenceNotifier);
     }
 }
--- a/cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
+++ b/cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
@@ -17,6 +17,8 @@ import com.sequenceiq.cloudbreak.cloud.event.resource.LaunchStackResult;
 import com.sequenceiq.cloudbreak.cloud.init.CloudPlatformConnectors;
 import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
 import com.sequenceiq.cloudbreak.cloud.model.CloudResourceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+import com.sequenceiq.cloudbreak.cloud.model.ResourceStatus;
 import com.sequenceiq.cloudbreak.cloud.notification.PersistenceNotifier;
 import com.sequenceiq.cloudbreak.cloud.scheduler.SyncPollingScheduler;
 import com.sequenceiq.cloudbreak.cloud.task.PollTask;
@@ -58,17 +60,14 @@ public class LaunchStackHandler implements CloudPlatformEventHandler<LaunchStack
         LOGGER.debug("Received event: {}", launchStackRequestEvent);
         LaunchStackRequest request = launchStackRequestEvent.getData();
         CloudContext cloudContext = request.getCloudContext();
+        CloudStack cloudStack = request.getCloudStack();
+        CloudConnector<Object> connector = cloudPlatformConnectors.get(cloudContext.getPlatformVariant());
+        AuthenticatedContext ac = connector.authentication().authenticate(cloudContext, request.getCloudCredential());
         try {
-            CloudConnector<Object> connector = cloudPlatformConnectors.get(cloudContext.getPlatformVariant());
-            AuthenticatedContext ac = connector.authentication().authenticate(cloudContext, request.getCloudCredential());
-            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, request.getCloudStack(), persistenceNotifier,
+            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, cloudStack, persistenceNotifier,
                     request.getAdjustmentType(), request.getThreshold());
-            List<CloudResource> resources = ResourceLists.transform(resourceStatus);
-            PollTask<ResourcesStatePollerResult> task = statusCheckFactory.newPollResourcesStateTask(ac, resources, true);
-            ResourcesStatePollerResult statePollerResult = ResourcesStatePollerResults.build(cloudContext, resourceStatus);
-            if (!task.completed(statePollerResult)) {
-                statePollerResult = syncPollingScheduler.schedule(task);
-            }
+            ResourcesStatePollerResult statePollerResult = waitForResources(ac, resourceStatus, cloudContext);
+            launchLoadBalancers(cloudStack, connector, ac, persistenceNotifier, cloudContext, statePollerResult);
             LaunchStackResult result = ResourcesStatePollerResults.transformToLaunchStackResult(request, statePollerResult);
             request.getResult().onNext(result);
             eventBus.notify(result.selector(), new Event<>(launchStackRequestEvent.getHeaders(), result));
@@ -82,5 +81,44 @@ public class LaunchStackHandler implements CloudPlatformEventHandler<LaunchStack
             request.getResult().onNext(failure);
             eventBus.notify(failure.selector(), new Event<>(launchStackRequestEvent.getHeaders(), failure));
         }
+    }
+
+    /**
+     * Launches the desired load balancers for the stack using the cloud provider resource connector.
+     * Adds the results of the launch to the existing state poller result from the launch of the main stack instances.
+     */
+    private void launchLoadBalancers(CloudStack cloudStack, CloudConnector<Object> connector, AuthenticatedContext authenticatedContext,
+            PersistenceNotifier persistenceNotifier, CloudContext cloudContext, ResourcesStatePollerResult statePollerResult) throws Exception {
+        if (statePollerResult.getResults().stream().anyMatch(result -> ResourceStatus.FAILED.equals(result.getStatus()) ||
+                ResourceStatus.DELETED.equals(result.getStatus()))) {
+            LOGGER.error("Some cloud resources failed creation. Load balancers will not be created.");
+            return;
+        }
+
+        if (!cloudStack.getLoadBalancers().isEmpty()) {
+            List<CloudResourceStatus> loadBalancerResourceStatus = connector.resources().launchLoadBalancers(authenticatedContext, cloudStack,
+                    persistenceNotifier);
+            if (!loadBalancerResourceStatus.isEmpty()) {
+                ResourcesStatePollerResult loadBalancerStatePollerResult = waitForResources(authenticatedContext, loadBalancerResourceStatus, cloudContext);
+                statePollerResult.addResults(loadBalancerStatePollerResult.getResults());
+            }
+        }
+    }
+
+    /**
+     * Creates a poll task which waits for the resources to be fully up and running, or more specifically,
+     * for the resources to be in a permanent successful state.
+     *
+     * Returns the result of the poll task.
+     */
+    private ResourcesStatePollerResult waitForResources(AuthenticatedContext ac, List<CloudResourceStatus> resourceStatuses, CloudContext cloudContext)
+            throws Exception {
+        List<CloudResource> resources = ResourceLists.transform(resourceStatuses);
+        PollTask<ResourcesStatePollerResult> task = statusCheckFactory.newPollResourcesStateTask(ac, resources, true);
+        ResourcesStatePollerResult statePollerResult = ResourcesStatePollerResults.build(cloudContext, resourceStatuses);
+        if (!task.completed(statePollerResult)) {
+            statePollerResult = syncPollingScheduler.schedule(task);
+        }
+        return statePollerResult;
     }
 }
--- a/cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/ApplicationNameUtil.java
+++ b/cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/ApplicationNameUtil.java
@@ -1,7 +1,13 @@
 package com.sequenceiq.cloudbreak.cloud.yarn;
 
+import java.util.List;
+
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
+import com.sequenceiq.cloudbreak.cloud.context.CloudContext;
+import com.sequenceiq.common.api.type.LoadBalancerType;
 
 @Component
 public class ApplicationNameUtil {
@@ -9,21 +15,53 @@ public class ApplicationNameUtil {
     @Value("${cb.max.yarn.resource.name.length:}")
     private int maxResourceNameLength;
 
+    public String createApplicationName(AuthenticatedContext authenticatedContext) {
+        List<String> applicationNameAndUser = initializeApplicationName(authenticatedContext.getCloudContext());
+        return decorateName(applicationNameAndUser.get(0), applicationNameAndUser.get(1));
+    }
+
+    public String createLoadBalancerName(AuthenticatedContext authenticatedContext) {
+        List<String> applicationNameAndUser = initializeApplicationName(authenticatedContext.getCloudContext());
+        return decorateLoadBalancerName(applicationNameAndUser.get(0), applicationNameAndUser.get(1));
+    }
+
     public String decorateName(String name, String userName) {
         if (name.endsWith("-cb")) {
-            name += "-" + userName;
+            name = name + "-" + userName;
         }
         if (!name.endsWith("-cb-" + userName)) {
             name = name + "-cb-" + userName;
         }
-        return cutToMaxLength(name, userName);
+        return cutToMaxLength(name, userName, "-cb-");
     }
 
-    private String cutToMaxLength(String name, String userName) {
+    public String decorateLoadBalancerName(String name, String userName) {
+        if (name.endsWith("-lb")) {
+            name = name + "-" + userName;
+        }
+        if (!name.endsWith("-lb-" + userName)) {
+            name = name + "-lb-" + userName;
+        }
+        return cutToMaxLength(name, userName, "-lb-");
+    }
+
+    public String createLoadBalancerComponentName(String applicationName, LoadBalancerType type) {
+        return applicationName + type.name();
+    }
+
+    private List<String> initializeApplicationName(CloudContext cloudContext) {
+        String name = cloudContext.getName();
+        String id = cloudContext.getId().toString();
+        name += "-" + id;
+        String user = cloudContext.getUserName().split("@")[0].replaceAll("[^a-z0-9-_]", "");
+        return List.of(name, user);
+    }
+
+    private String cutToMaxLength(String name, String userName, String postfixBeginning) {
         if (name.length() <= maxResourceNameLength) {
             return name;
         }
-        String postfix = "-cb-" + userName;
+        String postfix = postfixBeginning + userName;
         String clusterName = name.substring(0, name.indexOf(postfix));
         int newLength = maxResourceNameLength - postfix.length();
         clusterName = clusterName.substring(0, newLength);
--- a/cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
+++ b/cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
@@ -1,7 +1,7 @@
 package com.sequenceiq.cloudbreak.cloud.yarn;
 
 import java.net.MalformedURLException;
-import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
@@ -32,6 +32,7 @@ import com.sequenceiq.cloudbreak.cloud.yarn.client.model.request.ApplicationDeta
 import com.sequenceiq.cloudbreak.cloud.yarn.client.model.response.ApplicationDetailResponse;
 import com.sequenceiq.cloudbreak.cloud.yarn.client.model.response.ApplicationErrorResponse;
 import com.sequenceiq.cloudbreak.cloud.yarn.client.model.response.ResponseContext;
+import com.sequenceiq.cloudbreak.cloud.yarn.loadbalancer.service.launch.YarnLoadBalancerLaunchService;
 import com.sequenceiq.common.api.type.LoadBalancerType;
 import com.sequenceiq.common.api.type.ResourceType;
 
@@ -39,6 +40,12 @@ import com.sequenceiq.common.api.type.ResourceType;
 public class YarnMetadataCollector implements MetadataCollector {
     @Inject
     private YarnClientUtil yarnClientUtil;
+
+    @Inject
+    private ApplicationNameUtil applicationNameUtil;
+
+    @Inject
+    private YarnLoadBalancerLaunchService yarnLoadBalancerLaunchService;
 
     @Override
     public List<CloudVmMetaDataStatus> collect(AuthenticatedContext authenticatedContext, List<CloudResource> resources, List<CloudInstance> vms,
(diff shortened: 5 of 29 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
```diff
@@ -102,6 +103,13 @@
         return awsLaunchService.launch(ac, stack, resourceNotifier, adjustmentType, threshold);
     }
 
+    @Override
+    public List<CloudResourceStatus> launchLoadBalancer(AuthenticatedContext authenticatedContext, CloudStack stack, PersistenceNotifier persistenceNotifier)
```
brycederriso, line 107: **question:** Why are we adding this `launchLoadBalancer` method? I suppose it makes sense conceptually to include a separate API handle for it, but we generally don't have that for other "parts" of the cloud stack.

The `launch` method just handles the whole `CloudStack` object, of which the load balancers are a part.

brycederriso: I can see the utility of something like this when we implement an _upgrade_ path for LBs though. @hreeve-cloudera probably has better ideas here than I do.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/ApplicationNameUtil.java
```diff
@@ -1,29 +1,61 @@
...
         if (!name.endsWith("-cb-" + userName)) {
             name = name + "-cb-" + userName;
         }
-        return cutToMaxLength(name, userName);
+        return cutToMaxLength(name, userName, "-cb-");
+    }
+
+    public String decorateLoadBalancerName(String name, String userName) {
+        if (name.endsWith("-lb")) {
+            name += "-" + userName;
```
brycederriso, line 46: **nit:** Use assignment and plus rather than the `+=` operator.
```suggestion
            name = name + "-" + userName;
```

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
```diff
@@ -138,7 +143,21 @@ private CloudResource getYarnApplcationResource(Iterable<CloudResource> resource
...
-        // no-op
-        return Collections.emptyList();
+        YarnClient yarnClient = yarnClientUtil.createYarnClient(ac);
+        String loadBalancerApplicationName = applicationNameUtil.createLoadBalancerName(ac);
+        Iterable<Container> loadBalancerContainers = getContainers(loadBalancerApplicationName, yarnClient);
+        Iterator<Container> containerIterator = loadBalancerContainers.iterator();
+
+        List<CloudLoadBalancerMetadata> loadBalancerMetadata = Lists.newArrayList();
+        for (LoadBalancerType loadBalancerType : loadBalancerTypes) {
+            Container container = containerIterator.next();
```
brycederriso, line 153: **question:** I don't think I'm following this method, will we always have the same number of Containers as `LoadBalancerType`s?

Here, the metadata is responsible for _assigning_ rather than looking up the container information?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnResourceConnector.java
```diff
@@ -110,11 +123,7 @@ private void createApplication(YarnClient yarnClient, CreateApplicationRequest c
     }
 
     private CreateApplicationRequest createRequest(CloudStack stack, String applicationName) {
-        CreateApplicationRequest createApplicationRequest = new CreateApplicationRequest();
-        createApplicationRequest.setName(applicationName);
-        createApplicationRequest.setQueue(stack.getParameters().getOrDefault(YarnConstants.YARN_QUEUE_PARAMETER, defaultQueue));
-        String lifeTimeStr = stack.getParameters().get(YarnConstants.YARN_LIFETIME_PARAMETER);
-        createApplicationRequest.setLifetime(lifeTimeStr != null ? Integer.parseInt(lifeTimeStr) : defaultLifeTime);
+        CreateApplicationRequest createApplicationRequest = initializeRequest(stack, applicationName, defaultQueue, defaultLifeTime);
```
brycederriso, line 126: **praise:** Good method extraction.

brycederriso (commented):
Left some comments, but didn't really get to dig in to `LoadBalancerLaunchService`.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
sxxgrc: At least for the Yarn case this was kind of necessary to get all of the logic in the right place. Essentially, we wanted to be able to launch the loadbalancer after the other containers were launched so we could get their IP addresses, but still have the logic in a place where we had access to the stack and could persists the resource we would create for the LB, and this was the best place for it.
Haley mentioned that this would be better in order to isolate the logic completely to help with future changes and the like.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/ResourceConnector.java
```diff
@@ -60,6 +60,22 @@
     List<CloudResourceStatus> launch(AuthenticatedContext authenticatedContext, CloudStack stack, PersistenceNotifier persistenceNotifier,
             AdjustmentType adjustmentType, Long threshold) throws Exception;
 
+    /**
+     * Launches a loadbalancer instance on the Cloud platform.
+     * <br>
+     * This method shall initiate the infrastructure creation on Cloud platform and shall return of a list of {@link CloudResourceStatus} values. It does not
```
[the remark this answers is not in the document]
sxxgrc: Will do, thank you.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/ApplicationNameUtil.java
sxxgrc: Will do!

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
sxxgrc: Good question, for Yarn this will always be true since Yarn will only ever have public loadbalancers.
In general though, as this method is for getting all of the loadbalancer metadata, and the only inpus is a list of the types, I kind of assume that at most there will be the same number of loadbalancers as the amount of unique types used. @hreeve-cloudera can clarify this some more.

I was just trying to make sure I future proof this in case Yarn does end up using private loadbalancers and the like in the future.

sxxgrc:
Made @brycederriso 's desired changes and answered questions.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
hreeve-cloudera: We do have an upgrade path for load balancers. Bryce, you and I have talked about it briefly. But basically in CB-10612 I added a new API that, if called, will add load balancers to an existing environment. As part of that I separated out the AWS load balancer logic into a new `AwsLoadBalancerLaunchService`. So for AWS rolling with this change will be simple, because we can simply move where `AwsLoadBalancerLaunchService` is being currently called to be called here instead.

Ultimately we need to be able to call the LB creation logic separately of the stack creation anyway, to support that in-place upgrade API. So I didn't see any harm in separating things out like that, and told Santiago to go for it, and we would figure out how to implement it across the other cloud platforms as necessary.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/LoadBalancerLaunchService.java
```diff
@@ -0,0 +1,259 @@
...
+            }
+        });
+        return gatewayIPs;
+    }
+
+    /**
+     * Gets all of the containers currently running in the YCloud ecosystem within the application
+     * given by the application name provided.
+     */
+    public static Iterable<Container> getContainers(String applicationName, YarnClient yarnClient) {
```
hreeve-cloudera, line 133: Since `LoadBalancerLaunchService` is an injectable object, don't make this static. Instead just inject `LoadBalancerLaunchService` where this needs to be called, and call it through that object.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
```diff
@@ -138,7 +143,21 @@ private CloudResource getYarnApplcationResource(Iterable<CloudResource> resource
...
+            CloudLoadBalancerMetadata metadata = new CloudLoadBalancerMetadata.Builder()
+                    .withType(loadBalancerType)
+                    .withName(loadBalancerApplicationName + loadBalancerType.name())
```
hreeve-cloudera, line 156: Where is this naming convention being set? If it's used in other places, then we should have a shared method that can be called to generate this, so that it doesn't get out of sync if we change it one place and forget to update it elsewhere.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/LoadBalancerLaunchService.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Service
+public class LoadBalancerLaunchService {
```
hreeve-cloudera, line 50: Could you name this `YarnLoadBalancerLaunchService`?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/LoadBalancerLaunchService.java
```diff
@@ -0,0 +1,259 @@
...
+    private List<YarnComponent> createLoadBalancerComponents(CloudStack cloudStack, List<String> gatewayIPs, String applicationName) {
+        LOGGER.debug("Creating the loadbalancer components for application " + applicationName + " with gatewayIPs: " + gatewayIPs.toString() + ".");
+        Artifact artifact = createLoadBalancerArtifact();
+        Resource resource = createLoadBalancerResource();
+        String launchCommand = createLoadBalancerLaunchCommand(cloudStack);
+        Configuration configuration = createLoadBalancerConfiguration(gatewayIPs);
+
+        List<YarnComponent> loadBalancerComponents = Lists.newArrayList();
+        for (CloudLoadBalancer loadBalancer : cloudStack.getLoadBalancers()) {
+            String componentName = applicationName + loadBalancer.getType().name();
```
hreeve-cloudera, line 181: Is this the other place you're setting the LB name? Referring to my comment in `YarnMetadataCollector.collectLoadBalancer`. Yeah, please make some shared method both of these can call to generate the LB name. I realize it's just a simple string concatenation, but we don't want these to end up mismatched.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnMetadataCollector.java
hreeve-cloudera: It looks to me like in `LoadBalancerLaunchService`, you're creating one component per load balancer in the CloudStack, and each component has one container. So logically if there were two load balancers, there would be two components with a combined two containers. Is that right? That makes logical sense to me, and would match what seems to be being done in the for loop here.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsResourceConnector.java
hreeve-cloudera: Actually, Santiago, I don't think you need a new method here. This is functionally just a duplicate of the `updateLoadBalancers` method I already added to the `ResourceConnector` interface. So let's consolidate that. (We can rename the existing method to `launchLoadBalancer` if you like, to match the other `launch` method.)

sxxgrc: When in the flow is that called? It might require a little bit of work to get it working in that order, but not sure.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/LoadBalancerLaunchService.java
sxxgrc: Ahh that makes sense, sorry still getting used to Spring stuff.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -187,6 +188,9 @@ public boolean isLoadBalancerCreationConfigured(Stack stack, DetailedEnvironment
                 } else {
                     LOGGER.debug("External load balancer creation is disabled.");
                 }
+                if (CloudPlatform.YARN.equalsIgnoreCase(stack.getCloudPlatform())) {
+                    setupLoadBalancer(dryRun, stack, loadBalancers, knoxTargetGroup.get(), LoadBalancerType.PUBLIC);
+                }
```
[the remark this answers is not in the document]
sxxgrc: Yes that is true.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -246,13 +250,13 @@ private boolean isLoadBalancerEnabled(StackType type, DetailedEnvironmentRespons
 
     private boolean isLoadBalancerEnabledForDatalake(StackType type, DetailedEnvironmentResponse environment) {
         return StackType.DATALAKE.equals(type) && environment != null &&
-            (entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId()) ||
-                isLoadBalancerEntitlementNotRequiredForCloudProvider(environment.getCloudPlatform()) ||
+            (!isLoadBalancerEntitlementRequiredForCloudProvider(environment.getCloudPlatform()) ||
+                entitlementService.datalakeLoadBalancerEnabled(ThreadBasedUserCrnProvider.getAccountId()) ||
```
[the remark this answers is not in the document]
sxxgrc: Whoops sorry, I remember you had said in my previous PR for removing the entitlement that this order was more optimal since it just checks a string first, and then checks the entitlement which is a more expensive operation.
I can revert this if you want.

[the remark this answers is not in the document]
hreeve-cloudera: I suspected that may be the reason :). That is generally true, yes, but in this particular case I did it for a reason. You couldn't have known that though, so I appreciate you trying to optimize the code. But yes, please put it back.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
```diff
@@ -58,16 +66,20 @@ public void accept(Event<LaunchStackRequest> launchStackRequestEvent) {
...
             AuthenticatedContext ac = connector.authentication().authenticate(cloudContext, request.getCloudCredential());
-            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, request.getCloudStack(), persistenceNotifier,
+            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, cloudStack, persistenceNotifier,
                     request.getAdjustmentType(), request.getThreshold());
-            List<CloudResource> resources = ResourceLists.transform(resourceStatus);
-            PollTask<ResourcesStatePollerResult> task = statusCheckFactory.newPollResourcesStateTask(ac, resources, true);
-            ResourcesStatePollerResult statePollerResult = ResourcesStatePollerResults.build(cloudContext, resourceStatus);
-            if (!task.completed(statePollerResult)) {
-                statePollerResult = syncPollingScheduler.schedule(task);
+            ResourcesStatePollerResult statePollerResult = waitForResources(ac, resourceStatus, cloudContext);
```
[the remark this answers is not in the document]
sxxgrc: Will do before I merge.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
```diff
@@ -83,4 +95,34 @@ public void accept(Event<LaunchStackRequest> launchStackRequestEvent) {
...
+            statePollerResult = syncPollingScheduler.schedule(task);
+        }
+        return statePollerResult;
+    }
+
+    /**
+     * Checks to see if load balancers are enabled for this account, also verifying that the entitlement
+     * can be obtained successfully so as to keep errors out of the main accept call.
+     */
+    private boolean isLoadBalancerEnabled(Platform cloudPlatform) {
```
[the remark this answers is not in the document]
sxxgrc: So this method is in another package than the one for the LaunchStackHandler, which would require a dependency to be added which tbh I have not done before. How should I handle that?

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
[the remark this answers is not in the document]
hreeve-cloudera: Ah, so it isn't. Thinking on it more, I think you can just check if `cloudStack.getLoadBalancers().isEmpty()`. The load balancers objects are only created if all the validation checks in `LoadBalancerConfigService` passed, so I don't think we need to check them again.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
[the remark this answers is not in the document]
hreeve-cloudera: Ah, so it isn't. Thinking on it more, I think you can just check if `cloudStack.getLoadBalancers().isEmpty()`. The load balancers objects are only created if all the validation checks in `LoadBalancerConfigService` passed, so I don't think we need to check them again.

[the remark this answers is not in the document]
sxxgrc: Okay I'll do that then.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/YarnResourceConnector.java
```diff
@@ -213,7 +202,7 @@ private boolean checkApplicationAlreadyCreated(YarnClient yarnClient, String app
                     DeleteApplicationRequest deleteApplicationRequest = new DeleteApplicationRequest();
                     deleteApplicationRequest.setName(yarnApplicationName);
                     yarnClient.deleteApplication(deleteApplicationRequest);
-                    LOGGER.debug("Yarn Applicatin has been deleted");
+                    LOGGER.debug("Yarn Application has been deleted");
```
biharitomi, line 205: Can we have this on INFO level?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+
+    public CloudResource launch(AuthenticatedContext authenticatedContext, CloudStack cloudStack, PersistenceNotifier persistenceNotifier, YarnClient yarnClient)
+            throws Exception {
+        String applicationName = applicationNameUtil.createLoadBalancerName(authenticatedContext);
+
+        if (!yarnApplicationCreationService.checkApplicationAlreadyCreated(yarnClient, applicationName)) {
+            LOGGER.debug("Creating the load balancer application for the Yarn datalake.");
+            CreateApplicationRequest createApplicationRequest = createLoadBalancerRequest(cloudStack, applicationName, authenticatedContext, yarnClient);
+            yarnApplicationCreationService.createApplication(yarnClient, createApplicationRequest);
+            LOGGER.debug("Successfully created the Yarn load balancer application.");
```
biharitomi, line 82: Can we have INFO level here?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+        Iterable<Container> foundContainers = getContainers(applicationName, yarnClient);
+        Set<String> gatewayGroupNames = getGatewayGroupNames(cloudStack);
+
+        List<String> gatewayIPs = Lists.newArrayList();
+        foundContainers.forEach(container -> {
+            if (gatewayGroupNames.contains(container.getComponentName())) {
+                gatewayIPs.add(container.getIp() + ":443");
+            }
+        });
+        LOGGER.debug("Successfully found the following gateway IPs for the load balancer: " + gatewayIPs + ".");
```
biharitomi, line 125: Info level?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+        LOGGER.debug("Getting the Yarn containers for application " + applicationName + ".");
+        ApplicationDetailRequest applicationDetailRequest = new ApplicationDetailRequest();
+        applicationDetailRequest.setName(applicationName);
+        ResponseContext responseContext;
+
+        try {
+            responseContext = yarnClient.getApplicationDetail(applicationDetailRequest);
+            LOGGER.debug("Successfully for a response for Yarn application " + applicationName + " from the Yarn client.");
+        } catch (MalformedURLException ex) {
+            LOGGER.error("Failed to get information for the Yarn loadbalancer! Application name: " + applicationName + " Error: " + ex.getMessage());
```
biharitomi, line 143: Warn level?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+            throw new CloudConnectorException("Failed to get information for the Yarn loadbalancer.", ex);
+        }
+
+        if (responseContext.getStatusCode() == YarnResourceConstants.HTTP_SUCCESS) {
+            LOGGER.debug("Successfully retrieved container information for the Yarn application " + applicationName + ".");
+            ApplicationDetailResponse applicationDetailResponse = (ApplicationDetailResponse) responseContext.getResponseObject();
+            return applicationDetailResponse.getContainers();
+        } else {
+            LOGGER.error("Failed to get yarn container! Application name: " + applicationName);
```
biharitomi, line 152: Warn level?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+        try {
+            responseContext = yarnClient.getApplicationDetail(applicationDetailRequest);
+            LOGGER.debug("Successfully for a response for Yarn application " + applicationName + " from the Yarn client.");
+        } catch (MalformedURLException ex) {
+            LOGGER.error("Failed to get information for the Yarn loadbalancer! Application name: " + applicationName + " Error: " + ex.getMessage());
+            throw new CloudConnectorException("Failed to get information for the Yarn loadbalancer.", ex);
+        }
+
+        if (responseContext.getStatusCode() == YarnResourceConstants.HTTP_SUCCESS) {
+            LOGGER.debug("Successfully retrieved container information for the Yarn application " + applicationName + ".");
```
biharitomi, line 148: Info level?

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/YarnLoadBalancerLaunchService.java
```diff
@@ -0,0 +1,264 @@
...
+                .map(Group::getName).collect(Collectors.toSet());
+    }
+
+    /**
+     * Creates a Yarn component for each of the loadbalancers that are to be created, using a set of pre-defined
+     * constants for the various component parameters.
+     *
+     * Each container is pointed at a specific docker image which contains the loadbalancing service and logic.
+     */
+    private List<YarnComponent> createLoadBalancerComponents(CloudStack cloudStack, List<String> gatewayIPs, String applicationName) {
```
biharitomi, line 174: I would move the creation of the components into an individual service/factory. The `YarnLoadBalancerLaunchService.java` would be much more clean.

biharitomi (changes requested):

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
```diff
@@ -58,16 +59,20 @@ public void accept(Event<LaunchStackRequest> launchStackRequestEvent) {
...
             AuthenticatedContext ac = connector.authentication().authenticate(cloudContext, request.getCloudCredential());
-            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, request.getCloudStack(), persistenceNotifier,
+            List<CloudResourceStatus> resourceStatus = connector.resources().launch(ac, cloudStack, persistenceNotifier,
                     request.getAdjustmentType(), request.getThreshold());
-            List<CloudResource> resources = ResourceLists.transform(resourceStatus);
-            PollTask<ResourcesStatePollerResult> task = statusCheckFactory.newPollResourcesStateTask(ac, resources, true);
-            ResourcesStatePollerResult statePollerResult = ResourcesStatePollerResults.build(cloudContext, resourceStatus);
-            if (!task.completed(statePollerResult)) {
-                statePollerResult = syncPollingScheduler.schedule(task);
+            ResourcesStatePollerResult statePollerResult = waitForResources(ac, resourceStatus, cloudContext);
```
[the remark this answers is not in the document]
sxxgrc: What exactly do you mean by this? Should I check to see if the status is equal to FAILED or DELETED?
And if so, should I just not do the load balancer stuff, or handle that differently?

sxxgrc:
Made most of the changes requested by @biharitomi 
I went ahead and resolved most of the conversations here as well to make it easier to go through this PR. I left only the ones I would like to be double checked as well as the single requested change from Tamas that I want further clarification on before I work on.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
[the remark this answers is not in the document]
hreeve-cloudera: @sxxgrc The `ResourcesStatePollerResult` contains the state of all the resources created in the cloud. At this point in the process, that should be all the instances, network resources, etc. What Tamas means is that we should check the state of those resources, and if they're failed we shouldn't do the load balancer stuff, since the creation will ultimately fail anyway. I think that would look something like this:
```
ResourcesStatePollerResult statePollerResult = waitForResources(ac, resourceStatus, cloudContext);
if (statePollerResult.getResults().stream().noneMatch(result -> ResourceStatus.FAILED.equals(result.getStatus()))) {
    launchLoadBalancers(cloudStack, connector, ac, persistenceNotifier, cloudContext, statePollerResult);
} else {
    LOGGER.error("Some cloud resources failed creation. Load balancers will not be created.");
}
LaunchStackResult result = ResourcesStatePollerResults.transformToLaunchStackResult(request, statePollerResult);
```
This isn't really something you've introduced. It should have been done earlier, but wasn't.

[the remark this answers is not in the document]
biharitomi: @sxxgrc as @hreeve-cloudera mentioned can we get that additional condition? I can imagine that condition in the new `launchLoadBalancers` method too.

biharitomi:
I checked the code after you have processed the review items and only one more review item left:
https://github.com/hortonworks/cloudbreak/pull/10230/files#r599465559
Otherwise it looks good to me.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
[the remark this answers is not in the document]
sxxgrc: Okay I added this condition to the launchLoadBalancers method. Let me know if y'all need anything else.

sxxgrc:
Made the last change requested by @biharitomi

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchStackHandler.java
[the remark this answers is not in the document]
sxxgrc: By the way, the resulting LaunchStackResult object will contain the statuses for both the initial datalake instance launch and the loadbalancer launch, so that it will be aware of any failures across both launches and that will be handled by the general launch failure handling done later in the flow.

sxxgrc:
Added the final docker image endpoint, which can be found here: 
https://github.com/hortonworks/cloudbreak/pull/10230/files#diff-4ed62beae7d4bfe342a477f7137505bb4f17d6860d5043cd505353369ef5580dR37

I will run some final tests on this, but I will make this mergeable now.

cloud-yarn/src/main/java/com/sequenceiq/cloudbreak/cloud/yarn/loadbalancer/service/launch/LoadBalancerLaunchService.java
```diff
@@ -0,0 +1,259 @@
...
+    LoadBalancerLaunchService(ApplicationNameUtil applicationNameUtil) {
+        Objects.requireNonNull(applicationNameUtil);
+        this.applicationNameUtil = applicationNameUtil;
+    }
+
+    public CloudResource launch(AuthenticatedContext authenticatedContext, CloudStack cloudStack, PersistenceNotifier persistenceNotifier, YarnClient yarnClient)
+            throws Exception {
+        String applicationName = applicationNameUtil.createLoadBalancerName(authenticatedContext);
+
+        if (!checkApplicationAlreadyCreated(yarnClient, applicationName)) {
```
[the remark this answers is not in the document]
sxxgrc: I created the `YarnApplicationCreationService` class to remedy this, putting all of the methods they both call there. Now only the `YarnResourceConnector` injects the `LoadBalancerLaunchService` but not vice versa.

hreeve-cloudera (approved):

biharitomi (approved):
LGTM

PR #10231, opened 2021-03-11
CB-11070 CDP PC missing firewall for egress/ingress

hreeve-cloudera:
This is the second version of this change, which fixes a performance problem in the original commit. The
updated change no longer issues a DescribeSubnetRequest recursively on all subnets. Instead, a single
DescribeSubnetRequest is issued to fetch results for all relevant subnets, and the results are then iterated
over to find the description of each desired subnet. Original commit message is below:

Adds additional pass conditions to the logic that validates the AWS subnets provided for the public endpoint
gateway load balancer. Previously a subnet could only be used if it had an attached internet gateway. Now
it can be used if it passes one of the following three conditions:
  1. The subnet is tagged with "cdp-public-endpoint-gateway-lb" : "override".
  2. The subnet has an internet gateway attached.
  3. The subnet is routed through an AWS network firewall, and that firewall has an attached IGW.

Tested with unit tests and by enabling the feature with subnets with the override tag, and subnets routed
though a public network firewall.

Original PR for reference: https://github.com/hortonworks/cloudbreak/pull/10149

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.cloud.aws;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -11,15 +12,32 @@ import org.springframework.stereotype.Component;
 
 import com.amazonaws.services.ec2.model.Route;
 import com.amazonaws.services.ec2.model.RouteTable;
+import com.amazonaws.services.ec2.model.Subnet;
+import com.amazonaws.services.ec2.model.Tag;
+import com.amazonaws.services.networkfirewall.model.AWSNetworkFirewallException;
+import com.amazonaws.services.networkfirewall.model.DescribeFirewallRequest;
+import com.amazonaws.services.networkfirewall.model.FirewallMetadata;
+import com.amazonaws.services.networkfirewall.model.ListFirewallsRequest;
+import com.amazonaws.services.networkfirewall.model.SyncState;
+import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonNetworkFirewallClient;
+import com.sequenceiq.cloudbreak.cloud.aws.util.AwsPageCollector;
 
 @Component
 public class AwsSubnetIgwExplorer {
+
+    static final String ENDPOINT_GATEWAY_OVERRIDE_KEY = "cdp-public-endpoint-gateway-lb";
+
+    static final String ENDPOINT_GATEWAY_OVERRIDE_VALUE = "override";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AwsSubnetIgwExplorer.class);
 
     private static final String OPEN_CIDR_BLOCK = "0.0.0.0/0";
 
     private static final String IGW_PREFIX = "igw-";
+
+    private static final String FW_PREFIX = "vpce-";
+
+    private static final String ACCESS_DENIED = "AccessDeniedException";
 
     public boolean hasInternetGatewayOfSubnet(List<RouteTable> routeTables, String subnetId, String vpcId) {
         Optional<RouteTable> routeTable = getRouteTableForSubnet(routeTables, subnetId, vpcId);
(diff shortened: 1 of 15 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
```diff
@@ -68,4 +86,140 @@ public boolean hasInternetGatewayOfSubnet(List<RouteTable> routeTables, String s
...
+
+    private List<String> getPotentialFirewallEndpointIds(List<RouteTable> routeTables, String subnetId, String vpcId) {
+        List<String> gatewayIds = new ArrayList<>();
+        LOGGER.info("Searching subnet {} route table for potential firewall routing", subnetId);
+        Optional<RouteTable> subnetRouteTable = getRouteTableForSubnet(routeTables, subnetId, vpcId);
+        LOGGER.info("Searching subnet route table for any routes that start with the firewall gateway prefix ({})", FW_PREFIX);
+        subnetRouteTable.ifPresent(
+            routeTable ->
+                gatewayIds.addAll(subnetRouteTable.get().getRoutes().stream()
+                    .filter(route -> StringUtils.isNotEmpty(route.getGatewayId()) && route.getGatewayId().startsWith(FW_PREFIX))
```
jamisonbennett, line 162: I am not sure if other cloudbreak code operations this way or not. But I have a minor concern that using the prefix might catch things we don't intend. Maybe this is a common practice in cloudbreak, if so then I am OK with it.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
```diff
@@ -68,4 +86,140 @@ public boolean hasInternetGatewayOfSubnet(List<RouteTable> routeTables, String s
...
+                    .map(Route::getGatewayId)
+                    .collect(Collectors.toList())));
+        return gatewayIds;
+    }
+
+    private List<FirewallMetadata> getAllFirewallMetadataForVpc(AmazonNetworkFirewallClient nfwClient, String vpcId) {
+        ListFirewallsRequest request = new ListFirewallsRequest()
+            .withVpcIds(vpcId);
+        return AwsPageCollector.getAllFirewallMetadata(nfwClient, request);
```
jamisonbennett, line 171: Do you have to worry about setting the limits for the query?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
```diff
@@ -68,4 +86,140 @@ public boolean hasInternetGatewayOfSubnet(List<RouteTable> routeTables, String s
...
+        return result;
+    }
+
+    private List<AwsFirewall> getFirewalls(List<FirewallMetadata> firewallMetadata, AmazonNetworkFirewallClient nfwClient) {
+        List<String> firewallArns = firewallMetadata.stream()
+            .map(FirewallMetadata::getFirewallArn)
+            .collect(Collectors.toList());
+        LOGGER.debug("Issuing DescribeFirewallRequest for firewalls with arns {}", firewallArns);
+        return firewallArns.stream()
+            .map(arn -> nfwClient.describeFirewall(new DescribeFirewallRequest().withFirewallArn(arn)))
```
jamisonbennett, line 208: How many RPCs does this perform and do you think a bulk API might be useful (assuming one exists)?

jamisonbennett (commented):
It looks good. I have a few minor questions, nothing major. Thanks.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsSubnetIgwExplorer.java
hreeve-cloudera: It will run this request for each firewall in the VPC. There's unfortunately a bit of a disconnect in the way the data's presented that means we need to make this call. The subnet's route table will have all gateway ids, including firewalls, and the ListFirewallsRequest will give us the firewall arn and names. But the firewall gateway id is only visible in the DescribeFirewallResult. So we can't tell if the gateway ids found in the subnet's route table are firewall or not until we do this call.

I couldn't find a bulk API to use. However, most customers will never hit this code. If order for this check to work, you need to add additional permissions to your default AWS credential. If you don't have those, you'll get an `ACCESS_DENIED` earlier in the code and it will just move on. So this shouldn't be a general performance hit, and will only be encountered by customers who specifically enable it by adding the additional permissions.

jamisonbennett: Ok, thanks for checking.

jamisonbennett (approved):

PR #10236, opened 2021-03-11
CB-11518 Return more meaningful error code and message in FreeIPA whe…

lacikaaa:
…n it's deleted

- `KerberosMgmtV1Controller` won't wrap `NotFoundException` anymore -> will return 404 and meaningful message when there is no FreeIPA for env
- replaced FMS specific `NotFoundException` with the one from common module as they are the same and in FMS we used both


jamisonbennett (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
```diff
@@ -165,18 +166,18 @@ private String checkActorCrn() {
...
+        } catch (DeleteException | InvalidFreeIpaStateException | NotFoundException e) {
             throw e;
         } catch (RetryableFreeIpaClientException e) {
             LOGGER.error("internal server error", e);
             Exception exceptionForRestApi = e.getExceptionForRestApi();
             if (exceptionForRestApi instanceof DeleteException) {
                 throw (DeleteException) exceptionForRestApi;
             }
-            throw new DeleteException(DELETION_FAILED);
+            throw new DeleteException(DELETION_FAILED + e.getMessage());
```
topolyai5, line 177: What is the difference between the `e` and the `exceptionForRestApi`. Maybe it better to throw with that message and log here the retry exception

jamisonbennett:
Thanks for making the error message more meaningful.

topolyai5 (commented):
A `DeleteExceptionMapper` would be good to avoid the 500 error on that case.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtV1Controller.java
lacikaaa: I had to check the code. We wrap most of the exceptions from the original one into a new one in case of KerberosMgmt. This means if we can't create keytab in FreeIPA, we would throw a `KeytabCreationException` instead of `FreeIpaClientException`. And we put this `KeytabCreationException` into `RetryableFreeIpaClientException` so our code would try it again, but if it fails multiple times we take the inner/original exception from it and return that to the API.

topolyai5: So, would be better to add the original message to the exception, instead of the wrapped

topolyai5 (changes requested):
Could you please add a mapper to DeleteException

lacikaaa:
> Could you please add a mapper to DeleteException

I'm not sure what you would like to map it to.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/RetryableFreeIpaClientException.java
```diff
@@ -35,4 +35,9 @@ public RetryableFreeIpaClientException(String message, RetryableFreeIpaClientExc
     public Exception getExceptionForRestApi() {
         return exceptionForRestApi;
     }
+
+    @Override
+    public String toString() {
+        return super.toString() + System.lineSeparator() + "ExceptionForRestApi: "  + exceptionForRestApi.toString();
```
lacikaaa, line 41: @jamisonbennett does this change is useful in your opinion? I just want to be sure we won't miss anything when logging this exception

jamisonbennett (approved):

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/RetryableFreeIpaClientException.java
jamisonbennett: I think that should catch everything.

topolyai5:
> > Could you please add a mapper to DeleteException
> 
> I'm not sure what you would like to map it to.

I mean the exception mapper. In this case, we respond with an internal error, but I think it is an expected exception

topolyai5 (approved):
As we talked, we cannot map to other status code, because we wrap a lot of exception type into this reception type

PR #10235, opened 2021-03-11
CB-11535 it looks like that due to the circular dependency, Spring st…

akanto:
…arts to initialise the AwsSessionCredentialClient bean very early in the process, even before it starts to initialise the caching system.Since the caching system is not initialised therefore the @Cacheable annotation is not going to have any effect.

This is not a proper fix, just a workaround and most probably it reverts the benefits of "CB-10938 Decode authorization errors"

See detailed description in the commit message.

PR #10237, opened 2021-03-11
CB-11533 Fix AWS credential cache

Bajzathd:
AwsSessionCredentialClient was not caching credentials, because when setting it to AwsClient it did not have any proxies.
Fixed it by creating the sts client there instead and injecting it to AwsClient (so we do not have circular dependency issues), and disabled the checkstyle check to allow aws client class imports in that file.


lajosrodek:
Follow-up to #10235.

Bajzathd:
Fixed in #10235

PR #10239, opened 2021-03-11
CB-11341 refactor cluster install flow for different segments. Now it…

doktoric:
… is not a big giant method but multiple step.


bergerdenes (changes requested):
I still have not checked your largest files yet

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationFlowConfig.java
```diff
@@ -60,34 +108,134 @@
         implements RetryableFlowConfiguration<ClusterCreationEvent> {
     private static final List<Transition<ClusterCreationState, ClusterCreationEvent>> TRANSITIONS =
             new Builder<ClusterCreationState, ClusterCreationEvent>().defaultFailureEvent(CLUSTER_CREATION_FAILED_EVENT)
-            .from(INIT_STATE).to(CLUSTER_PROXY_REGISTRATION_STATE).event(CLUSTER_CREATION_EVENT).noFailureEvent()
-            .from(CLUSTER_PROXY_REGISTRATION_STATE).to(BOOTSTRAPPING_MACHINES_STATE).event(CLUSTER_PROXY_REGISTRATION_SUCCEEDED_EVENT)
+            .from(INIT_STATE)
+                    .to(CLUSTER_PROXY_REGISTRATION_STATE)
+                    .event(CLUSTER_CREATION_EVENT)
+                    .noFailureEvent()
```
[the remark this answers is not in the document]
doktoric: this is the same as before just formatted the file

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/ClusterCreationEvent.java
```diff
@@ -48,10 +80,42 @@
...
+    SUPPRESS_WARNINGS_FAILED_EVENT(EventSelectorUtil.selector(SupressClusterWarningsFailed.class)),
+    CONFIGURE_KERBEROS_SUCCESS_EVENT(EventSelectorUtil.selector(ClusterManagerConfigureKerberosSuccess.class)),
+    CONFIGURE_KERBEROS_FAILED_EVENT(EventSelectorUtil.selector(ClusterManagerConfigureKerberosFailed.class)),
+    EXECUTE_POST_INSTALL_RECIPES_FINISHED_EVENT(EventSelectorUtil.selector(ExecutePostInstallRecipesSuccess.class)),
+    EXECUTE_POST_INSTALL_RECIPES_FAILED_EVENT(EventSelectorUtil.selector(ExecutePostInstallRecipesFailed.class)),
+    HANDLE_CLUSTER_CREATION_SUCCESS_FINISHED_EVENT(EventSelectorUtil.selector(HandleClusterCreationSuccessSuccess.class)),
+    HANDLE_CLUSTER_CREATION_SUCCESS_FAILED_EVENT(EventSelectorUtil.selector(HandleClusterCreationSuccessFailed.class)),
+    PREPARE_DATALAKE_RESOURCE_FINISHED_EVENT(EventSelectorUtil.selector(PrepareDatalakeResourceSuccess.class)),
+    PREPARE_DATALAKE_RESOURCE_FAILED_EVENT(EventSelectorUtil.selector(PrepareDatalakeResourceFailed.class)),
+    START_AMBARI_FINISHED_EVENT(EventSelectorUtil.selector(StartClusterSuccess.class)),
```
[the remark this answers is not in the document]
doktoric: nothing changed on the old steps just added new steps

doktoric:
retest this please

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
```diff
@@ -191,13 +266,47 @@ public void buildCluster(Long stackId) throws CloudbreakException, ClusterClient
...
     }
 
-    private void setupProxy(ClusterApi connector, ProxyConfig proxyConfig) {
-        if (proxyConfig != null) {
-            LOGGER.info("proxyConfig is not null, setup proxy for cluster: {}", proxyConfig);
-            connector.clusterSetupService().setupProxy(proxyConfig);
-        } else {
-            LOGGER.info("proxyConfig was not found by proxyConfigCrn");
-        }
+    private String sdxStackCrn(Stack stack) {
```
gregito, line 269: I would rename this and all the other **new** "getter" functions below this to a more descriptive one.
Let's say this from sdxStackCrn -> getSdxStackCrnByStack
Or for example getClusterSetupService -> getClusterSetupServiceByStackId
Or datalakeResources -> getDatalakeResourcesByStack

And so on. 

These are private function be the have to have as descriptive names as possible to avoid unnecessary jumping inside to code (to figure out what am I actually calling)

gregito (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerConfigureKerberosHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+import reactor.bus.Event;
+import reactor.bus.EventBus;
+
+@Component
+public class ClusterManagerConfigureKerberosHandler implements EventHandler<ClusterManagerConfigureKerberosRequest> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClusterManagerConfigureKerberosHandler.class);
+
+    @Inject
```
gregito, line 26: I would use constructor injection instead of field since this is the suggested way - by the spring guys - of doing it

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerConfigureKerberosHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    private EventBus eventBus;
+
+    @Inject
```
gregito, line 29: same goes here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerPrepareProxyConfigHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import reactor.bus.Event;
+import reactor.bus.EventBus;
+
+@Component
+public class ClusterManagerPrepareProxyConfigHandler implements EventHandler<ClusterManagerPrepareProxyConfigRequest> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClusterManagerPrepareProxyConfigHandler.class);
+
+    @Inject
```
gregito, line 25: also here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerPrepareProxyConfigHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    private EventBus eventBus;
+
+    @Inject
```
gregito, line 28: and here

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerRefreshParcelClusterHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import reactor.bus.Event;
+import reactor.bus.EventBus;
+
+@Component
+public class ClusterManagerRefreshParcelClusterHandler implements EventHandler<ClusterManagerRefreshParcelRequest> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClusterManagerRefreshParcelClusterHandler.class);
+
+    @Inject
```
gregito, line 25: oh c'mon, you get it 😄  please dont' make me to comment all the **@Inject**'s 😄

topolyai5: Maybe we shouldn't mix the injections

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerSetupMonitoringHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void accept(Event<ClusterManagerSetupMonitoringRequest> event) {
+        Long stackId = event.getData().getResourceId();
+        Selectable response;
+        try {
+            clusterBuilderService.setupMonitoring(stackId);
+            response = new ClusterManagerSetupMonitoringSuccess(stackId);
+        } catch (RuntimeException | CloudbreakException e) {
```
topolyai5, line 44: what happens if an `Exception` occurs? our flow will be stuck

Same all of the handlers

topolyai5 (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerRefreshParcelClusterHandler.java
topolyai5: * by force, if someone creates a new class and makes it with constructor injection, ok, but forcing to mix the methods, is not a good way

gregito: who said that I would like to make anybody to mix it? :)
in new classes, there is no question about this. use injection in a preferred way.
IMO one should use injection in one and only one way in any class. but as you know my opinion is that field injection is less preferred hence if we introduce new fields and one also prefers to use constructor injection then all the other fields should be moved into the constructor. this is not the case here since this is a completely new class.

but let me note that this comment is just a comment, not a change request :)

core/src/main/java/com/sequenceiq/cloudbreak/core/cluster/ClusterBuilderService.java
doktoric: how about prepareSdxStackCrn?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/install/ClusterManagerSetupMonitoringHandler.java
doktoric: @topolyai5 why it will be stucked?

topolyai5: Because we won't send an event and the flow will wait until the pod will not restarted

topolyai5 (approved):

bergerdenes (approved):

PR #10242, opened 2021-03-11
CB-11626 Fix integration-test project check-results failure conditions

aszegedi:
We had error at starting ApplicationContext for Integration Test project, because of 
`Cannot find class in classpath: com.sequenceiq.it.cloudbreak.testcase.authorization.listfiltering.ImageCatalogListFiltering`.
So the related `artifact/integration-test` folder contains only the:
```
integcb
suites_log
```
sub folders. Beyond this the `suites_log` folder contains only the `spring.log` file with the above exception. Even the `integration-test_test_1 exited with code 1` the result was PASSED, because of:
```
./scripts/check-results.sh
Thu Mar 11 02:22:10 UTC 2021
Mar 11, 2021 2:22:12 AM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
[0;92m+++ Postgres traffic is below  limit.
[0;92m+++ INTEGRATION TEST SUCCESSFULLY FINISHED +++
RESULT=0
```
So we need to fix the `integration-test/scripts/check-results.sh` to handle this case correctly. Besides this the `integration-test/scripts/docker-compose.sh` was improved as well with:
```    
set -o pipefail ; $INTEGCB_LOCATION/.deps/bin/docker-compose --compatibility up --remove-orphans --exit-code-from test test | tee test.out
```

### DIFF
--- a/integration-test/scripts/check-results.sh
+++ b/integration-test/scripts/check-results.sh
@@ -4,19 +4,29 @@
 
 status_code=0
 
+set -ex
+
 if [[ "$CIRCLECI" ]]; then
 
     # Check integration test results
     date
     if [[ ! -d test-output ]]; then
-        echo -e "\033[0;91m--- !!! NO test-output DIRECTORY !!! ---\n"; status_code=1;
+        echo -e "\033[0;91m--- !!! NO test-output DIRECTORY !!! ---\n";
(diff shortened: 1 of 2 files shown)


foldik (approved):
Very good!

PR #10243, opened 2021-03-12
CB-11540 Encryption supported on instance type should use string valu…

doktoric:
…e and not boolean.


doktoric:
@lnardai already approved the change https://github.com/hortonworks/cloudbreak/pull/10240 which went to master. This must go to 2.39 because of ENGESC-7353

PR #10246, opened 2021-03-12
CB-11561 - removing the usage of Inject annotation from EnvDeleteClustersFlowEventChainFactory

gregito:
Removing the usage of Inject annotation from EnvDeleteClustersFlowEventChainFactory


bergerdenes:
To be honest I like @Inject annotation more. But I don't know what is the general policy here.

gregito:
> To be honest I like @Inject annotation more. But I don't know what is the general policy here.

It is not just me but the spring guys' recommendation too to use constructor-based injection instead of field injection.
Back then at the birth of the env service, we were agreed that we should use constructor-based injection everywhere to avoid such a monster injection hells like in the case of StackCreationService.
Surely this class wouldn't go that way but why not move it to the constructor and handle the problem at its root? :slightly_smiling_face:

biharitomi (approved):
LGTM

PR #10253, opened 2021-03-12
CB-11568 We use the same boilerplate code in every freeipa creation. …

topolyai5:
…Easier if we do this in valid() method and we set the image catalog URL at the mock provider, and set null at the other providers


aszegedi (changes requested):
Could you run all the tests on this with the `aws_e2e_test` and `real_ums_test` labels that has recently been included?
There are related tests for `authorization` suite and the E2E is going to be affected also.

topolyai5:
retest pls

topolyai5:
retest this please

aszegedi (approved):

PR #10257, opened 2021-03-12
CB-11416 Replace DefaultRestartAction for flows in freeipa service

lacikaaa:
- created and set `FillInMemoryStateStoreRestartAction` for flows running salt. It will set MDCContext and make the the `Stack` pollable if not deleted/terminated
- created and set `InitializeMDCContextRestartAction` for flows which doesn't use `InMemoryStore`
- Ensured events can be recreated from DB after service restart
- Fixed salt related issue where reading the highstate returned empty result when minion didn't return. Now missing minions are in the response causing exception/failure as it should.

Tested:
- FreeIPA provision with service restart. HA tested too
- FreeIPA termination with service restart
- FreeIPA running cleanup flow with service restart
- DL creation
- DH creation, scaling, termination

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
@@ -145,7 +145,7 @@ public class SaltStates {
     }
 
     private static Multimap<String, Map<String, String>> highStateJidInfo(SaltConnector sc, String jid) {
-        Map<String, List<Map<String, Object>>> jidInfo = sc.run("jobs.lookup_jid", RUNNER, Map.class, "jid", jid);
+        Map<String, List<Map<String, Object>>> jidInfo = sc.run("jobs.lookup_jid", RUNNER, Map.class, "jid", jid, "missing", "True");
         Map<String, List<RunnerInfo>> states = JidInfoResponseTransformer.getHighStates(jidInfo);
         return collectMissingTargets(states);
     }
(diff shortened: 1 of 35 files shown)


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
```diff
@@ -145,7 +145,7 @@ public static ApplyFullResponse showState(SaltConnector sc, String state) {
     }
 
     private static Multimap<String, Map<String, String>> highStateJidInfo(SaltConnector sc, String jid) {
-        Map<String, List<Map<String, Object>>> jidInfo = sc.run("jobs.lookup_jid", RUNNER, Map.class, "jid", jid);
+        Map<String, List<Map<String, Object>>> jidInfo = sc.run("jobs.lookup_jid", RUNNER, Map.class, "jid", jid, "missing", "True");
```
jamisonbennett, line 148: I don't quite understand what the purpose of `missing` or `True` are in this context. Is there existing code which checks for these values?

lacikaaa: this flag is passed to SaltStack: https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.jobs.html#salt.runners.jobs.lookup_jid
so `jid` paramater will take the value of the `jid` variable
`missing` parameter takes the value `True`
salt will include missing nodes in the response after this is set.

jamisonbennett (approved):
It looks like you will have to rebase due to the the `toString()` method being committed in a different PR.

keyki:
Your welcome :D

PR #10258, opened 2021-03-12
CB-11576: Fix FreeIPA race condition when adding a role

jamisonbennett:
Fix FreeIPA race condition when adding a role. It was possible for the
FreeIPA management service to fail to add a host if multiple hosts were added at the same time with the same role assignment.

This was manually tested in a debugger by forcing the failure
condition to occur. Both hosts were successfully added.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
@@ -51,9 +51,18 @@ public class KerberosMgmtRoleComponent {
             FreeIpaClient ipaClient)
             throws FreeIpaClientException {
         if (roleRequest != null && StringUtils.isNotBlank(roleRequest.getRoleName())) {
-            Set<Role> allRole = ipaClient.findAllRole();
-            Optional<Role> optionalRole = allRole.stream().filter(role -> role.getCn().equals(roleRequest.getRoleName())).findFirst();
-            Role role = optionalRole.isPresent() ? optionalRole.get() : ipaClient.addRole(roleRequest.getRoleName());
+            Role role;
+            try {
+                Optional<Role> optionalRole = findRole(roleRequest.getRoleName(), ipaClient);
+                role = optionalRole.isPresent() ? optionalRole.get() : ipaClient.addRole(roleRequest.getRoleName());
+            } catch (FreeIpaClientException e) {
+                if (!FreeIpaClientExceptionUtil.isDuplicateEntryException(e)) {
+                    LOGGER.error("Failed to add the role [{}]", roleRequest.getRoleName(), e);
+                    throw e;
+                }
+                LOGGER.debug("The role [{}] was recently created in a different thread, retrieve it", roleRequest.getRoleName(), e);
+                role = findRole(roleRequest.getRoleName(), ipaClient).get();
+            }
             addPrivilegesToRole(roleRequest.getPrivileges(), ipaClient, role);
             role = ipaClient.showRole(role.getCn());
             Set<String> servicesToAssignRole = service.stream()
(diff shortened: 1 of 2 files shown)


christmasferret:
nice catch. keep up the good work! ahahahah

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
```diff
@@ -53,7 +53,18 @@ public void addRoleAndPrivileges(Optional<Service> service, Optional<Host> host,
         if (roleRequest != null && StringUtils.isNotBlank(roleRequest.getRoleName())) {
             Set<Role> allRole = ipaClient.findAllRole();
             Optional<Role> optionalRole = allRole.stream().filter(role -> role.getCn().equals(roleRequest.getRoleName())).findFirst();
-            Role role = optionalRole.isPresent() ? optionalRole.get() : ipaClient.addRole(roleRequest.getRoleName());
+            Role role;
```
akanto, line 56: logic is getting more and more complex here, we shall add a bit more test coverage, focusing also on the edge cases

jamisonbennett: I added some tests for the race condition and also for some of the prior corner cases that were previously insufficiently unit tested.

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/model/Host.java
```diff
@@ -81,4 +81,17 @@ public Boolean getHasKeytab() {
     public void setMemberOfRole(List<String> memberOfRole) {
         this.memberOfRole = memberOfRole;
     }
+
+    @Override
+    public String toString() {
+        return "Host{"
```
aakanser, line 87: You could also use `String.format("....", dn, krbprincipal,...)`.

You could have a `public static char/String` for the apostrophe instead of retyping.

aakanser: maybe even a helper method `apostropheMe(string)` that just lets you do `apostropheMe(dn)`.

aakanser (approved):
looks good to me, left one style comment

freeipa-client/src/main/java/com/sequenceiq/freeipa/client/model/Host.java
jamisonbennett: There are many other files in cloudbreak which just use the `'`. So I think I am going to leave it as is, at least for now.

jamisonbennett:
It appears that the jenkins jobs are failing to pull maven dependencies right now. I can't seem to get them to pass, so I am going to just give it some time, and hopefully, it starts working soon.

freeipa/src/main/java/com/sequenceiq/freeipa/kerberosmgmt/v1/KerberosMgmtRoleComponent.java
```diff
@@ -53,7 +53,18 @@ public void addRoleAndPrivileges(Optional<Service> service, Optional<Host> host,
...
+            try {
+                role = optionalRole.isPresent() ? optionalRole.get() : ipaClient.addRole(roleRequest.getRoleName());
+            } catch (FreeIpaClientException e) {
+                if (!FreeIpaClientExceptionUtil.isDuplicateEntryException(e)) {
+                    LOGGER.error("Failed to add the role [{}]", roleRequest.getRoleName(), e);
+                    throw e;
+                }
+                LOGGER.debug("The role [{}] was recently created in a different thread, retrieve it", roleRequest.getRoleName(), e);
+                allRole = ipaClient.findAllRole();
+                role = allRole.stream().filter(r -> r.getCn().equals(roleRequest.getRoleName())).findFirst().get();
```
lacikaaa, line 66: this looks like a duplicate of line 54-55. Could you move this to a common method?

lacikaaa (changes requested):
thanks for the change! good catch!

lacikaaa (approved):
please resolve conflicts

PR #10259, opened 2021-03-12
CB-11297. Enhance validations for diagnostics collection.

oleewere:
details:
- do not run diagnostics on stopped/requested or deleted clusters
- check dbus s3 based urls during creating support bundle

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
@@ -1,5 +1,8 @@
 package com.sequenceiq.cloudbreak.telemetry;
 
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
@@ -11,6 +14,11 @@ public class DataBusEndpointProvider {
 
     private static final String CNAME_ENDPOINT = "https://dbusapi.us-west-1.altus.cloudera.com";
 
+    private static final Map<String, String> DBUS_API_S3_ENDPOINT_PATTERN_PAIRS =
+            Map.of("dbusapi.us-west-1", "https://cloudera-dbus-prod.s3.amazonaws.com",
+                    "dbusapi.sigma-dev", "https://cloudera-dbus-dev.s3.amazonaws.com",
+                    "dbusapi.sigma-stage", "https://cloudera-dbus-stage.s3.amazonaws.com");
+
     public String getDataBusEndpoint(String endpoint, boolean useCname) {
         String databusUrl = endpoint;
         if (useCname) {
@@ -18,5 +26,15 @@ public class DataBusEndpointProvider {
             databusUrl = CNAME_ENDPOINT;
         }
         return databusUrl;
+    }
+
+    public String getDatabusS3Endpoint(String endpoint) {
+        String dbusS3Endpoint = null;
+        for (Map.Entry<String, String> entry : DBUS_API_S3_ENDPOINT_PATTERN_PAIRS.entrySet()) {
+            if (StringUtils.isNotBlank(endpoint) && endpoint.contains(entry.getKey())) {
+                return entry.getValue();
+            }
+        }
+        return dbusS3Endpoint;
     }
 }
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/diagnostics/DiagnosticsCollectionValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/diagnostics/DiagnosticsCollectionValidator.java
@@ -2,8 +2,9 @@ package com.sequenceiq.cloudbreak.controller.validation.diagnostics;
 
 import org.springframework.stereotype.Component;
 
-import com.sequenceiq.cloudbreak.common.exception.BadRequestException;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
 import com.sequenceiq.cloudbreak.telemetry.support.SupportBundleConfiguration;
+import com.sequenceiq.cloudbreak.validation.AbstractCMDiagnosticsCollectionValidator;
 import com.sequenceiq.cloudbreak.validation.ValidationResult;
 import com.sequenceiq.common.api.diagnostics.BaseCmDiagnosticsCollectionRequest;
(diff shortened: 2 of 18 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
```diff
@@ -11,6 +14,11 @@
 
     private static final String CNAME_ENDPOINT = "https://dbusapi.us-west-1.altus.cloudera.com";
```
doktoric, line 15: I think this should be a config property

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
```diff
@@ -11,6 +14,11 @@
...
 
+    private static final Map<String, String> DBUS_API_S3_ENDPOINT_PATTERN_PAIRS =
```
doktoric, line 17: I think this should be a config property

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/diagnostics/DiagnosticsCollectionValidator.java
```diff
@@ -19,16 +21,17 @@ public DiagnosticsCollectionValidator(SupportBundleConfiguration supportBundleCo
         this.supportBundleConfiguration = supportBundleConfiguration;
     }
 
-    public void validate(BaseDiagnosticsCollectionRequest request, Telemetry telemetry, String stackCrn) {
-        validate(request.getDestination(), telemetry, stackCrn, false);
+    public void validate(BaseDiagnosticsCollectionRequest request, Stack stack, Telemetry telemetry, String stackCrn) {
```
doktoric, line 24: do we really need the whole stack object?

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/diagnostics/DiagnosticsCollectionValidator.java
```diff
@@ -48,6 +51,16 @@ public void validate(DiagnosticsDestination destination, Telemetry telemetry, St
         }
     }
 
+    private void validateStackStatus(Stack stack) {
+        Status stackStatus = stack.getStatus();
+        if (Status.STOPPED.equals(stackStatus)
+                || Status.DELETED_ON_PROVIDER_SIDE.equals(stackStatus)
+                || Status.REQUESTED.equals(stackStatus)) {
+            throw new BadRequestException(String.format("Cannot run diagnostics on stack (crn: '%s') if the stack is in %s state",
```
doktoric, line 59: I think the crn is not useful for the enduser also do not call it stack but use CDP terms like Data Hub or Data Lake

datalake/src/main/java/com/sequenceiq/datalake/service/validation/diagnostics/DiagnosticsCollectionValidator.java
```diff
@@ -50,6 +52,16 @@ public void validate(DiagnosticsDestination destination, StackV4Response stackV4
         }
     }
 
+    private void validateStackStatus(StackV4Response stackV4Response) {
+        Status stackStatus = stackV4Response.getStatus();
+        if (Status.STOPPED.equals(stackStatus)
+                || Status.DELETED_ON_PROVIDER_SIDE.equals(stackStatus)
+                || Status.REQUESTED.equals(stackStatus)) {
+            throw new BadRequestException(String.format("Cannot run diagnostics on stack (crn: '%s') if the stack is in %s state",
```
doktoric, line 60: I think the crn is not useful for the enduser also do not call it stack but use CDP terms like Data Hub or Data Lake

this class seems to me a duplication.

freeipa/src/main/java/com/sequenceiq/freeipa/service/diagnostics/DiagnosticsCollectionValidator.java
```diff
@@ -45,6 +48,13 @@ public void validate(BaseDiagnosticsCollectionRequest request, Telemetry telemet
         }
     }
 
+    private void validateStackStatus(Stack stack) {
+        if (stack.getStackStatus().getStatus().isFreeIpaUnreachableStatus()) {
+            throw new BadRequestException(String.format("Cannot run diagnostics on stack (crn: '%s') if the stack is in %s state",
```
doktoric, line 53: I think the crn is not useful for the enduser also do not call it stack but use CDP terms like Data Hub or Data Lake

this class seems to me a duplication.

doktoric (changes requested):

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
oleewere: do we have any common config that can be applied for many of our services? (like a common application.yml) because it would make sense here, if there is no I can create one

doktoric: we dont have I think

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/diagnostics/DiagnosticsCollectionValidator.java
oleewere: stack is loaded before that anyway, also it's better to use stack instead of other object to not increase the number of parameters

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
oleewere: @doktoric , i have eliminated the duplications (abstract class -> common, abstract CM based class -> core-api), regarding the common configuration change, maybe that is out of the context of this change as that would be useful for other static configs as well, if you are ok with it, I would do that in ta different PR

oleewere: resolving this as CB-11658 will cover that

common/src/main/java/com/sequenceiq/cloudbreak/telemetry/DataBusEndpointProvider.java
oleewere: resolving this as CB-11658 will cover that

attilapalfi92 (approved):

PR #10260, opened 2021-03-12
CB-11621 There are cases when we remove less node than demanded node count at downscale

sodre90:
eg: customer select 3 nodes for downscale but we removed 0 nodes

It can happen because we collect nodes in the following way:

 - collect nodes without FQDN (so they are not registered to CM)
 - collect nodes from CM by some ordering algorithm if we don't have enough nodes from the first step

Problem: there are nodes with FQDN but not registered into CM. It can happen because we assign FQDN to them, but somehow we were not able to join them to the CM cluster.

Solution: we have to collect them also, so the downscale collection will happen in the following way:

 - collect nodes without FQDN
 - collect nodes with FQDN but not registered into CM
 - collect nodes from CM by some ordering algorithm if we don't have enough nodes from the 1. and 2. step


biharitomi (approved):
LGTM, but some additional unit tests would be nice to have for the affected part of SaltOrchestrator.java

PR #10263, opened 2021-03-15
DISTX-584 Move CM to its own hostgroup in DE HA Template

smaniraju:
1. Manager HostGroup to contain only CM.
2. Old manager HostGroup renamed to masterx.
3. Manager HostGroup made Gateway type.
4. The change is applied to 7.2.8, 7.2.9 and 7.2.10 DE HA Templates

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
@@ -496,7 +496,7 @@
         ]
       },
       {
-        "refName": "manager",
+        "refName": "masterx",
         "cardinality": 1,
         "roleConfigGroupsRefNames": [
           "hdfs-BALANCER-BASE",
@@ -535,6 +535,13 @@
         "cardinality": 0,
         "roleConfigGroupsRefNames": [
           "yarn-NODEMANAGER-COMPUTE"
+        ]
+      },
+      {
+        "refName": "manager",
+        "cardinality": 1,
+        "roleConfigGroupsRefNames": [
+          "knox-KNOX_GATEWAY-BASE"
         ]
       }
     ]
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/gcp/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/gcp/dataengineering-ha.json
@@ -91,6 +91,22 @@
         "nodeCount": 0,
         "type": "CORE",
         "recoveryMode": "MANUAL"
+      },
+      {
+        "name": "masterx",
+        "template": {
+          "attachedVolumes": [
+            {
+              "count": 1,
+              "size": 100,
+              "type": "pd-standard"
+            }
+          ],
+          "instanceType": "e2-standard-8"
+        },
+        "nodeCount": 1,
+        "type": "CORE",
+        "recoveryMode": "MANUAL"
       }
     ]
   }
--- a/core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
@@ -15,7 +15,7 @@
     "instanceGroups": [{
       "nodeCount": 1,
       "name": "manager",
-      "type": "CORE",
+      "type": "GATEWAY",
       "recoveryMode": "MANUAL",
       "template": {
         "aws": {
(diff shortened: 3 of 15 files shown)


core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
```diff
@@ -496,7 +496,7 @@
         ]
       },
       {
-        "refName": "manager",
+        "refName": "main",
```
sidseth, line 499: Can we use a name like 'mangerGroup02'? Will make it a little more consistent if we add additional manager service hostGroups in the future (e.g. HS2/HMS). 'main' doesn't convey a lot.

core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
```diff
@@ -103,7 +103,31 @@
       {
         "nodeCount": 2,
         "name": "master",
-        "type": "GATEWAY",
+        "type": "CORE",
```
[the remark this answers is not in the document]
cegganesh84: @sidseth One good thing about the HA blueprint is it has the Knox gateways added explicitly in the blueprint. So Knox will be HA after this change. But what will be the effect of this change is that the Raz will not be HA, which I was anticipating to be HA. Adding @doktoric. If we move to a model where Raz can be added to an instance group with at least two instances then we can achieve RAZ HA, and we can track that separately. I know there is an email thread about this. So will post it there as well.

core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
smaniraju: manager name was used because CM was being installed on this as per Ganesh. But if we see the services in this hostGroup there isn't anything to link it to manager.

core/src/main/resources/defaults/clustertemplates/7.2.10/gcp/dataengineering-ha.json
```diff
@@ -91,6 +91,22 @@
...
+        "name": "main",
+        "template": {
+          "attachedVolumes": [
+            {
+              "count": 1,
+              "size": 100,
+              "type": "pd-standard"
+            }
+          ],
+          "instanceType": "e2-standard-8"
```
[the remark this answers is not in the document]
smaniraju: I will create separate jira to update all GCP instances to 16x64GB

core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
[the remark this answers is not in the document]
smaniraju: 7.2.8 was the only template which had to be changed for this, all others templates master was CORE only. This was the approach discussed on call with you and ganesh.  Maybe we can increase nodeCount of the new manager hostgrup with only CM to 2, so that RAZ is also HA ?  cc @cegganesh84 , @sidseth

core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
sidseth: Sorry. I meant masterGroup02. Yes, this is not a 'manager' (CM), and 'main' is not very helpful.

core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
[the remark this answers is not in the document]
sidseth: Ok. There's also https://github.com/hortonworks/cloudbreak/pull/10266/ .
End of the day - I think the 7.2.8, 7.2.9, and 7.2.10 templates need to reach the same state.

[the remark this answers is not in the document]
sidseth: Commit after renaming the group, and then a separate change to rationalize the 3 versions?

core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
smaniraju: Since the hostgroup name is also part of endpoint, the hostgroup name cannot be very big. Some services like hue require max hostgroup name of 7 characters. Also endpoint generated based on masterGroup02 would be something like https://<cluster>mastergroup020, https://<cluster>mastergroup021..... since number index is used to refer node. So dont think hostGroup name should contain numbers to avoid confusion between hostgroupname and nodeIndex.  I tried various other 6-7 letter words but main seemed better than most other options.  If you any other word in mind without numerals and length of upto 6 to 7 characters  I can change to that.

sidseth: I don't have a good name. 'masterx'? That's what LinkedIn ended up using. 'main' would confuse me ... Have master01, master02, and main01 vs mastere01, master02, masterx01 | The second at least says that these are the same category of nodes, running master services.

cegganesh84: Auxiliary is the name used by datalake HA. I am fine with masterx, and vote for it. Using the same name as datalake HA, could confuse some folks coming from a different perspective if the intentions of the name are slightly different.

smaniraju: ok will change this hostgroup name to masterx.

core/src/main/resources/defaults/clustertemplates/7.2.8/aws/dataengineering-ha.json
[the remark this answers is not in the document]
cegganesh84: Raz HA is taken care of by #10290

cegganesh84 (approved):
LGTM

sidseth (approved):
Looks good.

PR #10264, opened 2021-03-15
CB-9717 FreeIPA should log X-CDP-Request-ID to /var/log/httpd/error_log

christmasferret:
As described in https://docs.google.com/document/d/1krz-82-wY9PwUcJcof2SH9EtF7KkYcPVmWGIzN2akkA/edit#, the header X-CDP-Request-ID should be logged by FreeIPA into /var/log/httpd/error_log

This can be done by adding logging of environ.get('HTTP_X_CDP_Request_ID') to the info level logs statemetns in /usr/lib/python2.7/site-packages/ipaserver/rpcserver.py
WSGIExecutioner::wsgi_execute(). Salt to distribute update this file and restart ipa.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/common-install.sls
@@ -17,6 +17,15 @@ one_week_next_update_grace_period:
       - service pki-tomcatd@pki-tomcat start
     - unless: grep "^ca[.]crl[.]MasterCRL[.]nextUpdateGracePeriod=10080$" /var/lib/pki/pki-tomcat/ca/conf/CS.cfg
 
+add-httpd-x-cdp-trace-id:
+   file.line:
+     - name: /usr/lib/python2.7/site-packages/ipaserver/rpcserver.py
+     - mode: ensure
+     - content: "        logger.info('X-cdp-request-ID : %s', environ.get('HTTP_X_CDP_REQUEST_ID'))"
+     - after: "return self.marshal(result, RefererError(referer=environ['HTTP_REFERER']), _id)"
+     - indent: False
+     - backup: False
+
 /usr/lib/python2.7/site-packages/ipaserver/plugins/getkeytab.py:
   file.managed:
     - makedirs: True
@@ -32,6 +41,7 @@ restart_freeipa_after_plugin_change:
     - onlyif: test -f /etc/ipa/default.conf
     - watch:
       - file: /usr/lib/python2.7/site-packages/ipaserver/plugins/getkeytab.py
+      - file: /usr/lib/python2.7/site-packages/ipaserver/rpcserver.py
 
 set_number_of_krb5kdc_workers:
   file.replace:


christmasferret:
this is the first part of tracing id work which adds X-cdp-request-ID in freeipa httpd logging. 

I only added one line to the rpcserver.py
```
logger.info('X-cdp-request-ID : %s', environ.get('HTTP_X_CDP_REQUEST_ID'))
```
trace

lacikaaa:
I will leave this review mostly to you @jamisonbennett as I'm not really familiar with python code.
Only 2 notes:
- this is a huge file. Would it be possible to split into multiple?
- as this contains a bunch of logic at first glance, shall we cover this with tests?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/rpcserver.py
```diff
@@ -0,0 +1,1258 @@
...
+        args = ()
+        options = {}
+        command = None
+
+        e = None
+        if not 'HTTP_REFERER' in environ:
+            return self.marshal(result, RefererError(referer='missing'), _id)
+        if not environ['HTTP_REFERER'].startswith('https://%s/ipa' % self.api.env.host) and not self.env.in_tree:
+            return self.marshal(result, RefererError(referer=environ['HTTP_REFERER']), _id)
+        logger.info('X-cdp-request-ID : %s', environ.get('HTTP_X_CDP_REQUEST_ID'))
```
jamisonbennett, line 349: Due to licensing reasons, I think we may need to use salt to add the new lines to this file rather than load the entire file.

jamisonbennett (changes requested):

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/rpcserver.py
christmasferret: @jamisonbennett @lacikaaa  Instead of a file drop, now I am doing a line write. It will only write one line to the python file to log the trace-id. And I run salt multiple times, it will only write once if the line does not exist. 

Re tested everything and works.

lacikaaa:
just 2 questions:
- have you tested when there is no `HTTP_X_CDP_Request_ID` in the header?
- if I see it from your attached screenshot we will be able to correlate the logs by pid and not every log line will have this info. Am I right? Is it possible to add this to every related lines or it would take too much effort and doesn't worth it?

christmasferret:
@lacikaaa 

1. I tested with no `HTTP_X_CDP_Request_ID` in the header, it will show as `X-cdp-request-ID: None`. Nothing breaks. 

2. Yes. not all log line will have this info. We are logging this line whenever there is a request entering the wsgi server. So it is one log per one request. I think it is not possible to log all related lines as there are too many logging lines in the python code base and it doesn't worth it. The main purpose was just to correlate the request between freeipa and cluster-proxy to see if the request has landed on freeipa. The freeipa internal processing logic would be something different from this purpose and not an urgent need compare to what we want.

lacikaaa (approved):
thanks for the clarifications! LGTM
my only issue is we might won't notice if it breaks like if there is a slightly different.
One more thought about this. Is this file exist when the freeipa server package installed? Or only after the freeipa server installation process has run? I mean we might add this change to the freeipa image itself

jamisonbennett (approved):

christmasferret:
I think we can add this change to the freeipa image itself. But I am unfamiliar with that process and may take some time to go that route. Since the change here use `ensure`, so even later we add to freeipa image, the change here will not break. It will just skip and not duplicate the change. 

I will merge this one so we can start to use this x-cdp-request-id.

PR #10265, opened 2021-03-16
DISTX-557 Overcome master memory overcommit problem with default DE d…

cegganesh84:
…efinition by using m5.4xlarge

1. There are 33 roles in master, and 32 GB is clearly not enough.
2. This will solve the memory overcommit problem.
3. This will give more memory for services like HS2.

Tested that memory overcommit is gone by creating a cluster with custom definition like this.

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering.json
@@ -24,7 +24,7 @@
               "type": "NONE"
             }
           },
-          "instanceType": "m5.2xlarge"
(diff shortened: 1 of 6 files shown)


sidseth (commented):
Change for Azure as well?
Other than that, +1.

doktoric:
retest this please

cegganesh84:
For now updated Aws and Azure. It follows the same instance types as HA template. Whether to move to memory-optimized instances and update GCP also can be done in the followups.

sidseth (approved):

cegganesh84:
@sidseth Could you please merge this?

PR #10266, opened 2021-03-16
DISTX-546 Enable load-balancer for 7.2.9 and 7.2.10 by default

cegganesh84:
1. 7.2.8 is full tested in dev with both public and private subnets individually.
2. Promoting this change for all the future releases.
3. The template is still in preview state.

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering-ha.json
@@ -11,6 +11,7 @@
     "externalDatabase": {
       "availabilityType": "HA"
     },
+    "enableLoadBalancer": true,
     "instanceGroups": [{
       "nodeCount": 1,
       "name": "manager",
--- a/core/src/main/resources/defaults/clustertemplates/7.2.9/aws/dataengineering-ha.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.9/aws/dataengineering-ha.json
@@ -11,6 +11,7 @@
     "externalDatabase": {
       "availabilityType": "HA"
     },
+    "enableLoadBalancer": true,
     "instanceGroups": [{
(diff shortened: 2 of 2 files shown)


core/src/main/resources/defaults/clustertemplates/7.2.9/aws/dataengineering-ha.json
```diff
@@ -11,6 +11,7 @@
     "externalDatabase": {
       "availabilityType": "HA"
     },
+    "enableLoadBalancer": true,
```
sidseth, line 14: Along with this, does the manager need to be changed to type CORE, and the master to GATEWAY. That is at least what 7.2.8 does. Unless that change is expected in the change @smaniraju is currently making.

cegganesh84: Yeah, I am relying on @smaniraju's change.

cegganesh84: Yes, particularly I did not want to introduce merge conflicts for @smaniraju

core/src/main/resources/defaults/clustertemplates/7.2.10/aws/dataengineering-ha.json
```diff
@@ -11,6 +11,7 @@
     "externalDatabase": {
       "availabilityType": "HA"
     },
+    "enableLoadBalancer": true,
```
doktoric, line 14: I thin kthis feature is under entitlement

cegganesh84: No Richard. This is like DL HA LB. No entitlement. Unline DL HA, DE HA template is still in "PREVIEW" state.

doktoric (approved):

PR #10267, opened 2021-03-16
CB-11415 Replace DefaultRestartAction for flows in redbeams service

gergopapi2:
Upon restart the DBStack of running flows is not loaded into the in memory store, causing possible problems in the heartbeatService when all deleting flows are queried.
A restart action was added that fills the in memory store with currently executing flow's DBStack, if present (at the final step of the delete flow DBStack is deleted, thus after a restart at that place there is be no more DBStack).


keyki:
retest this please

PR #10268, opened 2021-03-16
CB-11442 Sdx upgrade: even if image change fails, Sdx flow continues

horadla23:
- if image change fails then we should set stack status to UPDATE_FAILED instead of AVAILABLE
- in this case, SDX syncer won't update SDX status from Upgrade failed to Available
- also change image failure handling should not roll back image information in Component table, since CM is already upgraded
- SDX change image wait logging fixes


foldik (approved):
lgtm

pdarvasi (approved):

tiborpopovics (approved):

PR #10269, opened 2021-03-16
CB-9504 Replace FreeIPA error page

daszabo:

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/patch-pki-tomcat.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/patch-pki-tomcat.sls
@@ -1,0 +1,25 @@
+restart_pki_tomcat:
+  service.running:
+      - name: pki-tomcatd@pki-tomcat.service
+      - watch:
+        - file: /var/lib/pki/pki-tomcat/conf/web.xml
+
+replace_default_tomcat_error_page:
+  file.replace:
+    - name: /var/lib/pki/pki-tomcat/conf/web.xml
+    - pattern: '</welcome-file-list>\s*</web-app>'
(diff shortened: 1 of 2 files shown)


pdarvasi (approved):

freeipa/src/main/resources/freeipa-salt/salt/freeipa/patch-error-page.sls
```diff
@@ -0,0 +1,24 @@
+restart_ipa:
+  service.running:
+      - name: ipa
```
lacikaaa, line 3: I think restarting `pki-tomcatd@pki-tomcat.service` should be enough

freeipa/src/main/resources/freeipa-salt/salt/freeipa/patch-error-page.sls
```diff
@@ -0,0 +1,24 @@
...
+      - watch:
+        - file: /var/lib/pki/pki-tomcat/conf/web.xml
+
+replace_default_tomcat_error_page:
+  file.replace:
+    - name: /var/lib/pki/pki-tomcat/conf/web.xml
+    - pattern: '</web-app>'
+    - repl: "<error-page>
+             <error-code>404</error-code>
+                <location>/error.jsp</location>
```
lacikaaa, line 13: is this page exists?

lacikaaa (changes requested):
replicas are missing this fix

pdarvasi:
@daszabo could you finish the testing or would you need some help? Thanks

daszabo:
Tested with FreeIPA HA, it works properly. Also made sure multiple runs yields the same result.

lacikaaa (approved):

PR #10270, opened 2021-03-16
CB-11379 Extend RightV4 with additional rights

agov:
Added DH and SDX describe along with DH upgrade rights to be used on UI


horadla23 (approved):

foldik (approved):
lgtm

PR #10203, opened 2021-03-08
CB-11419 Fix waiting edge cases in integration tests

Bajzathd:
While WaitService waits for a certain criteria to be met (depedning on StatusChecker implementation), it should refresh its data only once per round. This way we can not encounter an edge case where a check runs on different data then the check before it, causing changes to not handled properly.

See detailed description in the commit message.

PR #10271, opened 2021-03-16
CB-11419 Fix waiting edge cases in integration tests

Bajzathd:
While WaitService waits for a certain criteria to be met (depedning on StatusChecker implementation), it should refresh its data only once per round. This way we can not encounter an edge case where a check runs on different data then the check before it, causing changes to not handled properly.


Bajzathd:
Duplicate of #10203

Bajzathd:
Duplicate of #10203

Bajzathd:
Duplicate of #10203

PR #10278, opened 2021-03-16
CB-11542 Cross-account policy has s3:HeadBucket, but HeadBucket doesn…

roohisyeda:
…'t have a corresponding IAM action

### DIFF
--- a/cloud-aws/src/main/resources/definitions/aws-cb-policy.json
+++ b/cloud-aws/src/main/resources/definitions/aws-cb-policy.json
@@ -133,7 +133,6 @@
         "s3:CreateBucket",
         "s3:DeleteObject",
         "s3:GetObject",
-        "s3:HeadBucket",
         "s3:ListBucket",
         "s3:GetBucketLocation",
         "s3:PutObject",


odraese (approved):
lgtm
The description of the PR doesn't contain a section about how this was tested, so I assume that you ran an Env-Activation after that policy change...

roohisyeda:
> 
Screen Shot 2021-03-16 at 6 07 43 PM
Like mentioned on the JIRA there is no corresponding `action` for HeadBucket in s3.
Also when testing reduced permissions mode, I was able to activate successfully w/o HeadBucket.
There are the permissions needs for elevated permissions mode (normal activation) 
https://docs.cloudera.com/data-warehouse/cloud/aws-environments/topics/dw-aws-standard-iam-perms-required.html
and reduced permissions mode https://docs.cloudera.com/data-warehouse/cloud/aws-environments/topics/dw-aws-reduced-perms-mode-min-iam-perms-req.html
and if you see there is no HeadBucket listed there

doktoric (approved):

PR #10279, opened 2021-03-16
DISTX-397 Use the default setting for HS2 transport mode which is all…

cegganesh84:
… (both HTTP and binary)

1. OPSAPS-54505 made all as the default transport mode for 7.2.7+ clusters.
2. This is needed if we want to enable load-balancer from Hue to HS2 integration.
3. Tested this in a cluster by reverting the setting to default in CM in 7.2.8 cluster, and made sure both beeline and Hue worked.


cegganesh84:
Hue will support HS2 failover using ZK. So closing this PR for now.

PR #10280, opened 2021-03-17
CB-11635 checkright and hasrights should throw an error if provided c…

horadla23:
…rn is not valid

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -1,11 +1,13 @@
 package com.sequenceiq.cloudbreak.auth.altus;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -460,6 +465,9 @@ private boolean makeCheckRightCall(String actorCrn, String userCrn, String right
     public List<Boolean> hasRightsNoCache(String actorCrn, String memberCrn, List<RightCheck> rightChecks, Optional<String> requestId) {
         LOGGER.info("Checking whether member [{}] has rights [{}]", memberCrn,
                 rightChecks.stream().map(this::rightCheckToString).collect(Collectors.toList()));
+        checkArgument(rightChecks.stream().map(RightCheck::getResource).allMatch(Crn::isCrn),
```
foldik, line 468: The resource can be `null` or '*' as well, which case UMS does an account level check, so I would use the condition `c -> c == null || "*".equals(c) || Crn.isCrn(c)`.

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -497,6 +505,9 @@ public String rightCheckToString(RightCheck rightCheck) {
     }
 
     public List<Boolean> hasRightsOnResources(String actorCrn, String memberCrn, List<String> resourceCrns, String right, Optional<String> requestId) {
+        checkArgument(resourceCrns.stream().allMatch(Crn::isCrn),
```
foldik, line 508: The same as above.

foldik (changes requested):

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: null is not allowed on the interface, so it does not work anyway, i am suprised that '*' is allowed, since it is not a check for a specific right, but i'll fix this, but I think we will never send null or '*' on this interface

horadla23: fixed

foldik (approved):
LGTM

PR #10281, opened 2021-03-17
DISTX-571:set hive.server2.idle.session.timeout to 4h

satyabollineni:
The setting "hiveserver2_idle_session_timeout" used to be 1 day by default for DE clusters and now the default value is changed to 4 hours and the screenshots attached for the same:
 
before change:
Screenshot 2021-03-17 at 4 47 27 PM

Post the change:
Screenshot 2021-03-17 at 4 03 23 PM
Screenshot 2021-03-17 at 4 47 27 PM


doktoric (approved):
if it was tested than I am fine with it

cegganesh84 (approved):
LGTM

PR #10282, opened 2021-03-17
CB-11654 - Azure: Deployment failed, but the flow in CB continued (No…

pdarvasi:
… VM resource found amongst cloud resources)

In case the ARM deployment failed, we still tried to collect all the resources, for no reason.
Hence this commit introduces and extra check to only do this on non-failing deployments.


keyki (changes requested):
Looks good. Doesn't work.

keyki (approved):

PR #10285, opened 2021-03-17
CFM-1834 Streamline instance types for heavy-duty DataFlow templates.…

vbrodetskyi:
…Updated instance type and volume for ZK nodes to make Heavy Duty cluster cheaper.

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.9/gcp/flow-management.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.9/gcp/flow-management.json
@@ -55,7 +55,7 @@
         "template": {
           "instanceType": "e2-standard-4",
           "rootVolume": {
-            "size": 50
+            "size": 100
           },
           "attachedVolumes": [
             {
--- a/core/src/main/resources/defaults/clustertemplates/7.2.9/gcp/flow-management.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.9/gcp/flow-management.json
@@ -53,9 +53,9 @@
         "type": "CORE",
         "recoveryMode": "MANUAL",
         "template": {
-          "instanceType": "e2-standard-8",
+          "instanceType": "e2-standard-4",
           "rootVolume": {
-            "size": 150
+            "size": 50
           },
           "attachedVolumes": [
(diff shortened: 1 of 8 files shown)


core/src/main/resources/defaults/clustertemplates/7.2.9/gcp/flow-management.json
```diff
@@ -53,9 +53,9 @@
         "type": "CORE",
         "recoveryMode": "MANUAL",
         "template": {
-          "instanceType": "e2-standard-8",
+          "instanceType": "e2-standard-4",
           "rootVolume": {
-            "size": 150
+            "size": 50
```
doktoric, line 58: why it was modified?

vbrodetskyi: @doktoric according to discussion in jira, starting from here https://jira.cloudera.com/browse/CFM-1834?focusedCommentId=4378566&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4378566

doktoric: @vbrodetskyi I think this is not a good idea because we would like to support upgrades and the root disk size in this case is very small cc @keyki

vbrodetskyi: @doktoric got it, thank you for information, what is the minimal size we can set 100GB or 150GB?

keyki: We went with 50 GB as the default for a long time and recently we increased it to 100 to all data hub types. Upgrade requires lots of space since we need to download multiple parcels that are large. For a period of time, it also copied to multiple locations on the root disk so we need double the parcel size space. Setting it to 100 would align with other data hubs.

vbrodetskyi: @keyki thank you for clarification.

doktoric (approved):

PR #10287, opened 2021-03-17
CB-11667 Use ListBucket action instead of HeadBucket action in AWS IAM policy

joshelser:
s3:HeadBucket is not a valid IAM action. s3:ListBucket is the IAM action
which grants access to call the HeadBucket API on S3.


doktoric:
thanks @joshelser

doktoric (approved):

PR #10290, opened 2021-03-17
CB-9217 Ability to deploy multiple Raz instances on DH cluster. We ad…

doktoric:
…ding RAZ to hostgroups which has ZK master if the cluster contains at least 2 ZK server and also adding to the GW group.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazDatahubConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/raz/RangerRazDatahubConfigProvider.java
@@ -1,25 +1,28 @@
 package com.sequenceiq.cloudbreak.cmtemplate.configproviders.raz;
 
-import com.cloudera.api.swagger.model.ApiClusterTemplateService;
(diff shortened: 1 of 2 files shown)


cegganesh84 (approved):
I like the check for the ZK group for deciding on RAZ HA.

sidseth (approved):
Looks good. Not sure the >=3 ZK check is really required (The alternate is to say - install RAZ if ZK is present, and RAZ potentially has at least 2 instances on all clusters - or at least on clusters where CM runs on a separate node)

doktoric:
> Looks good. Not sure the >=3 ZK check is really required (The alternate is to say - install RAZ if ZK is present, and RAZ potentially has at least 2 instances on all clusters - or at least on clusters where CM runs on a separate node)

@sidseth should we then modify the jira comment ? 2 is enough ?

sidseth:
> > Looks good. Not sure the >=3 ZK check is really required (The alternate is to say - install RAZ if ZK is present, and RAZ potentially has at least 2 instances on all clusters - or at least on clusters where CM runs on a separate node)
> 
> @sidseth should we then modify the jira comment ? 2 is enough ?

I think this works reasonably well also, so continue to be +1. RAZ goes HA on shapes which are likely to be HA (multiple ZKs)

sidseth (approved):

PR #10292, opened 2021-03-18
CB-11672 GCP: Mark the existing 7.2.8 datahub templates as GA

cegganesh84:
1. Mark all the templates except for DE HA, DM and RTDM as released from 7.2.8
2. If 7.2.8 gets released before GCP is GA, customers with GCP entitlement will see this version as released before the announcement.
3. But marking it as released before, knocks out one item to track from the launch checklist.


doktoric (approved):

doktoric:
retest this please

PR #10294, opened 2021-03-18
CB-11655 Not allow to delete instances from data lake clusters.

foldik:


horadla23 (commented):
a simple unit test would be good for this

horadla23 (approved):

PR #10297, opened 2021-03-18
CB-11682 Remove internal actor from integration-test

horadla23:
- removing explicit internal actor usage from integration tests
- only using internal crn client for specific actions


horadla23:
retest this please

aszegedi (approved):

PR #10301, opened 2021-03-18
CB-11694 - Added a new Knox topology called cdp-proxy-token

smolnar82:
This new topology must be in synch with `cdp-proxy-api` in terms of services and gateway providers except for the authentication/federation provider they use.


keyki:
retest this please

keyki (approved):

PR #10303, opened 2021-03-19
CDPCP-3702 Add flag to force update of workload credentials by user sync

gitmfox:
Adds a flag to the SynchronizeAllUsers request that forces user sync to update
workload credentials for all users, whether or not they have changed. The flag
is a no-op when the CDP_USER_SYNC_CREDENTIALS_UPDATE_OPTIMIZATION entitlement
is not enabled.

The flag will be exposed as a '--force-workload-credentials-update' option to
the 'cdp environments sync-all-users' API. This is the backend implementation.

I added the flag to both public entry points to UserSyncService -- synchronizeUsers
and synchronizeUsersWithCustomPermissionCheck, even though callers of the former
always set it to false. It seemed inconsistent for this option to be accepted
when doing a custom permission check, but not otherwise.

I also moved the computation of UserSyncOptions up the call chain, to avoid
adding more parameters to lower-level methods, and to simplify adding new
options in the future.

### DIFF
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/model/SynchronizeAllUsersRequest.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/model/SynchronizeAllUsersRequest.java
@@ -24,6 +24,9 @@ public class SynchronizeAllUsersRequest extends SynchronizeOperationRequestBase 
     @ApiModelProperty(value = UserModelDescriptions.USERSYNC_ACCOUNT_ID)
     private String accountId;
 
+    @ApiModelProperty(value = UserModelDescriptions.WORKLOAD_CREDENTIALS_UPDATE_TYPE)
+    private WorkloadCredentialsUpdateType workloadCredentialsUpdateType = WorkloadCredentialsUpdateType.UPDATE_IF_CHANGED;
+
     public SynchronizeAllUsersRequest() {
     }
 
@@ -56,6 +59,14 @@ public class SynchronizeAllUsersRequest extends SynchronizeOperationRequestBase 
         this.accountId = accountId;
     }
 
+    public WorkloadCredentialsUpdateType getWorkloadCredentialsUpdateType() {
+        return workloadCredentialsUpdateType;
+    }
+
+    public void setWorkloadCredentialsUpdateType(WorkloadCredentialsUpdateType workloadCredentialsUpdateType) {
+        this.workloadCredentialsUpdateType = workloadCredentialsUpdateType;
+    }
+
     public Set<String> getDeletedWorkloadUsers() {
         return deletedWorkloadUsers;
     }
@@ -71,6 +82,7 @@ public class SynchronizeAllUsersRequest extends SynchronizeOperationRequestBase 
                 + ", users=" + users
                 + ", deletedWorkloadUsers=" + deletedWorkloadUsers
                 + ", accountId=" + accountId
+                + ", workloadCredentialsUpdateType=" + workloadCredentialsUpdateType
                 + ", " + super.fieldsToString()
                 + '}';
     }
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
@@ -35,7 +35,6 @@ import com.sequenceiq.authorization.resource.AuthorizationResourceAction;
 import com.sequenceiq.authorization.service.CommonPermissionCheckingUtils;
 import com.sequenceiq.authorization.service.CustomCheckUtil;
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
-import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
 import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
 import com.sequenceiq.cloudbreak.common.json.Json;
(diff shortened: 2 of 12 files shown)


handavid (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncService.java
```diff
@@ -154,21 +153,31 @@
...
-        return performSyncForStacks(accountId, actorCrn, userSyncFilter, stacks);
+        UserSyncOptions options = getUserSyncOptions(accountId, userSyncFilter.isFullSync(), forceWorkloadCredentialsUpdate);
+        return performSyncForStacks(accountId, actorCrn, userSyncFilter, options, stacks);
     }
 
-    private Operation performSyncForStacks(String accountId, String actorCrn, UserSyncRequestFilter userSyncFilter, List<Stack> stacks) {
+    private UserSyncOptions getUserSyncOptions(String accountId, boolean fullSync, boolean forceWorkloadCredentialsUpdate) {
+        boolean credentialsUpdateOptimizationEnabled = !forceWorkloadCredentialsUpdate &&
+                entitlementService.usersyncCredentialsUpdateOptimizationEnabled(accountId);
+        LOGGER.info("Credentials update optimization is{} enabled for account {}", credentialsUpdateOptimizationEnabled ? "" : " not", accountId);
```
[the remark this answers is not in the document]
gitmfox: Good point. I reworded the message.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/model/SynchronizeAllUsersRequest.java
```diff
@@ -24,6 +24,9 @@
     @ApiModelProperty(value = UserModelDescriptions.USERSYNC_ACCOUNT_ID)
     private String accountId;
 
+    @ApiModelProperty(value = UserModelDescriptions.FORCE_WORKLOAD_CREDENTIALS_UPDATE)
+    private Boolean forceWorkloadCredentialsUpdate = Boolean.FALSE;
```
lacikaaa, line 28: `Boolean` is not too flexible. Shouldn't we use some enum here so we could add other options if necessary?

lacikaaa (changes requested):

horadla23 (commented):
Based on the jira, I cannot see what is the real use case here from customer point of view. Jira mention only `in case something goes wrong`.
I also not sure if the flag `--force-workload-credentials-update` will make sense for customer, since in case of entitlement is off, then we update every user, in case of entitlement is on, we update only affected users, so we update workload credentials anyway, in this case if I were the customer, I wouldn't know what "force workload credentials update" means. I would introduce this change only if we see any concrete use case which makes this change necessary.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/model/SynchronizeAllUsersRequest.java
gitmfox: This seems like a simple on/off switch to me. I'm trying to think of other possible options, but not coming up with anything. Do you have some ideas of what the options might look like?

gitmfox:
> Based on the jira, I cannot see what is the real use case here from customer point of view. Jira mention only `in case something goes wrong`.
> I also not sure if the flag `--force-workload-credentials-update` will make sense for customer, since in case of entitlement is off, then we update every user, in case of entitlement is on, we update only affected users, so we update users who has workload credentials changed anyway, in this case if I were the customer, I wouldn't know what "force workload credentials update" means. I would introduce this change only if we see any concrete use case which makes this change necessary.

Sorry, I should have mentioned that this will be added as a hidden option to the 'environments sync-all-users' CLI/API, not visible to customers -- at least initially, and probably forever, unless there is a use case for making it publicly visible.

I agree that it doesn't make sense to users; I was thinking of it as something that support/services could have in their toolkit. To make "something goes wrong" less vague: the problem could be a bug, or some corruption of the IPA user record. It might also be useful, when debugging an authentication problem on the workload side, to be able to rule out the credentials update optimization as a factor. In general, when you have a sync operation, it seems to me that it's a good idea to have a way to do a 'force push.'

This flag would only be relevant when the entitlement is granted.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/user/model/SynchronizeAllUsersRequest.java
lacikaaa: force all, force only machine user, force only non machine users, force users with ssh-key. my point is that we saw similar cases that a boolean seems the right choice, as it can only be on/off and a few month later there came a third option. it gives us more options in the future as the API must be backward compatible.

gitmfox: > force all, force only machine user, force only non machine users, force users with ssh-key. my point is that we saw similar cases that a boolean seems the right choice, as it can only be on/off and a few month later there came a third option. it gives us more options in the future as the API must be backward compatible.

That makes sense. I will change it to be an enum.

gitmfox: Done.

gitmfox:
retest this please

lacikaaa (approved):

gitmfox:
retest this please

gitmfox:
retest this please

gitmfox:
@lacikaaa or @handavid, now that all of the checks have finally passed, would one of you mind merging this PR? Thanks.

PR #10304, opened 2021-03-19
CB-11700. Zookeeper should not be on the 'worker' nodes in the Spark3

sidseth:
template.
- Moves Zookeeper over to the 'master' node.


cegganesh84 (approved):
Good catch. I have seen worker scale-up fails in this scenario.

keyki (approved):

keyki:
retest this please

PR #10310, opened 2021-03-19
CB-11675 - Added Knox's own KNOXTOKEN service and the new tokengen application into the cdp-proxy topology

smolnar82:
Two new artifacts have to be added in cdp-proxy topology:
- Knox's own KNOXTOKEN service (it's proprietary to Knox, not like a regular service such Hive, YARN, ATLAS,...) with a custom configuration that is not meant to be changed from the UI
- A new Knox application called `tokengen`. It's 100% guaranteed that all versions after CDH 7.2.9 will have it

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -811,4 +811,32 @@
         {%- endif %}
     {%- endif %}
 
+    <service>
+      <role>KNOXTOKEN</role>
+      <param>
+         <name>knox.token.ttl</name>
+         <value>31540000000</value> <!-- 1 year -->
+      </param>
+      <param>
+         <name>knox.token.exp.server-managed</name>
+         <value>true</value>
+      </param>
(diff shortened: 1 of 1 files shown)


lajosrodek:
@smolnar82  I wonder what is the case with pre-7.2.9 runtimes. We still support launching DLs with earlier runtimes like 7.1.0, 7.2.0 etc, which would all use the same Knox topology definition.

smolnar82:
> @smolnar82 I wonder what is the case with pre-7.2.9 runtimes. We still support launching DLs with earlier runtimes like 7.1.0, 7.2.0 etc, which would all use the same Knox topology definition.

Yes, this is a valid concern. I'm sure that adding the `KNOXTOKEN` service should not cause an issue as it's a known KNOX service since the beginning.
I'm going to check the `tokengen` application. If that's a problem, then @keyki will prepare the `version based filtering` in the Jinja template next week.

smolnar82:
So we definitely need a version based checking because an unknown application breaks topology deployment.

keyki:
Let's continue here: https://github.com/hortonworks/cloudbreak/pull/10330

PR #10312, opened 2021-03-19
CB-11691 change the Datanode address and https address to a secure po…

keyki:
…rt >=1024

By default when Kerberos is enabled the following ports are set by CM:
DataNode Transceiver Port              dfs.datanode.address         1004
Secure DataNode Web UI Port (TLS/SSL)  dfs.datanode.https.address   9865

In this case CM will show a warning that HDFS has invalid configurations:
Secure DataNode configuration is invalid. There are two recommended configurations:
(1) DataNode Transceiver Port and Secure DataNode Web UI Port (TLS/SSL) both >= 1024,
DataNode Data Transfer Protection set, Hadoop TLS/SSL enabled;
(2) DataNode Transceiver Port and DataNode HTTP Web UI Port both < 1024,
DataNode Data Transfer Protection not set, Hadoop TLS/SSL disabled.

To fix this warning we can se these 2 parameters in the instantiator section for Kerberos.
This config is valid with all the runtime version we have. Tested from 7.1.0+.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
@@ -60,6 +60,7 @@ import com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn.YarnConstants;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn.YarnRoles;
 import com.sequenceiq.cloudbreak.common.json.JsonUtil;
 import com.sequenceiq.cloudbreak.common.type.ClusterManagerType;
+import com.sequenceiq.cloudbreak.dto.KerberosConfig;
 import com.sequenceiq.cloudbreak.template.BlueprintProcessingException;
 import com.sequenceiq.cloudbreak.template.TemplatePreparationObject;
 import com.sequenceiq.cloudbreak.template.model.ServiceAttributes;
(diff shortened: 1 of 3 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -78,6 +78,10 @@
 
     private static final String ROLE_TYPE = "roleType";
 
+    private static final BigDecimal SECURE_DATANODE_ADDRESS = BigDecimal.valueOf(9866L);
```
doktoric, line 81: should we add these ports to the deploymentconfig?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -446,7 +450,10 @@ private void addCmVersionDependantConfigs(ClouderaManagerRepo cmRepoDetails, Tem
         }
         if (Objects.nonNull(cmRepoDetails) && CMRepositoryVersionUtil.isEnableKerberosSupportedViaBlueprint(cmRepoDetails)
                 && templatePreparationObject.getKerberosConfig().isPresent()) {
-            instantiator.setEnableKerberos(new ApiConfigureForKerberosArguments());
+            instantiator.setEnableKerberos(new ApiConfigureForKerberosArguments()
+                    .datanodeTransceiverPort(SECURE_DATANODE_ADDRESS)
```
doktoric, line 454: do we need to check any cm version?

keyki: No, this exists since the very first version. I tested with 7.1.0

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
keyki: What do you mean?

keyki: Done

doktoric (approved):

PR #10313, opened 2021-03-19
CFM-1850 Datahub cluster creation issue due to Nifi Registry client p…

vbrodetskyi:
…roperties on CDH 7.2.9.0 stack. Removed client property which is not actual anymore and added new client properties to NiFi config, to create automatically NiFi Registry client in NiFi.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
@@ -137,8 +137,12 @@
                 "value": "-Xms8g"
               },
               {
-                "name": "add.internal.nifi.registry.client",
-                "value": "true"
+                "name": "nifi.registry.client.name",
+                "value": "Default NiFi Registry Client"
+              },
+              {
+                "name": "nifi.registry.client.url",
+                "value": "https://placeholder:18433"
               },
               {
                 "name": "rlimit_fds",
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management.bp
@@ -137,8 +137,12 @@
                 "value": "-Xms8g"
               },
               {
-                "name": "add.internal.nifi.registry.client",
-                "value": "true"
+                "name": "nifi.registry.client.name",
+                "value": "Default NiFi Registry Client"
+              },
+              {
+                "name": "nifi.registry.client.url",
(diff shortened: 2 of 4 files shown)


core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
```diff
@@ -137,8 +137,12 @@
...
               },
               {
-                "name": "add.internal.nifi.registry.client",
-                "value": "true"
+                "name": "nifi.registry.client.name",
+                "value": "Default NiFi Registry Client"
+              },
+              {
+                "name": "nifi.registry.client.url",
+                "value": "https://placeholder:18433"
```
lajosrodek, line 145: Out of curiosity, what is the purpose of this fake URL? Will it be simply ignored or later updated (by CM etc.) to some valid value?

vbrodetskyi: @lajosrodek yes, after NiFi will be started, this value will be updated in createRequiredNiFiObjects script (in CSD).

lajosrodek: Got it, thanks.

lajosrodek (approved):
LGTM

core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
doktoric: I think should be part of the csd right ?

adenes (approved):
LGTM, thanks @vbrodetskyi

core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
vbrodetskyi: @doktoric no, it should not be a part of CSD. For private cloud clusters, user should set values for these properties manually.

doktoric: @vbrodetskyi do we always need this property? if yes should we create configprovider for this?

vbrodetskyi: @doktoric we definitely need these two properties for 7.2.8 (CFM-2.0.8.0), but in future versions, it could be changed. I don't think we need configprovider, these properties should be enough.

doktoric (approved):

PR #10316, opened 2021-03-19
CB-11692: Fix FreeIPA HA repair when master instance is stopped

jamisonbennett:
Fix FreeIPA HA repair when master instance is stopped or doesn't
respond.

A unit test was added and it was manually tested using a local
deployment of cloudbreak.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
@@ -61,12 +61,9 @@ public class ChangePrimaryGatewayService {
                     newPrimaryGatewayInstanceId = assignNewPrimaryGatewayInstanceId(stack, repairInstanceIds);
                 }
             }
-        } catch (NotFoundException e) {
-            LOGGER.debug("No primary gateway found, searching for an different primary gateway");
+        } catch (NotFoundException | CloudbreakOrchestratorException e) {
+            LOGGER.debug("No primary gateway found, searching for a different primary gateway");
             newPrimaryGatewayInstanceId = assignNewPrimaryGatewayInstanceId(stack, repairInstanceIds);
-        } catch (CloudbreakOrchestratorException e) {
-            LOGGER.debug("The orchestrator failed to identify the primary gateway node", e);
-            throw new NotFoundException("No FreeIPA primary gateway could be selected", e);
         }
         LOGGER.debug("The new primary gateway will be {}", newPrimaryGatewayInstanceId);
         return newPrimaryGatewayInstanceId;
--- a/freeipa/src/test/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayServiceTest.java
+++ b/freeipa/src/test/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayServiceTest.java
@@ -21,6 +21,7 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
+import com.sequenceiq.cloudbreak.orchestrator.exception.CloudbreakOrchestratorFailedException;
 import com.sequenceiq.cloudbreak.orchestrator.host.HostOrchestrator;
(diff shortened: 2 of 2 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
```diff
@@ -62,11 +62,11 @@ public String selectNewPrimaryGatewayInstanceId(Stack stack, List<String> repair
                 }
             }
         } catch (NotFoundException e) {
-            LOGGER.debug("No primary gateway found, searching for an different primary gateway");
+            LOGGER.debug("No primary gateway found, searching for a different primary gateway");
             newPrimaryGatewayInstanceId = assignNewPrimaryGatewayInstanceId(stack, repairInstanceIds);
```
jasonwmcswain, line 66: @jamisonbennett , Does it make sense to add a follow-up message to know which is the new primary gateway instance?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
```diff
@@ -62,11 +62,11 @@ public String selectNewPrimaryGatewayInstanceId(Stack stack, List<String> repair
...
         } catch (CloudbreakOrchestratorException e) {
-            LOGGER.debug("The orchestrator failed to identify the primary gateway node", e);
-            throw new NotFoundException("No FreeIPA primary gateway could be selected", e);
+            LOGGER.debug("The orchestrator failed to identify the primary gateway node, searching for a different primary gateway", e);
+            newPrimaryGatewayInstanceId = assignNewPrimaryGatewayInstanceId(stack, repairInstanceIds);
```
jasonwmcswain, line 69: same question as above.

jasonwmcswain (approved):

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
christmasferret: what is the scenario difference between NotFoundException and CloudbreakOrchestratorException? 

catch(NotFoundException | CloudbreakOrchestratorException e) {
and just print the internal exception msg?

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
jamisonbennett: https://github.com/hortonworks/cloudbreak/blob/d67401acd4164577b44bff9a992113ac7556c518/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java#L95
The next log statement which will print for this flow will log "Changing the primary gateway metadata from _ to _" (or it will log "The primary gateway _ metadata did not change" for the case that there isn't a change.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
jamisonbennett: I was considering combining these before. I will do that now.

For reference:`NotFoundException` is thrown when the FreeIPA primary gateway node has been deleted. While `CloudbreakOrchestratorException` is thrown when the FreeIPA primary gateway node still exists but there is a failure to communicate with it (e.g. stopped, unreachable, unhealthy, etc.).

freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/repair/changeprimarygw/ChangePrimaryGatewayService.java
jamisonbennett: It looks like line 71 already has this follow-up message. Let me know if there is something else that you think is needed.
```
        LOGGER.debug("The new primary gateway will be {}", newPrimaryGatewayInstanceId);
```

lacikaaa (approved):

PR #10317, opened 2021-03-19
CB-11690 redbeams tags now sended and not requested by microservices.…

doktoric:
… With this approach any experience can send us requests with tags and will be able to provision resources for them.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/TagUtil.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/TagUtil.java
@@ -1,0 +1,27 @@
+package com.sequenceiq.datalake.service;
+
+import java.util.HashMap;
+import java.util.Map;
+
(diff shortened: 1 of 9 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/TagUtil.java
```diff
@@ -0,0 +1,29 @@
...
+public class TagUtil {
+
+    private TagUtil() {
+
+    }
+
+    public static Map<String, String> getTags(Json tag) {
+        try {
+            if (tag != null && tag.getValue() != null) {
+                return tag.get(new TypeReference<>() {
```
biharitomi, line 19: Can't we get a HashMap instance of TypeReference?

biharitomi (changes requested):

biharitomi (approved):

PR #10327, opened 2021-03-22
CB-11660 Log more info for Cloudformation stack when it's failing or …

lacikaaa:
…timing out

In this commit in case of Cloudformation timeout or error we will collect the CF related events and log it.
This will help us to investigate if the template has been modified from outside.

### DIFF
--- a/cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsCloudFormationErrorMessageProviderTest.java
+++ b/cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsCloudFormationErrorMessageProviderTest.java
@@ -15,6 +15,8 @@ import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
+import com.amazonaws.services.cloudformation.model.DescribeStackEventsRequest;
+import com.amazonaws.services.cloudformation.model.DescribeStackEventsResult;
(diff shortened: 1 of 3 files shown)


cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/util/AwsCloudFormationErrorMessageProviderTest.java
```diff
@@ -69,6 +71,7 @@ void shouldReportAddressLimitExceededOnCreate() {
                         .withResourceStatus(ResourceStatus.CREATE_COMPLETE.toString())
                         .withResourceStatusReason("Created")
         ));
+        when(cfRetryClient.describeStackEvents(any(DescribeStackEventsRequest.class))).thenReturn(new DescribeStackEventsResult());
```
Bajzathd, line 74: I think you should add an event here or add a new test for it.

lacikaaa: as the result is logged only and not returned for the UI it would be possible to test only if I move it to a separate class. But in that case all the others should be moved to a separate class. If you think it worth the effort I will do it.

Bajzathd: I see, then you can leave it as is.

Bajzathd (approved):

PR #10329, opened 2021-03-22
CB-11659 Add logrotate config for /pki/pki-tomcat/ca/debug on FreeIPA

christmasferret:
Currently /pki/pki-tomcat/ca/debug doesn't have logrotate configuration. It should have similarly to ipabackup.log. We add logrotate rule to avoid the log to fill the disk space.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/logrotate/conf/pkidebug
+++ b/freeipa/src/main/resources/freeipa-salt/salt/logrotate/conf/pkidebug
@@ -1,0 +1,10 @@
+/var/log/pki/pki-tomcat/ca/debug {
+    missingok
+    notifempty
+    copytruncate
+    daily
+    size 128M
+    rotate 6
+    su pkiuser pkiuser
+    nocompress
+}
--- a/freeipa/src/main/resources/freeipa-salt/salt/logrotate/init.sls
+++ b/freeipa/src/main/resources/freeipa-salt/salt/logrotate/init.sls
@@ -13,3 +13,11 @@ logrotate-ipabackup:
     - user: root
     - group: root
     - mode: 644
+
+logrotate-pkidebug:
+  file.managed:
+    - name: /etc/logrotate.d/pkidebug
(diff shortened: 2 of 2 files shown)


christmasferret:
tested the logrotate rule.

I have to add service restart at the end, because the service will hold the fd. If I don't restart, the service will still write to debug.1 instead of debug as it is open by the service. 

    postrotate
        /bin/systemctl restart pki-tomcatd@pki-tomcat > /dev/null 2>/dev/null || true

freeipa/src/main/resources/freeipa-salt/salt/logrotate/conf/pkidebug
```diff
@@ -0,0 +1,12 @@
...
+    missingok
+    notifempty
+    create 0640 pkiuser pkiuser
+    daily
+    size 128M
+    rotate 6
+    su pkiuser pkiuser
+    nocompress
+    postrotate
+        /bin/systemctl restart pki-tomcatd@pki-tomcat > /dev/null 2>/dev/null || true
```
jamisonbennett, line 11: I am thinking this just will impact the CA & CRL creation for a short time. Do you know if it impacts anything else?

lacikaaa: please try with `copytruncate` instead of restarting the service: https://linux.die.net/man/8/logrotate

lacikaaa (changes requested):

freeipa/src/main/resources/freeipa-salt/salt/logrotate/conf/pkidebug
christmasferret: Thanks! I tested copytruncate. Now it is working good without restart the service.

christmasferret:
I tested copytruncate. Instead of creating a new log file with the same name, it will keep the original file and move content to the rotation file. Now the service can keep writing to the same file without restarting the service.

lacikaaa (approved):

PR #10330, opened 2021-03-22
CB-11675 - Added Knox's own KNOXTOKEN service and the new tokengen ap…

keyki:
…plication into the cdp-proxy topology

Two new artifacts have to be added in cdp-proxy topology:

Knox's own KNOXTOKEN service (it's proprietary to Knox, not like a regular service such Hive, YARN, ATLAS,...) with a custom configuration that is not meant to be changed from the UI
A new Knox application called tokengen. It's 100% guaranteed that all versions after CDH 7.2.9 will have it

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -1,13 +1,12 @@
 package com.sequenceiq.cloudbreak.core.bootstrap.service.host;
 
+import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CDP_VERSION_7_2_9;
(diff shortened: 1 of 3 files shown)


smolnar82 (commented):
Thanks, @keyki !
LGTM +1

keyki:
The PR is ready, but I put it to Draft mode, because we need a new 7.2.9 other the URLs won't work
```
Failed to deploy topology cdp-proxy: org.apache.knox.gateway.deploy.DeploymentException: Failed to find application in /var/lib/knox/gateway/data/applications/tokengen
```
@smolnar82 will let us know once L0s pass and we have a GBN.

smolnar82:
@keyki - based on the latest changes in [CB-11706](https://jira.cloudera.com/browse/CB-11706) we do not need this commit to be merged.
The reason being is that from now on we're going to have access to the Knox Home page via the new `Token Integration` link where the `KNOXTOKEN` service and the new `tokengen` application will be available for CDH 7.2.9 and above.
Thanks for your time you spend on this item it was at least very valuable for me.

PR #10332, opened 2021-03-22
CB-11706 Add a Link to the Topology Generation page in cdp-proxy Knox…

doktoric:
… topology

### DIFF
--- a/core-api/src/main/resources/definitions/exposed-services.json
+++ b/core-api/src/main/resources/definitions/exposed-services.json
@@ -639,6 +639,21 @@
       "apiIncluded": false,
       "visibleForDatalake": false,
       "visibleForDatahub": true
+    },
+    {
+      "name": "KNOX_TOKEN_INTEGRATOR",
+      "displayName": "Token Integration",
+      "serviceName": "KNOX",
+      "knoxService": "KNOX_TOKEN_INTEGRATOR",
+      "knoxUrl": "/homepage/home",
+      "ssoSupported": true,
+      "apiOnly": false,
+      "apiIncluded": false,
+      "entitlement": "OJDBC_TOKEN_DH",
+      "visibleForDatalake": true,
+      "visibleForDatahub": true,
+      "withoutProxyPath": true,
+      "minVersion": "7.2.9"
     }
   ]
 }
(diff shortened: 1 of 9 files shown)


core-api/src/main/resources/definitions/exposed-services.json
```diff
@@ -639,6 +639,20 @@
       "apiIncluded": false,
       "visibleForDatalake": false,
       "visibleForDatahub": true
+    },
+    {
+      "name": "KNOX_TOKEN_GENERATOR",
+      "displayName": "Knox Token Generation",
```
smolnar82, line 645: Larry suggested `Token Integration`. What do you think?

smolnar82 (commented):
I've only had a minor comment on the display name. Please check and change if you agree with the proposed name.

core-api/src/main/resources/definitions/exposed-services.json
```diff
@@ -639,6 +639,20 @@
...
+      "serviceName": "KNOX",
+      "knoxService": "KNOX_TOKEN_GENERATOR",
+      "knoxUrl": "/tokengen/index.html",
```
lmccay, line 648: As I mentioned in the slack channel, I think that we want this to be more of a Token Integrations link at this point. This will point to the Knox /homepage/home link which will display the homepage with default settings focused on the cdp-proxy-token topology and related links. It will also have the link to Token Generation page on the home page. This allows us to NOT have to add a whole other set of API endpoints for cdp-proxy-token and the user having to know when to use which.

I know this is different than was described in the JIRA - we only came to this conclusion on Friday. 

I am also curious whether that knoxUrl link works like that without being fully qualified - maybe you fully qualify it internally upon deployment? I don't think where that link lives that it would work as a root relative path from CP.

lmccay (changes requested):

core-api/src/main/resources/definitions/exposed-services.json
```diff
@@ -639,6 +639,20 @@
       "apiIncluded": false,
       "visibleForDatalake": false,
       "visibleForDatahub": true
+    },
+    {
+      "name": "KNOX_TOKEN_INTEGRATOR",
+      "displayName": "Token Integration",
+      "serviceName": "KNOX",
+      "knoxService": "KNOX_TOKEN_INTEGRATOR",
+      "knoxUrl": "/tokengen/index.html",
```
smolnar82, line 648: @doktoric - the `knoxUrl` should be updated to `/homepage/home` too.
We really appreciate your help with this!

smolnar82 (changes requested):

smolnar82 (commented):
LGTM +1

smolnar82 (approved):

PR #10334, opened 2021-03-22
CB-11680 Improve root cause detection after a complex operation fails

lturcsanyi:
If something wrong happens during flow execution (currently cluster upscale and cluster downscale), we should run some checks in order to find the root cause of the error.
Currently salt minion check is implemented.
If the conclusion checker finds something it will send an event to the UI.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/SaltMinionCheckerConclusionStep.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/SaltMinionCheckerConclusionStep.java
@@ -1,0 +1,44 @@
+package com.sequenceiq.cloudbreak.conclusion.step;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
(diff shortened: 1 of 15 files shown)


horadla23 (approved):
LGTM

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/SaltMinionCheckerConclusionStep.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public ConclusionStepResult check(Long resourceId) {
+        Stack stack = stackService.getByIdWithListsInTransaction(resourceId);
+        Set<String> allNodes = stackUtil.collectNodes(stack).stream().map(Node::getHostname).collect(Collectors.toSet());
+        try {
+            stackUtil.collectAndCheckReachableNodes(stack, allNodes);
+        } catch (NodesUnreachableException e) {
+            Set<String> unreachableNodes = e.getUnreachableNodes();
+            LOGGER.error("Unreachable salt minions: {}", unreachableNodes);
+            String conclusion = String.format("Unreachable nodes: %s. Please check the instances on your cloud provider for further details.",
```
lnardai, line 38: Wouldn't it be better to include the logging inside the failed method, or make it so that if you set up a failure case, it's not possible to forget logging.
Its might useful the add another `evalute`, or finalize method for the ConclusionStepResult which logs what we found.

lnardai (approved):

PR #10336, opened 2021-03-23
CDPCP-4195. Allow invoking user sync against non-existent users

handavid:
Partial user sync failed to retrieve the UMS state if a specified
user or machineUser does not exist in UMS. This could be a problem
if a user sync invocation races against deletion of the actor. This
commit fixes this issue by reverting to retrieving users and
machineUsers individually if we receive a NOT_FOUND exception retrieving
the actors as a group. The missing actor will be recorded in the
warnings for the user sync.

Along with this change, parameter lists for some methods have been
simplified. Whether a user sync is a full sync or partial sync is dependent
upon the parameters in the UserSyncRequestFilter. I've removed
the "full sync" boolean parameter from methods that include the
request filter as a parameter. I've also changed methods that
include user and machine user crn lists to include the request
filter instead.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncRequestFilter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/UserSyncRequestFilter.java
@@ -14,10 +14,13 @@ public class UserSyncRequestFilter {
 
     private final Optional<String> deletedWorkloadUser;
 
+    private final boolean fullSync;
+
     public UserSyncRequestFilter(Set<String> userCrnFilter, Set<String> machineUserCrnFilter, Optional<String> deletedWorkloadUser) {
         this.userCrnFilter = ImmutableSet.copyOf(userCrnFilter);
         this.machineUserCrnFilter = ImmutableSet.copyOf(machineUserCrnFilter);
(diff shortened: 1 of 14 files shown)


handavid:
@hsheinblatt Let's discuss this approach here. 
The benefit of this change is that we can sync some of the users specified in the partial sync request. This change also provides us information about which CRN failed to sync.
The old behavior did fail fast, but would not provide any information about which actor is missing when multiple actors are included in the sync.
From the WorkloadIAM standpoint, we'd eventually like to use the failure details to respond in a more precise manner. We aren't there yet, and will have to treat the success or failure as a whole anyways for the time being.

horadla23:
@handavid Since this change looks quite big and makes the user sync even a little bit more complex, I would approach this issue from another point of view, for example in case of partial user sync, when users/machine users specified, we could check for those users in UMS before the actual usersync as a validation and throw a bad request if some of the user missing. Is it possible?

handavid:
@horadla23 I think that rejecting the request would also be a valid approach to improving this behavior. However, the current behavior will already fail early in the user sync. We will work around this in workload iam once we improve our error handling codepaths so we could also just leave the behavior as-is and wait for the overall automated user sync story.

handavid:
I'm discarding this PR in preference for keeping current behavior.

PR #10343, opened 2021-03-23
CB-11666 in case of missing network on provider side we are getting a…

doktoric:
… nullpointer. Adding a check for subnetworks.


biharitomi (approved):

doktoric:
unit tests will go to master

PR #10352, opened 2021-03-24
CB-11778 Fix NullPointer for `RetryableFreeIpaClientException#toString`

lacikaaa:
When `exceptionForRestApi` is not set `toString` could cause exception. This is fixed by handling null gracefully.
Also `FreeIpaClientExceptionWrapper` contructor has been edited to include the message, so it won't call `toString` on the wrapped exception.


doktoric:
this must go as a hotfix when we will update production.

jamisonbennett (approved):
Thanks for finding this and fixing it.

PR #10357, opened 2021-03-24
CB-11572 add datalakecrn to stack in order to eliminate datalake resource; all datalakeresource activity moved to datalakeservice

afarsang:
…urce; all datalakeresource activity moved to datalakeservice


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -550,19 +550,28 @@ private void saveCustomNameservers(Stack stack, KerberosConfig kerberosConfig, M
...
-            Optional<DatalakeResources> datalakeResource = datalakeResourcesService.findById(datalakeResourceId);
-            if (datalakeResource.isPresent() && datalakeResource.get().getDatalakeStackId() != null) {
-                Long datalakeStackId = datalakeResource.get().getDatalakeStackId();
-                Stack dataLakeStack = stackService.getByIdWithListsInTransaction(datalakeStackId);
-                String datalakeDomain = dataLakeStack.getGatewayInstanceMetadata().get(0).getDomain();
-                List<String> ipList = dataLakeStack.getGatewayInstanceMetadata().stream().map(InstanceMetaData::getPrivateIp).collect(Collectors.toList());
-                servicePillar.put("forwarder-zones", new SaltPillarProperties("/unbound/forwarders.sls",
-                        singletonMap("forwarder-zones", singletonMap(datalakeDomain, singletonMap("nameservers", ipList)))));
-            }
+        Optional<Stack> datahubStack = datalakeService.getDatalakeStackByDatahubStack(stack);
```
lajosrodek, line 555: I suggest renaming it to `dataLakeStackOptional`.

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
```diff
@@ -108,6 +108,7 @@
                 newRdsConfig = createNewRdsConfig(stack, cluster, getDb(), getDbUser(), getDbPort());
             }
             rdsConfigs = populateNewRdsConfig(rdsConfigs, stack, cluster, newRdsConfig);
+            rdsConfigs = populateNewRdsConfig(rdsConfigs, stack, cluster, newRdsConfig);
```
lajosrodek, line 111: Why the duplication?

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
afarsang: Ok

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/AbstractRdsConfigProvider.java
afarsang: it was a mistake

core/src/main/resources/schema/app/20210323131901_CB-11572_datahub_connected_to_datalake_without_datalakeresource.sql
```diff
@@ -0,0 +1,17 @@
...
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE stack ADD COLUMN IF NOT EXISTS datalakecrn varchar(255);
+
+CREATE INDEX idx_stack_datalakecrn ON stack (datalakecrn);
+
+UPDATE stack s
+SET datalakecrn = dl.resourcecrn
+FROM (
+  SELECT s1.id, s2.resourcecrn FROM stack s1, datalakeresources d, stack s2 WHERE s1.type = 'WORKLOAD' AND s1.datalakeresourceid = d.id AND d.datalakestack_id = s2.id
```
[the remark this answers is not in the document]
afarsang: datahubs (s1) are connected to datalake (s2) through datalakeresource

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,268 @@
...
+        return Optional.empty();
+    }
+
+    private String getHostForComponentInService(DatalakeResources datalakeResources, String service, String serviceComponent) {
+        if (datalakeResources.getServiceDescriptorMap().containsKey(service)) {
+            ServiceDescriptor serviceDescriptor = datalakeResources.getServiceDescriptorMap().get(service);
+            Map<String, Object> componentHostMap = serviceDescriptor.getComponentsHosts().getMap();
+            return String.valueOf(componentHostMap.get(serviceComponent));
+        } else {
+            return null;
```
gregito, line 193: Like in the case of other methods it would be nice to return an optional instead of null just to warn the developer about that this value can be null and should be handled that way later on

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,268 @@
...
+
+    private String getServiceRelatedSecret(Map<String, ServiceDescriptor> serviceDescriptorMap, String service, String secretKey) {
+        if (serviceDescriptorMap.containsKey(service)) {
+            ServiceDescriptor serviceDescriptor = serviceDescriptorMap.get(service);
+            Map<String, Object> secretMap = serviceDescriptor.getBlueprintSecretParams().getMap();
+            if (secretMap.containsKey(secretKey)) {
+                return String.valueOf(secretMap.get(secretKey));
+            }
+        }
+        return null;
```
gregito, line 218: same here for the optional-null topic

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,268 @@
...
+
+    private DatalakeResources getDatalakeResourceId(StackV4Request source, Workspace workspace) {
+        try {
+            if (source.getSharedService() != null && isNotBlank(source.getSharedService().getDatalakeName())) {
+                return datalakeResourcesService.getByNameForWorkspace(source.getSharedService().getDatalakeName(), workspace);
+            }
+        } catch (NotFoundException nfe) {
+            LOGGER.debug("No datalake resource found for data lake: {}", source.getSharedService().getDatalakeName());
+        }
+        return null;
```
gregito, line 237: same here for the optional-null topic

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    private SharedServiceConfigsView createSharedServiceConfigs(Blueprint blueprint, String ambariPassword,
+            Optional<DatalakeResources> datalakeResources) {
```
gregito, line 241: in general, passing an Optional as a method argument is not a very welcomed way since it can be tricky because one has to check whether the actual Optional is not null or not, and only after that can be checked with isPresent.

core/src/main/resources/schema/app/20210323131901_CB-11572_datahub_connected_to_datalake_without_datalakeresource.sql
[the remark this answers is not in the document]
lajosrodek: Ah, indeed, sorry. Need more coffee. 😃

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
afarsang: it is a private function, it won't be a problem

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/view/ClusterApiViewToClusterViewV4ResponseConverter.java
```diff
@@ -31,7 +29,7 @@ public ClusterViewV4Response convert(ClusterApiView source) {
         clusterViewResponse.setStatus(source.getStatus());
         clusterViewResponse.setHostGroups(convertHostGroupsToJson(source.getHostGroups()));
         clusterViewResponse.setCertExpirationState(source.getCertExpirationState());
-        addSharedServiceResponse(source, clusterViewResponse);
+        datalakeService.addSharedServiceResponse(source, clusterViewResponse);
```
doktoric, line 32: this will be very slow when we list clusters

afarsang: we discussed it

doktoric (approved):

usage-collection/src/main/proto/usage.proto
```diff
@@ -825,14 +825,26 @@ message CDPStatusDetails {
...
+}
+
 // Generated when Datalake creation has been requested or finished
 message CDPDatalakeRequested {
   // Operation metadata
   CDPOperationDetails operationDetails = 1;
   // Environment Crn
   string environmentCrn = 2;
-  // Details for Datalake
+  // Cluster details for Datalake
```
doktoric, line 844: do we need this in the current change?

core/src/main/resources/schema/app/20210323131901_CB-11572_datahub_connected_to_datalake_without_datalakeresource.sql
```diff
@@ -0,0 +1,17 @@
+-- // CB-11572 datahub connected to the datalake without datalakeresource
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE stack ADD COLUMN IF NOT EXISTS datalakecrn varchar(255);
```
doktoric, line 4: can we increase this and not use 255 ?

usage-collection/src/main/proto/usage.proto
afarsang: I guess I need a rebase...

core/src/main/resources/schema/app/20210323131901_CB-11572_datahub_connected_to_datalake_without_datalakeresource.sql
afarsang: I can. What do you suggest?

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+                .findDatalakeResourcesByWorkspaceAndEnvironment(datahubStack.getWorkspace().getId(), datahubStack.getEnvironmentCrn());
+        Optional<Stack> datalakeStack = Optional.ofNullable(datalakeResources)
+                .map(Set::stream).flatMap(Stream::findFirst)
+                .map(DatalakeResources::getDatalakeStackId)
+                .map(stackService::getByIdWithListsInTransaction);
+
+        return datalakeStack;
+    }
+
+    public Long getDatalakeResourceId(Long datalakeStackId) {
```
bergerdenes, line 143: you could use the Optional pattern, like in another places

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+        return null;
+    }
+
+    public SharedServiceConfigsView createSharedServiceConfigs(Stack source) {
+        Cluster cluster = source.getCluster();
+        Optional<DatalakeResources> datalakeResources = getDataLakeResource(source);
+        return createSharedServiceConfigs(cluster.getBlueprint(), cluster.getPassword(), datalakeResources);
+    }
+
+    public String getDatalakeCrn(StackV4Request source, Workspace workspace) {
```
bergerdenes, line 157: you could use the Optional pattern, like in another places

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+                }
+            } else {
+                LOGGER.debug("Unable to find DatalakeResources by the stack's id: {}", stack.getId());
+            }
+        }
+        stackResponse.setSharedService(sharedServiceResponse);
+    }
+
+    public Optional<Stack> getDatalakeStackByDatahubStack(Stack datahubStack) {
+        Long datalakeResourceId = datahubStack.getDatalakeResourceId();
```
bergerdenes, line 119: This is only working with the previous version which generates the datalakeResourceId
But it should be the fallback solution only.
First we should try to read the new field `datalakeCrn` and then fallback

core/src/main/resources/schema/app/20210323131901_CB-11572_datahub_connected_to_datalake_without_datalakeresource.sql
doktoric: 512

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+        LOGGER.info("There is no datalake resource id has set for the cluster.");
+        return Optional.empty();
+    }
+
+    public Optional<Stack> getDatalakeStackByStackEnvironmentCrn(Stack datahubStack) {
+        Set<DatalakeResources> datalakeResources = datalakeResourcesService
+                .findDatalakeResourcesByWorkspaceAndEnvironment(datahubStack.getWorkspace().getId(), datahubStack.getEnvironmentCrn());
+        Optional<Stack> datalakeStack = Optional.ofNullable(datalakeResources)
```
bergerdenes, line 135: This is only working with the previous version which generates the datalakeResourceId
But it should be the fallback solution only.
First we should try to read the new field datalakeCrn and then fallback

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+                SharedServiceV4Request sharedServiceRequest = new SharedServiceV4Request();
+                sharedServiceRequest.setDatalakeName(datalakeResources.get().getName());
+                stackRequest.setSharedService(sharedServiceRequest);
+            }
+        }
+    }
+
+    public void addSharedServiceResponse(ClusterApiView cluster, ClusterViewV4Response clusterResponse) {
+        SharedServiceV4Response sharedServiceResponse = new SharedServiceV4Response();
+        if (cluster.getStack().getDatalakeId() != null) {
```
bergerdenes, line 74: this should only be a fallback option, no?

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+            if (result.isPresent()) {
+                return result.get().getResourceCrn();
+            } else {
+                LOGGER.debug("No datalake resource found for data lake: {}", source.getSharedService().getDatalakeName());
+            }
+        }
+        return null;
+    }
+
+    public void setDatalakeIdOnStack(Stack stack, StackV4Request source, Workspace workspace) {
```
bergerdenes, line 170: the setting could be pushed to the side where this method is called (e.g. the other methods returning values, this is setting a stack property)

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
```diff
@@ -0,0 +1,265 @@
...
+    private DatalakeResourcesService datalakeResourcesService;
+
+    @Inject
+    private StackService stackService;
+
+    @Inject
+    private BlueprintUtils blueprintUtils;
+
+    public void prepareDatalakeRequest(Stack source, StackV4Request stackRequest) {
+        if (source.getDatalakeResourceId() != null) {
```
bergerdenes, line 62: this should be a fallback only
what if `databaseCrn` is present?

bergerdenes (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/sharedservice/DatalakeService.java
afarsang: Why it is necessary? I think in the second release we can implement it without fallback logic

PR #10360, opened 2021-03-24
DISTX-595 Release a first version of Oozie HA without the internal lo…

cegganesh84:
…ad-balancer

1. In public cloud all the requests go through Knox, so we do not need a load-balancer for external clients.
3. This is for Hue to access the Oozie in HA fashion.
4. Hue can not use Knox for this because the URL is different from what Hue expects it to be.
5. As a workaround use the first FQDN in the master hosts to configure it as a load-balancer to itself.
6. This means Oozie will be HA for the schedules, but can not accept modifications from Hue when the first FQDN in the master host is down. It still can accept modifications from external clients via Knox.
7. Added a TODO pointing to the parent task to add an internal load-balancer for Hue to work with Oozie in HA fashion.


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final String OOZIE_HTTP_PORT = "11000";
+
+    private static final String OOZIE_HTTPS_PORT = "11443";
+
+    @Override
+    public List<ApiClusterTemplateConfig> getServiceConfigs(CmTemplateProcessor templateProcessor, TemplatePreparationObject source) {
+        // TODO: DISTX-548 Add an internal load-balancer for Oozie. As a workaround use the first FQDN in the hostgroup as load-balancer.
+        return List.of(
+            config(OOZIE_LB_HOST, firstOozieFQDN(source)),
```
sidseth, line 32: CM config used by Hue only?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
```diff
@@ -25,18 +27,28 @@
...
-                        config(OOZIE_DATABASE_PASSWORD, oozieRdsView.getConnectionPassword())
-                );
+                List<ApiClusterTemplateConfig> config = new ArrayList<>(List.of(
+                    config(OOZIE_DATABASE_HOST, oozieRdsView.getHost()),
+                    config(OOZIE_DATABASE_NAME, oozieRdsView.getDatabaseName()),
+                    config(OOZIE_DATABASE_TYPE, oozieRdsView.getSubprotocol()),
+                    config(OOZIE_DATABASE_USER, oozieRdsView.getConnectionUserName()),
+                    config(OOZIE_DATABASE_PASSWORD, oozieRdsView.getConnectionPassword())));
+                if (isOozieHA(source)) {
+                    config.add(config(OOZIE_PLUGINS_LIST, OOZIE_HA_PLUGINS));
```
sidseth, line 49: Have no idea what this is. Assuming it is needed :)

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
cegganesh84: Yes.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
cegganesh84: Looks like Oozie uses ZK for coordination in HA mode. This plugin list enables that coordination.

sidseth (approved):

PR #10361, opened 2021-03-25
CB-11726 Deletion of stack is retried even when then stack does not e…

lturcsanyi:
…xist

If ActionFailedException is thrown, it will retry the describeStacks request (maximum) 15 times.
When the AmazonServiceException contains the 'stackName does not exist' message, there is no need to retry.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsTerminateService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsTerminateService.java
@@ -97,8 +97,8 @@ public class AwsTerminateService {
         DescribeStacksRequest describeStacksRequest = new DescribeStacksRequest().withStackName(cFStackName);
         try {
             retryService.testWith2SecDelayMax15Times(() -> isStackExist(amazonCloudFormationClient, cFStackName, describeStacksRequest));
-        } catch (ActionFailedException ignored) {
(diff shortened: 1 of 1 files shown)


cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsTerminateService.java
```diff
@@ -97,8 +97,8 @@ private void waitAndDeleteCloudformationStack(AuthenticatedContext ac, CloudStac
         DescribeStacksRequest describeStacksRequest = new DescribeStacksRequest().withStackName(cFStackName);
         try {
             retryService.testWith2SecDelayMax15Times(() -> isStackExist(amazonCloudFormationClient, cFStackName, describeStacksRequest));
-        } catch (ActionFailedException ignored) {
-            LOGGER.debug("Stack not found with name: {}", cFStackName);
+        } catch (ActionFailedException | AmazonServiceException e) {
```
lnardai, line 100: Can we have a separate catch for this particular case, that clearly states we did not found the stack?
Isn't it possible that for some reason we get ActionFailed, like "expired credential", the we proceed with the delete?

lturcsanyi: Yes, you are right.

lnardai (approved):

PR #10362, opened 2021-03-25
CDPD-16919 Phoenix 5.1 DataHub integration

richardantal:
To test this change I created Custom cluster template (called OPDB omid testv3)
I used this template to create a datahub cluster (https://console.dps.mow-dev.cloudera.com/cloud/workloads/details/richarda-opdb-omid/events)

In the first attempt I faced an issue that Omid-191 was not yet merged to cdpd-master. I pushed that change and it is in since  build number 7.2.9.0-154 so I had to specify the GBN as well.

Omid stated on the cluster and I ran some queries to test transactions. create transactional table, insert, commit, select

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
@@ -5,6 +5,17 @@
     "displayName": "opdb",
     "services": [
       {
+        "refName": "omid",
+        "serviceType": "OMID",
+        "roleConfigGroups": [
+          {
+            "refName": "omid-OMID_TSO-BASE",
+            "roleType": "OMID_TSO",
+            "base": true
+          }
+        ]
+      },
+      {
         "refName": "zookeeper",
         "serviceType": "ZOOKEEPER",
         "roleConfigGroups": [
@@ -120,7 +131,7 @@
             "configs": [
               {
                 "name": "hbase_client_config_safety_valve",
-                "value": "<property><name>phoenix.schema.isNamespaceMappingEnabled</name><value>true</value></property><property><name>phoenix.functions.allowUserDefinedFunctions</name><value>true</value></property>"
+                "value": "<property><name>phoenix.schema.isNamespaceMappingEnabled</name><value>true</value></property><property><name>phoenix.functions.allowUserDefinedFunctions</name><value>true</value></property><property><name>phoenix.transactions.enabled</name><value>true</value></property>"
               }
             ],
             "base": true
@@ -313,7 +324,8 @@
           "yarn-JOBHISTORY-BASE",
           "yarn-RESOURCEMANAGER-BASE",
           "yarn-GATEWAY-BASE",
-          "zookeeper-SERVER-BASE"
+          "zookeeper-SERVER-BASE",
+          "omid-OMID_TSO-BASE"
         ]
       },
       {
--- a/core/src/main/resources/defaults/blueprints/7.2.9/cdp-opdb.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.9/cdp-opdb.bp
@@ -5,6 +5,17 @@
     "displayName": "opdb",
     "services": [
       {
+        "refName": "omid",
+        "serviceType": "OMID",
+        "roleConfigGroups": [
+          {
+            "refName": "omid-OMID_TSO-BASE",
+            "roleType": "OMID_TSO",
+            "base": true
+          }
+        ]
+      },
+      {
         "refName": "zookeeper",
         "serviceType": "ZOOKEEPER",
         "roleConfigGroups": [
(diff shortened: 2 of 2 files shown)


stoty (commented):
+1 LGTM

doktoric:
@richardantal do we have a UI for phoenix? do want to expose that?

richardantal:
In this change I added Omid to the cluster template. One of the new feature brought by Phoenix 5.1 is the transaction support, for which we need Omid.

richardantal:
AFAIK this ticket blocks the phoenix and omid testing in 7.2.9

core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
```diff
@@ -99,7 +110,7 @@
           },
           {
             "name": "hbase_service_config_safety_valve",
-            "value": "<property><name>phoenix.schema.isNamespaceMappingEnabled</name><value>true</value></property><property><name>phoenix.functions.allowUserDefinedFunctions</name><value>true</value></property>"
+            "value": "<property><name>phoenix.schema.isNamespaceMappingEnabled</name><value>true</value></property><property><name>phoenix.functions.allowUserDefinedFunctions</name><value>true</value></property><property><name>phoenix.transactions.enabled</name><value>true</value></property>"
```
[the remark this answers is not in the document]
richardantal: Thanks @chrajeshbabu, I have updated the PR

tamaashu:
> @richardantal do we have a UI for phoenix?

No, we don't have a UI for Phoenix.

core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
```diff
@@ -313,7 +324,8 @@
           "yarn-JOBHISTORY-BASE",
           "yarn-RESOURCEMANAGER-BASE",
           "yarn-GATEWAY-BASE",
-          "zookeeper-SERVER-BASE"
+          "zookeeper-SERVER-BASE",
+          "omid-OMID_TSO-BASE"
```
joshelser, line 328: Can you double-check the max heap which is allocated for each process on the leader node? Want to make sure that we don't over-commit memory on this host.

CM tries to do this but is more-or-less guessing about it, IIRC.

joshelser (commented):
Looks fine to me. Only concern is about adding it to the leader role (and if we have enough heap size to not cause physical memory issues). As long as we still have 1GB of memory on the machine for the operating system to use, we should be OK.

Can you double-check that please, @richardantal ? Otherwise, lgtm.

doktoric (changes requested):
phoenix.transactions.enabled is coming with safety valve. We prefer that if you find such property which is not exposed as fixed property in CM then raise an OPSAPS issue for that.

core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
richardantal: Memory usage seems to be fine, we still have more memory on the leader node than on the workers for example.
https://richard-opdb-omid-gateway0.richarda.xcu2-8y8x.dev.cldr.work/cmf/hardware/hosts/1546333375/status

richardantal:
> phoenix.transactions.enabled is coming with safety valve. We prefer that if you find such property which is not exposed as fixed property in CM then raise an OPSAPS issue for that.

I don't think we want to do this in CM side. Setting this property only makes sense when we have Omid on the cluster, but even then its not sure we ( or customers) want to set it.
With this cluster definition we want to provide transactions support in Phoenix, for which it is necessary to set this property.

richardantal:
> phoenix.transactions.enabled is coming with safety valve. We prefer that if you find such property which is not exposed as fixed property in CM then raise an OPSAPS issue for that.

I've created a ticket for that: https://jira.cloudera.com/browse/OPSAPS-60064

core/src/main/resources/defaults/blueprints/7.2.10/cdp-opdb.bp
richardantal: The leader node is an m5.2xlarge one, it has 32 GB memory.
According to CM the sum of max heap sizes (+ 30% Overhead for every service) is 18,8 Gb, but it writes 10,4 for the ZooKeeper server.
These values are collected from the configs set by blueprint or CM params.

joshelser: Perfect! Thanks for the great analysis.

richardantal:
I think it would be better to get this in before the CM change because it blocks the testing of Omid.
After this we can do the change in a follow up commit. Is it okey for you @doktoric ?

chrajeshbabu (approved):

doktoric (approved):

PR #10365, opened 2021-03-25
CB-11783 Data Hub upgrade should suggest a Data Lake upgrade if there…

keyki:
… are no upgrade options

Currently, the Data Hub upgrade is limited to the same version as the Data Lake. Whenever
there are no upgrade options available for Data Hub it means 2 things:
- it is already on the latest version within the give major.minor.path version, e.g no newer 7.1.0 image is available
- upgrade options are available, e.g 7.2.0, but the Data Lake version is still 7.1.0

In the second case we're going to suggest to the user to upgrade their Data Lake first. In the first case
we're not suggesting an upgrade if there are no options anyways.

### DIFF
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeAvailabilityService.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeAvailabilityService.java
@@ -26,6 +26,7 @@ import com.sequenceiq.cloudbreak.cluster.service.ClusterComponentConfigProvider;
 import com.sequenceiq.cloudbreak.common.exception.BadRequestException;
 import com.sequenceiq.cloudbreak.controller.validation.stack.StackRuntimeVersionValidator;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.domain.view.StackView;
 import com.sequenceiq.cloudbreak.service.stack.StackService;
 import com.sequenceiq.cloudbreak.service.stack.StackViewService;
 import com.sequenceiq.distrox.v1.distrox.StackOperations;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeAvailabilityServiceTest.java
```diff
@@ -199,6 +200,70 @@ public void testOnlyReturnCandidatesWithDatalakeVersion() {
...
+        Stack stackWithEnv = new Stack();
+        stackWithEnv.setEnvironmentCrn("envcrn");
+        StackView stackView = new StackView();
+        ClusterView clusterView = new ClusterView();
+        clusterView.setId(1L);
+        ReflectionTestUtils.setField(stackView, "cluster", clusterView);
+        ClouderaManagerProduct product = new ClouderaManagerProduct();
+        product.setVersion("7.1.0");
+        product.setName("CDH");
+        List<ClouderaManagerProduct> products = List.of(product);
```
pdarvasi: this is not used

core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeAvailabilityService.java
```diff
@@ -69,6 +69,10 @@ public UpgradeV4Response checkForUpgrade(NameOrCrn nameOrCrn, Long workspaceId,
         UpgradeV4Response response = stackOperations.checkForClusterUpgrade(stack, workspaceId, request);
         List<ImageInfoV4Response> filteredCandidates = filterCandidates(stack, request, response.getUpgradeCandidates());
         response.setUpgradeCandidates(filteredCandidates);
+        if (filteredCandidates.isEmpty()) {
```
pdarvasi, line 72: UpgradeV4Response response could be passed instead of response.getUpgradeCandidates() so that 
- we could set the reason during filtering in method filterForDatalakeVersion() --> message will always be 100% accurate
- we could add the versions of the DH and the DL to provide more info --> message will be more useful

pdarvasi (approved):

PR #10370, opened 2021-03-27
CB-11772 Security vulnerability: Bind 8080, 8443 to localhost only

schfeca75:
CB-11772 Security vulnerability: Bind 8080, 8443 to localhost only


schfeca75:
This pullrequest is draft only, because I would like to test the Freeipa HA scenario and the Freeipa repair flows.

PR #10371, opened 2021-03-28
CB-11154 Use DES to enable SSE with CMK during Azure DH cluster creation

lajosrodek:
* When present in `DetailedEnvironmentResponse` (`getDiskEncryptionSetId()`), associate the provided Disk Encryption Set (DES) Resource ID with all the Managed Disks of the DH cluster being created. This applies to the OS disks as well as data (attached) disks, and will make all such disks encrypted by SSE using the customer-managed CMK linked to the DES.
  * `cloud-api` receives the SSE with CMK and DES Resource ID as two new dynamic parameters in `InstanceTemplate`. The parameter name constants are defined in the new class `AzureInstanceTemplate`.
* No entitlement check (regarding `EntitlementService.isAzureDiskSSEWithCMKEnabled()`) is made; the DES Resource ID will be always used and applied to all new DH VM disks when present in `DetailedEnvironmentResponse`.
* StackV4 API has been extended with `AzureEncryptionV4Parameters` that encompasses the `diskEncryptionSetId`. These parameters are contained by the new property `AzureInstanceTemplateV4Parameters.encryption`.
* A small bug has been fixed in `AwsInstanceTemplateV4Parameters.asMap()` around how `AwsInstanceTemplate.EBS_ENCRYPTION_ENABLED` gets populated. This also undoes the related (and ineffective) change of #9378.
* `DistroXV1RequestToStackV4RequestConverter.convertAsTemplate()` was a dead code so has been removed to avoid confusion. The related defunct tests in `DistroXV1RequestToStackV4RequestConverterTest` have been also cleaned up.
* `InstanceTemplateV1ToInstanceTemplateV4Converter` contained duplicated statements `response.setRootVolume(getIfNotNull(source.getRootVolume(), volumeConverter::convert));` in both `convert()` methods. The superfluous invocations have been dropped.
* The `apiVersion` of the `Microsoft.Compute/virtualMachines` resource in `arm-v2.ftl` has been updated from `2018-04-01` to `2019-07-01`; the latter is the earliest version supporting DES for VM disks. (Only the OS disk is created via ARM deployments in our case.) Unfortunately, this version bump came with a backward incompatible change around user-assigned MSIs; see `identity.identityIds` (array of strings) being changed to `identity.userAssignedIdentities` (object of empty objects).
* The legacy Azure Java SDK (`com.microsoft.azure:azure-mgmt-compute:1.38.0`) does not expose DES related functionality for VM disks. Not even the (currently) latest version 1.40.0 does so. We, therefore, have to rely on SDK internal implementation classes to apply the DES Resource ID to newly created attached disks. See `AzureClient.setupDiskEncryptionWithDesIfNeeded()` (called from `AzureClient.createManagedDisk()`).
* Testing:
  * Manually verified with local CB, brand new env, DH with and without DES Resource ID, with simple DES as well as double-encryption DES.
  * Added unit tests and updated existing ones, using JUnit 5 and AssertJ whenever possible. Since the Azure Java SDK has some `final` classes (e.g. `com.microsoft.azure.management.Azure`) that cannot be easily mocked, `AzureClient` validations in `AzureClientTest` are rather limited. Moreover, there are some reflection-based checks in `AzureClientTest.setupDiskEncryptionWithDesIfNeededTestWhenAzureSdkInternalTypeChecks()` that aim to guard against unexpected API changes in case of a future SDK version upgrade.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AzureInstanceTemplate.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AzureInstanceTemplate.java
@@ -1,0 +1,55 @@
+package com.sequenceiq.cloudbreak.cloud.model.instance;
+
+import java.util.Collection;
+import java.util.Map;
+
+import com.sequenceiq.cloudbreak.cloud.model.InstanceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.InstanceTemplate;
+import com.sequenceiq.cloudbreak.cloud.model.Volume;
+
+public class AzureInstanceTemplate extends InstanceTemplate {
+
+    /**
+     * Key of the optional dynamic parameter denoting whether managed disk encryption with a customer-managed encryption key is enabled or not. This applies
+     * to both root & attached (data) volumes.
+     *
+     * <p>
+     *     Permitted values:
+     *     <ul>
+     *         <li>{@code Boolean.TRUE} instance, {@code "true"} (ignoring case): Encryption with a customer-managed encryption key is enabled.</li>
+     *         <li>{@code Boolean.FALSE} instance, {@code "false"} (or any other {@code String} not equal to {@code "true"} ignoring case), {@code null}:
+     *         Encryption with a customer-managed encryption key is disabled. This implies that managed disks will be subject to the default Storage Service
+     *         Encryption with a platform managed key.</li>
+     *     </ul>
+     * </p>
+     *
+     * If enabled, the encryption key to use is determined by the disk encryption set specified by {@link #DISK_ENCRYPTION_SET_ID}. In particular,
+     * {@link #VOLUME_ENCRYPTION_KEY_TYPE} and {@link #VOLUME_ENCRYPTION_KEY_ID} are currently ignored for managed disk encryption.
+     *
+     * @see #DISK_ENCRYPTION_SET_ID
+     * @see #VOLUME_ENCRYPTION_KEY_ID
+     * @see #VOLUME_ENCRYPTION_KEY_TYPE
+     * @see #putParameter(String, Object)
+     * @see Boolean#parseBoolean(String)
+     */
+    public static final String MANAGED_DISK_ENCRYPTION_WITH_CUSTOM_KEY_ENABLED = "encryptionWithCustomKey";
+
+    /**
+     * Key of the dynamic parameter denoting the Resource ID of the disk encryption set to be used for managed disk encryption. Relevant and required only
+     * when {@link #MANAGED_DISK_ENCRYPTION_WITH_CUSTOM_KEY_ENABLED} is {@code true}. Its value will be ignored otherwise.
+     *
+     * <p>
+     *     When set, the value shall be a nonempty {@link String} containing the Resource ID of the disk encryption set.
+     * </p>
+     *
+     * @see #MANAGED_DISK_ENCRYPTION_WITH_CUSTOM_KEY_ENABLED
+     * @see #putParameter(String, Object)
+     */
+    public static final String DISK_ENCRYPTION_SET_ID = "diskEncryptionSetId";
+
+    public AzureInstanceTemplate(String flavor, String groupName, Long privateId, Collection<Volume> volumes, InstanceStatus status,
+            Map<String, Object> parameters, Long templateId, String imageId) {
+        super(flavor, groupName, privateId, volumes, status, parameters, templateId, imageId);
+    }
+
+}
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
@@ -23,6 +23,7 @@ import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.microsoft.azure.CloudException;
 import com.microsoft.azure.PagedList;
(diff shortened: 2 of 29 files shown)


core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
```diff
@@ -58,14 +60,26 @@ public GcpInstanceTemplateV1Parameters convert(GcpInstanceTemplateV4Parameters s
...
         AzureInstanceTemplateV4Parameters response = new AzureInstanceTemplateV4Parameters();
         response.setEncrypted(source.getEncrypted());
+        initEncryption(response, environment);
         response.setManagedDisk(source.getManagedDisk());
         response.setPrivateId(source.getPrivateId());
         return response;
     }
 
+    private void initEncryption(AzureInstanceTemplateV4Parameters response, DetailedEnvironmentResponse environment) {
+        // FIXME Fetch DES ID from environment; environment can be null!
```
bergerdenes, line 73: environment can not be null as your outer converter checks for that with the getIfNotNull

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -284,16 +287,25 @@ public void deleteBlobInStorageContainer(String resourceGroup, String storageNam
...
+        Disk.DefinitionStages.WithCreate withCreate = azure.disks().define(diskName)
                 .withRegion(Region.findByLabelOrName(region))
                 .withExistingResourceGroup(resourceGroupName)
                 .withData()
                 .withSizeInGB(diskSize)
                 .withTags(tags)
-                .withSku(convertAzureDiskTypeToDiskSkuTypes(diskType))
-                .create();
+                .withSku(convertAzureDiskTypeToDiskSkuTypes(diskType));
+        if (diskEncryptionSetId != null) {
```
bergerdenes, line 300: extract this please

bergerdenes (changes requested):
updating ARM version from 2018-04-01 to 2019-07-01 can be solved by updating the `component` table in `cbdb` like this way: `UPDATE component c
SET attributes = <some clever string manipulation>
FROM stack s
WHERE c.stack_id = s.id
AND c.componenttype = 'STACK_TEMPLATE'
AND s.terminated IS NULL;`

doktoric (changes requested):

cloud-azure/src/main/resources/templates/arm-v2.ftl
```diff
@@ -293,16 +293,17 @@
                    }
                  },
                  {
-                   "apiVersion": "2018-04-01",
+                   "apiVersion": "2019-07-01",
```
doktoric, line 296: I guess this does not cause any problem but I need to ask :)

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
```diff
@@ -58,14 +60,26 @@ public GcpInstanceTemplateV1Parameters convert(GcpInstanceTemplateV4Parameters s
...
+        AzureEncryptionV4Parameters encryption = new AzureEncryptionV4Parameters();
+//        encryption.setDiskEncryptionSetId("/subscriptions/3ddda1c7-d1f5-4e7b-ac81-0523f483b3b3/resourceGroups/lrodek-daily-system-rg/providers/" +
```
doktoric, line 75: can we remove this?

cloud-azure/src/main/resources/templates/arm-v2.ftl
lajosrodek: Here are the docs for these API versions:
* [2018-04-01](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2018-04-01/virtualmachines?tabs=json)
* [2019-07-01](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-07-01/virtualmachines?tabs=json)

The only incompatible change is this `identity.identityIds` vs `identity.userAssignedIdentities` switch. There were a few additional properties added here and there, but all of them seem optional. In any case, I did not encounter any problems with DL & DH launch.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
lajosrodek: Surely will be removed, once https://jira.cloudera.com/browse/CB-11874 gets resolved that exposes the new field inside `DetailedEnvironmentResponse`.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
lajosrodek: Correct, the code has been updated to reflect this.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
lajosrodek: Done, see `setupDiskEncryptionWithDesIfNeeded()`.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AzureInstanceTemplate.java
```diff
@@ -0,0 +1,55 @@
...
+     * If enabled, the encryption key to use is determined by the disk encryption set specified by {@link #DISK_ENCRYPTION_SET_ID}. In particular,
+     * {@link #VOLUME_ENCRYPTION_KEY_TYPE} and {@link #VOLUME_ENCRYPTION_KEY_ID} are currently ignored for managed disk encryption.
+     *
+     * @see #DISK_ENCRYPTION_SET_ID
+     * @see #VOLUME_ENCRYPTION_KEY_ID
+     * @see #VOLUME_ENCRYPTION_KEY_TYPE
+     * @see #putParameter(String, Object)
+     * @see Boolean#parseBoolean(String)
+     */
+    public static final String MANAGED_DISK_ENCRYPTION_WITH_CUSTOM_KEY_ENABLED = "encryptedWithCustomKey";
```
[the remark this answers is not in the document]
lajosrodek: Fixed.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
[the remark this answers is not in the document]
lajosrodek: Fixed.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
lajosrodek: Please note that there are still some FIXME stuffs that will be cleaned up before merging.

lajosrodek:
@bergerdenes  Thanks for the hints on ARM API version update in the DB. I think this feature need not work with pre-existing clusters (I mean we have no intention to make it available to them), so hopefully no DB update is required this time. 😃

bergerdenes (approved):

doktoric (approved):

lajosrodek:
Depends on #10406.

PR #10308, opened 2021-03-19
CB-11153: Use encryptionKeyUrl stored in CB vault to create DES.

juhi-09:
CB-11153: Use encryptionKeyUrl stored in CB vault to create DES.
    As part of this PR - a new state is added to both of `EnvCreationFlowConfig` and `EnvDeleteFlowConfig`. These new states `ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_STARTED_STATE` and `ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_STARTED_STATE` creates and deletes the DiskEncryptionSet(DES) for Azure cloudPlatform respectively. Methods to create and delete DES will be added in subsequent PRs of CB-11153.
    For more details, Please refer CB-9920
    
Screenshot 2021-03-23 at 5 23 19 PM

PR #10372, opened 2021-03-29
CB-11872 Implement ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_STARTED_STATE - to create the DES for Azure

juhi-09:
Note: The branch name differs from the JIRA - this is done in order to keep the review comments
CB-11872 Implement ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_STARTED_STATE - to create the DES for Azure.
This PR implements the state of creation Encryption Resources for Azure cloud - specifically for disk encryption with CMK.

     1. Check if singleResourceGroup - create Disk encryption set(DES) in that resourceGroup.
      `vaultResourceGroupName` is required to attach access policies to the keyVault for the created DES resource.
         For singleRG - it should be same as the RG provided in create environment request.
         For multi-RG, throw IllegalArgumentException error - "Customer Managed Key Encryption for managed Azure disks is supported for single resource group only." - this will be resolved/fixed as part of CB-11927
     2. Create DES in Azure cloud using encryptionKeyUrl.
     3. Update the Vault access policies for the DES created.
     4. Store DES in environment_parameters table as "disk_encryption_set_id".

Disk Encryption Set created with tags - 

Screenshot 2021-04-12 at 7 34 47 PM

Entries made in resource table for environment - 

Screenshot 2021-04-12 at 7 32 58 PM

Entries made in environment_parameters table - 

Screenshot 2021-04-12 at 7 36 09 PM


environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -78,6 +116,47 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+            CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+            AzureClient azureClient = azureClientService.getClient(cloudCredential);
+            String region = environment.getRegionSet().stream().findFirst().get().getName();
+            Map<String, String> tags = Collections.unmodifiableMap(environment.getEnvironmentTags().getUserDefinedTags());
+            ResourceGroup resourceGroup;
+            String resourceGroupNameForCreation = azureUtils.generateResourceGroupNameByNameAndId(
+                    String.format("%s-", environment.getName()),
+                    UUID.randomUUID().toString());
+            LOGGER.debug("Creating resource group {}", resourceGroupNameForCreation);
+            resourceGroup = azureClient.createResourceGroup(resourceGroupNameForCreation, region, tags);
```
lajosrodek, line 154: The RG name generation logic shall be moved over to `cloud-azure`, presumably into `AzureEncryptionResources` or into some new service invoked therein. The `environment.getName()` thing is the "base name" I mentioned above.

By the way, you will have to generate the DES name in a similar way, using the base name, adding some suffix (e.g. `-des`) then appending the UUID like for the RG name here.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -78,6 +116,47 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+        return resourceGroupUsagePattern.isSingleResourceGroup();
+    }
+
+    private Object getOrCreateResourceGroup(EnvironmentDto environmentDto, Environment environment) {
+        Optional<String> resourceGroupName = getAzureResourceGroupName(environmentDto, environment);
+        if (resourceGroupName.isPresent()) {
+            return resourceGroupName;
+        } else {
+            CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+            AzureClient azureClient = azureClientService.getClient(cloudCredential);
```
lajosrodek, line 146: Not to be used here. Belongs to `cloud-azure`.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -2,17 +2,36 @@
...
+import java.util.regex.Pattern;
+
+import javax.inject.Inject;
+
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import com.microsoft.azure.CloudException;
```
lajosrodek, line 18: Not to be used here. Belongs to `cloud-azure`.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -2,17 +2,36 @@
...
+import com.sequenceiq.cloudbreak.cloud.azure.AzureUtils;
+import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClient;
+import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClientService;
```
lajosrodek, line 21: Not to be used here.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -42,7 +74,10 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
         EnvironmentDto environmentDto = environmentDeletionDto.getEnvironmentDto();
         EnvDeleteEvent envDeleteEvent = getEnvDeleteEvent(environmentDeletionDto);
         try {
-            environmentService.findEnvironmentById(environmentDto.getId()).ifPresent(this::deleteEncryptionResources);
+            environmentService.findEnvironmentById(environmentDto.getId()).ifPresent(environment -> {
+                String diskEncryptionSetUrl = environment.getEnvironmentEncryptionParameters();
```
lajosrodek, line 78: It is a Resource ID, not an URL. But its placement is wrong anyway as mentioned above.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+            LOGGER.error("Failed to delete diskEncryptionSet");
+        }
+        if (!isSingleResourceGroup(environmentDto)) {
+            String resourceGroupName = getResourceGroupNameFromDES(diskEncryptionSetUrl);
+            if (StringUtils.isNotEmpty(resourceGroupName)) {
+                deleteAzureResourceGroup(resourceGroupName, environment);
+            } else {
+                LOGGER.error("Failed to parse Encryption parameters resource group");
+            }
+        }
```
lajosrodek, line 108: The DES + RG deletion shall be considered a single operation at the `cloud-api` level. This is an implementation detail. Both procedures have to be executed in the actual Azure impementation class, i.e. in `AzureEncryptionResources` or some service called therein.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -58,10 +88,18 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
-                            TO DO - create the Disk Encryption Set and save environment.
-                            */
+                        LOGGER.debug("Validating if encryptionKeyUrl exists in Azure cloud.");
+                        boolean keyExistsInCloud = environmentEncryptionService.validateEncryptionParameters(encryptionKeyUrl);
+                        if (keyExistsInCloud) {
+                            LOGGER.debug("{} exists in Aure cloud, Creating DiskEncryptionSet for environment.", encryptionKeyUrl);
+
+                            String resourceGroupName = (String) getOrCreateResourceGroup(environmentDto, environment);
+                            String diskEncryptionSetUrl = environmentEncryptionService.createEncryptionSet(encryptionKeyUrl, resourceGroupName);
+                            environment.setEnvironmentEncryptionParameters(diskEncryptionSetUrl);
```
lajosrodek, line 98: The placement and name of this property is misleading and unclear as mentioned above.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+    }
+
+    private String getResourceGroupNameFromDES(String diskEncryptionSetUrl) {
+        String regexString = Pattern.quote("resourceGroups/") + "(.*?)" + Pattern.quote("/providers");
```
lajosrodek, line 112: Is the usage of `Pattern.quote()` justified here? Do those strings contain any characters that need escaping in the regex?

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+            String resourceGroupName = getResourceGroupNameFromDES(diskEncryptionSetUrl);
+            if (StringUtils.isNotEmpty(resourceGroupName)) {
+                deleteAzureResourceGroup(resourceGroupName, environment);
+            } else {
+                LOGGER.error("Failed to parse Encryption parameters resource group");
+            }
+        }
+    }
+
+    private String getResourceGroupNameFromDES(String diskEncryptionSetUrl) {
```
lajosrodek, line 111: Rename `diskEncryptionSetUrl` to `diskEncryptionSetId` or something similar. It is a Resource ID string.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+        String regexString = Pattern.quote("resourceGroups/") + "(.*?)" + Pattern.quote("/providers");
+        Pattern pattern = Pattern.compile(regexString);
```
lajosrodek, line 113: The above pattern is static, especially without the `Pattern.quote()` that are not needed here in my opinion. So the `Pattern` instance can be made a constant field in the class, instead of being instantiated here.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+        Matcher matcher = pattern.matcher(diskEncryptionSetUrl);
+
+        if (matcher.find()) {
+            return matcher.group(1);
+        } else {
+            return "";
+        }
```
lajosrodek, line 120: Perhaps better to throw something like an `IllegalStateException` or `IllegalArgumentException`. Not having a match means that that the Resource ID is malformed, the RG part is mandatory.

environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -57,10 +92,63 @@ public void accept(Event<EnvironmentDeletionDto> environmentDtoEvent) {
...
+            LOGGER.debug("Deleting Encryption parameters resource group");
+            CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+            AzureClient azureClient = azureClientService.getClient(cloudCredential);
+            if (resourceGroupExists(azureClient, resourceGroupName)) {
+                azureClient.deleteResourceGroup(resourceGroupName);
+            }
+        } catch (CloudException e) {
+            throw azureUtils.convertToCloudConnectorException(e, "Encryption parameters resource group deletion");
+        }
     }
```
lajosrodek, line 152: Belongs to `cloud-azure`.

lajosrodek (changes requested):
This is still in a very immature and incomplete state, so apologies in advance for the many comments. 😃

bergerdenes:
Is this a draft PR? If so, please change the type to that.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/EncryptionParameters.java
```diff
@@ -0,0 +1,15 @@
...
+import org.slf4j.LoggerFactory;
+
+public interface EncryptionParameters extends CloudPlatformAware {
+    Logger LOGGER = LoggerFactory.getLogger(EncryptionParameters.class);
+
+    String createEncryptionParameters(String encryptionKeyUrl);
+
+    boolean deleteEncryptionParameters(String encryptionSet);
+
+    boolean validateEncryptionParameters(String encryptionKeyUrl);
```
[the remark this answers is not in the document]
juhi-09: yes, `validateEncryptionParameters` is suppose to validate the encryption resources. Renaming it to `validateEncryptionResources`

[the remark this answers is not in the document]
lajosrodek: It should be more concrete. `validateEncryptionResources` is too wide, it is like saying `validateThings`. 😃 Please find a better name. `validateEncryptionKey` sounds better, and it should do nothing more than checking if the provided key (& version, depending on whatever the `encryptionKeyId` parameter specifies) exists.

`createEncryptionParameters` and `deleteEncryptionParameters` clearly deal with disk encryption (and nothing else), so why not call them `createDiskEncryptionResource` / `deleteDiskEncryptionResource`? Or even better, despite my earlier advice of avoiding cloud provider specific terminology, we can make an exception here and use DiskEncryptionSet; "resource" is simply too general and does not say anything about the methods purpose. As I see, `PlatformResources.resourceGroups()` already violated the above rule, even resource groups do not make much sense outside Azure.

lajosrodek:
@juhi-09  Will check the thing in the afternoon. Till then, I have one comment regarding the commit message. The commit title (first line before the line break) is the one that is displayed in the commit history. Currently it shows "Note: The branch name differs from the JIRA - this is done in order t… ". Could you please remove that whole comment so that the "CB-11872 Implement ..." line becomes the title?

The note regarding the different branch name must not be part of the commit message, for the commit itself will be part of multiple branches eventually (`master` and subsequent release branches). You can mention that instead here in the PR, that is where it makes sense.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public CreatedEncryptionResources createEncryptionResources(EnvironmentDto environmentDto, Environment environment) {
+        EncryptionResources encryptionResources = getCloudConnector(environment.getCloudPlatform());
+        return encryptionResources.createDiskEncryptionSet(createEncryptionResourcesCreationRequest(environmentDto));
+    }
+
+    private EncryptionResources getCloudConnector(String cloudPlatform) {
+        CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(Platform.platform(cloudPlatform), Variant.variant(cloudPlatform));
+        return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant).encryptionResources())
+                .orElseThrow(() -> new NotFoundException("Encryption resources component not found!"));
```
lajosrodek, line 47: Not sure but the general CB `NotFoundException` might not be the best here. I would rather suggest creating a dedicated exception type `EncryptionResourcesNotFoundException` (direct copy-paste of [NetworkConnectorNotFoundException](https://github.com/hortonworks/cloudbreak/blob/master/environment/src/main/java/com/sequenceiq/environment/network/NetworkConnectorNotFoundException.java)) in this package and throwing that here.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,78 @@
...
+        this.credentialToCloudCredentialConverter = credentialToCloudCredentialConverter;
+        this.cloudPlatformConnectors = cloudPlatformConnectors;
+    }
+
+    public CreatedEncryptionResources createEncryptionResources(EnvironmentDto environmentDto, Environment environment) {
+        EncryptionResources encryptionResources = getCloudConnector(environment.getCloudPlatform());
+        return encryptionResources.createDiskEncryptionSet(createEncryptionResourcesCreationRequest(environmentDto));
+    }
+
+    private EncryptionResources getCloudConnector(String cloudPlatform) {
```
lajosrodek, line 44: Perhaps rename the method to `getEncryptionResources` (or `getCloudConnectorEncryptionResources` if you prefer so).

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -58,10 +65,16 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+                                environment);
+                        if (!Objects.isNull(createdEncryptionResources)) {
+                            AzureParameters azureParameters = (AzureParameters) environment.getParameters();
+                            azureParameters.setDiskEncryptionSetId(createdEncryptionResources.getEncryptionResourceId());
+                            environmentService.save(environment);
+                        } else {
+                            LOGGER.error("Failed to create the Disk Encryption set for environment {}.", environment.getName());
+                        }
                     } else {
                         LOGGER.debug("Environment {} has not requested for SSE with CMK.", environment.getName());
```
lajosrodek, line 79: I think this can be at INFO level.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
```diff
@@ -36,6 +36,9 @@
     @SecretValue
     private Secret encryptionKeyUrl = Secret.EMPTY;
```
lajosrodek, line 37: Huh, I just noticed the column name is `encryption_keyUrl`. It probably should be called `encryption_key_url` to be consistent with the rest of this class. The only issue is backward compatibility; has this column addition been already released?

To answer my question: The column was added in #10177 (CB  2.40), and the `implements AccountIdAwareResource` part was fixed in #10308 (CB  2.41). Looks like there is no way to remove / rename that column anymore. 😞 

CC @doktoric , @biharitomi, @topolyai5, @bergerdenes

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -58,10 +65,16 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
-                            */
+                        LOGGER.info("Creating Encryption resources for environment.");
+                        CreatedEncryptionResources createdEncryptionResources = environmentEncryptionService.createEncryptionResources(environmentDto,
+                                environment);
+                        if (!Objects.isNull(createdEncryptionResources)) {
+                            AzureParameters azureParameters = (AzureParameters) environment.getParameters();
+                            azureParameters.setDiskEncryptionSetId(createdEncryptionResources.getEncryptionResourceId());
+                            environmentService.save(environment);
+                        } else {
+                            LOGGER.error("Failed to create the Disk Encryption set for environment {}.", environment.getName());
```
lajosrodek, line 76: Perhaps change `set` to `Set`.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
```diff
@@ -36,6 +36,9 @@
...
 
+    @Column(name = "disk_encryption_set_id")
+    private String diskEncryptionSetId;
+
```
lajosrodek, line 41: As mentioned earlier, we probably will have to add one more field & DB column here for the Key Vault RG name or Resource ID.

lajosrodek (changes requested):
Sorry, a few dozens more comments. Mostly concerning naming and logic placement. There is one blocking issue, namely the lack of the Key Vault RG name / Resource ID details in the environment creation request.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,124 @@
...
+
+        return diskEncryptionSet;
+    }
+
+    private void grantAccessPolicyToDes(ApplicationTokenCredentials applicationTokenCredentials, String principalId, String subscriptionId,
+            String vaultName, String resourceGroupName) {
+        Azure azure = Azure.authenticate(applicationTokenCredentials).withSubscription(subscriptionId);
+        azure.vaults().getByResourceGroup(resourceGroupName, vaultName).update().defineAccessPolicy().forObjectId(principalId)
+                .allowKeyPermissions(List.of(KeyPermissions.WRAP_KEY, KeyPermissions.UNWRAP_KEY, KeyPermissions.GET)).attach().apply();
+    }
```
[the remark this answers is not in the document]
lajosrodek: One more to this: For better readability, please put each chained invocation on a separate line like this:
```
azure.vaults()
    .getByResourceGroup(resourceGroupName, vaultName)
    .update()
...
```
Of course the indentation might look different depending on what the IDE chooses.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
juhi-09: Do we really need to store Key Vault RG name or Resource ID ? 
`Key Vault RG name` is required during creation flow in case of multi-RG. Not sure if we would need it going forward.
If we decide to store it, would it make sense to (fetch and) store it during EnvironmentCreationDto from EnvironmentRequest ?

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
juhi-09: Maybe to drop column and add updated one?

lajosrodek: Dropping / renaming columns in general is not allowed for backwards compatibility. That is what the  schema-compatibility-test step is checking for. This case, however, is very special as the new column added earlier has never been actually in use so far. I let CB folks decide on its fate.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
lajosrodek: Yes, I agree that it is currently needed only once during the DES creation flow step, but storing it would still make sense to keep track of all user input given during environment creation.

Please note, however, that the Key Vault RG (or Resource ID) is always needed, irrespective of Single-RG vs Multi-RG. The Key Vault may be located in a totally different RG even when using Single-RG; that is what my setup looks like, and customers may decide to follow this approach as well in order to separate persistent resources (e.g. Key Vault, Storage Account, MSI) from volatile ones like clusters.

+1 on setting & persisting this during the initial EnvironmentRequest -> DTO -> AzureParameters conversion.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public CreatedDiskEncryptionSet createDiskEncryptionSet(DiskEncryptionSetCreationRequest diskEncryptionSetCreationRequest) {
+        AzureClient azureClient = azureClientService.getClient(diskEncryptionSetCreationRequest.getCloudCredential());
+        String encryptionKeyUrl = diskEncryptionSetCreationRequest.getEncryptionKeyUrl();
+        String envName = diskEncryptionSetCreationRequest.getEnvName();
+        String location = diskEncryptionSetCreationRequest.getRegion().getRegionName();
+        Map<String, String> tags = diskEncryptionSetCreationRequest.getTags();
+        String vaultName = ENCRYPTION_KEY_URL_PATTERN.matcher(encryptionKeyUrl).group(1);
+
+        String desResourceGroupName = null;
```
lajosrodek, line 54: Please remove the initializer. The variable will be definitely assigned after the `if` / `else` block.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,98 @@
...
+        String vaultResourceGroupName = null;
+        if (diskEncryptionSetCreationRequest.isSingleResourceGroup()) {
+            desResourceGroupName = diskEncryptionSetCreationRequest.getResourceGroup();
+            vaultResourceGroupName = desResourceGroupName;
+        } else {
+            desResourceGroupName = (String) createResourceGroup(azureClient, envName, location, tags);
+            vaultResourceGroupName = azureClient.getVaultResourceGroup(vaultName);
```
lajosrodek, line 61: As mentioned elsewhere, the Key Vault RG / Resource ID always has to be determined (or supplied by the user via the environment create request), irrespective of Single-RG vs Multi-RG. Single-RG only means that CB will create all the resources in a dedicated pre-existing RG, but it does not say anything about the location of other resources provided & managed by the user (e.g. Storage Account, MSI, Key Vault).

Therefore, the statement of line 61 can be moved here, and line 58 shall be deleted.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -837,4 +847,31 @@ private void handleAuthException(Runnable function) {
...
+            String resourceGroupName, String sourceVaultId, Map<String, String> tags) {
+        DiskEncryptionSetInner desIn = getDiskEncryptionSetInner(sourceVaultId, encryptionKeyUrl, location, tags);
+        DiskEncryptionSetsInner dSetsIn = azureClientCredentials.getComputeManager().inner().diskEncryptionSets();
+        return dSetsIn.createOrUpdate(resourceGroupName, diskEncryptionSetName, desIn);
+    }
+
+    public String getVaultResourceGroup(String vaultName) {
+        VaultsInner inner = (VaultsInner) ((GroupableResourcesImpl) azureClientCredentials.getKeyVaultManager().vaults()).inner();
+        PagedList<VaultInner> vaultsInSubscription = inner.listBySubscription();
+        return vaultsInSubscription.stream().filter(vInner -> vInner.name().equals(vaultName)).map(VaultInner::id).findFirst().orElse(null);
```
lajosrodek, line 875: Have you verified if this actually works as expected? In particular, `VaultInner.id()` was supposed to contain the whole Resource ID, from which the RG has to be extracted in a separate step. On the other hand, if this is indeed the Resource ID, why not return and use it directly, without extracting the RG part? Then the method name shall be changed to `getVaultResourceIdByVaultName`.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        String sourceVaultId = String.format("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.KeyVault/vaults/%s",
+                azureClient.getAzure().subscriptionId(), vaultResourceGroupName, vaultName);
```
lajosrodek, line 64: As mentioned in my comment in `AzureClient.getVaultResourceGroup()`, the current logic there possibly returns the whole Resource ID, so there is no need to do this magic here. In fact, if the output of that method is indeed the Resource ID, the string constructed here will be invalid. Could you please verify how this works locally?

lajosrodek (changes requested):
A few preliminary notes; a full review will come in 2 days.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
lajosrodek: @doktoric , @biharitomi, @topolyai5, @bergerdenes Please confirm that we can go with this workaround, given it may have performance hits. The actual impact is unknown at the moment.

lajosrodek: UPDATE: After some thinking, extra wide permissions (possibly at subscription scope?) might be needed for the above `listBySubscription()` call. It first needs to be confirmed if this is indeed the case. If it turns out to be valid, this approach will have to be dropped for infeasibility since customers will typically not grant subscription-level access.

doktoric: I think the real problem here is the required permission right  @lajosrodek ?

doktoric: @juhi-09 please do the api change for introducing the vault resource group name. We dont want to request full acount permission from customers.

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
doktoric: @juhi-09 create a new column with name encryption_key_url and then in the next CB release you can drop the current column.

juhi-09: Adding a new column `encryption_key_url` as part of this PR. Will drop `encryption_keyUrl` in future release

environment/src/main/java/com/sequenceiq/environment/parameters/dao/domain/AzureParameters.java
juhi-09: To support multi-RG, KeyVaultResourceId will be added as part of CB-11927.
As a workaround, using single resource group and keyVault should be present in the same resource group as mentioned in environment creation request.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,84 @@
...
+    public EncryptionResources getEncryptionResources(String cloudPlatform) {
+        CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(Platform.platform(cloudPlatform), Variant.variant(cloudPlatform));
+        return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant))
+                .map(CloudConnector<Object>::encryptionResources)
+                .orElseThrow(() -> new EncryptionResourcesNotFoundException("No Encryption resources component found for cloud platform: " + cloudPlatform));
+    }
+
+    public DiskEncryptionSetCreationRequest createEncryptionResourcesCreationRequest(EnvironmentDto environment) {
+        CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+        DiskEncryptionSetCreationRequest.Builder builder = new DiskEncryptionSetCreationRequest.Builder()
```
lajosrodek, line 58: Related to the "DES flow idempotency": The new ID field has to be populated in the `DiskEncryptionSetCreationRequest` in an idempotent & deterministic way (while still resulting in a unique ID suffix for the DES & RG resources). TODO Needs some elaboration, I will update this comment later.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
lajosrodek: Ping, make it INFO please.

environment/src/main/resources/schema/app/20210401084656_CB-11872_Implement_ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_STARTED_STATE_-_to_create_the_DES_for_Azure.sql
```diff
@@ -0,0 +1,10 @@
+-- // CB-11872 Implement ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_STARTED_STATE - to create the DES for Azure
```
lajosrodek, line 1: Please do not forget to drop this script after merging #10406.

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private EnvironmentEncryptionService underTest;
+
+    @BeforeEach
+    void setUp() {
+        underTest = new EnvironmentEncryptionService(
+                credentialToCloudCredentialConverter,
+                cloudPlatformConnectors,
+                environmentTagProvider
+        );
```
lajosrodek, line 66: Hint. No need to instantiate this way. Just add `@InjectMocks` to `underTest` and Mockito will instantiate the class before each test method using either of three ways: constructor injection, field injection, property setter injection. The constructor will win this time, automatically receiving the correct mock instances from the above fields.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,84 @@
...
+        this.cloudPlatformConnectors = cloudPlatformConnectors;
+        this.environmentTagProvider = environmentTagProvider;
+    }
+
+    public CreatedDiskEncryptionSet createEncryptionResources(EnvironmentDto environmentDto, Environment environment) {
+        EncryptionResources encryptionResources = getEncryptionResources(environment.getCloudPlatform());
+        return encryptionResources.createDiskEncryptionSet(createEncryptionResourcesCreationRequest(environmentDto));
+    }
+
+    public EncryptionResources getEncryptionResources(String cloudPlatform) {
```
lajosrodek, line 49: Please make this either `private` or package private. The latter makes sense if you want to test this method logic in isolation. If making it package private, please also add the `@VisibleForTesting` annotation to the method.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,84 @@
...
+        CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(Platform.platform(cloudPlatform), Variant.variant(cloudPlatform));
+        return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant))
+                .map(CloudConnector<Object>::encryptionResources)
+                .orElseThrow(() -> new EncryptionResourcesNotFoundException("No Encryption resources component found for cloud platform: " + cloudPlatform));
+    }
+
+    public DiskEncryptionSetCreationRequest createEncryptionResourcesCreationRequest(EnvironmentDto environment) {
```
lajosrodek, line 56: Like for `getEncryptionResources()`, this should be either `private` or package private, and bear the `@VisibleForTesting` annotation in the latter case.

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+                                        .withResourceGroupUsagePattern(ResourceGroupUsagePattern.USE_SINGLE)
+                                        .withName("dummyRG").build())
+                                .build())
+                        .build())
+                .withTags(new EnvironmentTags(new HashMap<>(), new HashMap<>()))
+                .build();
+        Environment env = new Environment();
+        env.setCloudPlatform("AZURE");
+        underTest.createEncryptionResources(environmentDto, env);
+        verify(encryptionResources, times(1)).createDiskEncryptionSet(any());
```
lajosrodek, line 121: Please add a stubbing (`when(..)`) logic for the `encryptionResources.createDiskEncryptionSet()` invocation and check if the response of `underTest.createEncryptionResources()` is as expected. That way there is no need for the `verify()` check.

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandlerTest.java
```diff
@@ -137,6 +157,40 @@ void acceptTestEnvironmentShouldNotBeUpdatedWhenEncryptionKeyUrlIsEmpty() {
...
+        when(environmentService.findEnvironmentById(ENVIRONMENT_ID)).thenReturn(Optional.of(environment));
+        CreatedDiskEncryptionSet createdDiskEncryptionSet = new CreatedDiskEncryptionSet.Builder()
+                .withDiskEncryptionSetId("dummmy-set-id")
+                .build();
+        when(environmentEncryptionService.createEncryptionResources(any(), any())).thenReturn(createdDiskEncryptionSet);
+
+        underTest.accept(environmentDtoEvent);
+
+        verify(eventSender).sendEvent(baseNamedFlowEventCaptor.capture(), headersArgumentCaptor.capture());
+        verify(environmentService, times(1)).save(any());
```
lajosrodek, line 174: Instead of simply verifying the act of the `save()` invocation, please capture the method argument and then check of the DESid inside the capture `Environment` matches the one returned above.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/encryption/DiskEncryptionSetCreationRequest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.sequenceiq.cloudbreak.cloud.CloudPlatformAware;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
+import com.sequenceiq.cloudbreak.cloud.model.Platform;
+import com.sequenceiq.cloudbreak.cloud.model.Region;
+import com.sequenceiq.cloudbreak.cloud.model.Variant;
+
+public class DiskEncryptionSetCreationRequest implements CloudPlatformAware {
```
lajosrodek, line 12: Related to the "DES flow idempotency": A new ID field has to be added. I will update this comment later. TODO exact name & type.

lajosrodek (changes requested):
Some further comments. Please hold on with the "DES flow idempotency" comments until I update them. Some additional comments will also come tomorrow regarding `CloudResource` persisting. 😃

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -0,0 +1,106 @@
...
+import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClient;
+import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClientService;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
+import com.sequenceiq.cloudbreak.cloud.model.Platform;
+import com.sequenceiq.cloudbreak.cloud.model.Region;
+import com.sequenceiq.cloudbreak.cloud.model.Variant;
+import com.sequenceiq.cloudbreak.cloud.model.encryption.DiskEncryptionSetCreationRequest;
+
+@RunWith(MockitoJUnitRunner.class)
+public class AzureEncryptionResourcesTest {
```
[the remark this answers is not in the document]
juhi-09: yes, it is so because the cloud call `azureClient.createOrUpdateDiskEncryptionSet` returns `DiskEncryptionSetInner`, not sure on how to mock this?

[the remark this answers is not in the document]
lajosrodek: Not sure what you exactly mean. Neither `AzureClient` nor `DiskEncryptionSetInner` are `final` classes, `AzureClient.createOrUpdateDiskEncryptionSet()` is a non-final instance method, so everything involved can be mocked. You can try adding a stubbing (`when(..)`) for the `createOrUpdateDiskEncryptionSet()` invocation.

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,90 @@
...
+        return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant))
+                .map(CloudConnector<Object>::encryptionResources)
+                .orElseThrow(() -> new EncryptionResourcesNotFoundException("No Encryption resources component found for cloud platform: " + cloudPlatform));
+    }
+
+    @VisibleForTesting
+    DiskEncryptionSetCreationRequest createEncryptionResourcesCreationRequest(EnvironmentDto environment) {
+        CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+        String uniqueId = Crn.safeFromString(environment.getResourceCrn()).getResource();
+        DiskEncryptionSetCreationRequest.Builder builder = new DiskEncryptionSetCreationRequest.Builder()
```
lajosrodek, line 63: Please populate the new field `cloudContext` with a `CloudContext` instance that you can build like the example here in [NetworkCreationRequestFactory](https://github.com/hortonworks/cloudbreak/blob/master/environment/src/main/java/com/sequenceiq/environment/network/service/NetworkCreationRequestFactory.java#L150-L161).

lajosrodek (changes requested):
3 asks regarding `CloudResource` persisting.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,107 @@
...
+                diskEncryptionSetCreationRequest.getId());
+        DiskEncryptionSetInner createdSet;
+        LOGGER.info("Checking if Disk Encryption Set {} exists on cloud Azure", diskEncryptionSetName);
+        createdSet = azureClient.getDiskEncryptionSet(desResourceGroupName, diskEncryptionSetName);
+        if (Objects.isNull(createdSet)) {
+            LOGGER.info("Creating Disk Encryption Set {}", diskEncryptionSetName);
+            createdSet = azureClient.createOrUpdateDiskEncryptionSet(diskEncryptionSetName, encryptionKeyUrl,
+                    diskEncryptionSetCreationRequest.getRegion().getRegionName(), desResourceGroupName, sourceVaultId,
+                    diskEncryptionSetCreationRequest.getTags());
+        }
```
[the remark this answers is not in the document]
lajosrodek: Note: Steps 2. & 4. are to be carried out only when `createdSet != null`. Furthermore, nullity checks are preferred using the `==` / `!=` relational operators; `Objects.isNull()` is only employed inside stream operations involving predicates, like `Stream.filter()`.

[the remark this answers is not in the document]
lajosrodek: Updated step 2. above.

lajosrodek (changes requested):
Further asks:

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandler.java
```diff
@@ -53,19 +61,32 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+                                LOGGER.info("Initializing Encryption resources for environment.");
+                                CreatedDiskEncryptionSet createdDiskEncryptionSet = environmentEncryptionService.createEncryptionResources(environmentDto,
+                                        environment);
+                                if (!Objects.isNull(createdDiskEncryptionSet) && StringUtils.isNotEmpty(createdDiskEncryptionSet.getDiskEncryptionSetId())) {
+                                    AzureParameters azureParameters = (AzureParameters) environment.getParameters();
+                                    azureParameters.setDiskEncryptionSetId(createdDiskEncryptionSet.getDiskEncryptionSetId());
+                                    environment.setStatus(EnvironmentStatus.ENVIRONMENT_ENCRYPTION_RESOURCES_INITIALIZED);
+                                } else {
+                                    environment.setStatus(EnvironmentStatus.ENVIRONMENT_ENCRYPTION_RESOURCES_INITIALIZATION_FAILED);
+                                    LOGGER.error("Failed to initialize the Disk Encryption Set for environment {}.", environment.getName());
```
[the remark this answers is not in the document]
juhi-09: Shall we retry and eventually fail the environment creation with error after some 'x' retries?

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandlerTest.java
```diff
@@ -137,6 +158,42 @@ void acceptTestEnvironmentShouldNotBeUpdatedWhenEncryptionKeyUrlIsEmpty() {
...
+                .withDiskEncryptionSetId("dummmy-set-id")
+                .build();
+        when(environmentEncryptionService.createEncryptionResources(any(), any())).thenReturn(createdDiskEncryptionSet);
+
+        underTest.accept(environmentDtoEvent);
+
+        verify(eventSender).sendEvent(baseNamedFlowEventCaptor.capture(), headersArgumentCaptor.capture());
+        AzureParameters azureParameters = (AzureParameters) environment.getParameters();
+        assertEquals(azureParameters.getDiskEncryptionSetId(), "dummmy-set-id");
+        assertEquals(environment.getStatus(), EnvironmentStatus.ENVIRONMENT_ENCRYPTION_RESOURCES_INITIALIZED);
```
lajosrodek, line 177: What happened to the `environmentService.save()` invocation? Could you please add a verification (with an `ArgumentCaptor`) like I asked earlier?

> Instead of simply verifying the act of the `save()` invocation, please capture the method argument and then check if the DESid inside the captured Environment matches the one returned above.

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandlerTest.java
```diff
@@ -137,6 +158,42 @@ void acceptTestEnvironmentShouldNotBeUpdatedWhenEncryptionKeyUrlIsEmpty() {
...
+        verifyEnvCreationEvent();
+    }
+
+    @Test
+    void acceptTestEnvironmentShouldNotBeUpdatedWhenCreatedDiskEncryptionSetIdIsNull() {
```
lajosrodek, line 182: As mentioned above, I do not feel it is a valid scenario to have `EnvironmentEncryptionService.createEncryptionResources()` return `null` (and with the current logic, it seems to be impossible to encounter such a situation).

lajosrodek (changes requested):
Thanks, some further comments below.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+                    .withDiskEncryptionSetId(createdSet.id())
+                    .withDiskEncryptionSetPrincipalId(createdSet.identity().principalId())
+                    .withDiskEncryptionSetLocation(createdSet.location())
+                    .withDiskEncryptionSetName(createdSet.name())
+                    .withTags(createdSet.getTags())
+                    .withDiskEncryptionSetResourceGroup(desResourceGroupName)
+                    .build();
+        } else {
+            throw new CloudConnectorException("Failed to create Disk Encryption Set.");
+        }
```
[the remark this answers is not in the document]
juhi-09: Azure SDK call either returns the object(filled(with DES id and other values)) or throws `IllegalArgumentException`.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+        return AzureConstants.PLATFORM;
+    }
+
+    @Override
+    public Variant variant() {
+        return AzureConstants.VARIANT;
+    }
+
+    @Override
+    public CreatedDiskEncryptionSet createDiskEncryptionSet(DiskEncryptionSetCreationRequest diskEncryptionSetCreationRequest) {
```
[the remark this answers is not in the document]
juhi-09: This method either returns `CreatedDiskEncryptionSet` or the exception thrown by `getOrCreateDiskEncryptionSetOnCloud` or  `IllegalArgumentException` in case of single RG.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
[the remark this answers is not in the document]
lajosrodek: OK, so this is mainly for the test case. We can leave this behavior, but it would be a really bad Azure SDK design fault if we ever encountered `null` in practice instead of getting an exception.

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandlerTest.java
lajosrodek: To line 176: It is the 3rd time I am repeating this: do not just verify the act of `save()` invocation, but also capture its actual argument & check its expected contents. Moreover, the `, times(1)` part shall be dropped, that is the default invocation count if not specified.

lajosrodek (changes requested):
One UT ask.

lajosrodek (approved):
Please ignore the above. 😄  LGTM, now let is wait for an approval from other folks.

environment/src/test/java/com/sequenceiq/environment/environment/flow/creation/handler/ResourceEncryptionInitializationHandlerTest.java
lajosrodek: Notes to line 176:

```
verify(environmentService, times(1)).save(environment);
```

1. Mockito matches method arguments using object equality based on `equals()` by default (i.e. as if `.save(eq(environment))`); see [the documentation](https://javadoc.io/static/org.mockito/mockito-core/3.9.0/org/mockito/Mockito.html#argument_matchers). This check only works because the `Environment` and `AzureParameters` entity classes do not currently override `equals()`, hence falling back to the default implementation based on reference equality. In other words: Since, the contents of `environment` is expected to change after executing `underTest.accept(environmentDtoEvent);` (notably `environment.parameters.diskEncryptionSetId` will be updated), this check will break as soon as someone adds a custom `equals()` implementation to `Environment` and `AzureParameters`. My original suggestion was to use an explicit `ArgumentCaptor` to capture the argument of the `save()` invocation, that would be immune to any changes regarding the `equals()` implementations of the above classes.
2. `, times(1)` is redundant because 1 is the default repetition count.

lajosrodek (commented):
@juhi-09  One note regarding the UT. **Update: The concern in point 1 does not really hold here, so feel free to ignore it. Point 2 is still relevant though.**

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -863,4 +871,41 @@ private void handleAuthException(Runnable function) {
...
+        return (DiskEncryptionSetInner) new DiskEncryptionSetInner()
+                .withEncryptionType(DiskEncryptionSetType.ENCRYPTION_AT_REST_WITH_CUSTOMER_KEY)
+                .withActiveKey(keyUrl)
+                .withIdentity(eSetId)
+                .withLocation(location)
+                .withTags(tags);
+    }
+
+    public DiskEncryptionSetInner getDiskEncryptionSet(String resourceGroupName, String diskEncryptionSetName) {
+        DiskEncryptionSetsInner dSetsIn = azureClientCredentials.getComputeManager().inner().diskEncryptionSets();
```
lajosrodek, line 888: Add comment before:
```
// The Disk encryption set operations are not exposed in public API, so have to rely on underlying DiskEncryptionSetInner
```

lajosrodek (commented):
Not a big deal, but could you please fix this along with anything else requested by the other folks? 😃

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+        String vaultName;
+        String vaultResourceGroupName;
+        String desResourceGroupName;
+        AzureClient azureClient = azureClientService.getClient(diskEncryptionSetCreationRequest.getCloudCredential());
+
+        Matcher matcher = ENCRYPTION_KEY_URL_PATTERN.matcher(diskEncryptionSetCreationRequest.getEncryptionKeyUrl());
+        if (matcher.matches()) {
+            vaultName = matcher.group(1);
+        } else {
+            throw new IllegalArgumentException("vaultName cannot be fetched from encryptionKeyUrl. encryptionKeyUrl should be of format - " +
```
doktoric, line 63: Do we show this to the customer? If yes then we should add the provided name into the error message and some link for azure docs as well.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+                    "'https://<vaultName>.vault.azure.net/keys/<keyName>/<keyVersion>'");
+        }
+        if (diskEncryptionSetCreationRequest.isSingleResourceGroup()) {
+            desResourceGroupName = diskEncryptionSetCreationRequest.getResourceGroup();
+            vaultResourceGroupName = desResourceGroupName;
+        } else {
+            throw new IllegalArgumentException("Customer Managed Key Encryption for managed Azure disks is supported for single resource group only.");
```
doktoric, line 70: single resource group only -> what is this means? I think we should meantion something like "if the datahub/datalake is in the same resource group where the vault is"

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+        LOGGER.info("Checking if Disk Encryption Set {} exists on cloud Azure", diskEncryptionSetName);
+        DiskEncryptionSetInner createdSet = azureClient.getDiskEncryptionSet(desResourceGroupName, diskEncryptionSetName);
+        if (createdSet == null) {
+            LOGGER.info("Creating Disk Encryption Set {}", diskEncryptionSetName);
+            createdSet = azureClient.createDiskEncryptionSet(diskEncryptionSetName, diskEncryptionSetCreationRequest.getEncryptionKeyUrl(),
+                    diskEncryptionSetCreationRequest.getRegion().getRegionName(), desResourceGroupName, sourceVaultId,
+                    diskEncryptionSetCreationRequest.getTags());
+        }
+        if (createdSet != null) {
+            CloudResource desCloudResource = CloudResource.builder()
```
doktoric, line 99: log for diskencryption set "name" exist on azure side. btw do we need to check not just the name but the resource group ?

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -0,0 +1,119 @@
...
+            return new CreatedDiskEncryptionSet.Builder()
+                    .withDiskEncryptionSetId(createdSet.id())
+                    .withDiskEncryptionSetPrincipalId(createdSet.identity().principalId())
+                    .withDiskEncryptionSetLocation(createdSet.location())
+                    .withDiskEncryptionSetName(createdSet.name())
+                    .withTags(createdSet.getTags())
+                    .withDiskEncryptionSetResourceGroup(desResourceGroupName)
+                    .build();
+        } else {
+            throw new CloudConnectorException("Failed to create Disk Encryption Set.");
```
doktoric, line 116: do we sometimes hitting this else branch ? on the top I see 
if (createdSet == null)  and if (createdSet != null) when we touch this part of the code ?

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -863,4 +871,42 @@ private void handleAuthException(Runnable function) {
     public Completable deleteGenericResourceByIdAsync(String databaseServerId) {
         return handleAuthException(() -> azure.genericResources().deleteByIdAsync(databaseServerId));
     }
+
+    private DiskEncryptionSetInner createDiskEncryptionSetInner(String sourceVaultId, String encryptionKeyUrl, String location, Map<String, String> tags) {
+        SourceVault sourceVault = new SourceVault().withId(sourceVaultId);
+        KeyVaultAndKeyReference keyUrl = new KeyVaultAndKeyReference().withKeyUrl(encryptionKeyUrl).withSourceVault(sourceVault);
+        EncryptionSetIdentity eSetId = new EncryptionSetIdentity().withType(DiskEncryptionSetIdentityType.SYSTEM_ASSIGNED);
+        return (DiskEncryptionSetInner) new DiskEncryptionSetInner()
```
doktoric, line 879: why we need this cast?

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -0,0 +1,190 @@
...
+                .withRegion(Region.region("dummyRegion"))
+                .withEnvironmentName("dummyEnvName")
+                .withEnvironmentId(1L)
+                .withSingleResourceGroup(true)
+                .withResourceGroupName("dummyResourceGroup")
+                .withTags(new HashMap<>())
+                .withEncryptionKeyUrl("wrongKeyUrl")
+                .build();
+        when(azureClientService.getClient(any())).thenReturn(azureClient);
+        assertThrows(IllegalArgumentException.class, () -> underTest.createDiskEncryptionSet(requestedSet));
```
doktoric, line 97: I think we should check the error as well like the message

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -0,0 +1,190 @@
...
+                .withResourceGroupName("dummyResourceGroup")
+                .withTags(new HashMap<>())
+                .withEncryptionKeyUrl("https://dummyVaultName.vault.azure.net/keys/dummyKeyName/dummyKeyVersion")
+                .build();
+        Subscription subscription = mock(Subscription.class);
+        when(subscription.subscriptionId()).thenReturn("dummySubscriptionId");
+        when(azureClientService.getClient(any())).thenReturn(azureClient);
+        when(azureClient.getCurrentSubscription()).thenReturn(subscription);
+        //Call to createDiskEncryptionSet is made and exception is thrown because of dummy parameters.
+        assertThrows(CloudConnectorException.class, () -> underTest.createDiskEncryptionSet(requestedSet));
```
doktoric, line 117: error message should be checked

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -0,0 +1,190 @@
...
+                .withLocation("dummyRegion")
+                .withTags(new HashMap<>());
+        Subscription subscription = mock(Subscription.class);
+        when(persistenceNotifier.notifyAllocation(any(), any())).thenReturn(new ResourcePersisted());
+        when(subscription.subscriptionId()).thenReturn("dummySubscriptionId");
+        when(azureUtils.generateDesNameByNameAndId(any(), any())).thenReturn("dummyEnvName-DES-uniqueId");
+        when(azureClientService.getClient(any())).thenReturn(azureClient);
+        when(azureClient.getCurrentSubscription()).thenReturn(subscription);
+        when(azureClient.getDiskEncryptionSet(any(), any())).thenReturn(null);
+        when(azureClient.createDiskEncryptionSet(any(), any(), any(), any(), any(), any())).thenReturn(des);
```
doktoric, line 184: I think it would be nice to add the class like any(String.class) just to make sure that if the parameters are changing we still double checking the unit tests

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
```diff
@@ -39,6 +39,7 @@
 
     ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_IN_PROGRESS("Initializing Resource Encryption settings for environment"),
     ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_IN_PROGRESS("Deleting Resource Encryption settings of environment"),
+    ENVIRONMENT_ENCRYPTION_RESOURCES_INITIALIZED("Created Resource Encryption settings for environment"),
```
doktoric, line 42: this probably requires UI update as well

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EncryptionResourcesNotFoundException.java
```diff
@@ -0,0 +1,18 @@
+package com.sequenceiq.environment.environment.encryption;
+
+import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
+
+public class EncryptionResourcesNotFoundException extends CloudbreakServiceException {
```
doktoric, line 5: this is probably requires ExceptionMapper right @topolyai5 ?

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
lajosrodek: Note: The `encryptionKeyUrl` is a secret. Such fields are sanitized / redacted in logs, so not sure what exactly to do when composing the error msg.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
lajosrodek: `.withLocation()` and `.withTags()` are actually inherited from the superclass of `DiskEncryptionSetInner`, `com.microsoft.azure.Resource`, and they return that type instead of `DiskEncryptionSetInner`. We could get away with this cast by explicitly storing the `DiskEncryptionSetInner` in a local variable (up to the `.withIdentity()` invocation) and calling `.withLocation()` and `.withTags()` on that variable separately.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
juhi-09: `azureClient.getDiskEncryptionSet` checks if the des exists with that name in provided resourceGroup.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
juhi-09: `azureClient.getDiskEncryptionSet` returns null if DES doesn't exists on cloud. If it exists it means `(createdSet != null)`. During testing with dummyParameters it was observed that `azureClient.createDiskEncryptionSet` returned `null`.
Exception is created to handle unforeseen cases of failures while creation, mostly from cloud.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
juhi-09: Updated CB-11791 to include this change

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
doktoric: Can we somehow then specify a better error message?

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EncryptionResourcesNotFoundException.java
topolyai5: Yes, mapper required

doktoric (approved):

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EncryptionResourcesNotFoundException.java
juhi-09: `CloudbreakServerExceptionMapper` exists for `CloudbreakServiceException`

PR #10373, opened 2021-03-29
CB-11379 subscriptionRequired flag for notifications

agov:
Extend CloudbreakNotification class with an optional subscriptionRequired boolean property, set based on PERSONAL_VIEW_CB_BY_RIGHT entitlement.
The flag will be used by Uluwatu to determine if notification filtering is enabled for the tenant.


foldik (approved):
lgtm

PR #10378, opened 2021-03-30
CB-11517 Create Authz Util Endpoint For List Subscription

foldik:


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -799,6 +811,20 @@ protected boolean isReadRight(String action) {
...
 
+    private List<Boolean> hasRightsBasedOnSingleCheckRightRequests(String memberCrn, List<RightCheck> rightChecks, AuthorizationClient client,
+            String requestId) {
+        return rightChecks.stream()
+                .map(check -> {
+                    try {
+                        client.checkRight(requestId, memberCrn, check.getRight(), check.getResource());
+                        return true;
+                    } catch (Exception ex) {
+                        return false;
```
horadla23, line 822: `makeCheckRightCall` has the same exception driven return statement, is it possible to reuse that?

authorization-common/src/main/java/com/sequenceiq/authorization/service/UtilAuthorizationService.java
```diff
@@ -60,23 +81,24 @@ public CheckResourceRightsV4Response getResourceRightsResult(CheckResourceRights
         Multimap<ImmutablePair<String, RightV4>, AuthorizationProto.RightCheck> resourceRightsChecks = LinkedListMultimap.create();
```
horadla23, line 81: I know that possibly I created this method, but maybe it is worth to split into multiple methods, looks hard to understand it

authorization-common/src/main/java/com/sequenceiq/authorization/service/UtilAuthorizationService.java
```diff
@@ -92,6 +114,35 @@ public CheckResourceRightsV4Response getResourceRightsResult(CheckResourceRights
         return generateResponse(resourceRightsChecks, rightCheckResultMap);
     }
 
+    public CheckRightOnResourcesV4Response getRightOnResourcesResult(CheckRightOnResourcesV4Request request) {
```
horadla23, line 117: same here, worth to split into at least 2 methods

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crns.java
```diff
@@ -0,0 +1,23 @@
+package com.sequenceiq.cloudbreak.auth.altus;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.sequenceiq.cloudbreak.auth.altus.Crn.ResourceType.MACHINE_USER;
+import static com.sequenceiq.cloudbreak.auth.altus.Crn.ResourceType.USER;
+import static com.sequenceiq.cloudbreak.auth.altus.Crn.Service.IAM;
+
+import java.util.Objects;
+
+public class Crns {
```
horadla23, line 10: are you sure we need a separate class for this one method? I think Crn class also can have it. Also the name of the class is too general especially when we see that the method is specialized for user related CRN

authorization-common/src/main/java/com/sequenceiq/authorization/service/list/AuthorizationResourceProvider.java
```diff
@@ -0,0 +1,7 @@
+package com.sequenceiq.authorization.service.list;
+
+import java.util.List;
+
+public interface AuthorizationResourceProvider {
```
horadla23, line 5: so in authorization-common we have `ResourceNameProvider`, `ResourceBasedCrnProvider`, `ResourceNameAndCrnProvider`, `ResourceCrnAndNameProvider` and now we have this, at this point we need to rethink these providers to avoid confusion

horadla23 (changes requested):
as requested, please add couple of lines in commit message about what is the change in high level and why we need this

common/src/main/java/com/sequenceiq/cloudbreak/auth/altus/Crns.java
foldik: I can put it into the Crn class.

authorization-common/src/main/java/com/sequenceiq/authorization/service/UtilAuthorizationService.java
foldik: Sure.

authorization-common/src/main/java/com/sequenceiq/authorization/service/UtilAuthorizationService.java
foldik: Will do.

horadla23 (approved):

PR #10381, opened 2021-03-30
CB-11850 Error message in case of delete node from orchestrator is mi…

horadla23:
…sleading


pdarvasi:
@horadla23 Is there a reason this remained in draft?

horadla23:
> @horadla23 Is there a reason this remained in draft?

yes, there is a blocking jira, I think we need a doc for customer about how to reenable minions and we need that in error message based on guideline of error messages. If this is enough, then we can merge this, but based on this error message, customer won't know, how to readd salt keys and restart minions

horadla23:
discussed with @sodre90, we will go with this error message for now
cc @pdarvasi @foldik

sodre90 (approved):

pdarvasi (approved):

PR #10382, opened 2021-03-30
CB-11823 - changing message due to: The ‘Standard_D4s_v3’ instance type isn’t supported by ‘AZURE’ platform

gregito:
CB-11823 - changing message due to: The ‘Standard_D4s_v3’ instance type isn’t supported by ‘AZURE’ platform


core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/template/TemplateValidator.java
```diff
@@ -80,7 +80,7 @@ public void validate(Credential credential, InstanceGroup instanceGroup, Stack s
                 }
                 if (vmType == null) {
                     validationBuilder.error(
-                            String.format("The '%s' instance type isn't supported by '%s' platform",
+                            String.format("Our platform currently not supporting the '%s' instance type for '%s'",
```
doktoric, line 83: can we somehow highlight what are supported?

gregito: unfortunately not, or it wouldn't be nice at least since, for example, we're supporting 60 different instance types for the eu-central-1a location. other locations may have more so the result message would be enormous.

doktoric (approved):
overall I am not sure we need app yaml for it

doktoric (approved):

PR #10385, opened 2021-03-30
CB-11760 Datalake upgrade: NPE when CM does not send host status

gergopapi2:
During datalake upgrade in the cluster sync step (part of the image change flow) the code asks CM for host statuses. CM server sometimes returns an empty response. This was not expected and an NPE was thrown.

After the NPE, even though the imafge change flow failed, the calling datalake flow proceeded and failed at the replaceVm - upscale step, as the new image was not copied.

This commit fixes the NPE itself only.


keyki (approved):

lacikaaa:
these instances would stay in `SERVICES_RUNNING` state. Is this the right state? Would it cause any issue later?
I think the sync might set the right state eventually.

keyki:
> these instances would stay in `SERVICES_RUNNING` state. Is this the right state? Would it cause any issue later?
> I think the sync might set the right state eventually.

Basically what CM says is that the metrics are NOT_AVAILABLE. Based on that we can't determine whether we should change the state of the instances. I've only seen this happen during upgrade so I think it's safe. The regular sync should set the correct host status if it changes.

PR #10386, opened 2021-03-30
CB-11839 Cloudwatch alarms are not deleted in case of sdx delete

hreeve-cloudera:
Adding additional logging to try and trace the cloudwatch SDX deletion failures.

I wasn't able to recreate a failure to delete the cloudwatch alarms in my local CB environment, nor was I able to find a clear code path that would lead to deletion being skipped. I also don't see any "Unable to delete cloudwatch alarms" in Kibana at all over the last 30 days, which makes me think deletion is being skipped, not failing. This change adds more logging to the full call stack, so that we can better identify why deletion isn't always happening.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -2,6 +2,7 @@ package com.sequenceiq.cloudbreak.cloud.aws.connector.resource;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


hreeve-cloudera:
> @hreeve-cloudera have you tested the scenario when you delete the instance on the provider side, let CB to recognize it (state changes to DELETED_ON_PROVIDER) then terminate the cluster?

I did try deleting a single EC2 instance and then doing a repair, to trigger the downscale operation. But I didn't try a terminate of the cluster with a deleted instance. I'll try that today.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -92,17 +94,18 @@ public void deleteCloudWatchAlarmsForSystemFailures(CloudStack stack, String reg
                 .collect(Collectors.toList());
         if (!instanceIdsNotInStack.isEmpty()) {
             LOGGER.warn("Instance IDs [{}] are not part of cloud stack {}, these instances may have already been deleted on the cloud provider side.",
-                    instanceIdsFromStack, stack);
+                instanceIdsNotInStack, stack);
```
brycederriso, line 97: **praise:** Thanks for fixing confusing logs!

brycederriso (approved):
Looks good to me.

hreeve-cloudera:
Abandoning this because I think I may have found a proper fix thanks to the suggestion from @lacikaaa

PR #10387, opened 2021-03-31
CB-11868 Create DistroX internal API for COD usecase

foldik:

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
@@ -61,6 +61,7 @@ import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.RetryableFlowRe
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackStatusV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackV4Response;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.StackViewV4Responses;
+import com.sequenceiq.cloudbreak.auth.security.internal.AccountId;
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
```diff
@@ -105,6 +106,12 @@ StackViewV4Responses list(
             nickname = "postDistroXV1")
     StackV4Response post(@Valid DistroXV1Request request);
 
+    @POST
+    @Path("/operational_database")
```
horadla23, line 110: I would not put OPDB into the path or method name, since this can be used for internal distrox creation also

core/src/main/java/com/sequenceiq/distrox/v1/distrox/controller/DistroXV1Controller.java
```diff
@@ -157,6 +159,12 @@ public StackV4Response post(@Valid @RequestObject DistroXV1Request request) {
         return distroxService.post(request);
```
horadla23, line 159: usually when we are creating an equivalent internal api, then one calls the other in controller to avoid double administration in the future

horadla23 (changes requested):

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
lnardai: It could even be it's own Controller, who know what else needs to be duplicated.
We should have some form of pattern for defining internal API's.

We already have examples like Flow and CB legacy API.

foldik: I reused the other method call. It's easy to move this into a new controller later if we need to.

horadla23 (approved):

PR #10388, opened 2021-03-31
CB-11879 do not get in project subnets if customer give us shared vpc…

doktoric:
…. also we expose permission issues now.

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpNetworkConnector.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpNetworkConnector.java
@@ -149,21 +149,25 @@ public class GcpNetworkConnector extends AbstractGcpResourceBuilder implements D
         String projectId = GcpStackUtil.getProjectId(credential);
         Subnetwork subnet = null;
         try {
-            SubnetworkList ownProjectSubnets = compute.subnetworks().list(projectId, region).execute();
-            Set<Subnetwork> collect = new HashSet<>();
-            if (ownProjectSubnets.getItems() != null) {
-                collect = ownProjectSubnets.getItems()
-                        .stream()
-                        .filter(e -> e.getName().equals(subnetId) || e.getId().toString().equals(subnetId))
-                        .collect(Collectors.toSet());
-            }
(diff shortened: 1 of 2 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpNetworkConnector.java
```diff
@@ -149,21 +149,21 @@ public NetworkCidr getNetworkCidr(Network network, CloudCredential credential) {
...
-            if (ownProjectSubnets.getItems() != null) {
-                collect = ownProjectSubnets.getItems()
-                        .stream()
-                        .filter(e -> e.getName().equals(subnetId) || e.getId().toString().equals(subnetId))
-                        .collect(Collectors.toSet());
-            }
-            if (collect.isEmpty() && !Strings.isNullOrEmpty(sharedProjectId)) {
-                subnet = compute.subnetworks().get(sharedProjectId, region, subnetId).execute();
-            } else {
+            if (Strings.isNullOrEmpty(sharedProjectId)) {
```
topolyai5, line 152: a log would be good, the subnet searched by sharedProjectId or not

doktoric: done

topolyai5 (approved):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/service/checker/AbstractGcpBaseResourceChecker.java
```diff
@@ -30,25 +30,53 @@ protected CloudResource createNamedResource(ResourceType type, String name) {
...
+                LOGGER.info("Resource {} not found: {}", resourceType, name);
+            } else {
+                throw extractedPermissionIssue(ex, jsonError);
+            }
+        } else {
+            throw extractIfNoJson(ex);
+        }
+    }
+
+    protected GcpResourceException exceptionHandlerWithThrow(GoogleJsonResponseException ex, String name, ResourceType resourceType) {
```
bergerdenes, line 45: name is misleading as this method does not throw, as opposed to the one above, which throws
it should be named that suggest this method produces a new exception instance

PR #10392, opened 2021-03-31
CB-10888 and CB-10889 Introduce new test case for SSH and CM access to SDX with new workload password

aszegedi:
The related test (`testSdxSshAndCmAccessWithNewWorkloadPassword`) tends to cover:
1. `test_cm_access_with_new_password`
2. `test_cm_access_with_new_password_non_power_users`
3. `test_ssh_access_to_datalake_node`
4. `test_new_password_for_new_environment`

test cases from `CB promotion L0/basic tests` ([cdpmc-qe/tests/regression/users/test_reset_password_flow.py](https://github.infra.cloudera.com/CDH/cdpmc-qe/blob/master/tests/regression/users/test_reset_password_flow.py))


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .then((tc, testDto, client) -> clouderaManagerUtil.checkClouderaManagerUserDetails(tc, testDto, client, workloadUsername, newWorkloadPassword))
+                .validate();
+    }
+
+    private String generateRecipeContent() throws IOException {
+        String recipeContentFromFile = ResourceUtil.readResourceAsString(applicationContext, getRecipePath());
+        return Base64.encodeBase64String(recipeContentFromFile.getBytes());
+    }
+
+    private String getWorkloadUsername(TestContext testContext) {
```
horadla23, line 132: do we necessarily need this method? I think we should configure the sso user for this test from parameter, but checking for it is not necessary, test will fail later anyway because of missing user

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/clouderamanager/action/ClouderaManagerClientActions.java
```diff
@@ -0,0 +1,188 @@
...
+    @Value("${integrationtest.clouderamanager.defaultUser:}")
+    private String cmUser;
+
+    @Value("${integrationtest.clouderamanager.defaultPassword:}")
+    private String cmPassword;
+
+    @Value("${integrationtest.cloudProvider:}")
+    private String cloudProvider;
+
+    public StackTestDto checkCmTestUser(TestContext testContext, StackTestDto stackTestDto, CloudbreakClient cloudbreakClient) {
```
horadla23, line 49: is this needed?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/clouderamanager/action/ClouderaManagerClientActions.java
```diff
@@ -0,0 +1,188 @@
...
+            LOGGER.error("Exception when calling UsersResourceApi#readUsers2", e);
+            throw new TestFailException("Exception when calling UsersResourceApi#readUsers2 at " + apiClient.getBasePath(), e);
+        } catch (Exception e) {
+            LOGGER.error("Can't get users' list at: '{}'!", apiClient.getBasePath());
+            throw new TestFailException("Can't get users' list at: " + apiClient.getBasePath(), e);
+        }
+        return sdxInternalTestDto;
+    }
+
+    public DistroXTestDto checkClouderaManagerUsers(TestContext testContext, DistroXTestDto distroXTestDto, CloudbreakClient cloudbreakClient) {
```
horadla23, line 127: is this needed?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
```diff
@@ -0,0 +1,144 @@
...
+        String idbrokerInstanceGroup = "idbroker";
+        String newWorkloadPassword = "Admin123@";
+        String workloadUsername = getWorkloadUsername(testContext);
+
+        testContext
+                .given(clouderaManager, ClouderaManagerTestDto.class)
+                .given(cluster, ClusterTestDto.class).withBlueprintName(getDefaultSDXBlueprintName()).withValidateBlueprint(Boolean.FALSE)
+                .withClouderaManager(clouderaManager)
+                .given(RecipeTestDto.class).withName(recipeName).withContent(generateRecipeContent())
+                .withRecipeType(POST_CLOUDERA_MANAGER_START)
```
horadla23, line 93: why do we need a recipe?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .when(recipeTestClient.createV4())
+                .given(masterInstanceGroup, InstanceGroupTestDto.class).withHostGroup(MASTER).withNodeCount(1).withRecipes(recipeName)
+                .given(idbrokerInstanceGroup, InstanceGroupTestDto.class).withHostGroup(IDBROKER).withNodeCount(1).withRecipes(recipeName)
```
horadla23, line 96: Is necessary to define master and idbroker group count? I think in light duty the nodecount is 1 by default.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .given(stack, StackTestDto.class).withCluster(cluster).withInstanceGroups(masterInstanceGroup, idbrokerInstanceGroup)
+                .given(sdxInternal, SdxInternalTestDto.class)
+                .withCloudStorage(getCloudStorageRequest(testContext))
+                .withStackRequest(key(cluster), key(stack))
+                .when(sdxTestClient.createInternal(), key(sdxInternal))
+                .await(SdxClusterStatusResponse.RUNNING)
+                .awaitForInstance(getSdxInstancesHealthyState())
+                .then((tc, testDto, client) -> sshJUtil.checkFilesOnHostByNameAndPath(testDto, client, List.of(MASTER.getName(), IDBROKER.getName()),
```
horadla23, line 104: I would check for a log file or something like that, for this I would not create a recipe

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .given(FreeIpaUserSyncTestDto.class)
+                .when(freeIpaTestClient.getLastSyncOperationStatus())
+                .await(OperationState.COMPLETED)
+                .given(UmsTestDto.class).assignTarget(EnvironmentTestDto.class.getSimpleName())
+                .when(umsTestClient.setWorkloadPassword(newWorkloadPassword))
+                .given(FreeIpaUserSyncTestDto.class)
+                .when(freeIpaTestClient.syncAll())
+                .await(OperationState.COMPLETED)
+                .given(sdxInternal, SdxInternalTestDto.class)
+                .then((tc, testDto, client) -> sshJUtil.checkFilesOnHostByNameAndPath(testDto, client, List.of(MASTER.getName()),
```
horadla23, line 121: this file check is duplicated, I would do it only once

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/clouderamanager/action/ClouderaManagerClientActions.java
aszegedi: Yep, the https://github.com/hortonworks/cloudbreak/blob/dae4debd77cf45027ba34b41761958ec5a91ea11/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/smoke/YarnSmokeTest.java#L81 uses it through the https://github.com/hortonworks/cloudbreak/blob/dae4debd77cf45027ba34b41761958ec5a91ea11/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/clouderamanager/ClouderaManagerUtil.java#L25 reference. I've just updated the related function and did not rewrite it.

Should I rewrite/refactor? How about?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
aszegedi: User has been configured at https://github.com/hortonworks/cloudbreak/blob/dae4debd77cf45027ba34b41761958ec5a91ea11/integration-test/src/main/resources/application.yml#L43-L47
However the Workload Username is not present here and I do not think we should add this as a parameter, because of we can get this from Manowar.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
aszegedi: We do not have E2E recipe test. So I think this will be useful as a side effect for SSH test. Beyond this I think a proper flow if we not just briefly SSH to the nodes even more checking something there (a newly created file is present).

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/clouderamanager/action/ClouderaManagerClientActions.java
horadla23: can you check if we actually use YarnSmokeTest or not? I found a testsuit yaml for that but I am not sure if we are still using it, the test itself was updated only once in 2020, not sure if it is used anymore

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
horadla23: if we need a recipe e2e test, i would create separate one for it and not mix different test cases. From authorization point of view, just SSHing is enough, checking files is not needed in these test cases I think.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
aszegedi: In this second try we SSH with user and password (namely with workload username and updated workload password). If this is success, that means our test user also can SSH to the `MASTER` with its new workload password.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
horadla23: by default every user can read log files, etc, on nodes, there is no specific authorization logic for files, so I would not bring file permission into these test cases.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
horadla23: makes sense, thank you!

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
horadla23: and what if we do not have such sso user `csso_cloudbreak-qe` in one of our environment? Anyway I would avoid hardcoding an user in the test code, if the email or something changes, we need a new version

horadla23: in my opinion the user is a prerequisite for test, if the user is not exists, the test would fail anyway, so  that's why I think we could spare a GRPC call if we would not check for user details here, also test code would be smaller

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
aszegedi: I'll give a try. Frankly we have not used this on another way yet.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java
aszegedi: `csso_cloudbreak-qe` just a default/sample value here what we are overwriting with the real acting workload username based on 
```workloadUsername.set(client.getDefaultClient().getUserDetails(userCrn, userCrn, Optional.of("")).getWorkloadUsername());```
So this is not a hardcoded/static value here. This is the result of the getter from UMS Test Client.

Comment (horadla23) [reply integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java:132]: i am sorry, you are right, cool, then this is ok for me

File: integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java (hunk 5 above)
Comment (horadla23) [reply integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/InternalSdxSshAndCmAccessTest.java:96]: I think we do not even SdxInternalTestDto here, SdxTestDto is enough, for example SdxRepairTests uses SdxTestDto without any extra configuration. I see that SshJUtil can handle SdxInternalTestDto only, but SdxTestDto should be enough too to get an IP address for SSH.

**Review by horadla23 — APPROVED** [review]

File: integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java [hunk 8]
```diff
@@ -155,14 +164,16 @@ protected void createDefaultEnvironmentWithNetwork(TestContext testContext) {
 
     protected void createDatalake(TestContext testContext) {
         SdxDatabaseRequest sdxDatabaseRequest = new SdxDatabaseRequest();
-        sdxDatabaseRequest.setAvailabilityType(SdxDatabaseAvailabilityType.NONE);
+        sdxDatabaseRequest.setAvailabilityType(SdxDatabaseAvailabilityType.NON_HA);
```
horadla23, line 167: why is this needed here?

PR #10393, opened 2021-03-31
CB-11799 Add API to check users and groups in ipa server

horadla23:

### DIFF
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/ClientTestV1Endpoint.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/ClientTestV1Endpoint.java
@@ -1,13 +1,19 @@
 package com.sequenceiq.freeipa.api.v1.freeipa.test;
 
+import javax.validation.Valid;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
 import com.sequenceiq.cloudbreak.jerseyclient.RetryAndMetrics;
+import com.sequenceiq.freeipa.api.v1.freeipa.test.model.CheckGroupsV1Request;
+import com.sequenceiq.freeipa.api.v1.freeipa.test.model.CheckUsersInGroupV1Request;
+import com.sequenceiq.freeipa.api.v1.freeipa.test.model.CheckUsersV1Request;
+import com.sequenceiq.freeipa.api.v1.freeipa.test.model.ClientTestBaseV1Response;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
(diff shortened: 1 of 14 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/client/FreeipaClientTestService.java
```diff
@@ -0,0 +1,105 @@
...
+
+        try {
+            User user = freeIpaClient.userShow(name);
+            LOGGER.info("Groups: {}", user.getMemberOfGroup());
+            LOGGER.info("Success: {}", user);
+        } catch (Exception e) {
+            LOGGER.error("Error showing user {}", name, e);
+            return "FAILED TO SHOW USER";
+        }
+        return "END";
```
horadla23, line 48: I want to declare, I just moved this logic into the service, but I do not wanted to touch it, even if I am not agree with the desing of it

lacikaaa: it wasn't designed :D

horadla23: yes, it definitely seems it is not created for actual customer use cases :D

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/ClientTestV1Endpoint.java
```diff
@@ -17,9 +22,29 @@
...
     @Path("{id}/{name}")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "Retrieves user information", produces = MediaType.APPLICATION_JSON, nickname = "userShowV1")
     String userShow(@PathParam("id") Long id, @PathParam("name") String name);
+
+    @POST
+    @Path("check_users")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Checks if users exists in FreeIPA of the environment", produces = MediaType.APPLICATION_JSON, nickname = "checkUsers")
+    Boolean checkUsers(@Valid CheckUsersRequest checkUsersRequest);
```
lacikaaa, line 36: are you sure a boolean will be sufficient here later? I would return a response object which could be enhanced later if needed

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/model/CheckGroupsRequest.java
```diff
@@ -0,0 +1,29 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel("CheckGroupsRequest")
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CheckGroupsRequest extends ClientTestBase {
+
+    @NotNull
```
lacikaaa, line 17: `@NotNull` is included in `@NotEmpty`, isn't it?

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/model/CheckUsersInGroupRequest.java
```diff
@@ -0,0 +1,41 @@
...
+@ApiModel("CheckUsersInGroupRequest")
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CheckUsersInGroupRequest extends ClientTestBase {
+
+    @NotNull
+    @NotEmpty
+    @ApiModelProperty(value = "Users to check for", required = true)
+    private Set<String> users;
+
+    @NotNull
```
lacikaaa, line 22: `@NotEmpty` would be better if empty string is invalid

freeipa/src/main/java/com/sequenceiq/freeipa/service/client/FreeipaClientTestService.java
```diff
@@ -0,0 +1,105 @@
...
+    public Boolean checkUsers(String environmentCrn, Set<String> requestedUsers) {
+        FreeIpaClient freeIpaClient = getClientByEnvironmentCrn(environmentCrn);
+        try {
+            LOGGER.info("Checking for users [{}] in environment {}", Joiner.on(",").join(requestedUsers), environmentCrn);
+            Set<String> freeipaUsers = freeIpaClient.userFindAll().stream().map(User::getUid).collect(Collectors.toSet());
+            LOGGER.debug("Users in freeipa: [{}]", Joiner.on(",").join(freeipaUsers));
+            return freeipaUsers.containsAll(requestedUsers);
+        } catch (FreeIpaClientException e) {
+            LOGGER.error("Find user FreeIPA call failed!", e);
+            throw new InternalServerErrorException("Find user FreeIPA call failed!");
```
lacikaaa, line 60: you could use `FreeIpaClientExceptionWrapper`

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/ClientTestV1Endpoint.java
horadla23: makes sense, i wasn't sure if i need that or not, but maybe it is worth to use response object

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/model/CheckGroupsRequest.java
horadla23: yes you're right

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/test/model/CheckUsersInGroupRequest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import javax.validation.constraints.NotEmpty;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.sequenceiq.service.api.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel("CheckUsersInGroupRequest")
```
lacikaaa, line 13: please add `v1` to the name of the requests as we do it with the others

horadla23: done

lacikaaa (approved):

PR #10395, opened 2021-03-31
CB-11645: Added Yarn to the supported load balancer cloud platform list.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-11645

Copying extended commit message here:

This change was originally meant to track upscale/downscale support for YCloud load balancers, however as upscale and downscale operations are not supported for Yarn stacks, this just tracks the official support for YCloud load balancers. All other load balancer logic for YCloud has been implemented and tested.

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -670,7 +670,7 @@ cb:
     readTimeoutMs: 5000
 
   loadBalancer:
-    supportedPlatforms: AWS
+    supportedPlatforms: AWS,YARN
 
 clusterProxy:
   url: http://localhost:10180/cluster-proxy


hreeve-cloudera:
Can you add something more descriptive to the commit message, so we don't lose the context of why this was a separate commit? Something like, "Because upscale and downscale operations are not supported for Yarn stacks, all supported load balancer logic has been implemented and tested, and support can be officially enabled."

sxxgrc:
Sure will do.

sxxgrc:
@hreeve-cloudera Done!

hreeve-cloudera (approved):

PR #10396, opened 2021-04-01
CB-11711 Open NGINX for VPC CIDR with CCMv2

bergerdenes:
Only on data hubs and data lakes.
This way all ingress traffic is enabled from within the VPC towards
NGINX port 9443 if CCMv2 is enabled.
This is achieved by adding CIDR list to gateway pillar and using it
with NGINX SSL config.


zabhi:
We are switching to a model where on the server side, we can pick any of the open connections and route a request via it. In case of FreeIPA HA, this means that request for node A (healthy) might get routed via node B (also healthy). Because of this, the change would also be required for FreeIPA.

biharitomi:
> We are switching to a model where on the server side, we can pick any of the open connections and route a request via it. In case of FreeIPA HA, this means that request for node A (healthy) might get routed via node B (also healthy). Because of this, the change would also be required for FreeIPA.

@zabhi Is it necessary to replicate states between all the CCMv2 agents on A, B and C nodes? What if A node is unhealthy, is CCMv2 able to continue working with the agents on B or C?

biharitomi (approved):
LGTM

PR #10398, opened 2021-04-01
DISTX-587 Integrate Periscope with audit service

smaniraju:
1. Add Control Plane API audit for periscope rest events.
2. Add Control Plane Service audit for periscope autoscaling trigger events.


smaniraju:
Periscope integrates with Control Plane Audit Service directly and not via CB Structured Events, because CB Structured events module is tightly integrated with Flows which is integrated with various other CB Modules.This would introduce of lot of unused flow tables and duplicate tables related to nodes into periscope schema. It also introduces conflicts with existing periscope services\implementation related to HA, Caching, Leader Allocation...

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
```diff
@@ -0,0 +1,96 @@
...
+                    .withMutating(mutating)
+                    .withRequestParameters(new Json(requestParameters).getValue())
+                    .withUserAgent(userAgent)
+                    .build();
+            event = AuditEvent.builder()
+                    .withAccountId(accountId)
+                    .withActor(ActorCrn.builder().withActorCrn(userCrn).build())
+                    .withEventData(apiRequestData)
+                    .withEventName(AuditEventName.MANAGE_AUTOSCALE_DATAHUB_CLUSTER)
+                    .withEventSource(Crn.Service.DATAHUB)
```
smaniraju, line 60: Audit service only recognizes Datahub as an event source and not individual services like cb,periscope.

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    @Async
+    public void auditAutoscaleServiceEvent(ScalingStatus scalingStatus, String auditMessage, String clusterCrn, String accountId) {
+        AuditEvent event = null;
+        try {
+            Map<String, Object> eventDetails = new HashMap<>();
+            eventDetails.put(CLUSTER_CRN, clusterCrn);
+            eventDetails.put(TIMESTAMP, System.currentTimeMillis());
```
sidseth, line 75: Can timestamp references be passed in from the caller? That is primarily for the scenario where the async audit handling is backlogged. (e.g. Queued up on the Audit service being down, which can cause timestamps which are quite different from the actual timestamp)

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static final String AUTOSCALE_TRIGGER_STATUS = "autoscaleTriggerStatus";
+
+    @Inject
+    private AuditClient auditClient;
+
+    @Value("${info.app.version:}")
+    private String periscopeVersion;
+
+    @Async
```
sidseth, line 44: Is this configured properly with SpringBoot - i.e. thread counts, names etc (Will it spawn an unlimited number of threads, or controlled - again in case the destination service is down.)

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
```diff
@@ -0,0 +1,96 @@
...
+            eventDetails.put(AUTOSCALE_TRIGGER_STATUS, scalingStatus);
+
+            ServiceEventData serviceEventData = ServiceEventData.builder()
+                    .withEventDetails(new Json(eventDetails).getValue())
+                    .withVersion(periscopeVersion)
+                    .build();
+
+            event = AuditEvent.builder()
+                    .withAccountId(accountId)
+                    .withEventData(serviceEventData)
```
sidseth, line 86: Any chance this could be a structured event of some kind, for easy parsing (e.g. currentNodeCount, newNodeCount, reason, etc). Probably a future patch.

sidseth (commented):
Mostly looks good. Have a few questions on the thread handling in case the destination service is down.

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
smaniraju: This is integration with control plane auditing service and this audit service expects ServiceEventData as the structured event.  This is basically for auditing actions and for end user usage who can query all audit events of the account.

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/service/AuditService.java
smaniraju: Yes it is configured in spring boot with default core thread size as 50 and max as 500.  In Periscope Onprem they have changed this to core 500 and max as 1000. I will submit a different PR for this change.

sidseth (approved):
LGTM

PR #10403, opened 2021-04-01
CB-11839 Cloudwatch alarms are not deleted in case of sdx delete

hreeve-cloudera:
Adds a list of deleted instances to the CloudStack object. During a stack termination, we'll send both the
running instances ids and terminated instances ids to the deleteCloudWatchAlarmsForSystemFailures. This
will allow us to delete any lingering cloudwatch alarms for instances that have been termianted on the
provider side.

Tested with unit tests and running CB locally, deleting the cloud instances, and verifying the deleted alarm
request succeeded for the deleted instances.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Group.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Group.java
@@ -1,6 +1,8 @@
 package com.sequenceiq.cloudbreak.cloud.model;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -32,24 +34,32 @@ public class Group extends DynamicModel {
 
     private final Optional<CloudFileSystemView> identity;
 
+    private final List<CloudInstance> deletedInstances;
+
     public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
             InstanceAuthentication instanceAuthentication, String loginUserName, String publicKey,
             int rootVolumeSize, Optional<CloudFileSystemView> identity) {
-        this.name = name;
-        this.type = type;
-        this.instances = ImmutableList.copyOf(instances);
-        this.security = security;
-        this.skeleton = Optional.ofNullable(skeleton);
-        this.instanceAuthentication = instanceAuthentication;
-        this.publicKey = publicKey;
-        this.loginUserName = loginUserName;
-        this.rootVolumeSize = rootVolumeSize;
-        this.identity = identity;
+        this(name, type, instances, security, skeleton, new HashMap<>(), instanceAuthentication, loginUserName, publicKey,
+            rootVolumeSize, identity, new ArrayList<>());
+    }
+
+    public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
+            InstanceAuthentication instanceAuthentication, String loginUserName, String publicKey,
+            int rootVolumeSize, Optional<CloudFileSystemView> identity, List<CloudInstance> deletedInstances) {
+        this(name, type, instances, security, skeleton, new HashMap<>(), instanceAuthentication, loginUserName, publicKey,
+            rootVolumeSize, identity, deletedInstances);
     }
 
     public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
             Map<String, Object> parameters, InstanceAuthentication instanceAuthentication, String loginUserName,
             String publicKey, int rootVolumeSize, Optional<CloudFileSystemView> identity) {
+        this(name, type, instances, security, skeleton, parameters, instanceAuthentication, loginUserName, publicKey,
+            rootVolumeSize, identity, new ArrayList<>());
+    }
+
+    public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
+            Map<String, Object> parameters, InstanceAuthentication instanceAuthentication, String loginUserName,
+            String publicKey, int rootVolumeSize, Optional<CloudFileSystemView> identity, List<CloudInstance> deletedInstances) {
         super(parameters);
         this.name = name;
         this.type = type;
@@ -61,6 +71,7 @@ public class Group extends DynamicModel {
         this.loginUserName = loginUserName;
         this.rootVolumeSize = rootVolumeSize;
         this.identity = identity;
+        this.deletedInstances = deletedInstances;
     }
 
     public CloudInstance getReferenceInstanceConfiguration() {
@@ -114,6 +125,10 @@ public class Group extends DynamicModel {
         return identity;
     }
 
+    public List<CloudInstance> getDeletedInstances() {
+        return deletedInstances;
+    }
+
     @Override
     public String toString() {
         return "Group{"
@@ -125,6 +140,7 @@ public class Group extends DynamicModel {
                 + ", loginUserName='" + loginUserName + '\''
                 + ", instanceAuthentication=" + instanceAuthentication
                 + ", skeleton=" + skeleton
+                + ", deletedInstances=" + deletedInstances
                 + '}';
     }
 }
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
@@ -2,6 +2,7 @@ package com.sequenceiq.cloudbreak.cloud.aws.connector.resource;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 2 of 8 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Group.java
```diff
@@ -32,24 +34,32 @@
...
-        this.type = type;
-        this.instances = ImmutableList.copyOf(instances);
-        this.security = security;
-        this.skeleton = Optional.ofNullable(skeleton);
-        this.instanceAuthentication = instanceAuthentication;
-        this.publicKey = publicKey;
-        this.loginUserName = loginUserName;
-        this.rootVolumeSize = rootVolumeSize;
-        this.identity = identity;
+        this(name, type, instances, security, skeleton, new HashMap<>(), instanceAuthentication, loginUserName, publicKey,
```
brycederriso, line 42: **praise:** Thank you for delegating to another constructor here!

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Group.java
```diff
@@ -32,24 +34,32 @@
...
     }
 
     public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
             Map<String, Object> parameters, InstanceAuthentication instanceAuthentication, String loginUserName,
             String publicKey, int rootVolumeSize, Optional<CloudFileSystemView> identity) {
+        this(name, type, instances, security, skeleton, parameters, instanceAuthentication, loginUserName, publicKey,
+            rootVolumeSize, identity, new ArrayList<>());
+    }
+
+    public Group(String name, InstanceGroupType type, Collection<CloudInstance> instances, Security security, CloudInstance skeleton,
```
brycederriso, line 60: **suggestion:** The number of arguments and overloadings of this constructor makes me think it would be good to associate it with a Builder.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/instance/InstanceGroup.java
```diff
@@ -130,6 +130,12 @@ public InstanceGroup replaceInstanceMetadata(Set<InstanceMetaData> instanceMetaD
                 .collect(Collectors.toSet());
     }
 
+    public Set<InstanceMetaData> getDeletedInstanceMetaDataSet() {
+        return instanceMetaData.stream()
+            .filter(metaData -> metaData.isTerminated() || metaData.isDeletedOnProvider())
```
brycederriso, line 135: **praise:** Great use of streams and filter!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -74,35 +75,51 @@ public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances,
...
+        List<String> deletedInstanceIds = stack.getGroups().stream()
+            .flatMap(group -> group.getDeletedInstances().stream())
+            .map(CloudInstance::getInstanceId)
+            .collect(Collectors.toList());
+        LOGGER.info("Found ids for deleted instances: [{}]. Deletion will be attempted on the cloudwatch alarms for these instances.", deletedInstanceIds);
+        instanceIds.addAll(deletedInstanceIds);
         deleteCloudWatchAlarmsForSystemFailures(stack, regionName, credentialView, instanceIds);
     }
 
     public void deleteCloudWatchAlarmsForSystemFailures(CloudStack stack, String regionName, AwsCredentialView credentialView, List<String> instanceIds) {
```
brycederriso, line 93: **question:** Do we need to call this method in `AwsDownscaleService`? Could we use the `deleteAllCloudWatchAlarmsForSystemFailures` method instead?

I think the naming of this method and the one above is a bit confusing, and thought we could just make this one `private` and run everything through the delete _all_.

Would we want to keep cloudwatch alarms around for downscaled instances?

brycederriso: Ah, I see you've already made one method private, so presumably you already looked into this?

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -74,35 +75,51 @@ public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances,
...
-        List<String> instanceIdsFromStack = stack.getGroups().stream()
+        if (instanceIds == null || instanceIds.isEmpty()) {
+            LOGGER.warn("No instance ids provided for cloudwatch alarm deletion. No alarms will be deleted.");
+        } else {
+            LOGGER.info("Deleting alarms for instance ids [{}]", instanceIds);
+            List<String> instanceIdsFromStack = stack.getGroups().stream()
```
brycederriso, line 98: **suggestion:** From what I can tell, the body of this `else` is entirely to create the log warning statement, and it uses the `stack` only.

I think you can consider extracting this into a method like `private void logMissingInstances(CloudStack stack)`.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -74,35 +75,51 @@ public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances,
...
+            }
+            deleteCloudWatchAlarmsForSystemFailures(regionName, credentialView, instanceIds);
         }
-        deleteCloudWatchAlarmsForSystemFailures(regionName, credentialView, instanceIds);
     }
 
     private void deleteCloudWatchAlarmsForSystemFailures(String regionName, AwsCredentialView credentialView, List<String> instanceIds) {
+        // The list of alarms received may be longer than 100 items, but alarms can only be deleted in batches of max
+        // size 100. To work around this, break the instanceIds list into chunks no greater than 100, and process each
+        // chunk. See https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_DeleteAlarms.html
```
brycederriso, line 116: **praise:** Thanks for explaining this batch size, which would be pretty obtuse otherwise!

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
```diff
@@ -74,35 +75,51 @@ public void addCloudWatchAlarmsForSystemFailures(List<CloudResource> instances,
...
         final AtomicInteger counter = new AtomicInteger(0);
-        instanceIds.stream()
+        Map<Integer, List<String>> grouping = instanceIds.stream()
                 .map(instanceId -> instanceId + alarmSuffix)
-                .collect(Collectors.groupingBy(s -> counter.getAndIncrement() / maxBatchsize))
-                .values()
+                .collect(Collectors.groupingBy(s -> counter.getAndIncrement() / maxBatchsize));
```
brycederriso, line 120: **question:** Do you know if this is idiomatic for how to break a `Collection` into groups?

TIL this.

cloud-aws/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchServiceTest.java
```diff
@@ -207,4 +207,49 @@ void testDeleteCloudWatchAlarmsForSystemFailuresDoesNotBatchDeletionWhenDescribe
...
+        DescribeAlarmsResult describeAlarmsResult = mock(DescribeAlarmsResult.class);
+        AmazonCloudWatchClient cloudWatchClient = mock(AmazonCloudWatchClient.class);
+        AwsCredentialView credentialView = mock(AwsCredentialView.class);
+        CloudStack stack = mock(CloudStack.class);
+        Group group = mock(Group.class);
+        Group deletedGroup = mock(Group.class);
+        CloudInstance instance1 = mock(CloudInstance.class);
+        CloudInstance instance2 = mock(CloudInstance.class);
+        CloudInstance deletedInstance = mock(CloudInstance.class);
+        List<Group> groups = List.of(group, deletedGroup);
```
brycederriso, line 231: **suggestion:** Consider breaking up this initialization block by separating the declarations + initializations above from the remaining `when(...)` setup.

brycederriso (approved):
Changes look good to me!

I left a smattering of suggestions and questions, but none of them are blocking -- mostly style nits and questions for my own edification.

brycederriso:
retest this please

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
hreeve-cloudera: I didn't change the downscale path. I also didn't make any method private that wasn't private already. I renamed `deleteAllCloudWatchAlarmsForSystemFailures` to include the `All` because that method is only used when we're terminating the instance, and so should be deleting all cloudwatch alarms associated with it. That's different than the downscale case, where we're deleting the cloudwatch alarms for the downscaled instances, but want to leave them for the other instances. So calling `deleteAllCloudWatchAlarmsForSystemFailures` wouldn't make sense in that case, because we don't want to delete all.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
hreeve-cloudera: I didn't write the code, I just tried not to break it. :)

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsCloudWatchService.java
hreeve-cloudera: It's mostly logging, but it also calls the next method in the call stack that's needed to actually do the delete:
```
deleteCloudWatchAlarmsForSystemFailures(regionName, credentialView, instanceIds);
```
I would prefer not extracting it out, since there's already a 3-4 overloaded method deep call stack, and I don't want to add another method to it unless it's really necessary.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/Group.java
hreeve-cloudera: I don't disagree, but these constructors are called _so many_ places. Changing this to a builder would dwarf the actual changes to fix the problem. That probably needs to be its own commit.

PR #10407, opened 2021-04-02
CB-11873: Implement `ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_STARTED_STATE` - to delete the DES from Azure cloud

juhi-09:
CB-11873: Implement `ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_STARTED_STATE` - to delete the DES from Azure cloud
 1. If single Resource group - delete the DES.
 2. If multi RG - delete the DES and then delete the resource group created for DES specifically.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
@@ -1,0 +1,78 @@
+package com.sequenceiq.environment.environment.encryption;
+
+import static com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern.USE_MULTIPLE;
+
+import java.util.Optional;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.cloud.EncryptionResources;
+import com.sequenceiq.cloudbreak.cloud.init.CloudPlatformConnectors;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
+import com.sequenceiq.cloudbreak.cloud.model.CloudPlatformVariant;
+import com.sequenceiq.cloudbreak.cloud.model.Platform;
+import com.sequenceiq.cloudbreak.cloud.model.Variant;
+import com.sequenceiq.cloudbreak.cloud.model.encryption.CreatedEncryptionResources;
+import com.sequenceiq.cloudbreak.cloud.model.encryption.EncryptionResourcesCreationRequest;
+import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
+import com.sequenceiq.environment.credential.v1.converter.CredentialToCloudCredentialConverter;
+import com.sequenceiq.environment.environment.domain.Environment;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.parameter.dto.AzureParametersDto;
+import com.sequenceiq.environment.parameter.dto.AzureResourceGroupDto;
+import com.sequenceiq.environment.parameter.dto.ParametersDto;
+import com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern;
+
+@Component
+public class EnvironmentEncryptionService {
+
+    private final CredentialToCloudCredentialConverter credentialToCloudCredentialConverter;
+
+    private final CloudPlatformConnectors cloudPlatformConnectors;
+
+    public EnvironmentEncryptionService(CredentialToCloudCredentialConverter credentialToCloudCredentialConverter,
+            CloudPlatformConnectors cloudPlatformConnectors) {
+        this.credentialToCloudCredentialConverter = credentialToCloudCredentialConverter;
+        this.cloudPlatformConnectors = cloudPlatformConnectors;
+    }
+
+    public CreatedEncryptionResources createEncryptionResources(EnvironmentDto environmentDto, Environment environment) {
+        EncryptionResources encryptionResources = getCloudConnector(environment.getCloudPlatform());
+        return encryptionResources.createDiskEncryptionSet(createEncryptionResourcesCreationRequest(environmentDto));
+    }
+
+    private EncryptionResources getCloudConnector(String cloudPlatform) {
+        CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(Platform.platform(cloudPlatform), Variant.variant(cloudPlatform));
+        return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant).encryptionResources())
+                .orElseThrow(() -> new NotFoundException("Encryption resources component not found!"));
+    }
+
+    private EncryptionResourcesCreationRequest createEncryptionResourcesCreationRequest(EnvironmentDto environment) {
+        CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+        EncryptionResourcesCreationRequest.Builder builder = new EncryptionResourcesCreationRequest.Builder()
+                .withCloudCredential(cloudCredential)
+                .withRegion(environment.getLocation().getName())
+                .withEnvName(environment.getName())
+                .withEnvId(environment.getId())
+                .withTags(environment.getTags().getUserDefinedTags())
+                .withEncryptionKeyUrl(environment.getParameters().getAzureParametersDto()
+                        .getAzureResourceEncryptionParametersDto().getEncryptionKeyUrl());
+        if (isSingleResourceGroup(environment)) {
+            builder.withSingleResourceGroup(true);
+            builder.withResourceGroupName(environment.getParameters().getAzureParametersDto().getAzureResourceGroupDto().getName());
+        } else {
+            builder.withSingleResourceGroup(false);
+        }
+        return builder.build();
+    }
+
+    private boolean isSingleResourceGroup(EnvironmentDto environmentDto) {
+        ResourceGroupUsagePattern resourceGroupUsagePattern = Optional.ofNullable(environmentDto.getParameters())
+                .map(ParametersDto::azureParametersDto)
+                .map(AzureParametersDto::getAzureResourceGroupDto)
+                .map(AzureResourceGroupDto::getResourceGroupUsagePattern)
+                .orElse(USE_MULTIPLE);
+        return resourceGroupUsagePattern.isSingleResourceGroup();
+    }
+
+}
--- a/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
@@ -14,6 +14,7 @@ import com.sequenceiq.cloudbreak.cloud.model.Platform;
 import com.sequenceiq.cloudbreak.cloud.model.Variant;
 import com.sequenceiq.cloudbreak.cloud.model.encryption.CreatedEncryptionResources;
 import com.sequenceiq.cloudbreak.cloud.model.encryption.EncryptionResourcesCreationRequest;
+import com.sequenceiq.cloudbreak.cloud.model.encryption.EncryptionResourcesDeletionRequest;
 import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
 import com.sequenceiq.environment.credential.v1.converter.CredentialToCloudCredentialConverter;
 import com.sequenceiq.environment.environment.domain.Environment;
@@ -41,6 +42,11 @@ public class EnvironmentEncryptionService {
         return encryptionResources.createDiskEncryptionSet(createEncryptionResourcesCreationRequest(environmentDto));
     }
 
+    public void deleteEncryptionResources(EnvironmentDto environmentDto, Environment environment) {
+        EncryptionResources encryptionResources = getCloudConnector(environment.getCloudPlatform());
+        encryptionResources.deleteDiskEncryptionSet(createEncryptionResourcesDeletionRequest(environmentDto));
+    }
+
     private EncryptionResources getCloudConnector(String cloudPlatform) {
         CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(Platform.platform(cloudPlatform), Variant.variant(cloudPlatform));
         return Optional.ofNullable(cloudPlatformConnectors.get(cloudPlatformVariant).encryptionResources())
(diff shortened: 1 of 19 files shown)


doktoric:
retest this please

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -0,0 +1,76 @@
...
+                .orElseThrow(() -> new EncryptionResourcesNotFoundException("No Encryption resources component found for cloud platform: " + cloudPlatform));
+    }
+
+    @VisibleForTesting
+    DiskEncryptionSetDeletionRequest createEncryptionResourcesDeletionRequest(EnvironmentDto environment) {
+        CloudCredential cloudCredential = credentialToCloudCredentialConverter.convert(environment.getCredential());
+        DiskEncryptionSetDeletionRequest.Builder builder = new DiskEncryptionSetDeletionRequest.Builder()
+                .withCloudCredential(cloudCredential)
+                .withDiskEncryptionSetId(environment.getParameters().getAzureParametersDto()
+                        .getAzureResourceEncryptionParametersDto().getDiskEncryptionSetId());
```
lajosrodek, line 58: 1. Populate the `CloudContext` like in #10372 (including copying over `getCloudContext()`).
2. Instead of setting the DES ID, populate `DiskEncryptionSetDeletionRequest.resources` as follows:
    1. `@Inject` a new instance field of type `ResourceRetriever resourceRetriever` into this class. This will be wired as the [CloudResourceRetrieverService](https://github.com/hortonworks/cloudbreak/blob/master/environment/src/main/java/com/sequenceiq/environment/resourcepersister/CloudResourceRetrieverService.java) instance from the Environment service.
    2. Create a helper method `getResourcesForDeletion(EnvironmentDto environment)` as shown below.
    3. Invoke `withResources()` with the output of `getResourcesForDeletion(environment)`.

`List<CloudResource> getResourcesForDeletion(EnvironmentDto environment)` should be something like this:
```
Optional<CloudResource> desCloudResourceOptional = resourceRetriever.findByResourceReferenceAndStatusAndType(environment.getParameters().getAzureParametersDto().getAzureResourceEncryptionParametersDto().getDiskEncryptionSetId(), CommonStatus.CREATED, ResourceType.AZURE_DISK_ENCRYPTION_SET);
return desCloudResourceOptional.map(r -> List.of(r)).orElse(List.of());
```

environment/src/test/java/com/sequenceiq/environment/environment/EnvironmentStatusTest.java
```diff
@@ -33,6 +33,7 @@
                 { "UPDATE_FAILED",                          EnvironmentStatus.UPDATE_FAILED,                        false },
                 { "ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_IN_PROGRESS", EnvironmentStatus.ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_IN_PROGRESS, false},
                 { "ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_IN_PROGRESS", EnvironmentStatus.ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_IN_PROGRESS, true},
+                { "ENVIRONMENT_ENCRYPTION_RESOURCES_DELETED", EnvironmentStatus.ENVIRONMENT_ENCRYPTION_RESOURCES_DELETED, false},
```
lajosrodek, line 36: I suspect this status should also result in `true`; any status after the start of the env deletion flow (i.e. anything after `EnvironmentStatus.DELETE_INITIATED`) is essentially "deletion in progress" until the environment gets into the `ARCHIVED` status.

@doktoric @bergerdenes @biharitomi Please confirm.

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+import com.sequenceiq.environment.credential.v1.converter.CredentialToCloudCredentialConverter;
+import com.sequenceiq.environment.environment.dto.EnvironmentDto;
+import com.sequenceiq.environment.parameter.dto.AzureParametersDto;
+import com.sequenceiq.environment.parameter.dto.AzureResourceEncryptionParametersDto;
+import com.sequenceiq.environment.parameter.dto.AzureResourceGroupDto;
+import com.sequenceiq.environment.parameter.dto.ParametersDto;
+import com.sequenceiq.environment.parameter.dto.ResourceGroupUsagePattern;
+
+@ExtendWith(MockitoExtension.class)
+class EnvironmentEncryptionServiceTest {
```
lajosrodek, line 30: Please add at least one test case for `deleteEncryptionResources()`.

environment/src/test/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandlerTest.java
```diff
@@ -50,6 +57,9 @@
     @Mock
     private Event<EnvironmentDeletionDto> environmentDtoEvent;
 
+    @Mock
+    private EnvironmentEncryptionService environmentEncryptionService;
```
lajosrodek, line 61: This is not used anywhere below. Please add at least one test method that verifies that `environmentEncryptionService.deleteEncryptionResources()` gets invoked when needed, and another that verifies that that method does not get invoked at all (if there is no DES to delete).

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/encryption/CreatedDiskEncryptionSet.java
```diff
@@ -0,0 +1,107 @@
+package com.sequenceiq.cloudbreak.cloud.model.encryption;
+
+import java.util.Map;
+
+public class CreatedDiskEncryptionSet {
```
lajosrodek, line 5: I guess this class is not needed for the deletion flow.

lajosrodek (changes requested):
First bunch of asks.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
```diff
@@ -39,6 +39,7 @@
 
     ENVIRONMENT_RESOURCE_ENCRYPTION_INITIALIZATION_IN_PROGRESS("Initializing Resource Encryption settings for environment"),
     ENVIRONMENT_RESOURCE_ENCRYPTION_DELETE_IN_PROGRESS("Deleting Resource Encryption settings of environment"),
+    ENVIRONMENT_ENCRYPTION_RESOURCES_DELETED("Deleted Resource Encryption settings for environment"),
```
doktoric, line 42: ui change is required

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -863,4 +865,16 @@ private void handleAuthException(Runnable function) {
     public Completable deleteGenericResourceByIdAsync(String databaseServerId) {
         return handleAuthException(() -> azure.genericResources().deleteByIdAsync(databaseServerId));
     }
+
+    public void deleteDiskEncryptionSet(String resourceGroup, String diskEncryptionSetName) {
+        // The Disk encryption set operations are not exposed in public API, so have to rely on underlying DiskEncryptionSetInner
+        DiskEncryptionSetsInner dSetsIn = azureClientCredentials.getComputeManager().inner().diskEncryptionSets();
+        dSetsIn.delete(resourceGroup, diskEncryptionSetName);
```
doktoric, line 872: what happens if the encryption set is deleted by the customer

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/response/EnvironmentStatus.java
juhi-09: Updated CB-11791 to include this UI change.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
juhi-09: If environment exists and DES is deleted by customer, Datahub(as of now only datahub uses des) create should result in failure. cc: @lajosrodek 
During deletion of environment, DES deletion will not happen from our side as `AzureClient.getDiskEncryptionSet` will return `null`. However, as we have DES stored in `environment_parameters` - respective cloudResource will be deleted.

doktoric: - what if customer deletes the DES permission and we try to delete
- are you sure we get a null ? How we check that the delete is success ?

juhi-09: `Microsoft.Compute/diskEncryptionSets/delete` is required to delete DES. If this is not present we would not be able to delete.
yes, `null` is returned when DES doesn't exists in cloud. To check that DES is indeed deleted - wait for sometime(as there is latency to get correct status) and recheck if `AzureClient.getDiskEncryptionSet` returns `null`.

doktoric: but we need to check the deletion status right ?

environment/src/test/java/com/sequenceiq/environment/environment/EnvironmentStatusTest.java
doktoric: I think it should be true yes

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
lajosrodek: > what happens if the encryption set is deleted by the customer
Subsequent DH launches will surely fail in that case. 😄 This is indeed a possibility as long as the DES is not associated with any disk resource. The handling of that check is planned in [CB-11789](https://jira.cloudera.com/browse/CB-11789); that shall provide sufficient safeguards to block env deletion if the DES is in use by experiences. (The exact logic needs some elaboration, however.)

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -133,4 +143,75 @@ void testShouldThrowExceptionWhenCloudPlatformNotAzure() {
...
+        assertNotNull(dummyResource);
+        assertEquals(dummyResource.get().getReference(), "dummyDesId");
+    }
+
+    @Test
+    void testDeleteEncryptionResourcesShouldCallDeleteDiskEncryptionSetWhenCloudPlatformAzure() {
+        when(cloudPlatformConnectors.get(any(CloudPlatformVariant.class))).thenReturn(cloudConnector);
+        when(cloudConnector.encryptionResources()).thenReturn(encryptionResources);
+        Environment env = new Environment();
+        env.setCloudPlatform("AZURE");
```
lajosrodek, line 187: Not needed. Please also remove from `testCreateEncryptionResourcesShouldCallCreateDiskEncryptionSetWhenCloudPlatformAzure()`.

environment/src/test/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandlerTest.java
```diff
@@ -68,6 +79,15 @@ void setUp() {
...
+                .withParameters(ParametersDto.builder()
+                        .withAzureParameters(AzureParametersDto.builder()
+                                .withEncryptionParameters(AzureResourceEncryptionParametersDto.builder()
+                                        .withDiskEncryptionSetId("/subscriptions/dummySubscriptionId/resourceGroups/dummyResourceGroup/" +
+                                                "providers/Microsoft.Compute/diskEncryptionSets/dummyDesId")
+                                        .build())
+                                .build())
+                        .build())
                 .build();
         EnvironmentDeletionDto build = EnvironmentDeletionDto
```
lajosrodek, line 92: `build` is not an expressive name; something else like `eventData` would look better.

environment/src/test/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandlerTest.java
```diff
@@ -100,6 +120,43 @@ void acceptTestEnvironmentFailure() {
...
+    }
+
+    @Test
+    void testDeletionContinuesIfEnvironmentEncryptionResourcesAreAlreadyDeleted() {
+        Environment environment = new Environment();
+        environment.setStatus(EnvironmentStatus.ENVIRONMENT_ENCRYPTION_RESOURCES_DELETED);
+        when(environmentService.findEnvironmentById(ENVIRONMENT_ID)).thenReturn(Optional.of(environment));
+        underTest.accept(environmentDtoEvent);
+        verify(environmentEncryptionService, never()).deleteEncryptionResources(environmentDtoEvent.getData().getEnvironmentDto(), environment);
+        verify(eventSender).sendEvent(baseNamedFlowEventCaptor.capture(), headersArgumentCaptor.capture());
```
lajosrodek, line 139: Please also check the value of `environment.getStatus()`; it should not change.

lajosrodek (changes requested):
Some further notes.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
juhi-09: The cloud API call to delete the DES doesn’t return anything. However, it may result in
```
@throws IllegalArgumentException thrown if parameters fail the validation      
@throws CloudException thrown if the request is rejected by server
@throws RuntimeException all other wrapped checked exceptions if the request fails to be sent
```
As this is a blocking call, we need not to check the status of deletion - it either succeeds or results in error.

doktoric (approved):

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -133,4 +147,88 @@ void testShouldThrowExceptionWhenCloudPlatformNotAzure() {
...
+    private void verifyDiskEncryptionSetDeletionRequest() {
+        DiskEncryptionSetDeletionRequest deletionRequest = diskEncryptionSetDeletionRequestCaptor.getValue();
+        CloudContext cloudContext = deletionRequest.getCloudContext();
+        List<CloudResource> cloudResources = deletionRequest.getCloudResources();
+        Optional<CloudResource> desCloudResourceOptional = cloudResources.stream()
+                .filter(r -> r.getType() == ResourceType.AZURE_DISK_ENCRYPTION_SET)
+                .findFirst();
+        assertEquals(desCloudResourceOptional.get().getReference(), "dummyDesId");
+        assertEquals(desCloudResourceOptional.get().getName(), "Des");
+        assertEquals(desCloudResourceOptional.get().getType(), ResourceType.AZURE_DISK_ENCRYPTION_SET);
```
lajosrodek, line 230: Nit:

- `desCloudResourceOptional.get()` could be extracted to reduce repetitions.
- Before these checks it might be worth asserting something like `assertTrue(desCloudResourceOptional.isPresent())`. Without this, the first invocation of `desCloudResourceOptional.get()` will fail with `NoSuchElementException` if there is no `AZURE_DISK_ENCRYPTION_SET` resource found in `cloudResources`. The test case will surely abort in any case, so this is more of a personal preference to make the `isPresent()` check first. 😄

lajosrodek (approved):
One minor issue, but it can be fixed separately if you agree. Approving for now, thanks for the changes! 😃

PR #10408, opened 2021-04-03
CB-11906: Fixed creation of datalake with unique image ID when catalog name not present.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-11906

Changes I made:
- Added a new endpoint path for the image catalog for getting an image without a catalog name
- Needed to do this since with the previous path, not having a name would result in a malformed path that would be rejected by the validator

Tested this with the CDP CLI change I made and seems to be working correctly for the two use cases.

Let me know if there is anything we want to change for this.

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/imagecatalog/ImageCatalogV4Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/imagecatalog/ImageCatalogV4Endpoint.java
@@ -118,6 +118,13 @@ public interface ImageCatalogV4Endpoint {
             @QueryParam("stackName") String stackName, @QueryParam("platform") String platform) throws Exception;
 
     @GET
+    @Path("image")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = ImageCatalogOpDescription.GET_IMAGE_BY_ID, produces = MediaType.APPLICATION_JSON,
+            notes = IMAGE_CATALOG_NOTES, nickname = "getImageById")
+    ImagesV4Response getImageByImageId(@PathParam("workspaceId") Long workspaceId, @QueryParam("imageId") String imageId) throws Exception;
+
+    @GET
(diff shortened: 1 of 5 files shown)


LinaAtAustin:
@sxxgrc Based on
https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java#L77, have you tried to set the catalog to be "cdp-default" in CLI if the user of CLI does not provide the catalog? In this way, we don't need to add new endpoint in this code update. And in CLI we can output the catalog we use for the input image ID. This can make trouble shooting easier if user uses an image that is not in "cdp-default" without specifying the catalog, and the creation of DL fails because in CLI error output, you can suggest user to check if the image ID is in the catalog.  

Can you also include CB team in the code review? Such as Krisztian Horvath and Gergely Topolyai? They need to agree with the code change.

sxxgrc:
@LinaAtAustin I have not tried that but it would most likely work. I did not try that originally because I assumed that we wanted the Cloudbreak code to accept null or empty catalog names, and at the moment it will not. This is fine if the only access point for specifying a custom image is through the CDP CLI, but if we do it with anything else, and use a null or empty catalog name that will not work. Let me know and I can just make that change in the CLI code and we will no longer need this change.

As for outputting the image catalog selected in the CLI, I can do that with the describe-datalake result and possibly with the result outputted when you first create a datalake. I could also output the image ID selected if you think that is helpful. All of this would be internal only of course. Let me know if that sounds good for you and I can make those changes as well.

LinaAtAustin:
@sxxgrc I prefer for your CLI code to set the catalog if user does not provide it. In this way, the logic is close to user and you can provide the default value in description, output the info to CB, and in error message to guide user on how to fix the issue.

If we leave the CB default value to be used inside of CB code itself, if it changes in the future, it is very hard to trouble shooting without seeing the code itself. You know how much effort you took. 

You can ask for the feedback from CB team, and then make the agreed change at once.

topolyai5 (approved):

PR #10411, opened 2021-04-03
CB-11907 Extend environment api with freeipa image parameters

pkedvessy:

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
@@ -19,6 +19,7 @@ import java.util.Set;
 
 import com.sequenceiq.common.api.backup.request.BackupRequest;
 import com.sequenceiq.environment.environment.v1.converter.BackupConverter;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.image.ImageSettingsRequest;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -225,6 +226,8 @@ public class FreeIpaCreationHandler extends EventSenderAwareHandler<EnvironmentD
         setTelemetry(environment, createFreeIpaRequest);
         setBackup(environment, createFreeIpaRequest);
         setTags(environment, createFreeIpaRequest);
+        setImage(environment, createFreeIpaRequest);
+
         SecurityGroupRequest securityGroupRequest = null;
         if (environment.getSecurityAccess() != null) {
             securityGroupRequest = createSecurityGroupRequest(environment.getSecurityAccess());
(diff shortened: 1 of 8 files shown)


TheTinkerDad (approved):

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
```diff
@@ -367,6 +370,19 @@ private EnvCreationEvent getNextStepObject(EnvironmentDto environmentDto) {
                 .build();
     }
 
+    private void setImage(EnvironmentDto environment, CreateFreeIpaRequest createFreeIpaRequest) {
+        if (environment.getFreeIpaCreation() != null) {
+            String imageCatalog = environment.getFreeIpaCreation().getImageCatalog();
```
doktoric, line 375: idea: the catalog and the id should be in a Image object

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
```diff
@@ -367,6 +370,19 @@ private EnvCreationEvent getNextStepObject(EnvironmentDto environmentDto) {
...
+            String imageId = environment.getFreeIpaCreation().getImageId();
+            if (!Strings.isNullOrEmpty(imageCatalog) && !Strings.isNullOrEmpty(imageId)) {
```
doktoric, line 377: if null should we fill up the request with default values ?

environment/src/test/java/com/sequenceiq/environment/environment/v1/converter/FreeIpaConverterTest.java
```diff
@@ -66,4 +71,17 @@ public void testConvertWithTwoInstancesAndOnlySpotInstances() {
         assertEquals(0.9, result.getAws().getSpot().getMaxPrice());
     }
 
+    @Test
+    public void testConvertWithImage() {
```
doktoric, line 75: negativ test?

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
pkedvessy: I'd leave it as it is. This structure is already the part of the freeipa api.

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
pkedvessy: If is already handled in freeipa service.

environment/src/test/java/com/sequenceiq/environment/environment/v1/converter/FreeIpaConverterTest.java
pkedvessy: good catch, implemented

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
doktoric: check DistroXImageV1Request we should be more consistent I think. So in this case because this is a new API I suggest to use such what we use for Distrox

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
doktoric: maybe a log would be nice then

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
pkedvessy: fixed

environment/src/main/java/com/sequenceiq/environment/environment/flow/creation/handler/freeipa/FreeIpaCreationHandler.java
pkedvessy: fixed

doktoric (approved):

PR #10412, opened 2021-04-04
CB-11910 Remove subnet validation logic for endpoint gateway

hreeve-cloudera:
Puts the logic that validations endpoint gateway subnets are public behind the CDP_ENDPOINT_GATEWAY_VALIDATION
entitlement. If the entitlement is disabled, all subnets provided for the endpoint gateway will be accepted
regardless of their type. This change also removes the last of the logic first introduced in CB-11070, and
some related validation logic in AwsLoadBalancerLaunchService.

Tested with unit tests, and by providing private subnets for the endpoint gateway and verifying the environment
created successfully, and the load balancers were created with the expected subnets.


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
```diff
@@ -50,11 +76,19 @@
                 } else {
                     subnetsToParse = source.getGatewayEndpointSubnetMetas();
                 }
-                Map<String, CloudSubnet> publicSubnetMetas = subnetsToParse.entrySet().stream()
-                    .filter(entry -> !entry.getValue().isPrivateSubnet())
-                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
-                endpointGatewayCloudSubnet = chooseSubnet(null, publicSubnetMetas, selectedAZ, false);
+                Map<String, CloudSubnet> publicSubnetMetas;
```
brycederriso, line 80: **question:** What are the key `String` values in this `Map`? Are they the names of the cloud subnets?

**suggestion:** Consider naming this `publicSubnetsMetadata`?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
```diff
@@ -17,21 +22,42 @@
...
+    @Inject
+    private EntitlementService entitlementService;
+
+    public Optional<CloudSubnet> chooseSubnet(String preferredSubnetId, Map<String, CloudSubnet> subnetMetas, String availabilityZone,
+            boolean allowRandom) {
+        return chooseSubnet(preferredSubnetId, subnetMetas, availabilityZone, allowRandom, false);
+    }
+
+    @VisibleForTesting
+    Optional<CloudSubnet> chooseSubnetPreferPublic(String preferredSubnetId, Map<String, CloudSubnet> subnetMetas, String availabilityZone) {
```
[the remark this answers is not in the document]
brycederriso: I thought about this a bit more, and think it would be nice to just drop this method and open up the `chooseSubnet` method as package-private and `@VisibleForTesting`.

I think we can get a similar level of readability by replacing the booleans with enums:
```java
public enum SubnetStrategy {
    PREFER_PUBLIC,
    DEFAULT;
}
```

```java
public Optional<CloudSubnet> chooseSubnet(String preferredSubnetId, Map<String, CloudSubnet> subnetMetas, String availabilityZone,
        boolean allowRandom) {
    return chooseSubnet(preferredSubnetId, subnetMetas, availabilityZone, allowRandom, SubnetStrategy.PREFER_PRIVATE);
}

@VisibleForTesting
Optional<CloudSubnet> chooseSubnetPreferPublic(String preferredSubnetId, Map<String, CloudSubnet> subnetMetas, String availabilityZone) {
    return chooseSubnet(preferredSubnetId, subnetMetas, availabilityZone, false, SubnetStrategy.DEFAULT);
}
```

```java
if (strategy == SubnetStrategy.PREFER_PUBLIC) {
```

And then dropping the additional methods.

brycederriso (approved):
Thanks for having a diff that removes more lines than it adds!

A few suggestions, but changes look good.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudSubnet.java
```diff
@@ -123,16 +115,8 @@ public void setType(SubnetType type) {
...
-    public boolean isRoutableToInternet() {
-        return routableToInternet;
-    }
-
-    public void setRoutableToInternet(boolean routableToInternet) {
-        this.routableToInternet = routableToInternet;
-    }
-
     public CloudSubnet withId(String newId) {
-        return new CloudSubnet(newId, name, availabilityZone, cidr, privateSubnet, mapPublicIpOnLaunch, igwAvailable, type, routableToInternet);
```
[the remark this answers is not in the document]
hreeve-cloudera: It was intended to mean that the subnet could receive traffic over the internet, even if there wasn't an internet gateway in its route table. It was originally added as part of CB-11070, but did not work as intended, and so it being removed now.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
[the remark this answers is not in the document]
hreeve-cloudera: `preferredSubnetId` is always null when it's called through `chooseSubnetPreferPublic`, but is not always null when  it's called through `chooseSubnet`.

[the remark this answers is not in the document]
hreeve-cloudera: I can change it to enum to make it more readable. But it's not clear to me what methods you're suggesting dropping. The method signatures in your example have the same accessibility of the existing ones, so I don't understand what's being changed outside of exchanging the boolean value for an enum.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
hreeve-cloudera: I believe they're the subnets ids. I didn't define the format of the Map, and am just using the format defined in `EnvironmentNetworkResponse.subnetMetas`.

Naming it `publicSubnetsMetadata` matches the pattern in `EnvironmentNetworkResponse`, which uses `Metas` instead of `Metadata`.

environment/src/main/java/com/sequenceiq/environment/environment/service/network/NetworkMetadataValidationService.java
```diff
@@ -54,13 +54,18 @@ public NetworkMetadataValidationService(CloudNetworkService cloudNetworkService,
     }
 
     private Map<String, CloudSubnet> removePrivateSubnets(Map<String, CloudSubnet> endpointGatewaySubnetMetas) {
-        LOGGER.debug("Removing any private subnets from the provided endpoint gateway list because they won't be used.");
-        if (endpointGatewaySubnetMetas == null || endpointGatewaySubnetMetas.isEmpty()) {
-            return Map.of();
+        if (entitlementService.endpointGatewayValidationEnabled(ThreadBasedUserCrnProvider.getAccountId())) {
```
biharitomi, line 57: Does it mean that every customer needs to have this entitlement as default?
As the original behaviour was the validation of the subnets.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
```diff
@@ -17,21 +22,42 @@
...
+        Optional<CloudSubnet> cloudSubnet = Optional.empty();
         if (StringUtils.isNotEmpty(preferredSubnetId)) {
             cloudSubnet = findSubnetById(subnetMetas, preferredSubnetId);
         } else if (StringUtils.isNotEmpty(availabilityZone)) {
             LOGGER.debug("Choosing subnet by availability zone {}", availabilityZone);
-            cloudSubnet = subnetMetas.values().stream()
-                .filter(s -> StringUtils.isNotEmpty(s.getAvailabilityZone()) &&
-                    s.getAvailabilityZone().equals(availabilityZone))
-                .findFirst();
+            if (preferPublic) {
```
biharitomi, line 45: Can we have debug logs here that indicates which path selected the the subnet based on what flags?

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
```diff
@@ -17,21 +22,42 @@
...
+                        s.getAvailabilityZone().equals(availabilityZone))
+                    .findFirst();
+            }
         } else if (allowRandom) {
             LOGGER.debug("Fallback to choose random subnet");
             cloudSubnet = subnetMetas.values().stream().findFirst();
-        } else {
-            cloudSubnet = Optional.empty();
         }
         return cloudSubnet;
```
biharitomi, line 62: Could you please add an info log what has been selected before the return statement?

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelectorTest.java
```diff
@@ -71,39 +80,98 @@ public void testChooseEndpointGatewaySubnetFromEnvironmentSubnetst() {
...
-            underTest.chooseSubnetForEndpointGateway(source, PRIVATE_ID_1);
+            ThreadBasedUserCrnProvider.doAs(USER_CRN, () -> underTest.chooseSubnetForEndpointGateway(source, PRIVATE_ID_1));
 
         assert subnet.isPresent();
         assertEquals(PUBLIC_ID_1, subnet.get().getId());
     }
 
     @Test
-    public void testChooseEndpointGatewaySubnetNoPublicEndpointSubnets() {
+    public void testChooseEndpointGatewaySubnetNoPublicEndpointSubnetsEntitlementEnabled() {
```
biharitomi, line 93: Shouldn't the test names be updated with new entitlement(`entitlementService.endpointGatewayValidationEnabled`)?

biharitomi (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/service/network/NetworkMetadataValidationService.java
hreeve-cloudera: The plan is that no customers will need this. By default we will not do the validation. However, @kmanamcheri suggested that instead of simply deleting the validation logic, we put it behind an entitlement, so that if it turns out it's better to have it than to not we can re-enable it quickly be enabling the entitlement. That's just a failsafe, and I don't anticipate we'll ever have to enable this entitlement.

I'm working on a separate effort to add new metric information around the endpoint gateway logic, so that we can see how customers are really using it and use that to see if we should bring back the validation logic in a different form in a later release. But that's not part of this, because I'm trying to get the initial version finished for Exxon.

core/src/test/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelectorTest.java
hreeve-cloudera: That would be more clear. Thanks for the suggestion.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/SubnetSelector.java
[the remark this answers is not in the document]
hreeve-cloudera: Added `SelectionFallbackStrategy` and `SelectionVisibilityStrategy` enums.

environment/src/main/java/com/sequenceiq/environment/environment/service/network/NetworkMetadataValidationService.java
hreeve-cloudera: Re: The original behavior, this logic is only called during initial environment creation. If any existing environments are using this feature, then they'll be past this part of the code already and won't revisit it. If a customer creates a new environment after this change, then we will not do the validation as part of that environment's creation.

biharitomi: Yeah but the validation is now disabled by default and the customer needs to ask for an entitlement to have the validation before env creation.
I would prefer to have the original behaviour which was the validation by default and have an entitlement for disabling it. 
@doktoric @lacikaaa @keyki @akanto What do you think?

biharitomi:
I checked the changes and looks okay, except the behaviour change of validation. That's the only open question.

environment/src/main/java/com/sequenceiq/environment/environment/service/network/NetworkMetadataValidationService.java
biharitomi: I think there will be env creation failures without the validation for customers who doesn't have extremely specific network setup and in this case they need to ask for an entitlement to have the validation on their env creations. I believe the better way is the opposite and have the validation by default but with entitlement we could disable it for customers with specific network setups.

doktoric: I think disable the validation by default is not a good idea because we preffer to prevent problems, with this solution we potentialy hide a problem in case of cluster deployment.

hreeve-cloudera: @biharitomi @doktoric I sent an email about why this design was chosen last Thursday. I realize that was right before your holidays so you may not have seen it. But we discussed this quite extensively in the semi-private network group, and it was decided that this is the option that makes the most sense right now.

1. The validation is only useful if it can accurately validate the network setup. This logic can't do that, because there isn't currently a way in CB to evaluate if a subnet has a route to the internet outside of its direct route table. That's the logic I tried to add in CB-11070, but it caused all those problems with the firewall service.
2. This is already an advanced network feature, and should only be used by advanced users who understand what it means for a subnet to have a route to the internet.
3. This validation is doing more harm than good. We only have evidence of one customer actively using this feature, and this validation has prevented them from being able to create an environment for over a month.
4. This validation can cause an environment to fail to be created. But if it's not used, then the environment will be created successfully. The worst case scenario in that situation is that the customer didn't use public subnets and so can't access their UIs and APIs from outside their internal network, but they can use every other feature of feature of CDP. So they are significantly less blocked even if they don't set up the public subnets correctly. And I'm going to work with our docs writer to document troubleshooting procedures for that potential scenario.

The reasons you're stating are the reasons I added the validation in the first place. And it's also why it's behind an entitlement, so that we can quickly re-enable it if we need to. But this is the approved design by the people who are responsible for and have worked on this feature.

doktoric: I would preffer 2 possible choice here:
- remove the validation completely 
- create a disable_validation entitlement and only grant for exxon

hreeve-cloudera:
@doktoric @biharitomi I've switched the logic so that it still uses the validation by default, but it can be disabled by granting the `CDP_ENDPOINT_GATEWAY_SKIP_VALIDATION` entitlement.

environment/src/main/java/com/sequenceiq/environment/environment/service/network/NetworkMetadataValidationService.java
lacikaaa: I agree with @doktoric . It doesn't make sense to add a code that is not used. I'm fine with dropping the validation if we can't make it work well/effectively now and we should be able to add it later.

> we can quickly re-enable it if we need to

what is the use-case for this?

hreeve-cloudera: I have already made the change to make the entitlement disable the validation, as per Richard's second suggestion.

biharitomi (approved):
LGTM
@hreeve-cloudera  thanks for picking our fundamental change requests on the behaviour of this feature!

PR #10418, opened 2021-04-06
CB-11769. Add new network checks for CDP nodestatus monitor.

oleewere:
details:
- update only models
- fix toString issue with RPCMessage during freeipa check
- skip neighbour scan for network checks
- include cloud specific network checks if the telemetry binary version is proper for that


freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -113,7 +113,7 @@ private void checkNodeHealthReports(Stack stack, Set<InstanceMetaData> checkable
     private void logReportResult(InstanceMetaData instanceMetaData, RPCResponse<NodeStatusProto.NodeStatusReport> nodeStatusRpcResponse, String reportType) {
         if (isSuccessfulRequest(nodeStatusRpcResponse)) {
             LOGGER.info("FreeIPA " + reportType + " reports for instance: [{}], report: [{}]",
-                    instanceMetaData.getInstanceId(), nodeStatusRpcResponse.getMessages().get(0));
+                    instanceMetaData.getInstanceId(), nodeStatusRpcResponse.getFirstTextMessage());
```
doktoric, line 116: I think a unit test would be nice

daszabo (approved):

PR #10419, opened 2021-04-06
CFM-1861 Nifi service data and settings are gone after datahub upgrad…

vbrodetskyi:
…ed from 7.2.6 to 7.2.9. Disabled file provider and enabled database provider, because NiFI Registry in CFM-2.2.0.0, should work with DB, on datahub. For private could it should work with file provider, that is why added changes into CB template.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-flow-management-small.bp
@@ -269,6 +269,14 @@
               {
                 "name": "xml.authorizers.userGroupProvider.composite-configurable-user-group-provider.property.User Group Provider 2",
                 "value": "shell-user-group-provider"
+              },
+              {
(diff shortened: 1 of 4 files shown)


adenes (approved):
thanks @vbrodetskyi , +1 for the change, it's in line with what's been discussed in the cfm data hub upgrade doc. cc @pvillard31

vbrodetskyi:
@doktoric could you please review/merge current changes? Also could you please clarify, is it enough to merge this change only to master?

doktoric (approved):

doktoric:
> @doktoric could you please review/merge current changes? Also could you please clarify, is it enough to merge this change only to master?

if you sen it to master then the next release will contains this fix. The branch cut will happen today.

vbrodetskyi:
> > @doktoric could you please review/merge current changes? Also could you please clarify, is it enough to merge this change only to master?
> 
> if you sen it to master then the next release will contains this fix. The branch cut will happen today.

@doktoric we need that change in 7.2.9. As i understand it's next release, so that's fine.

PR #10421, opened 2021-04-06
CB-11662 Environment service's start Freeipa flow step is not idempot…

topolyai5:
…ent. Step which is responsible for starting the related Freeipa is not idempotent and even the flow is restarted it fails

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/StartFreeIpaHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/StartFreeIpaHandler.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.environment.environment.flow.start.handler;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.environment.environment.EnvironmentStatus;
@@ -20,6 +22,8 @@ import reactor.bus.Event;
 @Component
 public class StartFreeIpaHandler extends EventSenderAwareHandler<EnvironmentStartDto> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(StartFreeIpaHandler.class);
+
     private final FreeIpaPollerService freeIpaPollerService;
 
     private final FreeIpaService freeIpaService;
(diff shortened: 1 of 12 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
```diff
@@ -43,12 +43,16 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
         EnvironmentDto environmentDto = environmentDtoEvent.getData();
         try {
             if (Strings.isNullOrEmpty(environmentDto.getParentEnvironmentCrn())) {
-                freeIpaService.describe(environmentDto.getResourceCrn()).ifPresent(freeIpa -> {
-                    if (freeIpa.getStatus() != null && !freeIpa.getStatus().isStoppable()) {
+                freeIpaService.describe(environmentDto.getResourceCrn()).ifPresentOrElse(freeIpa -> {
+                    if (freeIpa.getStatus() != null && (freeIpa.getStatus().isStoppedPhase() || freeIpa.getStatus().isStopInProgressPhase())) {
+                        LOGGER.info("Freeipa is already {}, so it won't started", freeIpa.getStatus());
```
biharitomi, line 48: Grammar: `it won't be started` I think it should be something like `Stop has already been trigger continuing without new stop trigger.

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
```diff
@@ -43,12 +43,16 @@ public void accept(Event<EnvironmentDto> environmentDtoEvent) {
...
+                    } else if (freeIpa.getStatus() != null && !freeIpa.getStatus().isStoppable()) {
```
biharitomi, line 49: I would move the null check one layer up as it used in two condition paths and in case of stop there is no status for the FreeIPA stack that should be a critical error.

environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/StartFreeIpaHandler.java
```diff
@@ -39,12 +43,16 @@ public String selector() {
     public void accept(Event<EnvironmentStartDto> environmentStartDtoEvent) {
         EnvironmentDto environmentDto = environmentStartDtoEvent.getData().getEnvironmentDto();
         try {
-            freeIpaService.describe(environmentDto.getResourceCrn()).ifPresent(freeIpa -> {
-                if (freeIpa.getStatus() != null && !freeIpa.getStatus().isStartable()) {
+            freeIpaService.describe(environmentDto.getResourceCrn()).ifPresentOrElse(freeIpa -> {
+                if (freeIpa.getStatus() != null && (freeIpa.getStatus().isAvailable() || freeIpa.getStatus().isStartInProgressPhase())) {
+                    LOGGER.info("FreeIPA is already {}, so it won't started", freeIpa.getStatus());
```
biharitomi, line 48: Grammar: `it won't be started` I think it should be something like `Start has already been trigger continuing without new start trigger.

environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/StartFreeIpaHandler.java
```diff
@@ -39,12 +43,16 @@ public String selector() {
...
+                } else if (freeIpa.getStatus() != null && !freeIpa.getStatus().isStartable()) {
```
biharitomi, line 49: I would move the null check one layer up as it used in two condition paths and in case of start if there is no status for the FreeIPA stack that should be a critical error.

biharitomi (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
topolyai5: I didn't want to add a new level to here

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
topolyai5: I didn't want to add a new level to here

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
topolyai5: This message does not contain the status. I think it is a useful information

environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/StartFreeIpaHandler.java
topolyai5: Like the previous version, we tried to start the freeipa event if the status was null, I didn't check this. 

I don't know it is a good choice  or not to interrupt the env flow

environment/src/main/java/com/sequenceiq/environment/environment/flow/stop/handler/StopFreeIpaHandler.java
biharitomi: I meant only the last sentence which was faulty.

biharitomi (approved):

PR #10425, opened 2021-04-07
CB-11252 Repair puts DL in an inconsistent state if the DB is stopped

lturcsanyi:
Before starting a repair the external database state should be validated.
Minor copy-paste naming fixes.


horadla23 (approved):
LGTM

PR #10431, opened 2021-04-07
Update GCP templates for performance

cegganesh84:
1. DISTX-557 -> To overcome memory overcommit problem in default DE template AWS and Azure was already moved to 64 GB memory, this is the GCP followup
2. DISTX-575 -> DE AWS and Azure was already handled to move to SSDs by default. The GCP DE HA template was also updated. This is just the GCP followup for other templates.
3. CB-9521 -> Moved the root volume of GCP to SSD, this should help in custom parcel handling, among other perf improvement and gets it on par with AWS and Azure.


sidseth (approved):
+1. LGTM

PR #10432, opened 2021-04-07
CFM-1861 Nifi service data and settings are gone after datahub upgrad…

vbrodetskyi:
…ed from 7.2.6 to 7.2.9. Added workaround for scaling up NiFi node (TLS/SSL properties and NiFi node with cardinality=0), because it's still not fixed in CM/CDH.


pvillard31:
LGTM - are we sure this can still get into 7.2.9 release?

doktoric (approved):

PR #10433, opened 2021-04-08
DISTX-602 DH Autoscaling Resource Level Authorization

smaniraju:
DISTX-602 DH Autoscaling should use Resource Level Authorization …

DH Autoscaling should only use resource level authorization checks, since Account Level authorization has been deprecated.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
@@ -17,8 +17,11 @@ import org.slf4j.LoggerFactory;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.stereotype.Controller;
 
-import com.sequenceiq.authorization.annotation.CheckPermissionByAccount;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
```diff
@@ -175,7 +178,8 @@ public AutoscaleRecommendationV4Response getRecommendation(String crn) {
     }
 
     @Override
-    @CheckPermissionByAccount(action = AuthorizationResourceAction.DATAHUB_READ)
+    @DisableCheckPermissions
+    @PreAuthorize("hasRole('AUTOSCALE')")
```
horadla23, line 182: have you tested this? in the future we will remove this `hasRole('AUTOSCALE')` and switch to internal api and periscope can call it internally. But for now, this will be good I think, just want to make sure that you have tested it and normal user (not periscope, not poweruser) cannot call this method

smaniraju: yes all **_hasRole('AUTOSCALE')_** methods are only called with internalcrn of periscope and these api are working.

horadla23 (approved):

PR #10435, opened 2021-04-08
CB-11658. Create config that holds common configurations for all serv…

oleewere:
…ices.

details:
- create new CommonConfig that holds common static configurations. (instead of any hard coded constant that is the same for every services)
- use common config with spec. objects for databus s3 endpoints.
- use the same way of implementation that experiences config used.
- update UTs


lnardai (approved):
Looks like a good start, we need a similar functionality since there are lot of configurations other than this that should be the same for all services.
I'm kind of on board with this, but let's wait one more approval.

attilapalfi92 (approved):

PR #10436, opened 2021-04-08
CB-7972 Increase datalake start timeout to 120 min

lturcsanyi:
There may be a situation where a 40 min timeout is not enough. For example if something is slow on the provider side, the polling could timeout on datalake side before the flow in the core could finish, thus the stack remains in UPDATE_IN_PROGRESS state. Increasing the timeout lets CB to finish the flow. Since we have flow tracking it should not cause any problems.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopWaitHandler.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopWaitHandler.java
@@ -6,6 +6,7 @@ import javax.inject.Inject;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import com.dyngr.exception.PollerException;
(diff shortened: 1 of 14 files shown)


sodre90:
please increase timeout in every SDX flow

lturcsanyi:
As agreed, all datalake polling timeout values can be configured now.

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/handler/SdxStopWaitHandler.java
```diff
@@ -24,11 +25,13 @@
...
 
-    public static final int SLEEP_TIME_IN_SEC = 20;
+    private static final Logger LOGGER = LoggerFactory.getLogger(SdxStopWaitHandler.class);
 
-    public static final int DURATION_IN_MINUTES = 40;
+    @Value("${sdx.stack.stop.sleeptime_sec:20}")
+    private int sleepTimeInSec;
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(SdxStopWaitHandler.class);
+    @Value("${sdx.stack.stop.duration_min:40}")
```
lnardai, line 33: Might worth to increase this as well with the Start value.

lnardai (approved):
LGTM!

PR #10443, opened 2021-04-08
CB-11964 Track previous flow state on finished and failed messages

nadamdb:

### DIFF
--- a/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentRequestProcessingStepMapper.java
+++ b/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentRequestProcessingStepMapper.java
@@ -25,6 +25,6 @@ public class EnvironmentRequestProcessingStepMapper {
     }
 
     public boolean isLastStep(FlowDetails flow) {
-        return "FINAL_STATE".equals(flow.getNextFlowState());
+        return flow.getNextFlowState() != null && (flow.getNextFlowState().endsWith("_FAILED_STATE") || flow.getNextFlowState().endsWith("_FINISHED_STATE"));
     }
 }
--- a/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentUseCaseMapper.java
+++ b/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentUseCaseMapper.java
@@ -20,11 +20,14 @@ public class EnvironmentUseCaseMapper {
     // At the moment we need to introduce a complex logic to figure out the use case
     public UsageProto.CDPEnvironmentStatus.Value useCase(FlowDetails flow) {
         UsageProto.CDPEnvironmentStatus.Value useCase = UsageProto.CDPEnvironmentStatus.Value.UNSET;
-        if (environmentRequestProcessingStepMapper.isFirstStep(flow)) {
-            useCase = firstStepToUseCaseMapping(flow.getFlowType());
-        } else if (environmentRequestProcessingStepMapper.isLastStep(flow)) {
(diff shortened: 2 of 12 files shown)


structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentRequestProcessingStepMapper.java
```diff
@@ -25,6 +25,6 @@ public boolean isFirstStep(FlowDetails flow) {
     }
 
     public boolean isLastStep(FlowDetails flow) {
-        return "FINAL_STATE".equals(flow.getNextFlowState());
+        return flow.getNextFlowState() != null && (flow.getNextFlowState().endsWith("_FAILED_STATE") || flow.getNextFlowState().endsWith("_FINISHED_STATE"));
```
akanto, line 28: This is going to miss a few flows, in some cases the flows are not ending with FAILED_STATE or FINISHED_STATE, but some cases they are ending with FINALIZE etc.

structuredevent-service-legacy/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/ClusterRequestProcessingStepMapper.java
```diff
@@ -25,6 +25,9 @@ public boolean isFirstStep(FlowDetails flow) {
     }
 
     public boolean isLastStep(FlowDetails flow) {
-        return "FINAL_STATE".equals(flow.getNextFlowState());
+        return flow.getNextFlowState() != null
+                && (flow.getNextFlowState().contains("_FAIL")
```
akanto, line 29: I guess, the _FAILED_STATE is missing

akanto:
I think that I don't agree with the concept. My preference is to always rely on "FINAL_STATE", because if you rely on that you cannot miss the exit of the flow.

I think I can agree, with the concept, if we are enforcing that flow can exit only with "_FAILED_STATE" and "_FINISHED_STATE", but even if we do it in this way, you cannot guarantee that validation flow steps are named uniformly.

akanto (changes requested):
I think that I don't agree with the concept. My preference is to always rely on "FINAL_STATE", because if you rely on that you cannot miss the exit of the flow.

I think I can agree, with the concept, if we are enforcing that flow can exit only with "_FAILED_STATE" and "_FINISHED_STATE", but even if we do it in this way, you cannot guarantee that validation flow steps are named uniformly.

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentUseCaseMapper.java
```diff
@@ -51,9 +54,9 @@
         return useCase;
     }
 
-    private UsageProto.CDPEnvironmentStatus.Value lastStepToUseCaseMapping(String flowState) {
+    private UsageProto.CDPEnvironmentStatus.Value lastStepToUseCaseMapping(String nextFlowState) {
         UsageProto.CDPEnvironmentStatus.Value useCase = UsageProto.CDPEnvironmentStatus.Value.UNSET;
-        switch (flowState) {
+        switch (nextFlowState) {
```
akanto, line 59: Might be confusing because in such a case the "UMS_RESOURCE_DELETE_STARTED_STATE" will be written to EDH instead of ENV_DELETE_FINISHED_STATE.

structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/service/telemetry/mapper/EnvironmentRequestProcessingStepMapper.java
akanto: Oh, this is Environment, so we might be good here.

akanto (approved):

PR #10444, opened 2021-04-08
CB-11843 Create new endpoint and flows for creating ldap/kerberos config

lacikaaa:
A new flow and endpoint introduced to create bind users used by clusters.
In this implentation if the bind user already exists on the FreeIPA we will replace it's password, while in the old implementation
this case resulted in an exception.
Requests are rejected using operation to ensure only one flow is running per cluster.
If the configs already exists the flow/operation is still started, but would finish faster as no connection to FreeIPA is necessary.
The status of the flow can be tracked by polling the operation status. Operation would be rejected if there is already a running operation
in the same environment for the same cluster. If the flow can't be started the operation will have failed state.
If the flow fails the operation will have failed state also.

### DIFF
--- a/flow/src/main/java/com/sequenceiq/flow/reactor/api/handler/HandlerEvent.java
+++ b/flow/src/main/java/com/sequenceiq/flow/reactor/api/handler/HandlerEvent.java
@@ -1,0 +1,34 @@
+package com.sequenceiq.flow.reactor.api.handler;
+
+import reactor.bus.Event;
+
+public class HandlerEvent<T> {
+
+    private Event<T> event;
+
+    private int eventCounter;
+
+    public HandlerEvent(Event<T> event) {
+        this.event = event;
+    }
+
+    public Event<T> getEvent() {
+        return event;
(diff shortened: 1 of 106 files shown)


flow/src/main/java/com/sequenceiq/flow/reactor/api/handler/HandlerEvent.java
```diff
@@ -0,0 +1,34 @@
+package com.sequenceiq.flow.reactor.api.handler;
+
+import reactor.bus.Event;
+
+public class HandlerEvent<T> {
+
+    private Event<T> event;
+
+    private int eventCounter;
```
keyki, line 9: Should this be atomic?

lacikaaa: haven't modified this, it's extracted as it was as an inner class

horadla23 (commented):
authz part looks good

PR #10447, opened 2021-04-08
CB-11981: Fixed YCloud datalake creation NPE due to not checking if stack network is null.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-11981

Simply added check to ensure network is not null before trying to access the internal CIDRs.


sxxgrc:
... did not mean to merge from branch hold up.

PR #10452, opened 2021-04-09
CB-11802 Rollback of failed upscale resources due to AWS quota limit fails when there is activity still in progress on the auto scaling group

foldik:


foldik:
Tested with fault injection. After a few retries it successfully recovered the original state.

topolyai5 (approved):

PR #10456, opened 2021-04-09
CB-11992:  Add zookeeper service to Datamart templates for 7.2.9 and …

doktoric:
…later

    Add instance of zookeeper service as required by Knox

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-mart.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-mart.bp
@@ -21,6 +21,23 @@
         ]
       },
       {
+        "refName": "zookeeper",
+        "serviceType": "ZOOKEEPER",
+        "serviceConfigs": [
+          {
+            "name": "service_config_suppression_server_count_validator",
+            "value": "true"
+          }
+        ],
+        "roleConfigGroups": [
+          {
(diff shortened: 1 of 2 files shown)


kmanamcheri:
@doktoric please make this template change for 7.2.10 as well. I found a few more issues where we made changes to 7.2.9 but didn't make it in 7.2.10 and we run into problems again.

doktoric:
@drorke @smolnar82 do we need this for 7.2.10 ?

smolnar82:
@lmccay We need this for 7.2.9, right?

smolnar82:
@doktoric Is your question if we need ZK for Knox in future releases? I believe the answer is yes.

drorke:
The commit already updates the 7.2.10 version of the blueprint (in addition to 7.2.9).  Is there anything else that would need to be done to get it into 7.2.10?

lmccay:
> @lmccay We need this for 7.2.9, right?

Yes, this is required for 7.2.9 for UBS token based access to JDBC. It is customer committed and needs the storage durability provided by ZK.

kmanamcheri:
To reiterate the plan, this is not a blocker for 2.41. This is a blocker for 7.2.9. CB 2.41 is currently in promotion stages and once that goes out, we'll merge this and get this out as a hotfix to support 7.2.9.

PR #10457, opened 2021-04-09
CB-11997 Provide additional internal endpoints for COD based on authZ…

horadla23:
… requirement


horadla23:
retest this please

foldik (approved):
If this is all the endpoints that we need to implement then it's fine for me.

PR #10459, opened 2021-04-09
CB-11514 DFX to use redbeams to create databaseServers

gergopapi2:
When redbeams databaseserver create endpoint is called, a clusterCrn has to be provided. Currently it is allowed to be datalake or datahub only. As a result of that services, that do not posses such CRNs are not allowed to call redbeams.

The present commit allows a databaseServer to be created with any CRN except of an environmentCrn.

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/validation/CrnValidator.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/validation/CrnValidator.java
@@ -12,16 +12,17 @@ import com.sequenceiq.cloudbreak.auth.altus.Crn;
 public class CrnValidator extends AbstractCrnValidator<String> {
 
     @Override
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/sequenceiq/cloudbreak/validation/CrnValidator.java
```diff
@@ -12,16 +12,17 @@
 public class CrnValidator extends AbstractCrnValidator<String> {
 
     @Override
-    protected String getErrorMessageIfServiceOrResourceTypeInvalid(String req, Set<Pair> serviceAndResourceTypePairs) {
```
doktoric, line 15: any unit test ?

gergopapi2: It is transitively tested in CrnValidatorTest (I extended that), but can add tests for this method directly as well.

PR #10460, opened 2021-04-10
DISTX-603 For AWS DE HA from 7.2.9 remove the Knox service from non gateway nodes

cegganesh84:
1. If not removed, the load-balancer would still route the requests to master nodegroup but there is no Nginx running there. So the user would get connection refused for a few tries. The load-balancer checks Knox port instead of Nginx port. Nginx would run only in one GATEWAY nodegroup.
2. This along with the fix for DISTX-604 must be released as hot fix before 7.2.9 release. So I will make a separate PR for 2.41 branch for this.


cegganesh84:
Architecture diagram for reference https://cloudera.atlassian.net/wiki/spaces/ENG/pages/1621459043/Data+Hub+Data+Engineering+High+Availability

hreeve-cloudera (approved):

sidseth (approved):
Looks good. Understand this a little better from an offline conversation.

PR #10461, opened 2021-04-12
CB-12000-allow-localhost-on-dl-dx-nginx

bergerdenes:

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/nginx/conf/ssl.conf
+++ b/orchestrator-salt/src/main/resources/salt/salt/nginx/conf/ssl.conf
@@ -3,6 +3,7 @@ server {
     {%- if salt['pillar.get']('gateway:enable_ccmv2', False) %}
         {% if salt['pillar.get']('gateway:cidrBlocks') %}
             listen       9443;
+            allow        127.0.0.1;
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/nginx/conf/ssl.conf
```diff
@@ -3,6 +3,7 @@ server {
     {%- if salt['pillar.get']('gateway:enable_ccmv2', False) %}
         {% if salt['pillar.get']('gateway:cidrBlocks') %}
             listen       9443;
+            allow        127.0.0.1;
```
lacikaaa, line 6: same for freeipa is missing

lacikaaa (changes requested):

orchestrator-salt/src/main/resources/salt/salt/nginx/conf/ssl.conf
bergerdenes: nope, it is already there

lacikaaa: thanks!

lacikaaa (approved):

PR #10466, opened 2021-04-12
CB-11963 Handle recipe execution failure on hosts without metadata or…

Bajzathd:
… fqdn


bbihari (approved):
LGTM

PR #10467, opened 2021-04-12
CB-11889 add Retryable annotation to host orchestrator teardown

sodre90:


doktoric:
the target branch is 2.41 the ticket is related to 2.42. Also some unit test would be nice

sodre90:
I rebased the PR to master because 2.41 will not be shipped

PR #10484, opened 2021-04-14
CB-11996 Support Runtime upgrade without OS upgrade (image swap) for all DH clusters

schfeca75:
CB-11996 Support Runtime upgrade without OS upgrade (image swap) for all DH clusters

* In case of distrox runtime upgrade the replaceVms parameter in the request will be overridden to false.
  This means that the vm replacement will be always skipped in case of distrox runtime upgrade.
* In case of distrox os upgrade the replaceVms parameter won't be changed,
  so the vm replacement can be controllable via this parameter.

### DIFF
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeService.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeService.java
@@ -9,21 +9,29 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
-import com.sequenceiq.cloudbreak.auth.ClouderaManagerLicenseProvider;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.dto.NameOrCrn;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.StackImageChangeV4Request;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeService.java
```diff
@@ -109,4 +139,24 @@ private void verifyCMLicenseValidity(String userCrn) {
...
+            try {
+                Image currentImage = componentConfigProviderService.getImage(stack.getId());
+                CloudbreakImageCatalogV3 imageCatalog = imageCatalogProvider.getImageCatalogV3(currentImage.getImageCatalogUrl());
+                com.sequenceiq.cloudbreak.cloud.model.catalog.Image currentCatalogImage =
+                        imageProvider.getCurrentImageFromCatalog(currentImage.getImageId(), imageCatalog);
+                com.sequenceiq.cloudbreak.cloud.model.catalog.Image targetCatalogImage =
+                        imageProvider.getCurrentImageFromCatalog(targetImage.getImageId(), imageCatalog);
+                if (!lockedComponentChecker.isUpgradePermitted(
+                        currentCatalogImage, targetCatalogImage, imageFilterParamsFactory.getStackRelatedParcels(stack))) {
+                    LOGGER.info("ReplaceVms parameter has been overridden to false in case of distrox runtime upgrade.");
```
keyki, line 154: Can you give more context in the log, like CRN or cluster name and what is the source and target image id?

PR #10491, opened 2021-04-15
CB-11747 External database create flow is not restartable

bergerdenes:
In case flow was stopped and then restarted just after the external
DB provision started, it might happen that upon launching a new
create request to RedBeams caused and exception. This is fixed now.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
@@ -5,6 +5,7 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import javax.ws.rs.BadRequestException;
@@ -64,17 +65,24 @@ public class ExternalDatabaseService {
 
     public void provisionDatabase(Cluster cluster, DatabaseAvailabilityType externalDatabase, DetailedEnvironmentResponse environment) {
         LOGGER.info("Create external {} database server in environment {} for DataHub {}", externalDatabase.name(), environment.getName(), cluster.getName());
-        AllocateDatabaseServerV4Request request = getDatabaseRequest(environment, externalDatabase, cluster);
         String databaseCrn;
         try {
-            databaseCrn = redbeamsClient.create(request).getResourceCrn();
+            Optional<DatabaseServerV4Response> existingDatabase = findExistingDatabase(cluster, environment.getCrn());
+            if (existingDatabase.isPresent()) {
+                String dbCrn = existingDatabase.get().getCrn();
+                LOGGER.debug("Found existing database with CRN {}", dbCrn);
+                databaseCrn = dbCrn;
+            } else {
+                LOGGER.debug("Requesting new database server creation");
+                AllocateDatabaseServerV4Request request = getDatabaseRequest(environment, externalDatabase, cluster);
+                databaseCrn = redbeamsClient.create(request).getResourceCrn();
+            }
             updateClusterWithDatabaseServerCrn(cluster, databaseCrn);
         } catch (BadRequestException badRequestException) {
             LOGGER.error("Redbeams create request failed, bad request", badRequestException);
             throw badRequestException;
         }
         waitAndGetDatabase(cluster, databaseCrn, DatabaseOperation.CREATION, true);
-
     }
 
     public void terminateDatabase(Cluster cluster, DatabaseAvailabilityType externalDatabase, DetailedEnvironmentResponse environment, boolean forced) {
@@ -122,6 +130,21 @@ public class ExternalDatabaseService {
         } catch (NotFoundException notFoundException) {
             LOGGER.info("Database server not found on redbeams side {}", databaseCrn);
         }
+    }
+
+    private Optional<DatabaseServerV4Response> findExistingDatabase(Cluster cluster, String environmentCrn) {
+        if (cluster.getStack() != null) {
+            String clusterCrn = cluster.getStack().getResourceCrn();
+            LOGGER.debug("Trying to find existing database server for environment {} and cluster {}", environmentCrn, clusterCrn);
+            try {
+                return Optional.ofNullable(redbeamsClient.getByClusterCrn(environmentCrn, clusterCrn));
+            } catch (NotFoundException ignore) {
+                LOGGER.debug("External database in environment {} for cluster {} does not exist.", environmentCrn, cluster);
+                return Optional.empty();
+            }
+        }
+        LOGGER.warn("[INVESTIGATE] Stack is empty for cluster '{}' in environment {}", cluster.getName(), environmentCrn);
+        return Optional.empty();
     }
 
     private boolean externalDatabaseReferenceExist(String databaseCrn) {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/RedbeamsClientService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/RedbeamsClientService.java
@@ -5,6 +5,7 @@ import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
@@ -76,6 +77,30 @@ public class RedbeamsClientService {
             String message = String.format("Failed to stop DatabaseServer with CRN %s", crn);
             LOGGER.error(message, e);
             throw new CloudbreakServiceException(message, e);
+        }
+    }
+
+    public DatabaseServerV4Response getByClusterCrn(String environmentCrn, String clusterCrn) {
+        validateForGetByClusterCrn(environmentCrn, clusterCrn);
+        try {
+            return ThreadBasedUserCrnProvider.doAsInternalActor(() -> redbeamsServerEndpoint.getByClusterCrn(environmentCrn, clusterCrn));
+        } catch (NotFoundException e) {
+            String message = String.format("DatabaseServer with Environment CRN %s and Cluster CRN %s was not found", environmentCrn, clusterCrn);
+            LOGGER.debug(message, e);
+            throw e;
+        } catch (WebApplicationException | ProcessingException e) {
+            String message = String.format("Failed to get DatabaseServer with Environment CRN %s and Cluster CRN %s", environmentCrn, clusterCrn);
+            LOGGER.error(message, e);
+            throw new CloudbreakServiceException(message, e);
+        }
+    }
+
+    private void validateForGetByClusterCrn(String environmentCrn, String clusterCrn) {
+        if (StringUtils.isBlank(environmentCrn)) {
+            throw new CloudbreakServiceException("Environment CRN is empty");
+        }
+        if (StringUtils.isBlank(clusterCrn)) {
+            throw new CloudbreakServiceException("Cluster CRN is empty");
         }
     }
 }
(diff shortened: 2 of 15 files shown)


redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
```diff
@@ -77,6 +77,18 @@ DatabaseServerV4Response getByName(
...
+    @GET
+    @Path("clusterCrn/{clusterCrn}")
+    @ApiOperation(value = DatabaseServerOpDescription.GET_BY_CLUSTER_CRN,
+            notes = DatabaseServerNotes.GET_BY_CLUSTER_CRN,
+            nickname = "getDatabaseServerByClusterCrn")
+    DatabaseServerV4Response getByClusterCrn(
+            @ValidCrn(resource = CrnResourceDescriptor.ENVIRONMENT) @NotNull @ApiParam(value = DatabaseServerParamDescriptions.ENVIRONMENT_CRN, required = true)
+            @QueryParam("environmentCrn") String environmentCrn,
+            @ValidCrn(resource = { CrnResourceDescriptor.DATALAKE, CrnResourceDescriptor.DATAHUB })
+            @NotNull @ApiParam(value = DatabaseServerParamDescriptions.NAME, required = true) @PathParam("clusterCrn") String clusterCrn
```
lajosrodek, line 89: Note: I think this will have to be updated soon to permit any non-environment CRNs for sake of looking up DB servers for non-CB "clusters" (e.g. DFX). See [this](https://github.com/hortonworks/cloudbreak/pull/10459/files#diff-11d87b3ace4084117083bb0317790b6c0f709d00603e662c2a6d9581247ce5c0) change in #10459.

CC @gergopapi2

bergerdenes: but does it block the merge?

lajosrodek: Certainly not, I just flagged it as something to be fixed once both PRs got merged. 😄

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -124,6 +133,20 @@ public void stopDatabase(Cluster cluster, DatabaseAvailabilityType externalDatab
         }
     }
 
+    private Optional<DatabaseServerV4Response> findExistingDatabase(Cluster cluster, String environmentCrn) {
+        Optional<DatabaseServerV4Response> existingDatabase = Optional.empty();
```
lacikaaa, line 137: drop this line

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -124,6 +133,20 @@ public void stopDatabase(Cluster cluster, DatabaseAvailabilityType externalDatab
...
+        if (cluster.getStack() != null) {
+            String clusterCrn = cluster.getStack().getResourceCrn();
+            LOGGER.debug("Trying to find existing database server for environment {} and cluster {}", environmentCrn, clusterCrn);
+            try {
+                existingDatabase = Optional.ofNullable(redbeamsClient.getByClusterCrn(environmentCrn, clusterCrn));
```
lacikaaa, line 142: add return here

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -124,6 +133,20 @@ public void stopDatabase(Cluster cluster, DatabaseAvailabilityType externalDatab
...
+            } catch (NotFoundException ignore) {
+                // it means DB does not exist
```
lacikaaa, line 144: add debug or info log here and return with `Optional#empty`

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -67,7 +68,15 @@ public void provisionDatabase(Cluster cluster, DatabaseAvailabilityType external
...
         try {
-            databaseCrn = redbeamsClient.create(request).getResourceCrn();
+            Optional<DatabaseServerV4Response> existingDatabase = findExistingDatabase(cluster, environment.getCrn());
+            if (existingDatabase.isPresent()) {
+                String dbCrn = existingDatabase.get().getCrn();
+                LOGGER.debug("Found existing database with CRN {}", dbCrn);
+                databaseCrn = dbCrn;
+            } else {
+                LOGGER.debug("Requesting new database server creation");
+                databaseCrn = redbeamsClient.create(request).getResourceCrn();
```
lacikaaa, line 78: request creation should be moved right before this line

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
```diff
@@ -67,7 +68,15 @@ public void provisionDatabase(Cluster cluster, DatabaseAvailabilityType external
         AllocateDatabaseServerV4Request request = getDatabaseRequest(environment, externalDatabase, cluster);
         String databaseCrn;
```
lacikaaa, line 69: this should be inside the try. also would be nicer if the `if` part would be in a separate method and there would be only one value assignment for this variable

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/RedbeamsClientService.java
```diff
@@ -78,4 +79,24 @@ public void stopByCrn(String crn) {
...
         }
     }
+
+    public DatabaseServerV4Response getByClusterCrn(String environmentCrn, String clusterCrn) {
+        if (Strings.isNullOrEmpty(environmentCrn)) {
+            throw new CloudbreakServiceException("Environment CRN is empty");
+        }
+        if (Strings.isNullOrEmpty(clusterCrn)) {
+            throw new CloudbreakServiceException("Cluster CRN is empty");
+        }
```
lacikaaa, line 89: move to a validate method please

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
lacikaaa: as the API can't be modified later I would say it would be better to call this resourceCrn or similar which can be meaningful for other services. But I will leave it to @gergopapi2

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
```diff
@@ -220,8 +220,18 @@ public DatabaseServerConfig getByCrn(String resourceCrn) {
         return repository.findByResourceCrn(resourceCrn);
     }
 
-    public Optional<DatabaseServerConfig> getByClusterCrn(String clusterCrn) {
-        return repository.findByClusterCrn(clusterCrn);
+    public DatabaseServerConfig getByClusterCrn(String environmentCrn, String clusterCrn) {
+        Optional<DatabaseServerConfig> resourceOpt = getByClusterCrnOptional(environmentCrn, clusterCrn);
+        if (resourceOpt.isEmpty()) {
```
lacikaaa, line 225: you could use `Optional#orElseThrow` to get the value and throw exception

lacikaaa (changes requested):
please add unit test for the changes in `ExternalDatabaseService`

redbeams-api/src/main/java/com/sequenceiq/redbeams/api/endpoint/v4/databaseserver/DatabaseServerV4Endpoint.java
doktoric: resourceCrn would be nicer yes because other services also want to use redbeams

gergopapi2: I do agree, however the name clusterCrn is already out in the AllocateDatabaseServerV4Request. For consistency it would make sense to keep the name I think.

bergerdenes: leaved as `clusterCrn` but changed the `@ValidCrn` as per @lajosrodek  suggested

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
bergerdenes: if possible I like the single point of return, but, whatever. 🤷🏼‍♂️

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
bergerdenes: nit, but ok

core/src/main/java/com/sequenceiq/cloudbreak/service/rdsconfig/RedbeamsClientService.java
bergerdenes: ok

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
bergerdenes: not possible to move the variable declaration into the `try` block as the `waitAndGetDatabase` needs it too and that is outside the try `block`

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/externaldatabase/ExternalDatabaseService.java
bergerdenes: ok

bergerdenes:
unit test added

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
```diff
@@ -220,8 +220,16 @@ public DatabaseServerConfig getByCrn(String resourceCrn) {
...
-        return repository.findByClusterCrn(clusterCrn);
+    public DatabaseServerConfig getByClusterCrn(String environmentCrn, String clusterCrn) {
+        DatabaseServerConfig databaseServerConfig = getByClusterCrnOptional(environmentCrn, clusterCrn)
+                .orElseThrow(() -> new NotFoundException(String.format("No %s found with cluster CRN '%s' in environment '%s'",
+                        DatabaseServerConfig.class.getSimpleName(), clusterCrn, environmentCrn)));
+        MDCBuilder.buildMdcContext(databaseServerConfig);
+        return databaseServerConfig;
+    }
+
+    public Optional<DatabaseServerConfig> getByClusterCrnOptional(String environmentCrn, String clusterCrn) {
```
lacikaaa, line 231: it would be nice to follow the repository's pattern here: `find` returns `Optional`, while `get` returns the pojo and fails in case there is none. so the method name would look like here: `findByEnvironmentIdAndClusterCrn`, same as the repo

lacikaaa (approved):
that only naming nit is left, otherwise LGTM, feel free to merge after fixing that one. thanks for the changes

redbeams/src/main/java/com/sequenceiq/redbeams/service/dbserverconfig/DatabaseServerConfigService.java
doktoric: findByEnvironmentCrnAndClusterCrn right ?

lacikaaa: right, I copied the repository method name. unfortunately in redbeams it called `environmentId` instead of crn...

doktoric: I see you @lacikaaa . I see you ...

PR #10493, opened 2021-04-16
DISTX-610 Revert : DISTX-595 Release a first version of Oozie HA without the in…

cegganesh84:
…ternal load-balancer

1. In public cloud all the requests go through Knox, so we do not need a load-balancer for external clients.
3. This is for Hue to access the Oozie in HA fashion.
4. Hue can not use Knox for this because the URL is different from what Hue expects it to be.
5. As a workaround use the first FQDN in the master hosts to configure it as a load-balancer to itself.
6. This means Oozie will be HA for the schedules, but can not accept modifications from Hue when the first FQDN in the master host is down. It still can accept modifications from external clients via Knox.
7. Added a TODO pointing to the parent task to add an internal load-balancer for Hue to work with Oozie in HA fashion.

This commit reverts 9b256f9d87a19054746dc7aab63e9f1bb133e563


sidseth (approved):
Looks good. I think this rolls back the Oozie aspects only.

PR #10495, opened 2021-04-18
CB-12039 Add default freeipa image to the core svc and merge custom image with the source image where the source image can be a FreeIPA or a Cloudbreak image as well

pkedvessy:


core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageType.java
```diff
@@ -0,0 +1,31 @@
...
+
+public enum ImageType {
+    FREEIPA,
+    DATAHUB,
+    DATALAKE,
+    INVALID;
+
+    private static final ImmutableMap<String, ImageType> VALUEMAP;
+
+    static {
```
pkedvessy, line 14: We have a generic implementation for this: DefaultEnumConverter
You should create a converter and reference it in the entity as a @Converter so hibernate will know how to map String to ImageType and vice versa. Sample what you can use as a pattern:  CommonStatusConverter

core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
```diff
@@ -95,6 +97,8 @@
 
     private static final String V2_CATALOG_FILE = "com/sequenceiq/cloudbreak/service/image/cb-image-catalog-v2.json";
 
+    private static final String V3_CATALOG_FILE = "com/sequenceiq/cloudbreak/service/image/cb-image-catalog-v3.json";
```
pkedvessy, line 100: Please test the solution with a freeipa image catalog as well. It is needed as freeipa contains less parameters for a particular image so we should make sure we won't fail on it eg by a NPE.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/imagecatalog/ImageToImageV4ResponseConverter.java
```diff
@@ -0,0 +1,37 @@
...
+import com.sequenceiq.cloudbreak.converter.AbstractConversionServiceAwareConverter;
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+
+@Component
+public class ImageToImageV4ResponseConverter extends AbstractConversionServiceAwareConverter<Image, ImageV4Response> {
+
+    @Override
+    public ImageV4Response convert(Image source) {
```
pkedvessy, line 14: At least stackDetails conversion is missing. It was not needed for freeipa, it is the reason it was not implemented by me. Please make sure all ImageV4Response paremeters are set up.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
```diff
@@ -675,4 +686,103 @@ public String getResourceCrnByResourceName(String resourceName) {
     public EnumSet<Crn.ResourceType> getSupportedCrnResourceTypes() {
         return EnumSet.of(Crn.ResourceType.IMAGE_CATALOG);
     }
+
+    private StatedImage getStatedImage(long workspaceId, String catalogName, String imageId)
```
pkedvessy, line 690: This class is too complex as it is without our changes as well. I'd move out our custom image implementation into a separated component (as it was discussed during the handover)

jszeles: The possible parts were moved to the CustomImageProvider class.

core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
jszeles: New freeipa catalog is provided for the unit tests.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageType.java
jszeles: ImageTypeConverter is used for the conversation.

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/imagecatalog/ImageToImageV4ResponseConverter.java
jszeles: The conversation is extended with the stackDetails param.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/catalog/VersionBasedImageProvider.java
```diff
@@ -73,22 +77,25 @@ public StatedImages getImages(CloudbreakImageCatalogV3 imageCatalogV3, ImageFilt
...
 
         List<Image> defaultImages = defaultVMImageUUIDs.stream()
                 .map(imageId -> getImage(imageId, imageCatalogV3.getImages()))
                 .flatMap(Optional::stream)
                 .collect(Collectors.toList());
 
         Collection<String> latestDefaultImageUuids = latestDefaultImageUuidProvider.getLatestDefaultImageUuids(imageFilter.getPlatforms(), defaultImages);
-        Stream.of(baseImages.stream(), cdhImages.stream())
+        (!freeipaImages.isEmpty() ?
+                Stream.of(freeipaImages.stream()) :
```
pkedvessy, line 89: Minor: you don't need to wrap` freeipaImage.stream()` with `Stream.of()`

jszeles: Stream.of wrapping is removed.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageProvider.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.toMap;
+
+@Component
+public class CustomImageProvider {
+
+    private static final String INTERNAL_BASE_URL = "https://archive.cloudera.com/p/";
```
Bajzathd, line 25: I think this should be an application property, not a constant.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageProvider.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public StatedImage mergeSourceImageAndCustomImageProperties(
```
Bajzathd, line 27: Is it a source or a stated image? Not really clear from the method signature

core/src/main/java/com/sequenceiq/cloudbreak/service/image/catalog/VersionBasedImageCatalogService.java
```diff
@@ -57,39 +60,45 @@ public void validate(CloudbreakImageCatalogV3 imageCatalogV3) throws CloudbreakI
...
     }
 
     private void validateUuids(CloudbreakImageCatalogV3 imageCatalogV3) throws CloudbreakImageCatalogException {
         Stream<String> baseUuids = imageCatalogV3.getImages().getBaseImages().stream().map(Image::getUuid);
         Stream<String> cdhUuids = imageCatalogV3.getImages().getCdhImages().stream().map(Image::getUuid);
-        Stream<String> uuidStream = Stream.of(baseUuids, cdhUuids).
-                reduce(Stream::concat).
-                orElseGet(Stream::empty);
+        Stream<String> freeipaUuids = imageCatalogV3.getImages().getFreeIpaImages().stream().map(Image::getUuid);
+        Stream<String> uuidStream = freeipaUuids.findAny().isPresent() ?
```
pkedvessy, line 87: This fails in case of freeipa as freeipaUuids stream get processed double times. Please fix it and cover the case with unit test to make sure it works for freeipa as well.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/catalog/AdvertisedImageCatalogService.java
```diff
@@ -50,8 +53,11 @@ private void validateVersion(CloudbreakImageCatalogV3 imageCatalogV3) throws Clo
...
     }
 
-    private void validateAdvertisedCdhImageExistance(CloudbreakImageCatalogV3 imageCatalogV3) throws CloudbreakImageCatalogException {
-        if (imageCatalogV3.getImages().getCdhImages().stream().noneMatch(i -> i.isAdvertised())) {
+    private void validateAdvertisedImageExistance(CloudbreakImageCatalogV3 imageCatalogV3) throws CloudbreakImageCatalogException {
+        List<Image> cdhImages = imageCatalogV3.getImages().getCdhImages();
+        List<Image> freeipaImages = imageCatalogV3.getImages().getFreeIpaImages();
+        List<Image> images = !freeipaImages.isEmpty() ? freeipaImages : cdhImages;
+        if (images.stream().noneMatch(i -> i.isAdvertised())) {
             throw new CloudbreakImageCatalogException("There should be at least one advertised cdh image.");
```
Bajzathd, line 61: The message is no longer correct, it can be thrown for freeipa images too

core/src/main/resources/application.yml
```diff
@@ -121,6 +121,7 @@ cb:
     url: http://localhost:8088
     contextPath: /environmentservice
   freeipa:
+    image.catalog.url: https://cloudbreak-imagecatalog.s3.amazonaws.com/v3-prod-freeipa-image-catalog.json
```
Bajzathd, line 124: Image catalogs are defined in profile specific application properties, like application-dev.yml, please remove this

pkedvessy: No, we should not move this to anywhere as overrides are going to be applied by environment specific helm configuration - as they are already in place for other image catalog urls. Tbh we should do the opposite of what you are suggesting with other image urls :-)

Bajzathd: That is also fine with me, just do it consistently

core/src/main/java/com/sequenceiq/cloudbreak/service/image/catalog/VersionBasedImageCatalogService.java
jszeles: This issue is solved in such a way that the three image lists were given together, because or only the cdh or the freeipa part is filled. Furthermore more unit test are implemented for this validation part which is affected.

jszeles (approved):

PR #10496, opened 2021-04-19
CB-12053 Improve Periscope Logging

smaniraju:
1. Periscope logback configuration updated to log resourceName, resourceCrn, tenant and relevant MDC Context.
2. Initialize Periscope Logger MDC Context correctly.
3. Propogate CB API Client Error Messages during Autoscaling trigger.

### DIFF
--- a/autoscale-api/src/main/resources/messages/messages.properties
+++ b/autoscale-api/src/main/resources/messages/messages.properties
@@ -10,7 +10,7 @@ autoscale.enabled=Autoscaling has been enabled.
 autoscale.schedule.config.overlap=''SCHEDULE-BASED'' Autoscaling Configuration ''{0}'' not triggered, overlaps with ''{1}''.
 
 autoscale.cluster.exists.for.crn=Cluster exists for the same CRN ''{0}'' and CM Variant ''{1}''.
-autoscale.cluster.not.available=''{0}'' Autoscaling could not be triggered; cluster was not in 'AVAILABLE' status.
+autoscale.cluster.scaling.failed=''{0}'' Autoscaling could not be triggered. {1}
 autoscale.cluster.update.in.progress=''{0}'' Autoscaling could not be triggered; cluster update was in progress.
 
 autoscale.activity.success=''{0}'' Autoscaling triggered for Host Group ''{1}'' from ''{2}'' to ''{3}''.
@@ -25,4 +25,4 @@ autoscale.validation.time.negative.adjustment=Schedule-Based Autoscaling Node Co
 autoscale.notification.history.update=Autoscaling History Updated
 autoscale.notification.config.update=Autoscaling Config Updated
 
-autoscale.trigger.failure=Autoscaling Trigger Failed. Please verify Cluster Connectivity and Environment User(s) Synchronization.
+autoscale.trigger.failure=Autoscaling Trigger Failed. Autoscaling metrics collection from cluster failed.
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
@@ -3,8 +3,11 @@ package com.sequenceiq.periscope.monitor.handler;
 import static com.sequenceiq.periscope.common.MessageCode.AUTOSCALING_ACTIVITY_SUCCESS;
 
 import java.util.List;
+import java.util.Optional;
 
 import javax.inject.Inject;
+import javax.ws.rs.ClientErrorException;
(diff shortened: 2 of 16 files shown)


autoscale-api/src/main/resources/messages/messages.properties
```diff
@@ -10,7 +10,7 @@ autoscale.enabled=Autoscaling has been enabled.
 autoscale.schedule.config.overlap=''SCHEDULE-BASED'' Autoscaling Configuration ''{0}'' not triggered, overlaps with ''{1}''.
 
 autoscale.cluster.exists.for.crn=Cluster exists for the same CRN ''{0}'' and CM Variant ''{1}''.
-autoscale.cluster.not.available=''{0}'' Autoscaling could not be triggered; cluster was not in 'AVAILABLE' status.
+autoscale.cluster.scaling.failed=''{0}'' Autoscaling could not be triggered. {1}
 autoscale.cluster.update.in.progress=''{0}'' Autoscaling could not be triggered; cluster update was in progress.
```
smaniraju, line 14: Propogate CB API Message which has more context to user for reason of autoscale trigger failure like.
_SCHEDULE-BASED' Autoscaling could not be triggered. Stack smani-as has flows under operation, request not allowed._
_'SCHEDULE-BASED' Autoscaling could not be triggered. Upscale is not allowed because the following instances are not in running state: i-095969c15d35ee472. Please remove them first!._

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -156,14 +160,14 @@ private void scaleDown(int scalingAdjustment, int totalNodes) {
...
+            statusReason = getMessageForCBSuccess();
             metricService.incrementMetricCounter(MetricType.CLUSTER_DOWNSCALE_SUCCESSFUL);
         } catch (Exception e) {
             scalingStatus = ScalingStatus.FAILED;
             metricService.incrementMetricCounter(MetricType.CLUSTER_DOWNSCALE_FAILED);
-            statusReason = getMessageForCBStatus(e.getMessage());
+            statusReason = getMessageForCBException(e);
             LOGGER.error("Couldn't trigger downscaling for host group '{}', cluster '{}', desiredNodeCount '{}', error '{}' ",
-                    hostGroup, cluster.getStackCrn(), desiredNodeCount, e);
+                    hostGroup, cluster.getStackCrn(), desiredNodeCount, statusReason, e);
```
sidseth, line 170: slf4j handles Exceptions without requiring a separate '{}' in the log message, correct?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -201,19 +206,21 @@ private void createHistoryAndNotify(int adjustmentCount, int totalNodes, String
...
+    private String getMessageForCBException(Exception cbApiException) {
+        String cbApiError = cbApiException.getMessage();
+        if (cbApiException instanceof ClientErrorException) {
+            try (Response exceptionResponse = ((ClientErrorException) cbApiException).getResponse()) {
+                cbApiError = exceptionResponse.readEntity(ExceptionResponse.class).getMessage();
+            } catch (Exception ex) {
+                LOGGER.error("Error processing CB API Exception '{}'", cbApiException, ex);
+            }
         }
+        return messagesService.getMessage(MessageCode.CLUSTER_SCALING_FAILED,
```
sidseth, line 218: Should 404 be handled explicitly? (Extremely unlikely though - since that would mean a trigger before Periscope syncs a deleted cluster?)
What happens when a cluster is in STOPPED state?

sidseth (approved):
Looks good. Couple of questions.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: slf4j logs exception stack trace without '{}' , but doesn't extract server error text from BadRequest and similar exceptions.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: 404 need not handled explicity since it just needs to be logged similar to other error codes.
When cluster is stopped, LOAD-Based autoscaling also stops. For TIME-Based autoscaling a message is added in event history saying "Autoscaling could not be triggered because cluster was not available"

PR #10499, opened 2021-04-19
CB-11963 Handle recipe execution failure on hosts without metadata or…

Bajzathd:
… fqdn

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
@@ -24,7 +24,9 @@ import org.springframework.core.convert.ConversionService;
 import org.springframework.stereotype.Component;
 
 import com.google.api.client.util.Joiner;
+import com.google.common.annotations.VisibleForTesting;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.Status;
+import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.ClusterDeletionBasedExitCriteriaModel;
 import com.sequenceiq.cloudbreak.domain.Recipe;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
```diff
@@ -176,30 +177,42 @@ private String getRecipeTimeoutErrorMessage(CloudbreakOrchestratorTimeoutExcepti
...
-        }).collect(Collectors.joining("\n ---------------------------------------------- \n")
-        );
+        String message = failures.stream()
+                .map(failure -> getSingleRecipeExecutionFailureMessage(instanceMetaData, failure))
+                .collect(Collectors.joining("\n ---------------------------------------------- \n"));
         return new StringBuilder("Failed to execute recipe(s): \n").append(message).toString();
     }
 
+    @VisibleForTesting
+    String getSingleRecipeExecutionFailureMessage(Set<InstanceMetaData> instanceMetaData, RecipeFailure failure) {
```
horadla23, line 195: I would prefer not using this `@VisibleForTesting` but unit test it through public method and make this private

topolyai5 (commented):
What will the result if the host and fqdn are null as well?

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
Bajzathd: This is called from another private method, that handles error from 4 different public methods, this is why I have done it this way. To eliminate this annotation a better solution would be to extract this logic to its own class (or into RecipeExecutionFailureCollector) and test it instead, I can do it if you think that would be better.

Bajzathd:
> What will the result if the host and fqdn are null as well?

If both values are null than they would equal, but the host can not be null, it is what we get back from the recipe execution failure.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/recipe/OrchestratorRecipeExecutor.java
pkedvessy: +1 on reducing the complexity of this class by moving out the failure message generation logic into a separated component

horadla23: i think testing only one public method's error handling would be enough to test this method

horadla23 (approved):

PR #10503, opened 2021-04-19
CB-12065: Fix FreeIPA HA repair to update cached kerberos config

jamisonbennett:
Fix FreeIPA HA repair so that the cached kerberos configuration is
updated. Specifically the DNS nameservers were stale which caused
DataLakes and DataHubs to have DNS break after all of the original
FreeIPA servers have been repaired.

This was tested by validating the cached kerberos configuration is
updated during downscale and upscale operations.


christmasferret (approved):
LGTM

freeipa/src/main/java/com/sequenceiq/freeipa/service/config/KerberosConfigUpdateService.java
```diff
@@ -0,0 +1,43 @@
...
+    @Inject
+    private KerberosConfigService kerberosConfigService;
+
+    @Inject
+    private StackService stackService;
+
+    public KerberosConfig updateNameservers(Long stackId) {
+        Stack stack = getStackWithInstanceMetadata(stackId);
+        String environmentCrn = stack.getEnvironmentCrn();
+        KerberosConfig kerberosConfig = kerberosConfigService.get(environmentCrn);
```
lacikaaa, line 32: this won't return all. you have to use `kerberosConfigRepository.findByAccountIdAndEnvironmentCrnAndArchivedIsFalse` on repository level, and this is not exposed on service level right now

lacikaaa (changes requested):

lacikaaa (approved):

lacikaaa:
this LGTM. but there are old LDAP configs, for old FreeIPA installation where we don't have the CNAME. see: `LdapConfigRegisterService#addServerHost`
we might need to handle those if there is any. we should check that before making any changes

jamisonbennett:
@lacikaaa Thanks for mentioning this. `LdapConfigRegisterService#addServerHost` was supported from 2.20 and later. The repair flow only is possible for HA configurations which are much later (2.31?). So this shouldn't be a problem because we can't repair something that old anyway.

lacikaaa:
@jamisonbennett thanks for the clarification
what about upscale, when we are migrating from single to HA in the foreseeable future?

PR #10507, opened 2021-04-19
CB-12174 Logic to create in-place public load balancer is using private subnet

hreeve-cloudera:
When the stack network is updated during the in-place load balancer creation, persist those changes to
the database, and to the new stack object passed along during creation. Tested by calling the API on an
existing environment and verifying the database network table was updated, the new stack object had the
updated network information in subsequent steps, and the created load balancer used a public subnet
instead of a private one.

### DIFF
--- a/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
+++ b/cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
@@ -19,7 +19,6 @@ import org.springframework.stereotype.Service;
 
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.cloudformation.model.DescribeStacksRequest;
-import com.amazonaws.services.cloudformation.model.ListStackResourcesResult;
 import com.amazonaws.services.cloudformation.model.ResourceStatus;
 import com.amazonaws.services.ec2.model.DescribeKeyPairsRequest;
 import com.amazonaws.services.ec2.model.ImportKeyPairRequest;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/CreateLoadBalancerEntityHandler.java
```diff
@@ -114,10 +118,13 @@ protected Selectable doAccept(HandlerEvent<CreateLoadBalancerEntityRequest> even
                 measure(() -> instanceGroupService.saveAll(newLoadBalancers.stream()
                         .flatMap(lb -> lb.getAllInstanceGroups().stream()).collect(Collectors.toSet())),
                     LOGGER, "Instance groups saved in {} ms for stack {}", stackName);
+                measure(() -> networkService.pureSave(stack.getNetwork()), LOGGER,
```
sxxgrc, line 121: This won't be affected by the network being null right? I assume not but just making sure since Yarn has a null network field for its stack objects.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsLaunchService.java
```diff
@@ -164,26 +163,6 @@
         return awsResourceConnector.check(ac, instances);
     }
 
-    private ListStackResourcesResult updateCloudFormationStack(AuthenticatedContext ac, CloudStack stack, ModelContext modelContext) {
```
sxxgrc, line 167: So removing this was necessary to persist the updated stack during load balancer creation? I'm sure it was but I can't tell why you removed it nor if it was doing anything important initially.

hreeve-cloudera: Ah, no, this is unrelated. It just so happened when I was testing this code I realized this method wasn't being used anymore, and decided to clean it up.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/CreateLoadBalancerEntityHandler.java
hreeve-cloudera: Good call. I'll put a null check around it.

orngejaket (approved):

PR #10508, opened 2021-04-19
CB-11944 AWS Datalake provisioning on mow-priv fails when creation flow restarts

hreeve-cloudera:
Moves the load balancer creation to a separate flow step from the rest of the resource creation.
This is because if all the resources are created together, CB cannot successfully restart the flow
step if it fails after load balancer creation has been initialized, but before it's completed.

Tested by running stack creation in local CB, verifying load balancers weren't created in initial
provisioning step, and were successfully provisioned in new step.


yg-gd:
Test

sxxgrc (approved):
Looks good to me. I like seeing everything isolated in that way too since it's probably better in the long run. Thank you for making this change.

cloud-reactor/src/main/java/com/sequenceiq/cloudbreak/cloud/handler/LaunchLoadBalancerHandler.java
```diff
@@ -0,0 +1,116 @@
...
+                result = ResourcesStatePollerResults.transformToLaunchLoadBalancerResult(request, statePollerResult);
+                request.getResult().onNext(result);
+                LOGGER.debug("Launching the load balancers successfully finished for {}", cloudContext);
+            } else {
+                LOGGER.debug("No load balancers configured for stack.");
+                result = new LaunchLoadBalancerResult(request.getResourceId(), List.of());
+            }
+            eventBus.notify(result.selector(), new Event<>(launchLoadBalancerRequestEvent.getHeaders(), result));
+        } catch (Exception e) {
+            if (ExceptionUtils.getRootCause(e) instanceof InterruptedException) {
```
lacikaaa, line 79: Have you seen such issue? I think we could add these to everywhere, so either we should not add it here or handle it generally.
@sodre90 what do you think?

sodre90: We did some improvements, so now it should not cause any problem. Eventbus is stopped as the first step in graceful shutdown so this message will not be consumed.
Anyway as I see it does not do anything, just logging, but it will send the message to the event bus :) So we can leave it as it is.

hreeve-cloudera: I haven't personally seen this myself. It was part of the `LaunchStackHandler`, and since the load balancer launch logic go through a very similar path through the cloud resource connectors as the stack launch logic does, I thought it was possible to run into this exception and that I should handle it.

hreeve-cloudera: @sodre90 It sounds like this might not be necessary in either case anymore? If so I'll clean it up next time I'm in that part of the code.

lacikaaa: no, you don't need this anywhere.
also there is no `else` branch here, so this code has no effect other than logging. I would say it should be deleted just to clean it up

PR #10509, opened 2021-04-19
CB-10579 Clean up unused targetgroup and loadbalancer columns

hreeve-cloudera:
Removes a column and index that were replaced in CB-10578. Tested by running database migrate up
and down on local database.

### DIFF
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/LoadBalancer.java
@@ -5,7 +5,6 @@ import java.util.HashSet;
 import java.util.Set;
 
 import java.util.stream.Collectors;
-import javax.persistence.CascadeType;
 import javax.persistence.Convert;
 import javax.persistence.Entity;
 import javax.persistence.FetchType;
@@ -14,7 +13,6 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.ManyToMany;
 import javax.persistence.ManyToOne;
-import javax.persistence.OneToMany;
 import javax.persistence.SequenceGenerator;
 
 import com.sequenceiq.cloudbreak.domain.ProvisionEntity;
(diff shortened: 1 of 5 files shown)


brycederriso (approved):
This looks right to me. The hard stuff is in the `UNDO` steps:
1. Add the `loadbalancer_id` column back
2. Add the foreign key constraint back to the `loadbalancer_id` column
3. Update all of the `loadbalancer_id` columns with the value that's in `targetgroup_loadbalancer.loadbalancerset_id`.

The changes for [CB-10578](https://github.com/hortonworks/cloudbreak/pull/9961/files) copied the values from `targetgroup.loadbalancer_id` into `targetgroup_loadbalancer.loadbalancer_id`.

I do wonder if we should include dropping the column in the same change as the copying. It makes sense to me conceptually, since the `UNDO` in _this_ changeset is sort of detached from the previous change -- except that you called out it was related to another PR!

Did you do it this way for particular reasons/convenience?

hreeve-cloudera:
> This looks right to me. The hard stuff is in the `UNDO` steps:
> 
>     1. Add the `loadbalancer_id` column back
> 
>     2. Add the foreign key constraint back to the `loadbalancer_id` column
> 
>     3. Update all of the `loadbalancer_id` columns with the value that's in `targetgroup_loadbalancer.loadbalancerset_id`.
> 
> 
> The changes for [CB-10578](https://github.com/hortonworks/cloudbreak/pull/9961/files) copied the values from `targetgroup.loadbalancer_id` into `targetgroup_loadbalancer.loadbalancer_id`.
> 
> I do wonder if we should include dropping the column in the same change as the copying. It makes sense to me conceptually, since the `UNDO` in _this_ changeset is sort of detached from the previous change -- except that you called out it was related to another PR!
> 
> Did you do it this way for particular reasons/convenience?

Yes. It's following the guidelines for [Database schema changes with zero downtime](https://cloudera.atlassian.net/wiki/spaces/ENG/pages/182452362/Database+schema+changes+with+zero+downtime). Anything that replaces existing objects in the database happens in three steps:

1. Create the new columns/tables.
2. Copy data from the old columns/tables to the new ones.
3. Delete the now unused columns/tables.

This process is to prevent the database from getting out of sync with the code. If it's done all together then, for example, the database may be updated before the services that uses the database, and so when the service tries to access the columns that were deleted as part of the database update the service fails. So this is just the third step in that process.

orngejaket (approved):

PR #10513, opened 2021-04-20
CB-6227 Optimize CB queries for stack

bbihari:
Replace the expensive stack queries to more efficient ones.


core/src/main/java/com/sequenceiq/cloudbreak/repository/InstanceMetaDataRepository.java
```diff
@@ -46,6 +46,9 @@
     @Query("SELECT i FROM InstanceMetaData i WHERE i.instanceGroup.stack.id= :stackId")
     Set<InstanceMetaData> findAllInStack(@Param("stackId") Long stackId);
 
+    @Query("SELECT i FROM InstanceMetaData i WHERE i.instanceGroup.stack.id= :stackId")
```
foldik, line 49: This query looks the same as the above. What is the difference other than naming?

foldik: Ahh I see, different entity graph. Sorry.

foldik (approved):
lgtm

PR #10514, opened 2021-04-20
CB-11754 change compute.storageAdmin role to storage.admin in the GCP credential create command

biharitomi:
The compute.storageAdmin role is completely superseded by compute.instanceAdmin.v1 so it could be removed, but storage.admin role is neccesary for CDP related storage operations.

I think it should go into 2.42 line of CB, so master is fine for now...


topolyai5 (approved):

doktoric (approved):

bergerdenes (approved):
unittest failed

PR #10520, opened 2021-04-21
CB-12078 autoscale and clusterdnsclient should use internal actor

horadla23:


horadla23:
contains: autoscale uses internal actor again, also fixed issues in DISTX-611

foldik (approved):
LGTM

smaniraju (approved):
Autoscaler changes LGTM

PR #10526, opened 2021-04-21
CB-9881 Introduce imagevalidation Spring profile

Bajzathd:


doktoric:
do we have enough quota there ?

Bajzathd:
> do we have enough quota there ?

I asked around in e2e channel if there can be any problems with it, and this concern was not raised. How can I verify if we have enough quota?

Bajzathd:
Application properties moved to https://github.com/hortonworks/cloudbreak-ansible-playbooks/pull/151/files

PR #10528, opened 2021-04-21
CB-12111: Update FreeIPA instance to "stopped" immediately after stac…

christmasferret:
…k was stopped

This is a bug that FreeIPA instance is "running" after stack was stopped. 
There is a 3 minutes delay in UI until next health check. 
We fixed by updating FreeIPA instance to "stopped" immediately after stack was stopped.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
@@ -8,9 +8,11 @@ import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.cloud.event.instance.StopInstancesResult;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.DetailedStackStatus;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
 import com.sequenceiq.freeipa.entity.Stack;
 import com.sequenceiq.freeipa.flow.stack.StackFailureEvent;
 import com.sequenceiq.freeipa.flow.stack.StackStartStopService;
+import com.sequenceiq.freeipa.service.stack.instance.InstanceUpdater;
 import com.sequenceiq.freeipa.service.stack.StackUpdater;
 
 @Component
@@ -23,6 +25,9 @@ public class StackStopService {
 
     @Inject
     private StackStartStopService stackStartStopService;
+
+    @Inject
+    private InstanceUpdater instanceUpdater;
 
     public void startStackStop(StackStopContext context) {
         stackUpdater.updateStackStatus(context.getStack(), DetailedStackStatus.STOP_IN_PROGRESS, "Stack infrastructure is now stopping.");
@@ -39,6 +44,7 @@ public class StackStopService {
         stackStartStopService.validateResourceResults(context.getCloudContext(),
                 stopInstancesResult.getErrorDetails(), stopInstancesResult.getResults(), false);
         stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.STOPPED, "Stack infrastructure stopped successfully.");
+        instanceUpdater.updateStatuses(stack, InstanceStatus.STOPPED);
     }
 
     public boolean isStopPossible(Stack stack) {
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/InstanceUpdater.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/InstanceUpdater.java
@@ -1,0 +1,39 @@
+package com.sequenceiq.freeipa.service.stack.instance;
+
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.InstanceStatus;
+import com.sequenceiq.freeipa.entity.InstanceMetaData;
+import com.sequenceiq.freeipa.entity.Stack;
+
+import javax.inject.Inject;
+
+@Component
+public class InstanceUpdater {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(InstanceUpdater.class);
+
+    @Inject
+    private InstanceMetaDataService instanceMetaDataService;
+
+    public Set<InstanceMetaData> updateStatuses(Stack stack, InstanceStatus instanceStatus) {
+        Set<InstanceMetaData> checkableInstances = stack.getNotDeletedInstanceMetaDataSet();
+        checkableInstances.forEach(instanceMetaData -> {
+            setStatusIfNotTheSame(instanceMetaData, instanceStatus);
+            instanceMetaDataService.save(instanceMetaData);
+        });
+        return checkableInstances;
+    }
+
+    private void setStatusIfNotTheSame(InstanceMetaData instanceMetaData, InstanceStatus newStatus) {
+        InstanceStatus oldStatus = instanceMetaData.getInstanceStatus();
+        if (oldStatus != newStatus) {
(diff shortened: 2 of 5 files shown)


christmasferret:
Currently, the freeipa instance status has a three minutes out of sync at the end of a stop flow. 
image-2021-04-13-16-07-10-158

It is fixed by immediately pulling and updating the status of the instances.
Screenshot from 2021-04-21 11-02-48

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
```diff
@@ -12,6 +12,7 @@
 import com.sequenceiq.freeipa.flow.stack.StackFailureEvent;
```
jamisonbennett, line 12: nit: the commit message line length is pretty long, can you add some new lines?

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -40,12 +43,22 @@
     @Value("${freeipa.autosync.update.status:true}")
     private boolean updateStatus;
 
+    public List<ProviderSyncResult> updateAndGetStatuses(Stack stack, boolean updateStatusFromFlow) {
+        Set<InstanceMetaData> notTerminatedForStack = stack.getAllInstanceMetaDataList().stream()
+                .filter(Predicate.not(InstanceMetaData::isTerminated))
+                .collect(Collectors.toSet());
+        Set<InstanceMetaData> checkableInstances = notTerminatedForStack.stream()
```
jamisonbennett, line 50: It seems that these two collections could be combined into a single collection that is filtered twice.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
christmasferret: Thx! Updated.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: Thx! Updated it accordingly.

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
```diff
@@ -39,6 +43,7 @@ public void finishStackStop(StackStopContext context, StopInstancesResult stopIn
         stackStartStopService.validateResourceResults(context.getCloudContext(),
                 stopInstancesResult.getErrorDetails(), stopInstancesResult.getResults(), false);
         stackUpdater.updateStackStatus(stack.getId(), DetailedStackStatus.STOPPED, "Stack infrastructure stopped successfully.");
+        providerChecker.updateAndGetStatuses(stack, true);
```
lacikaaa, line 46: the `finishStackStop` method is called from an action and not a handler
`ProviderChecker#updateAndGetStatuses` is calling the cloud provider which means it can take significant time (more likely if any retry can happen)
Actions should have only short running logic, like updating the DB.
I would prefer having this called from a handler. (if you think we need this)

Is there any reason why we don't just rewrite the instance status instead of going to the provider? The previous step (I haven't checked the actual code) should wait for the instances to be stopped on the provider, I don't think checking it again has any additional benefit.

lacikaaa (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/flow/stack/stop/StackStopService.java
christmasferret: @lacikaaa   I made the change according to your suggestion. Now instead of polling from the provider, we update the instance status directly. Internally, it waits till the instances stopped and stack status stopped, and then it updates the instance status to stopped at last. Tested again and everything works. Thx!

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -77,6 +79,28 @@
         }, LOGGER, ":::Auto sync::: provider is checked in {}ms");
     }
 
+    public Set<InstanceMetaData> updateStatuses(Stack stack, InstanceStatus instanceStatus, boolean updateStatusFromFlow) {
```
lacikaaa, line 82: - `updateStatusFromFlow` doesn't make sense here as it is called only from one place
- I don't think this is the right class for this method as this method doesn't check anything on the provider side
- please cover it with unit tests

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -77,6 +79,28 @@
...
+        Set<InstanceMetaData> checkableInstances = stack.getAllInstanceMetaDataList().stream()
+                .filter(Predicate.not(InstanceMetaData::isTerminated))
+                .filter(Predicate.not(InstanceMetaData::isDeletedOnProvider))
+                .collect(Collectors.toSet());
+
+        if (!updateStatusFromFlow && flowLogService.isOtherFlowRunning(stack.getId())) {
```
lacikaaa, line 88: I don't think this is necessary as it is only called from one flow step

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
```diff
@@ -77,6 +79,28 @@
...
+            throw new InterruptSyncingException("interrupt updating instance status in updateStatuses, flow is running on freeipa stack " +
+                    stack.getName());
+        } else {
+            checkableInstances.forEach(instanceMetaData -> {
+                if (updateStatus) {
```
lacikaaa, line 93: as this is not part of the sync, it should be always updated when the instances are stopped. So this condition is not necessary

lacikaaa (changes requested):

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: Thx! I moved the method to a separate InstanceUpdater.java class which is similar to the StackUpdater.java class. Now it is more clear.

Also, I created unit tests to cover it.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: Yes. You are right. I removed it.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/ProviderChecker.java
christmasferret: Yes. You are right. I removed it.

freeipa/src/main/java/com/sequenceiq/freeipa/service/stack/instance/InstanceUpdater.java
```diff
@@ -0,0 +1,45 @@
...
+@Component
+public class InstanceUpdater {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(InstanceUpdater.class);
+
+    @Inject
+    private InstanceMetaDataService instanceMetaDataService;
+
+    public Set<InstanceMetaData> updateStatuses(Stack stack, InstanceStatus instanceStatus) {
+        Set<InstanceMetaData> checkableInstances = stack.getAllInstanceMetaDataList().stream()
```
lacikaaa, line 26: `Stack` already has a `getNotDeletedInstanceMetaDataSet` method, you could use that one

christmasferret: Thanks! I wasn't aware of this existing function. Now I reuse it for this purpose.

lacikaaa (approved):

PR #10529, opened 2021-04-21
CB-12224: Add FreeIPA availability status

jamisonbennett:
Provide workloads a summary of whether FreeIPA is available and able
to respond to API requests. Several checks were modified to use this
new status, since it is more accurate than the old way of checking if
FreeIPA is available.

Unit tests were updated.
The FreeIPA management service API was manually checked with a local
deployment of cloudbreak.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/start/SdxStartService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/start/SdxStartService.java
@@ -167,7 +167,7 @@ public class SdxStartService {
 
     private void checkFreeipaRunning(String envCrn) {
         DescribeFreeIpaResponse freeipa = freeipaService.describe(envCrn);
-        if (freeipa != null && !freeipa.getStatus().isAvailable()) {
+        if (freeipa != null && freeipa.getAvailabilityStatus() != null && !freeipa.getAvailabilityStatus().isAvailable()) {
             throw new BadRequestException("Freeipa should be in Available state but currently is " + freeipa.getStatus().name());
         }
 
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/stop/SdxStopService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/stop/SdxStopService.java
@@ -179,7 +179,7 @@ public class SdxStopService {
 
     private void checkFreeipaRunning(String envCrn) {
         DescribeFreeIpaResponse freeipa = freeipaService.describe(envCrn);
-        if (freeipa != null && !freeipa.getStatus().isAvailable()) {
+        if (freeipa != null && freeipa.getAvailabilityStatus() != null && !freeipa.getAvailabilityStatus().isAvailable()) {
             throw new BadRequestException("Freeipa should be in Available state but currently is " + freeipa.getStatus().name());
         }
     }
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/SynchronizeUsersHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/SynchronizeUsersHandler.java
@@ -50,7 +50,7 @@ public class SynchronizeUsersHandler extends EventSenderAwareHandler<Environment
         try {
             if (synchronizeOnStartEnabled) {
                 freeIpaService.describe(environmentDto.getResourceCrn()).ifPresent(freeIpa -> {
-                    if (freeIpa.getStatus() != null && !freeIpa.getStatus().isAvailable()) {
+                    if (freeIpa.getStatus() != null && freeIpa.getAvailabilityStatus() != null && !freeIpa.getAvailabilityStatus().isAvailable()) {
                         throw new FreeIpaOperationFailedException("FreeIPA is not in AVAILABLE state to synchronize users! Current state is: " +
                                 freeIpa.getStatus().name());
                     }
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
@@ -1,80 +1,99 @@
 package com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common;
 
+import java.util.Collection;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 public enum DetailedStackStatus {
-    UNKNOWN(Status.UNKNOWN),
+    UNKNOWN(Status.UNKNOWN, AvailabilityStatus.UNKNOWN),
     // Provision statuses
-    PROVISION_REQUESTED(Status.REQUESTED),
-    PROVISION_SETUP(Status.CREATE_IN_PROGRESS),
-    IMAGE_SETUP(Status.CREATE_IN_PROGRESS),
-    CREATING_INFRASTRUCTURE(Status.CREATE_IN_PROGRESS),
-    METADATA_COLLECTION(Status.CREATE_IN_PROGRESS),
-    TLS_SETUP(Status.CREATE_IN_PROGRESS),
-    REGISTERING_WITH_CLUSTER_PROXY(Status.CREATE_IN_PROGRESS),
-    STACK_PROVISIONED(Status.STACK_AVAILABLE),
-    PROVISIONED(Status.AVAILABLE),
-    PROVISION_FAILED(Status.CREATE_FAILED),
+    PROVISION_REQUESTED(Status.REQUESTED, AvailabilityStatus.UNAVAILABLE),
+    PROVISION_SETUP(Status.CREATE_IN_PROGRESS, AvailabilityStatus.UNAVAILABLE),
+    IMAGE_SETUP(Status.CREATE_IN_PROGRESS, AvailabilityStatus.UNAVAILABLE),
+    CREATING_INFRASTRUCTURE(Status.CREATE_IN_PROGRESS, AvailabilityStatus.UNAVAILABLE),
+    METADATA_COLLECTION(Status.CREATE_IN_PROGRESS, AvailabilityStatus.UNAVAILABLE),
+    TLS_SETUP(Status.CREATE_IN_PROGRESS, AvailabilityStatus.UNAVAILABLE),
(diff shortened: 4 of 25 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/start/SdxStartService.java
```diff
@@ -167,7 +167,7 @@ public void waitCloudbreakCluster(Long sdxId, PollingConfig pollingConfig) {
 
     private void checkFreeipaRunning(String envCrn) {
         DescribeFreeIpaResponse freeipa = freeipaService.describe(envCrn);
-        if (freeipa != null && !freeipa.getStatus().isAvailable()) {
+        if (freeipa != null && !freeipa.getAvailabilityStatus()) {
```
sxxgrc, line 170: Might be an obvious question but if FreeIPA is null does that not inherently mean it is not in an available state? Or is this verified someplace else.

jamisonbennett: Good point, there are probably several other places that should have the check too.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
```diff
@@ -1,80 +1,99 @@
...
+    DOWNSCALE_REQUESTED(Status.UPDATE_REQUESTED, true),
+    DOWNSCALE_IN_PROGRESS(Status.UPDATE_IN_PROGRESS, true),
+    DOWNSCALE_COMPLETED(Status.AVAILABLE, true),
+    DOWNSCALE_FAILED(Status.DOWNSCALE_FAILED, false),
     // Repair statuses
-    REPAIR_REQUESTED(Status.UPDATE_REQUESTED),
-    REPAIR_IN_PROGRESS(Status.UPDATE_IN_PROGRESS),
-    REPAIR_COMPLETED(Status.AVAILABLE),
-    REPAIR_FAILED(Status.REPAIR_FAILED),
+    REPAIR_REQUESTED(Status.UPDATE_REQUESTED, true),
```
sxxgrc, line 51: Would this not imply that when an instance is in a faulty state (and therefore is _not_ available) and a user then attempts to repair it, it would then become available? Is that okay?

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
```diff
@@ -1,80 +1,99 @@
...
+    CLUSTER_OPERATION(Status.UPDATE_IN_PROGRESS, true),
     // Wait for sync
-    WAIT_FOR_SYNC(Status.WAIT_FOR_SYNC),
-    UNREACHABLE(Status.UNREACHABLE),
-    DELETED_ON_PROVIDER_SIDE(Status.DELETED_ON_PROVIDER_SIDE),
-    UNHEALTHY(Status.UNHEALTHY);
+    WAIT_FOR_SYNC(Status.WAIT_FOR_SYNC, false),
+    UNREACHABLE(Status.UNREACHABLE, false),
+    DELETED_ON_PROVIDER_SIDE(Status.DELETED_ON_PROVIDER_SIDE, false),
+    UNHEALTHY(Status.UNHEALTHY, true);
```
sxxgrc, line 73: I am definitely missing something but figured I'd ask, why is it available when it is unhealthy?

environment/src/main/java/com/sequenceiq/environment/environment/flow/start/handler/SynchronizeUsersHandler.java
```diff
@@ -50,7 +50,7 @@ public void accept(Event<EnvironmentStartDto> environmentStartDtoEvent) {
         try {
             if (synchronizeOnStartEnabled) {
                 freeIpaService.describe(environmentDto.getResourceCrn()).ifPresent(freeIpa -> {
-                    if (freeIpa.getStatus() != null && !freeIpa.getStatus().isAvailable()) {
+                    if (freeIpa.getStatus() != null && !freeIpa.getAvailabilityStatus()) {
```
sxxgrc, line 53: Same null question here.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/stop/SdxStopService.java
```diff
@@ -179,7 +179,7 @@ public void stopAllDatahub(Long sdxId) {
 
     private void checkFreeipaRunning(String envCrn) {
         DescribeFreeIpaResponse freeipa = freeipaService.describe(envCrn);
-        if (freeipa != null && !freeipa.getStatus().isAvailable()) {
+        if (freeipa != null && !freeipa.getAvailabilityStatus()) {
```
sxxgrc, line 182: Same null question here.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/stack/StackToDescribeFreeIpaResponseConverter.java
```diff
@@ -123,4 +125,8 @@ private PlacementResponse convert(Stack source) {
         placementResponse.setRegion(source.getRegion());
         return placementResponse;
     }
+
+    private Boolean convertAvailabilityStatus(StackStatus status) {
+        return status == null ? Boolean.FALSE : status.getDetailedStackStatus().isAvailable();
+    }
```
christmasferret, line 131: Do you want to log the detailedStackStatus to know what status exactly was converted to available/unavailable? So if it is unavailable, we know what the real status is. Or if you can search easily now in log already, then ignore.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
jamisonbennett: During a repair operation, at least 1 freeipa must be available otherwise there isn't a node the replicate from.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
jamisonbennett: Let's say we are running FreeIPA HA and there are 3 instances. One is in a bad state and two are responding correctly. Then the cluster would be unhealthy but that means that someone needs to repair it but FreeIPA is able to respond to all requests in this case. It definitely isn't perfect. For example, if all nodes are unhealthy then it is possible FreeIPA is unable to respond. But usually, when all nodes are unhealthy we see that FreeIPA is unreachable, stopped, failed, deleted, etc.

jamisonbennett: Maybe we should introduce two unhealthy states. One that means FreeIPA is not able to respond and another indicating it is still available.

freeipa/src/main/java/com/sequenceiq/freeipa/converter/stack/StackToDescribeFreeIpaResponseConverter.java
jamisonbennett: No, because the detailed status is also in the converted response.

sxxgrc (approved):
Aside from my small comments and possible edge cases which can be fixed in the future, I think this is a good change that is a great move in the right direction.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
sxxgrc: Ah okay I did not know that, thanks for the info.

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
sxxgrc: Okay that makes sense, I understand why you did it this way. 
Adding another unhealthy state might be ideal, something like being "unresponsive" vs being "unavailable", but that can definitely come with a future improvement.

freeipa/src/main/java/com/sequenceiq/freeipa/repository/StackRepository.java
```diff
@@ -76,6 +77,9 @@
     @Query("SELECT s FROM Stack s WHERE s.stackStatus.status IN :stackStatuses AND s.terminated = -1 ")
     List<Stack> findAllWithStatuses(@Param("stackStatuses") Collection<Status> stackStatuses);
 
+    @Query("SELECT s FROM Stack s WHERE s.stackStatus.detailedStackStatus IN :detailedStackStatuses AND s.terminated = -1 ")
```
lacikaaa, line 80: while `StackStatus#status` has index `detailedStackStatus` doesn't. Could you add it if we are going to use it in this where clause?

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/describe/DescribeFreeIpaResponse.java
```diff
@@ -62,6 +62,8 @@
     @ApiModelProperty(value = FreeIpaModelDescriptions.FREEIPA_SERVER_SETTINGS, required = true)
     private FreeIpaServerResponse freeIpa;
 
+    private Boolean availabilityStatus;
```
lacikaaa, line 65: I know we agreed it should be a 1 bit thing, but for the API I still prefer enums. just in case. it's easier to extend if we must.
I'm fine handling this as a boolean in the background

lacikaaa (changes requested):

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/model/common/DetailedStackStatus.java
jamisonbennett: I filed CB-12245 to handle this case.

lacikaaa (approved):

PR #10534, opened 2021-04-22
CB-11556 - extract code duplication from delete flow exception handlings

gregito:


lajosrodek:
@gregito  Erm, there is one more handler to take care of. 😃 Please [ResourceEncryptionDeleteHandler.accept()](https://github.com/hortonworks/cloudbreak/blob/master/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java#L68-L78) that got merged in yesterday (#10407).

Speaking of which, `deleteEncryptionResources()` of said class seems to simply swallow the exception in the `catch` clause; this apparently slipped through our rushed reviews. 😞 

CC @doktoric @juhi-09

lajosrodek:
@gregito Thanks for the change. I believe `ResourceEncryptionDeleteHandlerTest` has a test case checking the error path as well, and probably fails with an NPE due to the missing `HandlerExceptionProcessor` dependency.

lajosrodek:
Looks like there is also one more class with duplication: [IdBrokerMappingsDeleteHandler](https://github.com/hortonworks/cloudbreak/blob/master/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/IdBrokerMappingsDeleteHandler.java).

bergerdenes (changes requested):
I had some private questions: `HandlerFailureConjoiner` could take `Event` only
and  when verifying `mockExceptionProcessor`, the instances of the `HandlerFailureConjoiner` could be captured with an `ArgumentCaptor` and check the values of it

afarsang (approved):

doktoric (approved):

PR #10538, opened 2021-04-22
CB-12191 Add userTags field to EnvironmentDetails and Clusterdetails …

nadamdb:
…fields

Additionally updating the usage.proto file with changes from the original file in thunderhead.


environment/src/main/java/com/sequenceiq/environment/environment/dto/EnvironmentDto.java
```diff
@@ -338,6 +338,10 @@ public void setTags(EnvironmentTags tags) {
         this.tags = tags;
     }
 
+    public Map<String, String> getUserDefinedTags() {
```
doktoric, line 341: I think this field is duplicated with EnvironmentTags

akanto (approved):

PR #10539, opened 2021-04-23
CB-12243: DES delete flow exceptions are swallowed and do not result in flow failure

juhi-09:
CB-12243: DES delete flow exceptions are swallowed and do not result in flow failure
Fix the issue of swallowing of exceptions - by re-throwing the exception after logging.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
@@ -11,6 +11,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 import com.sequenceiq.environment.environment.EnvironmentStatus;
 import com.sequenceiq.environment.environment.domain.Environment;
 import com.sequenceiq.environment.environment.dto.EnvironmentDeletionDto;
@@ -86,6 +87,7 @@ public class ResourceEncryptionDeleteHandler extends EventSenderAwareHandler<Env
             environmentService.save(environment);
         } catch (Exception e) {
             LOGGER.error("Failed to delete Encryption resources for environment {} with error {}", environment.getName(), e);
+            throw new CloudbreakServiceException("Error occured while deleting encryption resources: " + e.getMessage(), e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandler.java
```diff
@@ -86,6 +86,7 @@ private void deleteEncryptionResources(EnvironmentDto environmentDto, Environmen
             environmentService.save(environment);
         } catch (Exception e) {
             LOGGER.error("Failed to delete Encryption resources for environment {} with error {}", environment.getName(), e);
+            throw e;
```
doktoric, line 89: I guess we need to be more specific like new Cloudbreak...Exception(e, "message")

lajosrodek: Agreed, `CloudbreakServiceException` could be a good choice.

environment/src/test/java/com/sequenceiq/environment/environment/flow/deletion/handler/ResourceEncryptionDeleteHandlerTest.java
```diff
@@ -160,6 +160,17 @@ void testDeletionContinuesIfDiskEncryptionSetIdIsNotPresent() {
         verifyEnvDeleteEvent();
     }
 
+    @Test
+    void testErrorWhenDeleteEncryptionResourcesResultsInErrorWhileEnvironmentSave() {
+        IllegalArgumentException errorOnSave = new IllegalArgumentException("error");
+        Environment environment = new Environment();
+        when(environmentService.findEnvironmentById(ENVIRONMENT_ID)).thenReturn(Optional.of(environment));
+        when(environmentService.save(any(Environment.class))).thenThrow(errorOnSave);
```
lajosrodek, line 168: Please replace `any(Environment.class)` with `environment`. We know exactly which instance to expect in the invocation, so let us use it. This is a general advice for all unit tests.

lajosrodek (changes requested):

lajosrodek:
@juhi-09 Since this needs to be fixed in 2.42, please rebase to `CB-2.42.0` & force-push, and also change the target branch in the PR.

lajosrodek:
The target branch is still not alright.

lajosrodek (approved):
Thanks for the changes. LGTM

topolyai5 (approved):

PR #10541, opened 2021-04-23
CB-12046 s3 calls should not leave the s3 region

gergopapi2:
A customer has a restriction that calls to s3 have to remain within their region (done with a policy from AWS Landing Zone). During datalake install, CM first run the task HDFS dir create failed.
Analysis revealed a connection timeout to s3. The connection, however, was made to s3 region east-us, the default s3 region in the absence of region info, to resolve the region of the bucket. Because of the barrage, however, the call was blocked.

The solution is based on a setting offered by the hadoop-aws component: adding the fs.s3.bucket.BUCKET-NAME.endpoint parameter with the regional endpoint will let hdfs know that it should use the given endpoint for the bucket (storage location base) given.

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/cloud/storage/LocationHelper.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/cloud/storage/LocationHelper.java
@@ -1,0 +1,21 @@
+package com.sequenceiq.cloudbreak.cloud.storage;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class LocationHelper {
+
+    private static final Pattern EXTRACT_BUCKET_PATTERN = Pattern.compile("^s3.?://([^/]+)");
+
+    public String parseS3BucketName(String cloudStorageLocation) {
+        String result = "";
+        Matcher m = EXTRACT_BUCKET_PATTERN.matcher(cloudStorageLocation);
+        if (m.find()) {
+            result = m.group(1);
+        }
+        return result;
+    }
+}
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProvider.java
@@ -3,17 +3,22 @@ package com.sequenceiq.cloudbreak.cmtemplate.configproviders.s3guard;
 import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.hive.HiveMetastoreCloudStorageServiceConfigProvider.HMS_METASTORE_DIR;
 
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
 
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 
+import com.sequenceiq.cloudbreak.cloud.storage.LocationHelper;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils;
 import com.sequenceiq.cloudbreak.template.TemplatePreparationObject;
 import com.sequenceiq.cloudbreak.template.filesystem.s3.S3FileSystemConfigurationsView;
 import com.sequenceiq.common.model.FileSystemType;
 
 @Component
-public class S3GuardConfigProvider {
+public class S3ConfigProvider {
 
     private static final String S3GUARD_METADATASTORE_IMPL_PARAM = "fs.s3a.metadatastore.impl";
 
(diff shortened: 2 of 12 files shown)


template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProviderTest.java
```diff
@@ -68,7 +80,8 @@ public void testGetHdfsServiceConfigsWithS3GuardWithAuthoriativeWarehousePath()
                 "<property><name>fs.s3a.s3guard.ddb.table</name><value>dynamoTable</value></property>" +
                 "<property><name>fs.s3a.s3guard.ddb.region</name><value>region</value></property>" +
                 "<property><name>fs.s3a.authoritative.path</name>" +
-                "<value>s3a://bucket/warehouse/managed</value></property>", sb.toString());
+                "<value>s3a://bucket/warehouse/managed</value></property>" +
+                "<property><name>fs.s3a.bucket.bucket.endpoint</name><value>s3.region.amazonaws.com</value></property>", sb.toString());
     }
```
lajosrodek, line 85: I suggest adding a test case that sets multiple FS locations with different buckets, so that the safety valve XML can contain more than one bucket endpoint overrides.

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProviderTest.java
```diff
@@ -29,12 +35,18 @@
...
+    @Mock
+    private LocationHelper locationHelper;
+
+    @InjectMocks
+    private S3ConfigProvider underTest;
 
     @Before
     public void setUp() {
-        underTest = new S3GuardConfigProvider();
+        MockitoAnnotations.initMocks(this);
```
lajosrodek, line 48: This call is not necessary when using `@RunWith(MockitoJUnitRunner.class)`. On the other hand, I suggest migrating the class to JUnit 5. 😄

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProvider.java
```diff
@@ -35,9 +40,17 @@
 
     private static final String S3GUARD_TABLE_TAG_VALUE = "s3guard";
 
+    private static final String S3_BUCKET_ENDPOINT_PARAM_TEMPLATE = "fs.s3a.bucket.%s.endpoint";
+
+    private static final String S3_ENDPOINT_TEMPLATE = "s3.%s.amazonaws.com";
```
lajosrodek, line 45: Have not checked this, just asking out of curiosity: Is the endpoint DNS name always formed this way? Does it always exist for every possible region?

common/src/main/java/com/sequenceiq/cloudbreak/cloud/storage/LocationHelper.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class LocationHelper {
+
+    private static final Pattern EXTRACT_BUCKET_PATTERN = Pattern.compile("^s3.?://([^/]*)");
```
lajosrodek, line 11: Not a big deal but I suggest replacing `*` with `+`.

lajosrodek (commented):
Some unwanted comments. 😄

gergopapi2:
> Some unwanted comments.

And actually warmly welcome! :D

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProvider.java
gergopapi2: The doc gives the general format of regional endpoints: https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints. It does, however, not state if it works for all regions.
Specifically for s3 endpoints I scrolled through the s3 endpoints and all (except for the two Chinese regions) seemed to comply: https://docs.aws.amazon.com/general/latest/gr/s3.html

doktoric:
@gergopapi2 I seen account-id.s3-control.us-east-2.amazonaws.com endpont as well. Does this mean that the request not even leaving the account ? :D

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProviderTest.java
gergopapi2: I extended a test case to have to different buckets.

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/s3guard/S3ConfigProviderTest.java
gergopapi2: Ok, it is now JUnit 5.

common/src/main/java/com/sequenceiq/cloudbreak/cloud/storage/LocationHelper.java
gergopapi2: Done.

gergopapi2:
This should go to an earlier branch, closing this PR

PR #10542, opened 2021-04-23
CB-12180 Ignore `no modifications to be performed` when creating DNS …

lacikaaa:
…records

In some rare cases when retry is performed both by FMS and the calling service it's possible the ADD request is sent
twice to FreeIPA. This would cause one of them to fail with `EMPTY_MODLIST(4202)` error. This can be ignored.

Tested manually with using debug on real FreeIPA and covered with unit tests.

### DIFF
--- a/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
+++ b/freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
@@ -157,6 +157,22 @@ public class FreeIpaClientExceptionUtil {
         }
     }
 
+    public static <T> Optional<T> ignoreEmptyModExceptionWithValue(FreeIpaClientCallable<T> callable, String message, Object... messageParams)
+            throws FreeIpaClientException {
+        try {
+            return Optional.ofNullable(callable.run());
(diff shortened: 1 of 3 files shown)


freeipa-client/src/main/java/com/sequenceiq/freeipa/client/FreeIpaClientExceptionUtil.java
```diff
@@ -157,6 +157,22 @@ public static void ignoreNotFoundException(FreeIpaClientRunnable runnable, Strin
         }
     }
 
+    public static <T> Optional<T> ignoreEmptyModExceptionWithValue(FreeIpaClientCallable<T> callable, String message, Object... messageParams)
+            throws FreeIpaClientException {
+        try {
+            return Optional.ofNullable(callable.run());
+        } catch (FreeIpaClientException e) {
+            if (isEmptyModlistException(e)) {
+                Optional.ofNullable(message).ifPresentOrElse(
```
christmasferret, line 166: Just curious. 
Is it the same to write 

if (message != null) {
} else {
}

lacikaaa: yes it is the same, just a compact format

PR #10543, opened 2021-04-23
CB-12078 Revert - CB Scaling API that works with internal crns.

foldik:

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterCommonService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterCommonService.java
@@ -152,7 +152,7 @@ public class AutoScaleClusterCommonService  implements ResourcePropertyProvider 
     protected Cluster getClusterByCrnOrName(NameOrCrn nameOrCrn) {
         return nameOrCrn.hasName() ?
                 clusterService.findOneByStackNameAndTenant(nameOrCrn.getName(), restRequestThreadLocalService.getCloudbreakTenant())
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/AutoscaleV4Controller.java
```diff
@@ -119,27 +120,31 @@ public AutoscaleStackV4Response getInternalAutoscaleClusterByName(String name, @
     }
 
     @Override
-    @CheckPermissionByAccount(action = AuthorizationResourceAction.POWERUSER_ONLY)
+    @DisableCheckPermissions
```
lnardai, line 123: I think these can be removed:
    @PreAuthorize("hasRole('AUTOSCALE')")

    @CheckPermissionByAccount(action = AuthorizationResourceAction.POWERUSER_ONLY)

foldik: It was the original version before the change, I wouldn't risk it to introduce a new failure.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackApiViewService.java
```diff
@@ -123,6 +124,7 @@ public StackApiView save(StackApiView stackApiView) {
         ));
     }
 
+    @PreAuthorize("hasRole('INTERNAL')")
```
horadla23, line 127: this isn't needed

horadla23 (approved):
LGTM

foldik:
Tested with schedule based scaling. Worked fine, the original API level exception disappeared.

PR #10544, opened 2021-04-23
CB-12243: DES delete flow exceptions are swallowed and do not resultin flow failure

juhi-09:
CB-12243: DES delete flow exceptions are swallowed and do not resultin flow failure

Fix the issue of swallowing of exceptions - by re-throwing the exception after logging.


lajosrodek:
@juhi-09 Please close this and update #10539 as I suggested there.

lajosrodek (changes requested):
This PR has to be closed.

PR #10546, opened 2021-04-23
CDPCP-3974. Update gradle dependencies (v3 - jackson version updates).

oleewere:
details:
- CB-11608. Upgrade package jackson-datatype-jsr310 to version 2.10.5.1 or above. (2.11.4)
- CB-11600. Upgrade package jackson-databind to version 2.10.5.1 or above (2.11.4)
- CB-11615. Upgrade package jackson-dataformat-cbor to version 2.11.4 or above. (2.11.4)


oleewere:
note: this PR can be related with "cannot convert to object" exceptions on integration test side, although it happened earlier as well (in March), but if those errors will be more common it can be considered to revert this, or would be better to check the integration test or jenkins logs about that

attilapalfi92 (approved):

PR #10550, opened 2021-04-23
CB-12270: Add logging instance metadata for FreeIPA health checks

jamisonbennett:
The FreeIPA instance IDs was added to the logging of the health
checks.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
@@ -45,18 +45,23 @@ public class FreeipaChecker {
         return checkedMeasure(() -> {
             Map<InstanceMetaData, DetailedStackStatus> statuses = new HashMap<>();
             List<RPCResponse<Boolean>> responses = new LinkedList<>();
+            LOGGER.info("Checking FreeIPA status for instance IDs {}",
+                    checkableInstances.stream().map(InstanceMetaData::getInstanceId).collect(Collectors.toList()));
             for (InstanceMetaData instanceMetaData : checkableInstances) {
                 try {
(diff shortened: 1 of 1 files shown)


christmasferret (commented):
this PR is a straightforward LGTM for some review activity.

freeipa/src/main/java/com/sequenceiq/freeipa/sync/FreeipaChecker.java
```diff
@@ -45,18 +45,23 @@
...
+                        newDetailedStackStatus = DetailedStackStatus.AVAILABLE;
                     } else {
-                        statuses.put(instanceMetaData, DetailedStackStatus.UNHEALTHY);
+                        newDetailedStackStatus = DetailedStackStatus.UNHEALTHY;
                     }
+                    LOGGER.info("FreeIpa health check reported {} for {}", newDetailedStackStatus, instanceMetaData);
+                    statuses.put(instanceMetaData, newDetailedStackStatus);
                 } catch (Exception e) {
-                    LOGGER.info("FreeIpaClientException occurred during status fetch: " + e.getMessage(), e);
+                    LOGGER.info("FreeIpaClientException occurred during status fetch for {}: {}", instanceMetaData, e.getMessage(), e);
```
lacikaaa, line 64: in the commit msg instance ID is written but you are logging the whole instancemetadata. Which should it be?

jamisonbennett: I updated the commit message. I logged all the instances in line 48/49 so that we know how many health checks will be performed. But after the health check is performed I wanted to include more than just the instance ID. I don't really need everything in the instance metadata but I think the instance ID, FQDN, and IP are useful.

lacikaaa: sure, thanks for the clarification!

lacikaaa (approved):

PR #10556, opened 2021-04-26
CB-12159 Solr UI load fails - 502 Service Con Error/ HA failover disabled

hreeve-cloudera:
Sticky sessions and noFallback were enabled for the Solr UI Knox endpoints, under the assumption that
the Solr UI requires a valid session. This prevented failover if one of the Solr instances was down.
The Solr team has since verified that the UI is stateless and there's no need for sticky sessions, and
no reason to disable fallback. Updating the topology to turn off sticky sessions and turn on falback.

Tested by manually update the topology.xml document on the data lake and verifying the Solr UI failed
over successfully.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -278,7 +278,7 @@
              {% if 'SOLR' in exposed and 'SOLR_SERVER' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>SOLR</name>
-                 <value>enableStickySession=true;noFallback=true;enableLoadBalancing=true</value>
+                 <value>enableStickySession=false;noFallback=false;enableLoadBalancing=true</value>
              </param>
              {%- endif %}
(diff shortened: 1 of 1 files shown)


hreeve-cloudera:
> Is this supposed to be a partial revert of [c54b3f4](https://github.com/hortonworks/cloudbreak/commit/c54b3f4aed45874c8dd7d0cc9e5c662e7717f684) ?

It's not intended as a revert. Recently the Knox service had some changes and added some new features, which could be enabled in the topology file using the new `enableStickySession`, `noFallback`, and `enableLoadBalancing` fields. I'm not going back to the old fields here. I'm just updating the values of the new fields.

> does moving the value back to `enabled=true;maxFailoverAttempts=3;failoverSleep=1000` make more sense?

No, because we do want to use the new logic associated with the new settings.

> does this value also need to be changed in `topology_api.xml.j2`?

No. The the topology.xml.js file contains the settings Knox uses when forwarding traffic to service UIs. The topology_api.xml.j2 file contains the settings Knox uses when forwarding traffic to service APIs. This problem was only impact the Solr UI, not the Solr API, so we don't need to make changes to the API settings. But good catch looking at that file too. :)

cegganesh84 (commented):
LGTM

brycederriso (approved):
This looks good to me!

PR #10557, opened 2021-04-27
CB-12292 EventBusConfigTest.uncaughtErrorButFlowFound unit test is flaky

sodre90:
The event bus is async this is why I added timeout(2000) to getLastFlowLog
method verification. The problem is sometimes the test verification thread
is faster than the async eventbus error handling thread so it can happen the
flowLogDBService method calls are not invoked when it was verified. I
hope this 1000 ms timeout can avoid the flakiness.


doktoric (approved):
thanks

lacikaaa:
could you add a bit more info to the commit message why would it fix the issue and what would this fix exactly? just in case someone else experience this again or edits the tests. thanks

sodre90:
> could you add a bit more info to the commit message why would it fix the issue and what would this fix exactly? just in case someone else experience this again or edits the tests. thanks

sure, added some explanation to the git commit

lacikaaa (approved):

PR #10559, opened 2021-04-27
CB-12301 Support different Flink port in exposed services according to CDH version

ferenc-csaky:


doktoric:
@ferenc-csaky was it changed or it was never correcT?

ferenc-csaky:
> @ferenc-csaky was it changed or it was never correcT?

It was changed, because 8082 can interfere with other services in some situations.

doktoric:
@ferenc-csaky from which version. How we can handle backward compatibility in this case ?

doktoric (approved):

doktoric:
retest this please

PR #10561, opened 2021-04-27
CB-11868 Create DistroX internal API for COD usecase

foldik:


horadla23 (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/recipe/RecipeService.java
```diff
@@ -116,6 +116,23 @@ public Recipe createForLoggedInUser(Recipe recipe, @Nonnull Long workspaceId, St
         }
     }
 
+    public Recipe createWithInternalUser(Recipe recipe, @Nonnull Long workspaceId, String accountId) {
+        if (recipeViewRepository.findByNameAndWorkspaceId(recipe.getName(), workspaceId).isPresent()) {
```
lnardai, line 120: Can we somehow differentiate recipes that are created by our application or by a regular user?
Simply thinking if some day we wish hide these, can we spare a problematic migration script at this point?

Question stands for all internally created resource.

lnardai (approved):
Looks good, well tested. I'm fine if we tackle the migration question in a separate story.

core/src/main/java/com/sequenceiq/cloudbreak/service/recipe/RecipeService.java
foldik: Good question. I can think of an extra column like `applicationCreated` or `internal`, that we will set true when it was created through an internal API. @lnardai @horadla23 what do you think about this solution?

foldik:
@lnardai I would solve it in this ticket so that we don't need to think about it later which can be a nightmare.

core/src/main/java/com/sequenceiq/cloudbreak/service/recipe/RecipeService.java
horadla23: I would prefer an enum for that (in case if we need something more in the future), like CreationType, with values USER and SERVICE, but anyway I like the idea

horadla23 (approved):
LGTM

PR #10562, opened 2021-04-27
CB-12307: Extract methods, add termination operations enum.

brycederriso:
Trying out some smaller PRs here.

Closes [CB-12307](https://jira.cloudera.com/browse/CB-12307).

* Extract `delete*` methods from the main termination method.
* Prefer two-element Enum to boolean flag value.
* Move `if` out of `deleteLoadBalancers` method.

This is a subtask for [CB-11644](https://jira.cloudera.com/browse/CB-11644), handle Azure load balancers in upscale/downscale operations.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
@@ -30,6 +30,11 @@ public class AzureTerminationHelperService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AzureTerminationHelperService.class);
 
+    private enum TerminationOperations {
+        DOWNSCALE,
+        TERMINATE;
+    }
+
     @Inject
     private AzureResourceGroupMetadataProvider azureResourceGroupMetadataProvider;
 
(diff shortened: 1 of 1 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
```diff
@@ -71,78 +77,97 @@
...
-        return terminateResources(ac, stack, vms, resourcesToRemove, networkResources, true);
+        return terminateResources(ac, stack, vms, resourcesToRemove, networkResources, TerminationOperations.TERMINATE);
     }
 
     private List<CloudResourceStatus> terminateResources(AuthenticatedContext ac, CloudStack stack, List<CloudInstance> vms,
-            List<CloudResource> resourcesToRemove, List<CloudResource> networkResources, boolean deleteWholeDeployment) {
+            List<CloudResource> resourcesToRemove, List<CloudResource> networkResources, TerminationOperations operation) {
         LOGGER.debug("Terminating the following resources: {}", resourcesToRemove);
-        LOGGER.debug("Operation is: {}", deleteWholeDeployment ? "terminate" : "downscale");
+        LOGGER.debug("Operation is: {}", operation.toString().toLowerCase(Locale.ROOT));
```
lacikaaa, line 93: don't over complicate please, just use `operation`
also these 2 logs could be joined into one. I don't see any reason for 2 statements

lacikaaa (changes requested):
thanks for making these changes. I still have the feeling this class/method is a bit procedural and the separate steps could benefit of being in a separate class but it's getting better

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureTerminationHelperService.java
brycederriso: I've changed this to use `operation`.

I'd argue the two log statements is more readable than:
```java
        LOGGER.debug("Terminating the following resources: {}\nOperation is: {}", resourcesToRemove, operation);
```

hreeve-cloudera: I'm going to side with Mihaly here. You could massage the language a little more.
```
LOGGER.debug("Initiating operation {} to terminate the following resources: {}", operation, resourcesToRemove);
```

lacikaaa (approved):

PR #10563, opened 2021-04-27
CB-12197 Introduce authorization tests regarding environment groups

aszegedi:
The related test (`testAddGroupsToEnvironment`) tends to cover:

1. Group
    1. test_add_group_to_env
    2. test_multiple_groups
    3. test_delete_group_from_env

test cases from `CB promotion L0/basic tests` ([cdpmc-qe/tests/regression/users/test_reset_password_flow.py](https://github.infra.cloudera.com/CDH/cdpmc-qe/blob/master/tests/regression/users/test_reset_password_flow.py)).

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
@@ -34,10 +34,10 @@ import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetUs
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetUserSyncStateModelRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.GetUserSyncStateModelResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.Group;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupMembersRequest;
+import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupMembersResponse;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupsForMemberRequest;
 import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupsForMemberResponse;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupMembersResponse;
-import com.cloudera.thunderhead.service.usermanagement.UserManagementProto.ListGroupMembersRequest;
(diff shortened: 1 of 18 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
```diff
@@ -1051,17 +1051,34 @@ public String getIdentityProviderMetadataXml(String requestId, String accountId)
...
-            assignee.setMachineUserNameOrCrn(userCrn);
-        } else {
-            assignee.setUserIdOrCrn(userCrn);
+        UserManagementProto.Assignee roleAssignee;
+
+        UserManagementProto.Assignee.Builder assignee = UserManagementProto.Assignee.newBuilder().setAccountId(accountId);
+        switch (crn.getResourceType()) {
+            case USER:
+                assignee.setUserIdOrCrn(assigneeCrn);
+                roleAssignee = assignee.build();
```
horadla23, line 1065: why do we need `roleAssignee` variable? we can return with `assignee.build();` after switch case as before I think

horadla23 (changes requested):
just one request

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/UmsClient.java
aszegedi: This is used in the case related logs, for example:
```LOGGER.info("Found user '{}' for assignee Crn: '{}'.", roleAssignee.getUserIdOrCrn(), assigneeCrn);```
or
```LOGGER.info("Found group '{}' for assignee Crn: '{}'.", roleAssignee.getGroupNameOrCrn(), assigneeCrn);```
So for the return variable is not needed to build the assignee as `roleAssignee`. However for logging purposes this is needed, because of we can get `roleAssignee.getUserIdOrCrn()`, `roleAssignee.getMachineUserNameOrCrn()` and `roleAssignee.getGroupNameOrCrn()` on this way based on the case `USER`, `MACHINE_USER` or `GROUP`.

horadla23: for logging you can use the assigneeCrn, which was set before

horadla23: but, i am not sure i get the logging itself, it logs the same crn twice, you can just log the assigneeCrn once

horadla23: so now you set the assigneeCrn and you log it twice, once from roleAssignee (which was set with assigneeCrn), once the assigneeCrn itself

horadla23 (approved):

PR #10565, opened 2021-04-27
CB-11720 Tie volume sets to discovery FQDNs

daszabo:
Some services are sensitive to have the volumes reattached to a machine
with the same FQDN.

* Discovery FQDN now also stored in volumesets
* Discovery FQDN saved the same time as mount info
* For already running clusters, discovery FQDN ges saved at the start of repair
* During bootstraping new nodes, discovery FQDN gets read from volume sets to
set the proper hostname based on instance id

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
@@ -11,7 +11,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -31,8 +30,10 @@ import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
 import com.sequenceiq.cloudbreak.cloud.model.CloudVmInstanceStatus;
 import com.sequenceiq.cloudbreak.cloud.model.Image;
 import com.sequenceiq.cloudbreak.cloud.model.InstanceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.VolumeSetAttributes;
 import com.sequenceiq.cloudbreak.cloud.scheduler.CancellationException;
 import com.sequenceiq.cloudbreak.cluster.service.ClusterComponentConfigProvider;
+import com.sequenceiq.cloudbreak.cluster.util.ResourceAttributeUtil;
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.common.service.HostDiscoveryService;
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -328,14 +333,19 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
         Set<String> clusterNodeNames = stack.getNotTerminatedInstanceMetaDataList().stream()
                 .map(InstanceMetaData::getShortHostname).collect(Collectors.toSet());
 
-        Iterator<String> iterator = recoveryHostNames.iterator();
+        List<Resource> diskResources = stack.getDiskResources();
+        Map<String, Optional<String>> imIdToFQDN = diskResources.stream()
+                .collect(Collectors.toMap(Resource::getInstanceId, volumeSet -> {
+                    Optional<VolumeSetAttributes> volumeSetAttributes = resourceAttributeUtil.getTypedAttributes(volumeSet, VolumeSetAttributes.class);
+                    return volumeSetAttributes.map(VolumeSetAttributes::getDiscoveryFQDN);
```
lacikaaa, line 340: would be nice to extract this as a separate method

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -328,14 +333,19 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
...
+                }));
         for (InstanceMetaData im : metaDataSet) {
             Node node = createNodeAndInitFqdnInInstanceMetadata(stack, im, clusterDomain, hostGroupNodeIndexes, clusterNodeNames);
             if (upscaleCandidateAddresses.contains(im.getPrivateIp())) {
                 // use the hostname of the node we're recovering instead of generating a new one
                 // but only when we would have generated a hostname, otherwise use the cloud provider's default mechanism
                 if (recoveredNodes && isNoneBlank(node.getHostname())) {
-                    node.setHostname(iterator.next().split("\\.")[0]);
+                    node.setHostname(imIdToFQDN.get(im.getInstanceId()).get().split("\\.")[0]);
```
lacikaaa, line 348: shouldn't `imIdToFQDN.get(im.getInstanceId()).isPresent()` be added to the if condition? it's strange calling get without a check first

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -81,6 +81,10 @@
             "LEFT JOIN FETCH ig.securityGroup WHERE s.id= :id AND (s.type is not 'TEMPLATE' OR s.type is null)")
     Optional<Stack> findOneWithLists(@Param("id") Long id);
 
+    @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.resources LEFT JOIN FETCH s.instanceGroups ig LEFT JOIN FETCH ig.instanceMetaData "
+            + "LEFT JOIN FETCH ig.template LEFT JOIN FETCH ig.securityGroup WHERE s.id= :id AND (s.type is not 'TEMPLATE' OR s.type is null)")
```
lacikaaa, line 85: all these extra stuff is really necessary? I expected to see only one join for the resources here

lacikaaa (changes requested):

keyki:
What about existing clusters?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
daszabo: Yes, there is still functionality that requires the joins unfrotunatley.

daszabo:
> What about existing clusters?

@keyki For clusters existed before this change, the fqdn will be filled during the downscale part of the repair process: https://github.com/hortonworks/cloudbreak/pull/10565/files#diff-3d0300379711862e75a354c93207f40645ea72c09869516b59d765d89e473c9fR134

lacikaaa (approved):

PR #10566, opened 2021-04-28
CB-12201 Stop the instance even if the CM stop is failed. Need to the…

topolyai5:
… cost saving. I had unified the cm error message fetcher. We had 3 solutions, I left one.


doktoric (approved):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/AbstractClouderaManagerApiCheckerTask.java
```diff
@@ -101,6 +103,17 @@ private boolean isToleratedError(ApiException e) {
                 || e.getCause() instanceof SocketTimeoutException;
     }
 
+    protected String getResultMessageWithDetailedMessagePostFix(ApiCommand apiCommand, CommandsResourceApi commandsResourceApi) {
```
gregito, line 106: not sure, that's why I'm asking but what ensures that the apiCommand is always non-null? since one line below we're accessing it and that could cause an NPE

topolyai5: What line do you mean?

But you are right, the NPE check worthy, in that case returns with an empty string.

PR #10569, opened 2021-04-28
CB-12293: Disable Nifi auto-restart in Ambari during downscale

tiborpopovics:


core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ambari/AmbariDecommissioner.java
```diff
@@ -650,27 +661,38 @@ private PollingResult waitForRegionServerDecommission(Stack stack, AmbariClient
...
+        hostsRunService.forEach(nifiNode -> {
             LOGGER.info("Retiring NiFi node {}", nifiNode);
             Pair<PollingResult, Exception> retireResult = Pair.of(SUCCESS, null);
             try {
                 int requestId = ambariClient.retire(List.of(nifiNode), "NIFI", NIFI);
                 retireResult = ambariOperationService.waitForOperations(stack, ambariClient,
                         singletonMap("Retiring NiFi node", requestId), RETIRE_NIFI_NODE_AMBARI_PROGRESS_STATE);
             } catch (Exception e) {
-                handleRetireError(nifiNode, e);
+                LOGGER.warn("Retire command failed.", e);
```
horadla23, line 672: if `ambariOperationService.waitForOperations` throws an exception, `retireResult` will be SUCCESS based on line 666 and we will not handle failure. I would set default result for `retireResult` to failed or I would give new value to `retireResult` here

horadla23 (changes requested):
please modify unit tests accordingly and also please add new unit tests, since there are more cases here

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/ambari/NifiDecommissionService.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private void retireNode(Stack stack, AmbariClient ambariClient, String nifiNode) {
+        LOGGER.info("Retiring NiFi node {}", nifiNode);
+        Pair<PollingResult, Exception> retireResult = Pair.of(SUCCESS, null);
+        int requestId = 0;
+        try {
+            requestId = ambariClient.retire(List.of(nifiNode), NIFI, NIFI_MASTER);
+        } catch (Exception e) {
+            LOGGER.warn("Calling retire command failed.", e);
```
horadla23, line 65: please extend the message with information about that we will decommission/downscale the node anyway in this case

horadla23 (approved):

PR #10571, opened 2021-04-28
CB-12229: Clean up Azure Load Balancer code

brycederriso:
Closes [CB-12229](https://jira.cloudera.com/browse/CB-12229).

* Add unit test to `AzureMetaDataCollector`, when collecting private ips.
* Fix imports for `AzureMetaDataCollectorTest`.
* Use an `Optional<String>` for LB ips instead of a `null` String.
* Use a switch/case for `AzureClient` when retrieving load balancer IPs based on the load balancer type.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureMetadataCollector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureMetadataCollector.java
@@ -10,7 +10,6 @@ import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


hreeve-cloudera (approved):
LGTM

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureMetadataCollector.java
```diff
@@ -143,17 +142,19 @@ private String getLocalityIndicator(Integer faultDomainCount, CloudContext cloud
             String loadBalancerName = AzureLoadBalancer.getLoadBalancerName(type, stackName);
             LOGGER.debug("Attempting to collect metadata for load balancer {}, type {}", loadBalancerName, type);
             try {
-                String ip = null;
+                Optional<String> ip;
```
lacikaaa, line 145: you could initialize it with empty

PR #10575, opened 2021-04-29
CB-12302 Move and migrate EdgeNode templates to CB code

foldik:
I add edge node templates to all of our runtimes appear to be in our code.

I tested the migration, it made the existing templates default, visible to other users as well, not just for the creator.

On mow-dev I found these edge node templates:

7.2.8 COD Edge Node
7.2.7 COD Edge Node
7.2.6 COD Edge Node
7.2.3 COD Edge Node
7.2.2 COD Edge Node
7.2.1 COD Edge Node
7.2.0 COD Edge Node
7.1.1 COD Edge Node

But we have 7.0.2, 7.1.0, 7.2.9, 7.2.10 runtimes in our code (I assume COD would create them anyway) so I add:

7.0.2 COD Edge Node
7.1.0 COD Edge Node
7.2.9 COD Edge Node
7.2.10 COD Edge Node

as well.

@lnardai @horadla23 @sodre90 @petersomogyi please correct me if some these are not needed and sure as hell that it won't break anything if I don't add them.

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -403,7 +403,10 @@ cb:
           7.2.2 - Flow Management Heavy Duty with Apache NiFi, Apache NiFi Registry=cdp-flow-management;
           7.2.2 - Data Discovery and Exploration=cdp-dde;
           7.2.2 - Streaming Analytics Light Duty with Apache Flink=cdp-flink-light;
-          7.2.2 - Streaming Analytics Heavy Duty with Apache Flink=cdp-flink-heavy
+          7.2.2 - Streaming Analytics Heavy Duty with Apache Flink=cdp-flink-heavy;
+          7.2.2 COD Edge Node=cdp-cod-edge-node
+        7.2.3: >
+          7.2.3 COD Edge Node=cdp-cod-edge-node
         7.2.6: >
           7.2.6 - SDX Light Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx;
           7.2.6 - SDX Medium Duty: Apache Hive Metastore, Apache Ranger, Apache Atlas=cdp-sdx-medium-ha;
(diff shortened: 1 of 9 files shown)


horadla23 (approved):
LGTM

lnardai:
So in real life scenario, where there are already an existing default templates with the same name, the migration simply happens without any error?

foldik:
@lnardai if there was an old template, created by COD, then it will be an other USER_MANAGED template and this change will make it a default template. If there is another DEFAULT template with different content then it is a different scenario that we need to specify what should happen.

petersomogyi (commented):
COD is not supported before 7.2.2 so those templates can be dropped.

core/src/main/resources/defaults/blueprints/7.0.2/cdp-cod-edge-node.bp
```diff
@@ -0,0 +1,53 @@
+{
+  "description": "COD Edge Node cluster template",
+  "blueprint": {
+    "cdhVersion": "7.0.2",
+    "displayName": "opdb",
+    "cmVersion": null,
+    "repositories": [],
+    "products": [],
```
petersomogyi, line 8: These values are not defined in our original blueprint but I assume these are the default values.

foldik: Yes, this is something I got from templates on mow-dev, so these are the default values.

lnardai (approved):

foldik:
@petersomogyi  One thing to note is that I saw the tag `is_cod_blueprint` for these blueprints.

```json
    {
      "name": "7.2.2 COD Edge Node",
      "description": "COD Edge Node cluster template",
      "stackType": "CDH",
      "stackVersion": "7.2.2",
      "hostGroupCount": 1,
      "status": "USER_MANAGED",
      "tags": {
        "is_cod_blueprint": "true"
      }
    },
```
Do we need to add those tags because the current migration doesn't add these to the edge node templates?

petersomogyi:
> @petersomogyi One thing to note is that I saw the tag `is_cod_blueprint` for these blueprints.
> 
> ```json
>     {
>       "name": "7.2.2 COD Edge Node",
>       "description": "COD Edge Node cluster template",
>       "stackType": "CDH",
>       "stackVersion": "7.2.2",
>       "hostGroupCount": 1,
>       "status": "USER_MANAGED",
>       "tags": {
>         "is_cod_blueprint": "true"
>       }
>     },
> ```
> 
> Do we need to add those tags because the current migration doesn't add these to the edge node templates?

I've confirmed with the team that the `is_cod_blueprint` is not needed. It was a side-effect of reusing the template creation login.

petersomogyi (approved):

PR #10577, opened 2021-04-29
CB-11923 Loadbalancer takes ~2.5 mins to failover requests when one of the Knox host is down

hreeve-cloudera:
Reduces the AWS target group HealthyThresholdCount to the minimum value of 2 from the default of 3,
and the HealthCheckIntervalSeconds to 10s from the default of 30s. This will reduce the failover
time for the load balancer to detect an unhealthy node to ~20s.


cegganesh84 (approved):
Thanks!

PR #10579, opened 2021-04-29
CB-10579 Clean up unused targetgroup and loadbalancer columns

hreeve-cloudera:
Removes the deprecated LoadBalancer and TargetGroup related fields and methods that were replaced
in CB-10578.


yg-gd (approved):

orngejaket:
Do we need any sql changes in the cleanup?

hreeve-cloudera:
> Do we need any sql changes in the cleanup?

Those need to be a separate step. So I realized the first time I tried this I did it wrong. I need to remove the references to the old columns from the code, which this change does. Then after this change is released, I'll remove the columns themselves in 2.45. That's to ensure the code that references the columns is officially gone before we remove the columns, to prevent upgrade problems.

PR #10581, opened 2021-04-30
CB-12345 Fix BindUsercreation for child environments

lacikaaa:
Child environment lookup was missing when creating bind users.
This caused an not found exception on FMS side which prevented YCLOUD cluster creation.


Bajzathd (approved):

lacikaaa:
tested locally, by creating 2 envs, one on ycloud and launching SDX in both of them:
image
LDAP configured:
image
Kerberos configured:
image

keyki (approved):

PR #10587, opened 2021-04-30
CB-12039 Add default freeipa image to the core svc and merge custom image with the source image where the source image can be a FreeIPA or a Cloudbreak image as well 

pkedvessy:


jszeles:
retest this please

PR #10595, opened 2021-05-03
CB-12244 Key Vault access permission grant to DES SP may fail

lajosrodek:
* Error handling enhancements:
  * `AzureEncryptionResources`:
    * Key Vault permission granting is executed as a fault-tolerant logic with retries. At most 15 retries will be performed using delays of 2, 4, 8, 10, 10... sec. (Hence the raw max duration for these operations is expected to be 124 sec.) All exceptions are caught and wrapped in a `Retry.ActionFailedException` for sake of retries.
* Code quality enhancements:
  * Add new constructor variants and a factory method to `Retry.ActionFailedException`.
* Testing:
  * Manual verification in local CB with various combinations of permissions in the Azure credential app role.
  * Added new unit tests and updated existing ones.

### DIFF
--- a/common/src/test/java/com/sequenceiq/cloudbreak/service/RetryTest.java
+++ b/common/src/test/java/com/sequenceiq/cloudbreak/service/RetryTest.java
@@ -1,0 +1,79 @@
+package com.sequenceiq.cloudbreak.service;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class RetryTest {
+
+    private static final String MESSAGE = "message";
+
+    @Test
+    void actionFailedExceptionConstructorTestWhenDefault() {
+        Retry.ActionFailedException underTest = new Retry.ActionFailedException();
+
+        assertThat(underTest.getMessage()).isNull();
+        assertThat(underTest.getCause()).isNull();
+    }
+
+    @Test
+    void actionFailedExceptionConstructorTestWhenMessage() {
+        Retry.ActionFailedException underTest = new Retry.ActionFailedException(MESSAGE);
+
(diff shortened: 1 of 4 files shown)


common/src/test/java/com/sequenceiq/cloudbreak/service/RetryTest.java
```diff
@@ -0,0 +1,79 @@
...
+    void actionFailedExceptionConstructorTestWhenMessageAndCauseAndEnableSuppressionAndWritableStackTrace() {
+        Throwable cause = new RuntimeException();
+        Throwable suppressed = new RuntimeException();
+
+        Retry.ActionFailedException underTest = new Retry.ActionFailedException(MESSAGE, cause, false, false);
+        underTest.addSuppressed(suppressed);
+
+        assertThat(underTest.getMessage()).isEqualTo(MESSAGE);
+        assertThat(underTest.getCause()).isSameAs(cause);
+        assertThat(underTest.getSuppressed()).isEmpty();
```
bergerdenes, line 57: why?

lajosrodek: Because that is how the `enableSuppression` flag works. TBH I have never used this constructor variant, but the impl in `Throwable` says that setting `enableSuppression = false` will make all `addSuppressed()` calls ignored. So here I am checking if exception suppression indeed works as expected.

There is no real alternative way to check this; the `enableSuppression` flag is not even saved in some `private` field, it is only used to initialize another field that controls the behavior of `addSuppressed()`.

topolyai5 (approved):

TheTinkerDad (approved):

PR #10599, opened 2021-05-04
CB-12396 Logging, error handling and code style refactor for DES logics

lajosrodek:
Covering DES creation & deletion.

* Error handling enhancements:
  * Wrap all DES and Key Vault related methods of `AzureClient` in `handleAuthException()`.
  * `AzureEncryptionResources`:
    * Catch all exceptions in the `public` methods and wrap them in a `CloudConnectorException`. `CloudException` (Azure) and `Retry.ActionFailedException` are treated in a special way; see the related methods of `AzureUtils` and the new class `CloudExceptionConverter`.
    * DES creation is executed in a sync poller to await the resource readiness. Poller failures will result in a `CloudConnectorException`. Note that the DES MSI SP existence is not checked; see below for more details.
    * The implementation of #10595 / [CB-12244](https://jira.cloudera.com/browse/CB-12244) has been updated so that all exceptions are caught and wrapped as a `CloudConnectorException`. (The latter will be still wrapped in a `Retry.ActionFailedException` for sake of retries, but the actual cause will be unwrapped later before exiting the caller `public` method.) The retry logic has been also applied to the deletion of DES resources; see the wrapped `AzureClient.deleteDiskEncryptionSet()` call in `deleteDiskEncryptionSetOnCloud()`.
* Code quality enhancements:
  * `AzureClient`:
    * Extract `ComputeManager` as a new field.
    * Rename `getDiskEncryptionSet()` to `getDiskEncryptionSetByName()`.
  * Add `toString()` implementations for `CloudCredential`, `DiskEncryptionSetCreationRequest`, `DiskEncryptionSetDeletionRequest`, and `ExtendedCloudCredential`.
  * Clean up superfluous fields in `DiskEncryptionSetCreationRequest` that are also available in the `CloudContext`.
  * Rename `principalId` fields and related accessor methods to `principalObjectId`; see `AzureClient` and `CreatedDiskEncryptionSet`.
  * Various refactorings, improved logging and exception handling in `AzureEncryptionResources`, `EnvironmentEncryptionService`, `ResourceEncryptionDeleteHandler`, and `ResourceEncryptionInitializationHandler`.
  * Add `PollPredicateStateTask`, a new specialized kind of `PollTask`.
* Testing:
  * Manual verification in local CB with various combinations of permissions in the Azure credential app role.
  * Added new unit tests and updated existing ones.

The existence of the DES SP cannot be easily checked. That would need powerful special AD API permissions for the credential app itself that most customers would never grant. In particular, `AzureClient` was supposed to be extended with the following method:

```
    public ServicePrincipal getServicePrincipalByObjectId(String principalObjectId) {
        return handleAuthException(() ->
                azure.accessManagement()
                        .servicePrincipals()
                        .getById(principalObjectId));
    }
```

Executing the above method typically resulted in an exception like this:

```
com.microsoft.azure.management.graphrbac.GraphErrorException: Status code 403, {"odata.error":{"code":"Authorization_RequestDenied","message":{"lang":"en","value":"Insufficient privileges to complete the operation."},"requestId":"b16d24e8-5b3a-4458-b66a-2b4f27bfb3c9","date":"2021-04-28T13:55:07"}}
```

In order for this to succeed, the credential app should have been modified to possess special AD API permissions (`Application.Read.All`, `Application.ReadWrite.All`, `Application.ReadWrite.OwnedBy`, `Directory.Read.All`), but getting those would have allowed the credential app to query the details of any other apps in the subcription. See https://docs.microsoft.com/en-us/graph/api/serviceprincipal-get?view=graph-rest-1.0&tabs=http and https://docs.microsoft.com/en-us/graph/permissions-reference#application-resource-permissions for the related REST API and permissions involved.

Though there is a system assigned managed identity (MSI) as well sitting behind the SP (with the MSI `displayName` matching the DES name), querying the properties of this identity (in order to check its existence) would also need an additional powerful `Action` (`Microsoft.ManagedIdentity/identities/read`) for the role assigned to the credential app, which is probably not desirable. See https://docs.microsoft.com/en-us/azure/role-based-access-control/resource-provider-operations#microsoftmanagedidentity for the permissions related to operations on MSIs.

### DIFF
--- a/cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
+++ b/cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
@@ -1,10 +1,14 @@
 package com.sequenceiq.cloudbreak.cloud.azure;
 
 import static com.sequenceiq.cloudbreak.cloud.model.Location.location;
+import static com.sequenceiq.cloudbreak.cloud.model.Region.region;
+import static com.sequenceiq.common.api.type.CommonStatus.CREATED;
+import static com.sequenceiq.common.api.type.ResourceType.AZURE_DISK_ENCRYPTION_SET;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
@@ -14,10 +18,14 @@ import static org.mockito.Mockito.when;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Supplier;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.function.Executable;
+import org.mockito.ArgumentCaptor;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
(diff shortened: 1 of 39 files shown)


cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -85,77 +116,109 @@ public void testVariantShouldReturnAzurePlatform() {
...
+        initExceptionConversion();
 
-        assertEquals(assertThrows(IllegalArgumentException.class, () -> underTest.createDiskEncryptionSet(requestedSet)).getMessage(),
+        verifyException(IllegalArgumentException.class, () -> underTest.createDiskEncryptionSet(requestedSet),
                 "Customer Managed Key Encryption for managed Azure disks is supported only if the CDP resources " +
                         "are in the same resource group as the vault.");
     }
 
+    private <T extends Throwable> void verifyException(Class<T> expectedType, Executable executable, String messageExpected) {
+        CloudConnectorException cloudConnectorException = assertThrows(CloudConnectorException.class, executable);
```
bergerdenes, line 162: hint: assertj has fancy methods for exception validation

bergerdenes (approved):
LGTM

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
lajosrodek: I know, but some folks (notably @doktoric 😄 ) prefer the JUnit `assertThrows()` over AssertJ `assertThatCode()`.

PR #10602, opened 2021-05-05
CB-12078 ums client should be created with original internal actor al…

horadla23:
…ways


smaniraju (commented):
Since this change is across all CB-UMS interactions better to get this PR reviewed from UMS perspective more than scaling perspective.

horadla23:
> Since this change is across all CB-UMS interactions better to get this PR reviewed from UMS perspective more than scaling perspective.

@smaniraju based on this https://github.infra.cloudera.com/thunderhead/thunderhead/blob/master/docs/cdp/grpc_metadata_contexts.md#actor-crn I think it is not required to involve UMS here, since in their readme there is a recommendation to use internal actor in case of service-service call

smaniraju (approved):
Based on the note shared about recommended usage of internal-crn for all service-to-service UMS interactions, this change LGTM.

PR #10612, opened 2021-05-10
CB-12050: Handle upgrade validation flow in sdx.

tiborpopovics:


keyki:
Generally, the PR looks fine, however, we need to separate the API changes from the rest of the code. This is because during a control plane upgraded it's not guaranteed which backend will receive the GET request and the operation might fail. The other option is to somehow handle this scenario as well when we try to fetch the flow log.

tiborpopovics:
I'm closing this PR because I'll create separate PR for the API change and for the feature as well.

PR #10614, opened 2021-05-10
CB-12484 remove unused actorcrn parameters from umsclient, cb uses internal actor always

horadla23:


lnardai (approved):
Somehow it would be nice to let the caller now that all these call are done with internal actor for a particular user.

PR #10615, opened 2021-05-10
CB-12280 Concluson checker parses and understand nodestatus response …

lturcsanyi:
…after failed flow

- Add new conclusion checker step for salt bootstrap nodes
- Improve logging
- Measure conclusion step durations
- Send events to the UI only if a conclusion found
- Enable conclusion checker event sending locally by default


core/src/main/java/com/sequenceiq/cloudbreak/conclusion/ConclusionChecker.java
```diff
@@ -25,25 +27,28 @@ public ConclusionChecker(List<ConclusionStep> conclusionSteps) {
...
             while (hasMoreSteps() && doNext()) {
                 ConclusionStep conclusionStep = conclusionSteps.get(actualStep++);
                 LOGGER.debug("Conclusion step: {}", conclusionStep);
 
-                ConclusionStepResult stepResult = conclusionStep.check(resourceId);
-                LOGGER.debug("Conclusion step {}, conclusion: {}", stepResult.isStepFailed() ? "failed" : "succeeded", stepResult.getConclusion());
-                if (stepResult.getConclusion() != null) {
-                    conclusions.add(stepResult.getConclusion());
+                Conclusion conclusion = measure(() -> conclusionStep.check(resourceId),
+                        LOGGER, "Conclusion step finished in {} ms for resourceId {}, step {}", resourceId, conclusionStep.getClass().getSimpleName());
```
lnardai, line 37: Do we have a a CRN here? If it's included in the MdcContext that's fine, but the resourceId carries little information.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/ConclusionChecker.java
```diff
@@ -25,25 +27,28 @@ public ConclusionChecker(List<ConclusionStep> conclusionSteps) {
...
+                if (conclusion.isStepFailed()) {
+                    LOGGER.debug("Conclusion step failed, conclusion: {}, details: {}", conclusion.getConclusion(), conclusion.getDetails());
```
lnardai, line 39: What does step failed mean in this context? 
Is it that we were not able to execute the check due too unexpected error or the result we were expecting was marking some form of error?
So in other words did it encounter a failure during it's execution, or properly detected an error?

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/Conclusion.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static Conclusion succeeded(Class<? extends ConclusionStep> conclusionStepClass) {
+        return new Conclusion(false, null, null, conclusionStepClass);
+    }
+
+    public static Conclusion failed(String conclusion, String details, Class<? extends ConclusionStep> conclusionStepClass) {
+        return new Conclusion(true, conclusion, details, conclusionStepClass);
+    }
+
+    public boolean isStepFailed() {
```
lnardai, line 28: This StepFailed notation is somewhat ambiguous to me. 
Does it detected a failure before it could finish or is there execution error?

Might be better with 2 different values? 
```suggestion
    public boolean isStepFailed() {
```
IsCompleted() and isFailureFound()

horadla23 (approved):
LGTM

core/src/test/java/com/sequenceiq/cloudbreak/conclusion/step/SaltBootstrapCheckerConclusionStepTest.java
```diff
@@ -0,0 +1,97 @@
+package com.sequenceiq.cloudbreak.conclusion.step;
```
lnardai, line 1: Nice!

lnardai (approved):
LGTM! I've just left that one comment about isStepFailed ambiguity.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/ConclusionChecker.java
lturcsanyi: The latter, it means the result we were expecting was marking some form of error.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/ConclusionChecker.java
lturcsanyi: MdcContext should contain related data.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/Conclusion.java
lturcsanyi: The conclusion object is created only when the execution is finished. IsCompleted always would be true.

PR #10618, opened 2021-05-10
CB-12428 Introduce custom UMS user Json path

aszegedi:
Manowar Dev API E2E requires special test users, these are available at `f8e2f110-fc7e-4e46-ae55-381aacc6718c` account. So we need to fetch these before API E2E tests are running on Mow-Dev.

Actual `fetch UMS users` solution was hard-coded. So I've introduced environment variables and move the default parameters to the project's `application.yml` as well. In the future we can change the fetching UMS secrets parameters externally on Jenkins and not require code change on Cloudbreak side.


horadla23 (commented):
the change looks good to me, I would add a section about how to use this in integration-test readme

horadla23 (approved):
readme will come in another PR

PR #10621, opened 2021-05-11
CB-12273. Set up YARN container placement preference rules.

sidseth:
- Placement rules for AM / non-AM conrtainers being set based on CDP
  version for now.

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/yarn/YarnResourceManagerRoleConfigProvider.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/yarn/YarnResourceManagerRoleConfigProvider.java
@@ -1,7 +1,13 @@
 package com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn;
 
 import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.config;
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn.YarnConstants.ATTRIBUTE_NAME_NODE_INSTANCE_TYPE;
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn.YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_COMPUTE;
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.yarn.YarnConstants.ATTRIBUTE_NODE_INSTANCE_TYPE_WORKER;
+import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CDPD_VERSION_7_2_11;
+import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.isVersionNewerOrEqualThanLimited;
 
+import java.util.LinkedList;
 import java.util.List;
 
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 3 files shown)


tangzhankun (commented):
Thanks for the patch! @sidseth 
It looks good to me. Just one below minor question:

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/yarn/YarnResourceManagerRoleConfigProvider.java
```diff
@@ -77,4 +95,21 @@ private String getCapacitySchedulerValveValue() {
...
+    private boolean isContainerPlacementConfigRequired(TemplatePreparationObject source) {
+        String cdpdVersion = source.getBlueprintView().getProcessor().getStackVersion() == null ?
+                "" : source.getBlueprintView().getProcessor().getStackVersion();
+        return isVersionNewerOrEqualThanLimited(cdpdVersion, CDPD_VERSION_7_2_10);
+    }
+
+    private String getYarnSiteContainerPlacementSafetyValveValues() {
+        return
+                ConfigUtils.getSafetyValveProperty("yarn.resourcemanager.am.placement-preference-with-node-attributes",
+                        String.format("ORDER NODES IN NodeInstanceType WITH %s > %s",
```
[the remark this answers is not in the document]
sidseth: Will switch this over as well. Thanks for the review.

sidseth:
Tried on a cluster. Will remove the DO_NOT_MERGE label once verified by yarn folks.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/yarn/YarnResourceManagerRoleConfigProvider.java
```diff
@@ -20,13 +26,22 @@
...
         switch (roleType) {
             case YarnRoles.RESOURCEMANAGER:
-                return List.of(
-                        config(CAPACITY_SCHEDULER_SAFETY_VALVE, ConfigUtils.getSafetyValveConfiguration(getCapacitySchedulerValveValue())),
-                        config(CAPACITY_SCHEDULER_CLASS, "org.apache.hadoop.yarn.server.resourcemanager.scheduler.capacity.CapacityScheduler"));
+                List<ApiClusterTemplateConfig> configs = new LinkedList<>();
+                if (isContainerPlacementConfigRequired(source)) {
+                    configs.add(config(RESOURCE_MANAGER_YARN_SITE_SAFETY_VALVE, getYarnSiteContainerPlacementSafetyValveValues()));
+                }
+                if (isLlapBasedConfigNeeded((CmTemplateProcessor) source.getBlueprintView().getProcessor())) {
```
cegganesh84, line 39: There is no LLAP in datahub clusters. It looks like there is a behavior change for pre-7.2.11 clusters with this change. Earlier we used to set this value always, but after this change, this will be only set if LLAP is enabled. A customer deleting and recreating the 7.2.9 cluster will see different values for the capacity scheduler.

cegganesh84 (changes requested):
Just want to clarify one thing before approving.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/yarn/YarnResourceManagerRoleConfigProvider.java
sidseth: Yes. There's no LLAP. This code, in fact, should not exist. Don't want to make that change in this jira though.

In terms of a behaviour change - that is not the case, even though it looks like it.

Earlier isConfigurationNeeded would filter out setting the LLAP configs (that has the LLAP check). Now, since there are 2 different aspects which could trigger YARN RM configs, isConfigurationNeeded is more permissive, and whether the LLAP configs are required or not is checked again in this method.

cegganesh84: Thanks!

cegganesh84 (approved):
LGTM

PR #10625, opened 2021-05-11
CB-12443 Decommission InstanceIds Should be part of RequestBody

smaniraju:
Instance Ids for decommission should be part of RequestBody so that it does not cause BufferOverflowException.


smaniraju:
Tested that 200 node downscale works on GCP with this change.

sidseth (approved):

PR #10626, opened 2021-05-11
CB-12402 CB controllers should use workspaceid from threadlocal

horadla23:


foldik (approved):
LGTM

PR #10627, opened 2021-05-11
CB-12397 Validate API call if user is internal but API is not interna…

horadla23:
…lly callable


foldik (approved):
LGTM

horadla23:
retest this please

PR #10630, opened 2021-05-11
CB-11892 Unit tests for google cloud modul

doktoric:


cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/AbstractGcpResourceBuilderTest.java
```diff
@@ -0,0 +1,30 @@
+package com.sequenceiq.cloudbreak.cloud.gcp;
+
+import org.junit.Assert;
+import org.junit.Test;
```
biharitomi, line 4: I think Junit5 jupiter API would be better choice to have new tests with.

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpAttachedDiskResourceBuilderTest.java
```diff
@@ -194,6 +197,7 @@ void setUp() throws Exception {
 
     @Test
     void testBuildWithDiskEncryption() throws Exception {
+        setUpBuild();
```
biharitomi, line 200: Why do we call this explicitly instead of '@BeforeEach'?

integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
```diff
@@ -52,7 +52,7 @@
 
 @EnableAutoConfiguration(exclude = {DataSourceAutoConfiguration.class, DataSourceTransactionManagerAutoConfiguration.class,
         HibernateJpaAutoConfiguration.class})
-@ComponentScan(basePackages = "com.sequenceiq.it")
+@ComponentScan(basePackages = { "com.sequenceiq.it", "com.sequenceiq.cloudbreak.cloud.gcp.util" })
```
biharitomi, line 55: Why do we need this?

integration-test/src/main/java/com/sequenceiq/it/config/IntegrationTestConfiguration.java
```diff
@@ -31,7 +31,7 @@
 import io.jaegertracing.spi.Sampler;
 
 @Configuration
-@ComponentScan("com.sequenceiq.it")
+@ComponentScan(basePackages = { "com.sequenceiq.it", "com.sequenceiq.cloudbreak.cloud.gcp.util" })
```
biharitomi, line 34: Why do we need this package for integration test module?

biharitomi (changes requested):

integration-test/src/main/java/com/sequenceiq/it/IntegrationTestApp.java
doktoric: because now we have a bean and not a static class

integration-test/src/main/java/com/sequenceiq/it/config/IntegrationTestConfiguration.java
doktoric: because now we have a bean and not a static class

biharitomi (approved):

PR #10636, opened 2021-05-11
CB-12444 Remove entitlement check for endpoint gateway feature

hreeve-cloudera:
The endpoint gateway feature is GA, so removing the entitlement checks. The entitlement itself
still exists and will be removed later.


sxxgrc (approved):
Looks good! I believe there are some checks for this in the UI code, so I will need to change those to accompany this.

hreeve-cloudera:
> Is it possible to remove the method in EntitlementService with this change?

I was planning on doing that with a separate change, because that method is used in various places for the mock tests and I didn't want to inadvertently break things there. I planned the overall change to be several parts in sequence:
1. Remove the entitlement checks in the CB code.
2. Remove the entitlement check from the UI code.
3. Mark the entitlement as deprecated on the control plane side in the thunderhead code.
4. Then, remove the actual entitlement and its related methods from `EntitlementService` and the related test classes.

PR #10638, opened 2021-05-11
CB-12288 Use LB frontend even if LB is not registered with PEM

yg-gd:
tested locally by standing up stack, now uses aws elb address for CM URL
Screen Shot 2021-05-11 at 16 32 50

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -124,16 +124,38 @@ public class LoadBalancerConfigService {
         }
     }
 
+    /*
+     * favors public over private
+     */
     public String getLoadBalancerUserFacingFQDN(Long stackId) {
         Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stackId).stream()
-            .filter(lb -> StringUtils.isNotEmpty(lb.getFqdn()))
-            .collect(Collectors.toSet());
-        LoadBalancer loadBalancer = loadBalancers.stream()
-            .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
-            .findAny()
-            .orElseGet(() -> loadBalancers.stream().findAny().orElse(null));
+                .filter(lb -> StringUtils.isNotBlank(lb.getDns()) ||
+                        StringUtils.isNotBlank(lb.getIp()) || StringUtils.isNotBlank(lb.getFqdn()))
+                .collect(Collectors.toSet());
 
-        return loadBalancer == null ? null : loadBalancer.getFqdn();
+        return findPublicLbName(loadBalancers)
+                .orElseGet(() -> findPrivateLbNameOrNull(loadBalancers));
+    }
+
+    private Optional<String> findPublicLbName(Set<LoadBalancer> loadBalancers) {
+        return loadBalancers.stream()
+                .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
+                .findAny().map(this::getBestAddressable);
+    }
+
+    private String findPrivateLbNameOrNull(Set<LoadBalancer> loadBalancers) {
+        return loadBalancers.stream().findAny()
+                .map(this::getBestAddressable).orElse(null);
+    }
+
+    private String getBestAddressable(LoadBalancer lb) {
+        if (StringUtils.isNotBlank(lb.getFqdn())) {
+            return lb.getFqdn();
+        }
+        if (StringUtils.isNotBlank(lb.getDns())) {
+            return lb.getDns();
+        }
+        return lb.getIp();
     }
 
     public Optional<LoadBalancer> selectLoadBalancer(Set<LoadBalancer> loadBalancers, LoadBalancerType preferredType) {
--- a/core/src/test/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigServiceTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigServiceTest.java
@@ -61,6 +61,14 @@ public class LoadBalancerConfigServiceTest extends SubnetTest {
 
     private static final String PRIVATE_DNS = "private.dns";
 
+    private static final String PRIVATE_FQDN = "private.fqdn";
+
+    private static final String PRIVATE_IP = "private.ip";
+
+    private static final String PUBLIC_FQDN = "public.fqdn";
+
+    private static final String PUBLIC_IP = "public.ip";
+
     private static final String PUBLIC_DNS = "public.dns";
 
     @Mock
@@ -469,7 +477,17 @@ public class LoadBalancerConfigServiceTest extends SubnetTest {
         when(loadBalancerPersistenceService.findByStackId(0L)).thenReturn(loadBalancers);
 
         String fqdn = underTest.getLoadBalancerUserFacingFQDN(0L);
-        assertEquals(PUBLIC_DNS, fqdn);
+        assertEquals(PUBLIC_FQDN, fqdn);
+    }
+
+    @Test
+    public void testGetLoadBalancerUserFacingFQDNNoLBs() {
+        Set<LoadBalancer> loadBalancers = Set.of();
+
+        when(loadBalancerPersistenceService.findByStackId(0L)).thenReturn(loadBalancers);
+
+        String fqdn = underTest.getLoadBalancerUserFacingFQDN(0L);
+        assertNull(fqdn);
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -124,16 +125,54 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
+            }
 
-        return loadBalancer == null ? null : loadBalancer.getFqdn();
+        }
+        return getBestAddressable(best);
+
+    }
+
+    private String getBestAddressable(LoadBalancer lb) {
+        Preconditions.checkNotNull(lb);
```
lacikaaa, line 165: how could this happen? I don't think this is a valid check and won't be better then a nullpointer on the next line

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -124,16 +125,54 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
+        if (StringUtils.isNotBlank(lb.getFqdn())) {
+            return lb.getFqdn();
+        }
+        if (StringUtils.isNotBlank(lb.getDns())) {
+            return lb.getDns();
+        }
+        if (StringUtils.isNotBlank(lb.getIp())) {
+            return lb.getIp();
+        }
+        return null;
```
lacikaaa, line 175: how could this happen?

lacikaaa (changes requested):
please refactor your code as it's hard to follow and not really readable
also your unit tests doesn't cover every case
thanks

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -124,16 +125,54 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
-            .orElseGet(() -> loadBalancers.stream().findAny().orElse(null));
+                .filter(lb -> StringUtils.isNotEmpty(lb.getDns()) ||
+                        StringUtils.isNotEmpty(lb.getIp()) || StringUtils.isNotEmpty(lb.getFqdn()))
+                .collect(Collectors.toSet());
+        if (loadBalancers.isEmpty()) {
+            return null;
+        }
+        Iterator<LoadBalancer> iterator = loadBalancers.iterator();
+        LoadBalancer best = iterator.next();
+        while (iterator.hasNext()) {
```
hreeve-cloudera, line 147: I'm not entirely sure what changing the original stream() to this iterator adds to the logic. It looks like what it's doing is iterating over the set of load balancers to get the PUBLIC load balancer. But that's already what the stream() was doing:
```
LoadBalancer loadBalancer = loadBalancers.stream()
    .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
    .findAny()
    .orElseGet(() -> loadBalancers.stream().findAny().orElse(null));
```
That'll look at the set, select the PUBLIC load balancer if one exists in the set, and select the first load balancer in the set if there isn't a PUBLIC one. So I think you can just take the load balancer is selects here and pass that to `getBestAddressable`.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -124,16 +125,54 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
+            LoadBalancer balancer = iterator.next();
+            if (!best.getType().equals(balancer.getType())) {
+                if (best.getType().equals(LoadBalancerType.PRIVATE)) {
+                    best = balancer;
+                }
+                continue;
+            }
+            if (StringUtils.isNotEmpty(balancer.getFqdn()) && StringUtils.isEmpty(best.getFqdn())) {
```
hreeve-cloudera, line 155: This if statement will only be hit if we have two load balancers in the set that are the same type. Functionally, that shouldn't happen. The load balancers in the set will always have different types, because we only create new load balancer objects if there isn't one of the desired type already saved. (This class also has does the initial load balancer creation, and that check is done in the `createLoadBalancerIfNotExists` method.) I understand what you were doing with this clause, but it's not a valid case due to how the code is written.

yg-gd: This was a piece of information I was missing, Is it correct to say there will only ever be at most 1 Public and 1 Private load balancer returned for a given stackId?

hreeve-cloudera: Correct.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
yg-gd: I believe all of these could be null, if the loadBalancer had just been initialized before any load balance metadata had been applied to it.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: Unresolving this because Mihaly has a point. This is only called when `findAny()` find a valid element, so when could it be null? I understand the caution of wanting to check that it's not, but I think a better path would be to make sure we don't call this method will a null argument.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
hreeve-cloudera: @lacikaaa That could theoretically happen if the user is looking in the datalake UI while the stack is still spinning up. When the UI tries to get the user facing FQDN for the CM server, it'll land here. If we land here before the metadata has been collected, we just want to return null as an indication that we don't have the information yet.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
yg-gd: That's a good point! I think it's also probably valid to return null in that case, which can signal to anyone else using the method that they should use a different load balancer.

hreeve-cloudera:
@yg-gd The change looks good for me. When merging things into CB, you should squash all your commits in the PR into a single commit before merging, and when you merge choose the "Rebase and merge" option. We'll still need to wait on @lacikaaa approval since he requested changes, but I'll add mine after you do the squash.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
lacikaaa: I was trying to point out a bit indirect way that this code is not reachable. neither this line nor the first if is reachable. if it is please cover with unit tests, as these are not covered yet. If these can't happen why we have these?
```
        if (StringUtils.isNotBlank(lb.getFqdn())) {
            return lb.getFqdn();
        } else if (StringUtils.isNotBlank(lb.getDns())) {
            return lb.getDns();
        } else {
            return lb.getIp();
        }
```
this should be sufficient if you replace the filters from `notEmpty` to `notBlank`

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -124,16 +124,35 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
-            .filter(lb -> StringUtils.isNotEmpty(lb.getFqdn()))
-            .collect(Collectors.toSet());
-        LoadBalancer loadBalancer = loadBalancers.stream()
-            .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
-            .findAny()
-            .orElseGet(() -> loadBalancers.stream().findAny().orElse(null));
+                .filter(lb -> StringUtils.isNotEmpty(lb.getDns()) ||
+                        StringUtils.isNotEmpty(lb.getIp()) || StringUtils.isNotEmpty(lb.getFqdn()))
+                .collect(Collectors.toSet());
+        return loadBalancers.stream()
```
lacikaaa, line 135: it's much better than previous, I would just move out the 2 methods to make things a bit more nice:
```
    public String getLoadBalancerUserFacingFQDN(Long stackId) {
        Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stackId).stream()
                .filter(lb -> StringUtils.isNotBlank(lb.getDns()) || StringUtils.isNotBlank(lb.getIp()) || StringUtils.isNotBlank(lb.getFqdn()))
                .collect(Collectors.toSet());
        return findPublicLbName(loadBalancers)
                .orElseGet(() -> findPrivateLbNameOrNull(loadBalancers));
    }

File: core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java (hunk 2 above)
Comment (hreeve-cloudera) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java:155]: The only difference between `isNotBlank` and `isNotEmpty` is `isNotBlank` will return false if the string contains only whitespace, while `isNotEmpty` will return true in that case. They both return false if the string is null or empty.

`testGetLoadBalancerUserFacingFQDN` verifies the first if statement can be reached, `testGetLoadBalancerUserFacingFQDNPublicMissingFQDN` verifies the else/if can be reached, `testGetLoadBalancerUserFacingFQDNIPOnlyMissingFQDN` verifies the final else can be reached, and `testGetLoadBalancerUserFacingFQDNNoFqdnSet` verifies that if none of those are hit then we return null at the end of the method.

Can you explain why you think the code is unreachable?

Comment (lacikaaa) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java:155]: thanks, I know the difference. I just wanted to make sure the filter in `getLoadBalancerUserFacingFQDN` is in sync with the conditions here.
first `if` is unreachable because this will never return `null`:
`loadBalancerPersistenceService.findByStackId(stackId)`

the last null is unreachable because of the filter, which filters out all the LB which doesn't have at least one of these field set.

Comment (hreeve-cloudera) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java:155]: Ah, I understand now. Thanks for clarifying. Since we're returning null in to different places, the unit test isn't accurately verifying which null return statement is being hit, and so isn't verifying what it's meant to. Thanks for that catch!

@yg-gd to that end I think things would be cleaner if you only had a single null return statement. It would also be good to have a unit test case where the LB exists but has null fqdn, dns, and ip, and a separate test case for when there are no load balancers in the stack at all. That's a test case I missed the first go-round, but it'd be good to have.

Comment (yg-gd) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java]: I took @lacikaaa's point In the case where all 3 are null, then it's safe to return the IP without checking since it's the same as returning null. I've left the initial null check in case this is used in the future in some other context where it's possible that lb is null.

Comment (lacikaaa) [reply core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java]: my opinion still: you should drop that check. it's a private method called from one other. this is unused unnecessary code. if we would code like this, we could have tons of null check in the beginning of all methods. the best is to avoid nulls as soon as possible.

File: core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java [hunk 6]
```diff
@@ -124,16 +124,41 @@ public String generateLoadBalancerEndpoint(Stack stack) {
...
+                .orElseGet(() -> findPrivateLbName(loadBalancers));
+    }
+
+    private Optional<String> findPublicLbName(Set<LoadBalancer> loadBalancers) {
+        return loadBalancers.stream()
+                .filter(lb -> LoadBalancerType.PUBLIC.equals(lb.getType()))
+                .findAny().map(this::getBestAddressable);
+    }
+
+    private String findPrivateLbName(Set<LoadBalancer> loadBalancers) {
```
lacikaaa, line 146: `OrNull` as it can return null if it's not found

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
yg-gd: A compiler is able to infer that it's not possible for lb to be null at this point and will optimize away this check regardless. What remains is whether or not if it's helpful for humans reading the code to understand that inference.

lacikaaa (approved):

PR #10644, opened 2021-05-12
CB-12497: Add availability status to the FreeIPA list API

jamisonbennett:
Add availability status to the FreeIPA list API.

A unit test was added. This was manually tested with a local
deploymnet of cloudbreak.


christmasferret (approved):
Straightforward change. LGTM!

TamasFulop (approved):
I am not familiar with CB in these details, but changes look good to me.
Our (DFX) goal is to have the `availabilityStatus` field available through the `/v1/freeipa/list` response so we don't need to query `/v1/freeipa` endpoint to check if FreeIPA service is available for a given environment.

PR #10645, opened 2021-05-12
CB-12415 Ability to specify min count for a node group in cluster def…

doktoric:
…inition. Right now in UI only GATEWAY node group has the ability to specify min node count as 1. We need extend this facility to another node group based on cluster definition. The first use-case for this will be in DE HA cluster one can not reduce the master node group or masterx node group to be zero.


core/src/test/java/com/sequenceiq/cloudbreak/service/stack/flow/UpdateNodeCountValidatorTest.java
```diff
@@ -0,0 +1,95 @@
...
+            BadRequestException badRequestException = assertThrows(BadRequestException.class, () -> {
+                underTest.validateScalabilityOfInstanceGroup(stack, instanceGroupAdjustmentV4Request);
+            });
+            Assert.assertTrue(badRequestException.getMessage().contains(errorMessageSegment.get()));
+        } else {
+            assertDoesNotThrow(() -> underTest.validateScalabilityOfInstanceGroup(stack, instanceGroupAdjustmentV4Request));
+        }
+    }
+
+    private static Stream<Arguments> testValidateScalabilityOfInstanceGroupData() {
```
topolyai5, line 74: Can you create some tests with 0, because that is the default value, but I don't understand correctly what will happen if the user use with 0 or null (in the request)

biharitomi (approved):

PR #10655, opened 2021-05-13
CB-12412 Update the cluster status to "available" even on datalake ba…

christmasferret:
…ckup failure

Currently, when data lake backup fails, cluster status is set to BACKUP_FAILED state. In this state, the cluster can not be stopped. There is no need to be in a failure state like BACKUP_FAILED when a backup fails. We should set the cluster state back to "Available" and log the result of the backup operation in the events.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
@@ -230,7 +230,7 @@ public class StackStatusCheckerJob extends StatusCheckerJob {
         clusterOperationService.reportHealthChange(stack.getResourceCrn(), newFailedNodeNames, newHealtyHostNames);
         if (!failedInstances.isEmpty()) {
             clusterService.updateClusterStatusByStackId(stack.getId(), Status.AMBIGUOUS);
-        } else if (statesFromAvailabeAllowed().contains(stack.getCluster().getStatus())) {
+        } else if (statesFromAvailableAllowed().contains(stack.getCluster().getStatus())) {
             clusterService.updateClusterStatusByStackId(stack.getId(), Status.AVAILABLE);
         }
     }
@@ -242,7 +242,7 @@ public class StackStatusCheckerJob extends StatusCheckerJob {
         function.run();
     }
 
-    private EnumSet<Status> statesFromAvailabeAllowed() {
+    private EnumSet<Status> statesFromAvailableAllowed() {
         return EnumSet.of(
                 Status.AMBIGUOUS,
                 Status.STOPPED,
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreStatusService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/dr/BackupRestoreStatusService.java
@@ -44,7 +44,7 @@ public class BackupRestoreStatusService {
     }
 
     public void handleDatabaseBackupFailure(long stackId, String errorReason, DetailedStackStatus detailedStatus) {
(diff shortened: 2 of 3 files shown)


christmasferret:
Now when backup failed. cluster status is BACKUP_FAILED.
Screenshot from 2021-05-12 23-19-01

The environment can not be stopped.
Screenshot from 2021-05-12 19-22-14

With this change, the stack status sync job will change the BACKUP_FAILED status back to AVAILABLE.
Screenshot from 2021-05-12 23-37-56

The environment can be stopped. 
Screenshot from 2021-05-12 23-57-30

Screenshot from 2021-05-13 00-17-01

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
```diff
@@ -246,6 +246,7 @@ private void ifFlowNotRunning(Runnable function) {
         return EnumSet.of(
```
brycederriso, line 246: To make sure I understand what's happening, `updateStates` checks this enum for a set of values that should automatically be switched to `AVAILABLE`, rather than whatever end state the cluster is in.

If so, this looks good to me.

brycederriso, line 246: **suggestion (non-blocking):** This isn't part of your change, but the misspelling of this method name feels insidious. Consider refactoring it. 

```diff
diff --git a/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java b/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
index 0c98b0e98e..6ec2104928 100644
--- a/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
@@ -245 +245 @@ public class StackStatusCheckerJob extends StatusCheckerJob {
-    private EnumSet<Status> statesFromAvailabeAllowed() {
+    private EnumSet<Status> statesFromAvailableAllowed() {
```

brycederriso (approved):
Change looks good to me.

Thanks for writing a thorough PR!

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
christmasferret: Yes, exactly what you said. The sync job sync every 3 minutes to bring the states in that ENUMS to available. We put the backup_failed in it so it won't block further datalake operation.

christmasferret: You are right and have a higher standard than me. HAHAHA . Corrected those now.

christmasferret:
@kkalvagadda1  See if you want to take one more look. Or I will click the merge button.

christmasferret:
After discussion, instead of using status sync 3-minute quartz job, we will bring the state to available right after flow and send notification event.  

I put DO_NOT_MERGE tag and will update this PR.

kkalvagadda1 (changes requested):
Please make as below
1. On backup failure, make the DATA LAKE STATUS as Running.
2. See if you add a event log that backup has failed.

christmasferret:
@kkalvagadda1 

1. On backup failure, make the cluster status as available.
The cluster status was marked as backup_failed. Now change it to available. So datalake can be stopped now. The escalation was due to that cluster status was marked as "backup_failed"
Screenshot from 2021-05-16 19-04-12


2. The is already event send to event history to show the backup failure. 
Screenshot from 2021-05-16 18-41-53
Screenshot from 2021-05-15 21-21-17

kkalvagadda1 (approved):

PR #10659, opened 2021-05-13
CB-12269 Remove CDP_FREEIPA_BACKUP_LOCATION_CONFIG entitlement check

christmasferret:
We have qe-tested freeipa backup change for AWS, AZURE, and GCP. We will remove CDP_FREEIPA_BACKUP_LOCATION_CONFIG entitlement check.


christmasferret:
@jamisonbennett @lacikaaa 

In this PR, I am only removing the entitlement check. I still kept CDP_FREEIPA_BACKUP_LOCATION_CONFIG in entitlement service and all tests in case we want to turn on the entitlement again. I think there is no need to remove CDP_FREEIPA_BACKUP_LOCATION_CONFIG now immediately.

Later, under the same JIRA, I can remove all related code in entitlement service and all tests here and also from UMS.

lacikaaa:
I think you can remove all related code safely. We would just revert the commit if we need this again. No need to keep the unused code around. So please remove those too.

christmasferret:
@lacikaaa 

OK. I removed all related code. This exactly reverted CB-11633 in which I created the entitlement. I've tested locally. Could you take another look? Thanks!

lacikaaa (approved):

PR #10663, opened 2021-05-13
CB-12188 Remove old salt minions before boostrap

orngejaket:
This removes old/dead salt minions that no longer exist on salt masters before bootstrapping new salt master nodes.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
@@ -355,11 +355,21 @@ public class ClusterBootstrapper {
         instanceMetaDataService.saveAll(metaDataSet);
         try {
             List<GatewayConfig> allGatewayConfigs = gatewayConfigService.getAllGatewayConfigs(stack);
+            cleanupOldSaltState(allGatewayConfigs, nodes);
             bootstrapNewNodesOnHost(stack, allGatewayConfigs, nodes, allNodes);
         } catch (CloudbreakOrchestratorCancelledException e) {
             throw new CancellationException(e.getMessage());
         } catch (CloudbreakOrchestratorException e) {
             throw new CloudbreakException(e);
+        }
+    }
+
+    private void cleanupOldSaltState(List<GatewayConfig> allGatewayConfigs, Set<Node> nodes) throws CloudbreakOrchestratorFailedException {
+        List<GatewayConfig> saltMastersToCorrect = allGatewayConfigs.stream()
+                .filter(gc -> nodes.stream().anyMatch(n -> !gc.getPrivateAddress().equals(n.getPrivateIp())))
+                .collect(Collectors.toList());
+        for  (GatewayConfig masterToCorrect: saltMastersToCorrect) {
+            hostOrchestrator.removeDeadSaltMinions(masterToCorrect);
         }
     }
 
--- a/core/src/test/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapperTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapperTest.java
@@ -2,6 +2,8 @@ package com.sequenceiq.cloudbreak.core.bootstrap.service;
 
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -19,7 +21,6 @@ import com.sequenceiq.cloudbreak.cloud.model.Image;
 import com.sequenceiq.cloudbreak.cluster.service.ClusterComponentConfigProvider;
 import com.sequenceiq.cloudbreak.cluster.util.ResourceAttributeUtil;
 import com.sequenceiq.cloudbreak.common.service.HostDiscoveryService;
-import com.sequenceiq.cloudbreak.core.CloudbreakImageNotFoundException;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.host.HostBootstrapApiCheckerTask;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.host.HostClusterAvailabilityCheckerTask;
 import com.sequenceiq.cloudbreak.core.bootstrap.service.host.context.HostBootstrapApiContext;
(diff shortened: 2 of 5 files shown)


core/src/test/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapperTest.java
```diff
@@ -111,6 +108,7 @@ public void shouldUseReachableInstances() throws CloudbreakException, Cloudbreak
         Cluster cluster = new Cluster();
         cluster.setGateway(new Gateway());
         when(stack.getCluster()).thenReturn(cluster);
+        GatewayConfig gatewayConfig = new GatewayConfig("host1", "1.1.1.1", "1.1.1.1", 22, "i-1839", false);
```
hreeve-cloudera, line 111: Was this an incidental change?

orngejaket: No it fixes an NPE.  The gateway config was a mock, the logic i added needed to compare ip's so it being null throws a NPE.

hreeve-cloudera (approved):
LGTM

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -359,6 +360,17 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
         }
     }
 
+    private void cleanupOldSaltState(List<GatewayConfig> allGatewayConfigs, Set<Node> nodes, Set<Node> allNodes) throws CloudbreakOrchestratorFailedException {
+        List<GatewayConfig> gatewayTargets = allGatewayConfigs.stream()
+                .filter(gc -> nodes.stream().anyMatch(n -> gc.getPrivateAddress().equals(n.getPrivateIp())))
```
jamisonbennett, line 365: What happens if the IP address happens to be reused after a repair? E.g. in Azure, I have seen that frequently a new VM that is being repaired will get the same IP address as the old VM had.

orngejaket: So we're looking for the repairing gateway in this line.  The gateway configs already have been updated with the new ip and nodes (nodes to be repaired) will match.  So if it's changed or is the same, we'll select them correctly.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -359,6 +360,17 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
         }
     }
 
+    private void cleanupOldSaltState(List<GatewayConfig> allGatewayConfigs, Set<Node> nodes, Set<Node> allNodes) throws CloudbreakOrchestratorFailedException {
```
sodre90, line 363: allNodes is not used in this method, please remove

sodre90:
How does it work in the case of Cluster Proxy + CCM? If I am correct we register only one node to CP so we are not able to communicate with every gateway node.
Anyway, wouldn't the logic implemented here solve this issue? https://github.com/hortonworks/cloudbreak/blob/fd15d6eb76745dbfccfc3a79cd2c62c1e6038b93/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/poller/join/MinionAcceptor.java#L59

orngejaket:
I'm not sure how this behaves with CCM + Cluster Proxy.  That's the next thing I'm looking at in a different jira.

So the problem is that we're trying to reach old minions and it fails before it gets to the MinionAcceptor.cleanupMinionIds.  I had tested it with the updated code and it still didn't accept it.

lacikaaa:
> So the problem is that we're trying to reach old minions and it fails before it gets to the MinionAcceptor.cleanupMinionIds. I had tested it with the updated code and it still didn't accept it.

Could you point me where it fails? I see you put it right before `bootstrapNewNodesOnHost`. This method would invoke the code with the `MinionAcceptor` and until that I couldn't find where it fails.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -746,7 +746,8 @@ public void tearDown(List<GatewayConfig> allGatewayConfigs, Map<String, String>
         }
     }
 
-    private void removeDeadSaltMinions(GatewayConfig gateway) throws CloudbreakOrchestratorFailedException {
+    @Override
+    public void removeDeadSaltMinions(GatewayConfig gateway) throws CloudbreakOrchestratorFailedException {
         try (SaltConnector saltConnector = saltService.createSaltConnector(gateway)) {
             MinionStatusSaltResponse minionStatusSaltResponse = SaltStates.collectNodeStatus(saltConnector);
```
sodre90, line 752: Please add some more logs. Print minions we will remove because they are dead.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -359,6 +360,17 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
         }
     }
 
+    private void cleanupOldSaltState(List<GatewayConfig> allGatewayConfigs, Set<Node> nodes) throws CloudbreakOrchestratorFailedException {
+        List<GatewayConfig> gatewayTargets = allGatewayConfigs.stream()
+                .filter(gc -> nodes.stream().anyMatch(n -> gc.getPrivateAddress().equals(n.getPrivateIp())))
```
sodre90, line 365: Can we simplify this logic to _.filter(gc -> nodes.stream().anyMatch(n -> gc.getPrivateAddress().equals(n.getPrivateIp())))_ ? Then we can delete these lines: 
`List<GatewayConfig> saltMastersToCorrect = new ArrayList<>(allGatewayConfigs);
 saltMastersToCorrect.removeAll(gatewayTargets);`

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/ClusterBootstrapper.java
```diff
@@ -359,6 +360,17 @@ public void bootstrapNewNodes(Long stackId, Set<String> upscaleCandidateAddresse
         }
     }
 
+    private void cleanupOldSaltState(List<GatewayConfig> allGatewayConfigs, Set<Node> nodes) throws CloudbreakOrchestratorFailedException {
```
sodre90, line 363: please write a unit test for this logic

sodre90 (approved):

PR #10664, opened 2021-05-14
CB-12533 Disable the upgrade for the Medium duty HA template

keyki:
The upgrade is broken for Medium Duty with CCM. Broken means that
the repair phase of the upgrade process will fail. In the repair phase
we're re-creating all the instances with the new cloud image. Until
this bug is not fixed it is not safe to do an upgrade on this cluster
type. In the blueprint there is a field where this can be controlled
however, it's only updated when certain operations are performed like
opening the Shared resources > Cluster templates page or the
Cluster Definitions page on the Data Lake page. There are other operations
that can refresh this field in the database like launching Data Hubs or
the cache expires.


lacikaaa (approved):

lnardai (approved):

tiborpopovics (approved):

biharitomi (approved):
lgtm

akanto (approved):

keyki:
It is merged to 2.42 instead

PR #10677, opened 2021-05-14
CFM-1975 7.2.9 Flow Management fails to launch in mow-dev. Added cond…

vbrodetskyi:
…ition to update working directories only if stack version equals or greater than 7.2.10. Also added unit tests for related classes.


doktoric (approved):

doktoric:
can you please squash the commits?

PR #10679, opened 2021-05-14
ENGESC-5190 Make Hue follow the HDFS failovers using WebHDFS

cegganesh84:
1. This was done upto 7.2.7 via #9628
2. 7.2.8 template must have been in code-review at that time.
3. So the template change did not reflect in the future versions.
4. Without this customers cannot use Hue Oozie editor in DE HA cluster.


sidseth (commented):
This adds HTTPFS as a component to the template. Does CM auto-configure Hue to use HTTPFS?

Adding HTTPFS itself, does it get deployed with security enabled, are there other config parameters which may need to be set?
Would be good to have the HDFS team vet this, and get this on the radar for testing.

cegganesh84:
@sidseth 

> Does CM auto-configure Hue to use HTTPFS?

Yes, it does.

> Adding HTTPFS itself, does it get deployed with security enabled, are there other config parameters which may need to be set?

I do not know the answer, but this was already vetted by HDFS team in ENGESC-5190. This jira simply fixes what was missed in the template copy for runtime releases.

sidseth (approved):

PR #10680, opened 2021-05-15
CB-10402:  Load balancer support on Azure to enable semi-private networks

brycederriso:
Closes [CB-10402](https://jira.cloudera.com/browse/CB-10402).

* Fix arm-template issue.
* Refactor AzureLoadBalancerModelBuilder.

## ARM template issue
I messed up nesting levels of the arm template and had a template that looked like:
```
[
  { "id": "value", "id": "different_value"}
]
```

instead of:
```
[
  {"id": "value"},
  {"id": "different_value"}
]
```
Causing NICs to be set up belonging to only a single backend address pool, rather than both the public and private pools.

## Refactor
The AzureLoadBalancerModelBuilder is just reordering method definitions to match the invocation and including a javadoc comment.

## Testing
Tested by creating an Azure Medium Duty DL using the `--no-use-public-ip` and forcing Cloudbreak to set up an Azure environment with the `PublicEndpointAccessGateway.ENABLED` option.

This should simulate the configuration of Azure semi-private networks.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
@@ -32,6 +32,20 @@ public class AzureLoadBalancerModelBuilder {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


hreeve-cloudera (approved):

biharitomi (approved):

brycederriso:
retest this!

brycederriso:
retest this please

PR #10684, opened 2021-05-17
CDPD-24463 Hive proto logs not cleared up

amitagarwal06:
[CDPD-24463](https://jira.cloudera.com/browse/CDPD-24463)
This sets hive.hook.proto.events.clean.freq=1h
for 7.2.10 and 7.2.11 DE, DE-HA, DE-Spark3 templates, so that logs are purged
every hour and saves storage cost. This will incur few list calls to object stores every hour.

Testing done:
Have created a datahub cluster using a modified DE template that includes above mentioned config change. Datahub cluster creation was successful on AWS and config change was applied.


rbalamohan (approved):

amitagarwal06:
@keyki This is approved too, kindly merge this as well.

PR #10692, opened 2021-05-17
CB-12395: DES SP access is not removed from Key Vault when deleting DES

juhi-09:
Inorder to to remove the Service principal(SP) access for Disk encryption set(DES) from key vault, following parameters are required -
1. key vault name
2. key vault resource group.
3. DES SP id.

With CB-9920(Phase 1), keyVault resource group is same as des resource group.
For phase 2, where desResourceGroup can be different from keyVaultResourceGroup, it will be taken care off as part of CB-11927.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
@@ -36,7 +36,7 @@ import com.sequenceiq.common.api.type.CommonStatus;
 public class AzureEncryptionResources implements EncryptionResources {
     private static final Logger LOGGER = LoggerFactory.getLogger(AzureEncryptionResources.class);
 
-    private static final Pattern ENCRYPTION_KEY_URL_PATTERN = Pattern.compile("https://([^.]+)\\.vault.*");
+    private static final Pattern ENCRYPTION_KEY_URL_VAULT_NAME = Pattern.compile("https://([^.]+)\\.vault.*");
 
     private static final Pattern DISK_ENCRYPTION_SET_NAME = Pattern.compile(".*diskEncryptionSets/([^.]+)");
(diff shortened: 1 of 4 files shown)


lajosrodek (commented):
My 2 cents.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -233,4 +235,21 @@ private void deleteDiskEncryptionSetOnCloud(AzureClient azureClient, String desR
...
+            vaultName = matcher.group(1);
+        } else {
+            throw new IllegalArgumentException(String.format("Failed to deduce vault name from given encryption key url \"%s\"", encryptionKeyUrl));
+        }
+        String description = String.format("access to Key Vault \"%s\" in Resource Group \"%s\" for Service Principal having object ID \"%s\" " +
+                        "associated with Disk Encryption Set \"%s\" in Resource Group \"%s\"", vaultName, desResourceGroupName, desPrincipalObjectId,
+                desName, desResourceGroupName);
+                LOGGER.info("Removing {}.", description);
+                azureClient.removeKeyVaultAccessPolicyFromServicePrincipal(desResourceGroupName, vaultName, desPrincipalObjectId);
+                LOGGER.info("Removed {}.", description);
```
lajosrodek, line 253: I know it is overly paranoid, but I would perhaps wrap the these lines in a `retryService.testWith2SecDelayMax15Times()` call, and inside the lambda add a `try` / `catch` around like for `grantKeyVaultAccessPolicyToDiskEncryptionSetServicePrincipal()` and `deleteDiskEncryptionSetOnCloud()`. The vault name extraction (lines 241-247) can stay outside of the retry logic, as there is no point retrying if the key URL is malformed (which should not happen in practice at this point).

doktoric (approved):

PR #10694, opened 2021-05-17
CB-12607: Allow deletion of data lake even when there are data-hub or experiences in the environment.

kkalvagadda1:
This is a key piece of change to figure out if deleting/re-creating the data lake works.

This change is also needed to test config refresh works with experiences.


lacikaaa:
I don't think we need this merged for testing. You can test in your local dev env. Until it's not decided how will the migration happen we shall not merge code for POC

lacikaaa (changes requested):
see my comment

lnardai:
I'll close this as obosolete.

PR #10697, opened 2021-05-18
CB-12535 cb event API should be authorized

horadla23:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/EventV4Controller.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/EventV4Controller.java
@@ -17,10 +17,12 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Sort;
 import org.springframework.stereotype.Controller;
 
-import com.sequenceiq.authorization.annotation.DisableCheckPermissions;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/EventV4Controller.java
```diff
@@ -67,12 +70,14 @@ private StackView getStackViewIfAvailable(String name) {
...
     @Override
-    public StructuredEventContainer structured(String name) {
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.POWERUSER_ONLY)
+    public StructuredEventContainer structured(String name, @AccountId String accountId) {
         return legacyStructuredEventService.getStructuredEventsForStack(name, workspaceService.getForCurrentUser().getId());
     }
 
     @Override
-    public Response download(String name) {
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.POWERUSER_ONLY)
```
lnardai, line 79: Shouldn't this be allowed for users with access? Or other services have overlays for this?

lnardai: If there is no button on the UI, it's fine for me.

lnardai (approved):
Approved if

PR #10698, opened 2021-05-18
CB-12613 Added temporaryStorage field to InstanceTemplate in DataHub post API

daszabo:
This field will controll if only ephemeral storage should be initialized for VMs

### DIFF
--- a/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/Template.java
+++ b/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/Template.java
@@ -19,6 +19,8 @@ import javax.persistence.UniqueConstraint;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.ResourceStatus;
 import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.common.json.JsonToString;
+import com.sequenceiq.cloudbreak.common.type.TemporaryStorage;
+import com.sequenceiq.cloudbreak.converter.TemporaryStorageConverter;
 import com.sequenceiq.cloudbreak.domain.converter.ResourceStatusConverter;
 import com.sequenceiq.cloudbreak.service.secret.SecretValue;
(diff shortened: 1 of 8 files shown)


horadla23 (approved):
LGTM

horadla23:
one minor thing: in entity, please add a `DefaultEnumConverter` to the new field, like here: https://github.com/hortonworks/cloudbreak/blob/89411c9e1676344c29ca813407f41694aed95fcf/core-model/src/main/java/com/sequenceiq/cloudbreak/domain/Template.java#L49

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/Template.java
```diff
@@ -68,6 +69,8 @@
     @OneToMany(mappedBy = "template", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
     private Set<VolumeTemplate> volumeTemplates;
 
+    private TemporalStorage temporalStorage;
```
horadla23, line 72: sql is missing for this, isn't it?
can this value be null?

horadla23 (changes requested):
i approved this first, then those questions above came into my mind

common/src/main/java/com/sequenceiq/cloudbreak/common/type/TemporalStorage.java
```diff
@@ -0,0 +1,5 @@
+package com.sequenceiq.cloudbreak.common.type;
+
+public enum TemporalStorage {
```
lajosrodek, line 3: Did not you rather mean `temporary`? "Temporal" means "time-dependent" or "related to time". 😄

horadla23 (approved):

common/src/main/java/com/sequenceiq/cloudbreak/common/type/TemporalStorage.java
daszabo: Fair point, thanks :)

lajosrodek (approved):
LGTM, nice work!

PR #10702, opened 2021-05-18
CB-12565: DES deletion and env termination fail if the RG name contains a dot

juhi-09:
CB-12565: DES deletion and env termination fail if the RG name contains a dot

This PR fixes the regex pattern to extract resource group name from resource id, specifically for disk encryption set (DES).
As per Azure documentation, resource group name - Can include alphanumeric, underscore, parentheses, hyphen, period (except at end), and Unicode characters that match the allowed characters. 'RESOURCE_GROUP_NAME' pattern is updated to extract the name correctly.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
@@ -13,6 +13,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.microsoft.azure.management.compute.implementation.DiskEncryptionSetInner;
 import com.sequenceiq.cloudbreak.cloud.EncryptionResources;
 import com.sequenceiq.cloudbreak.cloud.azure.client.AzureClient;
@@ -34,13 +35,15 @@ import com.sequenceiq.common.api.type.CommonStatus;
 
 @Service
 public class AzureEncryptionResources implements EncryptionResources {
+
+    @VisibleForTesting
+    static final Pattern RESOURCE_GROUP_NAME = Pattern.compile(".*resourceGroups/([^/]+)/providers.*");
+
     private static final Logger LOGGER = LoggerFactory.getLogger(AzureEncryptionResources.class);
 
     private static final Pattern ENCRYPTION_KEY_URL_VAULT_NAME = Pattern.compile("https://([^.]+)\\.vault.*");
 
     private static final Pattern DISK_ENCRYPTION_SET_NAME = Pattern.compile(".*diskEncryptionSets/([^.]+)");
-
-    private static final Pattern DISK_ENCRYPTION_SET_RESOURCE_GROUP = Pattern.compile(".*resourceGroups/([^.]+)/providers.*");
 
     @Inject
     private AzureClientService azureClientService;
@@ -135,7 +138,7 @@ public class AzureEncryptionResources implements EncryptionResources {
                     throw new IllegalArgumentException(String.format("Failed to deduce Disk Encryption Set name from given resource id \"%s\"",
                             diskEncryptionSetId));
                 }
-                matcher = DISK_ENCRYPTION_SET_RESOURCE_GROUP.matcher(diskEncryptionSetId);
+                matcher = RESOURCE_GROUP_NAME.matcher(diskEncryptionSetId);
                 if (matcher.matches()) {
                     desResourceGroupName = matcher.group(1);
                 } else {
--- a/cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
+++ b/cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
@@ -7,6 +7,7 @@ import static com.sequenceiq.common.api.type.ResourceType.AZURE_DISK_ENCRYPTION_
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doThrow;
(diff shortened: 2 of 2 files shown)


cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -139,6 +142,40 @@ private void initActionFailedExceptionConversion() {
                 .thenAnswer(invocation -> new Retry.ActionFailedException(new CloudConnectorException(invocation.getArgument(0, Exception.class))));
     }
 
+    private static Object[][] getDesResourceId() {
```
lajosrodek, line 145: As a convention, such methods shall be named so their purpose is clear. If used for a single test method, it makes sense to name them as _testMethod_`DataProvider`. So in this case something like `testDiskEncryptionSetResourceGroupNameExtractionDataProvider` would sound more appropriate.

Note: `DataProvider` is a TestNG terminology; it was the first test library that provided support for parameterized tests, way before JUnit.

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -139,6 +142,40 @@ private void initActionFailedExceptionConversion() {
...
+                        "dummyAlphaParenthesesPeriodResourceGroup---....)()(0910"},
+                {"/subscriptions/dummySubscriptionId/resourceGroups/dummyUnderscroceResourceGroup---___....)()(0910/providers/" +
+                        "Microsoft.Compute/diskEncryptionSets/dummyDes",
+                        "dummyUnderscroceResourceGroup---___....)()(0910"}
+        };
+    }
+
+    @ParameterizedTest
+    @MethodSource("getDesResourceId")
+    public void testDiskEncryptionSetResourceGroupNameExtraction(String desResourceId, String resourceGroupName) {
```
lajosrodek, line 170: - To make it easier to identify the individual test cases of a parameterized test, it is a convention to add some "test case description". This is done with `@ParameterizedTest(name = "{0}")` where the value of the `name` property can contain placeholders referencing method arguments; `{0}` is the first (leftmost) one. See https://github.com/hortonworks/cloudbreak/blob/master/environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/azure/AzureResourceEncryptionParametersTest.java#L79-L81 for an example. Either add a dedicated argument like `testCaseName` there, or simply reference one or more arguments that are used to construct the test case description.
- Again, as a convention, method arguments denoting expected values should be named accordingly. So `resourceGroupName` should be renamed to something like `resourceGroupNameExpected` or `expectedResourceGroupName`, whichever you prefer.

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -139,6 +142,40 @@ private void initActionFailedExceptionConversion() {
                 .thenAnswer(invocation -> new Retry.ActionFailedException(new CloudConnectorException(invocation.getArgument(0, Exception.class))));
     }
 
+    private static Object[][] getDesResourceId() {
+        return new Object[][]{
+                {"/subscriptions/dummySubscriptionId/resourceGroups/dummyAlphaResourceGroup/providers/" +
```
lajosrodek, line 147: Not a big deal but it is a helpful convention to add a comment just before the first `{arguments}` entry listing all the desired test method argument names. So, in this case, something like `// desResourceId resourceGroupName` (or adapted accordingly if adding or renaming arguments).

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -139,6 +142,40 @@ private void initActionFailedExceptionConversion() {
...
+                        "dummyUnderscroceResourceGroup---___....)()(0910"}
+        };
+    }
+
+    @ParameterizedTest
+    @MethodSource("getDesResourceId")
+    public void testDiskEncryptionSetResourceGroupNameExtraction(String desResourceId, String resourceGroupName) {
+        String rgName = "";
+        Matcher matcher = underTest.DISK_ENCRYPTION_SET_RESOURCE_GROUP.matcher(desResourceId);
+        if (matcher.matches()) {
```
lajosrodek, line 173: We always expect the pattern to match, do not we? So I would add an `assertTrue(matcher.matches())` and eliminate the `if` condition entirely. Moreover, the dummy `rgName = ""` initialization is also unnecessary.

lajosrodek (commented):
My 2 cents.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -34,14 +35,16 @@
 
 @Service
 public class AzureEncryptionResources implements EncryptionResources {
+
+    @VisibleForTesting
+    static final Pattern DISK_ENCRYPTION_SET_RESOURCE_GROUP = Pattern.compile(".*resourceGroups/([^/]+)/providers.*");
```
doktoric, line 40: why visible for testing?

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResourcesTest.java
```diff
@@ -139,6 +143,38 @@ private void initActionFailedExceptionConversion() {
...
+                {"testDiskEncryptionSetResourceGroupName", "/subscriptions/dummySubscriptionId/resourceGroups/" +
+                        "dummyAlphaParenthesesPeriodResourceGroup---....)()(0910/providers/Microsoft.Compute/diskEncryptionSets/dummyDes",
+                        "dummyAlphaParenthesesPeriodResourceGroup---....)()(0910"},
+                {"testDiskEncryptionSetResourceGroupName", "/subscriptions/dummySubscriptionId/resourceGroups/" +
+                        "dummyUnderscoreResourceGroup---___....)()(0910/providers/Microsoft.Compute/diskEncryptionSets/dummyDes",
+                        "dummyUnderscoreResourceGroup---___....)()(0910"}
+        };
+    }
+
+    @ParameterizedTest(name = "{0}")
```
doktoric, line 170: the name should show what you are testing with some description. This is just a suggestion

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
juhi-09: The change is to correct the pattern, using this pattern in test will ensure that future changes to pattern would be tested appropriately.

doktoric (approved):

doktoric:
rebase pls

PR #10703, opened 2021-05-18
CB-11868 Create default edge node cluster definitions.

foldik:
We migrated edge node blueprints as defaults. Now the user won't see them under the Create DataHub -> Custom Tab since we only list user managed blueprints there. This change fixes this by creating default cluster templates that the user will see under the Cluster Definitions tab which is what we promised on the first place.

### DIFF
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
@@ -1,0 +1,35 @@
+{
+  "name": "7.2.10 COD Edge Node for AWS",
+  "description": "",
+  "type": "OPERATIONALDATABASE",
+  "featureState": "RELEASED",
+  "cloudPlatform": "AWS",
+  "distroXTemplate": {
+    "cluster": {
+      "blueprintName": "7.2.10 COD Edge Node"
+    },
+    "instanceGroups": [
+      {
+        "name": "leader",
+        "nodeCount": 1,
+        "recoveryMode": "MANUAL",
+        "template": {
+          "attachedVolumes": [
+            {
+              "count": 1,
+              "size": 100,
+              "type": "standard"
+            }
+          ],
+          "aws": {
+            "encryption": {
+              "type": "NONE"
+            }
+          },
+          "instanceType": "m5.2xlarge"
+        },
+        "type": "GATEWAY"
+      }
+    ]
+  }
+}
--- a/core/src/main/resources/defaults/clustertemplates/7.2.10/azure/edgenode.json
+++ b/core/src/main/resources/defaults/clustertemplates/7.2.10/azure/edgenode.json
@@ -1,0 +1,38 @@
+{
+  "name": "7.2.10 COD Edge Node for Azure",
+  "description": "",
+  "type": "OPERATIONALDATABASE",
+  "featureState": "RELEASED",
+  "cloudPlatform": "AZURE",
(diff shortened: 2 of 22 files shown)


lnardai (approved):
Did we check what happen in case of name collision? Hopefully something similar like in case of Cluster templates.

core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "7.2.10 COD Edge Node for AWS",
+  "description": "",
+  "type": "OPERATIONALDATABASE",
+  "featureState": "RELEASED",
```
petersomogyi, line 5: I'm unfamiliar with the `type`'s values in the cluster templates. Does it have any _real_ effect on the created DataHub? These edge node cluster tempaltess are related to the OPERATIONALDATABASE area but are not directly OPERATIONALDATABASE DataHubs.

The "featureState": "RELEASED" is also questionable in this context but as I know the other value is IN_PREVIEW so going with RELEASED is better.

core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
```diff
@@ -0,0 +1,35 @@
...
+              "size": 100,
+              "type": "gp2"
+            }
+          ],
+          "aws": {
+            "encryption": {
+              "type": "NONE"
+            }
+          },
+          "instanceType": "m5.4xlarge"
```
petersomogyi, line 29: The m5.4xlarge has 16 vCPUs and 64 GiB memory but Standard_D8_v3 has half of these (8 vCPU, 32Gib).
Can we rather go with m5.2xlarge for AWS (8 vCPU, 32GiB) to be in sync? I don't think m5.4xlarge is really required and on Edge node creation the customer could still increase the instance size.

foldik:
@lnardai in case of name collision the template content remains the same we don't touch it during migration so existing cluster templates with the same name will remain the same.

core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
foldik: Sure, good point.

core/src/main/resources/defaults/clustertemplates/7.2.10/aws/edgenode.json
foldik: As far as I know OPERATIONALDATABASE is just an enum that we use on the UI, it gives more context to the user.

petersomogyi (approved):

pdarvasi (approved):

pdarvasi:
LGTM, pls wait for @keyki to check, too.

keyki:
How urgent is this to make it part of the 2.43 release?

foldik:
@lnardai can you comment on @keyki's comment if it is urgent or we can live without it for now?

lnardai:
@keyki 
we've agreed with the COD team, that it's not important to block or retest the full release for this, since few customers are using the Edge node solution, and we can provide a CLI based workaround.

But basically without this commit we are breaking the UI based solution provided here:
https://docs.cloudera.com/operational-database/cloud/configure-edge-node/topics/cod-edge-node-overview.html

So I would merge this to 2.43.0 in the hope of piggybacking in another more issues.

PR #10705, opened 2021-05-18
CB-12587 Implement network checker concluson step

lturcsanyi:
Checks:
- cloudera.com
- CCM health check
- any neighbours accessible check

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
@@ -1,0 +1,77 @@
+package com.sequenceiq.cloudbreak.conclusion.step;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.cloudera.thunderhead.telemetry.nodestatus.NodeStatusProto.HealthStatus;
+import com.cloudera.thunderhead.telemetry.nodestatus.NodeStatusProto.NetworkDetails;
+import com.cloudera.thunderhead.telemetry.nodestatus.NodeStatusProto.NodeStatus;
+import com.cloudera.thunderhead.telemetry.nodestatus.NodeStatusProto.NodeStatusReport;
+import com.sequenceiq.cloudbreak.client.RPCResponse;
+import com.sequenceiq.cloudbreak.node.status.NodeStatusService;
+
+@Component
+public class NetworkCheckerConclusionStep extends ConclusionStep {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetworkCheckerConclusionStep.class);
+
+    @Inject
+    private NodeStatusService nodeStatusService;
+
+    @Override
+    public Conclusion check(Long resourceId) {
+        RPCResponse<NodeStatusReport> networkReport;
+        try {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
```diff
@@ -0,0 +1,71 @@
...
+        for (NodeStatus nodeStatus : networkReport.getResult().getNodesList()) {
+            NetworkDetails networkDetails = nodeStatus.getNetworkDetails();
+            String host = nodeStatus.getStatusDetails().getHost();
+            if (networkDetails.getCcmEnabled() && !HealthStatus.OK.equals(networkDetails.getCcmAccessible())) {
+                networkFailures.add("CCM is not accessible from node " + host + ". Please check network settings!");
+                String details = String.format("CCM health status is %s for node %s", networkDetails.getCcmAccessible(), host);
+                networkFailureDetails.add(details);
+                LOGGER.warn(details);
+            }
+            if (!HealthStatus.OK.equals(networkDetails.getClouderaComAccessible())) {
```
oleewere, line 57: note that UNKNOWN is the default status, that does not mean there is a problem with the connection. NOK could mean that so better to check against that.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
```diff
@@ -0,0 +1,71 @@
...
+                networkFailures.add("Cloudera.com is not accessible from node: " + host + ". Please check network settings!");
+                String details = String.format("Cloudera.com accessibility status is %s for node %s", networkDetails.getClouderaComAccessible(), host);
+                networkFailureDetails.add(details);
+                LOGGER.warn(details);
+            }
+            if (!HealthStatus.OK.equals(networkDetails.getAnyNeighboursAccessible())) {
+                networkFailures.add("Node " + host + " cannot reach any neighbour nodes. Please check nodes and network settings!");
+                String details = String.format("Neighbours accessibility status is %s for node %s", networkDetails.getAnyNeighboursAccessible(), host);
```
oleewere, line 65: the background network checks (for nodestatus monitor) neighbours are not checked at all, it's costly (as happens from every host) +  some tools can consider that as an  attack (?) (+ if the network is not set up correctly that could ping nodes that are not inside even part of the cluster, that happened with GCP)

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
lturcsanyi: I will modify this.

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/NetworkCheckerConclusionStep.java
lturcsanyi: Thanks for clarification! What is your suggestion? Skip this neighbour checking or maybe use a different endpoint for that?

oleewere: the response object won't change here ... maybe the way should be which is used to ping other hosts (from every hosts)...at the moment this operation is a network scanning, if we could change that to do the scanning against a list of the exact host (if it stored somehow on the nodes??) or if salt can do similar checks between minions it can be even in a different endpoint

lnardai (approved):

PR #10713, opened 2021-05-20
CB-12614 Validate at cluster creation time, that instance types support instance storage, when cluster was posted with ephemeral volumes as temporary storage

daszabo:
Validate at cluster creation time, that instance types support instance storage, when cluster was posted with ephemeral volumes as temporary storage

https://github.com/hortonworks/cloudbreak/pull/10698 is a prerequisite for this PR.

### DIFF
--- a/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackValidator.java
+++ b/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackValidator.java
@@ -1,20 +1,38 @@
 package com.sequenceiq.cloudbreak.cloud.aws;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
 import javax.inject.Inject;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
 
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.cloudformation.model.DescribeStacksRequest;
+import com.amazonaws.services.ec2.model.DescribeInstanceTypesRequest;
+import com.amazonaws.services.ec2.model.DescribeInstanceTypesResult;
+import com.amazonaws.services.ec2.model.InstanceTypeInfo;
 import com.sequenceiq.cloudbreak.cloud.Validator;
 import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonCloudFormationClient;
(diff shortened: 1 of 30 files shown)


nadamdb (approved):

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackValidator.java
```diff
@@ -29,10 +47,46 @@ public void validate(AuthenticatedContext ac, CloudStack cloudStack) {
...
         String cFStackName = cfStackUtil.getCfStackName(ac);
         try {
+            LOGGER.debug("Checking stack name availability. [{}]", cFStackName);
             cfClient.describeStacks(new DescribeStacksRequest().withStackName(cFStackName));
             throw new CloudConnectorException(String.format("Stack is already exists with the given name: %s", cFStackName));
         } catch (AmazonServiceException ignored) {
+            LOGGER.warn("Exception while checking stack name availability.", ignored);
+        }
 
+        LOGGER.debug("Check instance storage availability on instance types");
```
akanto, line 57: This could be moved to a separate private method.

akanto (approved):
Sadly it feels like you had to touch every test class in the universe...

daszabo:
> Sadly it feels like you had to touch every test class in the universe...

I wanted to introduce a new constructor for InstanceTemplate, however the json deserialization does not support multiple constructors.

cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsStackValidator.java
```diff
@@ -24,15 +42,58 @@
...
         AwsCredentialView credentialView = new AwsCredentialView(ac.getCloudCredential());
         String regionName = ac.getCloudContext().getLocation().getRegion().value();
         AmazonCloudFormationClient cfClient = awsClient.createCloudFormationClient(credentialView, regionName);
         String cFStackName = cfStackUtil.getCfStackName(ac);
         try {
+            LOGGER.debug("Checking stack name availability. [{}]", cFStackName);
             cfClient.describeStacks(new DescribeStacksRequest().withStackName(cFStackName));
             throw new CloudConnectorException(String.format("Stack is already exists with the given name: %s", cFStackName));
         } catch (AmazonServiceException ignored) {
+            LOGGER.warn("Exception while checking stack name availability.", ignored);
```
biharitomi, line 59: can we add a check that the catched exception is notfound or not, if it is we should log only a single line on info that we are good to go with the stack name, otherwise we should log on warn level.

biharitomi (changes requested):
Sry, what I commented is not created by you, but I think we could fix it here as we modifying around that.
On the other hand unit test coverage would be nice to have on the new validation logic.

biharitomi (approved):
LGTM

PR #10711, opened 2021-05-20
Cb 12644 lock medium duty upgrade/repair behind entitlement

orngejaket:
See detailed description in the commit message.

PR #10716, opened 2021-05-20
CB-12665 Deny repairs and downscale for HA Datahubs

foldik:
Needs to be merged after https://github.com/hortonworks/cloudbreak/pull/10711 and altered to disable the validation in case of the account has entitlement for medium duty upgrade or repair.


lnardai (approved):

lnardai (approved):

cegganesh84:
@lnardai @foldik The commit message says deny repairs and downscale. But it seems to me only repair is handled here. Please correct me if I am wrong.

foldik:
> @lnardai @foldik The commit message says deny repairs and downscale. But it seems to me only repair is handled here. Please correct me if I am wrong.

@cegganesh84 repair and upgrade comes here and the validation will catch it if the customer is not entitled to do it anyway.

PR #10723, opened 2021-05-20
CB-12646 Add no_proxy host list to ProxyConfig Endpoint

bergerdenes:
Also propagate its value into User Data of FreeIPA and core hosts.
Placeholders are there for CM and Telemetry changes where the new configuration
value should be propagated.


oleewere (approved):
lgtm

doktoric (approved):

PR #10726, opened 2021-05-20
CB-12372: Create e2e test for distroX upgrade.

tiborpopovics:
This test case is required to test the functionality of the DistroX cluster upgrade.
After a proper environment has created, we need to upgrade the SDX first, then we can proceed with the DistroX upgrade.
When we call the DistroX upgrade endpoint with the proper runtime version the CB should select a new image and upgrade the CM and CDH components for the new runtime version.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
@@ -1,0 +1,128 @@
+package com.sequenceiq.it.cloudbreak.testcase.e2e.distrox;
+
+import static com.sequenceiq.it.cloudbreak.cloud.HostGroupType.COMPUTE;
+import static com.sequenceiq.it.cloudbreak.cloud.HostGroupType.MASTER;
+import static com.sequenceiq.it.cloudbreak.cloud.HostGroupType.WORKER;
+import static com.sequenceiq.it.cloudbreak.context.RunningParameter.key;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.inject.Inject;
+
+import org.testng.annotations.Test;
+
+import com.sequenceiq.it.cloudbreak.client.DistroXTestClient;
+import com.sequenceiq.it.cloudbreak.client.SdxTestClient;
+import com.sequenceiq.it.cloudbreak.cloud.v4.CommonClusterManagerProperties;
+import com.sequenceiq.it.cloudbreak.context.Description;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+import com.sequenceiq.it.cloudbreak.dto.distrox.cluster.DistroXUpgradeTestDto;
+import com.sequenceiq.it.cloudbreak.dto.sdx.SdxTestDto;
+import com.sequenceiq.it.cloudbreak.dto.sdx.SdxUpgradeTestDto;
+import com.sequenceiq.it.cloudbreak.testcase.e2e.AbstractE2ETest;
+import com.sequenceiq.it.cloudbreak.util.CloudFunctionality;
+import com.sequenceiq.it.cloudbreak.util.DistroxUtil;
+import com.sequenceiq.it.cloudbreak.util.InstanceUtil;
+import com.sequenceiq.it.cloudbreak.util.VolumeUtils;
+import com.sequenceiq.it.cloudbreak.util.spot.UseSpotInstances;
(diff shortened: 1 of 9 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
```diff
@@ -0,0 +1,130 @@
...
+    private List<String> expectedVolumeIds = new ArrayList<>();
+
+    @Override
+    protected void setupTest(TestContext testContext) {
+        createDefaultUser(testContext);
+        createDefaultCredential(testContext);
+        initializeDefaultBlueprints(testContext);
+        createEnvironmentWithNetworkAndFreeIpa(testContext);
+
+        String sdxName = resourcePropertyProvider().getName();
```
aszegedi, line 54: From this point I would move test steps to the actual test case.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
```diff
@@ -0,0 +1,130 @@
...
+        distroXName = resourcePropertyProvider().getName();
+        testContext
+                .given(sdxName, SdxTestDto.class)
+                .withCloudStorage()
+                .withRuntimeVersion("7.2.7")
```
aszegedi, line 59: I would use the `application.yml` and a related environment variable on Jenkins to parameterise this runtime version. If the version is going to be changed we do not need an additional commit to update the test as well:
```
  runtimeVersion: 7.2.8
  upgrade:
    currentRuntimeVersion: 7.2.0
    targetRuntimeVersion: 7.2.8
```
at latest [application.yml](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/resources/application.yml#L78-L82)

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
```diff
@@ -0,0 +1,130 @@
...
+                .when(sdxTestClient.create(), key(sdxName))
+                .await(SdxClusterStatusResponse.RUNNING, key(sdxName))
+                .awaitForInstance(getSdxInstancesHealthyState())
+                .validate();
+
+        testContext
+                .given(distroXName, DistroXTestDto.class)
+                .withTemplate("7.2.7 - Data Engineering: Apache Spark, Apache Hive, Apache Oozie")
```
aszegedi, line 67: Same as I mentioned above for the version.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
```diff
@@ -0,0 +1,130 @@
...
+        testContext
+                .given(distroXName, DistroXTestDto.class)
+                .when(distroXTestClient.stop(), key(distroXName))
+                .await(STACK_STOPPED)
+                .validate();
+
+        testContext
+                .given(SdxUpgradeTestDto.class)
+                .withReplaceVms(SdxUpgradeReplaceVms.DISABLED)
+                .withRuntime("7.2.9")
```
aszegedi, line 90: Here we can also use an external parameter.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXUpgradeTests.java
```diff
@@ -0,0 +1,130 @@
...
+    @Test(dataProvider = TEST_CONTEXT)
+    @UseSpotInstances
+    @Description(given = "there is a running Cloudbreak, and an environment with SDX and DistroX cluster in available state",
+            when = "upgrade called on the DistroX cluster", then = "DistroX upgrade should be successful, the cluster should be up and running")
+    public void testDistroXUpgrade(TestContext testContext) {
+        List<String> actualVolumeIds = new ArrayList<>();
+
+        testContext
+                .given(DistroXUpgradeTestDto.class)
+                .withRuntime("7.2.9")
```
aszegedi, line 114: Same external parameter as above.

aszegedi (changes requested):

aszegedi:
Could you add some description to this PR, a brief description about the goal of the change and the implemented solution...etc.

aszegedi (approved):

aszegedi:
Please add a detailed description to this PR before merge.

PR #10727, opened 2021-05-20
CB-12675: Allow force deletion of data lake

kmanamcheri:
Today, if any data hubs are attached, we block deletion of data lakes.
However, there are many use cases where we want to backup, delete,
recreate and restore on to a new datalake (with the same name). We
already have a --force option. Let's use this flag to force delete
a datalake when there are datahubs attached (and are stopped).

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
@@ -8,6 +8,7 @@ import static com.sequenceiq.sdx.api.model.SdxClusterShape.CUSTOM;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -852,8 +852,8 @@ private void checkIfSdxIsDeletable(SdxCluster sdxCluster, boolean forced) {
                 throw ex;
             }
         }
-        if (attachedDistroXClusters != null && !attachedDistroXClusters.isEmpty()) {
-            throw new BadRequestException(String.format("The following Data Hub cluster(s) must be terminated before SDX deletion [%s]",
+        if (!forced && attachedDistroXClusters != null && !attachedDistroXClusters.isEmpty()) {
+            throw new BadRequestException(String.format("The following Data Hub cluster(s) must be terminated before SDX deletion [%s]. Please use force delete to override this check.",
```
hreeve-cloudera, line 856: You mention in the JIRA, "We can check if the datahub is stopped, but we shouldn't block this operation." I think that'd be a good thing to check. Could we add that?

kkalvagadda1: I agree, we should check if the data hubs are stopped even when if it a forcible deletion.

kmanamcheri: Ok i'll add the check. So basically you can force this ONLY if the datahubs are stopped, right? Does that sound like the correct logic?

jamisonbennett: You might have situations where you can't stop datahubs due to a failure. I have a cluster where I can't stop FreeIPA and I have seen a datalake that couldn't be stopped too.

kmanamcheri:
@hreeve-cloudera Added check for stopped datahubs

hreeve-cloudera (approved):

PR #10740, opened 2021-05-21
CB-12325: DES resource may sometimes be left behind during environment delete

juhi-09:
CB-12325: DES resource may sometimes be left behind during environment delete

In few cases, the flow to create DES(Disk encryption set) during environment creation might result in failure after creating DES i.e; the resource is created at Azure cloud but the process itself failed to complete, leading to dangling DES in Azure cloud.
Cases where DES process can fail - Lack of adequate permissions to set DES correctly, environment service failure e.t.c

DES is created only when encryption key url is present in AzureResourceEncryptionParametersDto. Further, encryption key url can only be present in AzureResourceEncryptionParametersDto if entitlement to enable SSE with CMK is enabled for the account. So, Inorder to delete DES, checking for presence of encryption key url is sufficient.
During deletion DES resource is fetched from DB using environmentId and ResourceType.AZURE_DISK_ENCRYPTION_SET - only a single DES exists for any environment.
 If resource is found, deletion is processed further else not.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
@@ -22,7 +22,6 @@ import com.sequenceiq.cloudbreak.cloud.model.Variant;
 import com.sequenceiq.cloudbreak.cloud.model.encryption.CreatedDiskEncryptionSet;
 import com.sequenceiq.cloudbreak.cloud.model.encryption.DiskEncryptionSetCreationRequest;
 import com.sequenceiq.cloudbreak.cloud.model.encryption.DiskEncryptionSetDeletionRequest;
-import com.sequenceiq.common.api.type.CommonStatus;
 import com.sequenceiq.common.api.type.ResourceType;
 import com.sequenceiq.environment.credential.v1.converter.CredentialToCloudCredentialConverter;
 import com.sequenceiq.environment.environment.dto.EnvironmentDto;
@@ -121,9 +120,8 @@ public class EnvironmentEncryptionService {
     }
 
     private List<CloudResource> getResourcesForDeletion(EnvironmentDto environment) {
-        Optional<CloudResource> desCloudResourceOptional = resourceRetriever.findByResourceReferenceAndStatusAndType(
-                environment.getParameters().getAzureParametersDto().getAzureResourceEncryptionParametersDto().getDiskEncryptionSetId(),
-                CommonStatus.CREATED, ResourceType.AZURE_DISK_ENCRYPTION_SET);
+        Optional<CloudResource> desCloudResourceOptional = resourceRetriever.findByEnvironmentIdAndType(environment.getId(),
+                ResourceType.AZURE_DISK_ENCRYPTION_SET);
         return desCloudResourceOptional.map(List::of).orElse(List.of());
     }
 
--- a/environment/src/main/java/com/sequenceiq/environment/resourcepersister/CloudResourceRetrieverService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/resourcepersister/CloudResourceRetrieverService.java
@@ -32,4 +32,12 @@ public class CloudResourceRetrieverService implements ResourceRetriever {
         return optionalResource
                 .map(resource -> cloudResourceConverter.convert(resource));
     }
+
+    public Optional<CloudResource> findByEnvironmentIdAndType(Long environmentId, ResourceType resourceType) {
+        Optional<Resource> optionalResource = resourceService.findByEnvironmentIdAndType(environmentId, resourceType);
+        LOGGER.debug("Resource retrieved by environmentId: {} and type: {}. Is present: {}", environmentId, resourceType, optionalResource.isPresent());
+        return optionalResource
+                .map(resource -> cloudResourceConverter.convert(resource));
+    }
+
 }
--- a/environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
+++ b/environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
@@ -67,6 +67,8 @@ class EnvironmentEncryptionServiceTest {
 
     private static final String DISK_ENCRYPTION_SET_ID = "dummyDesId";
 
+    private static final String DISK_ENCRYPTION_SET_NAME = "dummyDesName";
+
     private static final long ENVIRONMENT_ID = 1L;
 
     private static final String KEY_URL = "dummyKeyUrl";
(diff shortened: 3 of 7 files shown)


doktoric:
please resolve the conflicts

environment/src/main/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionService.java
```diff
@@ -120,10 +121,9 @@ private CloudContext getCloudContext(EnvironmentDto environment) {
                 .build();
     }
 
-    private List<CloudResource> getResourcesForDeletion(EnvironmentDto environment) {
-        Optional<CloudResource> desCloudResourceOptional = resourceRetriever.findByResourceReferenceAndStatusAndType(
-                environment.getParameters().getAzureParametersDto().getAzureResourceEncryptionParametersDto().getDiskEncryptionSetId(),
-                CommonStatus.CREATED, ResourceType.AZURE_DISK_ENCRYPTION_SET);
+    private List<CloudResource> getResourcesForDeletion(EnvironmentDto environment, String diskEncryptionSetName) {
+        Optional<CloudResource> desCloudResourceOptional = resourceRetriever.findByEnvironmentIdAndNameAndType(
+                environment.getId(), diskEncryptionSetName, ResourceType.AZURE_DISK_ENCRYPTION_SET);
```
lajosrodek, line 126: A different lookup strategy is preferable. Surely it is the most convenient to find the resource by its name, but the naming strategy is an implementation detail in `cloud-azure` that we do not want to depend on.

Therefore, I suggest the following:
* Remove the `diskEncryptionSetName` argument.
* Add a new method to `CloudResourceRetrieverService`: `findByEnvironmentIdAndType(Long environmentId, ResourceType type)`. You shall, in turn, also add `com.sequenceiq.environment.resourcepersister.ResourceService.findByEnvironmentIdAndType()` and `com.sequenceiq.environment.resourcepersister.ResourceRepository.findByEnvironmentIdAndType()` (with the  appropriate `@Query`).
* Look up `desCloudResourceOptional` using `resourceRetriever.findByEnvironmentIdAndType()`. The DES will be either associated with an environment (the global one created here), or potentially later with a cluster (FreeIPA / DL / DH) if we decide to support that. In either case, there will be at most one DES for the respective parent resourceCrn.

environment/src/main/java/com/sequenceiq/environment/resourcepersister/CloudResourceRetrieverService.java
```diff
@@ -32,4 +32,13 @@
...
     }
+
+    @Override
+    public Optional<CloudResource> findByEnvironmentIdAndNameAndType(Long environmentId, String resourceName, ResourceType resourceType) {
+        Optional<Resource> optionalResource = resourceService.findByEnvironmentIdAndNameAndType(environmentId, resourceName, resourceType);
+        LOGGER.debug("Resource retrieved by optionalResource name: {}, type: {} and environmentId: {}. Is present: {}", resourceName, resourceType,
+                environmentId, optionalResource.isPresent());
+        return optionalResource
+                .map(resource -> cloudResourceConverter.convert(resource));
+    }
```
lajosrodek, line 43: Please revert & add another method `findByEnvironmentIdAndType(Long environmentId, ResourceType type)` without ` @Override`.

core/src/main/java/com/sequenceiq/cloudbreak/service/eventbus/CloudResourceRetrieverService.java
```diff
@@ -35,4 +35,9 @@
         return optionalResource
                 .map(resource -> cloudResourceConverter.convert(resource));
     }
+
+    @Override
+    public Optional<CloudResource> findByEnvironmentIdAndNameAndType(Long environmentId, String resourceName, ResourceType resourceType) {
+        return Optional.empty();
+    }
```
lajosrodek, line 42: To be reverted.

freeipa/src/main/java/com/sequenceiq/freeipa/service/cloud/CloudResourceRetrieverService.java
```diff
@@ -37,4 +37,9 @@
         return optionalResource
                 .map(resource -> conversionService.convert(resource, CloudResource.class));
     }
+
+    @Override
+    public Optional<CloudResource> findByEnvironmentIdAndNameAndType(Long environmentId, String resourceName, ResourceType resourceType) {
+        return Optional.empty();
+    }
```
lajosrodek, line 44: To be reverted.

redbeams/src/main/java/com/sequenceiq/redbeams/service/cloud/CloudResourceRetrieverService.java
```diff
@@ -16,4 +16,9 @@
     public Optional<CloudResource> findByResourceReferenceAndStatusAndType(String resourceReference, CommonStatus status, ResourceType resourceType) {
         return Optional.empty();
     }
+
+    @Override
+    public Optional<CloudResource> findByEnvironmentIdAndNameAndType(Long environmentId, String resourceName, ResourceType resourceType) {
+        return Optional.empty();
+    }
```
lajosrodek, line 23: To be reverted.

lajosrodek (changes requested):
Some asks. They are concerned with the general strategy for looking up the DES `CloudResource`.

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -204,19 +206,22 @@ void testCreateEncryptionResourcesDeletionRequestShouldReturnWithANewDeletionReq
...
         CloudResource desCloudResource = CloudResource.builder()
-                .name("Des")
+                .name(DISK_ENCRYPTION_SET_NAME)
                 .type(ResourceType.AZURE_DISK_ENCRYPTION_SET)
                 .reference(DISK_ENCRYPTION_SET_ID)
                 .status(CommonStatus.CREATED)
                 .build();
-        when(resourceRetriever.findByResourceReferenceAndStatusAndType(any(), any(), any())).thenReturn(Optional.ofNullable(desCloudResource));
+        when(resourceRetriever.findByEnvironmentIdAndType(ENVIRONMENT_ID, ResourceType.AZURE_DISK_ENCRYPTION_SET))
+                .thenReturn(Optional.ofNullable(desCloudResource));
```
lajosrodek, line 224: Nit: `desCloudResource` cannot be `null` here, so `Optional.of()` would be more appropriate.

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
```diff
@@ -269,20 +274,23 @@ void testDeleteEncryptionResourcesShouldCallDeleteDiskEncryptionSetWhenCloudPlat
...
```
lajosrodek, line 293: Like above, `Optional.of()` would be more appropriate.

lajosrodek (approved):
Thanks for the fixes, LGTM. The two minor comments are not critical, up to you fixing them now, later or ignoring them altogether. 😄

environment/src/test/java/com/sequenceiq/environment/environment/encryption/EnvironmentEncryptionServiceTest.java
juhi-09: Updated

PR #10741, opened 2021-05-21
CB-12525 Merge requested telemetry events into status changed events

nadamdb:


akanto (changes requested):
Can we still write the "requested" tables for the next 1-2 releases, please? The transition will be significantly simpler.

akanto (approved):

PR #10742, opened 2021-05-21
CB-12710: Fix closing FormDataMultiPart to not close it early

jamisonbennett:
FormDataMultiPart was being closed multiple times while it was still
being used. This was fixed so that it is only closed when it is done
being used.

Also some missing toString() methods were added.

This was tested manually with a local deployment of cloudbreak.


lacikaaa (changes requested):
could you modify the method definition a bit, so it would return with the modified MultiPart (and use it where called)? This way it would be obvious what does the method change

PR #10747, opened 2021-05-25
CRB-1587 Release a first version of Oozie HA from 7.2.11

cegganesh84:
1. In public cloud all the requests go through Knox, so we do not need a external load-balancer for external clients.
3. This is for Hue and Yarn to access the Oozie in HA fashion.
4. For AWS and Azure use an internal load-balancer.
5. For GCP we do not have any load-balancers, so as a workaround use the current Oozie server address for url and callbacks.
6. To make Hue to Oozie interaction work via load-balancer, Hue is moved to a new hostgroup. Since there is only one instance for GCP for manager node, Hue is also added to masterx hostgroup.
7. CM is given ability to add hosts.
8. LB reverse IP is added.
9. Knox proxy hosts are configured even if we failed to acquire Let's Encrypt certificate. Rationale is, the common workaround for Let's Encrypt outages is to let the cluster be created and users are asked to click the renew certificate button.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
@@ -7,6 +7,7 @@ import com.sequenceiq.cloudbreak.cloud.azure.loadbalancer.AzureLoadBalancingRule
 import com.sequenceiq.cloudbreak.cloud.model.CloudLoadBalancer;
 import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
 import com.sequenceiq.cloudbreak.cloud.model.Group;
+import com.sequenceiq.cloudbreak.cloud.model.TargetGroupPortPair;
 import com.sequenceiq.common.api.type.LoadBalancerType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -73,10 +74,13 @@ public class AzureLoadBalancerModelBuilder {
 
     private List<AzureLoadBalancingRule> collectLoadBalancingRules(CloudLoadBalancer cloudLoadBalancer) {
         return cloudLoadBalancer.getPortToTargetGroupMapping()
-                .keySet()
-                .stream()
-                .map(AzureLoadBalancingRule::new)
-                .collect(toList());
+            .entrySet().stream()
+            .map(entry -> {
+                TargetGroupPortPair pair = entry.getKey();
+                Group firstGroup = entry.getValue().stream().findFirst().orElse(null);
+                return new AzureLoadBalancingRule(pair, firstGroup);
+            })
+            .collect(toList());
     }
 
     private AzureLoadBalancer buildAzureLb(LoadBalancerType type, Set<String> instanceGroupNames, List<AzureLoadBalancingRule> rules, String stackName) {
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/loadbalancer/AzureLoadBalancingRule.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/loadbalancer/AzureLoadBalancingRule.java
@@ -1,5 +1,8 @@
 package com.sequenceiq.cloudbreak.cloud.azure.loadbalancer;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.sequenceiq.cloudbreak.cloud.model.Group;
 import com.sequenceiq.cloudbreak.cloud.model.TargetGroupPortPair;
 
 public final class AzureLoadBalancingRule {
@@ -11,11 +14,14 @@ public final class AzureLoadBalancingRule {
 
     private final AzureLoadBalancerProbe probe;
 
-    public AzureLoadBalancingRule(TargetGroupPortPair portPair) {
+    private final String groupName;
+
+    public AzureLoadBalancingRule(TargetGroupPortPair portPair, Group group) {
         this.backendPort = portPair.getTrafficPort();
         this.frontendPort = portPair.getTrafficPort();
         this.name = defaultNameFromPort(portPair.getTrafficPort());
         this.probe = new AzureLoadBalancerProbe(portPair.getHealthCheckPort());
+        this.groupName = checkNotNull(group, "Group must be provided.").getName();
     }
 
     private String defaultNameFromPort(int port) {
@@ -38,6 +44,10 @@ public final class AzureLoadBalancingRule {
         return probe;
     }
 
+    public String getGroupName() {
+        return groupName;
+    }
+
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
@@ -46,6 +56,7 @@ public final class AzureLoadBalancingRule {
         sb.append("    backendPort: ").append(backendPort).append("\n");
         sb.append("    frontendPort: ").append(frontendPort).append("\n");
         sb.append("    probe: ").append(toIndentedString(probe)).append("\n");
+        sb.append("    groupName: ").append(toIndentedString(groupName)).append("\n");
         sb.append("}");
 
         return sb.toString();
--- a/cloud-azure/src/main/resources/templates/arm-v2.ftl
+++ b/cloud-azure/src/main/resources/templates/arm-v2.ftl
@@ -201,7 +201,7 @@
                  <#assign createAndNoSecGroup = !isUpscale && (! securityGroups[instance.groupName]?? || ! securityGroups[instance.groupName]?has_content)>
                  <#if !noPublicIp>
                  {
-                   "apiVersion": "2015-05-01-preview",
+                   "apiVersion": "2017-08-01",
                    "type": "Microsoft.Network/publicIPAddresses",
                    "name": "[concat(parameters('publicIPNamePrefix'), '${instance.instanceId}')]",
                    "location": "[parameters('region')]",
@@ -212,8 +212,14 @@
                         </#list>
                      </#if>
                      },
+                   <#if instance.availabilitySetName?? && instance.availabilitySetName?has_content>
+                     "sku": {
+                         "name": "Standard",
+                         "tier": "Regional"
+                     },
+                   </#if>
                    "properties": {
-                       <#if instanceGroup == "GATEWAY">
+                       <#if instanceGroup == "GATEWAY" || (instance.availabilitySetName?? && instance.availabilitySetName?has_content)>
                        "publicIPAllocationMethod": "Static"
                        <#else>
                        "publicIPAllocationMethod": "Dynamic"
@@ -303,7 +309,7 @@
                                                only a single LB pool
                                            -->
                                        <#list loadBalancerMapping[instance.groupName] as loadBalancer>
-                                       { "id": "[resourceId('Microsoft.Network/loadBalancers/backendAddressPools', '${loadBalancer.name}', 'address-pool')]" }
+                                       { "id": "[resourceId('Microsoft.Network/loadBalancers/backendAddressPools', '${loadBalancer.name}', '${instance.groupName}-pool')]" }
                                        <#sep>, </#sep>
                                        </#list>
                                    ]
@@ -445,11 +451,13 @@
                   },
                   "properties": {
                     "backendAddressPools": [
-                      {
-<#--                todo: when additional lbs and pools are added we'll have to handle associating the appropriate address pool names-->
-                        "name": "address-pool",
-                        "properties": {}
-                      }
+                       <#list loadBalancer.instanceGroupNames as groupName>
+                       {
+                         "name": "${groupName}-pool",
+                         "properties": {
+                         }
+                       }<#sep>,</#sep>
+                       </#list>
                     ],
                     "frontendIPConfigurations": [
                       <#if loadBalancer.type == "PUBLIC">
@@ -488,7 +496,7 @@
                                 "name": "${rule.name}",
                                 "properties": {
                                     "backendAddressPool": {
-                                        "id": "[resourceId('Microsoft.Network/loadBalancers/backendAddressPools', '${loadBalancer.name}', 'address-pool')]"
+                                        "id": "[resourceId('Microsoft.Network/loadBalancers/backendAddressPools', '${loadBalancer.name}', '${rule.groupName}-pool')]"
                                     },
                                     "backendPort": ${rule.backendPort},
                                     "enableFloatingIP": false,
@@ -522,7 +530,7 @@
                     ]
                   },
                   "sku": {
-                    "name": "Basic"
+                    "name": "Standard"
                   }
                 }
                 <#if loadBalancer.type == "PUBLIC">
@@ -532,7 +540,7 @@
                     "name": "${loadBalancer.name}-publicIp",
                     "location": "[parameters('region')]",
                     "sku": {
-                        "name": "Basic"
+                        "name": "Standard"
                     },
                     "properties": {
                         "publicIPAddressVersion": "IPv4",
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -3,6 +3,7 @@ package com.sequenceiq.cloudbreak.core.bootstrap.service.host;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_0_2;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_2_0;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_2_1;
+import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.CLOUDERAMANAGER_VERSION_7_4_3;
 import static com.sequenceiq.cloudbreak.cmtemplate.CMRepositoryVersionUtil.isVersionNewerOrEqualThanLimited;
 import static com.sequenceiq.cloudbreak.core.bootstrap.service.ClusterDeletionBasedExitCriteriaModel.clusterDeletionBasedModel;
 import static java.util.Collections.singletonMap;
@@ -73,6 +74,7 @@ import com.sequenceiq.cloudbreak.domain.stack.cluster.gateway.GatewayTopology;
 import com.sequenceiq.cloudbreak.domain.stack.cluster.host.HostGroup;
 import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceGroup;
 import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceMetaData;
+import com.sequenceiq.cloudbreak.domain.stack.loadbalancer.LoadBalancer;
 import com.sequenceiq.cloudbreak.dto.KerberosConfig;
 import com.sequenceiq.cloudbreak.dto.LdapView;
 import com.sequenceiq.cloudbreak.kerberos.KerberosConfigService;
@@ -102,6 +104,7 @@ import com.sequenceiq.cloudbreak.service.rdsconfig.RdsConfigService;
 import com.sequenceiq.cloudbreak.service.sharedservice.DatalakeService;
 import com.sequenceiq.cloudbreak.service.stack.InstanceGroupService;
 import com.sequenceiq.cloudbreak.service.stack.InstanceMetaDataService;
+import com.sequenceiq.cloudbreak.service.stack.LoadBalancerPersistenceService;
 import com.sequenceiq.cloudbreak.service.stack.StackService;
 import com.sequenceiq.cloudbreak.service.stack.flow.MountDisks;
 import com.sequenceiq.cloudbreak.template.kerberos.KerberosDetailService;
@@ -111,6 +114,7 @@ import com.sequenceiq.cloudbreak.util.NodesUnreachableException;
 import com.sequenceiq.cloudbreak.util.StackUtil;
 import com.sequenceiq.common.api.telemetry.model.DataBusCredential;
 import com.sequenceiq.common.api.telemetry.model.Telemetry;
+import com.sequenceiq.common.api.type.LoadBalancerType;
 
 @Component
 public class ClusterHostServiceRunner {
@@ -144,6 +148,9 @@ public class ClusterHostServiceRunner {
 
     @Inject
     private InstanceMetaDataService instanceMetaDataService;
+
+    @Inject
+    private LoadBalancerPersistenceService loadBalancerPersistenceService;
 
     @Inject
     private ClusterComponentConfigProvider clusterComponentConfigProvider;
@@ -341,7 +348,7 @@ public class ClusterHostServiceRunner {
         postgresConfigService.decorateServicePillarWithPostgresIfNeeded(servicePillar, stack, cluster);
 
         if (blueprintService.isClouderaManagerTemplate(cluster.getBlueprint())) {
-            addClouderaManagerConfig(stack, cluster, servicePillar, clouderaManagerRepo);
+            addClouderaManagerConfig(stack, cluster, servicePillar, clouderaManagerRepo, primaryGatewayConfig);
         }
         ldapView.ifPresent(ldap -> saveLdapPillar(ldap, servicePillar));
 
@@ -381,7 +388,9 @@ public class ClusterHostServiceRunner {
                 && !kerberosDetailService.isClusterManagerManagedKrb5Config(kerberosConfig);
     }
 
-    private void addClouderaManagerConfig(Stack stack, Cluster cluster, Map<String, SaltPillarProperties> servicePillar, ClouderaManagerRepo clouderaManagerRepo)
+    private void addClouderaManagerConfig(Stack stack, Cluster cluster,
+        Map<String, SaltPillarProperties> servicePillar, ClouderaManagerRepo clouderaManagerRepo,
+        GatewayConfig primaryGatewayConfig)
             throws CloudbreakOrchestratorFailedException {
         Telemetry telemetry = componentConfigProviderService.getTelemetry(stack.getId());
         DataBusCredential dataBusCredential = null;
@@ -398,7 +407,7 @@ public class ClusterHostServiceRunner {
         Optional<String> licenseOpt = decoratePillarWithClouderaManagerLicense(stack.getId(), servicePillar);
         decoratePillarWithClouderaManagerRepo(clouderaManagerRepo, servicePillar, licenseOpt);
         decoratePillarWithClouderaManagerDatabase(cluster, servicePillar);
-        decoratePillarWithClouderaManagerCommunicationSettings(cluster, servicePillar);
+        decoratePillarWithClouderaManagerCommunicationSettings(stack, cluster, servicePillar, primaryGatewayConfig, clouderaManagerRepo);
         decoratePillarWithClouderaManagerAutoTls(cluster, servicePillar);
         csdParcelDecorator.decoratePillarWithCsdParcels(stack, servicePillar);
         decoratePillarWithClouderaManagerSettings(servicePillar, clouderaManagerRepo, stack);
@@ -460,14 +469,39 @@ public class ClusterHostServiceRunner {
                 new SaltPillarProperties("/cloudera-manager/database.sls", singletonMap("cloudera-manager", singletonMap("database", rdsView))));
     }
 
-    private void decoratePillarWithClouderaManagerCommunicationSettings(Cluster cluster, Map<String, SaltPillarProperties> servicePillar) {
+    private void decoratePillarWithClouderaManagerCommunicationSettings(Stack stack,
+        Cluster cluster,
+        Map<String, SaltPillarProperties> servicePillar,
+        GatewayConfig primaryGatewayConfig,
+        ClouderaManagerRepo clouderaManagerRepo) {
         Boolean autoTls = cluster.getAutoTlsEnabled();
         Map<String, Object> communication = new HashMap<>();
+        if (isVersionNewerOrEqualThanLimited(clouderaManagerRepo.getVersion(), CLOUDERAMANAGER_VERSION_7_4_3)) {
+            Set<LoadBalancer> loadBalancers = loadBalancerPersistenceService.findByStackId(stack.getId());
+            if (!loadBalancers.isEmpty()) {
+                Optional<LoadBalancer> loadBalancer = loadBalancerConfigService.selectLoadBalancer(loadBalancers, LoadBalancerType.PUBLIC);
+                Optional<String> san = loadBalancer.flatMap(this::getBestSANForLB);
+                if (san.isPresent()) {
+                    communication.put("internal_loadbalancer_san", san.get());
+                }
+            }
+        }
         communication.put("port", autoTls ? CM_HTTPS_PORT : CM_HTTP_PORT);
         communication.put("protocol", autoTls ? "https" : "http");
         communication.put("autotls_enabled", autoTls);
-        servicePillar.put("cloudera-manager-communication", new SaltPillarProperties("/cloudera-manager/communication.sls",
+        servicePillar.put("cloudera-manager-communication",
+            new SaltPillarProperties("/cloudera-manager/communication.sls",
                 singletonMap("cloudera-manager", singletonMap("communication", communication))));
+    }
+
+    private Optional<String> getBestSANForLB(LoadBalancer lb) {
+        if (isNotBlank(lb.getFqdn())) {
+            return Optional.of("DNS:" + lb.getFqdn());
+        }
+        if (isNotBlank(lb.getDns())) {
+            return Optional.of("DNS:" + lb.getDns());
+        }
+        return Optional.of("IP:" + lb.getIp());
     }
 
     private void decoratePillarWithClouderaManagerAutoTls(Cluster cluster, Map<String, SaltPillarProperties> servicePillar) {
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/CmServiceKeytabRequestFactory.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/CmServiceKeytabRequestFactory.java
@@ -27,7 +27,7 @@ public class CmServiceKeytabRequestFactory {
         request.setDoNotRecreateKeytab(Boolean.TRUE);
         RoleRequest roleRequest = new RoleRequest();
         roleRequest.setRoleName("hadoopadminrole-" + stack.getName());
-        roleRequest.setPrivileges(Set.of("Service Administrators", "Certificate Administrators", "CA Administrator"));
+        roleRequest.setPrivileges(Set.of("Service Administrators", "Certificate Administrators", "Host Administrators", "CA Administrator"));
         request.setRoleRequest(roleRequest);
         return request;
     }
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
@@ -1,5 +1,6 @@
 package com.sequenceiq.cloudbreak.service;
 
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.oozie.OozieHAConfigProvider.OOZIE_HTTPS_PORT;
 import static com.sequenceiq.cloudbreak.common.mappable.CloudPlatform.AWS;
 import static com.sequenceiq.cloudbreak.common.mappable.CloudPlatform.AZURE;
 import static java.util.Map.entry;
@@ -31,6 +32,7 @@ import com.sequenceiq.cloudbreak.cloud.model.TargetGroupPortPair;
 import com.sequenceiq.cloudbreak.cloud.model.instance.AzureInstanceGroupParameters;
 import com.sequenceiq.cloudbreak.cmtemplate.CmTemplateProcessor;
 import com.sequenceiq.cloudbreak.cmtemplate.configproviders.knox.KnoxRoles;
+import com.sequenceiq.cloudbreak.cmtemplate.configproviders.oozie.OozieRoles;
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 import com.sequenceiq.cloudbreak.common.json.Json;
 import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
@@ -107,6 +109,20 @@ public class LoadBalancerConfigService {
         return groupNames;
     }
 
+    public Set<String> getOozieGroups(Stack stack) {
+        LOGGER.debug("Fetching list of instance groups with Oozie installed");
+        Cluster cluster = stack.getCluster();
+        if (cluster != null) {
+            CmTemplateProcessor cmTemplateProcessor = new CmTemplateProcessor(cluster.getBlueprint().getBlueprintText());
+            Set<String> groupNames = cmTemplateProcessor.getHostGroupsWithComponent(OozieRoles.OOZIE_SERVER);
+            LOGGER.info("Oozie instance groups are {}", groupNames);
+            return groupNames;
+        } else {
+            LOGGER.info("No Oozie instance groups found");
+            return Set.of();
+        }
+    }
+
     public String generateLoadBalancerEndpoint(Stack stack) {
         StringBuilder name = new StringBuilder()
             .append(stack.getName())
@@ -119,6 +135,8 @@ public class LoadBalancerConfigService {
         switch (targetGroup.getType()) {
             case KNOX:
                 return Set.of(new TargetGroupPortPair(Integer.parseInt(httpsPort), Integer.parseInt(knoxServicePort)));
+            case OOZIE:
+                return Set.of(new TargetGroupPortPair(Integer.parseInt(OOZIE_HTTPS_PORT), Integer.parseInt(OOZIE_HTTPS_PORT)));
             default:
                 return null;
         }
@@ -194,17 +212,21 @@ public class LoadBalancerConfigService {
     }
 
     /**
-     * Adds availability sets to instance groups associated with Knox target groups.
+     * Adds availability sets to instance groups associated with Knox and Oozie target groups.
      * This method updates the JSON parameters associated with certain instance groups.
      *
      * It's only necessary to add availability sets to the Azure deployment while we
-     * use the {@code basic} Azure Load Balancer SKU.
+     * use the {@code basic} Azure Load Balancer SKU. We use {@code standard} Load Balancer
+     * we still use availability set to indicate that we need a {@code standard} IP addresses
+     * for the instances in those availability sets.
      *
      * @param availabilitySetPrefix A string prefix. Should be a stack name normally.
      * @param loadBalancers the list of load balancers to look up target groups from.
      */
     public void configureLoadBalancerAvailabilitySets(String availabilitySetPrefix, Set<LoadBalancer> loadBalancers) {
         getKnoxInstanceGroups(loadBalancers)
+                .forEach(instanceGroup -> attachAvailabilitySetParameters(availabilitySetPrefix, instanceGroup));
+        getOozieInstanceGroups(loadBalancers)
                 .forEach(instanceGroup -> attachAvailabilitySetParameters(availabilitySetPrefix, instanceGroup));
     }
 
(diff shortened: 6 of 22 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -389,6 +434,31 @@ private boolean isEndpointGatewayEnabled(EnvironmentNetworkResponse network) {
         return Optional.ofNullable(knoxTargetGroup);
     }
 
+    private Optional<TargetGroup> setupOozieHATargetGroup(Stack stack, boolean dryRun) {
+        TargetGroup oozieTargetGroup = null;
+        Set<String> oozieGroupNames = getOozieGroups(stack);
+        InstanceGroup oozieInstanceGroup = stack.getInstanceGroups().stream()
+            .filter(ig -> oozieGroupNames.contains(ig.getGroupName()))
+            .filter(ig -> ig.getNodeCount() > 1)
```
lacikaaa, line 442: so this can't change during an upscale, and LB only needed during HA, is this right?

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -43,10 +43,13 @@ public void registerLoadBalancerDomainWithFreeIPA(Stack stack) {
             try {
                 if (StringUtils.isNotEmpty(loadBalancer.getDns())) {
                     sendAddDnsCnameRecordRequest(stack, loadBalancer);
-                } else if (StringUtils.isNotEmpty(loadBalancer.getIp())) {
+                } else {
+                    LOGGER.error("Unable to find DNS for load balancer. Load balancer address will not be registered as CNAME with FreeIPA.");
```
lacikaaa, line 47: instead of creating a new `else` branch here you could add logs to every existing one and it should be obvious from the logs what happened

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
```diff
@@ -56,4 +70,17 @@ public String getServiceType() {
     protected DatabaseType dbType() {
         return DatabaseType.OOZIE;
     }
+
+    public static boolean isOozieHA(TemplatePreparationObject source) {
+        String cdhVersion = source.getBlueprintView().getProcessor().getStackVersion() == null ?
```
lacikaaa, line 75: `StringUtils#defaultString` does the same in a bit more compact way

cloud-azure/src/main/resources/templates/arm-v2.ftl
```diff
@@ -525,7 +533,7 @@
                     "name": "${loadBalancer.name}-publicIp",
                     "location": "[parameters('region')]",
                     "sku": {
-                        "name": "Basic"
+                        "name": "Standard"
```
lacikaaa, line 536: I think this wasn't changed earlier for a reason. @brycederriso and @hreeve-cloudera could confirm if this is safe to do

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -77,6 +80,7 @@ private void sendAddDnsARecordRequest(Stack stack, LoadBalancer loadBalancer) {
         request.setHostname(endpoint);
         request.setIp(ip);
         request.setEnvironmentCrn(stack.getEnvironmentCrn());
+        request.setCreateReverse(true);
```
lacikaaa, line 83: when this would be cleaned up from FreeIPA?

lacikaaa: @hreeve-cloudera are the loadbalancer related domain names added to the freeipa cleanup?

lacikaaa (changes requested):
some more unit tests are definitely need eg for `LoadBalancerConfigService`

cloud-azure/src/main/resources/templates/arm-v2.ftl
hreeve-cloudera: The main reason it wasn't changed was because we weren't sure of the impact of doing so, and also because the PR to add Azure load balancers was very large and we didn't want to add additional complexity. However, revisiting this and eventually changing it to standard was always meant to be a follow up item.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: This is just a reordering of some of my log messages. My preferences is to have explicit log messages, instead of trying to implicitly figure out what happened by looking for missing logs.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
hreeve-cloudera: Yes, when the stack is terminated. See `ClusterPublicEndpointManagementService.terminate()`.

Since adding the reverse lookup is part of the original registration request, I would assume the current deletion logic would delete both the forward and reverse entries. But we might want to doublecheck that.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hue/HueConfigProvider.java
```diff
@@ -138,7 +138,6 @@ private String getCdhVersionString(TemplatePreparationObject source) {
 
     private boolean externalFQDNShouldConfigured(GatewayView gateway, GeneralClusterConfigs generalClusterConfigs) {
         return gateway != null
-                && generalClusterConfigs.getKnoxUserFacingCertConfigured()
```
hreeve-cloudera, line 141: Why was this removed?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
```diff
@@ -0,0 +1,65 @@
+package com.sequenceiq.cloudbreak.cmtemplate.configproviders.oozie;
+
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.ConfigUtils.config;
+import static com.sequenceiq.cloudbreak.cmtemplate.configproviders.oozie.OozieRoleConfigProvider.isOozieHA;
+
+import com.sequenceiq.cloudbreak.template.model.GeneralClusterConfigs;
```
hreeve-cloudera, line 6: nit: Out of order import.

hreeve-cloudera:
For Azure LBs, we register the LB under an A record in FreeIPA. I see you've added the reverse DNS entry for the A records. But for AWS LB, we use a CNAME record. Did you have problems with reverse DNS in an AWS environment, or does that appear to be working?

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
cegganesh84: The logic change `else if` to `if` here is to register the FQDN to IP mapping and reverse mapping for both AWS and Azure clusters. Without this change, Hue will not start.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/hue/HueConfigProvider.java
cegganesh84: If it is not removed, and there is a Let's Encrypt outage, Hue knox proxy hosts property is not set. The common workaround for Let's Encrypt outages is to let the cluster be created and users are asked to click the renew certificate button. Unfortunately `generalClusterConfigs.getKnoxUserFacingCertConfigured()` is not set to true for those clusters. This means Hue will become unusable via load-balancer.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: Yup, there is also other enhancements in progress like specifying min node count to 2 so that the customer can not also downscale or start with 0 or 1 nodes.

cegganesh84:
@hreeve-cloudera @lacikaaa I have addressed most of the comments that are code related. Tested it with CCMv2, and no CCM. Tested with AWS, Azure. Tested with semi-private networks, private networks and public networks. The PR is still in draft state because one of the important CM jira is not fixed yet for auto-tls cert verification https://jira.cloudera.com/browse/OPSAPS-55825?page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel&focusedCommentId=4579428#comment-4579428

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
```diff
@@ -72,11 +72,11 @@ private AzureLoadBalancer convertCloudLoadBalancerToAzureLoadBalancer(CloudLoadB
     }
 
     private List<AzureLoadBalancingRule> collectLoadBalancingRules(CloudLoadBalancer cloudLoadBalancer) {
-        return cloudLoadBalancer.getPortToTargetGroupMapping()
-                .keySet()
-                .stream()
-                .map(AzureLoadBalancingRule::new)
-                .collect(toList());
+        List<AzureLoadBalancingRule> list = cloudLoadBalancer.getPortToTargetGroupMapping()
```
fritsi, line 75: No need for the interim variable

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/loadbalancer/AzureLoadBalancingRule.java
```diff
@@ -11,11 +14,15 @@
...
-    public AzureLoadBalancingRule(TargetGroupPortPair portPair) {
+    private final String groupName;
+
+    public AzureLoadBalancingRule(TargetGroupPortPair portPair, Group group) {
         this.backendPort = portPair.getTrafficPort();
         this.frontendPort = portPair.getTrafficPort();
         this.name = defaultNameFromPort(portPair.getTrafficPort());
         this.probe = new AzureLoadBalancerProbe(portPair.getHealthCheckPort());
+        checkNotNull(group);
+        this.groupName = group.getName();
```
fritsi, line 25: `this.groupName = checkNotNull(group, "Expected non-null 'checkNotNull(group)'").getName();`

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -107,6 +109,21 @@
         return groupNames;
     }
 
+    public Set<String> getOozieGroups(Stack stack) {
+        LOGGER.debug("Fetching list of instance groups with Oozie installed");
+        Cluster cluster = stack.getCluster();
+        if (cluster != null) {
+            Set<String> groupNames;
```
fritsi, line 116: Why do we need for the early definition of this variable? You could just do `Set<String> groupNames = cmTemplateProcessor ...` below.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -461,16 +465,34 @@ private void decoratePillarWithClouderaManagerDatabase(Cluster cluster, Map<Stri
...
     }
 
+    private Optional<String> getBestSANForLB(LoadBalancer lb) {
+        if (isNotBlank(lb.getFqdn())) {
+            return Optional.of("DNS:" + lb.getFqdn());
+        }
+        if (isNotBlank(lb.getDns())) {
+            return Optional.of("DNS:" + lb.getDns());
+        }
+        return Optional.of("IP:" + lb.getIp());
```
fritsi, line 493: Should we have a validation in place that the IP is indeed defined?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
```diff
@@ -0,0 +1,65 @@
...
+            && isOozieHA(source)
+            && oozieFQDN(source).isPresent();
+    }
+
+    private static Optional<String> oozieFQDN(TemplatePreparationObject source) {
+        GeneralClusterConfigs generalClusterConfigs = source.getGeneralClusterConfigs();
+        if (generalClusterConfigs != null && generalClusterConfigs.getInternalLoadBalancerEndpoint().isPresent()) {
+            return generalClusterConfigs.getInternalLoadBalancerEndpoint();
+        }
+        return source.getHostGroupsWithComponent(OozieRoles.OOZIE_SERVER)
```
fritsi, line 61: Is this branch even valid? Shouldn't we always have a LoadBalancer configured?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
```diff
@@ -25,18 +31,27 @@
...
+                    config(OOZIE_DATABASE_HOST, oozieRdsView.getHost()),
+                    config(OOZIE_DATABASE_NAME, oozieRdsView.getDatabaseName()),
+                    config(OOZIE_DATABASE_TYPE, oozieRdsView.getSubprotocol()),
+                    config(OOZIE_DATABASE_USER, oozieRdsView.getConnectionUserName()),
+                    config(OOZIE_DATABASE_PASSWORD, oozieRdsView.getConnectionPassword())));
+                if (isOozieHA(source) && CloudPlatform.GCP == source.getCloudPlatform()) {
+                    // There is no load-balancer for GCP yet. If the first Oozie server is down, the second Oozie server can process actions.
+                    // But without the load-balancer the callbacks from Yarn will fail. By default the action check delay is 10 mins.
+                    // Reducing this to 2 mins.
+                    config.add(config(OOZIE_CONFIG_SAFETY_VALVE,
```
fritsi, line 51: Wait, what? I think in this case we should simply say that we don't support Oozie HA for GCP.

What exactly do you mean by _"There is no load-balancer for GCP yet. If the first Oozie server is down, the second Oozie server can process actions."_ ? If no LoadBalancer is configured then how will the request routed to the 2nd instance?

I don't think we should do this. This would still mean a 2 minute delay in processing which is unacceptable to a lot of customers.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
```diff
@@ -56,4 +71,16 @@ public String getServiceType() {
     protected DatabaseType dbType() {
         return DatabaseType.OOZIE;
     }
+
+    public static boolean isOozieHA(TemplatePreparationObject source) {
+        String cdhVersion = StringUtils.defaultString(source.getBlueprintView().getProcessor().getStackVersion());
+
+        if (isVersionNewerOrEqualThanLimited(cdhVersion, CLOUDERA_STACK_VERSION_7_2_11)) {
+            return source.getHostGroupsWithComponent(OozieRoles.OOZIE_SERVER)
+                .filter(hg -> hg.getNodeCount() > 1)
```
fritsi, line 80: Shouldn't this be `>=1` ? What if I have 2 Host groups, each with a single Node, but both of them have an Oozie installed? In this case we'd have 2 total Oozie instances meaning Oozie is in HA.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
cegganesh84: > If no LoadBalancer is configured then how will the request routed to the 2nd instance?

If there is no load-balancer there are only two internal clients that are affected. Hue and Yarn callback. External clients are unaffected because Knox does client side failover. For Yarn callback this is just a temporary hack to reduce the delay in a failure scenario. This is much better than having Oozie in just one server. This is not supported in GA mode. This is a temporary workaround to overcome the problem of having 1 CM template for all three cloud providers. We had an agreement in the design discussion that it is OK for GCP to be in this mode when there is no cloud provider-based load-balancer support. CC: @sidseth @shwethags @mschumpert 

2 mins are just swag and I am ok with even 1 second, if there are no huge performance penalties, in the normal mode of operation.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
cegganesh84: That type of configuration is not supported. It should fail with load-balancer not configured error message if the customer themselves are not adding load-balancers via custom templates. I do not know how common such a setup is. If the customers raise a feature request then I need to re-architect the target groups for the load-balancers to be spread across multiple host groups. In Azure it is strictly checked to be in a single host group at the moment because we are still using availability sets.

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
cegganesh84: No need, getLoadBalancer already filters out the load-balancers that does not have any of these.

```
    public Optional<LoadBalancer> getLoadBalancer(Long stackId, LoadBalancerType loadBalancerType) {
        return loadBalancerPersistenceService.findByStackId(stackId).stream()
            .filter(lb -> StringUtils.isNotBlank(lb.getDns()) ||
                StringUtils.isNotBlank(lb.getIp()) || StringUtils.isNotBlank(lb.getFqdn()))
            .filter(lb -> loadBalancerType.equals(lb.getType()))
            .findAny();
    }
```

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
cegganesh84: This is the GCP case. Please see below.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
cegganesh84: I see that the reverse DNS entries are cleared. I have created 4 to 5 clusters in the same environment. Only one of the reverse entry is leaked. I strongly suspect I did some force delete.

```
  Record name: 75.6
  PTR record: deha2-gateway.gk1dev.xcu2-8y8x.dev.cldr.work.
```

cloud-azure/src/main/resources/templates/arm-v2.ftl
```diff
@@ -201,7 +201,7 @@
                  <#assign createAndNoSecGroup = !isUpscale && (! securityGroups[instance.groupName]?? || ! securityGroups[instance.groupName]?has_content)>
                  <#if !noPublicIp>
                  {
-                   "apiVersion": "2015-05-01-preview",
+                   "apiVersion": "2017-08-01",
```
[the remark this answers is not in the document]
cegganesh84: Hoping that it is not an issue. In my testing I did not find any.

cloud-aws-cloudformation/src/main/resources/templates/aws-cf-stack.ftl
```diff
@@ -315,6 +315,12 @@
           "TargetType" : "instance",
           "HealthCheckPort" : "${listener.targetGroup.healthCheckPort}",
           "HealthCheckIntervalSeconds" : 10,
+          <#if loadBalancer.awsScheme == "internal">
```
hreeve-cloudera, line 318: What's the purpose of this change? Is it just to get around not being able to loopback traffic? Keep in mind that with this, we lose all context for where the traffic originated. I know this is the internal load balancer, so it might not matter as much, but we're losing the ability to filter/set security rules for traffic based on where it originated, because when the traffic gets to the instance it looks like it originated at the load balancer.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -137,6 +155,18 @@ public String getLoadBalancerUserFacingFQDN(Long stackId) {
                 .orElseGet(() -> findPrivateLbNameOrNull(loadBalancers));
     }
 
+    public Optional<LoadBalancer> getLoadBalancer(Long stackId, LoadBalancerType loadBalancerType) {
```
hreeve-cloudera, line 158: Could we change this method name to `getLoadBalancerWithEndpointDefined` or something like that, to make it clear you're only getting a load balancer if the FQDN/DNS/IP are defined?

core/src/test/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigServiceTest.java
```diff
@@ -13,6 +13,7 @@
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
 
+import com.sequenceiq.cloudbreak.domain.stack.instance.InstanceMetaData;
```
hreeve-cloudera, line 16: nit: Import in the wrong order

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -405,9 +491,9 @@ private LoadBalancer createLoadBalancerIfNotExists(Set<LoadBalancer> loadBalance
         return loadBalancer;
     }
 
-    private void setupKnoxLoadBalancer(LoadBalancer loadBalancer, TargetGroup knoxTargetGroup) {
-        loadBalancer.addTargetGroup(knoxTargetGroup);
-        knoxTargetGroup.addLoadBalancer(loadBalancer);
+    private void setupLoadBalancerWithTargetGroup(LoadBalancer loadBalancer, TargetGroup targetGroup) {
+        loadBalancer.addTargetGroup(targetGroup);
+        targetGroup.addLoadBalancer(loadBalancer);
```
brycederriso, line 496: **praise:** Thanks for genericizing the parameter name `knoxTargetGroup` -> `targetGroup`.

brycederriso (approved):
Looks good to me.

cloud-aws-cloudformation/src/main/resources/templates/aws-cf-stack.ftl
cegganesh84: Actually, the security rules are a nonissue for internal load-balancer. We must allow all the traffic between instances in the same subnet. The IP for audit purposes is a legitimate concern. The only use case for the internal load-balancer is Hue/Yarn to Oozie - service to service communication. Which is more of a log to troubleshoot than to do a serious audit about it.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureLoadBalancerModelBuilder.java
```diff
@@ -73,10 +73,9 @@ private AzureLoadBalancer convertCloudLoadBalancerToAzureLoadBalancer(CloudLoadB
 
     private List<AzureLoadBalancingRule> collectLoadBalancingRules(CloudLoadBalancer cloudLoadBalancer) {
         return cloudLoadBalancer.getPortToTargetGroupMapping()
-                .keySet()
-                .stream()
-                .map(AzureLoadBalancingRule::new)
-                .collect(toList());
+            .entrySet().stream()
+            .map(mapping -> new AzureLoadBalancingRule(mapping.getKey(), mapping.getValue().stream().findFirst().orElse(null)))
```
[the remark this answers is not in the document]
cegganesh84: I wouldn't mind either approach. Actually, my first changeset included everything expanded, and after a code review, comment used IntelliJ to shrink it. Just more for my edification, is the single line lambda actually interfere with the readability?

cloud-aws-cloudformation/src/main/resources/templates/aws-cf-stack.ftl
hreeve-cloudera: The internal load balancer is also used for all data lakes that have private subnets, to proxy to Knox. All data lakes have a load balancer. The load balancer is of type internal if the data lake uses private subnets, and internet-facing if the data lake uses public subnets. Basically:
* We're registering the LB FQDN in a public AWS Route 53 zone, because that's what we do for all FQDNs, public and private.
* We're using an NLB, so we don't have any way to add a security policy to the LB, so we don't have a way to block traffic at the load balancer level and need to rely on the EC2 instance security policy.
* With this change, we're making it look like all traffic is coming from the NLB. Which means that as long as the security policy is set up to accept traffic from the load balancer, which is must be, then it will accept all traffic sent to the load balancer, regardless of that traffic's source.

What I think all that means is if someone can find the LB's FQDN, which again is technically public, they can make calls through it regardless of the EC2 instance security policy, because the instance can't tell where the traffic is coming from with this change.

From the testing I did around this previously: "I think I just confirmed the security group thing. I used the existing load balancer config, and modified my security group to only accept HTTPS traffic from the VPC CIDR. I verified I could make an API call through the LB successfully from one of the cluster nodes, but couldn't from my laptop. I then changed the target group to use IP addresses instead of instances ids, and was able to make an API call from my laptop"

What I don't entirely remember is whether or not I was using an internal or an internet-facing NLB there. But based on the JIRA (CB-11925) which was not about the public load balancer, I'm about 90% sure I was testing on an internal NLB. I really think this configuration is going to allow all traffic to flow through the NLB with no way to filter it. That might not matter too much for Hue, Yarn, and Oozie, but it matters a lot for Knox.

cegganesh84: The concept of an internal load-balancer is it is not publicly routable. Only the internet-facing load-balancers are publicly routable. We might have registered it publicly but that does not mean anyone can access it. The internal load-balancers have private IP. Only those who have access to that IP can access any service which is behind the IP. Those who have that access to the internal load-balancer already have access to the service endpoint also. That is why I deem it a nonissue.

> What I don't entirely remember is whether or not I was using an internal or an internet-facing NLB there. But based on the JIRA (CB-11925) which was not about the public load balancer, I'm about 90% sure I was testing on an internal NLB. I really think this configuration is going to allow all traffic to flow through the NLB with no way to filter it. That might not matter too much for Hue, Yarn, and Oozie, but it matters a lot for Knox.

I am 100% sure it is internal load-balancer. You are accessing a load-balancer port 443 which has private IP. Which is impossible unless there is a VPN connection or AWS transit gateway. So the trust is already established via a route with the help of firewall rules at the VPC layer. Now you want to specifically restrict a particular instance's port 443 to restrict inbound traffic from the load-balancer when the request origin is outside of VPC CIDR? What is the use of load-balancer here then?

> I then changed the target group to use IP addresses instead of instances ids, and was able to make an API call from my laptop

Though it does not matter, I would like to point out that, I am not using IP addresses instead of instance ids. All I am doing is not preserving the client IP at the load-balancer layer.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
fritsi: Let's take this conversation to the Jira.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieHAConfigProvider.java
cegganesh84: After Jira conversation addressed this by using the originating server address for Yarn callbacks https://github.com/hortonworks/cloudbreak/pull/10747/files#diff-999ddd6303a84e8956ee01a9ebbf5dc3127d0948e9e1879e784474120edd9a05R50

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
cegganesh84: After Jira conversation addressed this by using the originating server address for Yarn callbacks https://github.com/hortonworks/cloudbreak/pull/10747/files#diff-999ddd6303a84e8956ee01a9ebbf5dc3127d0948e9e1879e784474120edd9a05R50

cloud-aws-cloudformation/src/main/resources/templates/aws-cf-stack.ftl
cegganesh84: Resolved this by moving Hue away from the Oozie hostgroup.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -194,18 +224,22 @@ public boolean isLoadBalancerCreationConfigured(Stack stack, DetailedEnvironment
...
+     * we still use availability set to indicate that we need a {@code standard} IP addresses
+     * for the instances in those availability sets.
      *
      * @param availabilitySetPrefix A string prefix. Should be a stack name normally.
      * @param loadBalancers the list of load balancers to look up target groups from.
      */
     public void configureLoadBalancerAvailabilitySets(String availabilitySetPrefix, Set<LoadBalancer> loadBalancers) {
         getKnoxInstanceGroups(loadBalancers)
                 .forEach(instanceGroup -> attachAvailabilitySetParameters(availabilitySetPrefix, instanceGroup));
+        getOozieInstanceGroups(loadBalancers)
```
sidseth, line 241: This won't clobber anything set previously for the Knox IGs, right? (Looks like both set the same thing - so essentially needs to be called on a union of all the IGs)

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -119,6 +135,8 @@ public String generateLoadBalancerEndpoint(Stack stack) {
         switch (targetGroup.getType()) {
             case KNOX:
                 return Set.of(new TargetGroupPortPair(Integer.parseInt(httpsPort), Integer.parseInt(knoxServicePort)));
+            case OOZIE:
+                return Set.of(new TargetGroupPortPair(Integer.parseInt(OOZIE_HTTPS_PORT), Integer.parseInt(OOZIE_HTTPS_PORT)));
```
sidseth, line 139: Probably a dumb question. In the situation where the LB is shared (public / private) - does this end up opening up external access to Oozie endpoints via the LB (as against via Knox) (Was this port otherwise blocked by network security rules?)

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -389,6 +445,35 @@ private boolean isEndpointGatewayEnabled(EnvironmentNetworkResponse network) {
         return Optional.ofNullable(knoxTargetGroup);
     }
 
+    private Optional<TargetGroup> setupOozieHATargetGroup(Stack stack, boolean dryRun) {
+        Optional<InstanceGroup> oozieInstanceGroup = getOozieInstanceGroup(stack);
```
sidseth, line 449: Can this span across hostGroups? - similar to the way Knox can. Otherwise this ends up restricting Oozie to always be in the same host group, which becomes an unsaid limitation on the definition of custom templates.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
```diff
@@ -389,6 +445,35 @@ private boolean isEndpointGatewayEnabled(EnvironmentNetworkResponse network) {
...
+            LOGGER.info("Oozie HA instance group not found; not enabling Oozie load balancer configuration.");
+            return Optional.empty();
+        }
+    }
+
+    private Optional<InstanceGroup> getOozieInstanceGroup(Stack stack) {
+        Set<String> oozieGroupNames = getOozieGroups(stack);
+        return stack.getInstanceGroups().stream()
+            .filter(ig -> oozieGroupNames.contains(ig.getGroupName()))
+            .filter(ig -> ig.getNodeCount() > 1)
```
sidseth, line 473: Assuming this is the part which sets up the LB only if more than 1 node (specifically 1 HG with more than 1 node) is present.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/CmServiceKeytabRequestFactory.java
```diff
@@ -27,7 +27,7 @@ public ServiceKeytabRequest create(Stack stack, GatewayConfig primaryGatewayConf
         request.setDoNotRecreateKeytab(Boolean.TRUE);
         RoleRequest roleRequest = new RoleRequest();
         roleRequest.setRoleName("hadoopadminrole-" + stack.getName());
-        roleRequest.setPrivileges(Set.of("Service Administrators", "Certificate Administrators", "CA Administrator"));
+        roleRequest.setPrivileges(Set.of("Service Administrators", "Certificate Administrators", "Host Administrators", "CA Administrator"));
```
sidseth, line 30: What does this do?

core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
```diff
@@ -551,7 +551,9 @@
         "refName": "manager",
         "cardinality": 1,
```
sidseth, line 552: Should this be changed to 2, to avoid confusion?

How many of these instances actually exist? Hue has been added on masterx, and manager. Is there 1 or >1 instance of the manager role?

orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls
```diff
@@ -47,6 +48,19 @@ setup_autotls_token:
     - after: "use_tls=.*"
     - backup: False
 
+{% if internal_loadbalancer_san|length > 0 %}
+
+setup_autotls_san:
+  file.line:
+    - name: /etc/cloudera-scm-agent/config.ini
```
sidseth, line 55: What is this change doing?

sidseth (commented):
Assuming node repair / replacement is taken care of by the way the LBs are configured - pointing to a group rather than individual instances?

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: Yes, this will not affect the Knox settings for the default templates. Knox and Oozie are in different instance groups. Let's say there is a custom template with Knox and Oozie in the same host group, it will override the Knox settings with the same values, so it is harmless.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: These ports are always blocked by network security rules. Only the cluster nodes can communicate with each other through the Oozie https port.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: In AWS yes, but in Azure no in this patch. I am using the availability set as an abstraction similar to target groups in AWS. This simplified the Azure changes. While we can remove this restriction to open up for Oozie in multiple host groups by doing some more code changes in this patch, it will make the Azure Freemarker template code changes and corresponding java code changes, even more, harder to follow.

A little bit of history. Before this patch, it was no for Azure. Because we were using basic SKU and it does not support directly registering instances to the Azure load-balancer. In this patch we changed the SKU to register more than one availability set. Instead we could have chosen to register the instances directly without using any availability set. It was harder to code that way in the Azure Freemarker template.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: Yes.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/CmServiceKeytabRequestFactory.java
cegganesh84: When we configure load-balancer in Oozie, CM tries to add the host in FreeIPA. It was failing because it did not have permission to add the role for the host.

core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
cegganesh84: I never understood the meaning of cardinality. At the moment it is treated as minimum node count, but I have not seen any failure directly given if I give less than the cardinality in the cluster definition. The manager is 1 in GCP, 2 in AWS and Azure for 7.2.11. For node count, we should treat the cluster definition as the source of truth.

orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls
cegganesh84: The auto-tls certificates only have the hostname before this change. IF we are calling the Oozie API using load balancer name, it will fail for the hostname verification. So the load-balancer is added as a SAN to make the TLS certificate verification pass.

core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
sidseth: So this ends up with 3 instances of Hue? 2 on the manager nodes, and 1 on masterx?
OT: How does CB handle 2 nodes with CM on it. (Will ping offline)

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
sidseth: OK. So realistically, Knox would need to be on a single hostGroup as well, correct? (Before and after this patch). The code however is written to allow it to span hostGroups.

cegganesh84: Knox is also restricted to only one host group before this patch for Azure. In AWS, you are correct, it was allowed in multiple host groups while technically only one host group will work.

core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
cegganesh84: 3 instances of Hue for AWS and Azure. CM generates a warning that the recommended configuration is only max 2 load-balancers. We can revisit this warning later.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
```diff
@@ -43,10 +43,13 @@ public void registerLoadBalancerDomainWithFreeIPA(Stack stack) {
             try {
                 if (StringUtils.isNotEmpty(loadBalancer.getDns())) {
                     sendAddDnsCnameRecordRequest(stack, loadBalancer);
-                } else if (StringUtils.isNotEmpty(loadBalancer.getIp())) {
+                } else {
+                    LOGGER.error("Unable to find DNS for load balancer. Load balancer address will not be registered as CNAME with FreeIPA.");
+                }
+                if (StringUtils.isNotEmpty(loadBalancer.getIp())) {
                     sendAddDnsARecordRequest(stack, loadBalancer);
```
cegganesh84, line 50: After I moved to private load-balancers for public networks in Azure I am getting this error [1]. While technically I should not because the IP address is added as an A record.

[1]
```
/opt/cloudera/cm/bin/gen_credentials_ipa.sh failed with exit code 1 and output of <<
+ CMF_REALM=GK2DEV.XCU2-8Y8X.DEV.CLDR.WORK
+ export PATH=/usr/kerberos/bin:/usr/kerberos/sbin:/usr/lib/mit/sbin:/usr/sbin:/usr/lib/mit/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
+ PATH=/usr/kerberos/bin:/usr/kerberos/sbin:/usr/lib/mit/sbin:/usr/sbin:/usr/lib/mit/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
+ kinit -k -t /etc/cloudera-scm-server/cmf.keytab CM/test9-manager0.gk2dev.xcu2-8y8x.dev.cldr.work@GK2DEV.XCU2-8Y8X.DEV.CLDR.WORK
+ KEYTAB_OUT=/var/run/cloudera-scm-server/cmf2646773110601411.keytab
+ PRINCIPAL=HTTP/10.2.0.8@GK2DEV.XCU2-8Y8X.DEV.CLDR.WORK
+ MAX_RENEW_LIFE=0
+ '[' -z /etc/krb5.conf ']'
+ echo 'Using custom config path '\''/etc/krb5.conf'\'', contents below:'
+ cat /etc/krb5.conf
+ PRINC=HTTP
++ echo HTTP/10.2.0.8@GK2DEV.XCU2-8Y8X.DEV.CLDR.WORK
++ cut -d / -f 2
++ cut -d @ -f 1
+ HOST=10.2.0.8
+ set +e
+ ipa host-show 10.2.0.8
ipa: ERROR: 10.2.0.8: host not found
+ ERR=2
+ set -e
+ [[ 2 -eq 0 ]]
+ echo 'Adding new host: 10.2.0.8'
+ ipa host-add 10.2.0.8
ipa: ERROR: Host '10.2.0.8' does not have corresponding DNS A/AAAA record
```

cegganesh84 (commented):
Self review after few more combinations of testing.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
fritsi: So with this we are saying that only one group can contain oozie instances?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
fritsi: Follow-up question: if the template already contains a safety-valve, will this extend it, or override it? See: [7.2.11/cdp-data-engineering-ha.bp#L229-L234](https://github.com/hortonworks/cloudbreak/blob/d6d64bbb4ce24b40afa93290bb1a6e502a4e02be/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp#L229-L234)

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
cegganesh84: Yes.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
cegganesh84: It will extend it for different keys. For same keys the templates' value will be given preference, not the code.

core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigService.java
fritsi: Thanks

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/configproviders/oozie/OozieRoleConfigProvider.java
fritsi: So the `oozie.service.HadoopAccessorService.nameNode.whitelist` property coming from the template will also remain in the safety-valve, correct? And this will not remove that property.

cegganesh84: Yes, that is correct. It will *not* remove that property.

sidseth (approved):
Looks good to me.

core/src/main/java/com/sequenceiq/cloudbreak/service/publicendpoint/FreeIPAEndpointManagementService.java
cegganesh84: This is fixed by preferring the FQDN for internal load balancers.

PR #10750, opened 2021-05-25
CB-12543: sourceImageDate missing from custom images

jszeles:
In case of custom image's query operation the related source creation date is also provided.
One option was to persist the date in the database, but this value can change so it should always be updated, so only this value will be retrieved at the moment of the query and sent as part of the response.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageCatalogService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageCatalogService.java
@@ -2,9 +2,12 @@ package com.sequenceiq.cloudbreak.service.image;
 
 import com.sequenceiq.cloudbreak.auth.altus.Crn;
 import com.sequenceiq.cloudbreak.auth.altus.CrnResourceDescriptor;
+import com.sequenceiq.cloudbreak.cloud.model.catalog.Image;
 import com.sequenceiq.cloudbreak.common.exception.BadRequestException;
 import com.sequenceiq.cloudbreak.common.exception.NotFoundException;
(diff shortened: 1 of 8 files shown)


TheTinkerDad (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageCatalogService.java
```diff
@@ -61,6 +64,18 @@ public CustomImage getCustomImage(Long workspaceId, String imageCatalogName, Str
         return getCustomImageFromCatalog(imageCatalog, imageId);
     }
 
+    public Image getSourceImage(CustomImage image, String catalogName) {
+        String imageId = image.getCustomizedImageId();
+        LOGGER.debug(String.format("Get source image '%s' from default catalog", imageId));
+        try {
+            StatedImage sourceStatedImage = imageCatalogService.getSourceImageByImageType(image, catalogName);
```
pkedvessy, line 71: catalogName is the catalog name of the custom image. source image should come from the default image catalog

pkedvessy (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/image/CustomImageCatalogService.java
TheTinkerDad: Nice catch, I didn't notice it!

pkedvessy (approved):

keyki:
Converted it to draft. This will be a htofix and at this point it should not be merged to 2.43.

PR #10753, opened 2021-05-25
CB-12296 BP set to null during the finalization of termiantion but we…

topolyai5:
… want to access it and throw a NPE. Furthermore, the private ips are empty when the cluster does not have any hostgroup but we want to access by the servicename and it throw a NPE.

I have fixed the CM API issues


doktoric (approved):

keyki:
@topolyai5 do you want to push this out as a hotfix? Otherwise it should go to master now.

topolyai5:
> @topolyai5 do you want to push this out as a hotfix? Otherwise it should go to master now.

ok, it can merge into the master

PR #10763, opened 2021-05-25
CB-12098: explicit gcp backup location validation

christmasferret:
Currently user may use gcs://cloudbreak-dev/rdoktorics/telemetry insteand of gs:// url. We should be explicit to enforce a gs:// protocol.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/cloudstorage/CloudStorageLocationValidator.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/cloudstorage/CloudStorageLocationValidator.java
@@ -18,6 +18,7 @@ import com.sequenceiq.environment.credential.v1.converter.CredentialToCloudCrede
 import com.sequenceiq.environment.environment.domain.Environment;
 import com.sequenceiq.environment.environment.dto.EnvironmentBackup;
 import com.sequenceiq.environment.environment.dto.telemetry.EnvironmentLogging;
+import com.sequenceiq.environment.exception.EnvironmentServiceException;
 
 @Component
 public class CloudStorageLocationValidator {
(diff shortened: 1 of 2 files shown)


christmasferret:
Screenshot from 2021-05-25 16-40-10

Before this change, if user mistakenly input gcs:// URL. Cloudbreak will still provision environment and freeipa clusters and fail later when access the gs bucket. Now we want to fail it earlier as gcs:// is a common protocol mistake.

christmasferret:
the error message will be "Status: Creation Failed
The Google storage location [gcs://gcp-eng-sdx-daily-logs/yuwang-gcp-525-t7/] should be a gs:// URL. Refer to Cloudera documentation at https://docs.cloudera.com/cdp/latest/requirements-gcp/topics/mc-gcp_minimum_setup_for_cloud_storage.html for the required rights."

Now the creation will fail within one minutes at the beginning instead of 10 minutes in provision.

environment/src/main/java/com/sequenceiq/environment/environment/validation/cloudstorage/CloudStorageLocationValidator.java
```diff
@@ -34,6 +35,10 @@ public CloudStorageLocationValidator(CloudProviderServicesV4Endopint cloudProvid
 
     public void validate(String storageLocation, Environment environment, ValidationResultBuilder resultBuilder) {
         Optional<FileSystemType> fileSystemType = getFileSystemType(environment);
+        if (fileSystemType.isPresent() && "gs".equals(fileSystemType.get().getProtocol()) && !storageLocation.startsWith("gs://")) {
```
doktoric, line 38: use Filesystemtype enum here

christmasferret: Thanks for the suggestions. I made the change and retested with both CB-CLI and UI.

christmasferret:
@doktoric 
I tested with CB-CLI. Now it will fail immediately if filled with "gcs//" url within one second now instead of failing after 10 minutes during provision. I think this is the intention of the JIRA. Would you kindly take a look again? Thanks!

doktoric (approved):

PR #10767, opened 2021-05-26
ENGESC-8553 do not fail stack downscale if instance is not found on AWS

sodre90:


lnardai (approved):

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsDownscaleService.java
```diff
@@ -155,12 +156,15 @@ private void terminateInstances(Long stackId, List<String> instanceIds, AmazonEc
         try {
             amazonEC2Client.terminateInstances(new TerminateInstancesRequest().withInstanceIds(instanceIds));
         } catch (AmazonServiceException e) {
+            LOGGER.info("Termination failed, lets check if it is because instance was not found", e);
             if (!INSTANCE_NOT_FOUND_ERROR_CODE.equals(e.getErrorCode())) {
                 throw e;
             } else {
+                LOGGER.info("Instance was not found, lets terminate others");
                 List<String> runningInstances = instanceIds.stream()
```
hreeve-cloudera, line 164: Question: Instead of failing in both the `terminateIntances` and `waitForTerminateInstances` methods and retrying, if we reach this case can we just replace `instanceIds` with `runningInstances`, so that we only pass the ids in `runningInstances` to `waitForTerminateInstances`? We only want to wait for termination on instances if we successfully initiated termination, so we might as well pass that information along instead of rediscovering it in the next method.

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsDownscaleService.java
```diff
@@ -155,12 +156,15 @@ private void terminateInstances(Long stackId, List<String> instanceIds, AmazonEc
         try {
             amazonEC2Client.terminateInstances(new TerminateInstancesRequest().withInstanceIds(instanceIds));
         } catch (AmazonServiceException e) {
+            LOGGER.info("Termination failed, lets check if it is because instance was not found", e);
```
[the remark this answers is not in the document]
sodre90: makes sense, I didn't recognize it

cloud-aws/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/connector/resource/AwsDownscaleService.java
sodre90: I thought about this also but I was not sure if I want to modify the code because of this but because you suggested this also and it is not a big work to do then I did it, please check if it is the solution you thought.

sodre90: Yes I tested it on real datalake. Rewrote instance id to an old terminated one and did repair more times.

PR #10770, opened 2021-05-26
CB-12647 Use salt ping to improve salt related conclusion checks

lturcsanyi:
Checking order: salt master services locally, salt minions, running services (currently salt-bootstrap)
Fallback method (network.ipaddrs) for old versions or if nodestatus throws an error.
Use CDP_NODESTATUS_ENABLE_SALT_PING entitlement to enable salt ping reports.
Increase nodestatus client read timeout to 30s.


core/src/main/resources/application.yml
```diff
@@ -699,7 +699,7 @@ cb:
   paywall.url: "https://archive.cloudera.com/p/cdp-public/"
   nodestatus:
     connectionTimeoutMs: 5000
-    readTimeoutMs: 5000
+    readTimeoutMs: 30000
```
oleewere, line 702: hope it won't be too large as during node status checks we are using the same values, and there are 4-5 different request for 1 cluster

oleewere:
im wondering on that also, would it make sense to increase the required cdpTelemetryVersion for the salt ping configuration? just because you have found that bug/typo with the minions (statusReson -> statusReason) - fix was that on the images: https://github.com/hortonworks/cloudbreak-images/commit/776f6075d73388d9385ccfa6dac157fc96b352e6 (that would mean the version should be larger than 9)

otherwise the change is looking good

core/src/main/resources/application.yml
lturcsanyi: I increased the connection timeout to 15 sec.

lturcsanyi:
> im wondering on that also, would it make sense to increase the required cdpTelemetryVersion for the salt ping configuration? just because you have found that bug/typo with the minions (statusReson -> statusReason) - fix was that on the images: [hortonworks/cloudbreak-images@776f607](https://github.com/hortonworks/cloudbreak-images/commit/776f6075d73388d9385ccfa6dac157fc96b352e6) (that would mean the version should be larger than 9)
> 
> otherwise the change is looking good

You already did this :)

oleewere (approved):

core/src/main/java/com/sequenceiq/cloudbreak/conclusion/step/SaltCheckerConclusionStep.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private Conclusion checkUnreachableNodes(Long resourceId) {
+        Stack stack = stackService.getByIdWithListsInTransaction(resourceId);
+        Set<String> allNodes = stackUtil.collectNodes(stack).stream().map(Node::getHostname).collect(Collectors.toSet());
+        try {
+            stackUtil.collectAndCheckReachableNodes(stack, allNodes);
+        } catch (NodesUnreachableException e) {
+            Set<String> unreachableNodes = e.getUnreachableNodes();
+            String conclusion = String.format("Unreachable nodes: %s. Please check the instances on your cloud provider for further details.",
```
lnardai, line 79: I assume this message might get to the customer;
Unreachable nodes can be ambiguous, I would try to clarify the message here, where does the connection problematic, from where we can't reach nodes.

- Is it CDP can't reach the cluster nodes?  (here it might be useful to have another level of digdown, what's not available exactly.)
- Is it a connection problem inside the Cluster and nodes can't talk to each other.

lnardai (approved):

PR #10775, opened 2021-05-26
CB-12691 Introduce new step for default environment create method with Cluster Proxy

aszegedi:
The [default CB communication tunnel](cloudbreak/environment/src/main/resources/application.yml) type is `DIRECT`, but on [QA Cloud Test Environment](https://qa-cloud-canary.eng.hortonworks.com) it has been changed to `CLUSTER_PROXY`. However, all of our tests still use `DIRECT` connection. So we need to change our tests' default environment creation steps to include `with Cluster Proxy`.


horadla23:
lgtm

horadla23 (approved):

PR #10776, opened 2021-05-26
CB-12783. Use max_retry_interval for fluentd exponential backoff retr…

oleewere:
…ies.


vaibhavgandhi12 (approved):
🚢

PR #10783, opened 2021-05-27
ODX-684 Move hadoop.tmp.dir out of /tmp

anmolnar:
Added new `hadoop.tmp.dir` setting to all COD specific blueprints.


doktoric:
the only thing here is that it will not work on ycloud.

anmolnar:
@doktoric Do you know a good way to overcome that?

anmolnar:
@doktoric I'm done with the changes that we discussed.

doktoric:
retest this please

doktoric (approved):

PR #10784, opened 2021-05-27
CB-12821. Provide uuid parameter in the internal datalake list diagno…

oleewere:
…stics collection response


aakanser:
looks good to me

You can also consider renaming it to "datalake-flowid" since that would have a better meaning in this context than "uuid".

aakanser (approved):
looks good to me

PR #10787, opened 2021-05-28
CB-12054. Create databus client (grpc) in CB (for processing data from CP)

oleewere:
details:
- add new module: databus-connector (proto based, see sigmadbus.proto)
- create a SigmaDbus client - it won't close message channels on every call (as for other grpc based calls) as a lot of operations can be executed.
- in order to not block other operations, create blocking queue + worker pairs, for processing data is put into blocking queue and the worker runs on a different thread, it uses the grpc client for processing
- multiple workers (+ clients) can be used based on application configs (means more threads), also queues has a limit in order to not grow memoery too high if the communication would not work and hanging for some reason
- any new streaming processor can be added with providing a new implementation of AbstractDatabusRecordProcessor + using a new Databus streaming configurations
- if data to process is larger than 1MB -> use S3 api
- add entitlements for freeipa+DL and for DH
- added unit tests
- add processor: MetricsDatabusRecordProcessor -> that send system metrics and/or CM metrics reports to CP


core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerJob.java
```diff
@@ -97,43 +110,87 @@ private void executeNodeStatusCheck(Stack stack) {
             }
             CdpNodeStatusRequest request = requestBuilder.withCmMonitoring(true).build();
             CdpNodeStatuses statuses = nodeStatusService.getNodeStatuses(stack, request);
-            processNodeStatusReport(statuses.getNetworkReport(), NodeStatusProto.NodeStatus::getNetworkDetails, stack, "Network");
-            processNodeStatusReport(statuses.getServicesReport(), NodeStatusProto.NodeStatus::getServicesDetails, stack, "Services");
-            processNodeStatusReport(statuses.getSystemMetricsReport(), NodeStatusProto.NodeStatus::getSystemMetrics, stack, "System metrics");
-            processCmMetricsReport(statuses.getCmMetricsReport(), stack);
+            processNodeStatusReport(statuses.getNetworkReport(), NodeStatusProto.NodeStatus::getNetworkDetails, stack, "Network", accountId);
+            processNodeStatusReport(statuses.getServicesReport(), NodeStatusProto.NodeStatus::getServicesDetails, stack, "Services", accountId);
+            final boolean processMetrics;
```
aakanser, line 115: maybe consider `processMetricsToDatabus` here and within `processMnMetricsReport`.

core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerJob.java
```diff
@@ -97,43 +110,87 @@ private void executeNodeStatusCheck(Stack stack) {
...
+                    stack, "System metrics", accountId, processMetrics);
+            processCmMetricsReport(statuses.getCmMetricsReport(), stack, accountId, processMetrics);
         }
     }
 
-    private void processCmMetricsReport(Optional<RPCResponse<NodeStatusProto.CmMetricsReport>> cmMetricsReport, Stack stack) {
+    private void processCmMetricsReport(Optional<RPCResponse<NodeStatusProto.CmMetricsReport>> cmMetricsReport, Stack stack,
+            String accountId, boolean process) {
         if (cmMetricsReport.isPresent()) {
             LOGGER.debug("CM metrics report: {}", cmMetricsReport.get().getFirstTextMessage());
```
aakanser, line 131: is this log statement that important? Otherwise, you could get rid of the boolean process argument and just not call this method based on the boolean value.

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
```diff
@@ -0,0 +1,160 @@
...
+            uploadRecordToS3(s3BucketUrl, request, recordId);
+        } else {
+            throw new DatabusRecordProcessingException("Cannot process record to Sigma Databus.");
+        }
+    }
+
+    private void uploadRecordToS3(String s3Url, DatabusRequest request, String recordId) throws DatabusRecordProcessingException {
+            String payload = getPayload(request);
+        try {
+            HttpClient client = HttpClient.newHttpClient();
```
aakanser, line 90: does this need closing? you can use try-resources if auto-closeable with try(HttpClient  client = ...) {}

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
```diff
@@ -0,0 +1,160 @@
...
+            HttpRequest httpRequest = HttpRequest.newBuilder()
+                    .uri(new URI(s3Url))
+                    .PUT(HttpRequest.BodyPublishers.ofString(payload))
+                    .build();
+            HttpResponse<String> response = client.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            Response.Status.Family statusFamily = Response.Status.Family.familyOf(response.statusCode());
+            if (Response.Status.Family.SUCCESSFUL.equals(statusFamily) || Response.Status.Family.REDIRECTION.equals(statusFamily)) {
+                LOGGER.debug("Databus record with id {} successfully uploaded to s3.", recordId);
+            } else {
+                throw new DatabusRecordProcessingException(String.format("S3 upload failed for databus record with id %s", recordId));
```
aakanser, line 100: it would be good to emit a LOGGER.warn with the Response Status and maybe reason as a clue why it's failing.

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    public void putRecord(DatabusRequest request) throws DatabusRecordProcessingException {
+        ManagedChannelWrapper channelWrapper = getMessageWrapper();
+        DbusProto.PutRecordRequest recordRequest = convert(request, databusStreamConfiguration);
+        String requestId =  RequestIdUtil.newRequestId();
+        LOGGER.debug("Creating databus request with request id: {}", requestId);
+        buildMdcContext(request, requestId);
+        DbusProto.PutRecordResponse recordResponse = newStub(channelWrapper.getChannel(),
+                requestId, ThreadBasedUserCrnProvider.INTERNAL_ACTOR_CRN)
+                .putRecord(recordRequest);
```
aakanser, line 72: try-catch for this?

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
```diff
@@ -0,0 +1,160 @@
...
+            String s3BucketUrl = recordResponse.getRecord().getUploadUrl();
+            LOGGER.debug("Dbus record can be uploaded to s3 [record id: {}], [s3 url: {}]", recordId, s3BucketUrl);
+            uploadRecordToS3(s3BucketUrl, request, recordId);
+        } else {
+            throw new DatabusRecordProcessingException("Cannot process record to Sigma Databus.");
+        }
+    }
+
+    private void uploadRecordToS3(String s3Url, DatabusRequest request, String recordId) throws DatabusRecordProcessingException {
+            String payload = getPayload(request);
```
aakanser, line 88: maybe null/empty checks for the arguments just to be sure

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/processor/AbstractDatabusRecordProcessor.java
```diff
@@ -0,0 +1,155 @@
...
+    protected void handleDatabusRecordProcessingException(DatabusRequest input, DatabusRecordProcessingException e) {
+        LOGGER.debug(String.format("Exception during databus record processing [skip] - input: %s", input), e);
+    }
+
+    /**
+     * Override this to change the default behavior when an unexpected error happens.
+     * Default behaviour: log the exception.
+     */
+    protected void handleUnexpectedException(DatabusRequest input, Exception e) {
+        LOGGER.debug(String.format("Unexpected exception during databus record processing [skip] - input: %s", input), e);
```
aakanser, line 103: I think LOGGER.warn is better for this.

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/processor/AbstractDatabusRecordProcessor.java
```diff
@@ -0,0 +1,155 @@
...
+            LOGGER.debug("Databus processing is not enabled, skip text message processing to databus");
+        }
+    }
+
+    /**
+     * Override this to change the default behavior when an databus record processing error happens.
+     * Default behaviour: log the exception.
+     */
+    protected void handleDatabusRecordProcessingException(DatabusRequest input, DatabusRecordProcessingException e) {
+        LOGGER.debug(String.format("Exception during databus record processing [skip] - input: %s", input), e);
```
aakanser, line 95: I think LOGGER.warn is better for this.

aakanser (approved):
Looks good to me, left a few comments

core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerJob.java
oleewere: technically that is just processing/logging, data is only sent to databus if that is enabled during the processing, im keeping this for now

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
oleewere: https://stackoverflow.com/questions/53919721/close-java-http-client
you can see this one is not closable, based on the comments it should handle that after that is not referenced anymore

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
oleewere: we can't put nulls into the blocking queue, and its getters are optional, so we are good (I know it's more robust to check that at multiple places so even unit tests can't break that but i do not like to doing that)

core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerJob.java
oleewere: yes, so if databus processing is not enabled, which can be the case, it will logs out the response on CB side, which is still useful, also it's possible we are just getting json output but not fitting with the nodestatus proto (in earlier nodestatus monitor changes, there were no proto objects implemented ... there was a different json based response ... so in that case, proto object is not filled but we still have a text response that can be logged out)

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
oleewere: yeah, could be, i choosed to just logging out the status code itself. before that exception, which is useful for the success operations as well

core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerConfig.java
```diff
@@ -12,7 +12,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NodeStatusCheckerConfig.class);
 
-    @Value("${nodestatuschecker.intervalsec:1200}")
+    @Value("${nodestatuschecker.intervalsec:1800}")
```
attilapalfi92, line 15: why the increase here?

core/src/main/resources/application.yml
```diff
@@ -711,6 +711,14 @@ cb:
       sdx:
         enabled: false
 
+cluster:
+  monitoring:
+    enabled: false
+    databus:
+      processing:
+        workers: 1
```
attilapalfi92, line 719: 1 worker is enough for default? I don't know, I'm just asking.

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
attilapalfi92: imagine if we were using kotlin... :P

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
```diff
@@ -0,0 +1,162 @@
...
+        try {
+            HttpClient client = HttpClient.newHttpClient();
+            HttpRequest httpRequest = HttpRequest.newBuilder()
```
attilapalfi92, line 92: Does CB in our deployments have the permission to write to this s3 bucket?

attilapalfi92 (commented):
I just had a few questions but LGTM. Approval after answers received.

databus-connector/src/main/java/com/sequenceiq/cloudbreak/sigmadbus/SigmaDatabusClient.java
oleewere: the url response itself contains an authentication token to access that bucket, so yes

core/src/main/java/com/sequenceiq/cloudbreak/job/nodestatus/NodeStatusCheckerConfig.java
oleewere: we do not need these data too often, so 30 mins looked better choice

core/src/main/resources/application.yml
oleewere: for every cluster one record needs to be processed in every 30 min, so 1 worker should be good enough (like UMS calls are much more frequent for clusters), but if at any point we need to scale this (as the requests will be too frequent based on the number of the clusters), workers can be increased to send requests in parallel.

attilapalfi92 (approved):

PR #10789, opened 2021-05-31
CB-12564: Allow SDX upgrade when DataHubs with ephemeral disk or with spot instances are not stopped.

tiborpopovics:


core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/Stack.java
```diff
@@ -655,7 +655,7 @@ public StopRestrictionReason isInfrastructureStoppable() {
         if ("AWS".equals(cloudPlatform())) {
             for (InstanceGroup instanceGroup : instanceGroups) {
                 if (instanceGroup.getTemplate().getVolumeTemplates().stream()
-                        .filter(volume -> "ephemeral".equals(volume.getVolumeType())).findAny().isPresent()) {
+                        .anyMatch(volume -> "ephemeral".equals(volume.getVolumeType()))) {
```
lacikaaa, line 658: we should have `ephemeral` as a constant somewhere instead of a string multiple places

lacikaaa: is there any unit test for this?

lacikaaa (changes requested):
good job, please check those nits

keyki (approved):
Tested locally.

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/Stack.java
tiborpopovics: Yes, please check now.

PR #10792, opened 2021-05-31
CB-11570 SQL error in SDX service, because: value too long for type character varying(255)

foldik:


horadla23 (approved):
LGTM

PR #10793, opened 2021-05-31
CB-11720 Tie volume sets to discovery FQDNs

daszabo:
Some services are sensitive to have the volumes reattached to a machine
    with the same FQDN.

    * Discovery FQDN now also stored in volumesets
    * Discovery FQDN saved the same time as mount info
    * For already running clusters, discovery FQDN ges saved at the start of repair
    * During bootstraping new nodes, discovery FQDN gets read from volume sets to
    set the proper hostname based on instance id


daszabo:
Reopening the previously reverted commit. Removed the query, which fetch joins the stack table with resources (among others). Disk resources now fetched in a separate transaction, using `ResourceService::findByStackIdAndType`. This filters by stack id and resource type.

lacikaaa (approved):

PR #10802, opened 2021-05-31
CB-11797 - Added log cloud storage permission validation for deployment flows (FreeIPA/DataLake/DataHub)

attilapalfi92:
- Adding its own flow step for cloud storage validation on VMs by writing a test sample to the cloud storage.
- Executing the validation soon as possible.
- Skipping the validation if no telemetry logging is set.
- For DataHub deployments, randomly choosing 1 host, and doing the validation only on that node.
- Doing it similarly for Datalake, plus validation on the IDBroker node which does not cause the flow to fail, just sends a warning notification.

https://github.infra.cloudera.com/thunderhead/thunderhead/pull/4873


oleewere (approved):
lgtm

PR #10803, opened 2021-05-31
CB-12505 Mount ephemeral volumes to different file path based on temporary storage value

daszabo:
When the temporaryStorage attribute (coming from an internal datahub post request) is set to EPHEMERAL_VOLUMES, instance storage should be mounted on the nodes under /hadoopfs/ephfsX.

### DIFF
--- a/orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
+++ b/orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.orchestrator.model;
 
+import com.sequenceiq.cloudbreak.common.type.TemporaryStorage;
+
 public class Node {
     private final String privateIp;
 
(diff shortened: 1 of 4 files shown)


orchestrator-api/src/main/java/com/sequenceiq/cloudbreak/orchestrator/model/Node.java
```diff
@@ -17,13 +19,16 @@
...
     private NodeVolumes nodeVolumes;
 
+    private TemporaryStorage temporaryStorage;
+
     public Node(String privateIp, String publicIp, String instanceId, String instanceType, String fqdn, String hostGroup) {
         this(privateIp, publicIp, instanceId, instanceType, fqdn, null, hostGroup);
     }
 
-    public Node(String privateIp, String publicIp, String instanceId, String instanceType, String fqdn, String hostGroup, NodeVolumes nodeVolumes) {
+    public Node(String privateIp, String publicIp, String instanceId, String instanceType, String fqdn, String hostGroup, NodeVolumes nodeVolumes, TemporaryStorage temporaryStorage) {
```
nadamdb, line 28: Checkstyle is failing on this line, because it's too long

nadamdb (changes requested):
Besides the checkstyle failing line LGTM

nadamdb (approved):

PR #10806, opened 2021-06-01
CB-12840 Fix cluster provision in case of custom image where the image catalog url is null

pkedvessy:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
@@ -28,6 +28,8 @@ import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
 
+import com.google.common.base.Strings;
+import com.sequenceiq.common.api.type.ImageType;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
```diff
@@ -317,12 +319,41 @@ public StatedImages getImages(Long workspaceId, String name, String provider) th
     public StatedImages getImages(Long workspaceId, String name, Set<String> providers) throws CloudbreakImageCatalogException {
         try {
             ImageCatalog imageCatalog = get(workspaceId, name);
-            return getImages(new ImageFilter(imageCatalog, providers, cbVersion, baseImageEnabled(), null, null));
+            if (Strings.isNullOrEmpty(imageCatalog.getImageCatalogUrl()) && imageCatalog.getCustomImages() != null) {
```
TheTinkerDad, line 322: Conditions like this that clearly identify something, IMHO could be extracted into a function, even if only called a few times. Something like isCustomCatalog(imageCatalog) looks more readable.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeAvailabilityService.java
```diff
@@ -157,12 +157,17 @@ private UpgradeV4Response checkForUpgrades(Stack stack, boolean lockComponents)
...
             LOGGER.info(String.format("Retrieving images for upgrading stack %s", stack.getName()));
             com.sequenceiq.cloudbreak.cloud.model.Image currentImage = getImage(stack);
-            CloudbreakImageCatalogV3 imageCatalog = getImagesFromCatalog(stack.getWorkspace(), currentImage);
-            Image image = getCurrentImageFromCatalog(currentImage.getImageId(), imageCatalog);
-            ImageFilterParams imageFilterParams = imageFilterParamsFactory.create(image, lockComponents, stack);
-            ImageFilterResult filteredImages = filterImages(stack.getCreator().getTenant().getName(), imageCatalog, stack.cloudPlatform(), imageFilterParams);
-            LOGGER.info(String.format("%d possible image found for stack upgrade.", filteredImages.getAvailableImages().getCdhImages().size()));
-            upgradeOptions = createResponse(image, filteredImages, stack.getCloudPlatform(), stack.getRegion(), currentImage.getImageCatalogName());
+            if (currentImage.getImageCatalogUrl() == null) {
+                upgradeOptions.setReason(String.format("Upgarde is not supported by clusters with custom image"));
```
TheTinkerDad, line 161: I think a TODO or a FIXME would be nice here - currently it looks like we don't even plan to support it. Or at least add a "yet" to the end of the message :)

TheTinkerDad (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogService.java
pkedvessy: ACK

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterUpgradeAvailabilityService.java
pkedvessy: It was there but I have removed it as the message states the current behavior :-) But yes, I can agree.

TheTinkerDad (approved):

PR #10808, opened 2021-06-01
CB-8382 - [azure] ARM template is redeployed if it exists

pdarvasi:
This commit adds an extra check to recreate the ARM template instead of reusing it if it is already finished (meaning its status is not ResourceStatus.IN_PROGRESS)

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
@@ -133,7 +133,7 @@ public class AzureResourceConnector extends AbstractResourceConnector {
         boolean resourcesPersisted = false;
         try {
             List<CloudResource> instances;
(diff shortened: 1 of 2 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
```diff
@@ -133,7 +133,7 @@
         boolean resourcesPersisted = false;
         try {
             List<CloudResource> instances;
-            if (!client.templateDeploymentExists(resourceGroupName, stackName)) {
+            if (createTemplateDeployment(stackName, resourceGroupName, client)) {
```
gergopapi2, line 136: minor: I would change the name to
`shouldCreateTemplateDeplyoment` or to  `isTemplateDeploymentNeeded`
to reflect that this is a method deciding what to do and it will not create the template.

But given that my reveiw is late, I do not insist on this.

gergopapi2 (approved):

PR #10811, opened 2021-06-01
CB-12842 Azure disk SSE with CMK fix for DH request azure == null

lajosrodek:
* Handle the case when the input `InstanceTemplateV1Request` has `azure == null`; Azure disk SSE with CMK will be applied to the DH disks even in such scenarios if the environment has prescribed it so. This fixes the former broken logic of #10371 that failed to apply disk encryption for such requests.
* Testing: Manual verification in local CB; updated existing UT.

### DIFF
--- a/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
+++ b/core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
@@ -4,6 +4,8 @@ import static com.sequenceiq.cloudbreak.util.NullUtil.getIfNotNull;
 
 import java.util.Optional;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.parameter.template.AwsEncryptionV4Parameters;
@@ -28,6 +30,7 @@ import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvi
 
 @Component
 public class InstanceTemplateParameterConverter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(InstanceTemplateParameterConverter.class);
 
     public AwsInstanceTemplateV4Parameters convert(AwsInstanceTemplateV1Parameters source) {
         AwsInstanceTemplateV4Parameters response = new AwsInstanceTemplateV4Parameters();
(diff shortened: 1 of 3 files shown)


bergerdenes (approved):

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
```diff
@@ -80,11 +83,15 @@ private void initAzureEncryptionFromEnvironment(AzureInstanceTemplateV4Parameter
                 .map(AzureEnvironmentParameters::getResourceEncryptionParameters)
                 .map(AzureResourceEncryptionParameters::getDiskEncryptionSetId)
                 .orElse(null);
+        String environmentName = environment.getName();
         if (diskEncryptionSetId != null) {
+            LOGGER.info("Applying SSE with CMK for Azure managed disks as per environment \"{}\".", environmentName);
```
lacikaaa, line 88: shouldn't the log context contain every necessary information and env name is just duplication of data here?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateV1ToInstanceTemplateV4Converter.java
```diff
@@ -25,7 +28,10 @@ public InstanceTemplateV4Request convert(InstanceTemplateV1Request source, Detai
         response.setAttachedVolumes(getIfNotNull(source.getAttachedVolumes(), volumeConverter::convertTo));
         response.setEphemeralVolume(getIfNotNull(source.getEphemeralVolume(), volumeConverter::convert));
         response.setAws(getIfNotNull(source.getAws(), instanceTemplateParameterConverter::convert));
-        response.setAzure(getIfNotNull(source.getAzure(), environment, instanceTemplateParameterConverter::convert));
+        AzureInstanceTemplateV1Parameters azureParametersEffective = Objects.requireNonNullElse(source.getAzure(), new AzureInstanceTemplateV1Parameters());
+        // InstanceTemplateParameterConverter.convert(AzureInstanceTemplateV1Parameters, DetailedEnvironmentResponse) has to be always invoked for sake of
+        // Azure disk encryption!
```
lacikaaa, line 33: do we really need this comment? commit message and unit test should be enough to ensure we don't delete the code which is necessary. maybe some higher level test to make sure

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
lajosrodek: Good point, will remove the msg parameter.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateV1ToInstanceTemplateV4Converter.java
lajosrodek: Makes sense. Will remove the comment in a moment.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateParameterConverter.java
lajosrodek: Done.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/converter/InstanceTemplateV1ToInstanceTemplateV4Converter.java
lajosrodek: Done.

lacikaaa (approved):

PR #10818, opened 2021-06-02
CDPAM-1763 | Ship jumpgate agent logs from access and out logs at the time of env creation

SachinHg:
The log locations to collect the info is updated based on the recent changes where we introduced two log files(access, out) for jumpgate agent in [CDPAM-1553](https://jira.cloudera.com/browse/CDPAM-1553)


oleewere (approved):
lgtm

PR #10824, opened 2021-06-02
CB-12695-new-tunneltype

bergerdenes:


lacikaaa (changes requested):
this would switch all new envs with ccmv2 entitlement to the jumpgate solution. as I don't see any change regarding userdata code I'm worried this would break cluster creation when CCMv2 entitlement is enabled.

bergerdenes:
Added new tunnel type only for now. The next step is to add the conversion for environment response in public API. Then userdata will be modified along with defaulting to new CCMV2_JUMPGATE tunnel type.

lacikaaa (approved):

PR #10830, opened 2021-06-02
CB-12883 Create CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY_AZURE entitlement

hreeve-cloudera:
Creates a new CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY_AZURE for the Azure environment Endpoint Gateway feature. Also removes the CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY entitlement, which was being used for the AWS Endpoint Gateway feature which has gone GA.

This change reworks the logic that was checking the endpoint gateway entitlement. Originally we were checking it in many places, which didn't match best practices. Now, we're instead checking it in only two places: when the environment creation request is received, and when a request to enable the feature on an existing environment is received. If the request does not pass entitlement validation when initially received, the request will fail. If it passes, then we don't continue to chek the entitlement at other places in the code, which is what was being done previously.

Tested with unit tests.


environment/src/main/java/com/sequenceiq/environment/network/service/LoadBalancerEntitlementService.java
```diff
@@ -0,0 +1,34 @@
...
+import com.sequenceiq.cloudbreak.common.type.CloudConstants;
+import com.sequenceiq.common.api.type.PublicEndpointAccessGateway;
+
+@Component
+public class LoadBalancerEntitlementService {
+
+    @Inject
+    private EntitlementService entitlementService;
+
+    public void valiateNetworkForEndpointGateway(String cloudPlatform, String environmentName,
```
sxxgrc, line 19: validate is misspelled :( sorry since it's used in a few places

hreeve-cloudera: nooooooooo...

sxxgrc (approved):
Aside from the one misspelling, everything looks good. Thank you for the change.

PR #10845, opened 2021-06-03
CB-12688 - Modify credential validation to make sure we are only checking the combined minimal policy in case of validation

gregito:


cloud-aws-common/src/test/java/com/sequenceiq/cloudbreak/cloud/aws/common/AwsCredentialVerifierTest.java
```diff
@@ -102,7 +102,7 @@ public void verifyCredentialAndThrowFailExceptionTest() throws IOException {
             assertThat(e.getMessage(), not(CoreMatchers.containsString("accepted_action")));
         }
         List<SimulatePrincipalPolicyRequest> allSimulatePrincipalPolicyRequest = requestArgumentCaptor.getAllValues();
-        int simulateRequestNumber = 4;
+        int simulateRequestNumber = 2;
```
doktoric, line 105: why only 2?

gregito: @doktoric 
The key difference here is that in the old way we had 2 other Sids that has a resource other than '*': 

- The "LogsLimited" with the actions of ["logs:CreateLogStream", "logs:DescribeLogStreams", "logs:PutLogEvents"] for the resource "arn:aws:logs:\*:\*:log-group:/aws/eks/\*:\*" 
- And the "PerformanceInsightsLimited" for action "pi:\*" for the resource "arn:aws:pi:\*:\*:metrics/rds/\*"

The lack of these two entries make the test output different.

doktoric (approved):

PR #10847, opened 2021-06-03
CB-12843 Make sure we have PKs on every CB table

bbihari:
Public key is necessary for maintaining replica identity in postgres, in this pull request I've added pkey where it was missing.


foldik (approved):
LGTM

PR #10861, opened 2021-06-08
CB-12573: Disallow OS upgrade for flow management blueprints.

tiborpopovics:


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/image/BlueprintUpgradeOptionValidator.java
```diff
@@ -14,18 +18,31 @@
...
+    private boolean isCustomBlueprint(Blueprint blueprint) {
+        return !blueprint.getStatus().isDefault();
+    }
+
+    private boolean isEnabled(boolean lockComponents, Blueprint blueprint) {
+        return lockComponents ? isOsUpgradeEnabled(blueprint) : isRuntimeUpgradeEnabled(blueprint);
+    }
+
+    private boolean isOsUpgradeEnabled(Blueprint blueprint) {
+        BlueprintUpgradeOption upgradeOption = blueprint.getBlueprintUpgradeOption();
```
keyki, line 36: Shouldn't we use the `getBlueprintUpgradeOption` here as well?

core/src/main/resources/defaults/blueprints/7.2.11/cdp-flow-management-small.bp
```diff
@@ -3,6 +3,7 @@
   "blueprint": {
     "cdhVersion": "7.2.11",
     "displayName": "flow-management",
+    "blueprintUpgradeOption": "OS_UPGRADE_DISABLED",
```
keyki, line 6: For 7.2.11 this should be allowed. Anything that is below this version must be disabled.

keyki (approved):

PR #10868, opened 2021-06-08
CB-12943 Fix upgrade matrix to contain valid regex

lacikaaa:
- `.` is replaced with `\\.` so it would be escaped in Java
- `[0,1,2,3,10]` like has been replaced with `(0|1|2|3|10)`. The first mean any of these: `0` `1` `2` `3` `,`

### DIFF
--- a/core/src/main/resources/definitions/upgrade-matrix-definition.json
+++ b/core/src/main/resources/definitions/upgrade-matrix-definition.json
@@ -2,95 +2,95 @@
   "runtime_upgrade_matrix": [
     {
       "target_runtime": {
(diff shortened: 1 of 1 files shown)


lajosrodek (approved):
LGTM

akanto (changes requested):
Please extend the commit message with a proper explanation similar what is added to the pull request comment.

PleSE write unit tests, so test with 2 digits and also with 3 digits and also use negative tests with no-matching values like 7.2X10, etc.

akanto (approved):

PR #10869, opened 2021-06-08
CB-12906 whenException does not throw TestFailException in case of error

aszegedi:
TestContext's [whenException](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java#L266) was not failing in case of the expected exception has not been thrown and the doAction goes through with no error.

So we should put a related exception to the getExceptionMap() in case of the expected exception has not been thrown.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
@@ -263,7 +263,14 @@ public abstract class TestContext implements ApplicationContextAware {
         Log.whenException(LOGGER, action.getClass().getSimpleName() + " action on " + entity + " by " + who);
 
         try {
-            return doAction(entity, clientClass, action, who.getAccessKey());
+            String message = String.format("Expected exception with message (%s) has not been thrown at action (%s)!",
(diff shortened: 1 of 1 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/context/TestContext.java
```diff
@@ -263,7 +263,14 @@ public InstanceAwait getInstanceAwait() {
         Log.whenException(LOGGER, action.getClass().getSimpleName() + " action on " + entity + " by " + who);
 
         try {
-            return doAction(entity, clientClass, action, who.getAccessKey());
+            String message = String.format("Expected exception with message (%s) has not been thrown at action (%s)!",
```
topolyai5, line 266: What will happen if the `runningParameter.getExpectedMessage()` is empty or null?

aszegedi: This is going to show an `()` in the message. I think no other issue is going to be generated.

topolyai5 (approved):

PR #10871, opened 2021-06-08
CB-12962 set-endpoint-access-gateway fails on existing Medium duty environment

hreeve-cloudera:
When populating the instance metadata for the stack, include the metadata for all gateway instances
instead of only the primary gateway metadata. This will allow the set-endpoint-access-gateway
functionality to be run on stack with multiple gateway isstance, e.g. medium duty data lakes.

Tested by calling the set-endpoint-access-gateway API on a medium duty environment.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/RegisterPublicDnsHandler.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/RegisterPublicDnsHandler.java
@@ -1,7 +1,8 @@
 package com.sequenceiq.cloudbreak.reactor.api.event.stack.loadbalancer.handler;
 
-import java.util.Optional;
+import java.util.List;
 import java.util.Set;
+
 import javax.inject.Inject;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


horadla23:
which API is "set-endpoint-access-gateway API"?

hreeve-cloudera:
> which API is "set-endpoint-access-gateway API"?

This is the EnvironmentEndpoint.updateEnvironmentLoadBalancersByName API.

horadla23:
and have you tested it locally (i guess yes)? I just want to make sure that the test covers this change, since it seems we need to allows cert generation in config (`gateway.cert.generation.enabled`)

hreeve-cloudera:
> and have you tested it locally (i guess yes)? I just want to make sure that the test covers this change, since it seems we need to allow cert generation in config (`gateway.cert.generation.enabled`)

Yep. I created a medium duty data lake in my local CB services, and then invoked that API using a curl call. My local dev setup is configured to use the mow-dev PEM service for DNS registration and to generated certs, so I did have "gateway.cert.generation.enabled" set to true and the path to renew the certs was hit.

horadla23:
> > and have you tested it locally (i guess yes)? I just want to make sure that the test covers this change, since it seems we need to allow cert generation in config (`gateway.cert.generation.enabled`)
> 
> Yep. I created a medium duty data lake in my local CB services, and then invoked that API using a curl call. My local dev setup is configured to use the mow-dev PEM service for DNS registration and to generated certs, so I did have "gateway.cert.generation.enabled" set to true and the path to renew the certs was hit.

cool, that's what I wanted to ask, if you used mow-dev PEM. in this case that test should cover the change

horadla23 (approved):
LGTM

horadla23:
@hreeve-cloudera pls do not forget to create a merge PR to master or ask @pkedvessy to do that since he has a recent change committed into 2.44 too

PR #10872, opened 2021-06-09
CB-12922 Fix cluster upgrade in case of the image catalog url is missing.

pkedvessy:


jszeles (approved):

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/image/update/StackImageUpdateService.java
```diff
@@ -87,6 +91,10 @@ public Image getImageModelFromStatedImage(Stack stack, Image currentImage, State
 
     public StatedImage getNewImageIfVersionsMatch(Stack stack, String newImageId, String imageCatalogName, String imageCatalogUrl) {
         try {
+            if (restRequestThreadLocalService.getRequestedWorkspaceId() == null) {
```
horadla23, line 94: have you experienced that it can be null? we have a filter to fill that in

horadla23: discussed in slack

horadla23 (approved):
lgtm

PR #10877, opened 2021-06-09
CB-12774 its possible that a RG contains no VMs, 

lnardai:
e.g repair both hostgroups, we should allow it to continue

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureVirtualMachineService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureVirtualMachineService.java
@@ -44,9 +44,21 @@ public class AzureVirtualMachineService {
     @Inject
     private AzureResourceGroupMetadataProvider azureResourceGroupMetadataProvider;
 
-    @Retryable(backoff = @Backoff(delay = 1000, multiplier = 2, maxDelay = 10000), maxAttempts = 5)
+    @Retryable(backoff = @Backoff(delay = 1000, multiplier = 2, maxDelay = 10000), maxAttempts = 10)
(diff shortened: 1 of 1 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureVirtualMachineService.java
```diff
@@ -129,7 +145,7 @@ private AzureVirtualMachinesWithStatuses getVmsFromAzureAndFillStatusesIfResourc
         for (Map.Entry<String, Collection<String>> resourceGroupInstanceIdsMap : resourceGroupInstanceMultimap.asMap().entrySet()) {
             LOGGER.info("Get vms for resource group and add to all virtual machines: {}", resourceGroupInstanceIdsMap.getKey());
             try {
-                virtualMachines.putAll(getVirtualMachinesByName(azureClient,
+                virtualMachines.putAll(getVirtualMachinesByNameEmptyAllowed(azureClient,
```
lajosrodek, line 148: Just a note: since this is an intra-class invocation, the `@Retryable` on `getVirtualMachinesByNameEmptyAllowed()` will be ineffective. `getVmsAndVmStatusesFromAzure()` will surely provide retries (though with a different `maxAttempts` setting) , but e.g. `getVmsAndVmStatusesFromAzureWithoutRetry()` will not exhibit any retries at all.

lnardai: Yes that correct and in this context this might be misleading, but it's called by the context of the syncer where we don't want any retries.
Might be better to make this one private and remove the annotation.

horadla23 (approved):
LGTM

PR #10879, opened 2021-06-09
CB-12843 Make sure we have PKs on every CB table, test PR, should fail

bbihari:


lnardai:
LGTM! Closing this as verified.

PR #10890, opened 2021-06-10
CB-12965. Create individual API for env/DL/DH installation flow progress

oleewere:
details:
- separate progress service from new operations (creaate progress service for every remaining progress api)
- add operation endpoints + controllers
- operation api always get back the latest operation (not by type, can be done later)
- add operationType field for flowlogs -> represents named operations for flows or flow chains
- create operation view models that holds detailed and simple flow progress results
- update flow2 handlers for supporting new operation type
- add new query with an operationid in the response (operationid is flowid or flowchain - based on the flow is in chain or not)
- operation type of flow chains will override the operation type of flows

Additional change: CB-13103. Operation API: Calculate progress percentage from flow time statistics
details:
- create new flow operation statistics table
- store only not UNKNOWN operations (per cloud platform)
- calculate average from the statistics (to guess a flow/operation time) - or if no duration history -> use 12 minutes
- provide additional payload information (resourceCrn/envCrn/cloudPlatform) by a new object: PayloadContext
- store flow stats + running flows/flowchains in memory (cache) -> use this if there is anything in the cache + the type of running operation is not UNKNOWN
- add additional cache cleanup to the flow cleanup job (it should not have old flows in the cache, but in case of any newly added features to flows, it's possible - with an unexpected exception - a cache remove operation can be skipped)

### DIFF
--- a/common/src/main/java/com/sequenceiq/cloudbreak/common/event/PayloadContext.java
+++ b/common/src/main/java/com/sequenceiq/cloudbreak/common/event/PayloadContext.java
@@ -1,0 +1,47 @@
+package com.sequenceiq.cloudbreak.common.event;
+
+import java.io.Serializable;
+
+public class PayloadContext implements Serializable {
+
+    private final String resourceCrn;
+
+    private final String environmentCrn;
+
+    private final String cloudPlatform;
+
+    private PayloadContext(String resourceCrn, String environmentCrn, String cloudPlatform) {
+        this.resourceCrn = resourceCrn;
+        this.environmentCrn = environmentCrn;
+        this.cloudPlatform = cloudPlatform;
+    }
+
+    public static PayloadContext create(String resourceCrn, String cloudPlatform) {
+        return new PayloadContext(resourceCrn, null, cloudPlatform);
+    }
+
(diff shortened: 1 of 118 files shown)


datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/ProgressEndpoint.java
```diff
@@ -34,4 +35,11 @@
     @ApiOperation(value = "List recent flow operations progress details for resource by resource crn", produces = "application/json",
             notes = "Flow operations progress", nickname = "getSdxFlowLogsProgressByResourceCrn")
     SdxProgressListResponse getFlowLogsProgressByResourceCrn(@PathParam("resourceCrn") String resourceCrn);
+
+    @GET
+    @Path("/resource/crn/{resourceCrn}/provision")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Get provision flow operation progress details for resource by resource crn\"", produces = "application/json",
```
daszabo, line 42: Copy paste error \"

daszabo (changes requested):

daszabo (approved):

PR #10892, opened 2021-06-10
CB-13013 WaitOperationChecker exitWaiting does not throw exception in case of isCreateFailed

aszegedi:
Azure API E2E DistroXEncryptedVolumeTest/testCreateDistroXWithEncryptedVolume is does not failing even the `status=CREATE_FAILED`.

I found that the [exitWaiting method](https://github.com/hortonworks/cloudbreak/blob/CB-2.44.0/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/wait/service/WaitOperationChecker.java#L67-L70) returns only a boolean:
```
if (waitObject.isCreateFailed()) {
     LOGGER.info("'{}' the polled resource entered into creation failed state. Exit waiting!", name);
     return true;
}
```
However here a `throw new TestFailException` would be expected as well.


horadla23 (approved):
LGTM

PR #10894, opened 2021-06-10
[CB-12976] Add Resource Builder for GCP instance Groups

yg-gd:
Part of the first step for GCP loadbalancers, defining instance groups to be used to define service pools

Unit testing here, more verification coming.

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
@@ -25,6 +25,7 @@ import com.google.api.client.auth.oauth2.TokenResponseException;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.compute.Compute;
 import com.google.api.services.compute.Compute.Addresses;
+import com.google.api.services.compute.Compute.InstanceGroups.AddInstances;
 import com.google.api.services.compute.Compute.Instances.Get;
 import com.google.api.services.compute.Compute.Instances.Insert;
 import com.google.api.services.compute.model.AccessConfig;
@@ -32,6 +33,9 @@ import com.google.api.services.compute.model.AttachedDisk;
 import com.google.api.services.compute.model.CustomerEncryptionKey;
 import com.google.api.services.compute.model.CustomerEncryptionKeyProtectedDisk;
 import com.google.api.services.compute.model.Instance;
+import com.google.api.services.compute.model.InstanceGroupList;
+import com.google.api.services.compute.model.InstanceGroupsAddInstancesRequest;
+import com.google.api.services.compute.model.InstanceReference;
 import com.google.api.services.compute.model.InstancesStartWithEncryptionKeyRequest;
 import com.google.api.services.compute.model.Metadata;
 import com.google.api.services.compute.model.Metadata.Items;
@@ -89,6 +93,8 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
     private static final int MAX_TAG_LENGTH = 63;
 
     private static final int ORDER = 3;
+
+    private static final String INSTANCE_REFERENCE_URI = "https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/%s";
 
     @Inject
     private CustomGcpDiskEncryptionCreatorService customGcpDiskEncryptionCreatorService;
@@ -209,10 +215,53 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
             Operation operation = insert.execute();
             verifyOperation(operation, buildableResource);
             updateDiskSetWithInstanceName(auth, computeResources, instance);
+            assignToExistingInstanceGroup(context, group, instance, buildableResource);
             return singletonList(createOperationAwareCloudResource(buildableResource.get(0), operation));
         } catch (GoogleJsonResponseException e) {
             throw new GcpResourceException(checkException(e), resourceType(), buildableResource.get(0).getName());
         }
+    }
+
+    /**
+     * if a InstanceGroup was created in GCP for this Instance's group, then after creating this compute instance assign it to that group.
+     * the group in general can be used to manage all instances in the same group, specifiaclly one way is used to assign to a load balancer.
+     * also provides aggrigated monitoring
+     */
+    private void assignToExistingInstanceGroup(GcpContext context,
+            Group group, Instance instance, List<CloudResource> buildableResource) throws IOException {
+        Compute compute = context.getCompute();
+        String projectId = context.getProjectId();
+        String zone = context.getLocation().getAvailabilityZone().value();
+
+        List<CloudResource> instanceGroupResources = filterResourcesByType(context.getGroupResources(group.getName()), ResourceType.GCP_INSTANCE_GROUP);
+        if (!instanceGroupResources.isEmpty()  && doesGcpInstanceGroupExist(compute, projectId, zone, instanceGroupResources.get(0))) {
+            LOGGER.info("adding instance {} to group {} in project {}", instance.getName(), group.getName(), projectId);
+            InstanceGroupsAddInstancesRequest request = createAddInstancesRequest(instance, projectId, zone);
+            AddInstances addInstances = compute.instanceGroups().addInstances(projectId,
+                    zone, instanceGroupResources.get(0).getName(), request);
+            try {
+                Operation execute = addInstances.execute();
+                verifyOperation(execute, buildableResource);
+            } catch (GoogleJsonResponseException e) {
+                LOGGER.error("Error in Google response, unable to add instance {} to group {} : {} for {}",
+                        instance.getName(), group.getName(), e.getMessage(), e.getDetails().getMessage());
+                throw new GcpResourceException(checkException(e), resourceType(), buildableResource.get(0).getName());
+            }
+        } else {
+            LOGGER.info("skipping group assignment {} doesn't exist in project {}", group.getName(), projectId);
+        }
+
+    }
+
+    private InstanceGroupsAddInstancesRequest createAddInstancesRequest(Instance instance, String projectId, String zone) {
+        return new InstanceGroupsAddInstancesRequest().setInstances(List.of(new InstanceReference()
+                .setInstance(String.format(INSTANCE_REFERENCE_URI,
+                        projectId, zone, instance.getName()))));
+    }
+
+    private boolean doesGcpInstanceGroupExist(Compute compute, String projectId, String zone, CloudResource instanceGroupResource) throws IOException {
+        InstanceGroupList instanceGroupList = compute.instanceGroups().list(projectId, zone).execute();
+        return instanceGroupList.getItems().stream().anyMatch(item -> item.getName().equals(instanceGroupResource.getName()));
     }
 
     private void addTag(List<String> tagList, String actualTag) {
@@ -359,7 +408,7 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
         return ORDER;
     }
 
-    private Collection<AttachedDisk> getBootDiskList(Iterable<CloudResource> resources, String projectId, String zone) {
+    private Collection<AttachedDisk> getBootDiskList(List<CloudResource> resources, String projectId, String zone) {
         Collection<AttachedDisk> listOfDisks = new ArrayList<>();
         for (CloudResource resource : filterResourcesByType(resources, ResourceType.GCP_DISK)) {
             listOfDisks.add(createDisk(projectId, true, resource.getName(), zone, true));
@@ -367,7 +416,7 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
         return listOfDisks;
     }
 
-    private Collection<AttachedDisk> getAttachedDisks(Iterable<CloudResource> resources, String projectId) {
+    private Collection<AttachedDisk> getAttachedDisks(List<CloudResource> resources, String projectId) {
         Collection<AttachedDisk> listOfDisks = new ArrayList<>();
         for (CloudResource resource : filterResourcesByType(resources, ResourceType.GCP_ATTACHED_DISKSET)) {
             VolumeSetAttributes volumeSetAttributes = resource.getParameter(CloudResource.ATTRIBUTES, VolumeSetAttributes.class);
@@ -390,8 +439,9 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
         return attachedDisk;
     }
 
-    private List<NetworkInterface> getNetworkInterface(GcpContext context, Iterable<CloudResource> computeResources, Group group,
+    private List<NetworkInterface> getNetworkInterface(GcpContext context, List<CloudResource> computeResources, Group group,
             CloudStack stack, CloudInstance instance) throws IOException {
+
         boolean noPublicIp = context.getNoPublicIp();
         String projectId = context.getProjectId();
         Location location = context.getLocation();
@@ -436,14 +486,10 @@ public class GcpInstanceResourceBuilder extends AbstractGcpComputeBuilder {
         return String.format("https://www.googleapis.com/compute/v1/projects/%s/global/networks/%s", sharedProjectId, customNetworkId);
     }
 
-    private List<CloudResource> filterResourcesByType(Iterable<CloudResource> resources, ResourceType resourceType) {
-        List<CloudResource> resourcesTemp = new ArrayList<>();
-        for (CloudResource resource : resources) {
-            if (resourceType.equals(resource.getType())) {
-                resourcesTemp.add(resource);
-            }
-        }
-        return resourcesTemp;
+    private List<CloudResource> filterResourcesByType(List<CloudResource> resources, ResourceType resourceType) {
+        return Optional.ofNullable(resources).orElseGet(List::of).stream()
+                .filter(resource -> resourceType.equals(resource.getType()))
+                .collect(Collectors.toList());
     }
 
     private CloudVmInstanceStatus stopStart(GcpContext context, AuthenticatedContext auth, CloudInstance instance, boolean stopRequest) {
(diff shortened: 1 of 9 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/AbstractGcpGroupBuilder.java
```diff
@@ -16,4 +21,16 @@
...
+    protected CloudResource doOperationalRequest(CloudResource resource, ComputeRequest<Operation> request) throws IOException {
+        try {
+            Operation operation = request.execute();
+            if (operation.getHttpErrorStatusCode() != null) {
+                throw new GcpResourceException(operation.getHttpErrorMessage(), resourceType(), resource.getName());
+            }
+            return createOperationAwareCloudResource(resource, operation);
+        } catch (GoogleJsonResponseException e) {
+            exceptionHandler(e, resource.getName(), resourceType());
+            throw new GcpResourceException(checkException(e), resourceType(), resource.getName());
```
hreeve-cloudera, line 33: There's a case in `exceptionHandler` that does not thrown an exception, specifically the case where the resource was not found. Previously `GcpFirewallInResourceBuilder.delete` was only calling `exceptionHandler`, which means that in the case the resource wasn't found no exception was thrown, because you don't need to delete a resource that doesn't exist. With this change, that case will always through a `GcpResourceException`. I don't think that's a valid change because we're introduction an exception into a case that's actually not an error case.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
+import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
+import com.sequenceiq.cloudbreak.cloud.model.CloudResourceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.Group;
+import com.sequenceiq.cloudbreak.cloud.model.Location;
+import com.sequenceiq.cloudbreak.cloud.model.Network;
+import com.sequenceiq.cloudbreak.cloud.model.Security;
+import com.sequenceiq.common.api.type.ResourceType;
+
+public class GcpInstanceGroupResourceBuilder extends AbstractGcpGroupBuilder {
```
hreeve-cloudera, line 36: Maybe we should have a Javadoc here talking about the concept of a GCP intance group vs. a Cloudbreak model instance group, for anyone who comes later and doesn't have that context.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public CloudResource build(GcpContext context,
+            AuthenticatedContext auth, Group group,
+            Network network, Security security, CloudResource resource) throws Exception {
+
+        Insert insert = context.getCompute().instanceGroups().insert(context.getProjectId(),
+                context.getLocation().getAvailabilityZone().value(), new InstanceGroup().setName(resource.getName()));
+        CloudResource instanceGroupCloudResource = doOperationalRequest(resource, insert);
+
+        if (group.getInstances().isEmpty()) {
```
hreeve-cloudera, line 55: Is this something we should check before we make the GCP API call, for efficiency?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+import com.sequenceiq.cloudbreak.cloud.model.CloudResourceStatus;
+import com.sequenceiq.cloudbreak.cloud.model.Group;
+import com.sequenceiq.cloudbreak.cloud.model.Location;
+import com.sequenceiq.cloudbreak.cloud.model.Network;
+import com.sequenceiq.cloudbreak.cloud.model.Security;
+import com.sequenceiq.common.api.type.ResourceType;
+
+public class GcpInstanceGroupResourceBuilder extends AbstractGcpGroupBuilder {
+
+    private static final int ORDER = 1;
```
hreeve-cloudera, line 38: For my reference, what is ORDER doing here, and what does a value of 1 mean?

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilderTest.java
```diff
@@ -0,0 +1,169 @@
...
+    private GcpResourceNameService resourceNameService;
+
+    @Mock
+    private GcpStackUtil gcpStackUtil;
+
+    @InjectMocks
+    private GcpInstanceGroupResourceBuilder underTest;
+
+    @Test
+    public void testDeleteWhenEverythingGoesFine() throws Exception {
```
hreeve-cloudera, line 52: Do we have test cases for when things don't go fine?

hreeve-cloudera (changes requested):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
yg-gd: The previous call is creating the group. We then check that there are instances before adding any, if a group can exist with instances in it then this is where we check for that.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
yg-gd: These builders have a defined order in which they create thing. On the Compute side of things this is so that we've created the disk before the instance tries to attach it. This order in which the group builders doesn't affect much.

yg-gd:
There's one issue here that I've just noticed. The group builder are called before the compute builders. The existing case for the group resource, the network defined tag permissions this is desired. You'd want to create the tag before passing it in during instance creation. 
However for instance groups, we'd want the opposite. We need the instances to have already been created, before we try to add them to the group. I'll need to keep thinking on how to handle this.

yg-gd:
Screen Shot 2021-06-24 at 10 37 39
After running locally, instance groups are created in GCP for each group on the stack and instances are assigned to them

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/AbstractGcpGroupBuilder.java
yg-gd: Yes, Changed is back to keeps deletes as a special case. and kept the pattern in the new builder

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
yg-gd: Javadoc added

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilderTest.java
yg-gd: I've added a new test for just the updates. 

I have still yet to add a test for errors on the API during the build

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public CloudResourceStatus update(GcpContext context, AuthenticatedContext auth, Group group, Network network, Security security, CloudResource resource) {
+        return null;
+    }
+
+    /*
+     * compute difference between current Group and GCP Instance Group, remove any no longer needed then add new
+     */
+    protected void updateInstanceList(GcpContext context, AuthenticatedContext auth,
```
brycederriso, line 78: **suggestion:** I think this can actually be package-private as well, since the test class is in the same package.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+            InstanceGroupsListInstances listInstancesResponse = listInstances.execute();
+            if (listInstancesResponse.getItems() != null) {
+                for (InstanceWithNamedPorts item : listInstancesResponse.getItems()) {
+                    existingURLs.add(item.getInstance());
+                }
+            }
+        } catch (GoogleJsonResponseException e) {
+            throw new GcpResourceException(checkException(e), resourceType(), resource.getName());
+        }
+        Set<String> removed = Sets.difference(existingURLs, requestedURLs);
```
brycederriso, line 105: **praise:** TIL about `Sets.difference`, neat!

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -209,12 +212,35 @@
...
             verifyOperation(operation, buildableResource);
             updateDiskSetWithInstanceName(auth, computeResources, instance);
+            assignToInstanceGroup(context, auth, group, instance, buildableResource);
             return singletonList(createOperationAwareCloudResource(buildableResource.get(0), operation));
         } catch (GoogleJsonResponseException e) {
             throw new GcpResourceException(checkException(e), resourceType(), buildableResource.get(0).getName());
         }
     }
 
+    protected void assignToInstanceGroup(GcpContext context, AuthenticatedContext auth,
```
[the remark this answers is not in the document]
yg-gd: Yep, this and the other could have stricter access

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -209,12 +212,35 @@
             Operation operation = insert.execute();
             verifyOperation(operation, buildableResource);
             updateDiskSetWithInstanceName(auth, computeResources, instance);
+            assignToInstanceGroup(context, group, instance, buildableResource);
```
hreeve-cloudera, line 215: We're calling this for all instances? I don't see any references to what instances are behind the load balancer here.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -209,12 +212,35 @@
...
     }
 
+    private void assignToInstanceGroup(GcpContext context,
+            Group group, Instance instance, List<CloudResource> buildableResource) throws IOException {
+        Compute compute = context.getCompute();
+        String projectId = context.getProjectId();
+        Location location = context.getLocation();
+
+        List<InstanceReference> addReferences = new ArrayList<>();
+        addReferences.add(new InstanceReference().setInstance(String.format("https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/%s",
```
hreeve-cloudera, line 229: This string should be a constant instead of hardcoded here.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -209,12 +212,35 @@
...
             verifyOperation(operation, buildableResource);
             updateDiskSetWithInstanceName(auth, computeResources, instance);
+            assignToInstanceGroup(context, group, instance, buildableResource);
             return singletonList(createOperationAwareCloudResource(buildableResource.get(0), operation));
         } catch (GoogleJsonResponseException e) {
             throw new GcpResourceException(checkException(e), resourceType(), buildableResource.get(0).getName());
         }
     }
 
+    private void assignToInstanceGroup(GcpContext context,
```
hreeve-cloudera, line 222: Can you add a javadoc about what exactly this method is doing? This method also needs more logging and comments in general. There's no explanation for anything being done right now.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
yg-gd: Yes, I've added a check that it does exist before attempting to add it, should we want to, we could not create the groups. however there's probably a small benefit to applying this regardless of load balancing. mostly just for monitoring sake.

hreeve-cloudera: We should mention this to the wider CB team. It can probably just be a comment in the team room. But I know when Azure there was actually a case where Bryce was trying to organize all instances into what are called availability sets, and there was a specific reason it couldn't be done. We should make sure there isn't a similar restriction in Azure. @brycederriso do you remember the details of that?

hreeve-cloudera (approved):
Everything looks good to me. But @yg-gd, please don't merge this in until we've run it by the larger CB team. Reach out to Bryce about the availability set problem he ran into to make sure you won't see a similar problem with the instance groups.

hreeve-cloudera:
@pdarvasi  @lacikaaa This change is part of the work to add GCP load balancers to the data lake. It creates a resource build for GCP instance groups (which are different than CB instance groups) and assigns instances to a GCP instance group. Since you looked over the Azure LB changes, I was hoping you could look over this as well.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
lacikaaa: I don't know about it. maybe @pdarvasi @gergopapi2 or @keyki ?

lacikaaa (changes requested):
please don't forget to squash your commits

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
hreeve-cloudera: Dan and I followed up with Bryce. Bryce was doing something similar where he was putting all instances into availability sets by instance group. @doktoric told us we couldn't do that, because there was a 150 instance limit on availability sets, and at least one customer had an instance group with more than 150 instances. Dan looked into similar limitations for GCP instance groups. There the limitation vary depending on your configuration, but we can have at least 1,000 instances in an instance group. So I think we're okay there, and won't run into the limitation problem we had on Azure.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java
```diff
@@ -209,12 +215,52 @@
...
+
+        List<CloudResource> instanceGroupResources = filterResourcesByType(context.getGroupResources(group.getName()), ResourceType.GCP_INSTANCE_GROUP);
+        if (!instanceGroupResources.isEmpty()  && doesGcpInstanceGroupExist(compute, projectId, zone, instanceGroupResources.get(0))) {
+            LOGGER.info("adding instance {} to group {} in project {}", instance.getName(), group.getName(), projectId);
+            AddInstances addInstances = compute.instanceGroups().addInstances(projectId,
+                    zone,
+                    instanceGroupResources.get(0).getName(),
+                    new InstanceGroupsAddInstancesRequest().setInstances(List.of(new InstanceReference()
+                            .setInstance(String.format(INSTANCE_REFERENCE_URI,
+                            projectId, zone, instance.getName())))));
```
lacikaaa, line 244: this is a bit much inline stuff. I would recommend a bit refactor, like:
```
            InstanceGroupsAddInstancesRequest addInstanceRequest = createAddInstanceRequest(instance, projectId, zone);
            AddInstances addInstances = compute.instanceGroups().addInstances(projectId, zone, instanceGroupResources.get(0).getName(), addInstanceRequest);

File: cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpInstanceResourceBuilder.java [hunk 13]
```diff
@@ -440,9 +486,11 @@ private String getNetworkUrl(String sharedProjectId, String customNetworkId) {
...
     private List<CloudResource> filterResourcesByType(Iterable<CloudResource> resources, ResourceType resourceType) {
         List<CloudResource> resourcesTemp = new ArrayList<>();
-        for (CloudResource resource : resources) {
-            if (resourceType.equals(resource.getType())) {
-                resourcesTemp.add(resource);
+        if (resources != null) {
+            for (CloudResource resource : resources) {
+                if (resourceType.equals(resource.getType())) {
+                    resourcesTemp.add(resource);
+                }
```
lacikaaa, line 493: I would suggest to use stream instead, bit more compact and readable:
```
private List<CloudResource> filterResourcesByType(List<CloudResource> resources, ResourceType resourceType) {
        return Optional.ofNullable(resources).orElseGet(List::of).stream()
                .filter(resource -> resourceType.equals(resource.getType()))
    }

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
lacikaaa: still no logs, why is this resolved?

lacikaaa (changes requested):

lacikaaa (approved):

PR #10903, opened 2021-06-11
CB-13047 Add instance storage volume count field to Template and fill…

nadamdb:
… out during provisioning

### DIFF
--- a/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
+++ b/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
@@ -2,6 +2,7 @@ package com.sequenceiq.cloudbreak.cloud.aws;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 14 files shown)


cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
```diff
@@ -206,4 +212,25 @@ private void addKnownInstance(CloudInstance cloudInstance, Multimap<String, Inst
         }
         return cloudLoadBalancerMetadata;
     }
+
+    @Override
+    public Map<String, Integer> collectInstanceStorageCount(AuthenticatedContext ac, List<String> instanceTypes) {
```
akanto, line 217: Instead of Map<String, Int> can we return a class that wraps the Map<String, Int> with helper methods? That would be easier to extend in the future.

nadamdb: Done

akanto (approved):

PR #10908, opened 2021-06-14
CB-12485 crn region needs to be configurable

horadla23:

### DIFF
--- a/auth-connector/build.gradle
+++ b/auth-connector/build.gradle
@@ -34,6 +34,7 @@ dependencies {
   testCompile group: 'org.mockito',                     name: 'mockito-core',                   version: mockitoVersion
   testImplementation group: 'org.assertj',              name: 'assertj-core',                   version: assertjVersion
   testImplementation group: 'org.hamcrest',             name: 'hamcrest',                       version: hamcrestVersion
+  testCompile project(path: ':common', configuration: 'tests')
 }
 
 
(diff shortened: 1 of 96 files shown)


horadla23:
(sidenote: it seems code owner feature for authorization module is working :) )

common/src/main/java/com/sequenceiq/cloudbreak/auth/crn/CrnGeneratorService.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.cloudbreak.auth.altus.Crn;
+import com.sequenceiq.cloudbreak.auth.altus.CrnResourceDescriptor;
+
+@Service
+@ConfigurationProperties("crn")
+public class CrnGeneratorService {
```
lnardai, line 13: How about RegionAwareCrnGenerator? I would like to convey what is the usecase for this service vs. the regular Crn.Builder.

horadla23: I like it, I will change the class name

lnardai (approved):
Looks good! 
Only question in my mind is how do we enforce people to use the 'RegionAwareCrn' builder. It would be nice to restrict, at least in application code that only the proper CRN with correct region should be used. 
We should at least make it hard for people to create Crn's with the default region and partition. Any ideas on that?

horadla23:
retest this please

PR #10921, opened 2021-06-15
CB-13081 DL upgrade on azure: package version check still throws

gergopapi2:
Previous CB-12684 disabled package version check during upgrade only in one part. There was another invocation of check package versions. This commit turns of checking package versions also for that other invocation


lnardai (approved):
LGTM! wait for one more approval from our team.

horadla23 (approved):

PR #10923, opened 2021-06-15
CB-13080 check if the target runtime version is supported for upgrade

keyki:
It happened in production that a version (7.2.10) was offered as an upgrade
candidate, but the 7.2.10 version was not supported yet. This is because
the upgrade filter did not check whether a version is supported or not. It
only checks the upgrade matrix. However, there is a separate env variable
called cb.runtimes.latest that tells what is the latest supported version.
As part of this PR the upgrade filters will consider this variable also
when offering the new versions.

@lnardai @akanto this is a draft PR that we could deploy as a hotfix for 2.44.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
@@ -7,6 +7,7 @@ import org.springframework.stereotype.Service;
 import com.sequenceiq.cloudbreak.api.endpoint.v4.common.StackType;
 import com.sequenceiq.cloudbreak.cloud.model.catalog.Image;
 import com.sequenceiq.cloudbreak.cloud.model.catalog.ImagePackageVersion;
+import com.sequenceiq.cloudbreak.service.runtimes.SupportedRuntimes;
 import com.sequenceiq.cloudbreak.service.upgrade.image.ImageFilterParams;
 import com.sequenceiq.cloudbreak.service.upgrade.matrix.UpgradeMatrixService;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
```diff
@@ -22,12 +23,16 @@
...
+    private SupportedRuntimes supportedRuntimes;
+
     public boolean permitCmUpgrade(ImageFilterParams imageFilterParams, Image image) {
         return permitUpgrade(imageFilterParams, image, ImagePackageVersion.CM, ImagePackageVersion.CM_BUILD_NUMBER, false);
     }
 
     public boolean permitStackUpgrade(ImageFilterParams imageFilterParams, Image image) {
-        return permitUpgrade(imageFilterParams, image, ImagePackageVersion.STACK, ImagePackageVersion.CDH_BUILD_NUMBER, checkUpgradeMatrix(imageFilterParams));
+        return permitUpgrade(imageFilterParams, image, ImagePackageVersion.STACK, ImagePackageVersion.CDH_BUILD_NUMBER, checkUpgradeMatrix(imageFilterParams))
+                && isRuntimeVersionSupported(image);
```
tiborpopovics, line 35: Can you please flip the two parts of the `AND` statement to run the `isRuntimeVersionSupported ` first?

pdarvasi (approved):

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/UpgradePermissionProvider.java
keyki: done

tiborpopovics (approved):

PR #10927, opened 2021-06-15
CDPCP-4714. Add enabled state to FMS user model

handavid:
Previous commits added the fields that represent whether an
actor is disabled to the models retrieved from the UMS and
from IPA. This commit now propagates those fields to the FMS's
internal user model for comparing users between the UMS and IPA.
The code for calculating the difference between the UMS and IPA
states was updated to not use the FmsUser.equals() method since
that method now includes the new field. The unit tests were
updated to ensure that the new calculation of users to add to
IPA works correctly.


shubhangiKishore (commented):
lgtm!

PR #10930, opened 2021-06-16
CB-12939 Always request instance storage on AWS

daszabo:
Instance storage/ephemeral storage should be requested always, regardless of storage type selected.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/InstanceStoreMetadata.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/InstanceStoreMetadata.java
@@ -2,28 +2,33 @@ package com.sequenceiq.cloudbreak.cloud.model;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 public class InstanceStoreMetadata {
 
-    private Map<String, Integer> instaceStoreCountMap = new HashMap<>();
+    private Map<String, VolumeParameterConfig> instaceStoreConfigMap = new HashMap<>();
 
     public InstanceStoreMetadata() {
     }
 
-    public InstanceStoreMetadata(Map<String, Integer> instaceStoreCountMap) {
-        this.instaceStoreCountMap = instaceStoreCountMap;
+    public InstanceStoreMetadata(Map<String, VolumeParameterConfig> instaceStoreConfigMap) {
+        this.instaceStoreConfigMap = Optional.ofNullable(instaceStoreConfigMap).orElse(new HashMap<>());
     }
 
-    public void addInstanceStoreCountToInstanceType(String instanceType, Integer instanceStoreCount) {
-        instaceStoreCountMap.put(instanceType, instanceStoreCount);
+    public void addInstanceStoreConfigToInstanceType(String instanceType, VolumeParameterConfig instanceStoreConfig) {
+        instaceStoreConfigMap.put(instanceType, instanceStoreConfig);
     }
 
     public Integer mapInstanceTypeToInstanceStoreCount(String instanceType) {
-        return instaceStoreCountMap.get(instanceType);
+        return instaceStoreConfigMap.getOrDefault(instanceType, VolumeParameterConfig.EMPTY).maximumNumber();
     }
 
     public Integer mapInstanceTypeToInstanceStoreCountNullHandled(String instanceType) {
-        Integer instanceStoreCount = instaceStoreCountMap.get(instanceType);
+        Integer instanceStoreCount = instaceStoreConfigMap.getOrDefault(instanceType, VolumeParameterConfig.EMPTY).maximumNumber();
         return instanceStoreCount != null ? instanceStoreCount : 0;
+    }
+
+    public Map<String, VolumeParameterConfig> getInstaceStoreConfigMap() {
+        return instaceStoreConfigMap;
     }
 }
--- a/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
+++ b/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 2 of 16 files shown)


nadamdb (approved):

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/provision/service/StackCreationService.java
```diff
@@ -279,14 +284,45 @@ public void setInstanceStoreCount(StackContext stackContext) {
...
+                    .map(toVolumeTemplate(template))
+                    .ifPresent(volumeTemplate -> template.getVolumeTemplates().add(volumeTemplate));
+        }
+    }
+
+    private Function<VolumeParameterConfig, VolumeTemplate> toVolumeTemplate(Template template) {
+        return config -> {
+            LOGGER.debug("Adding instance storage to template. Template id: {}, instane type: {}", template.getId(), template.getInstanceType());
+            VolumeTemplate volumeTemplate = new VolumeTemplate();
+            volumeTemplate.setVolumeType("ephemeral");
```
lacikaaa, line 313: `"ephemeral"` is present as a magic string in several places. Could you refactor this to have as a constant somewhere?

cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
```diff
@@ -215,22 +221,14 @@ private void addKnownInstance(CloudInstance cloudInstance, Multimap<String, Inst
...
-                        instanceTypeInfo.getInstanceStorageInfo().getDisks() != null ?
-                        instanceTypeInfo.getInstanceStorageInfo().getDisks().stream().map(DiskInfo::getCount).mapToInt(Integer::intValue).sum() : 0 : 0);
-            }
-            return instanceStoreMetadata;
-        } catch (Exception e) {
-            LOGGER.warn("Failed to describe instance types: {}", instanceTypes, e);
-            throw new CloudConnectorException(e.getMessage(), e);
-        }
+        Location location = ac.getCloudContext().getLocation();
+        CloudVmTypes cloudVmTypes = awsPlatformResources.virtualMachines(ac.getCloudCredential(), location.getRegion(), Map.of());
```
biharitomi, line 225: It's a little bit defensive programming, but as I checked that AwsPlatformResources service doesn't really care about exceptions, so I would not remove the exception handling from the original code 😸

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/provision/service/StackCreationService.java
```diff
@@ -279,14 +284,45 @@ public void setInstanceStoreCount(StackContext stackContext) {
...
+            volumeTemplate.setVolumeCount(config.maximumNumber());
+            volumeTemplate.setVolumeSize(config.maximumSize());
+            volumeTemplate.setUsageType(VolumeUsageType.GENERAL);
+            volumeTemplate.setTemplate(template);
+            return volumeTemplate;
+        };
+    }
+
+    private boolean isTemplateSetUpForInstanceStorage(Template template) {
+        return template.getVolumeTemplates().stream().anyMatch(volumeTemplate -> "ephemeral".equalsIgnoreCase(volumeTemplate.getVolumeType()));
```
biharitomi, line 323: Can we have a static variable or enum for ephemeral volume type?

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/InstanceStoreMetadata.java
```diff
@@ -5,25 +5,29 @@
...
-    private Map<String, Integer> instaceStoreCountMap = new HashMap<>();
+    private Map<String, VolumeParameterConfig> instaceStoreConfigMap = new HashMap<>();
 
     public InstanceStoreMetadata() {
     }
 
-    public InstanceStoreMetadata(Map<String, Integer> instaceStoreCountMap) {
-        this.instaceStoreCountMap = instaceStoreCountMap;
+    public InstanceStoreMetadata(Map<String, VolumeParameterConfig> instaceStoreConfigMap) {
+        this.instaceStoreConfigMap = instaceStoreConfigMap;
```
biharitomi, line 14: I would add a null check here, to avoid null issues when `getInstaceStoreConfigMap()` is called. I think we should initialise an empty map if constructor is called explicitly with ` null`.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/provision/service/StackCreationService.java
```diff
@@ -279,14 +284,45 @@ public void setInstanceStoreCount(StackContext stackContext) {
...
+                templateService.savePure(template);
             }
         }
     }
 
+    private void addInstanceStorageVolumeTemplateToTemplate(InstanceStoreMetadata instanceStoreMetadata, InstanceGroup ig, Template template) {
+        if (!isTemplateSetUpForInstanceStorage(template)) {
+            LOGGER.debug("Instance storage was not requested. Attempting to add them implicitly. " +
+                    "Group name: {}, instance type: {}", ig.getGroupName(), template.getInstanceType());
+            Optional.ofNullable(instanceStoreMetadata.getInstaceStoreConfigMap().get(template.getInstanceType()))
```
biharitomi, line 303: If `getInstaceStoreConfigMap()` could not be null I am fine with this call, but otherwise null check would needed here. 😸

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/provision/service/StackCreationService.java
```diff
@@ -279,14 +284,45 @@ public void setInstanceStoreCount(StackContext stackContext) {
...
+                    .map(toVolumeTemplate(template))
+                    .ifPresent(volumeTemplate -> template.getVolumeTemplates().add(volumeTemplate));
+        }
+    }
+
+    private Function<VolumeParameterConfig, VolumeTemplate> toVolumeTemplate(Template template) {
+        return config -> {
+            LOGGER.debug("Adding instance storage to template. Template id: {}, instane type: {}", template.getId(), template.getInstanceType());
```
biharitomi, line 311: I would log this on `info` level as it is a fact that we are setting the volume type for the template.

biharitomi (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/provision/service/StackCreationService.java
doktoric: can you please also cover this logic with unit tests?

biharitomi (approved):

PR #10938, opened 2021-06-16
CB-12879: Fix Azure with SPN issue.

brycederriso:
Closes [CB-12879](https://jira.cloudera.com/browse/CB-12879).

* Add tests for implementations of EnvironmentBaseNetworkConverter
* Add tests for GcpEnvironmentNetworkConverterTest.
* Add tests for AwsEnvironmentNetworkConverterTest.
* Add tests for AzureEnvironmentNetworkConverter.
* Extract getCloudSubnet method.
* Extract buildAttributes and attachEndpointGatewaySubnet methods.
* Override attachEndpointGatewaySubnet for Azure case.
* Add comments to explain the AzureEnvironmentNetworkConverter

Tested by setting up an Azure data lake in an environment with `--no-use-public-ip`, and forcing the `PublicEndpointAccessGateway.ENABLED` check in `EnvironmentBaseNetworkConverter.java` to be true.

See attached diff for a patch that can be used to replicate the test.
[cb-12879-testing.txt](https://github.com/hortonworks/cloudbreak/files/6664451/cb-12879-testing.txt)


core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/environment/network/EnvironmentBaseNetworkConverter.java
```diff
@@ -66,5 +49,43 @@ public Network convertToLegacyNetwork(EnvironmentNetworkResponse source, String
...
+     * @param source contains source information to retrieve subnets from
+     * @param attributes a Map which we put the gateway subnet ID in
+     */
+    protected void attachEndpointGatewaySubnet(EnvironmentNetworkResponse source, Map<String, Object> attributes, CloudSubnet cloudSubnet) {
+        Optional<CloudSubnet> endpointGatewaySubnet = subnetSelector.chooseSubnetForEndpointGateway(source, cloudSubnet.getId());
+        if (endpointGatewaySubnet.isPresent()) {
+            attributes.put(ENDPOINT_GATEWAY_SUBNET_ID, endpointGatewaySubnet.get().getId());
+        } else {
+            throw new BadRequestException("Could not find public subnet in availability zone: " + cloudSubnet.getAvailabilityZone());
+        }
```
hreeve-cloudera, line 81: I think you could also change this logic to use the `orElseThrow()` like you did in `getCloudSubnet`.

hreeve-cloudera (approved):
Looks good to me.

brycederriso:
I wasn't able to completely test this locally. I did set up an environment, and the infrastructure for the data lake.

However, my DL never reported back to CB that it was created successfully.

This looks to be an issue with my local CB/CBD set up, as the same issue showed up when I tested when running off of master.

My changes do allow Azure to set up the VMs and load balancer, so we're going to merge this and see if it works.
@hreeve-cloudera

PR #10946, opened 2021-06-21
[CB-13154] create Entitlement for Gcp loadbalancers

yg-gd:
Entitlement to be used for gating GCP loadbalancer changes
tested with entitlement services unit tests


yg-gd:
I'm not entirely a fan of this entitlement name, open to suggestions.

other than the matching commit to thunderhead, are there other external steps I need to follow to create the entitlement?

hreeve-cloudera:
@yg-gd These changes look good for a new entitlement, although in application.yml we typically want the LB stuff to be `false` by default, since that's for the mock setup and we haven't implemented load balancers for mock environments.

However, we haven't been doing cloud provider specific entitlements for the datalake LB. We have the `CDP_DATA_LAKE_LOAD_BALANCER` entitlement, which all load balancers are gated behind until they're GA. When the feature goes GA for a cloud provider (only AWS so far), that cloud provider is checked in the `LoadBalancerConfigService.isLoadBalancerEntitlementRequiredForCloudProvider()` method. So you don't need to do anything for the datalake LB entitlement in the initial release. We just need to update that method when it goes GA.

However, we will need a cloud provider specific entitlement for the semi-private network feature once you're working on that. See the `CDP_PUBLIC_ENDPOINT_ACCESS_GATEWAY_AZURE` entitlement. So it looks like you're on the right track for when you get to that.

sxxgrc:
Haley addresses the comments I would have so I would follow that and then this can be merged in.

hreeve-cloudera:
Correction: This should not be merged in because we're not using entitlements this way for the datalake load balancer. This should be abandoned, and when Dan gets to the semi-private network work he can create a new PR for the GCP semi-private network entitlement.

PR #10954, opened 2021-06-22
CB-13088 Set x-cdp-request-id header for cluster-proxy to trace reque…

lturcsanyi:
…sts between CB and CP

### DIFF
--- a/client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/tracing/CmRequestIdProviderInterceptor.java
+++ b/client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/tracing/CmRequestIdProviderInterceptor.java
@@ -1,0 +1,25 @@
+package com.sequenceiq.cloudbreak.cm.client.tracing;
+
+import static com.sequenceiq.cloudbreak.client.RequestIdProvider.REQUEST_ID_HEADER;
+
+import java.io.IOException;
+
+import org.springframework.stereotype.Component;
+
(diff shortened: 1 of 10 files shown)


client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/tracing/CmRequestIdProviderInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final String REQUEST_ID_HEADER = "x-cdp-request-id";
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request originalRequest = chain.request();
+        String requestId = MDC.get(LoggerContextKey.REQUEST_ID.toString());
+        if (requestId == null) {
+            requestId = UUID.randomUUID().toString();
+            LOGGER.debug("Generated requestId for request: {}", requestId);
+        }
```
horadla23, line 30: it may worth to extract this code into some static method/class since I have seen this third times in this PR

horadla23 (approved):

client-cm/src/main/java/com/sequenceiq/cloudbreak/cm/client/tracing/CmRequestIdProviderInterceptor.java
lturcsanyi: You are right, I will extract this.

PR #10957, opened 2021-06-22
CB-13131 Fix for AWS CF trimming whitespaces from string parameters a…

bergerdenes:
…ffecting User Data concatenation

The issue is the following:
The user data script that is sent as parameter value to the AWS CloudFormation
template, is split into more (max 4) chunks.
The split boundary was the space character between the `export` keyword and
the variable name.
After joining back the chunks with Fn::Join in the CF template, it resulted
in an invalid state (the proper variable was not exported), but rather
a local variable is created with the following name and value:
exportCCM_V2_INVERTING_PROXY_HOST=hostname...
This is fixed by splitting the user data chunks between non-whitespace characters.


doktoric (approved):

horadla23 (approved):

horadla23:
@bergerdenes please merge this forward to master

PR #10962, opened 2021-06-23
CB-12971 Make freeipa backup work for GCP environment with log bucket…

cegganesh84:
… path root location already exists in operating system

1. To avoid storage.objects.list permission for logger we were syncing backups to the bucket instead of the object by staging it in a similar folder structure locally.
2. This is a dangerous idea to begin with because if the root location already exists in operating system then it will sync the entire OS system content.
3. In this change we avoid this mistake by checking if the folder already exists and do a cp instead of mv.
4. But this needs a new permission for logger service account which needs to be documented.
5. We keep the old behaviour as is in the normal scenarios for backward compatibility purposes.

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_backup
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/freeipa_backup
@@ -249,23 +249,29 @@ elif [[ "${config[backup_platform]}" = "AZURE" ]]; then
     fi
     remove_local_backups
 elif [[ "${config[backup_platform]}" = "GCP" ]]; then
-    # To avoid storage.objects.list permission for logger we must sync backups to the bucket instead of the object
     BACKUP_PATH="/${BACKUP_LOCATION#*://*/}"
     BACKUP_BUCKET=$(echo "${BACKUP_LOCATION}" | cut -f-3 -d '/')
-    doLog "INFO Copy backup to ${BACKUP_PATH} before moving it to cloud storage"
     if [[ "${BACKUP_PATH}" = / ]]; then
       BACKUP_PATH="/cluster-backups/freeipa"
     elif [[ "${BACKUP_PATH}" != /* ]]; then
       BACKUP_PATH="/${BACKUP_PATH}"
     fi
-    mkdir -p "${BACKUP_PATH}"
-    cp -r "${config[backup_path]}/${BACKUPDIR}" "${BACKUP_PATH}"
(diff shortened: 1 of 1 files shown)


lacikaaa (changes requested):
there is already a storage validation in FMS, like a preflight check, before we start installing and configuring the FreeIPA. I think it should validate if the backup script has to use the new logic and it has the necessary permissions.
Also please file a documentation jira for this new permission and when it is necessary. thanks

cegganesh84:
> I think it should validate if the backup script has to use the new logic and it has the necessary permissions.

Agreed. One trouble in doing that. In the bash script, it is easy to check for preexisting folders. If we have to do it outside the VM then we need to know what folders are preexisting apriori based on the image baking process it could differ (our centos image vs customer custom rhel image for example). I believe it can go in as an additional enhancement.

Yes documentation is part of this jira itself.

lacikaaa:
> If we have to do it outside the VM then we need to know what folders are preexisting apriori based on the image baking process it could differ

This validation is run when the VM is already up, but before we are installing and configuring freeipa. It would still take a few minutes to fail, and it won't fail in request time but at least the customer won't have to wait until the freeipa is installed and configured.
see: https://github.com/hortonworks/cloudbreak/blob/master/freeipa/src/main/java/com/sequenceiq/freeipa/flow/freeipa/provision/action/FreeIpaProvisionActions.java#L100-L115

jamisonbennett:
The cloud storage validation fails after about 5 minutes if there are not permissions. But it is a much faster failure message than waiting 35 minutes for the install to fail due to permissions issues.

cegganesh84:
Ah, I see where the check is https://github.com/hortonworks/cloudbreak/commit/e4fdf4eae35e1e9a83673ed926154d18b15f1919, I confused myself by thinking it is along the lines of DL IAM storage permissions check.

cegganesh84:
All I can tell is when I execute the backup validation shell script it works, that is fails for invalid permission and succeeds for correct permission. But the validation is not kicking in after 2.44 release. No "Completed validating cloud storage" after 21st June https://logs-dev-7x.sre.cloudera.com/_plugin/kibana/goto/906378e35df478da5f706a4580e3bef0?security_tenant=global

lacikaaa:
does it mean that the current validation logic covers the new use case too?

cegganesh84:
@lacikaaa yes, it should work automatically after this issue is fixed CB-13309

lacikaaa (approved):

PR #10969, opened 2021-06-24
CB-13210 Add custom nginx response header

bergerdenes:
Add a response header in our nginx so we can check if the request reached nginx,
or failed between CB - CCM


lacikaaa (changes requested):
please add `always` and validate the change in case of an error

lacikaaa (approved):

PR #10974, opened 2021-06-24
CB-13227 Create separate Ranger repo for each Kudu

attilabukor:
The Real Time DataMart blueprints use the default Ranger service
repository name for Kudu, which is cm_kudu. Each RTDM DataHub is
supposed to have its own Ranger service repo for Kudu.

Kudu CSD recognizes the "{{GENERATED_RANGER_SERVICE_NAME}}" magic string
and uses the CM-generated unique service repo name.


attilabukor:
cc @doktoric

doktoric (approved):
if it is tested then approved. Thanks

PR #10977, opened 2021-06-25
CB-12834: Improve current image selection for upgrade to after a successful runtime upgrade we should offer the same image as an OS upgrade.

tiborpopovics:


core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageProvider.java
```diff
@@ -1,19 +1,73 @@
...
+
     public Image getCurrentImageFromCatalog(String currentImageId, CloudbreakImageCatalogV3 imageCatalog)
             throws CloudbreakImageNotFoundException {
         return imageCatalog.getImages().getCdhImages().stream()
                 .filter(img -> currentImageId.equals(img.getUuid()))
                 .findFirst()
                 .orElseThrow(() -> new CloudbreakImageNotFoundException(String.format("Image not found with id: %s", currentImageId)));
     }
+
+    public com.sequenceiq.cloudbreak.cloud.model.Image getCurrentImageForUpgrade(Long stackId) throws CloudbreakImageNotFoundException {
```
lacikaaa, line 37: if you move this stuff to a new class you could avoid having the package name written everywhere

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageProvider.java
```diff
@@ -1,19 +1,73 @@
...
+
+    private Set<com.sequenceiq.cloudbreak.cloud.model.Image> getImagesFromInstanceMetadata(Long stackId) {
+        return instanceMetaDataService.getNotDeletedInstanceMetadataByStackId(stackId)
+                .stream()
+                .map(image -> {
+                    try {
+                        return image.getImage().get(com.sequenceiq.cloudbreak.cloud.model.Image.class);
+                    } catch (IOException e) {
+                        LOGGER.error("Failed to convert Json to Image", e);
+                        return null;
```
lacikaaa, line 62: wouldn't this cause adding `null` to the set in this case?

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageProvider.java
```diff
@@ -1,19 +1,73 @@
...
         return imageCatalog.getImages().getCdhImages().stream()
                 .filter(img -> currentImageId.equals(img.getUuid()))
                 .findFirst()
                 .orElseThrow(() -> new CloudbreakImageNotFoundException(String.format("Image not found with id: %s", currentImageId)));
     }
+
+    public com.sequenceiq.cloudbreak.cloud.model.Image getCurrentImageForUpgrade(Long stackId) throws CloudbreakImageNotFoundException {
+        com.sequenceiq.cloudbreak.cloud.model.Image currentImageFromStack = imageService.getImage(stackId);
+        Set<com.sequenceiq.cloudbreak.cloud.model.Image> imagesFromInstances = getImagesFromInstanceMetadata(stackId);
+        com.sequenceiq.cloudbreak.cloud.model.Image currentImage = getCurrentImage(currentImageFromStack, imagesFromInstances);
```
keyki, line 40: I would call it something different to avoid confusion with the current images. Something that says different than the current image used for the stack.

core/src/main/java/com/sequenceiq/cloudbreak/service/image/ImageProvider.java
```diff
@@ -1,19 +1,73 @@
...
+        LOGGER.debug("The current image of the stack: {}. The available images on instances: {}, The current image for upgrade: {}",
```
keyki, line 41: The current image for upgrade is not the same as `currentImage.getImageId()` right? It's the image that the instances are running on at the moment.

lacikaaa (approved):

keyki:
Unfortunately, the OS upgrade part fails for me with error coming from AWS:
```
Image change failed with message: Cannot execute method: updateStack. No updates are to be performed. (Service: AmazonCloudFormation; Status Code: 400; Error Code: ValidationError; Request ID: 265d27ee-6e67-440e-9452-e46ef366acfa; Proxy: null)
```

keyki (changes requested):
See https://github.com/hortonworks/cloudbreak/pull/10977#issuecomment-869654590

PR #10978, opened 2021-06-25
CB-13233 Prepare UnassignResourceRoleAction for NOT_FOUND case

aszegedi:
MOW Dev API E2E test build is failing, because of testAddGroupsToEnvironment test is failing with:
```
com.sequenceiq.it.cloudbreak.exception.TestFailException: EXCEEDING UMS ACTION MAX. RETRY (11) ::: UMS action has been failed, because of: NOT_FOUND: Cannot find resource role...
```
So I'm introducing an additional step to the `UnassignResourceRoleAction` with a pre-check for resource role has already been revoked from user.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignResourceRoleUserAction.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignResourceRoleUserAction.java
@@ -27,22 +27,23 @@ public class AssignResourceRoleUserAction extends AbstractUmsAction<UmsTestDto> 
     @Override
     protected UmsTestDto umsAction(TestContext testContext, UmsTestDto testDto, UmsClient client) throws Exception {
         CloudbreakUser user = testContext.getRealUmsUserByKey(userKey);
+        String userCrn = user.getCrn();
         String resourceRole = testDto.getRequest().getRoleCrn();
         String resourceCrn = testDto.getRequest().getResourceCrn();
-        Log.when(LOGGER, format(" Assigning resource role '%s' at resource '%s' for user '%s' ", resourceRole, resourceCrn, user.getCrn()));
-        Log.whenJson(LOGGER, format(" Assign resource role request:%n "), testDto.getRequest());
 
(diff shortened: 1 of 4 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/ums/AssignResourceRoleUserAction.java
```diff
@@ -27,22 +27,21 @@ public AssignResourceRoleUserAction(String userKey) {
...
+            LOGGER.info(format(" Resource role '%s' has already been assigned to user '%s' at resource '%s' ", resourceRole, userCrn, resourceCrn));
+            Log.when(LOGGER, format(" Resource role '%s' has already been assigned to user '%s' at resource '%s' ", resourceRole, userCrn, resourceCrn));
         } else {
-            client.getDefaultClient().assignResourceRole(user.getCrn(), resourceCrn, resourceRole, Optional.of(""));
-            // wait for UmsRightsCache to expire
-            Thread.sleep(7000);
-            LOGGER.info(format(" Resource role '%s' has been assigned at resource '%s' for user '%s' ", resourceRole, resourceCrn, user.getCrn()));
-            Log.when(LOGGER, format(" Resource role '%s' has been assigned at resource '%s' for user '%s' ", resourceRole, resourceCrn, user.getCrn()));
+            client.getDefaultClient().assignResourceRole(userCrn, resourceCrn, resourceRole, Optional.of(""));
+            LOGGER.info(format(" Resource role '%s' has been assigned to user '%s' at resource '%s' ", resourceRole, userCrn, resourceCrn));
```
lnardai, line 43: I'm fine with removing the Thread.sleep from here, but now we should somehow enforce that validation, because now it would be easy to misuse the framework since the assign and verification is not atomic.

```
                .when(umsTestClient.unAssignResourceRole(L0UserKeys.ENV_CREATOR_A))
                .then((tc, dto, client) -> validateAssignedResourceRole(tc, dto, client, userEnvCreatorA, getIamGroupAdminCrn(), false))
```

aszegedi: My bad, these lines were removed mistakenly. So I've reintroduced the removed lines. Thanks for pointer!

sodre90 (approved):

PR #10982, opened 2021-06-28
CDPCP-4714. User sync disables and enables users

handavid:
The UserSyncService now enables and disables users based on their state
in the UMS.

UserDisableOperations are invoked for users that exist in IPA that need to be
disabled. The UserAddOperation is updated to add disabled users to IPA. This happens
as a single operation to avoid any potential issues if user sync fails between
adding a user and disabling a user.

UserEnableOperations are invoked for users that exist in IPA that need to be enabled.

The FreeIpaUserStateProvider was fixed to correctly set the enablement state.
The state is returned as a negative flag (true indicates disabled). The previous
commit introducing this to the model had the flag reversed.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
@@ -22,24 +22,29 @@ import com.sequenceiq.freeipa.service.freeipa.user.UserSyncConstants;
 public class UsersStateDifference {
     private static final Logger LOGGER = LoggerFactory.getLogger(UsersStateDifference.class);
 
-    private ImmutableSet<FmsGroup> groupsToAdd;
+    private final ImmutableSet<FmsGroup> groupsToAdd;
 
-    private ImmutableSet<FmsGroup> groupsToRemove;
+    private final ImmutableSet<FmsGroup> groupsToRemove;
 
-    private ImmutableSet<FmsUser> usersToAdd;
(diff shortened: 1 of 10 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/model/UsersStateDifference.java
```diff
@@ -132,6 +153,42 @@ public static UsersStateDifference forDeletedUser(String deletedUser, Collection
...
+                .map(FmsUser::getName)
+                .collect(Collectors.toSet());
+        ImmutableSet<String> usersToEnable = ImmutableSet
+                .copyOf(Sets.intersection(enabledUmsUsers, disabledIpaUsers));
+
+        LOGGER.info("usersToEnable size = {}", usersToEnable.size());
+        LOGGER.debug("userToEnable = {}", usersToEnable);
+
+        return usersToEnable;
+    }
```
lacikaaa, line 190: these 2 method should use the same logic, and the only difference is whether we are looking for disabled users in freeipa which should be enabled, or enabled users which should be disabled. I think this logic could be in a separate method with parameters. What do you think?

handavid: thanks! they used to need slightly different logic before I update user-add. I'll take a look and factor it out.

handavid: updated!

lacikaaa (approved):

PR #10987, opened 2021-06-29
CB-13246 We are storing non-sensitive data in the attributes field of…

akanto:
… the instancegroup objects. We shall make sure that these are sent o EDH to make it easier to reconstruct how the customer has created the cluster.


nadamdb (approved):

lajosrodek:
I am late to the party, just wondering if `template.getAttributes().getMap()` is prone to NPE. Is `attributes` always non-null?

PR #10988, opened 2021-06-29
CB-12979 introduce a hidden internal flag on the upgrade request to a…

keyki:
…llow internal API requests

A new hidden field is introduced on the `UpgradeV4Request` so we can build features around it
or simply bypass some of the validations. This flag can't be set on the API, but an internal API
call is needed to set it. As part of this PR I've introduced the internal counterparts of the
upgrade APIs for Data Hub. This can be called from the experiences with the internal actor crn.
The current purpose of this field is to bypass the blueprint validation even if it's `DISABLED`.
The scenario is that the COD clusters are disabled for upgrade on the Data Hub page, but the `ODX`
experience would like to drive the upgrade for these clusters so they can call the internal API.
This is to prevent customers to directly upgrade the COD clusters without the `ODX` app knowing
about it.

Tested with curl:
```
curl -X POST \
-H "Content-Type: application/json" \
-H "x-cdp-actor-crn: crn:cdp:iam:us-west-1:altus:user:__internal__actor__" \
-d '{"runtime":"7.2.10"}' \
http://localhost:9091/cb/api/v1/distrox/internal/krisz-wl-1/upgrade\?initiatorUserCrn\=crn:cdp:iam:us-west-1:pepsi:user:khorvath
```

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/request/tags/upgrade/UpgradeV4Request.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/request/tags/upgrade/UpgradeV4Request.java
@@ -1,9 +1,11 @@
 package com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.tags.upgrade;
 
 import java.util.Objects;
+import java.util.StringJoiner;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.sequenceiq.cloudbreak.doc.ModelDescriptions;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.request.InternalUpgradeSettings;
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions.UpgradeModelDescription;
 import com.sequenceiq.cloudbreak.validation.ValidUpgradeRequest;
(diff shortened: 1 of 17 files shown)


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/request/tags/upgrade/UpgradeV4Request.java
```diff
@@ -15,23 +15,26 @@
...
     private Boolean dryRun;
 
     private Boolean replaceVms = Boolean.TRUE;
 
-    @ApiModelProperty(ModelDescriptions.UpgradeModelDescription.SHOW_AVAILABLE_IMAGES)
+    @ApiModelProperty(UpgradeModelDescription.SHOW_AVAILABLE_IMAGES)
     private UpgradeShowAvailableImages showAvailableImages = UpgradeShowAvailableImages.DO_NOT_SHOW;
 
+    @ApiModelProperty(hidden = true)
+    private boolean internal;
```
lacikaaa, line 36: could you replace this with a class which makes it posible to extend for later unforeseen scenarios? boolean is too restrictive and not extendable.
also I think this should be validated. If it's set and not called with internal crn then the request should be rejected

lacikaaa (approved):

PR #10990, opened 2021-06-30
CB-10602 Propogate telemetry metrics from periscope.

smaniraju:
Telemetry events are propogated after triggering a scaling action.
Telemetry events are propogated after autoscaling config is updated.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterCommonService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterCommonService.java
@@ -39,10 +39,11 @@ import com.sequenceiq.periscope.service.ClusterService;
 import com.sequenceiq.periscope.service.EntitlementValidationService;
 import com.sequenceiq.periscope.service.HistoryService;
 import com.sequenceiq.periscope.service.NotFoundException;
+import com.sequenceiq.periscope.service.UsageReportingService;
 import com.sequenceiq.periscope.service.configuration.ClusterProxyConfigurationService;
 
 @Component
-public class AutoScaleClusterCommonService  implements ResourcePropertyProvider {
+public class AutoScaleClusterCommonService implements ResourcePropertyProvider {
     private static final Logger LOGGER = LoggerFactory.getLogger(AutoScaleClusterCommonService.class);
 
     @Inject
@@ -50,6 +51,9 @@ public class AutoScaleClusterCommonService  implements ResourcePropertyProvider 
 
     @Inject
     private HistoryService historyService;
+
+    @Inject
+    private UsageReportingService usageReportingService;
 
     @Inject
     private CloudbreakCommunicator cloudbreakCommunicator;
@@ -111,19 +115,26 @@ public class AutoScaleClusterCommonService  implements ResourcePropertyProvider 
         Cluster cluster = clusterService.findById(clusterId);
         if (!cluster.isAutoscalingEnabled().equals(enableAutoScaling)) {
             cluster = clusterService.setAutoscaleState(clusterId, enableAutoScaling);
-            createAutoscalingStateChangedHistoryAndNotify(cluster);
+            processAutoscalingStateChanged(cluster);
         }
         return cluster;
     }
 
     public void deleteAlertsForClusterCrn(String stackCrn) {
         Cluster cluster = getClusterByCrnOrName(NameOrCrn.ofCrn(stackCrn));
-        clusterService.deleteAlertsForCluster(cluster.getId());
+        cluster = clusterService.deleteAlertsForCluster(cluster.getId());
+        processAutoscalingStateChanged(cluster);
     }
 
     public void deleteAlertsForClusterName(String stackName) {
         Cluster cluster = getClusterByCrnOrName(NameOrCrn.ofName(stackName));
-        clusterService.deleteAlertsForCluster(cluster.getId());
+        cluster = clusterService.deleteAlertsForCluster(cluster.getId());
+        processAutoscalingStateChanged(cluster);
+    }
+
+    public void processAutoscalingStateChanged(Cluster cluster) {
+        createAutoscalingConfigChangedHistoryAndNotify(cluster);
+        usageReportingService.reportAutoscalingConfigChanged(restRequestThreadLocalService.getCloudbreakUser().getUserCrn(), cluster);
     }
 
     public void createLoadAlerts(Long clusterId, List<LoadAlert> loadAlerts) {
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
@@ -158,7 +158,7 @@ public class DistroXAutoScaleClusterV1Controller implements DistroXAutoScaleClus
         }
 
         Cluster updatedCluster = clusterService.findById(cluster.getId());
-        asClusterCommonService.createAutoscalingConfigChangedHistoryAndNotify(updatedCluster);
+        asClusterCommonService.processAutoscalingStateChanged(updatedCluster);
         return createClusterJsonResponse(updatedCluster);
     }
 }
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
@@ -33,6 +33,7 @@ import com.sequenceiq.periscope.notification.HttpNotificationSender;
 import com.sequenceiq.periscope.service.AuditService;
 import com.sequenceiq.periscope.service.HistoryService;
 import com.sequenceiq.periscope.service.PeriscopeMetricService;
+import com.sequenceiq.periscope.service.UsageReportingService;
 import com.sequenceiq.periscope.utils.LoggingUtils;
 
 @Component("ScalingRequest")
@@ -76,6 +77,9 @@ public class ScalingRequest implements Runnable {
 
     @Inject
     private AuditService auditService;
+
+    @Inject
+    private UsageReportingService usageReportingService;
 
     public ScalingRequest(Cluster cluster, ScalingPolicy policy, int totalNodes, int desiredNodeCount, List<String> decommissionNodeIds) {
         this.cluster = cluster;
@@ -136,7 +140,7 @@ public class ScalingRequest implements Runnable {
                     hostGroup, cluster.getStackCrn(), desiredNodeCount, statusReason, e);
             metricService.incrementMetricCounter(MetricType.CLUSTER_UPSCALE_FAILED);
         } finally {
-            createHistoryAndNotify(scalingAdjustment, totalNodes, statusReason, scalingStatus);
+            processAutoscalingTriggered(scalingAdjustment, totalNodes, statusReason, scalingStatus);
         }
     }
 
@@ -169,7 +173,7 @@ public class ScalingRequest implements Runnable {
             LOGGER.error("Couldn't trigger downscaling for host group '{}', cluster '{}', desiredNodeCount '{}', error '{}' ",
                     hostGroup, cluster.getStackCrn(), desiredNodeCount, statusReason, e);
         } finally {
-            createHistoryAndNotify(scalingAdjustment, totalNodes, statusReason, scalingStatus);
+            processAutoscalingTriggered(scalingAdjustment, totalNodes, statusReason, scalingStatus);
         }
     }
 
@@ -194,16 +198,17 @@ public class ScalingRequest implements Runnable {
                     "decommissionNodeIds '{}', error '{}' ", hostGroup, cluster.getStackCrn(), decommissionNodeIds.size(),
                     decommissionNodeIds, statusReason, e);
         } finally {
-            createHistoryAndNotify(-decommissionNodeIds.size(), totalNodes, statusReason, scalingStatus);
+            processAutoscalingTriggered(-decommissionNodeIds.size(), totalNodes, statusReason, scalingStatus);
         }
     }
 
-    private void createHistoryAndNotify(int adjustmentCount, int totalNodes, String statusReason, ScalingStatus scalingStatus) {
+    private void processAutoscalingTriggered(int adjustmentCount, int totalNodes, String statusReason, ScalingStatus scalingStatus) {
         History history = historyService.createEntry(scalingStatus,
                 StringUtils.left(statusReason, STATUSREASON_MAX_LENGTH), totalNodes, adjustmentCount, policy);
         notificationSender.sendHistoryUpdateNotification(history, cluster);
         auditService.auditAutoscaleServiceEvent(scalingStatus, statusReason, cluster.getStackCrn(),
                 cluster.getClusterPertain().getTenant(), System.currentTimeMillis());
+        usageReportingService.reportAutoscalingTriggered(adjustmentCount, totalNodes, scalingStatus, statusReason, policy.getAlert(), cluster);
     }
 
     private String getMessageForCBException(Exception cbApiException) {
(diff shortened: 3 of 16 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/service/UsageReportingService.java
```diff
@@ -0,0 +1,87 @@
...
+                .flatMap(Set::stream)
+                .forEach(alert -> cdpAutoscaleConfigChangedBuilder.addAutoscalingPolicyDefinition(withAlert(alert)));
+
+        usageReporter.cdpDatahubAutoscaleConfigChanged(cdpAutoscaleConfigChangedBuilder.build());
+    }
+
+    public UsageProto.CDPAutoscalePolicyDefinition withAlert(BaseAlert alert) {
+        Map parameters = new HashMap<>();
+        if (alert instanceof TimeAlert) {
+            TimeAlert ta = (TimeAlert) alert;
```
sidseth, line 70: Optionally move these into TimeAlert/LoadAlert - so that any changes in parameters, are handled correctly.

autoscale/src/test/java/com/sequenceiq/periscope/service/UsageReportingServiceTest.java
```diff
@@ -0,0 +1,109 @@
...
+public class UsageReportingServiceTest {
+
+    @InjectMocks
+    private UsageReportingService underTest;
+
+    @Mock
+    private UsageReporter usageReporter;
+
+    @Test
+    public void testReportAutoscalingTriggered() {
```
sidseth, line 38: Similar test for ScheduleBasedScaling?

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
```diff
@@ -138,6 +146,12 @@ private DistroXAutoscaleClusterResponse createClusterJsonResponse(Cluster cluste
         return distroXAutoscaleClusterResponseConverter.convert(cluster);
     }
 
+    private Cluster updateAutoscaleState(Long clusterId, AutoscaleClusterState autoscaleState) {
+        Cluster updatedCluster = asClusterCommonService.setAutoscaleState(clusterId, autoscaleState);
```
sidseth, line 150: Can we possibly capture the user, etc details - as to who made this change?
Independent of this CR: Does that need to be captured into AuditEvents?

usage-collection/src/main/proto/usage.proto
```diff
@@ -1168,3 +1172,52 @@ message CDPUDXBundleDetails {
   // The type of the diagnostics bundle.
   string type = 5;
 }
+
+//Autoscale Policy Definition
+message CDPAutoscalePolicyDefinition {
+  // The autoscale policy type
+  string autoscalePolicyType = 1;
```
sidseth, line 1179: Should this be an enum?

usage-collection/src/main/proto/usage.proto
```diff
@@ -1168,3 +1172,52 @@ message CDPUDXBundleDetails {
...
+  // The Autoscaling Policy Definition.
+  CDPAutoscalePolicyDefinition autoscalingPolicyDefinition = 4;
+  // The Original HostGroup Node Count.
+  int32 originalHostGroupNodeCount = 5;
+  // The Desired HostGroup Node Count.
+  int32 desiredHostGroupNodeCount = 6;
+   // The Autoscaling Action that was performed.
+  string autoscalingAction = 7;
+  // The status of the trigger
+  string autoscaleTriggerStatus = 8;
```
sidseth, line 1206: enum?

usage-collection/src/main/proto/usage.proto
```diff
@@ -1168,3 +1172,52 @@ message CDPUDXBundleDetails {
   // The type of the diagnostics bundle.
   string type = 5;
 }
+
```
sidseth, line 1175: Is the timestamp of these events captured automatically?
From looking at LoggingUsageReporter - it looks like some events set it explicitly, while others don't (guessing the rest have some kind of a timestamp within the event itself?)

sidseth (commented):
Have left some comments. A little concerned about the amount of load this may generate (as compared to other operations on the cluster which are far less frequent). I think it is important to discuss this aspect with @nadamdb

smaniraju:
> Have left some comments. A little concerned about the amount of load this may generate (as compared to other operations on the cluster which are far less frequent). I think it is important to discuss this aspect with @nadamdb

The telemetry is basically a LOG to console, so I dont think it adds any huge load to the service.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -194,16 +198,17 @@ private void scaleDownByNodeIds(List<String> decommissionNodeIds) {
...
     }
 
-    private void createHistoryAndNotify(int adjustmentCount, int totalNodes, String statusReason, ScalingStatus scalingStatus) {
+    private void processAutoscalingTriggered(int adjustmentCount, int totalNodes, String statusReason, ScalingStatus scalingStatus) {
         History history = historyService.createEntry(scalingStatus,
                 StringUtils.left(statusReason, STATUSREASON_MAX_LENGTH), totalNodes, adjustmentCount, policy);
         notificationSender.sendHistoryUpdateNotification(history, cluster);
         auditService.auditAutoscaleServiceEvent(scalingStatus, statusReason, cluster.getStackCrn(),
                 cluster.getClusterPertain().getTenant(), System.currentTimeMillis());
+        usageReportingService.reportAutoscalingTriggered(adjustmentCount, totalNodes, scalingStatus, statusReason, policy.getAlert(), cluster);
```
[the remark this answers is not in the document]
smaniraju: If scaling operation is already in progress LoadBased autoscaling wont trigger another event so there would not be that metric here. If time-based scaling is configured by user such that every 2 mins another scaling event is triggered this metric will help us to fine tune the validation required for schedule-based autoscaling.
Other flows generate similar events for all CB Flow changes and the number of events generated here will be far less. 
I have tested multiple times and there is no difference between actual scaling node count and yarn recommended node count. This information we already get from max node count in LoadAlertDefinition.

autoscale/src/main/java/com/sequenceiq/periscope/service/UsageReportingService.java
smaniraju: done

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
smaniraju: user details belongs to audit events and these are already captured in auditevents.

usage-collection/src/main/proto/usage.proto
smaniraju: timestamp is captured automatically

autoscale/src/test/java/com/sequenceiq/periscope/service/UsageReportingServiceTest.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/controller/AutoScaleClusterCommonService.java
```diff
@@ -51,6 +52,9 @@
     @Inject
     private HistoryService historyService;
 
+    @Inject
+    private UsageReportingService usageReportingService;
```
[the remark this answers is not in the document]
smaniraju: Fixed

autoscale/src/main/java/com/sequenceiq/periscope/controller/DistroXAutoScaleClusterV1Controller.java
```diff
@@ -96,14 +104,14 @@ public DistroXAutoscaleClusterResponse getClusterByName(@ResourceName String clu
...
-        return createClusterJsonResponse(asClusterCommonService.setAutoscaleState(cluster.getId(), autoscaleState));
+        return createClusterJsonResponse(updateAutoscaleState(cluster.getId(), autoscaleState));
     }
 
     @Override
     @CheckPermissionByResourceName(action = AuthorizationResourceAction.SCALE_DATAHUB)
     public DistroXAutoscaleClusterResponse enableAutoscaleForClusterName(@ResourceName String clusterName, AutoscaleClusterState autoscaleState) {
         Cluster cluster = asClusterCommonService.getClusterByStackName(clusterName);
-        return createClusterJsonResponse(asClusterCommonService.setAutoscaleState(cluster.getId(), autoscaleState));
+        return createClusterJsonResponse(updateAutoscaleState(cluster.getId(), autoscaleState));
```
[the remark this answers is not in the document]
smaniraju: It was already done for all updates, added even for deletes.

usage-collection/src/main/proto/usage.proto
smaniraju: done

usage-collection/src/main/proto/usage.proto
smaniraju: done

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
sidseth: The difference would come in when the YARN recommended node count is much higher than the maxNodeCount configured for the cluster?
On the downscale, the difference will be - YARN recommends N nodes, we need to decommission >N nodes - so which ones/how many were a result of YARN, and which ones/how many were randomly picked after this.

sidseth:
> > Have left some comments. A little concerned about the amount of load this may generate (as compared to other operations on the cluster which are far less frequent). I think it is important to discuss this aspect with @nadamdb
> 
> The telemetry is basically a LOG to console, so I dont think it adds any huge load to the service.

Would be good to understand if there will be any bottlenecks in the entire chain (i.e. consumption of the LOG)

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
smaniraju: The forced nodes autoscale downsized in case of LoadAlertConfig limits vs the ones recommended by YARN are all already available in Kibana. I dont agree such detailed information ( node-ids, ips, routine operation logs)  are all should be part of telemetry metrics. This will add volume to telemetry storage also.

autoscale/src/main/java/com/sequenceiq/periscope/service/UsageReportingService.java
```diff
@@ -0,0 +1,92 @@
...
+                UsageProto.CDPDatahubAutoscaleTriggered.newBuilder().setAutoscaleTriggerDetails(
+                        UsageProto.CDPAutoscaleTriggerDetails.newBuilder()
+                                .setAccountId(cluster.getClusterPertain().getTenant())
+                                .setClusterCrn(cluster.getStackCrn())
+                                .setClusterName(cluster.getStackName())
+                                .setAutoscalingAction(scalingAction)
+                                .setAutoscaleStatus(getAutoscaleStatusType(scalingStatus))
+                                .setAutoscalingPolicyDefinition(withAlert(alert))
+                                .setOriginalHostGroupNodeCount(clusterSize)
+                                .setDesiredHostGroupNodeCount(clusterSize + adjustmentCount)
```
nadamdb, line 42: The setter methods of UsageProto builder fail if a null value is passed to them so if you are not sure that these values cannot be null then it's advised to do a null check or set defaults (for example: .setClusterName(defaultIfEmpty(cluster.getStackName(), "")) )

nadamdb (commented):
> > > Have left some comments. A little concerned about the amount of load this may generate (as compared to other operations on the cluster which are far less frequent). I think it is important to discuss this aspect with @nadamdb
> > 
> > 
> > The telemetry is basically a LOG to console, so I dont think it adds any huge load to the service.
> 
> Would be good to understand if there will be any bottlenecks in the entire chain (i.e. consumption of the LOG)

On cloudbreak side it shouldn't cause a bottleneck because, as mentioned, it's a simple console log action. The bottleneck might come on the EDH team side when they consume the logs and transfer it to the database. It might be wise to discuss the frequency of these new events with Bryan Branstetter and Alan Jackoway (sorry, I couldn't find their GH usernames).

Also I added some comments regarding the use of usage.proto. These are not strictly change requests just some considerations. Otherwise the change (at least the logging part, what I'm familiar with) LGTM.

autoscale/src/main/java/com/sequenceiq/periscope/service/UsageReportingService.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    private UsageProto.AutoscalePolicyType.Value getAutoscalePolicyType(AlertType alertType) {
+        switch (alertType) {
+            case LOAD:
+                return UsageProto.AutoscalePolicyType.Value.LOAD_BASED;
+            case TIME:
+                return UsageProto.AutoscalePolicyType.Value.TIME_BASED;
+            default:
+                return UsageProto.AutoscalePolicyType.Value.UNRECOGNIZED;
```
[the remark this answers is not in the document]
smaniraju: ok changed.

autoscale/src/main/java/com/sequenceiq/periscope/service/UsageReportingService.java
smaniraju: ok changed.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
sidseth: At the same time, it is the best mechanism we have at the moment to correlate scaling operations/failures with what YARN had to say. Ideally, would like the YARN API to return metrics it used to come to the decision, which could also be part of a string dump (and a non queryable field in the final data consumption path).

Can add that as a follow up - I think this can be quite useful though, especially if YARN can give back the metrics that it used for the specific decision.

usage-collection/src/main/proto/usage.proto
```diff
@@ -1171,3 +1175,77 @@ message CDPUDXBundleDetails {
...
+}
+
+//Autoscale Trigger Status
+message AutoscaleScalingStatusType {
+  enum Value {
+    // A value indicating the enum is unset.
+    UNSET = 0;
+    // Scaling action successfully requested.
+    SUCCESS = 1;
+    // Scaling action failed.
```
sidseth, line 1210: What's the difference between FAILED and TRIGGER_FAILED?

sidseth (approved):
Looks good to me.

One additional question: The logging events are marked as 'Aync' - which presumably means they are handled by some part of the Spring system.
1. What else is handled by the same thread pool? (i.e. can this block other critical services).
2. If there are errors (e.g. NPEs) while constructing PB events - how will that show up. / and more generally errors from the async handler.

usage-collection/src/main/proto/usage.proto
smaniraju: Updated the status for easier understanding.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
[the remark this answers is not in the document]
smaniraju: ok.

smaniraju:
> Looks good to me.
> 
> One additional question: The logging events are marked as 'Aync' - which presumably means they are handled by some part of the Spring system.
> 
> 1. What else is handled by the same thread pool? (i.e. can this block other critical services).
> 2. If there are errors (e.g. NPEs) while constructing PB events - how will that show up. / and more generally errors from the async handler.

Async pool is currently used for auditing and telemetry.
Async Handler by default logs the exception and moves on.

PR #10991, opened 2021-06-30
CB-13252 Create new L0 test for Datalake Backup

aszegedi:
A new SDX Backup test case for L0 tests.

Beyond the new test case I've introduced:
- new operations at SDX API Endpoint:
   - `getBackupDatalakeStatusByName`: is needed for the `validateDatalakeBackupStatus` of `SdxBackupTest` 
   - `getBackupDatalakeStatusById`: this is for clarify, because of the `backupDatalakeStatusByName` operation is actually points to the `getBackupStatusByBackupId` at the `datalakeDrClient`.
- fixed some typos
- new request at Datalake DR Client:
   - `getBackupStatusByDatalakeName`: is needed for the `getBackupDatalakeStatusByName` new SDX API operation
- updated the SDX Controller and SDX Backup/Restore Service as well based on the above changes
- updated the Flow Util and Service with `getLatestFlowFinalizedAndFailed`: to stabilise the IT module wait for flow feature

### DIFF
--- a/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
+++ b/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
@@ -31,6 +31,8 @@ import com.sequenceiq.cloudbreak.validation.ValidStackNameLength;
 import com.sequenceiq.flow.api.model.FlowIdentifier;
 import com.sequenceiq.sdx.api.model.AdvertisedRuntime;
 import com.sequenceiq.sdx.api.model.RangerCloudIdentitySyncStatus;
+import com.sequenceiq.sdx.api.model.SdxBackupResponse;
+import com.sequenceiq.sdx.api.model.SdxBackupStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxClusterDetailResponse;
 import com.sequenceiq.sdx.api.model.SdxClusterRequest;
 import com.sequenceiq.sdx.api.model.SdxClusterResponse;
@@ -40,15 +42,14 @@ import com.sequenceiq.sdx.api.model.SdxDatabaseBackupStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxDatabaseRestoreResponse;
 import com.sequenceiq.sdx.api.model.SdxDatabaseRestoreStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxRepairRequest;
-import com.sequenceiq.sdx.api.model.SdxValidateCloudStorageRequest;
-import com.sequenceiq.sdx.api.model.SetRangerCloudIdentityMappingRequest;
-import com.sequenceiq.sdx.api.model.SdxBackupResponse;
-import com.sequenceiq.sdx.api.model.SdxBackupStatusResponse;
 import com.sequenceiq.sdx.api.model.SdxRestoreResponse;
 import com.sequenceiq.sdx.api.model.SdxRestoreStatusResponse;
+import com.sequenceiq.sdx.api.model.SdxValidateCloudStorageRequest;
+import com.sequenceiq.sdx.api.model.SetRangerCloudIdentityMappingRequest;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
 
 @Validated
 @Path("/sdx")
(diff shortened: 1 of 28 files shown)


datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
```diff
@@ -206,6 +206,20 @@ SdxBackupStatusResponse backupDatalakeStatusByName(@PathParam("name") String nam
             @QueryParam("backupId") String backupId,
             @QueryParam("backupName") String backupName);
 
+    @GET
+    @Path("/datalakename/{datalakeName}/backupDatalakeStatus")
```
lturcsanyi, line 210: I think we should keep the url convention for the REST endpoints.
I see the original endpoint `{name}/backupDatalakeStatus` uses the backupId, so I suggest something like this for the new endpoints:
`{name}/backupStatus` or `{name}/backupStatusByDatalakeName`
and
`{name}/backupStatusByBackupId/{backupId}`

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/gcp/action/GcpClientActions.java
```diff
@@ -180,12 +183,17 @@ public void stopHostGroupInstances(List<String> instanceIds) {
         return operation == null ? null : operation.getError();
     }
 
-    public void listBucketSelectedObject(String baseLocation, String selectedObject, boolean zeroContent) {
+    public void listBucketSelectedObject(String baseLocation, String selectedObject, Boolean zeroContent) {
```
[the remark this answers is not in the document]
aszegedi: `listBucketSelectedObject` has three implementations in IT module based on the provider:
- [S3ClientActions.listBucketSelectedObject(String baseLocation, String selectedObject, Boolean zeroContent)](https://github.com/hortonworks/cloudbreak/blob/bea15989757a942fe829eb61a4336cfd61de40a6/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/aws/amazons3/action/S3ClientActions.java#L72)
- [GcpClientActions.listBucketSelectedObject(String baseLocation, String selectedObject, boolean zeroContent)](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/gcp/action/GcpClientActions.java#L183)
- [AzureCloudBlobClientActions.listSelectedDirectory(String baseLocation, String selectedDirectory, Boolean zeroContent)](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/azure/azurecloudblob/action/AzureCloudBlobClientActions.java#L317)
In case of AWS and Azure the type was `Boolean` expect GCP where this was `boolean`. So I had to made a common solution in the [CloudFunctionality interface cloudStorageListContainer](https://github.com/hortonworks/cloudbreak/blob/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/CloudFunctionality.java#L53). Beyond this the `zeroContent` variable is required here. So cannot be `null`.

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
aszegedi: These were introduced originally next to [@Path("{name}/backupDatalakeStatus")](https://github.com/hortonworks/cloudbreak/blob/master/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java#L201-L207). If you check, for example the [SdxUpgradeEndpoint](https://github.com/hortonworks/cloudbreak/blob/master/datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxUpgradeEndpoint.java) there are also similar operation id implementations:
```
 @Path("{name}/upgrade")
...
@Path("/crn/{crn}/upgrade")
```

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/util/gcp/action/GcpClientActions.java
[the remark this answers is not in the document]
lturcsanyi: If zeroContent cant be null, then we can use  boolean instead of Boolean. There is a potential NPE right now if somebody calls the method with null value.

datalake-api/src/main/java/com/sequenceiq/sdx/api/endpoint/SdxEndpoint.java
lturcsanyi: I see that there is a convention for crn's, but in _SdxEndpoint_ all other methods that use the datalake name, use the same **{name}/operation** format.
Also the datalake name is required in all requests, it can be confusing to have @PathParam in one request and @QueryParam in another.

datalake-dr-connector/src/main/java/com/sequenceiq/cloudbreak/datalakedr/DatalakeDrClient.java
```diff
@@ -97,6 +97,38 @@ public DatalakeDrStatusResponse triggerRestore(String datalakeName, String backu
...
+        }
+
+        checkNotNull(datalakeName);
+        checkNotNull(actorCrn);
+
+        try (ManagedChannelWrapper channelWrapper = makeWrapper()) {
+            BackupDatalakeStatusRequest.Builder builder = BackupDatalakeStatusRequest.newBuilder()
+                    .setDatalakeName(datalakeName);
+            if (!Strings.isNullOrEmpty(backupId)) {
+                builder.setBackupName(backupId);
```
lturcsanyi, line 120: It should be builder.setBackupId(backupId);

aszegedi: Good catch! Thanks, I'm fixing this.

lturcsanyi (approved):

PR #10997, opened 2021-07-01
CB-13176 Gather AWS medium duty datalake load balancer configs and re…

hreeve-cloudera:
…turn via cdp cli

The goal of this change is to provide load balancer details that can be used to find and query the
load balancer configuration on the cloud provider side. This change includes:
  1. New columns in the loadbalancer and targetgroup tables where cloud provider specific metadata
     can be stored, via a serialized JSON class.
  2. LoadBalancerConfigDbWrapper and TargetGroupConfigDbWrapper, the wrapper classes that will be
     serialized and stored in the database, and contain the cloud provider metadata.
  3. A new converter, LoadBalancerToLoadBalanacerResponseConverter, that will convert the database
     objects into a LoadBalancerResponse, included as part of the StackV4Response.
  4. The AWS implementation of the metadata collection, storage, and conversion.

### DIFF
--- a/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
+++ b/cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
@@ -33,6 +33,7 @@ import com.sequenceiq.cloudbreak.cloud.aws.client.AmazonCloudFormationClient;
 import com.sequenceiq.cloudbreak.cloud.aws.common.AwsPlatformResources;
 import com.sequenceiq.cloudbreak.cloud.aws.common.client.AmazonEc2Client;
 import com.sequenceiq.cloudbreak.cloud.aws.common.loadbalancer.AwsLoadBalancer;
+import com.sequenceiq.cloudbreak.cloud.aws.common.loadbalancer.AwsLoadBalancerScheme;
 import com.sequenceiq.cloudbreak.cloud.aws.common.loadbalancer.LoadBalancerTypeConverter;
 import com.sequenceiq.cloudbreak.cloud.aws.common.util.AwsLifeCycleMapper;
 import com.sequenceiq.cloudbreak.cloud.aws.common.view.AuthenticatedContextView;
@@ -69,6 +70,9 @@ public class AwsMetadataCollector implements MetadataCollector {
 
     @Inject
     private AwsPlatformResources awsPlatformResources;
+
+    @Inject
+    private AwsLoadBalancerMetadataCollector awsLoadBalancerMetadataCollector;
 
     @Override
     public List<CloudVmMetaDataStatus> collect(AuthenticatedContext ac, List<CloudResource> resources, List<CloudInstance> vms,
@@ -309,15 +313,20 @@ public class AwsMetadataCollector implements MetadataCollector {
 
         List<CloudLoadBalancerMetadata> cloudLoadBalancerMetadata = new ArrayList<>();
         for (LoadBalancerType type : loadBalancerTypes) {
-            String loadBalancerName = AwsLoadBalancer.getLoadBalancerName(loadBalancerTypeConverter.convert(type));
+            AwsLoadBalancerScheme scheme = loadBalancerTypeConverter.convert(type);
+            String loadBalancerName = AwsLoadBalancer.getLoadBalancerName(scheme);
             LOGGER.debug("Attempting to collect metadata for load balancer {}, type {}", loadBalancerName, type);
             try {
                 LoadBalancer loadBalancer = cloudFormationStackUtil.getLoadBalancerByLogicalId(ac, loadBalancerName);
+                LOGGER.debug("Parsing all listener and target group information for load balancer {}", loadBalancerName);
+                Map<String, Object> parameters = awsLoadBalancerMetadataCollector.getParameters(ac, loadBalancer, scheme);
+
                 CloudLoadBalancerMetadata loadBalancerMetadata = new CloudLoadBalancerMetadata.Builder()
                     .withType(type)
                     .withCloudDns(loadBalancer.getDNSName())
                     .withHostedZoneId(loadBalancer.getCanonicalHostedZoneId())
                     .withName(loadBalancerName)
+                    .withParameters(parameters)
                     .build();
                 cloudLoadBalancerMetadata.add(loadBalancerMetadata);
                 LOGGER.debug("Saved metadata for load balancer {}: DNS {}, zone id {}", loadBalancerName, loadBalancer.getDNSName(),
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/AwsLoadBalancerResponse.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/AwsLoadBalancerResponse.java
@@ -1,0 +1,24 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.loadbalancer;
+
+import java.io.Serializable;
+
+import javax.validation.constraints.NotNull;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions.StackModelDescription;
+
+public class AwsLoadBalancerResponse implements Serializable {
+
+    @ApiModelProperty(StackModelDescription.AWS_LB_ARN)
+    @NotNull
+    private String arn;
+
+    public String getArn() {
+        return arn;
+    }
+
+    public void setArn(String arn) {
+        this.arn = arn;
+    }
+}
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/LoadBalancerResponse.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/LoadBalancerResponse.java
@@ -1,0 +1,85 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.loadbalancer;
+
+import java.io.Serializable;
+import java.util.List;
+
+import javax.validation.constraints.NotNull;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions.StackModelDescription;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.sequenceiq.common.api.type.LoadBalancerType;
+
+@JsonInclude(Include.NON_NULL)
+public class LoadBalancerResponse implements Serializable {
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_FQDN)
+    private String fqdn;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_CLOUD_DNS)
+    private String cloudDns;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_IP)
+    private String ip;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_TARGETS)
+    @NotNull
+    private List<TargetGroupResponse> targets;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_TYPE)
+    @NotNull
+    private LoadBalancerType type;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_AWS)
+    private AwsLoadBalancerResponse awsResourceId;
+
+    public String getFqdn() {
+        return fqdn;
+    }
+
+    public void setFqdn(String fqdn) {
+        this.fqdn = fqdn;
+    }
+
+    public String getCloudDns() {
+        return cloudDns;
+    }
+
+    public void setCloudDns(String cloudDns) {
+        this.cloudDns = cloudDns;
+    }
+
+    public String getIp() {
+        return ip;
+    }
+
+    public void setIp(String ip) {
+        this.ip = ip;
+    }
+
+    public List<TargetGroupResponse> getTargets() {
+        return targets;
+    }
+
+    public void setTargets(List<TargetGroupResponse> targets) {
+        this.targets = targets;
+    }
+
+    public LoadBalancerType getType() {
+        return type;
+    }
+
+    public void setType(LoadBalancerType type) {
+        this.type = type;
+    }
+
+    public AwsLoadBalancerResponse getAwsResourceId() {
+        return awsResourceId;
+    }
+
+    public void setAwsResourceId(AwsLoadBalancerResponse awsResourceId) {
+        this.awsResourceId = awsResourceId;
+    }
+}
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/TargetGroupResponse.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/TargetGroupResponse.java
@@ -1,0 +1,51 @@
+package com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.loadbalancer;
+
+import java.io.Serializable;
+import java.util.Set;
+
+import javax.validation.constraints.NotNull;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions.StackModelDescription;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+@JsonInclude(Include.NON_NULL)
+public class TargetGroupResponse implements Serializable {
+
+    @ApiModelProperty(StackModelDescription.TARGET_GROUP_PORT)
+    @NotNull
+    private long port;
+
+    @ApiModelProperty(StackModelDescription.TARGET_GROUP_INSTANCES)
+    @NotNull
+    private Set<String> targetInstances;
+
+    @ApiModelProperty(StackModelDescription.TARGET_GROUP_AWS)
+    private AwsTargetGroupResponse awsResourceIds;
+
+    public long getPort() {
+        return port;
+    }
+
+    public void setPort(long port) {
+        this.port = port;
+    }
+
+    public Set<String> getTargetInstances() {
+        return targetInstances;
+    }
+
+    public void setTargetInstances(Set<String> targetInstances) {
+        this.targetInstances = targetInstances;
+    }
+
+    public AwsTargetGroupResponse getAwsResourceIds() {
+        return awsResourceIds;
+    }
+
+    public void setAwsResourceIds(AwsTargetGroupResponse awsResourceIds) {
+        this.awsResourceIds = awsResourceIds;
+    }
+}
--- a/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/loadbalancer/LoadBalancerToLoadBalancerResponseConverter.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/loadbalancer/LoadBalancerToLoadBalancerResponseConverter.java
@@ -1,0 +1,125 @@
+package com.sequenceiq.cloudbreak.converter.v4.stacks.loadbalancer;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.loadbalancer.AwsLoadBalancerResponse;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.loadbalancer.AwsTargetGroupResponse;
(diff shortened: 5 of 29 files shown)


yg-gd (changes requested):

core-model/src/main/java/com/sequenceiq/cloudbreak/domain/stack/loadbalancer/aws/AwsTargetGroupArns.java
```diff
@@ -0,0 +1,27 @@
+package com.sequenceiq.cloudbreak.domain.stack.loadbalancer.aws;
+
+/**
+ * Contains the ARNs for both a listener and the target group that listener forwards traffic to.
+ */
+public class AwsTargetGroupArns {
```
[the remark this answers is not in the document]
hreeve-cloudera: I don't disagree that there's a lot of hoops to jump though. In fact, there should be exactly the three you mentioned, so I'm not sure what the fourth hoop is. If you can point it out specifically I can try to remove it.

We're mostly running into dependency problems when it comes to object optimization. We need to convert AWS metadata -> database object -> response object. So we do `AwsLoadBalancerMetadata` (metadata object) -> `AwsLoadBalancerConfig` (database object) -> `AwsLoadBalancerSettings` (response object). One possible optimization might be to just use the DB classes directly instead of the `CloudLoadBalancerMetadata` object. But the problem is, the metadata collector modules don't have a dependency on the database entity module, so I can't create objects of that type there. Similarly, the database entity modeul doesn't have a dependency on the metadata collector modules, so I can't create an object in those modules and use it directly as part of the database entries. In these various objects seem to meet in the `core` module, which does have the appropriate dependencies to use all the objects. But we need a way to get the information to `core`, which is where the different types of objects come from. They're originated in parts of the code that can't share information, so we need a module specific way to propagate that information back to the `core` module that can put it all together.

Man, the names of these things are still terrible. I'm going to try again.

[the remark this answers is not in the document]
hreeve-cloudera: Okay:
* Metadata object: `LoadBalancerMetadataBase` for base object, `AwsLoadBalancerMetadata` for specific implementation (unchanged)
* Database object: `LoadBalancerConfigDbWrapper` for base object, `AwsLoadBalancerConfigDb` for specific object. All objects the ultimate end up in the database now have `Db` as part of their name.
* Response object: `AwsLoadBalancerResponse`
Now hopefully it's actually clear what part each objects belongs to from the name.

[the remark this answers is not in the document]
yg-gd: I understand now. makes sense and the direction dependencies flow in is probably the only way to do it.

yg-gd (changes requested):
Just one thing otherwise looks good

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsLoadBalancerMetadata.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public List<AwsTargetGroupMetadata> getTargetGroupMetadata() {
+        return targetGroupMetadata;
+    }
+
+    public void setTargetGroupMetadata(List<AwsTargetGroupMetadata> targetGroupMetadata) {
+        this.targetGroupMetadata = targetGroupMetadata;
+    }
+
+    public void addTargetGroupMetadata(AwsTargetGroupMetadata targetGroupConfig) {
```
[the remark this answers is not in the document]
hreeve-cloudera: /flips table

Thought I caught all those :laughing:

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsLoadBalancerMetadata.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * AWS specific extension of LoadBalancerMetadataBase. For AWS, the resource information we need is the ARN
+ * for the load balancer, and information about any target groups and listeners attached to the load
+ * balancer. This information is represented in the AwsTargetGroupMetadata object. Using the provided information,
+ * a user can use the AWS console and CLI to look up additional information about the load balancer as needed.
+ */
+public class AwsLoadBalancerMetadata extends LoadBalancerMetadataBase {
```
lacikaaa, line 12: this doesn't seem right to have provider specific stuff in the cloud-api modul. the purpose of this is to provide an abstraction layer

lacikaaa (changes requested):

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsLoadBalancerMetadata.java
hreeve-cloudera: Then do you know what module it should go in? I understand the purpose of the CB model is abstraction. But the load balancers are different enough between cloud providers that we couldn't find a way to complete abstract it out. We need this information to return anything useful about the load balancers, but I'm happy to move it if it belongs in a different part of the code.

lacikaaa: I would expect something similar what we have for credentials. Those are pretty different per providers.
@biharitomi @doktoric do you have any better suggestion/idea?

hreeve-cloudera: Do you mean the `AwsCredentialPrerequisites`, `AzureCredentialPrerequisites`, and `GcpCredentialPrerequisites` classes? Because those are also in the cloud-api module.

yg-gd (approved):

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsLoadBalancerMetadata.java
lacikaaa: no, I meant `CloudCredential` which is kinda uniqe for the providers. I'm not sure why the prerequisite check generalized.
I'm asking the guys for input, because I think they might have better knowledge about the cloud modules. Let me ping them

doktoric: LoadBalancerMetadata can be an extends of DynamicModel and then you can use the parameter map for storing cloud provider related data. You can create a view object where you want to use the map elements e.g AwsCredentialView. 
In this way you can store the information in a class which is abstract enough.

biharitomi: I would also prefer the approach which would not contain any provider specific info in the `API` module. 
I think the credential prerequisite stuff is there by a mistake and probably didn't catch during the review process.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsTargetGroupMetadata.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Objects;
+
+/**
+ * Contains information about listeners and target groups attached to a load balancer. In AWS each listener
+ * listens on a single port, and forwards traffic received on that port to a target group, which distributes
+ * the traffic among the instances in the target group on a defined traffic port. Listeners and target groups
+ * each have their own ARN. In the cloudbreak implementation, the listener port and the target group traffic
+ * port are the same, so they are represented by a single 'port' member.
+ */
+public class AwsTargetGroupMetadata {
```
biharitomi, line 12: I would use the same approach here, not to store AWS related stuff in cloud model directly.

cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
```diff
@@ -252,4 +276,50 @@ private void addKnownInstance(CloudInstance cloudInstance, Multimap<String, Inst
...
+    /**
+     * Parses the stack resource summary for the cloudformation stack and pulls out all listener and target group ARNs
+     * associated with a particular load balancer, and creates AwsTargetGroupMetadata objects using the provided
+     * resources summaries.
+     * @param scheme The scheme of the load balancer being processed, either INTERNAL or INTERNET-FACING
+     * @param summaries The list of resource summaries from cloud formation that contain both the logical resource id
+     *                  (the name), and the physical resource id (ARN).
+     * @return A list of metadata pulled from the resource summaries.
+     */
+    private List<AwsTargetGroupMetadata> parseTargetGroupCloudConfig(AwsLoadBalancerScheme scheme, List<StackResourceSummary> summaries) {
```
biharitomi, line 289: Can we create a separate (util) class to cover this functionality instead of extending the metadata collector?
I think it would result in cleaner code and easier to be unit tested.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/LoadBalancerResponse.java
```diff
@@ -0,0 +1,85 @@
...
+import com.sequenceiq.common.api.type.LoadBalancerType;
+
+@JsonInclude(Include.NON_NULL)
+public class LoadBalancerResponse implements Serializable {
+
+    @ApiModelProperty
+    private String fqdn;
+
+    @ApiModelProperty
+    private String cloudDns;
```
biharitomi, line 21: Can we have model description for the new request/response objects?
It would be easier to understand what value the field could contain.

An example on StackV4Response: https://github.com/hortonworks/cloudbreak/blob/5d33992d0bc8b2a278a6c5ecf45a0876534a262c/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/StackV4Response.java#L82

biharitomi (changes requested):
Overall, usually a cluster has one LB with one target-group and listener on AWS side this approach is more future-proof but handling multiple LB with multiple listeners and target-group has it's own overhead and complexity from code quality point of view.
On the other hand this functionality should be implemented for the native AWS variant, but it will be much more easier as resources already persist their provider side reference at creation, the only thing that needs to be implemented is gathering them in the appropriate metadata collection.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsLoadBalancerMetadata.java
hreeve-cloudera: I will look into that. I admit to being frustrated that this review has been up for three weeks, and I'm only now getting feedback to significantly rework it.

lacikaaa: I understand the frustration, but I got to you in ~5 workdays after I got the email notification from Github. I know it's still a lot.

biharitomi: Sorry, but we needed to finish the demoable implementation of an immediate feature before I could allocate a ~2 hours to read through a ~50 files PR.

hreeve-cloudera: Yeah, the fact that I changed the package of one of the classes meant a lot of files that weren't otherwise changed ended up with different imports. I've actually undone that move, so there are less files touched now.

I've made `CloudLoadBalancerMetadata` extend `DynamicModel`, and created an `AwsLoadBalancerMetadataView` to parse the parameters. Everything still converts as expected according to the unit tests. Let me know how it looks.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/loadbalancer/AwsTargetGroupMetadata.java
hreeve-cloudera: This class doesn't exist anymore.

cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsMetadataCollector.java
hreeve-cloudera: Moved this to a new `AwsLoadBalancerMetadataCollectorClass`.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/LoadBalancerResponse.java
hreeve-cloudera: Done.

hreeve-cloudera:
> Overall, usually a cluster has one LB with one target-group and listener on AWS side this approach is more future-proof but handling multiple LB with multiple listeners and target-group has it's own overhead and complexity from code quality point of view.

The day where we have more than one target group is coming. If we ever get CM HA working, we'll have two target groups, one for Knox and one for CM. It does make the code more complex, but I designed it that one from the beginning because I knew that's the direction we were heading.

> On the other hand this functionality should be implemented for the native AWS variant, but it will be much more easier as resources already persist their provider side reference at creation, the only thing that needs to be implemented is gathering them in the appropriate metadata collection.

Yep, I opened CB-13514 for that. From a quick look it does look like ARNs are more accessible and it'll be simpler there. I'll work on that once this is merged in, since this has the CB model code I need.

biharitomi (approved):

biharitomi:
> > Overall, usually a cluster has one LB with one target-group and listener on AWS side this approach is more future-proof but handling multiple LB with multiple listeners and target-group has it's own overhead and complexity from code quality point of view.
> 
> The day where we have more than one target group is coming. If we ever get CM HA working, we'll have two target groups, one for Knox and one for CM. It does make the code more complex, but I designed it that one from the beginning because I knew that's the direction we were heading.
> 
> > On the other hand this functionality should be implemented for the native AWS variant, but it will be much more easier as resources already persist their provider side reference at creation, the only thing that needs to be implemented is gathering them in the appropriate metadata collection.
> 
> Yep, I opened CB-13514 for that. From a quick look it does look like ARNs are more accessible and it'll be simpler there. I'll work on that once this is merged in, since this has the CB model code I need.

Yeah, I think this is the more future-proof approach as I wrote it in my previous comment.
I approved the PR, and move of a class has been removed, if you put it in a separate commit, the move, then I can process the nearly 60 files affected PR much more easily :)

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/AwsLoadBalancerResponse.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.Serializable;
+
+import javax.validation.constraints.NotNull;
+
+import io.swagger.annotations.ApiModelProperty;
+
+public class AwsLoadBalancerResponse implements Serializable {
+
+    @ApiModelProperty
```
lacikaaa, line 11: I think should add some value to this annotation, same for `AwsTargetGroupResponse`

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/LoadBalancerResponse.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_TARGETS)
+    @NotNull
+    private List<TargetGroupResponse> targets;
+
+    @ApiModelProperty(StackModelDescription.LOAD_BALANCER_TYPE)
+    @NotNull
+    private LoadBalancerType type;
+
+    // null for non-AWS stacks
```
lacikaaa, line 35: not necessary comment

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/loadbalancer/TargetGroupResponse.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @ApiModelProperty(StackModelDescription.TARGET_GROUP_PORT)
+    @NotNull
+    private long port;
+
+    @ApiModelProperty(StackModelDescription.TARGET_GROUP_INSTANCES)
+    @NotNull
+    private Set<String> targetInstances;
+
+    // null for non-AWS stacks
```
lacikaaa, line 25: same here, please drop it

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/loadbalancer/LoadBalancerToLoadBalancerResponseConverter.java
```diff
@@ -0,0 +1,124 @@
...
+            TargetGroupResponse response = new TargetGroupResponse();
+            response.setPort(portPair.getTrafficPort());
+            response.setTargetInstances(instanceIds);
+            if (targetGroupConfig != null) {
+                response.setAwsResourceIds(convertAwsTargetGroup(targetGroupConfig.getAwsConfig(), portPair.getTrafficPort()));
+            }
+            responses.add(response);
+        }
+
+        return responses;
```
lacikaaa, line 99: this could be a bit improved I think, like:
```
        return portPairs.stream()
                .map(portPair -> mapPortPairToTargetGroup(instanceIds, targetGroupConfig, portPair))
    }

    private TargetGroupResponse mapPortPairToTargetGroup(Set<String> instanceIds, TargetGroupConfigDbWrapper targetGroupConfig, TargetGroupPortPair portPair) {
        TargetGroupResponse response = new TargetGroupResponse();
        response.setPort(portPair.getTrafficPort());
        response.setTargetInstances(instanceIds);
        if (targetGroupConfig != null) {
            response.setAwsResourceIds(convertAwsTargetGroup(targetGroupConfig.getAwsConfig(), portPair.getTrafficPort()));
        }
        return response;
    }

File: core/src/main/java/com/sequenceiq/cloudbreak/service/LoadBalancerConfigConverter.java [hunk 11]
```diff
@@ -0,0 +1,77 @@
...
+    static final String MISSING_CLOUD_RESOURCE = "Could not find cloud resource corresponding to this traffic port.";
+
+    @Inject
+    private LoadBalancerConfigService loadBalancerConfigService;
+
+    public LoadBalancerConfigDbWrapper convertLoadBalancer(String cloudPlatform, CloudLoadBalancerMetadata cloudLoadBalancerMetadata) {
+        switch (cloudPlatform) {
+            case AWS:
+                return buildAwsConfig(new AwsLoadBalancerMetadataView(cloudLoadBalancerMetadata));
+            // TODO: AZURE, GCP
```
lacikaaa, line 35: or you can just add a placeholder:
```
            case AZURE:
            case GCP:
            default:

File: core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/MetadataSetupService.java [hunk 12]
```diff
@@ -317,8 +326,17 @@ public void saveLoadBalancerMetadata(Stack stack, Iterable<CloudLoadBalancerMeta
                 String endpoint = loadBalancerConfigService.generateLoadBalancerEndpoint(stack);
                 LOGGER.info("Saving load balancer endpoint as: {}", endpoint);
                 loadBalancerEntry.setEndpoint(endpoint);
+                loadBalancerEntry.setProviderConfig(loadBalancerConfigConverter.convertLoadBalancer(stack.getCloudPlatform(),
```
lacikaaa, line 329: for a separate PR: could you please move out LB related stuff from `MetadataSetupService` to a separate class? like `LBMetadataSetupService`

core/src/main/resources/schema/app/20210629154850_CB-13176_Gather_AWS_load_balancer_configs.sql
```diff
@@ -0,0 +1,11 @@
+-- // CB-13176 Gather AWS load balancer configs
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE loadbalancer ADD COLUMN IF NOT EXISTS providerconfig TEXT NULL;
+ALTER TABLE targetgroup ADD COLUMN IF NOT EXISTS providerconfig TEXT NULL;
```
lacikaaa, line 5: why is the `NULL` necessary here?

lacikaaa (changes requested):
overall LGTM, just mentioned some nits

core/src/main/java/com/sequenceiq/cloudbreak/converter/v4/stacks/loadbalancer/LoadBalancerToLoadBalancerResponseConverter.java
hreeve-cloudera: Done. Thanks for the code.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/MetadataSetupService.java
hreeve-cloudera: Created CB-13645.

lacikaaa (approved):

PR #11004, opened 2021-07-02
CB-13260 Datalake cloud storage validation error on Azure in case of …

bbihari:
…non-subscription wide policies of the credential's service principal

In case of Azure single resourcegroup the roleassignment are queried from the resourcegroup instead of the subscription.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/objectstorage/ObjectStorageValidateRequest.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/objectstorage/ObjectStorageValidateRequest.java
@@ -24,6 +24,8 @@ public class ObjectStorageValidateRequest implements CloudPlatformAware {
 
     private String logsLocationBase;
 
+    private AzureParameters azure;
+
     private MockAccountMappingSettings mockAccountMappingSettings;
 
     public ObjectStorageValidateRequest() {
@@ -36,6 +38,7 @@ public class ObjectStorageValidateRequest implements CloudPlatformAware {
         this.spiFileSystem = builder.spiFileSystem;
         this.logsLocationBase = builder.logsLocationBase;
         this.mockAccountMappingSettings = builder.mockAccountMappingSettings;
+        this.azure = builder.azure;
     }
 
     public static Builder builder() {
@@ -90,6 +93,14 @@ public class ObjectStorageValidateRequest implements CloudPlatformAware {
         this.mockAccountMappingSettings = mockAccountMappingSettings;
     }
 
+    public AzureParameters getAzure() {
+        return azure;
+    }
+
+    public void setAzure(AzureParameters azure) {
+        this.azure = azure;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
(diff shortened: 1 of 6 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/objectstorage/ObjectStorageValidateRequest.java
```diff
@@ -175,6 +190,13 @@ public Builder withMockSettings(String region, String adminGroupName) {
             return this;
         }
 
+        public Builder withAzureParameters(String singleResourceGroupName) {
```
lajosrodek, line 193: It does not make a difference now, but I suggest changing the signature to take an argument of type `AzureParameters` instead.

bbihari: I would leave this one as long as the number of parameters are limited because this way we could create builder method for the important scenarios limiting the chance of not setting or setting too many parameters and this way we can keep AzureParameters as implementation detail.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/objectstorage/ObjectStorageValidateRequest.java
```diff
@@ -24,6 +24,8 @@
 
     private String logsLocationBase;
 
+    private AzureParameters azure;
```
pdarvasi, line 27: Don't you want to include AzureParameters in a CloudParamters class, for cleaner future extensibility or you'd prefer to put any provider-specific param here in the future?

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/validator/AzureIDBrokerObjectStorageValidatorTest.java
```diff
@@ -127,7 +127,7 @@ public void setup() {
     public void testValidateObjectStorageWithoutFileSystems() {
         SpiFileSystem fileSystem = new SpiFileSystem("test", FileSystemType.ADLS_GEN_2, null);
         ValidationResultBuilder resultBuilder = new ValidationResultBuilder();
-        underTest.validateObjectStorage(client, fileSystem, "", resultBuilder);
+        underTest.validateObjectStorage(client, fileSystem, "", null, resultBuilder);
```
pdarvasi, line 130: could you pls add a unit test to cover single RG case, too?

pdarvasi (changes requested):

cloud-azure/src/test/java/com/sequenceiq/cloudbreak/cloud/azure/validator/AzureIDBrokerObjectStorageValidatorTest.java
bbihari: sure

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/objectstorage/ObjectStorageValidateRequest.java
bbihari: I think it's not necessary here, there will be parameters related only to one provider at a time.

pdarvasi (approved):

pdarvasi:
retest this please

PR #11005, opened 2021-07-02
CB-13259 FreeIPA replica DNA ID range initialization

lacikaaa:
When a FreeIPA first instance is installed it allocates an id range with a size of 200000. The FreeIPA allocates ids from the beginning of this range when creating a user or a group.
After a replica is added to the existing FreeIPA domain it will ask for an id range when first time it needs an id, like the first time a user is added on that instance. In this case another FreeIPA master which already has an id range set, will give the upper half of it range to the one without a range. This behavior has 2 consequences:
- a FreeIPA instance/replica can exist without an id range for indefinitely long time
- with every new replica another instance can "loose" half of its range available

With this commit after replica install every replica will run a script to try to allocate ID range for itself.
Logic implemented:
- validate the new replica doesn't have a range assigned. If it has, then the script exists as it has nothing to do
- fetch the original whole ID range
- fetch the occupied ID ranges, both used and next ranges
- check if there is any free range with at least a 1000 ID
- if there is, try to assign one as next range. If one fails (maybe another replica was faster) then try the next one until there is no more
- create and delete a user to ensure the replica has an active range, either assigned by the script or by getting half range of another instance

It's possible when querying the already asigned ranges, instances return overlapping ranges. If this happens there is a retry after 3 to 10 seconds.

The script is prepared to run using `ipa console` command. This enforces some changes to the script which won't be necessary in case of a normal python script.

Developed on a separate repo for easier development and testing: lacikaaa/ipadnarange

### DIFF
--- a/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
+++ b/freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
@@ -1,0 +1,156 @@
+from ipalib import api
+from ipaserver.install import replication
+import logging
+from random import randint
+from time import sleep
+
+
+class RangeSet:
+    def __init__(self, elements):
+        self.ranges = list(elements)
+
+    def __iter__(self):
+        return iter(self.ranges)
+
+    def __repr__(self):
+        return 'RangeSet: %r' % self.ranges
+
+    def has(self, tup):
+        for pos, i in enumerate(self.ranges):
+            if i[0] <= tup[0] and i[1] >= tup[1]:
+                return pos, i
+        raise ValueError('Invalid range or overlapping range')
+
+    def minus(self, tup):
+        pos, (x, y) = self.has(tup)
+        out = []
+        if x < tup[0]:
+            out.append((x, tup[0] - 1))
+        if y > tup[1]:
+            out.append((tup[1] + 1, y))
+        self.ranges[pos:pos + 1] = out
+
+    def __sub__(self, r):
+        r1 = RangeSet(self)
+        for i in r: r1.minus(i)
+        return r1
+
+    def sub(self, r):  # inplace subtraction
+        for i in r:
+            self.minus(i)
+
+    def subWithResult(self, result, r):  # unfortunately ipa console won't work with the __sub__ solution, couldn't initialize RangeSet
+        result.ranges = self.ranges
+        for i in r:
+            result.minus(i)
+        return result
+
+
+class DnaRangeIniter:
+    """Needed a class as FreeIPA console couldn't resolved method to method calls"""
+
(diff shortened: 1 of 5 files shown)


freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
```diff
@@ -0,0 +1,156 @@
...
+        result.ranges = self.ranges
+        for i in r:
+            result.minus(i)
+        return result
+
+
+class DnaRangeIniter:
+    """Needed a class as FreeIPA console couldn't resolved method to method calls"""
+
+    MIN_RANGE_SIZE = 1000
```
jamisonbennett, line 52: I fear this isn't sufficient. If 1000 users/groups are added then how does a new range get reallocated?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
```diff
@@ -0,0 +1,156 @@
...
+            exit(0)
+
+        dna_next_range_start, dna_next_range_end = self.localReplManager.get_DNA_next_range(api.env.host)
+        print ("DNA next range on this server: {} - {}".format(dna_next_range_start, dna_next_range_end))
+        if dna_next_range_start is not None:
+            print ("DNA next range is already set on this server, exiting")
+            exit(0)
+
+    def get_all_servers_fqdn(self):
+        servers_resp = api.Command.server_find(None, pkey_only=True)['result']
```
jamisonbennett, line 73: I am not sure this handles ipa servers that were previously deleted. Is it possible that they the original ranges are reused?

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
lacikaaa: it will ask for a range from another instance after exhaustion. I noticed during testing that sometimes there are ranges with 1 free ID during the default behavior. This is a mitigation to avoid assigning too small ranges in the first place.

freeipa/src/main/resources/freeipa-salt/salt/freeipa/scripts/initdnarange.py
lacikaaa: the previously deleted FreeIPA instances' ranges won't show up when the ranges are listed. So they are free to take and they can be reassigned. I tested the scenario having 2 freeipa instance, A and B. Created users on A, repaired it, so now it's C. C got the ranges from A and it continued to work, using the next id from the range.

lacikaaa:
retest this please

PR #11013, opened 2021-07-06
CB-12871: Re-sizing data lake from light -> medium.

kkalvagadda1:
With the changes made one can re-build the data lake with a different shape. Below is the new flow chain for re-sizing.

**stop-existing DL -> Detach it from the environment -> Create new DL with requested shape -> delete the detached DL.**

There could be a failure in any of these flows. These changes don't have implementation for recovery from these failures.

There is a separate Jira created for changes needed to have a recovery mechanism.


sodre90 (changes requested):
can't see any integration test and e2e test, without these we can't merge it

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
```diff
@@ -230,6 +245,14 @@ protected void doExecute(SdxContext context, SdxCreateFailedEvent payload, Map<O
...
                     LOGGER.info("Can not set status to SDX_CREATION_FAILED because data lake was not found");
                 }
+                // When SDX is created as part of RESIZE flow chain, we should mark flow as failed to avoid
+                // subsequent flow's to be executed.
+                flowLogService.getLastFlowLog(context.getFlowParameters().getFlowId()).ifPresent(flowLog -> {
+                    if (flowLog.getFlowChainId() != null) {
+                        Flow flow = getFlow(context.getFlowParameters().getFlowId());
+                        flow.setFlowFailed(payload.getException());
+                    }
+                });
```
horadla23, line 255: marking flow failed can cause various issues, for example user may can click retry on it because of failed flow, also flow API will return with failed flow, but the creation actually not failed in this case, so I recommend other approach for this, because marking any flow failed without reason can be confusing

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
```diff
@@ -75,6 +85,11 @@ protected SdxContext createFlowContext(FlowParameters flowParameters, StateConte
 
             @Override
             protected void doExecute(SdxContext context, SdxEvent payload, Map<Object, Object> variables) throws Exception {
+                // When SDX is created as part of re-size flow chain, SDX in payload will not have ID.
+                if (!Strings.isNullOrEmpty(payload.getSdxName())) {
+                    SdxCluster sdxCluster = sdxService.getByNameInAccount(context.getUserId(), payload.getSdxName());
+                    context.setSdxId(sdxCluster.getId());
+                }
```
horadla23, line 92: what is the reason for this? based on description, this should be a new datalake in database and it should have id

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
```diff
@@ -17,13 +17,17 @@
...
-            LOGGER.info("Get datalake count in environment {}", stack.getEnvironmentCrn());
-            Long datalakesInEnv = stackService.countByEnvironmentCrnAndStackType(stack.getEnvironmentCrn(), StackType.DATALAKE);
-            if (datalakesInEnv > 1L) {
+            LOGGER.info("Get data lake count in environment {}", stack.getEnvironmentCrn());
+            Long datalakesInEnv = sdxClientService.getByEnvironmentCrn(stack.getEnvironmentCrn())
+                    .stream()
+                    .filter(sdxClusterResponse -> !sdxClusterResponse.isDetached())
+                    .filter(sdxClusterResponse -> !sdxClusterResponse.getCrn().equals(stack.getResourceCrn()))
+                    .count();
+            if (datalakesInEnv >= 1L) {
```
horadla23, line 30: why have you changed this? I think cloudbreak still contains information about datalake, and sdxClientService requires additional API call which makes validation longer

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/SdxStopActions.java
```diff
@@ -191,6 +192,8 @@ protected void doExecute(SdxContext context, SdxFailedEvent payload, Map<Object,
                 if (exception.getMessage() != null) {
                     statusReason = exception.getMessage();
                 }
+                Flow flow = getFlow(context.getFlowParameters().getFlowId());
+                flow.setFlowFailed(payload.getException());
```
horadla23, line 196: same here

datalake/src/main/java/com/sequenceiq/datalake/entity/SdxCluster.java
```diff
@@ -114,6 +114,9 @@
     @Column(name = "sdx_cluster_service_version")
     private String sdxClusterServiceVersion;
 
+    @Column(nullable = false)
+    private boolean detached;
+
```
horadla23, line 119: is it not possible to mark datalake as deleted and if it fails then mark it as available again? in this case we need to add this to every repository query every time

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -384,12 +390,68 @@ public SdxCluster getByNameInAccount(String userCrn, String name) {
         } catch (TransactionExecutionException e) {
             throw new TransactionRuntimeExecutionException(e);
         }
-
         FlowIdentifier flowIdentifier = sdxReactorFlowManager.triggerSdxCreation(savedSdxCluster);
-
         return Pair.of(savedSdxCluster, flowIdentifier);
     }
 
+    public Pair<SdxCluster, FlowIdentifier> resizeSdx(final String userCrn, final String clusterName, final SdxClusterResizeRequest sdxClusterResizeRequest) {
```
horadla23, line 397: this method has a lot in common with `create` method, is it not possible to create a common method for these two?

horadla23 (changes requested):
please add integration/e2e tests for this, as Peter suggested

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
kkalvagadda1: This piece of code is in "SDX_CREATION_FAILED_STATE" state. When the flow reaches this state, it is clear that the creation of SDX failed. I do not understand why you think creation flow enters this state without an actual failure?


Moreover, this logic is hit only when the create flow is part of a flow chain. This is will avoid any regression on current sdx creation.

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
kkalvagadda1: SdxCluster entity is created in SdxService.java but it is not persisted into the database immediately as there is a data lake already existing for the environment. It is persisted as part of detaching flow.

As it is not persisted at the flow chain creation, the payload(SdxCluster) here does not have the ID in it.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
kkalvagadda1: In case of re-sizing of the data lake, we are detaching the actual sdx and creating a new sdx with a different shape.

As far as CB is concerned, there is a data lake stack for the given environmentCrn. It does not anything about the sdx being dectached.

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/SdxStopActions.java
kkalvagadda1: This piece of code is in "SDX_STOP_FAILED_STATE" state. When the flow reaches this state, it is clear that the stop of SDX failed. I do not understand why you think stop flow enters this state without an actual failure?

datalake/src/main/java/com/sequenceiq/datalake/entity/SdxCluster.java
kkalvagadda1: I don't think it is the correct thing to mark it as deleted even it is not. Moreover, there will be other complications doing that.
1. There could multiple SDX Custers in a deleted state when you pick it in the recovery it will be tricky.
2.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
kkalvagadda1: There are similarities in initializing certain fields but overall there are a lot of diferences.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
horadla23: I see your point, but still I would think about some solution which can make us avoid API calls. In this case we would call datalake service in case of every datalake creation which can slow down all datalake creation.

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
horadla23: I see, it makes sense, thanks for the explanation!

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
horadla23: ohh, I haven't seen it is a FAILED state, then it should be ok, I think, sorry

datalake/src/main/java/com/sequenceiq/datalake/flow/stop/SdxStopActions.java
horadla23: same here, I haven't seen it is a FAILED state, then it should be ok, sorry

datalake/src/main/java/com/sequenceiq/datalake/entity/SdxCluster.java
horadla23: yeah, it makes sense, I just wanted to ask if it is possible to avoid usage of a new boolean field, because of repository calls, but if you think it is the best we can do, I think it is fine

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
horadla23: i see, then at least you can extract some similar parts into methods, like these:
```
        SdxCluster sdxCluster = new SdxCluster();
        sdxCluster.setInitiatorUserCrn(userCrn);
        sdxCluster.setCrn(createCrn(getAccountIdFromCrn(userCrn)));
        sdxCluster.setClusterName(name);
        sdxCluster.setAccountId(getAccountIdFromCrn(userCrn));
        sdxCluster.setClusterShape(sdxClusterRequest.getClusterShape());
        sdxCluster.setCreated(clock.getCurrentTimeMillis());
        sdxCluster.setEnvName(environment.getName());
        sdxCluster.setEnvCrn(environment.getCrn());
        sdxCluster.setRangerRazEnabled(sdxClusterRequest.isEnableRangerRaz());
        sdxCluster.setSdxClusterServiceVersion(sdxClusterServiceVersion);
```

```
        try {
            newSdxCluster.setStackRequest(JsonUtil.writeValueAsString(stackRequest));
        } catch (JsonProcessingException e) {
            LOGGER.error("Can not parse internal stackrequest", e);
            throw new BadRequestException("Can not parse internal stackrequest", e);
        }
```

and there are several method calls which is used in both methods, for example:
`prepareDefaultSecurityConfigs`
`MDCBuilder.buildMdcContext(sdxCluster)`
`validateEnv`
`validateMediumDutySdxEnablement`
`prepareCloudStorageForStack`
`stackRequest.setResourceCrn(sdxCluster.getCrn())`

I know there is lot of variables and method calls here, but I am sure it is possible to avoid some duplication in this case.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
horadla23: side note: you have changed the if condition from `if (datalakesInEnv > 1L)` to `if (datalakesInEnv >= 1L)` which leads to validation error in every case I think

horadla23: back to the `DataLakeValidator` topixc: it is used in `CloudConfigValidationActions`, I am not sure for what purpose, but I think, even if we call datalake service and validating if there is only 1 live datalake there, in CB it can cause issues if there is 2 non terminated datalake stacks for an environment, so since @doktoric created mostly the `CloudConfigValidationActions` class, I would ask him, if what should be done here to not break anything during cloud config validation

horadla23: in general I think it can cause issues if we have 2 non-terminated datalake stacks for an environment at the same time, like here:
https://github.com/hortonworks/cloudbreak/blob/6439d4fe3ef62d7d61cb5f3f813bf466902ec9b8/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackViewRepository.java#L28

sodre90: I aggree with @horadla23, you should check these kind of calls, it can cause troubles.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
sodre90: agree, there are lot of common things so we can put them into common methods, maybe we can move out these from SdxService to another service.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudStorageManifester.java
```diff
@@ -72,8 +74,21 @@ public CloudStorageRequest initCloudStorageRequest(DetailedEnvironmentResponse e
         return cloudStorageRequest;
     }
 
-    public CloudStorageRequest initSdxCloudStorageRequest(String cloudPlatform,
-            String blueprint, String clusterName, SdxCloudStorageRequest cloudStorage) {
+    public CloudStorageRequest initCloudStorageRequestFromExistingSdxCluster(DetailedEnvironmentResponse environment,
+            ClusterV4Response clusterV4Response, SdxCluster sdxCluster) {
+        boolean anyCloudStorageIsConfigured = !Strings.isNullOrEmpty(sdxCluster.getCloudStorageBaseLocation());
+        CloudStorageRequest cloudStorageRequest = new CloudStorageRequest();
```
sodre90, line 80: we can create a method in CloudStorageBase to copy base parameters from one CloudStorageBase to another CloudStorageBase. As I see request and response are the same so it would be enoght to call this "copy" method to copy all parameters.

sodre90 (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
kkalvagadda1: In this case, the stack for the data lake which is detached is not in the terminated state but is in the stopped state.
Trying to understand what kind of issue might be possible?
let me reach out to Richard on this

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
kkalvagadda1: will update.

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxController.java
```diff
@@ -138,6 +139,19 @@ public SdxClusterResponse create(String name, @Valid SdxCustomClusterRequest cre
         return sdxClusterResponse;
     }
 
+    @Override
+    @CheckPermissionByAccount(action = AuthorizationResourceAction.RESIZE_DATALAKE)
+    public SdxClusterResponse resize(String name, SdxClusterResizeRequest resizeSdxClusterRequest) {
+        String userCrn = ThreadBasedUserCrnProvider.getUserCrn();
+        Pair<SdxCluster, FlowIdentifier> result = sdxService.resizeSdx(userCrn, name, resizeSdxClusterRequest);
+        SdxCluster sdxCluster = result.getLeft();
+        metricService.incrementMetricCounter(MetricType.CUSTOM_SDX_REQUESTED, sdxCluster);
```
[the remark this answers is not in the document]
kkalvagadda1: Yes, it is correct. I overlooked it.

datalake/src/main/java/com/sequenceiq/datalake/flow/detach/event/DatalakeResizeFlowChainStartEvent.java
```diff
@@ -0,0 +1,23 @@
...
+public class DatalakeResizeFlowChainStartEvent extends SdxEvent {
+
+    private SdxCluster sdxCluster;
+
+    public DatalakeResizeFlowChainStartEvent(Long sdxId, SdxCluster newSdxCluster, String userId) {
+        super(sdxId, userId);
+        this.sdxCluster = newSdxCluster;
+    }
+
+    public SdxCluster getsdxCluster() {
```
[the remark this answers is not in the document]
kkalvagadda1: will update.

datalake/src/main/java/com/sequenceiq/datalake/flow/create/SdxCreateActions.java
```diff
@@ -230,6 +245,14 @@ protected void doExecute(SdxContext context, SdxCreateFailedEvent payload, Map<O
                 } catch (NotFoundException notFoundException) {
                     LOGGER.info("Can not set status to SDX_CREATION_FAILED because data lake was not found");
                 }
+                // When SDX is created as part of RESIZE flow chain, we should mark flow as failed to avoid
+                // subsequent flow's to be executed.
+                flowLogService.getLastFlowLog(context.getFlowParameters().getFlowId()).ifPresent(flowLog -> {
```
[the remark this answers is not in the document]
kkalvagadda1: I thought marking the flow as failed might impact the retry mechanism that is why I did I added the check to make sure no regression issue is introduced.

datalake/src/main/java/com/sequenceiq/datalake/flow/detach/SdxDetachActions.java
```diff
@@ -0,0 +1,175 @@
...
+
+            @Override
+            protected SdxContext createFlowContext(FlowParameters flowParameters, StateContext<FlowState, FlowEvent> stateContext, SdxStartDetachEvent payload) {
+                return SdxContext.from(flowParameters, payload);
+            }
+
+            @Override
+            protected void doExecute(SdxContext context, SdxStartDetachEvent payload, Map<Object, Object> variables) throws Exception {
+                LOGGER.info("Execute detach flow for SDX: {}", payload.getResourceId());
+                variables.put("SDX", payload.getSdxCluster());
```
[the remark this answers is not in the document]
kkalvagadda1: will do

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
horadla23: for example, this call will fail when an environment has a detached and a newer datalake: https://github.com/hortonworks/cloudbreak/blob/6439d4fe3ef62d7d61cb5f3f813bf466902ec9b8/core/src/main/java/com/sequenceiq/cloudbreak/repository/StackViewRepository.java#L28 (repository call will return more than one item)
this is just an example, but it was a general rule in CB that an environment has only 1 non-terminated datalake, but with this change in a small time interval there can be 2 non-terminated datalake for an environment.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/status/SdxStatusService.java
```diff
@@ -81,7 +81,7 @@ public SdxCluster setStatusForDatalakeAndNotify(DatalakeStatusEnum status, Resou
         }).orElseThrow(() -> new NotFoundException("SdxCluster was not found with ID: " + datalakeId));
     }
 
-    private void setStatusForDatalake(DatalakeStatusEnum status, String statusReason, SdxCluster sdxCluster) {
+    public void setStatusForDatalake(DatalakeStatusEnum status, String statusReason, SdxCluster sdxCluster) {
```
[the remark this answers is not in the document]
kkalvagadda1: SdxStatusService exposes API to update the status and notify the status to notify that status so that it can be rendered in the UI.

What I need in this case is an API that just updates the status and does not notify.

Till now there was no such need so setStatusForDatalake as a private method made sence. Now that that is need for an API why not make it a public API?

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
kkalvagadda1: yes. There will be 2 non-terminated data lakes for short period.

doktoric: Let me summarise the problem here:
The current DH creation logic works on a way that some of the DL config inherited into the DH cluster. Currently we provision a DH into and environment and automatically assign it the to DL cluster which is running in the Environment. IF we have n+1 DL in the Environment we can not handle this scenario because we don't know which DL should we use.
Question regarding this scenario:
- Can we provision a DH cluster under DL resize process?
- Can it cause any issue on experience side if we have multiple DL in the environment ? 
Also one more thing, I think the UI side also can handle multiple DL in an environment.

kkalvagadda1: You are right, DH uses the data lake stack associated with the environment CRN to fetch the remote data context of DL.  If DH fetches remote data context of DL while DL re-sizing is in progress it will cause issues. DL re-size operation is performed after the data hub is stopped. One of the first steps in the re-size flow chain is to stop DL and DH clusters.

Service discovery internally depends on getByEnvCrn API expose by SdxEndpoint to the SDx information. We are filtering out data lakes that are detached before returning the list. we are good on that front as well.

kkalvagadda1:
> Can't see any integration test and e2e test. Without these we can't merge it, I don't want to suffer the same as for HA medium-duty, where repair and upgrade scenarios were not tested and we had to disable and rework the feature after it was released.

Hello Peter, These changes are not ready to be delivered to the customers. This is one of the first commits. Other commits will follow this. I tried to get the e2e tests working locally but will have the tests added in the next few weeks. I have created a separate task for it. Let's not block this patch for e2e tests.

core/src/main/java/com/sequenceiq/cloudbreak/controller/validation/datalake/DataLakeValidator.java
horadla23: > One of the first steps in the re-size flow chain is to stop DL and DH clusters.

is this ensured somehow? and what about experiences? we need to make sure somehow that everything is stopped and we also need to make sure that it is not possible to start a new datahub during resize

horadla23: nvm,sdx stop will stop all datahubs

horadla23 (approved):
I am fine with this change, it is behind an entitlement, before we turn on entitlement, we need these tickets to be implemented:
CB-12872 Handle all the failure scenarions that occur while migrating from light->medium size.
CB-13414 Add e2e tests for re-sizing of the datalake.

sodre90:
We can merge it but please create a jira ticket, not only for e2e but please do mock integration tests also.

kkalvagadda1:
> We can merge it but please create a jira ticket, not only for e2e but please do mock integration tests also.

done.

horadla23 (approved):

sodre90 (approved):

PR #11014, opened 2021-07-06
CB-13244 Check cloud provider vm statuses during conclusion checks

lturcsanyi:
If CM is running get host statuses from CM, otherwise get status information from provider.


horadla23 (approved):
LGTM

PR #11016, opened 2021-07-06
CB-12696 Handle CCMV2_JUMPGATE tunnel type

bergerdenes:
Determine flag to send down to user data and Salt pillar.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
@@ -142,13 +142,13 @@ public class EnvironmentCreationService {
     private void initializeEnvironmentTunnel(Environment environment) {
         Tunnel tunnel = environment.getExperimentalFeaturesJson().getTunnel();
         boolean ccmV2Enabled = entitlementService.ccmV2Enabled(environment.getAccountId());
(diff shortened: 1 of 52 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -142,13 +142,13 @@ private Environment initializeEnvironment(EnvironmentCreationDto creationDto) {
     private void initializeEnvironmentTunnel(Environment environment) {
         Tunnel tunnel = environment.getExperimentalFeaturesJson().getTunnel();
         boolean ccmV2Enabled = entitlementService.ccmV2Enabled(environment.getAccountId());
-        if (Tunnel.CCMV2 == tunnel && !ccmV2Enabled) {
+        if ((Tunnel.CCMV2 == tunnel || Tunnel.CCMV2_JUMPGATE == tunnel) && !ccmV2Enabled) {
```
doktoric, line 145: useCcmV2OrJumpgate can be used here right ?

bergerdenes: right, if tunnel is not null (which I hope is not null)

doktoric (approved):

PR #11023, opened 2021-07-08
CB-12913: Add backup and restore policies for S3.

brycederriso:
Closes [CB-12913](https://jira.cloudera.com/browse/CB-12913).

Check the AWS policy for Datalake backup + restore into cloudbreak so that we can refer to the policies from the public facing documentation.

I arrived at these values by creating a bucket outside of the normal permissions, and running backup restore while whittling the permissions down.

These policies are more permissive than they absolutely have to be, but they're also simpler for it.

The most restrictive/specific policies would include separate policies for Admin, Logger, and Ranger roles when restoring.

This issue is related to updating the public facing docs in [CB-12981](https://jira.cloudera.com/browse/CB-12981).

### DIFF
--- a/cloud-aws-cloudformation/src/main/resources/definitions/aws-datalake-backup-policy.json
+++ b/cloud-aws-cloudformation/src/main/resources/definitions/aws-datalake-backup-policy.json
@@ -1,0 +1,17 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "DatalakeBackupPermissions",
+      "Effect": "Allow",
+      "Action": [
+        "s3:PutObject",
+        "s3:GetObject",
+        "s3:DeleteObject"
+      ],
+      "Resource": [
+        "arn:aws:s3:::<BACKUP_BUCKET>/*"
+      ]
+    }
+  ]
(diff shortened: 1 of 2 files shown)


kkalvagadda1 (approved):
Looks good.

doktoric:
this was merged but we already added every policy to the credential prerequisite json. Can you please add these @brycederriso ?

brycederriso:
@doktoric 
> this was merged but we already added every policy to the credential prerequisite json. Can you please add these @brycederriso ?

It looks like the [prerequisites](https://github.com/hortonworks/cloudbreak/blob/master/cloud-aws-common/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/common/AwsCredentialConnector.java#L111) are truly prerequisites for setting up a data lake, that's not quite what we intend for the backup location.

We want customers to have a place to reference these policies, but they don't need to be provided at Data Lake creation time, we're just planning on referring to them from the backup/restore documentation.

They're being added to the `cdp` folder of `policy.json`s because they mostly fit and we need a place to check them into VCS. But, they're not really intended to be used by the CF when setting things up.

PR #11024, opened 2021-07-08
CB-13108:Add validations for accepting CMK from create-aws-environment cli

satyabollineni:

### DIFF
--- a/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
+++ b/environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
@@ -1,0 +1,72 @@
+package com.sequenceiq.environment.api.v1.environment.model.request.aws;
+
+import java.io.Serializable;
+
+import javax.validation.constraints.Pattern;
+import com.google.common.annotations.VisibleForTesting;
+
+import com.sequenceiq.environment.api.doc.environment.EnvironmentModelDescription;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(value = "AwsDiskEncryptionV1Parameters")
+public class AwsDiskEncryptionParameters implements Serializable {
+
+    @VisibleForTesting
+    static final String AWS_ENCRYPTION_KEY_INVALID_MESSAGE = "The identifier of the AWS Key Management Service (AWS KMS) " +
+            "customer master key (CMK) to use for Amazon EBS encryption.\n" +
+            "You can specify the key ARN in the below format:\n" +
+            "Key ARN: arn:partition:service:region:account-id:resource-type/resource-id. " +
+            "For example, arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab.\n";
+
+    @VisibleForTesting
+    static final String REG_EXP_FOR_ENCRYPTION_KEY_ARN =
+            "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$";
+
+    @Pattern(regexp = REG_EXP_FOR_ENCRYPTION_KEY_ARN,
+            message = AWS_ENCRYPTION_KEY_INVALID_MESSAGE)
+    @ApiModelProperty(EnvironmentModelDescription.AWS_DISK_ENCRYPTION_PARAMETERS)
+    private String encryptionKeyArn;
+
+    public AwsDiskEncryptionParameters() {
+    }
+
+    private AwsDiskEncryptionParameters(Builder builder) {
+        this.encryptionKeyArn = builder.encryptionKeyArn;
+    }
+
+    public String getEncryptionKeyArn() {
+        return encryptionKeyArn;
+    }
+
+    public void setEncryptionKeyArn(String encryptionKeyArn) {
+        this.encryptionKeyArn = encryptionKeyArn;
+    }
+
+    @Override
+    public String toString() {
+        return "AwsDiskEncryptionParameters{" +
+                "encryptionKeyArn=" + encryptionKeyArn + '}';
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static final class Builder {
+        private String encryptionKeyArn;
+
+        private Builder() {
+        }
+
+        public Builder withEncryptionKeyArn(String encryptionKeyArn) {
+            this.encryptionKeyArn = encryptionKeyArn;
+            return this;
+        }
+
+        public AwsDiskEncryptionParameters build() {
+            return new AwsDiskEncryptionParameters(this);
+        }
+    }
+}
--- a/environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
+++ b/environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
@@ -1,0 +1,90 @@
+package com.sequenceiq.environment.api.v1.environment.model.request.aws;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class AwsDiskEncryptionParametersTest {
+
+    private static final String ENCRYPTION_KEY_ARN = "encryptionKeyArn";
+
+    @Test
+    void defaultConstructorTest() {
+        AwsDiskEncryptionParameters underTest = new AwsDiskEncryptionParameters();
+        underTest.setEncryptionKeyArn(ENCRYPTION_KEY_ARN);
+        verifyFields(underTest);
+    }
+
+    private void verifyFields(AwsDiskEncryptionParameters underTest) {
+        assertThat(underTest.getEncryptionKeyArn()).isEqualTo(ENCRYPTION_KEY_ARN);
+    }
+
+    @Test
+    void builderTest() {
+        AwsDiskEncryptionParameters underTest = AwsDiskEncryptionParameters.builder()
+                .withEncryptionKeyArn(ENCRYPTION_KEY_ARN)
+                .build();
+        verifyFields(underTest);
+    }
+
+    static Object[][] encryptionKeyArnValidationTestDataProvider() {
+        return new Object[][]{
+                // testCaseName encryptionKeyArn expectedValid
+                {"encryptionKeyArnWithoutKeyID/", "arn:aws:kms:us-east-1:012345678910:key/", false},
+
+                {"EncryptionKeyArnWithoutPartition",
+                        "aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithNonNumericAccountId",
+                        "arn:aws:kms:us-east-1:abc:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithInvalidServiceName",
+                        "arn:aws:role:us-east-1:012345678910/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithInvalidRegion",
+                        "arn:aws-us-gov:kms:us-east/2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithSpecialCharInAccountId",
+                        "arn:aws-us-gov:kms:us-east-2:0123456789-:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithoutRegion",
+                        "arn:aws-us-gov:kms::0123456789:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"encryptionKeyArnWithoutAccountId",
+                        "arn:aws-us-gov:kms:us-east-1::key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+
+                {"EncryptionKeyArnWithInvalidKeyIDSubStringCount",
+                        "arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd", false},
+
+                {"EncryptionKeyArnWithInvalidKeyIDSubStringLength",
+                        "arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56-1234567890ab", false},
+
+                {"encryptionKeyArnWithValidKey",
+                        "arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+
+                {"encryptionKeyArnWithPossibleValidPartition",
+                        "arn:aws-us-gov:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+
+                {"encryptionKeyArnWithPossibleValidPartition",
+                        "arn:aws-cn:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+
+                {"encryptionKeyArnWithPossibleValidRegion",
+                        "arn:aws-us-gov:kms:us-east-2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+        };
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("encryptionKeyArnValidationTestDataProvider")
+    void encryptionKeyArnValidationTest(String testCaseName, String encryptionKeyArn, boolean expectedValid) {
+        if (expectedValid) {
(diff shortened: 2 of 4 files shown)


lajosrodek:
@satyabollineni The commit history seems to be messed up, there are two unrelated commits.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AWSDiskEncryptionParametersTest.java
```diff
@@ -0,0 +1,101 @@
...
+                        "arn:aws-us-gov:role:us-east-2:0123456789-:key/1234abcd-12ab-34cd-56ef-1234567890ab", false},
+                {"encryptionKeyArn=arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+                        "arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+                {"encryptionKeyArn=arn:aws-us-gov:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+                        "arn:aws-us-gov:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+                {"encryptionKeyArn=arn:aws-cn:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+                        "arn:aws-cn:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+                {"encryptionKeyArn=arn:aws-us-gov:kms:us-east-2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+                        "arn:aws-us-gov:kms:us-east-2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+
```
lajosrodek, line 75: Perhaps add one or two more cases for the empty region & account ID scenario (both of which should result in a validation error).

lajosrodek (changes requested):
Some asks. Also a general q: As I see, processing of the new ARN field is still missing, so e.g. it is not set in the DTO nor in the entity, not persisted to the DB. I suppose these parts will come in subsequent PR(s), perhaps as part of CB-13110.

satyabollineni:
> Some asks. Also a general q: As I see, processing of the new ARN field is still missing, so e.g. it is not set in the DTO nor in the entity, not persisted to the DB. I suppose these parts will come in subsequent PR(s), perhaps as part of CB-13110.

Yes, Just splitting the PR's so that review would be easier and faster. I will make the changes in DTO files in parallel.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+class AWSDiskEncryptionParametersTest {
+
+    private static final String ENCRYPTION_KEY_ARN = "encryptionKeyArn";
+
+    private LocalValidatorFactoryBean localValidatorFactory;
+
+    @BeforeEach
+    void setUp() {
+        localValidatorFactory = new LocalValidatorFactoryBean();
```
bergerdenes, line 24: You are testing the whole Pattern validator here not only your validation regexp.
It would be more concise if you just extract your regexp and run the checks in `encryptionKeyArnValidationTest` against the regexp only (i.e. matches or not)

bergerdenes (changes requested):

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: I have made the changes accordingly. Hope now this looks good.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
```diff
@@ -97,19 +81,12 @@ void encryptionKeyArnValidationTest(String testCaseName, String encryptionKeyArn
                 .withEncryptionKeyArn(encryptionKeyArn)
                 .build();
 
-        Set<ConstraintViolation<AwsDiskEncryptionParameters>> constraintViolations = localValidatorFactory.validate(underTest);
+        String regex = "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key/[a-zA-Z0-9-]+$";
```
bergerdenes, line 84: the regex should come from the tested class!

bergerdenes (changes requested):

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: @bergerdenes I have explored from my end and It seems I could not extract just the regex from pattern annotator, Instead I should be using ValidatorFactory to validate this. Could you please help me If there is any other way that I am missing here to just extract the regex( I could not re-define a local variable of type pattern in the source file as I am already using the pattern annotator for validation  there)

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
lajosrodek: @bergerdenes  I am late to the party but would like to disagree. Testing the validator itself is more valuable (like an integration test), that way we can be sure that annotations are not screwed up and so the validation will be indeed applied correctly. Satya's original approach is a direct adaptation of [AzureResourceEncryptionParametersTest](https://github.com/hortonworks/cloudbreak/blob/master/environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/azure/AzureResourceEncryptionParametersTest.java), for which I must be blamed. 😄 There are also some further examples where `LocalValidatorFactoryBean` was used in a similar fashion.

In short, I suggest reverting to the original code.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
lajosrodek: As mentioned in my other comment, I do not feel it is right to rely on regex matching in the UT, since the Hibernate validator already takes care of that for us (if used, of course).

bergerdenes: I think by testing the validator, we are not testing purely the pattern, but the validator as well (which is totally unnecessary as there is no new validator implemented here)
You can extract the regex by adding a static final String to your class (a package-private one) and you can use that from your unit test.

lajosrodek (changes requested):
Two small asks. Please also apply Denes's fixes; I am happy with either approach (Hibernate validator or regex).

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
```diff
@@ -0,0 +1,68 @@
...
+public class AwsDiskEncryptionParameters implements Serializable {
+
+    @VisibleForTesting
+    static final String AWS_ENCRYPTION_KEY_INVALID_MESSAGE = "The identifier of the AWS Key Management Service (AWS KMS) " +
+            "customer master key (CMK) to use for Amazon EBS encryption.\n" +
+            "You can specify the key ARN in the below format:\n" +
+            "Key ARN: arn:partition:service:region:account-id:resource-type/resource-id. " +
+            "For example, arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab.\n";
+
+    @Pattern(regexp = "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key/[a-zA-Z0-9-]+$",
```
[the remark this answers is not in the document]
satyabollineni: Added the GUID validation in the regex for the key part

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: Changed the test case to regex check as per the review comments.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
```diff
@@ -0,0 +1,74 @@
...
+            "customer master key (CMK) to use for Amazon EBS encryption.\n" +
+            "You can specify the key ARN in the below format:\n" +
+            "Key ARN: arn:partition:service:region:account-id:resource-type/resource-id. " +
+            "For example, arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab.\n";
+
+    static final String REG_EXP_FOR_TESTING =
+            "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key\\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}+$";
+
+    @Pattern(regexp =
+            "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key\\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}+$",
```
lajosrodek, line 27: Sorry, why duplicate the regexp? `regexp = REG_EXP_FOR_TESTING` should be feasible right? 😄

Moreover, the name of the constant is not really expressive. It is a regexp for what? The fact that it is main purpose is to facilitate easier testing should be expressed with `@VisibleForTesting` as used for `AWS_ENCRYPTION_KEY_INVALID_MESSAGE` above.

lajosrodek (changes requested):
Generally looks okay, except for one minor issue. Please also squash your stuff into a single commit.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
satyabollineni: That was intentional while adding a new static variable for testing, Agreed that It could be simulated to AWS_ENCRYPTION_KEY_INVALID_MESSAGE. Made the changes accordingly.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @VisibleForTesting
+    static final String AWS_ENCRYPTION_KEY_INVALID_MESSAGE = "The identifier of the AWS Key Management Service (AWS KMS) " +
+            "customer master key (CMK) to use for Amazon EBS encryption.\n" +
+            "You can specify the key ARN in the below format:\n" +
+            "Key ARN: arn:partition:service:region:account-id:resource-type/resource-id. " +
+            "For example, arn:aws:kms:us-east-1:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab.\n";
+
+    static final String REG_EXP_FOR_ENCRYPTION_KEY_ARN =
+            "^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:key\\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}+$";
```
lajosrodek, line 24: Missing `@VisibleForTesting`

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParameters.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Pattern(regexp = REG_EXP_FOR_ENCRYPTION_KEY_ARN,
+            message = AWS_ENCRYPTION_KEY_INVALID_MESSAGE)
+
+    @ApiModelProperty(EnvironmentModelDescription.AWS_DISK_ENCRYPTION_PARAMETERS)
+
+    private String encryptionKeyArn;
```
lajosrodek, line 31: Unnecessary empty lines between annotations and their targets. Please remove the empty lines 28 and 30, it makes no sense for the annotations to stand separated.

lajosrodek (changes requested):

satyabollineni:
@bergerdenes My local gradle build was success but the PR bot is continuously failing for the same error. I tried adding escape characters  in the regex but no luck in PR bot(local build passes without escape characters as well. Could you please help me with some pointers to avoid this PR bot failure?

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: Resolving this conversation as per https://github.com/hortonworks/cloudbreak/pull/11024#discussion_r668716966

lajosrodek (approved):
LGTM

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
```diff
@@ -0,0 +1,84 @@
+package com.sequenceiq.environment.api.v1.environment.model.request.aws;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.Assert.assertFalse;
```
bergerdenes, line 7: please do not mix Junit4 and Junit5, use Junit5's Assertions.assertFalse and assertTrue

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
```diff
@@ -0,0 +1,84 @@
...
+                {"encryptionKeyArn=arn:aws-us-gov:kms:us-east-2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+                        "arn:aws-us-gov:kms:us-east-2:012345678910:key/1234abcd-12ab-34cd-56ef-1234567890ab", true},
+        };
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("encryptionKeyArnValidationTestDataProvider")
+    void encryptionKeyArnValidationTest(String testCaseName, String encryptionKeyArn, boolean expectedValid) {
+        if (expectedValid) {
+            assertTrue(AwsDiskEncryptionParameters.AWS_ENCRYPTION_KEY_INVALID_MESSAGE,
```
bergerdenes, line 77: If the assertion fails, then tell so. `AWS_ENCRYPTION_KEY_INVALID_MESSAGE` is for the end-user. Which also means that this constant can be `private static final`
Assertion message should state what and why failed. Like `encryptionKeyArn + " should match pattern " +AwsDiskEncryptionParameters.REG_EXP_FOR_ENCRYPTION_KEY_ARN`, and similar for the negative assert.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
```diff
@@ -0,0 +1,84 @@
...
+        AwsDiskEncryptionParameters underTest = AwsDiskEncryptionParameters.builder()
+                .withEncryptionKeyArn(ENCRYPTION_KEY_ARN)
+                .build();
+        verifyFields(underTest);
+    }
+
+    static Object[][] encryptionKeyArnValidationTestDataProvider() {
+        return new Object[][]{
+                // testCaseName encryptionKeyArn expectedValid
+                {"encryptionKeyArn=arn:aws:kms:us-east-1:012345678910:key/", "arn:aws:kms:us-east-1:012345678910:key/", false},
```
bergerdenes, line 36: Your test case name does not say anything about what going to be tested. Like: missing resource, missing servicename, missing accountid, wrong accountid, wrong UUID in key, etcetc.
I do not see test for a wrong UUID for example (maybe it is there but hard to see)

bergerdenes (changes requested):
Small glitches in the test

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: Added the meaningful test case names.

environment-api/src/test/java/com/sequenceiq/environment/api/v1/environment/model/request/aws/AwsDiskEncryptionParametersTest.java
satyabollineni: Addressed

bergerdenes (approved):

PR #11027, opened 2021-07-08
CB-6875 CM should give back services health for hosts and CB should u…

horadla23:
…se it to detect if a host is healthy or not


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
```diff
@@ -238,15 +242,36 @@ public ClusterStatusResult getStatus(boolean blueprintPresent) {
...
         Map<HostName, ClusterManagerState> clusterManagerStateByHost = hostHealth.entrySet().stream()
                 .flatMap(this::getScmHealthForHost)
                 .collect(toMap(Pair::getLeft, Pair::getRight));
         boolean hostCertExpiring = isHostCertExpiring(hostHealth);
+        checkServicesHealth(apiHostList, clusterManagerStateByHost);
         LOGGER.debug("Creating 'ExtendedHostStatuses' with {} and cert expiring [{}]", clusterManagerStateByHost, hostCertExpiring);
         return new ExtendedHostStatuses(clusterManagerStateByHost, hostCertExpiring);
     }
 
+    private void checkServicesHealth(List<ApiHost> apiHostList, Map<HostName, ClusterManagerState> clusterManagerStateByHost) {
```
lacikaaa, line 256: I did some refactoring:
```
    private void checkServicesHealth(List<ApiHost> apiHostList, Map<HostName, ClusterManagerState> clusterManagerStateByHost) {
        apiHostList.forEach(host -> changeStateForHostIfServiceInBadHealth(clusterManagerStateByHost, host));
    }

    private void changeStateForHostIfServiceInBadHealth(Map<HostName, ClusterManagerState> clusterManagerStateByHost, ApiHost host) {
        Set<String> servicesWithBadHealth = collectServicesWithBadHealthOnHost(host);
        if (!servicesWithBadHealth.isEmpty()) {
            updateServicesHealthForHost(clusterManagerStateByHost, host, servicesWithBadHealth);
        }
    }

    private Set<String> collectServicesWithBadHealthOnHost(ApiHost host) {
        return host.getRoleRefs().stream()
                .filter(roleRef -> ApiHealthSummary.BAD.equals(roleRef.getHealthSummary()))
                .map(ApiRoleRef::getServiceName)
                .collect(Collectors.toSet());
    }

    private void updateServicesHealthForHost(Map<HostName, ClusterManagerState> clusterManagerStateByHost, ApiHost host, Set<String> servicesWithBadHealth) {
        ClusterManagerState clusterManagerState = clusterManagerStateByHost.get(hostName(host.getHostname()));
        clusterManagerState.setClusterManagerStatus(ClusterManagerStatus.UNHEALTHY);
        String statusReason = String.format("The following services are in bad health: %s", Joiner.on(",").join(servicesWithBadHealth));
        if (StringUtils.isEmpty(clusterManagerState.getStatusReason())) {
            clusterManagerState.setStatusReason(statusReason);
        } else {
            StringUtils.appendIfMissing(clusterManagerState.getStatusReason(), statusReason);
        }
    }
```
but I still think this is not the right solution.
It might be better to modify `getScmHealthForHost` and `createClusterManagerStateFromApiHealthCheck` to include the service health check also, so you won't have to update the map after it's created.

lacikaaa (changes requested):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
horadla23: the pain here is that `ApiRoleRef` class does not have `ApiHealthCheck` field, only `ApiHealthSummary`, so making `getScmHealthForHost` able to check services too would require the refactor of at least half of this whole service. if you think it's worth to do so, I can do it

lacikaaa: not necessary if it's that painful

lacikaaa (approved):

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerClusterStatusService.java
horadla23: I would not refactor `getScmHealthForHost` to be honest

health of agent contains 
- name of the health check (known)
- health status 
- explanation

health of services consist only of 
- name of service (unknown) 
- health status

so the logic of checking the health and create error message based on health is quite different in case of agent check and in case of service check

horadla23:
retest this please

PR #11033, opened 2021-07-09
CB-13357. Remove 'all' param from IPA group find command

handavid:
Group-find with the 'all' parameter is very slow. Anecdotally, for ~7000
groups, it takes > 5 minutes while running without the 'all' parameter
only takes ~10 seconds. The 'all' parameter was removed from FreeIpaClient.groupFindAll
to improve user sync performance.

A search of the codebase revealed only two usages of the FreeIpaClient.groupFindAll
command: user sync and the freeipa client test service.

The freeipa client test service only needs to retrieve a single group at a time. This
code was updated to use group-show instead of group-find.

User sync only requires the group names when finding all the groups and does not need
the other infromation returned with the 'all' parameter.


gitmfox (approved):
Nice!

horadla23 (approved):
LGTM

PR #11034, opened 2021-07-09
CDPCP-4911. User user-show instead of user-find when retrieving a sin…

handavid:
…gle user from ipa

user-show is faster than user-find because user-find searches through multiple fields
and tries to return all matching users. Single user sync needs to retrieve individual
users and should use user-show instead of user-find.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/FreeIpaUsersStateProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/FreeIpaUsersStateProvider.java
@@ -6,6 +6,8 @@ import static com.sequenceiq.freeipa.client.FreeIpaChecks.IPA_UNMANAGED_GROUPS;
 import java.util.Optional;
 import java.util.Set;
 
+import com.sequenceiq.freeipa.client.FreeIpaClientExceptionUtil;
+import com.sequenceiq.freeipa.client.model.User;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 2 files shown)


gitmfox:
When we move to searching for users by CRN (CDPCP-2836), won't we need to go back to using user-find?

freeipa/src/main/java/com/sequenceiq/freeipa/service/freeipa/user/FreeIpaUsersStateProvider.java
```diff
@@ -66,9 +66,8 @@ public UsersState getFilteredFreeIpaState(FreeIpaClient freeIpaClient, Set<Strin
             if (IPA_PROTECTED_USERS.contains(userName)) {
                 continue;
             }
-            Optional<com.sequenceiq.freeipa.client.model.User> ipaUserOptional = freeIpaClient.userFind(userName);
-            if (ipaUserOptional.isPresent()) {
-                com.sequenceiq.freeipa.client.model.User ipaUser = ipaUserOptional.get();
+            try {
+                com.sequenceiq.freeipa.client.model.User ipaUser = freeIpaClient.userShow(userName);
```
lacikaaa, line 70: you can use `FreeIpaClientExceptionUtil#ignoreNotFoundExceptionWithValue`

handavid: done

handavid:
> When we move to searching for users by CRN (CDPCP-2836), won't we need to go back to using user-find?

Mark,
I was thinking about that also. I think we are still ok with user-show because the crn->workload username mapping will be consistent. The title is still returned as part of user-show so we can check if the crn matches at that point rather than finding using the crn.

PR #11039, opened 2021-07-12
CB-13367 network conversion in case of ycloud must be handled. Nullpo…

doktoric:
…inter noew fixed


lajosrodek (approved):
LGTM. Please add some UTs (potentially in a follow-up PR).

PR #11046, opened 2021-07-12
CB-6875 CM should give back services health for hosts and CB should use it to detect if a host is healthy or not

horadla23:


horadla23:
retest this please

lacikaaa:
where is the IT mock part?

horadla23:
> where is the IT mock part?

i decided to not add the new swagger, it generates 10000 lines of new code and i still can only generate fake data for services in mock, so it doesn't worth to add and the actual problem can be solved with a null-check in 

> where is the IT mock part?

I decided to not update swagger in mock, it generates 10000 lines of new code and the issue can be solved with a null-check in `collectServicesWithBadHealthOnHost`. Until we do not add integration test for this, regeneration of CM API in mock doesn't needed, and I think covering this implementation with unit test is enough.

lacikaaa (approved):

PR #11048, opened 2021-07-12
CB-13318: Automatic backup should be triggered conditionally.

kkalvagadda1:
Currently, backup is triggered on every upgrade if the necessary entitlement is enabled.

Instead, it should be invoked conditionally only in the use cases it is supported.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
@@ -14,26 +14,28 @@ import static com.sequenceiq.datalake.flow.repair.SdxRepairEvent.SDX_REPAIR_EVEN
 import static com.sequenceiq.datalake.flow.start.SdxStartEvent.SDX_START_EVENT;
 import static com.sequenceiq.datalake.flow.stop.SdxStopEvent.SDX_STOP_EVENT;
 
+import java.util.Comparator;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import javax.inject.Inject;
 
-import com.sequenceiq.datalake.flow.dr.restore.event.DatalakeTriggerRestoreEvent;
-import com.sequenceiq.datalake.service.EnvironmentClientService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.auth.altus.EntitlementService;
+import com.sequenceiq.cloudbreak.cloud.VersionComparator;
 import com.sequenceiq.cloudbreak.common.event.Acceptable;
+import com.sequenceiq.cloudbreak.common.type.Versioned;
+import com.sequenceiq.cloudbreak.datalakedr.config.DatalakeDrConfig;
 import com.sequenceiq.cloudbreak.exception.CloudbreakApiException;
 import com.sequenceiq.cloudbreak.exception.FlowNotAcceptedException;
 import com.sequenceiq.cloudbreak.exception.FlowsAlreadyRunningException;
-import com.sequenceiq.cloudbreak.datalakedr.config.DatalakeDrConfig;
 import com.sequenceiq.datalake.entity.SdxCluster;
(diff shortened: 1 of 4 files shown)


datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
```diff
@@ -105,7 +108,7 @@ public FlowIdentifier triggerDatalakeRuntimeUpgradeFlow(SdxCluster cluster, Stri
         LOGGER.info("Trigger Datalake runtime upgrade for: {} with imageId: {} and replace vm param: {}", cluster, imageId, replaceVms);
         String userId = ThreadBasedUserCrnProvider.getUserCrn();
         if (entitlementService.isDatalakeBackupOnUpgradeEnabled(ThreadBasedUserCrnProvider.getAccountId()) &&
-                datalakeDrConfig.isConfigured()) {
+                datalakeDrConfig.isConfigured() && shouldSdxBackupBePerformed(cluster)) {
```
brycederriso, line 111: **praise:** Thanks for extracting this boolean to a function!

datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
```diff
@@ -116,6 +119,35 @@ public FlowIdentifier triggerDatalakeRuntimeUpgradeFlow(SdxCluster cluster, Stri
...
+    private boolean shouldSdxBackupBePerformed(SdxCluster cluster) {
+        boolean retVal = true;
+        String reason = null;
+        if (cluster.isRangerRazEnabled()) {
+            retVal = false;
+            reason = "RAZ is enabled";
+        } else if (isVersionOlderThan(cluster, "7.2.1")) {
+            retVal = false;
+            reason = "Unsupported runtime: " + cluster.getRuntime();
+        } else if (cluster.getCloudStorageFileSystemType().isGcs()) {
```
brycederriso, line 131: **suggestion:** I think it would be worth calling out that we're using the storage file system type as a proxy for the cloud provider. Maybe in a javadoc comment on the method.

datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
```diff
@@ -116,6 +119,35 @@ public FlowIdentifier triggerDatalakeRuntimeUpgradeFlow(SdxCluster cluster, Stri
         }
     }
 
+    private boolean shouldSdxBackupBePerformed(SdxCluster cluster) {
+        boolean retVal = true;
+        String reason = null;
+        if (cluster.isRangerRazEnabled()) {
+            retVal = false;
+            reason = "RAZ is enabled";
```
brycederriso, line 127: **suggestion:** consider making the reason `"backup is not supported with RAZ enabled"`, I think that makes it clearer that we're explicitly _not_ supporting backups with RAZ.

brycederriso (approved):
I have two small suggestions, and the changes look good to me otherwise.

hreeve-cloudera (approved):

datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
kkalvagadda1: done.

datalake/src/main/java/com/sequenceiq/datalake/flow/SdxReactorFlowManager.java
```diff
@@ -116,6 +119,41 @@ public FlowIdentifier triggerDatalakeRuntimeUpgradeFlow(SdxCluster cluster, Stri
...
+            retVal = false;
+            reason = "Unsupported cloud provider Azure on runtime: " + cluster.getRuntime();
+        }
+        if (!retVal) {
+            LOGGER.info("Backup not triggered. Reason: " + reason);
+        }
+        return retVal;
+}
+
+    private static boolean isVersionOlderThan(SdxCluster cluster, String baseVersion) {
```
pdarvasi, line 151: pls use CMRepositoryVersionUtil#isVersionOlderThanLimited() instead

kkalvagadda1: Both of these using the same versionComparator. I initially started using the same API. I had to make the validation login complex if i had to use CMRepositoryVersionUtil that is why i deliberately not use it.

PR #11057, opened 2021-07-14
CB-13391. Fix circular dep for freeipa between flow and operation ser…

oleewere:
…vice beans.

details:
- there is already an OperationService that was used inseide flows
- new OperationApi was added notso long - this operationService was re-used, and overlap with curreent operations (by using different objects)
- From that point operationService has the flowService which has dependencies for FlowConfigs, but in many freeipa flow configs, operationService is referenced
- this issue can be solved by separate the OperationApi related code from the original operation usage.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/controller/OperationV1Controller.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/controller/OperationV1Controller.java
@@ -19,6 +19,7 @@ import com.sequenceiq.flow.api.model.operation.OperationView;
 import com.sequenceiq.freeipa.api.v1.operation.OperationV1Endpoint;
 import com.sequenceiq.freeipa.api.v1.operation.model.OperationStatus;
 import com.sequenceiq.freeipa.converter.operation.OperationToOperationStatusConverter;
+import com.sequenceiq.freeipa.service.operation.FlowOperationService;
 import com.sequenceiq.freeipa.service.operation.OperationService;
 
 @Controller
@@ -26,6 +27,9 @@ public class OperationV1Controller implements OperationV1Endpoint {
 
     @Inject
     private OperationService operationService;
+
+    @Inject
+    private FlowOperationService flowOperationService;
 
     @Inject
     private OperationToOperationStatusConverter operationToOperationStatusConverter;
@@ -40,6 +44,6 @@ public class OperationV1Controller implements OperationV1Endpoint {
     @Override
     @CheckPermissionByResourceCrn(action = DESCRIBE_ENVIRONMENT)
     public OperationView getOperationProgressByEnvironmentCrn(@ResourceCrn String resourceCrn, boolean detailed) {
-        return operationService.getOperationProgressByEnvironmentCrn(resourceCrn, detailed);
+        return flowOperationService.getOperationProgressByEnvironmentCrn(resourceCrn, detailed);
     }
 }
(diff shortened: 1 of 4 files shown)


oleewere:
issue that happened because of this:
```text
***************************
APPLICATION FAILED TO START
***************************
Description:
The dependencies of some of the beans in the application context form a cycle:
   flowLogsToListDiagnosticsCollectionResponseConverter (field private com.sequenceiq.flow.core.config.FlowProgressHolder com.sequenceiq.freeipa.converter.diagnostics.FlowLogsToListDiagnosticsCollectionResponseConverter.flowProgressHolder)
┌─────┐
|  flowProgressHolder defined in URL [jar:file:/freeipa.jar!/BOOT-INF/lib/flow.jar!/com/sequenceiq/flow/core/config/FlowProgressHolder.class]
↑     ↓
|  rebootFlowConfig
↑     ↓
|  REBOOT_FAILED_STATE (field private com.sequenceiq.freeipa.service.operation.OperationService com.sequenceiq.freeipa.flow.instance.reboot.action.RebootActions$4.operationService)
↑     ↓
|  operationService (field private com.sequenceiq.flow.service.FlowService com.sequenceiq.freeipa.service.operation.OperationService.flowService)
↑     ↓
|  flowService (field private com.sequenceiq.flow.converter.FlowProgressResponseConverter com.sequenceiq.flow.service.FlowService.flowProgressResponseConverter)
↑     ↓
|  flowProgressResponseConverter defined in URL [jar:file:/freeipa.jar!/BOOT-INF/lib/flow.jar!/com/sequenceiq/flow/converter/FlowProgressResponseConverter.class]
└─────┘
```

freeipa/src/main/java/com/sequenceiq/freeipa/service/operation/FlowOperationService.java
```diff
@@ -0,0 +1,43 @@
...
+@Service
+public class FlowOperationService {
+
+    @Inject
+    private FlowService flowService;
+
+    @Inject
+    private OperationDetailsPopulator operationDetailsPopulator;
+
+    public OperationView getOperationProgressByEnvironmentCrn(String resourceCrn, boolean detailed) {
```
lajosrodek, line 27: If this is really always for looking up by environment CRN (and not freeipa CRN, for example), perhaps rename `resourceCrn` to `environmentCrn`.

freeipa/src/test/java/com/sequenceiq/freeipa/service/operation/FlowOperationServiceTest.java
```diff
@@ -50,7 +50,7 @@
 
     @BeforeEach
     public void setUp() {
-        underTest = new OperationService();
+        underTest = new FlowOperationService();
         MockitoAnnotations.openMocks(this);
     }
```
lajosrodek, line 55: Why the need for an explicit instantiation & `openMocks()`? `@InjectMocks` takes care of both, so I think the whole `setUp()` can be dropped if no other activities are to be done.

lajosrodek (approved):
LGTM, only two minor remarks.

freeipa/src/main/java/com/sequenceiq/freeipa/controller/OperationV1Controller.java
```diff
@@ -40,6 +44,6 @@ public OperationStatus getOperationStatus(@NotNull String operationId, @AccountI
     @Override
     @CheckPermissionByResourceCrn(action = DESCRIBE_ENVIRONMENT)
     public OperationView getOperationProgressByEnvironmentCrn(@ResourceCrn String resourceCrn, boolean detailed) {
```
lajosrodek, line 46: Hmm, I did not notice this last time, `resourceCrn` could be renamed to `environmentCrn`. 😄 Sorry

oleewere: will do that in an addendum change as ITs passed

PR #11061, opened 2021-07-14
CB-12068 Introduce additional configuration parameters for Load-Based…

smaniraju:
… Autoscaling.

   1.maxScaleUpStepSize : The max step size for scaleup.
   2.maxScaleDownStepSize : The max step size for scaledown.
   3.scaleUpCoolDownMinutes : CoolDown Mins for scaleup after last autoscaling action.
   4.scaleDownCoolDownMinutes : CoolDown Mins for scaledown after last autoscaling action.
All the above parameters are optional.

### DIFF
--- a/autoscale-api/src/main/java/com/sequenceiq/periscope/doc/ApiDescription.java
+++ b/autoscale-api/src/main/java/com/sequenceiq/periscope/doc/ApiDescription.java
@@ -153,7 +153,12 @@ public class ApiDescription {
     public static class LoadAlertJsonProperties {
         public static final String LOAD_ALERT_CONFIGURATION_MIN_RESOUCE_VALUE = "The lower bound for the resource";
         public static final String LOAD_ALERT_CONFIGURATION_MAX_RESOUCE_VALUE = "The upper bound for the resource";
-        public static final String LOAD_ALERT_CONFIGURATION_COOL_DOWN_MINS_VALUE = "CoolDown between successive cluster autoscaling actions.";
+        public static final String LOAD_ALERT_CONFIGURATION_COOL_DOWN_MINS_VALUE = "CoolDown Mins between successive cluster autoscaling actions." +
+                " This is overridden by the scaleUpCoolDownMinutes/scaleDownCoolDownMinutes if they are specified";
+        public static final String LOAD_ALERT_CONFIGURATION_SCALE_UP_COOL_DOWN_MINS_VALUE = "CoolDown Mins for scaleup after last autoscaling action.";
+        public static final String LOAD_ALERT_CONFIGURATION_SCALE_DOWN_COOL_DOWN_MINS_VALUE = "CoolDown Mins for scaledown after last autoscaling action.";
+        public static final String LOAD_ALERT_CONFIGURATION_MAX_SCALEUP_STEP_SIZE_VALUE = "The max step size for scaleup.";
+        public static final String LOAD_ALERT_CONFIGURATION_MAX_SCALEDOWN_STEP_SIZE_VALUE = "The max step size for scaledown.";
         public static final String LOAD_ALERT_CONFIGURATION = "Configuration of Load Alert";
     }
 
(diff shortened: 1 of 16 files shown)


autoscale-api/src/main/java/com/sequenceiq/periscope/doc/ApiDescription.java
```diff
@@ -153,7 +153,11 @@ private ApiDescription() {
     public static class LoadAlertJsonProperties {
         public static final String LOAD_ALERT_CONFIGURATION_MIN_RESOUCE_VALUE = "The lower bound for the resource";
         public static final String LOAD_ALERT_CONFIGURATION_MAX_RESOUCE_VALUE = "The upper bound for the resource";
-        public static final String LOAD_ALERT_CONFIGURATION_COOL_DOWN_MINS_VALUE = "CoolDown between successive cluster autoscaling actions.";
+        public static final String LOAD_ALERT_CONFIGURATION_COOL_DOWN_MINS_VALUE = "CoolDown Mins between successive cluster autoscaling actions.";
```
sidseth, line 156: "This is overridden by the scale-up/scale-down cooldown parameters, if they are specified"

autoscale/src/main/java/com/sequenceiq/periscope/domain/LoadAlertConfiguration.java
```diff
@@ -35,8 +45,22 @@ public Integer getCoolDownMinutes() {
...
-        return TimeUnit.MILLISECONDS.convert(coolDownMinutes, TimeUnit.MINUTES);
+    public Long getPollingCoolDownMillis() {
+        Integer pollingCoolDown = coolDownMinutes != null ? coolDownMinutes :
+                Math.min(scaleUpCoolDownMinutes, scaleDownCoolDownMinutes);
+        return TimeUnit.MILLISECONDS.convert(pollingCoolDown, TimeUnit.MINUTES);
+    }
+
+    @JsonIgnore
+    public Long getScaleUpCoolDownMillis() {
+        Integer scaleUpCoolDown = coolDownMinutes != null ? coolDownMinutes : scaleUpCoolDownMinutes;
```
sidseth, line 56: I think this should be the opposite. If scaleUpCoolDownMinutes is set - use it, otherwise fallback to the default.
(Similarly for the scale-down)

i.e. A user can set either the scale-up or the scale-down config to override the default value for that specific scenario.

Would apply to getPollingCooldownMillis as well.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
```diff
@@ -129,15 +129,17 @@ protected void pollYarnMetricsAndScaleCluster() throws Exception {
...
-                        hostFqdnsToInstanceId, maxAllowedDownScale, mandatoryDownScaleCount);
+                        hostFqdnsToInstanceId, maxAllowedDownScale, mandatoryDownScaleCount, loadAlertConfiguration.getMaxScaleDownStepSize());
 
-        if (yarnRecommendedScaleUpCount > 0) {
+        if (yarnRecommendedScaleUpCount > 0 && isCoolDownTimeElapsed(cluster.getStackCrn(),
+                loadAlertConfiguration.getScaleUpCoolDownMillis(), cluster.getLastScalingActivity()))  {
             sendScaleUpEvent(existingHostGroupSize, yarnRecommendedScaleUpCount);
-        } else if (!yarnRecommendedDecommissionHosts.isEmpty()) {
+        } else if (!yarnRecommendedDecommissionHosts.isEmpty() && isCoolDownTimeElapsed(cluster.getStackCrn(),
+                loadAlertConfiguration.getScaleDownCoolDownMillis(), cluster.getLastScalingActivity()))  {
```
sidseth, line 142: Any chance the log message in
EvaluatorExecutor
"LOGGER.debug("Cluster {} cannot be scaled for {} min(s)", clusterCrn," 

could be made a little more specific to differentiate between scale-up/scale-down times for this particular scenario?

autoscale-api/src/main/java/com/sequenceiq/periscope/doc/ApiDescription.java
smaniraju: yes

autoscale/src/main/java/com/sequenceiq/periscope/domain/LoadAlertConfiguration.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/load/YarnLoadEvaluator.java
smaniraju: ok

sidseth (approved):

PR #11066, opened 2021-07-15
CB-13393 Image lookup by image uuid should not filter images by OS.

pkedvessy:


TheTinkerDad (approved):

doktoric (approved):

doktoric:
increased cluster proxy version and retried

PR #11069, opened 2021-07-15
CB-12805 volume deletion not to call instnace attribute update on AWS when no instance has been assigned to volume set

biharitomi:
Covering edge-case when volumes have been created but infra creation failed before volume attachment which is causing volume sets are not assigned to any instance and volume deletion fails on updating attributes of the related instance.


bergerdenes (approved):

lajosrodek (approved):
LGTM, perhaps add an UT for this scenario as well.

PR #11058, opened 2021-07-14
CB-13391. Addendum - rename resourceCrn to environmentCrn in freeipa …

oleewere:
…operation controller

See detailed description in the commit message.

PR #11070, opened 2021-07-15
CB-13391. Fix freeipa circular dependency (because of flow operation service)

oleewere:

cherry-picked from master


nadamdb (approved):

lajosrodek:
See #11057 and #11058.

PR #11071, opened 2021-07-15
CB-13413. Add status reason for list diagnostics output.

oleewere:


aakanser (approved):
looks good to me

PR #11072, opened 2021-07-15
Optimize cluster telemetry cleanup on termination.

oleewere:


biharitomi (approved):

bergerdenes:
should it go to 2.45 as well?

oleewere:
> should it go to 2.45 as well?

i do not think so it's that critical, i think it's fine on master

doktoric:
@oleewere how about with 2.46? do we fine with master?

oleewere:
> @oleewere how about with 2.46? do we fine with master?

same is true there, point of this change to remove unnecessary transaction (also api call happening inside that) - i do not think any 2.46 tests failed on this - so master should be fine

PR #11075, opened 2021-07-15
[AWS] Autoscale stack update failed after upgrading Data Hub

schfeca75:
Instead of update the cloudformation template, we create a new version of LaunchTemplate for all the autoscale groups with the new imageid


cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsLaunchTemplateImageUpdateService.java
```diff
@@ -1,58 +1,66 @@
...
-                LOGGER.debug(String.format("CloudFormation is not updated as it is already in the latest state, name: %s", cfResource.getName()), e);
-            } else {
-                throw e;
-            }
+        AmazonAutoScalingClient autoScalingClient = awsClient.createAutoScalingClient(credentialView, regionName);
+        AmazonEc2Client ec2Client = awsClient.createEc2Client(credentialView, regionName);
+        Map<AutoScalingGroup, String> autoScalingGroups = autoScalingGroupHandler.getAutoScalingGroups(cloudFormationClient, autoScalingClient, cfResource);
+        for (Map.Entry<AutoScalingGroup, String> asgEntry : autoScalingGroups.entrySet()) {
+            AutoScalingGroup autoScalingGroup = asgEntry.getKey();
+            LaunchTemplateSpecification launchTemplateSpecification = autoScalingGroup.getLaunchTemplate() == null ?
```
keyki, line 45: Add a log message before each AWS API call.

keyki:
Can you please collect the permissions required for these calls?

keyki:
Please add a new validation for the required permissions. If the customer doesn't have it then the upgrade should not be started.

cloud-aws-cloudformation/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/AwsLaunchTemplateImageUpdateService.java
```diff
@@ -1,58 +1,66 @@
...
+            ModifyLaunchTemplateRequest modifyLaunchTemplateRequest = new ModifyLaunchTemplateRequest()
+                    .withLaunchTemplateId(launchTemplateSpecification.getLaunchTemplateId())
+                    .withDefaultVersion(createLaunchTemplateVersionResult.getLaunchTemplateVersion().getVersionNumber().toString());
+            ec2Client.modifyLaunchTemplate(modifyLaunchTemplateRequest);
+            LaunchTemplateSpecification newLaunchTemplateSpecification = new LaunchTemplateSpecification()
+                    .withLaunchTemplateId(launchTemplateSpecification.getLaunchTemplateId())
+                    .withVersion(createLaunchTemplateVersionResult.getLaunchTemplateVersion().getVersionNumber().toString());
+            autoScalingClient.updateAutoScalingGroup(new UpdateAutoScalingGroupRequest()
+                    .withAutoScalingGroupName(autoScalingGroup.getAutoScalingGroupName())
+                    .withLaunchTemplate(newLaunchTemplateSpecification));
```
lacikaaa, line 61: could you split these into methods and add logging and result validations, like:
- `createNewLaunchTemplateVersion`
- `configureASGForNewVersion`

I would expect the `for` body has 3-4 statements at max, or even that one could be moved to a single method so the `updateImage` won't have too many lines.

lacikaaa (changes requested):

lacikaaa (approved):

PR #9368, opened 2020-11-03
CB-7898: Add SRM to Streaming DataHub cluster templates

Jugis:
See detailed description in the commit message.

PR #11082, opened 2021-07-16
CB-10916 CruiseControl in Streaming blueprint for 2.7.12

benyoka:
Cruise Control is added to the Streams Messaging blueprints as of 7.2.12

Testing: manual


doktoric:
I think we also need to rename the current blueprint in the database

benyoka:
> I think we also need to rename the current blueprint in the database

We explored this question when SRM was added in 7.2.6. The conclusion was that DB scripts are not needed. Please see https://github.com/hortonworks/cloudbreak/pull/9368

doktoric:
> > I think we also need to rename the current blueprint in the database
> 
> We explored this question when SRM was added in 7.2.6. The conclusion was that DB scripts are not needed. Please the [comment on that PR](https://github.com/hortonworks/cloudbreak/pull/9368#issuecomment-726960128).

I almost aleays forget this :)

doktoric (approved):

PR #11088, opened 2021-07-17
CB-13226 Making root disk to be of type SSD for FreeIPA in GCP

cegganesh84:
1. The previous PR #11076 changed the root disk of datalake and datahub.
2. There is a separate code path for FreeIPA which still sets the root disk type to HDD.
3. Found during end-to-end testing.
4. Will perform the tests in parallel.

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/template/InstanceTemplateRequestToTemplateConverter.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/template/InstanceTemplateRequestToTemplateConverter.java
@@ -26,6 +26,7 @@ import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.VolumeR
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.instance.aws.AwsInstanceTemplateParameters;
 import com.sequenceiq.freeipa.entity.Template;
 import com.sequenceiq.freeipa.service.DefaultRootVolumeSizeProvider;
+import com.sequenceiq.freeipa.service.DefaultRootVolumeTypeProvider;
 import com.sequenceiq.freeipa.service.stack.instance.DefaultInstanceTypeProvider;
 
 @Component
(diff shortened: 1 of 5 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/converter/instance/template/InstanceTemplateRequestToTemplateConverter.java
```diff
@@ -73,10 +73,15 @@ public Template convert(InstanceTemplateRequest source, CloudPlatform cloudPlatf
     }
 
     private void setVolumesProperty(Set<VolumeRequest> attachedVolumes, Template template, CloudPlatform cloudPlatform) {
+        String type = "HDD";
+        if (CloudPlatform.GCP == cloudPlatform) {
```
doktoric, line 77: maybe this should go the interface layer as a method like getFreeIPADefaultDiskType() and then it will be implemented in each cloudprovider modul ?

cegganesh84: @doktoric How about implementing like https://github.com/hortonworks/cloudbreak/blob/master/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/DefaultRootVolumeSizeProvider.java? That way I can contain the change within FreeIPA.

I could not find how gp2 and Standard SSD LRS are chosen for AWS and Azure. They are certainly not honoring the HDD value here.

cegganesh84: Thanks, @doktoric, followed that approach.

freeipa/src/main/java/com/sequenceiq/freeipa/service/DefaultRootVolumeTypeProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultRootVolumeTypeProvider.class);
+
+    private static final String DEFAULT_ROOT_VOLUME_TYPE = "HDD";
+
+    private static final String ROOT_VOLUME_TYPE_PROPERTY_PREFIX = "cb.platform.default.rootVolumeType.";
+
+    private final Map<String, String> platformVolumeTypeMap;
+
+    public DefaultRootVolumeTypeProvider(CloudPlatformConnectors cloudPlatformConnectors, Environment environment) {
```
doktoric, line 31: can you please add some unit test for this class?

cegganesh84: Thanks, added.

doktoric (approved):

PR #11090, opened 2021-07-18
CB-13283 - Add CloudWatch alarm to the instances

gregito:


cloud-aws-native/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/instance/alarm/AwsNativeCloudWatchResourceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    protected boolean isFinished(AwsContext context, AuthenticatedContext auth, CloudResource resource) {
+        return true;
+    }
+
+    @Override
+    public ResourceType resourceType() {
+        return ResourceType.AWS_INSTANCE;
```
doktoric, line 84: I think the resourcetype should be cloudwath alarm

doktoric (approved):

PR #11094, opened 2021-07-19
CB-13431: Add CDP entitlement for support of CMEK for GCP encryption

juhi-09:
CB-13431: Add CDP entitlement for support of CMEK for GCP encryption
If this entitlement is set to true, Customer on GCP will have an option to use CustomerManagedEncryptionKeys to Encrypt Managed Disks.
Changes which supports encryption with CMEK are taken as part of CB-10502.


cegganesh84 (approved):
LGTM

PR #11099, opened 2021-07-19
CB-13338 Add FreeIPA upgrade right

lacikaaa:


authorization-common/src/main/resources/legacyRights.json
```diff
@@ -11,6 +11,7 @@
   "environments/accessEnvironment": "environments/accessEnvironment",
   "environments/repairFreeIPA": "environments/write",
   "environments/adminFreeIPA": "environments/adminFreeIPA",
+  "environments/upgradeFreeIPA": "environments/upgradeFreeIPA",
```
horadla23, line 14: value should be `environments/write` I think

lacikaaa: right!

horadla23 (approved):

jamisonbennett (approved):

PR #11100, opened 2021-07-19
CB-13445 Unable to delete cluster template which has been created by …

horadla23:
…internal endpoint


core/src/main/java/com/sequenceiq/cloudbreak/controller/v4/BlueprintV4Controller.java
```diff
@@ -116,7 +116,7 @@ public BlueprintV4Response deleteByName(Long workspaceId, @NotNull @ResourceName
 
     @Override
     @CheckPermissionByResourceCrn(action = AuthorizationResourceAction.DELETE_CLUSTER_TEMPLATE)
-    public BlueprintV4Response deleteByCrn(Long workspaceId, @NotNull @ResourceCrn String crn) {
+    public BlueprintV4Response deleteByCrn(Long workspaceId, @NotNull @ResourceCrn @TenantAwareParam String crn) {
         Blueprint deleted = blueprintService.deleteByWorkspace(NameOrCrn.ofCrn(crn), workspaceId);
```
foldik, line 120: I think here if you use this with internal crn you need to get the workspaceId from `restRequestThreadLocalService.getRequestedWorkspaceId()`. As far as I know the framework doesn't change method param properly. Simple unit tests didn't catch it for me in the past. If this is not the case anymore than ignore this comment.

foldik (approved):
LGTM

PR #11102, opened 2021-07-19
CB-13443: Enabled Load balancing using Knox for Profiler Manager Service.

subrat-mishra:
Profiler Manager Services has 3 main services i.e. profiler_manager_service, profiler_metrics_service and data_discovery_service. They all supports active-active HA now. 
Added enableLoadBalancing explicitly for all the services which would help in scaling. 

Tested these configs on my local cluster.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/gateway/config/cm/topology.xml.j2
@@ -338,19 +338,19 @@
              {% if 'DATA-DISCOVERY-SERVICE-API' in exposed and 'DATA_DISCOVERY_SERVICE_AGENT' in salt['pillar.get']('gateway:location') -%}
              <param>
                  <name>DATA-DISCOVERY-SERVICE-API</name>
-                 <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
+                 <value>enableStickySession=false;noFallback=false;enableLoadBalancing=true</value>
              </param>
              {%- endif %}
              {% if 'PROFILER-ADMIN-API' in exposed and 'PROFILER_ADMIN_AGENT' in salt['pillar.get']('gateway:location') -%}
              <param>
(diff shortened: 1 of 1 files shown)


moresandeep:
@subrat-mishra the sticky session flag is `false` i am assuming this is because the services are stateless. Other than confirming that the services are stateless everything looks good.

moresandeep (approved):

subrat-mishra:
> @subrat-mishra the sticky session flag is `false` i am assuming this is because the services are stateless. Other than confirming that the services are stateless everything looks good.

Yes @moresandeep all the services are stateless.

kumariitr (approved):

doktoric:
@subrat-mishra  is this change backward compatible? We are using this knox configuration for older runtimes as well like 7.2.8? Will it work ?

subrat-mishra:
> @subrat-mishra is this change backward compatible? We are using this knox configuration for older runtimes as well like 7.2.8? Will it work ?

@doktoric I haven't tested on the older runtimes. CB-9745 has similar changes for other services like RANGER, SOLR etc.
@moresandeep can you please let us know the knox configs are backward compatible?

moresandeep:
@doktoric these changes are backwards compatible with 7.2.8.

subrat-mishra:
Thanks @moresandeep for confirming.
@doktoric could you please merge this PR?

doktoric:
@moresandeep we still support older runtimes then 7.2.8

moresandeep:
I see what you mean, i don't think they will cause any issue for older runtimes but it would be best if we can test it out. 
@subrat-mishra this should be easy to test on an older cluster <7.1.7

subrat-mishra:
@moresandeep, I have tested runtime version 7.1.0 on public cloud. I have manually added the knox configs to the
/var/lib/knox/gateway/conf/topologies/cdp-proxy.xml and verified its working fine. Although the enableLoadBalancing feature was not working which is expected.

doktoric (approved):

PR #11103, opened 2021-07-19
CB-12898 Filter the resources by private id as well during the collec…

topolyai5:
…t metadata status


biharitomi (approved):
LGTM, thanks

PR #11112, opened 2021-07-20
CB-13065 remove datalakeresource and servicedescriptor tables

afarsang:

### DIFF
--- a/core/src/main/resources/schema/app/20210719141901_CB-13065_remove_datalakeresource_and_servicedescriptor.sql
+++ b/core/src/main/resources/schema/app/20210719141901_CB-13065_remove_datalakeresource_and_servicedescriptor.sql
@@ -1,0 +1,69 @@
+-- // CB-13065 removing datalakeresources and servicedescriptor
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE stack DROP COLUMN datalakeresourceid;
+
+DROP TABLE datalakeresources_rdsconfig;
+
+DROP TABLE servicedescriptor;
+
+DROP TABLE datalakeresources;
+
+-- //@UNDO
+-- SQL to undo the change goes here.
+
+CREATE TABLE IF NOT EXISTS datalakeresources (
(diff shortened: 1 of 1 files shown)


core/src/main/resources/schema/app/20210719141901_CB-13065_remove_datalakeresource_and_servicedescriptor.sql
```diff
@@ -0,0 +1,69 @@
...
+    blueprintparams TEXT NOT NULL,
+    blueprintsecretparams TEXT,
+    componentshosts TEXT NOT NULL
+);
+ALTER TABLE ONLY servicedescriptor ADD CONSTRAINT servicedescriptor_pkey PRIMARY KEY (id);
+ALTER TABLE ONLY servicedescriptor ADD CONSTRAINT fk_servicedescriptor_datalakeresources_id FOREIGN KEY (datalakeresources_id) REFERENCES datalakeresources(id);
+CREATE INDEX IF NOT EXISTS idx_servicedescriptor_datalakeresources_id ON servicedescriptor (datalakeresources_id);
+
+DROP SEQUENCE IF EXISTS datalakeresources_datalakestack_id_seq CASCADE;
+ALTER TABLE datalakeresources ALTER COLUMN datalakestack_id DROP NOT NULL;
```
bergerdenes, line 50: are you sure this is part of the undo script?

afarsang: these were part of the original scripts which created and manipulated the datalakeresource table. from 2018 till recent time

bergerdenes: I would rather script out the CREATE TABLE scripts and paste those. At least please try out if the undo script produces the expected result, e.g. the tables are restored to the same state.

PR #11118, opened 2021-07-21
CB-13419 Revert back to Basic load-balancer in Azure for datalake and…

cegganesh84:
… DE HA for 7.2.10.

1. Basic load-balancer allows the VMs behind it talk to internet, Standard doesn't.
2. This does not entirely fix CB-13419, instead a stop gap till 7.2.11 DE HA is released.
3. To fix DE HA, we need to find a way for the nodes to communicate with internet using one of the options given in https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections

### DIFF
--- a/cloud-azure/src/main/resources/templates/arm-v2.ftl
+++ b/cloud-azure/src/main/resources/templates/arm-v2.ftl
@@ -530,7 +530,11 @@
                     ]
                   },
                   "sku": {
-                    "name": "Standard"
+                      <#if (loadBalancer.instanceGroupNames?size > 1)>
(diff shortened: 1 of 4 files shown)


doktoric (approved):

lacikaaa:
@cegganesh84 if something is neede in a release branch, like 2.46, then please open a PR there first, and merge it forward instead of opening a separate one for all branches. As I see this is already merged to 2.46, in that case 2.46 should be merged to master. Please refer to CB branching documentation

PR #11120, opened 2021-07-21
CB-13379 Ease conflict resolution problems in FreeIPA HA setup

cegganesh84:
1. There is a chance that ipa-client-install and ipa dnsrecord-add could go to different FreeIPA hosts.
2. The replication may not have been complete when the ipa dnsrecord-add was done.
3. Also the repeated addition of records could end up in similar case above and could even lead to deletion of records due to conflict.
4. The conflict can not be automatically resolved by FreeIPA because multi valued DNs are not possible to be converted to hostnames.
5. Also it is two record additions with same or different values.
6. To overcome all these obstacles in this change trying to use dnsrecord-mod instead of dnsrecord-add.
7. This way I hope the nsuniqueid is not added to the idnsname which is leading to multi valued DNs and which becomes an unresolvable conflict.

Tested at scale.

(cherry picked from commit b1c630ae8e339a432ace82937075dc17d30c2a8d)

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/sssd/ipa.sls
@@ -28,7 +28,7 @@ join_ipa:
 {% else %}
     - name: runuser -l root -c 'export PW="{{salt['pillar.get']('sssd-ipa:password')}}" && /opt/salt/scripts/join_ipa.sh && echo $(date +%Y-%m-%d:%H:%M:%S) >> /var/log/ipa-join-executed'
 {% endif %}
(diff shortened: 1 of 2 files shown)


lacikaaa:
are we sure we don't want to wait for some the full lifecycle to see if this is causing any issue?

sidseth:
> are we sure we don't want to wait for the full lifecycle to see if this is causing any issue?
That would be how long? I think this has been tried out on a few largish clusters, with a reasonable amount of scaling - so should be good to go.

cegganesh84:
Closing this PR. We did saw one instances where the A record disappeared even after dns-mod change.

PR #11123, opened 2021-07-21
CB-13483 Differentiate init.ftl for ccmv2 agent installation

bergerdenes:
CCMv2 jumpgate agent needs to be installed on FreeIPA host only, and will not be installed on datalake/datahub gateway node.

### DIFF
--- a/core/src/test/resources/azure-gateway-ccm-v2-jumpgate-init.sh
+++ b/core/src/test/resources/azure-gateway-ccm-v2-jumpgate-init.sh
@@ -26,6 +26,7 @@ export CCM_V2_AGENT_KEY_ID="agentKeyId"
 export CCM_V2_AGENT_CRN="agentCrn"
 export CCM_V2_AGENT_BACKEND_ID_PREFIX="agentCrn-"
 export IS_CCM_V2_JUMPGATE_ENABLED=true
+export START_CCM_V2_JUMPGATE_AGENT=false
 
 date >> /tmp/time.txt
 
--- a/core/src/main/resources/init/init.ftl
+++ b/core/src/main/resources/init/init.ftl
@@ -67,6 +67,7 @@ export IS_CCM_V2_JUMPGATE_ENABLED=true
 <#else>
 export IS_CCM_V2_JUMPGATE_ENABLED=false
 </#if>
+export START_CCM_V2_JUMPGATE_AGENT=false
 
 ${customUserData}
 
--- a/core/src/test/resources/azure-core-ccm-init.sh
+++ b/core/src/test/resources/azure-core-ccm-init.sh
@@ -19,6 +19,7 @@ export IS_PROXY_ENABLED=false
 export IS_CCM_ENABLED=false
 export IS_CCM_V2_ENABLED=false
 export IS_CCM_V2_JUMPGATE_ENABLED=false
+export START_CCM_V2_JUMPGATE_AGENT=false
 
 date >> /tmp/time.txt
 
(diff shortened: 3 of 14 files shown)


doktoric (approved):

lacikaaa:
while this looks ok, I think we are adding code to the core `init.ftl` because the image contains the same logic here for both freeipa and the clusters. Ideally cluster related images shouldn't have the jumpgate related stuff, and as the result we could omit the related code from the core too.

bergerdenes:
Well, this is true. I remove the core stuff.

lacikaaa (approved):

lacikaaa:
> Well, this is true. I remove the core stuff.

just make sure the image side handles well if this is not set

doktoric:
if we remove this from the core make sure that the DH/DL part also not using the legacy ccmv2 way. so no agent started there at all. I think we added it to the core because of the current userdata behaviour.

core/src/test/resources/azure-gateway-ccm-v2-jumpgate-init.sh
```diff
@@ -26,6 +26,7 @@ export CCM_V2_AGENT_KEY_ID="agentKeyId"
 export CCM_V2_AGENT_CRN="agentCrn"
 export CCM_V2_AGENT_BACKEND_ID_PREFIX="agentCrn-"
 export IS_CCM_V2_JUMPGATE_ENABLED=true
+export START_CCM_V2_JUMPGATE_AGENT=false
```
lajosrodek, line 29: Is not this in conflict with `IS_CCM_V2_JUMPGATE_ENABLED=true` in the line above?

bergerdenes: not

lacikaaa:
> if we remove this from the core make sure that the DH/DL part also not using the legacy ccmv2 way. so no agent started there at all. I think we added it to the core because of the current userdata behaviour.

Ok, I checked the user-data-helper script and I think it's getting a bit messy with all these flags.
We have:
```
IS_CCM_ENABLED
IS_CCM_V2_ENABLED
IS_CCM_V2_JUMPGATE_ENABLED
```
and now we would have
```
START_CCM_V2_JUMPGATE_AGENT
```
as a new variable.
I think it's time to rethink a bit this, because it's getting to grow into a flag hell. Can't we just split up the `user-data-helper`, and in case of clusters, there shouldn't be anything related starting a jumpgate agent. It shouldn't be possible, because it's a FreeIPA (image) only feature. So when we see `IS_CCM_V2_JUMPGATE_ENABLED = true` on a FreeIPA image, then it would start the agent as a jumpgate agent, while the same on a cluster would configure nginx to listen without mTLS.

bergerdenes:
How do you know on an image in the user-data-helper script that you are running or a FreeIPA or on a DL/DH cluster? This is the flag that determines it. True, I could call it `cluster_type` or `image-type` or whatever.

lacikaaa:
> How do you know on an image in the user-data-helper script that you are running or a FreeIPA or on a DL/DH cluster? This is the flag that determines it. True, I could call it `cluster_type` or `image-type` or whatever.

you place a different script in case of FreeIPA and DL/DH on the image during the burning process

core/src/test/resources/azure-gateway-ccm-v2-jumpgate-init.sh
lajosrodek: Alright 👍

PR #11133, opened 2021-07-22
[CB-13426]  create resourcebuilders for new load balancer specific resources

yg-gd:
Add in the other resources required for GCP load balancing, which are based on the number of load balancers on the stack, or the mapping within the load balancer.


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -44,15 +54,24 @@
...
+    public static List<CloudResource> buildReservedIp(GcpContext context, List<CloudResource> buildableResource,
+            CloudStack cloudStack, String type) throws Exception {
+        List<CloudResource> result = new ArrayList<>();
+        for (CloudResource resource : buildableResource) {
             String projectId = context.getProjectId();
             String region = context.getLocation().getRegion().value();
 
             Address address = new Address();
             address.setName(resource.getName());
-            address.setDescription(description());
```
hreeve-cloudera, line 55: Is there a reason this was removed?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -61,18 +80,23 @@
             try {
                 Operation operation = networkInsert.execute();
                 if (operation.getHttpErrorStatusCode() != null) {
-                    throw new GcpResourceException(operation.getHttpErrorMessage(), resourceType(), resource.getName());
+                    throw new GcpResourceException(operation.getHttpErrorMessage(), ResourceType.GCP_RESERVED_IP, resource.getName());
```
hreeve-cloudera, line 83: I'd change this back to `resourceType()`. We have the method available, and this changes it from hardcoding the resource type in one place to hardcoding it in four places.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -61,18 +80,23 @@
...
                 }
-                result = Collections.singletonList(createOperationAwareCloudResource(resource, operation));
+                result.add(createOperationAwareCloudResource(resource, operation));
             } catch (GoogleJsonResponseException e) {
-                throw new GcpResourceException(checkException(e), resourceType(), resource.getName());
+                throw new GcpResourceException(e.getDetails().getMessage(), ResourceType.GCP_RESERVED_IP, resource.getName());
```
hreeve-cloudera, line 87: Why did you change this line from using methods to pulling the logic out and hardcoding it? I know the methods are very simple, but why have this code more than one place when you can just call the methods available via the abstract classes?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -81,8 +105,7 @@ public CloudResource delete(GcpContext context, AuthenticatedContext auth, Cloud
             Operation operation = compute.addresses().delete(projectId, region, resource.getName()).execute();
             return createOperationAwareCloudResource(resource, operation);
         } catch (GoogleJsonResponseException e) {
-            exceptionHandler(e, resource.getName(), resourceType());
-            return null;
+            throw exceptionHandlerWithThrow(e, resource.getName(), ResourceType.GCP_RESERVED_IP);
```
hreeve-cloudera, line 108: Here, again, we're causing a thrown exception when the resource isn't found, whereas before we weren't throwing an exception in that case. Why the change?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/group/GcpInstanceGroupResourceBuilder.java
```diff
@@ -45,8 +45,6 @@ public CloudResource build(GcpContext context,
             AuthenticatedContext auth, Group group,
             Network network, Security security, CloudResource resource) throws Exception {
         LOGGER.info("Building GCP instancegroup {} for project {}", group.getName(), context.getProjectId());
-
```
hreeve-cloudera, line 48: Nit: I'm not oposed to adding whitespace to make something more readable, but you only need one line. :)

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpBackendServiceResourceBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+import com.sequenceiq.cloudbreak.cloud.gcp.context.GcpContext;
+import com.sequenceiq.cloudbreak.cloud.model.CloudLoadBalancer;
+import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
+import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+import com.sequenceiq.cloudbreak.cloud.model.Group;
+import com.sequenceiq.cloudbreak.cloud.model.TargetGroupPortPair;
+import com.sequenceiq.common.api.type.LoadBalancerType;
+import com.sequenceiq.common.api.type.ResourceType;
+
+public class GcpBackendServiceResourceBuilder extends AbstractGcpLoadBalancerBuilder {
```
hreeve-cloudera, line 30: Could we get some java docs on these new classes explaining what they do?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpForwardingRuleResourceBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+
+            if (!backendName.isPresent()) {
+                LOGGER.warn("backend not found for fowarding rule {}, port {}, project {}", buildableResource.getName(), trafficPort, projectId);
+                continue;
+            } else {
+                forwardingRule.setBackendService(String.format(GCP_BACKEND_SERVICE_REF_FORMAT, projectId, regionName, backendName.get()));
+            }
+
+            if (ipResources.isEmpty()) {
+                LOGGER.warn("No reserved IP address for loadbalancer {}-{}, using empmeral", loadBalancer.getType(), projectId);
```
hreeve-cloudera, line 90: Nit: Is "empmeral" a typo?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpForwardingRuleResourceBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+        String regionName = context.getLocation().getRegion().getRegionName();
+        List<CloudResource> results = new ArrayList<>();
+
+
+        List<CloudResource> backendResources = filterResourcesByType(context.getLoadBalancerResources(loadBalancer.getType()), ResourceType.GCP_BACKEND_SERVICE);
+        List<CloudResource> ipResources = filterResourcesByType(context.getLoadBalancerResources(loadBalancer.getType()), ResourceType.GCP_RESERVED_IP);
+
+        for (CloudResource buildableResource : buildableResources) {
+            Integer trafficPort = buildableResource.getParameter(TRAFFICPORT, Integer.class);
+            ForwardingRule forwardingRule = new ForwardingRule().setIPProtocol("TCP")
```
hreeve-cloudera, line 66: I generally think string literals should be constants instead of hardcoded wherever possible. That makes them easy to find in the scenario where they might have to be updated, and easier to share if we eventually find we're hardcoding the same values multiple places.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/service/checker/AbstractGcpBaseResourceChecker.java
```diff
@@ -46,7 +46,7 @@ protected void exceptionHandler(GoogleJsonResponseException ex, String name, Res
         }
     }
 
-    protected GcpResourceException exceptionHandlerWithThrow(GoogleJsonResponseException ex, String name, ResourceType resourceType) {
+    protected static GcpResourceException exceptionHandlerWithThrow(GoogleJsonResponseException ex, String name, ResourceType resourceType) {
```
hreeve-cloudera, line 49: Why were these methods made static? I don't see them referenced in a static context anywhere in this review.

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/LoadBalancerResourceBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+import com.sequenceiq.cloudbreak.cloud.CloudPlatformAware;
+import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
+import com.sequenceiq.cloudbreak.cloud.model.CloudLoadBalancer;
+import com.sequenceiq.cloudbreak.cloud.model.CloudResource;
+import com.sequenceiq.cloudbreak.cloud.model.CloudStack;
+import com.sequenceiq.cloudbreak.cloud.template.context.ResourceBuilderContext;
+import com.sequenceiq.common.api.type.ResourceType;
+
+/**
+ * Cloud providers which do not support template based deployments (like AWS Cloudformation, Azure ARM or OpenStack Heat)
```
hreeve-cloudera, line 15: These are some nice docs.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/service/checker/AbstractGcpBaseResourceChecker.java
yg-gd: I had made these static to be referenced staticly from the new Reserved IP resource builder, but then I realized it would be just as easy to inject the old one. Thanks for bringing this up.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
yg-gd: I had made this change with the other static methods, no longer needed with injection.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
yg-gd: originally changed to make it a static method. not needed with injection

yg-gd:
@pdarvasi @lacikaaa These are the other changes for GCP load balancer resource builders. @biharitomi this is what I had in mind for a load balancer resource builder, at some point we might want to align the way we launch load balancers in AWS/Azure

sxxgrc (approved):
Looks good to me based on what I know.

yg-gd:
Now includes changes from local testing, will squash after review.  part of CB-12822 also here

doktoric:
do we need entitlement for this feature?

hreeve-cloudera:
> do we need entitlement for this feature?

The entitlement that will be used is the `CDP_DATA_LAKE_LOAD_BALANCER` entitlement. The way it's intended to work for LB is that the entitlement is checked when first adding load balancer objects to the stack. If it's not enabled, they won't be added to the stack, and the logic the depends on the LB being in the stack will not trigger.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -27,6 +28,11 @@
 
 @Service
 public class GcpReservedIpResourceBuilder extends AbstractGcpComputeBuilder {
+
+    public static final String INTERNAL = "INTERNAL";
```
biharitomi, line 32: Can we have a enum to store the types of reserved IPs?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/compute/GcpReservedIpResourceBuilder.java
```diff
@@ -44,15 +50,20 @@
...
-        if (!buildableResource.isEmpty()) {
-            CloudResource resource = buildableResource.get(0);
+        return buildReservedIp(context, buildableResource, cloudStack, EXTERNAL);
+    }
+
+    public List<CloudResource> buildReservedIp(GcpContext context, List<CloudResource> buildableResource,
+            CloudStack cloudStack, String type) throws Exception {
+        List<CloudResource> result = new ArrayList<>();
+        for (CloudResource resource : buildableResource) {
             String projectId = context.getProjectId();
```
biharitomi, line 60: Could you please add logs about what resource is being built?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/AbstractGcpLoadBalancerBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+
+    protected CloudResource doOperationalRequest(CloudResource resource, ComputeRequest<Operation> request) throws IOException {
+        try {
+            Operation operation = request.execute();
+            if (operation.getHttpErrorStatusCode() != null) {
+                throw new GcpResourceException(operation.getHttpErrorMessage(), resourceType(), resource.getName());
+            }
+            return createOperationAwareCloudResource(resource, operation);
+        } catch (GoogleJsonResponseException e) {
+            throw exceptionHandlerWithThrow(e, resource.getName(), resourceType());
```
biharitomi, line 43: Can we have logs, at least for the unhappy paths?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/CreateLoadBalancerEntityHandler.java
```diff
@@ -42,6 +40,8 @@
 import com.sequenceiq.flow.reactor.api.handler.ExceptionCatcherEventHandler;
 import com.sequenceiq.flow.reactor.api.handler.HandlerEvent;
 
+import reactor.bus.Event;
```
biharitomi, line 43: As I checked only an additional space has been added to the code, why is this import needed?

core/src/main/java/com/sequenceiq/cloudbreak/reactor/api/event/stack/loadbalancer/handler/CreateLoadBalancerEntityHandler.java
```diff
@@ -98,7 +98,7 @@ protected Selectable doAccept(HandlerEvent<CreateLoadBalancerEntityRequest> even
             instanceGroups.forEach(ig -> ig.setTargetGroups(targetGroupPersistenceService.findByInstanceGroupId(ig.getId())));
             Set<LoadBalancer> existingLoadBalancers = loadBalancerPersistenceService.findByStackId(stack.getId());
             stack.setInstanceGroups(instanceGroups);
-            Set<LoadBalancer> newLoadBalancers = loadBalancerConfigService.createLoadBalancers(stack, environment, false);
+            Set<LoadBalancer> newLoadBalancers = loadBalancerConfigService. createLoadBalancers(stack, environment, false);
```
hreeve-cloudera, line 101: The extra space here is probably what's breaking all of the jenkins jobs.

hreeve-cloudera:
@yg-gd It looks like all my comments were addressed. I would like you to do one test where you create an environment without load balancers, just to make sure this new logic doesn't break that path. Once we've confirmed that I will approve the PR.

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
```diff
@@ -116,8 +120,12 @@
         //context
         ResourceBuilderContext context = contextBuilders.get(platform).contextInit(cloudContext, auth, stack.getNetwork(), cloudResources, false);
 
+        //loadBalancer
+        List<CloudResourceStatus> cloudResourceStatuses = loadBalancerResourceService.deleteResources(context, auth, cloudResources, false);
+
         //compute
-        List<CloudResourceStatus> cloudResourceStatuses = computeResourceService.deleteResources(context, auth, cloudResources, false);
+        List<CloudResourceStatus> computeStatuses = computeResourceService.deleteResources(context, auth, cloudResources, false);
```
biharitomi, line 127: What is the reason behind introducing a new group of resources?

Loadbalancers could be created with compute resource builders and the resource builder order could give the possibility to delete/create them at the right phase of resource building.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpBackendServiceResourceBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+import com.sequenceiq.cloudbreak.cloud.model.TargetGroupPortPair;
+import com.sequenceiq.common.api.type.LoadBalancerType;
+import com.sequenceiq.common.api.type.ResourceType;
+
+/**
+ * Responsible for CRUD opertaions of a GCP Backend Service Resource
+ * A backend service could have multiple definitions, but currently the only one used is based on instance groups
+ * Set the health check mentod to be used for the related instance groups
+ */
+@Service
```
biharitomi, line 36: Can we have logs with more details? 
I think we must have logging that will be handy at any time when error happens especially if it a producation issue or just would like to follow which resources have been created on the happy-path.

biharitomi (changes requested):

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
yg-gd: loadBalancers are created after the rest of the stack with the ResourceConnector.launchLoadBalancer. We don't want to accidentally recreate compute resources here. 
They could be ordered with compute resource builders. but they'd always come afterwards, so the fact that it happens in their own flow step is fine.


additionally, when building with a compute resource builder, it's tied to a CloudInstance which doesn't exist for loadBalancers, and we wouldn't want it to. Additionally building it for a specific Cloudloadbalancer lets us handle adding an additional one.

biharitomi: With Native AWS approach we did almost the same approach, used the `launchLoadBalancer` to create resources and created the compute resource builders for eliminating the load balancer related resources.
[AwsNativeLoadBalancerResourceBuilder](https://github.com/hortonworks/cloudbreak/blob/965345ea651fde41481d0cb1fb44fe7f84a3573c/cloud-aws-native/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/resource/loadbalancer/AwsNativeLoadBalancerResourceBuilder.java#L28)

If we introduce a new phase of builders for load balancers, then I would suggest to create and delete resources within the same resource builder and not using the `ResourceConnector.launchLoadBalancer` for resource creation **OR** extend the dedicated `ResourceConnector` interface with the deletion of load balancer resources and wiring it to the right flow step and in this case the modification of the resource builder engine is also avoidable. cc: @lacikaaa @doktoric @hreeve-cloudera 
On the other hand I would not touch the resource builder's logic if it is not necessary just for deleting resources that have been created by not a builder.

yg-gd: independent of where it's being used in the flow, having all the methods for creation and deletion of a resource in a single builder is useful for keeping a consistent image of how they are interacted with. 

Whether or not there needs to be a new flow step for the deletion of loadbalancers is probably outside the scope of this change.

yg-gd:
@hreeve-cloudera I was able to create an environment/datalake without loadbalancers as well just fine.

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
biharitomi: Yeah, it could be done this way, but this change is more risky as other cloud providers are also using the same logic in AbstractResourceConnector.java. So please create more unit tests for the new load balancer related builder modification in the 'common' logic and also make sure that clusters with and without load balancers could be created on AWS native variant for example.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpBackendServiceResourceBuilder.java
biharitomi: @yg-gd can we have logs in create and delete methods too??

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpForwardingRuleResourceBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+import com.sequenceiq.common.api.type.ResourceType;
+
+/**
+ * Responsible for the CRUD operations of a GCP Forwarding Rule.
+ * A Forwarding Rule maps an incoming request based on IP address, port, to a specific Backend Service
+ * Only one can exist for a given IP address and port combination in a subnet
+ *
+ */
+@Service
+public class GcpForwardingRuleResourceBuilder extends AbstractGcpLoadBalancerBuilder {
```
biharitomi, line 38: Can we have proper debug and info level logs for this class?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpHealthCheckResourceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Resposible for CRUD operations for a GCP health checker resource.
+ * Currently only regional health checks are supported.
+ * Generally only one exists for each port being used in a given stack, but not a hard rule
+ * Currently the health check is a graceful termination of TCP handshake, support exists for HTTP level health checks
+ *
+ */
+@Service
+public class GcpHealthCheckResourceBuilder extends AbstractGcpLoadBalancerBuilder {
```
biharitomi, line 33: Can we have proper warn, debug and info level logs for this class?

biharitomi (changes requested):
Please try to fix logging issues in this PR, as I found one previous review item as resolved, but the logging was almost the same as before the comment which was almost zero in the class.
Unit test could be found for resource builders, but with only testing the happy-path, can we have tests for the unhappy paths?

hreeve-cloudera (approved):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpBackendServiceResourceBuilder.java
yg-gd: I've added a general log statement for each resource builder as well as this specific builder

cloud-template/build.gradle
```diff
@@ -18,4 +18,8 @@ dependencies {
 
     compile group: 'org.slf4j',                     name: 'slf4j-api',                      version: slf4jApiVersion
 
+    testCompile        group: 'org.springframework.boot',  name: 'spring-boot-starter-test',       version: springBootVersion
```
biharitomi, line 21: Why is the `spring-boot-starter-test` needed for this module?

biharitomi (approved):

biharitomi:
I approved the PR only, one open question from my side that's been added as a single comment:
https://github.com/hortonworks/cloudbreak/pull/11133#pullrequestreview-732625781

cloud-template/build.gradle
yg-gd: added during  new tests ended up not being needed.

PR #11134, opened 2021-07-22
CB-13469 The validator of instance storage should be moved to a commo…

topolyai5:
…n service and enabled for native AWS variant.


biharitomi (approved):
LGTM

PR #11137, opened 2021-07-22
CB-13144 Implement mount and format of instance storage after instanc…

daszabo:
…e restart on AWS


nadamdb (approved):
LGTM

PR #11138, opened 2021-07-22
CB-7810 ERROR: status code: 409, message: Sdx cluster 628 has flows u…

bbihari:
…nder operation, request not allowed. ERROR: status code: 409, message: Sdx cluster 629 has flows under operation, request not allowed.

Adds better customer facing error message if operation is not allowed because flow is already running.


lnardai (approved):

lnardai:
Just a quick question:
Would it be possible to print the name of the flowchain? The reason I'm asking is because a customer might get confused if there is an ongoing repair, and receives an ScaleUp operation in progress? I'm not making this a hard requirement for this PR since this already improves on the error message greatly as is.

PR #11141, opened 2021-07-22
CB-13501. Send preflight check (network) results to EDH.

oleewere:


oleewere:
@akanto @nadamdb - because there were a lot of changes in usage.proto file in thunderhead, i have also updated usage.proto with the changes, but i have separated the commits, so it's enough to check the first commit about the proto changes, if namings are fine, i will open the PR which can be merged and do the same for thunderhead repo

nadamdb (approved):
LGTM

PR #11156, opened 2021-07-27
CB-13554 Gateways has to be repaired first (not only primary GW)

sodre90:
If we have 2 gateways and the primary is healthy, then CB are allowing to repair other host groups (like core host group / DL). It will fail because CB tries to communicate with the unhealthy GW node. As a quick fix we should validate if every GW group is healthy and do not let the repair request if the unhealthy GW node is not in the repair request.


foldik (approved):
lgtm

PR #11157, opened 2021-07-27
CB-13559. Flow stats are persisted for operation type if the flow fails inside a flow ch…

oleewere:
…ain.


tiborm (approved):
Thanks, looks good to me. +1

PR #11159, opened 2021-07-27
CB-13539 CollectDownscaleCandidatesHandler transaction is too long

horadla23:
- old behaviour: during collection of downscale candidates we called listHosts in CM then we called readHost in CM for every known instance with summary view
- new behaviour: during collection of downscale candidates we call readHosts in CM with summary view

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/CollectDownscaleCandidatesHandler.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/CollectDownscaleCandidatesHandler.java
@@ -27,7 +27,6 @@ import com.sequenceiq.cloudbreak.reactor.api.event.resource.CollectDownscaleCand
 import com.sequenceiq.cloudbreak.service.CloudbreakException;
 import com.sequenceiq.cloudbreak.service.cluster.ClusterApiConnectors;
 import com.sequenceiq.cloudbreak.service.hostgroup.HostGroupService;
-import com.sequenceiq.cloudbreak.service.stack.DefaultRootVolumeSizeProvider;
 import com.sequenceiq.cloudbreak.service.stack.InstanceMetaDataService;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/reactor/CollectDownscaleCandidatesHandler.java
```diff
@@ -160,17 +155,16 @@ private void fillDiscoveryFQDNInVolumeSetIfEmpty(List<InstanceMetaData> removabl
         }
     }
 
-    private Set<Long> collectCandidates(CollectDownscaleCandidatesRequest request, Stack stack, int defaultRootVolumeSize)
+    private Set<Long> collectCandidates(CollectDownscaleCandidatesRequest request, Stack stack)
             throws CloudbreakException {
-        LOGGER.debug("Collecting candidates for downscale based on " + CollectDownscaleCandidatesRequest.class.getSimpleName() + " [{}] and "
-                + Stack.class.getSimpleName() + " [{}] and default root volume size [{}]", request.toString(), "stackCrn: " + stack.getResourceCrn(),
-                defaultRootVolumeSize);
+        LOGGER.debug("Collecting candidates for downscale based on CollectDownscaleCandidatesRequest request [{}] and "
```
lturcsanyi, line 160: This will log `Collecting candidates for downscale based on CollectDownscaleCandidatesRequest request [CollectDownscaleCandidatesRequest{...}] ...`
Maybe a bit redundant.

lturcsanyi (approved):

core/src/main/java/com/sequenceiq/cloudbreak/reactor/CollectDownscaleCandidatesHandler.java
horadla23: yes, makes sensen

PR #11160, opened 2021-07-27
CB-13452 the namedquery was wrongly named now it is fixed and doing w…

doktoric:
…hat required. I tested with 2 node decimission from the ui.


foldik (approved):
LGTM

PR #11164, opened 2021-07-28
CB-13597 gcp can't create instances on default subnet

yg-gd:
small change to fallback to network parameters when the instance doesn't have a subnet defined yet.


sxxgrc (approved):
Looks good, thanks for making this change and adding testing!

lacikaaa:
could you add a bit more info to the commit message? like when this could happen, and also the part you added to the PR description. thanks

lacikaaa (approved):
thanks

PR #11169, opened 2021-07-29
CB-13042 FreeIPA upscale fix for GCP

cegganesh84:
1. For GCP we need to set the hostname during VM creation otherwise GCP will override it with an internal DNS name.
2. For the hostname to be set properly during the upscale operation we need to set the discovery FQDN in the instance metadata.
3. Thanks to Jamison for giving a pointer about getting the domain during the upscale operation.

Only unit testing is done, end to end test will be done in parallel.


lacikaaa:
I think this change in my PR makes this PR unnecessary: https://github.com/hortonworks/cloudbreak/pull/11084/files#diff-a0ea364c2c17c5188bb3e7b2f792ae0fd9d04e4d4a7b55b604e1e888d4f7c2d6R112

cegganesh84:
@lacikaaa Thanks for pointing out. You change should work. Have you tested in GCP?

lacikaaa:
@cegganesh84 I think I haven't. If I have time today I'll

cegganesh84:
Tested that it works for GCP also.

PR #11170, opened 2021-07-29
CB-8692 Cloud storage link on the test result page like Kibana link

aszegedi:
For easier navigation (find the related cluster logs) we need to introduce direct links to the cluster logs (freeipa, datalake, datahub) on the ReportNG Report test case result index page similarly as we have for Kibana logs.

Beyond this I've upgraded Kibana logs URL generation to the new SRE Kibana.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/listener/ReportListener.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/listener/ReportListener.java
@@ -4,24 +4,52 @@ import static com.sequenceiq.it.cloudbreak.log.Log.log;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import javax.inject.Inject;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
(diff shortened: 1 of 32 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/listener/ReportListener.java
```diff
@@ -81,6 +115,108 @@ private void logMeasurements(ITestResult tr) {
...
+                LOGGER.info("FreeIPA resource is not present in Cluster Logs");
+            } else {
+                String freeIpaCloudStorageUrl = storageUrl.getFreeIpaStorageUrl(resourceName, resourceCrn, baseLocation, cloudProvider);
+                tr.getTestContext().setAttribute(tr.getName() + FI_STORAGE_URL, freeIpaCloudStorageUrl);
+                LOGGER.info("FreeIpa cloud storage: {}:{}", tr.getName() + FI_STORAGE_URL, freeIpaCloudStorageUrl);
+            }
+        } catch (Exception e) {
+            LOGGER.info("FreeIPA resource is not present in Test Context.", e);
+        }
+    }
```
horadla23, line 136: these methods are quite similar, is it possible to create one common method with for them?

lnardai (approved):

PR #11172, opened 2021-07-29
CB-13511 adding the variant to the stack response

doktoric:


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/StackViewV4Response.java
```diff
@@ -176,4 +179,12 @@ public String getStackVersion() {
     public void setStackVersion(String stackVersion) {
         this.stackVersion = stackVersion;
     }
+
+    public String getVariant() {
```
lajosrodek, line 183: I know this is not a fault of your change, but perhaps it would make sense to add a `toString()` impl as well like for `StackV4Response`.

lajosrodek (approved):
LGTM. Only one note regarding adding a `toString()`, but it may also come in a separate PR if deemed useful.

nadamdb (approved):
LGTM

oleewere (approved):

PR #11173, opened 2021-07-29
CB-12810 add assertion to multiaz case

afarsang:

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/distrox/AvailabilityZoneAssertion.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/distrox/AvailabilityZoneAssertion.java
@@ -1,0 +1,66 @@
+package com.sequenceiq.it.cloudbreak.assertion.distrox;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.jetbrains.annotations.NotNull;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.instancegroup.InstanceGroupV4Response;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.instancegroup.instancemetadata.InstanceMetaDataV4Response;
+import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
+import com.sequenceiq.it.cloudbreak.CloudbreakClient;
+import com.sequenceiq.it.cloudbreak.assertion.Assertion;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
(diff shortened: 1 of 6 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/distrox/AvailabilityZoneAssertion.java
```diff
@@ -0,0 +1,66 @@
...
+import com.sequenceiq.it.cloudbreak.assertion.Assertion;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+import com.sequenceiq.it.cloudbreak.exception.TestFailException;
+
+public class AvailabilityZoneAssertion implements Assertion<DistroXTestDto, CloudbreakClient> {
+
+    @Override
+    public DistroXTestDto doAssertion(TestContext testContext, DistroXTestDto testDto, CloudbreakClient client) throws Exception {
+        if (testDto.getCloudPlatform() != CloudPlatform.AWS) {
```
doktoric, line 23: do we need to also check the variant?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXCustomTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Test(dataProvider = TEST_CONTEXT)
+    @UseSpotInstances
+    @Description(
+            given = "there is a running cloudbreak and existin environment, distrox, datalake",
+            when = "a valid DistroX",
+            then = "DistroX cluster is mutliaz")
+    public void testCreateDistroXWithEncryptedVolume(TestContext testContext) {
+
+        testContext
+                .given(CredentialTestDto.class).withName("aws-test-ee4d66dfa2")
```
doktoric, line 65: is this ok with fix name

doktoric (changes requested):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXCustomTest.java
afarsang: ok

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/assertion/distrox/AvailabilityZoneAssertion.java
afarsang: i haven't seen it in the response.. I chose to check the metadata availibitily zone size instead

doktoric (approved):

PR #11177, opened 2021-07-29
CB-13607 UI should show in events if a service is in bad health based…

horadla23:
… on CM response

- refactor ExtendedHostStatuses to handle various health checks
- refactor ClouderaManagerClusterStatusService to handle various health checks
- ClusterOperationService handleHelathChange can set reason for failed instanceMetadata too, also can add reason for failed intstance to event message


horadla23:
Screenshot 2021-08-02 at 12 49 58
Screenshot 2021-08-02 at 12 42 19
Screenshot 2021-08-02 at 12 42 07
Screenshot 2021-08-02 at 12 36 04

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterOperationService.java
```diff
@@ -316,55 +320,75 @@ private void handleHealthChange(Set<String> failedNodes, Set<String> newHealthyN
...
             Map<String, List<String>> autoRecoveryNodesMap, Map<String, InstanceMetaData> autoRecoveryHostMetadata,
-            Map<String, InstanceMetaData> failedHostMetadata) {
+            Map<InstanceMetaData, Optional<String>> failedHostMetadata) {
         try {
             boolean hasAutoRecoverableNodes = !autoRecoveryNodesMap.isEmpty();
             if (hasAutoRecoverableNodes) {
                 flowManager.triggerClusterRepairFlow(cluster.getStack().getId(), autoRecoveryNodesMap, false, false);
-                updateChangedHosts(cluster, autoRecoveryHostMetadata.keySet(), Set.of(SERVICES_HEALTHY), InstanceStatus.WAITING_FOR_REPAIR,
-                        CLUSTER_AUTORECOVERY_REQUESTED);
+                updateChangedHosts(cluster, autoRecoveryHostMetadata.keySet().stream().collect(Collectors.toMap(host -> host, host -> Optional.empty())),
```
lacikaaa, line 328: all the 3 `updateChangedHosts` calls and parameters are hard to follow and read. I would suggest to move all of them to separate methods, move the paramaters into variables in those methods.

also if `updateChangedHosts` handles when the 2nd parameter is empty we won't need those `if` condition like `!failedHostMetadata.isEmpty()`

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterOperationService.java
```diff
@@ -316,55 +320,75 @@ private void handleHealthChange(Set<String> failedNodes, Set<String> newHealthyN
...
             }
         } catch (TransactionExecutionException e) {
             throw new TransactionRuntimeExecutionException(e);
         }
     }
 
-    private void updateChangedHosts(Cluster cluster, Set<String> hostNames, Set<InstanceStatus> expectedState,
-            InstanceStatus newState, ResourceEvent resourceEvent) throws TransactionService.TransactionExecutionException {
-        String recoveryMessage = cloudbreakMessagesService.getMessage(resourceEvent.getMessage(), hostNames);
+    private void updateChangedHosts(Cluster cluster, Map<String, Optional<String>> hostNamesWithReason, Set<InstanceStatus> expectedState,
```
lacikaaa, line 348: it would be nice if you could do a bit of refactor here and move out some parts of this method

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterOperationService.java
```diff
@@ -316,55 +320,75 @@ private void handleHealthChange(Set<String> failedNodes, Set<String> newHealthyN
...
 
+    private String getMessageArgument(Map<String, Optional<String>> hostNamesWithReason) {
+        List<String> hostMessages = hostNamesWithReason.entrySet().stream().map(entry -> {
+            if (entry.getValue().isPresent()) {
+                return entry.getKey() + ": [" + entry.getValue().get() + "]";
+            }
+            return entry.getKey();
+        }).collect(Collectors.toList());
+        return Joiner.on(", ").join(hostMessages);
+    }
```
lacikaaa, line 390: I did a bit of refactor here, please consider it:
```
    private String getMessageArgument(Map<String, Optional<String>> hostNamesWithReason) {
        return hostNamesWithReason.entrySet().stream()
                .map(this::mapHostNameWithReasonToMessage)
                .collect(Collectors.joining(", "));
    }

    private String mapHostNameWithReasonToMessage(Map.Entry<String, Optional<String>> entry) {
        if (entry.getValue().isPresent()) {
            return String.format("%s: [%s]", entry.getKey(), entry.getValue().get());
        } else {
            return entry.getKey();
        }
    }

**Review by lacikaaa — CHANGES_REQUESTED** [review]

File: core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/flow/ClusterOperationService.java [hunk 4]
```diff
@@ -280,33 +284,33 @@ public void reportHealthChange(String crn, Set<String> failedNodes, Set<String>
         }
     }
 
-    private void handleHealthChange(Set<String> failedNodes, Set<String> newHealthyNodes, Stack stack) {
+    private void handleHealthChange(Map<String, Optional<String>> failedNodes, Set<String> newHealthyNodes, Stack stack) {
```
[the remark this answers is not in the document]
horadla23: I tried, it would be bigger change, since the autorecovery logic is a little messy

horadla23:
I'll retest this once again after changes based on Mihaly comments

horadla23:
> I'll retest this once again after changes based on Mihaly comments

retest is done

cluster-api/src/main/java/com/sequenceiq/cloudbreak/cluster/status/ExtendedHostStatuses.java
```diff
@@ -1,34 +1,73 @@
...
 
-    public Map<HostName, ClusterManagerState> getHostHealth() {
-        return hostHealth;
+    public String statusReasonForHost(HostName hostName) {
+        List<String> statusReasons = emptyIfNull(hostsHealth.get(hostName)).stream()
+                .sorted(Comparator.comparing(HealthCheck::getType))
+                .map(HealthCheck::getReason)
+                .flatMap(Optional::stream)
+                .collect(Collectors.toList());
+        return Joiner.on(" ").join(statusReasons);
```
lacikaaa, line 60: you could use `Collectors.join`

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
```diff
@@ -288,20 +288,25 @@ private boolean isCMRunning(ClusterApi connector) {
...
         return runningInstances.stream()
-                .filter(i -> failedHosts.contains(i.getDiscoveryFQDN()) || noReportHosts.contains(i.getDiscoveryFQDN()))
-                .collect(toSet());
+                .filter(i -> failedHosts.containsKey(i.getDiscoveryFQDN()) || noReportHosts.contains(i.getDiscoveryFQDN()))
+                .collect(Collectors.toMap(imd -> imd, imd -> getReasonForFailedInstance(failedHosts, imd)));
+    }
+
+    private Optional<String> getReasonForFailedInstance(Map<String, Optional<String>> failedHosts, InstanceMetaData imd) {
+        if (failedHosts.containsKey(imd.getDiscoveryFQDN())) {
+            return failedHosts.get(imd.getDiscoveryFQDN());
```
lacikaaa, line 307: I think Map#get would return null if it's not presetn, in that case you could just use:
`Optional.ofNullable(failedHosts.get(fqdn)`

horadla23: yes, it will `null`, good catch, i will check all `get`s to make sure

lacikaaa (approved):

core/src/main/java/com/sequenceiq/cloudbreak/job/StackStatusCheckerJob.java
horadla23: but there is an `if` to check if there is key for the given fqdn, so this should work

PR #11188, opened 2021-08-02
CB-13551 fix the permission of 2 Auto-TLS related files

keyki:
Cloudbreak generates 2 files to set up Auto-TLS:
- /etc/cloudera-scm-server/certs/auto-tls.init.txt
- /etc/cloudera-scm-agent/cmagent.token
Both of these files have the world-readable permission and as
part of this PR it is changed to 600. The auto-tls.init.txt can be used
by the cloudera-scm user and the cmagent.token can be used by the root user.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/start.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/start.sls
@@ -6,6 +6,15 @@ check_token:
     - name: /etc/cloudera-scm-agent/cmagent.token
     - retry:
         attempts: 10
+
+/etc/cloudera-scm-agent/cmagent.token:
+  file.managed:
+    - user: root
+    - group: root
+    - mode: 600
+    - replace: False
+    - require:
+      - file: check_token
 {% endif %}
 
 start_agent:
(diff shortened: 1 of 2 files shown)


orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/start.sls
```diff
@@ -6,6 +6,13 @@ check_token:
     - name: /etc/cloudera-scm-agent/cmagent.token
     - retry:
         attempts: 10
+
+/etc/cloudera-scm-agent/cmagent.token:
+  file.managed:
```
lacikaaa, line 11: shall we add a `require` for the previous state?

keyki: I can add if you want, but I think it will fail at that point and does not proceed with executing the following states.

lacikaaa: I'm not sure about that without failhard. I would like to avoid the case when the `check_token` state fails, but this one creates the file and next time it would succeed.
```
If source is left blank or None (use ~ in YAML), the file will be created as an empty file and the content will not be managed.

Comment (keyki) [reply orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/start.sls:11]: I added a required section
```
----------
          ID: /etc/cloudera-scm-agent/cmagent.token
    Function: file.managed
      Result: False
     Comment: One or more requisite failed: cloudera.agent.start.check_token
     Started: 15:37:03.777529
    Duration: 0.004 ms
     Changes:
----------
          ID: start_agent
    Function: service.running
        Name: cloudera-scm-agent
      Result: False
     Comment: One or more requisite failed: cloudera.agent.start.check_token
     Started: 15:37:03.778363
    Duration: 0.002 ms
     Changes:
```

lacikaaa (approved):

PR #11192, opened 2021-08-03
CB-13601 Using CM hosts api instead of cluster api to get host list for decommission

daszabo:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerDecomissioner.java
```diff
@@ -196,9 +196,9 @@ private void addInstancesNotKnownByCMWithLimit(Integer limit, Set<InstanceMetaDa
         if (hostsToRemove.size() != hostNames.size()) {
             LOGGER.debug("Not all hosts found in the given host group. [{}, {}]", hostGroup.getName(), hostNames);
         }
-        ClustersResourceApi clustersResourceApi = clouderaManagerApiFactory.getClustersResourceApi(client);
+        HostsResourceApi hostsResourceApi = clouderaManagerApiFactory.getHostsResourceApi(client);
         try {
-            ApiHostList hostRefList = clustersResourceApi.listHosts(stack.getName(), null, null, null);
+            ApiHostList hostRefList = hostsResourceApi.readHosts(null, null, null);
```
horadla23, line 201: what happens if third param (view) is null? Will it return with full view? Or with simplest view?

biharitomi: What happens if CB sends a node to be decommissioned when it is not part of CM cluster?

horadla23: it seems it is handled here in the filter:
```
Sets.newHashSet(hostsToRemove.keySet()).stream()
                    .filter(hostName -> !runningHosts.contains(hostName))
                    .forEach(hostsToRemove::remove);
```

daszabo: I set it to SUMMARY just to be sure.

daszabo: Decommission does not seem to be cluster based. It was able to decommission it with no problem.

horadla23 (approved):
LGTM

daszabo:
Do not merge yet please. There will be another change to actually delete the hosts. I just noticed that decommission really just decommission the hosts, but keeps them is CM.

daszabo:
> Do not merge yet please. There will be another change to actually delete the hosts. I just noticed that decommission really just decommission the hosts, but keeps them is CM.

DONE

biharitomi (approved):
If it is not a problem on CM side that a node is not part of the CM's cluster but decommission is triggered on it, then LGTM

PR #11193, opened 2021-08-03
CB-13523 CM agent token copy sporadically fails for base images.

lnardai:
In case of base images it's possible that the scm agent is not yet installed and the directory is not present. This causes the salt cp.get_file command to fail. Since the downscale fix introduced previously ignores whether the copy is successful or not this can lead to scaleing/provision failures.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/generate_agent_tokens.sh.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/generate_agent_tokens.sh.j2
@@ -35,7 +35,7 @@ do
   if [[ ! -e ${tokendir}/agent_token_copied_to_client ]]; then
     set +e
     # This can be fairly slow, for large clusters
-    COPY_RESULT=$(salt ${fqdn} cp.get_file salt://agent-tls-tokens/${AGENT_FQDN_TOKENDIR_MAP[${fqdn}]}/cmagent.token /etc/cloudera-scm-agent/cmagent.token)
+    COPY_RESULT=$(salt ${fqdn} cp.get_file salt://agent-tls-tokens/${AGENT_FQDN_TOKENDIR_MAP[${fqdn}]}/cmagent.token /etc/cloudera-scm-agent/cmagent.token makedirs=True)
     SALT_CP_RET=$?
     set -e
     if [[ "$COPY_RESULT" != *"False"* ]] && [[ "$SALT_CP_RET" -eq 0 ]] ; then


orchestrator-salt/src/main/resources/salt/salt/cloudera/manager/scripts/generate_agent_tokens.sh.j2
```diff
@@ -35,7 +35,7 @@ do
   if [[ ! -e ${tokendir}/agent_token_copied_to_client ]]; then
     set +e
     # This can be fairly slow, for large clusters
-    COPY_RESULT=$(salt ${fqdn} cp.get_file salt://agent-tls-tokens/${AGENT_FQDN_TOKENDIR_MAP[${fqdn}]}/cmagent.token /etc/cloudera-scm-agent/cmagent.token)
+    COPY_RESULT=$(salt ${fqdn} cp.get_file salt://agent-tls-tokens/${AGENT_FQDN_TOKENDIR_MAP[${fqdn}]}/cmagent.token /etc/cloudera-scm-agent/cmagent.token makedirs=True)
```
lacikaaa, line 38: what about the rights of the created directory? would it be the same as when it's created during installation?

lnardai: I've created 4 cluster with it, but let me check.

lajosrodek: A naive idea: Why not add a `require: - cmd: install-cloudera-manager-server` dependency to state `run_generate_agent_tokens`?

lacikaaa: this runs on the CM server instance, agent is installed everywhere. you can't wait for every instance to finish a state in saltstack (not strictly true, but the way we use it we can't enforce it)

lnardai: Checked on a fresh installment: 

On idbroker1 after a green provision, I removed the yum cloudera-manager-agent packed, remove the directory and reinstalled it with yum.
Compred to other nodes, it has the same user and group.

`
lnardai-test-dh4-broker1.lnardai.xcu2-8y8x.wl.cloudera.site:
    drwxr-xr-x    2 root     root         24 Aug  5 13:39 cloudera-scm-agent
lnardai-test-dh4-broker0.lnardai.xcu2-8y8x.wl.cloudera.site:
    drwxr-xr-x    2 root     root         92 Aug  3 16:03 cloudera-scm-agent
lnardai-test-dh4-broker2.lnardai.xcu2-8y8x.wl.cloudera.site:
    drwxr-xr-x    2 root     root         92 Aug  3 16:03 cloudera-scm-agent
`

lacikaaa (approved):

PR #11195, opened 2021-08-03
DISTX-678 - Set hive.metastore.try.direct.sql.ddl=true explicitly in Hive templates

juhi-09:
DISTX-678 - Set hive.metastore.try.direct.sql.ddl=true explicitly in Hive templates
    As part of CDPD-25578, "hive.metastore.try.direct.sql.ddl" was explicitly set to true for "hive_service_config_safety_valve" - used by hive
    This PR sets "hive.metastore.try.direct.sql.ddl" as true for "hive_metastore_config_safety_valve" to be used by hive metastore.

**Property set for "hive_metastore_config_safety_valve"**
Screenshot 2021-08-09 at 5 56 07 PM

**Property set for "hive_service_config_safety_valve" - already as part of CDPD-25578**
Screenshot 2021-08-09 at 5 56 59 PM

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
@@ -130,7 +130,13 @@
           {
             "refName": "hms-HIVEMETASTORE-BASE",
             "roleType": "HIVEMETASTORE",
-            "base": true
+            "base": true,
+            "configs": [
+              {
+                "name": "hive_metastore_config_safety_valve",
+                "value": "<property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>"
+              }
+            ]
           }
         ]
       },
--- a/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-spark3.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-spark3.bp
@@ -259,7 +259,13 @@
           {
             "refName": "hms-HIVEMETASTORE-BASE",
             "roleType": "HIVEMETASTORE",
-            "base": true
+            "base": true,
+            "configs": [
+              {
+                "name": "hive_metastore_config_safety_valve",
+                "value": "<property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>"
+              }
+            ]
           }
         ]
       },
--- a/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering.bp
@@ -249,7 +249,13 @@
           {
             "refName": "hms-HIVEMETASTORE-BASE",
             "roleType": "HIVEMETASTORE",
-            "base": true
+            "base": true,
+            "configs": [
+              {
+                "name": "hive_metastore_config_safety_valve",
+                "value": "<property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>"
+              }
+            ]
           }
         ]
       },
--- a/core/src/main/resources/defaults/blueprints/7.2.12/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.12/cdp-data-engineering-ha.bp
@@ -130,7 +130,13 @@
           {
             "refName": "hms-HIVEMETASTORE-BASE",
             "roleType": "HIVEMETASTORE",
-            "base": true
+            "base": true,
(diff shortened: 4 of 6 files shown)


amitagarwal06:
@juhi-09 You are setting hive.metastore.try.direct.sql=true in a safety valve of Hive service but images in this PR show this config from HMS service.
hive_metastore_database_datanucleus_try_direct_sql config in HMS by default is set to true, so does not require any changes.

Here is the beeline output of default values for both the configs from a 7.2.10 cluster

0: jdbc:hive2://p-temp-master0.perf-aws.xcu2-> set hive.metastore.try.direct.sql;
+-------------------------------------+
|                 set                 |
+-------------------------------------+
| hive.metastore.try.direct.sql=true  |
+-------------------------------------+
1 row selected (0.017 seconds)
0: jdbc:hive2://p-temp-master0.perf-aws.xcu2-> set hive.metastore.try.direct.sql.ddl;
+-----------------------------------------+
|                   set                   |
+-----------------------------------------+
| hive.metastore.try.direct.sql.ddl=true  |
+-----------------------------------------+

@rbalamohan If beeline shows correct value for both the configs, do you think any fix is needed?

cegganesh84 (approved):
LGTM.

Two observations no need to address them in the commit,

1. Newlines within a json string could be avoided. Breaks the custom template creator UI.
2. 7.2.10 is a released version. Sometimes changing a released version goes untested to production because the testing for the upcoming version has started.

@amitagarwal06 I do not see a way to set the service config for HMS except for setting it via Hive. I can only set role configs for HMS directly.

Per Rajesh both the setting needs to be explicitly set. So this change is needed. I guess it is related to the point above, as setting Hive is not propagated to HMS unless explicitly set via service config safety valve, which this change is doing correctly as shown by Juhi's screenshot.

CC: @sidseth

amitagarwal06:
@cegganesh84 If you look at the second screenshot Juhi has posted that is exposing hive.metastore.try.direct.sql as a service-wide config directly for HMS.

Anyway there is more to this than what is apparent. Here is my observation -
One can open hive-site.xml in CM UI via Hive & HMS processes separately. Both copies of hive-site.xml are different. Here are values of both the configs from their separate hive-site.xml -

From Hive
----------
  
  <property>
    <name>hive.metastore.try.direct.sql.ddl</name>
    <value>true</value>
  </property>
  <property>
    <name>hive.metastore.try.direct.sql</name>
    <value>true</value>
  </property>

From HMS
-----------
  <property>
    <name>hive.metastore.try.direct.sql.ddl</name>
    <value>false</value>
  </property>
  <property>
    <name>hive.metastore.try.direct.sql</name>
    <value>true</value>
  </property>

1. hive.metastore.try.direct.sql - As one can see its value is correctly set to true in both the places. So it does not require any change.
2. hive.metastore.try.direct.sql.ddl - This is set to true for Hive as it was changed in the prior JIRA CDPD-25578, but its false for HMS. So if any changes have to be done it should be for HMS only.

I dont know if both Hive & HMS need this config or only one. If both need it then it appears to me that it should be set for both the services individually via their own safety valves. If only one service needs it then we should set it only in the safety valve of that service. @rbalamohan can confirm what should be done.

cegganesh84:
@amitagarwal06 I see what you mean. The fix for CDPD-25578 did not propagate to HMS, that is the real issue here. I will add a comment like that.

core/src/main/resources/defaults/blueprints/7.2.10/cdp-data-engineering-ha.bp
```diff
@@ -147,7 +147,8 @@
             "name": "hive_service_config_safety_valve",
             "value": "<property><name>fs.s3a.ssl.channel.mode</name><value>openssl</value></property><property><name>hive.txn.acid.dir.cache.duration</name><value>0</value></property><property><name>hive.server2.tez.session.lifetime</name><value>30m</value></property>
                         <property><name>hive.blobstore.supported.schemes</name><value>s3,s3a,s3n,abfs,gs</value></property><property><name>hive.orc.splits.include.fileid</name><value>false</value></property><property><name>hive.hook.proto.events.clean.freq</name><value>1h</value></property>
-                        <property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>"
+                        <property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>
```
cegganesh84, line 150: Looks like this value was not propagated to HMS as shown by @amitagarwal06 .

You could set this value in HMS safety vavle and try to export the template using CM API. A nice tool is available in CM UI left hand side support tab.

sidseth:
I think the change needs to be made to the metastore component in the template - hms-HIVEMETASTORE-BASE - for it to be used by HMS.

sidseth (approved):

PR #11200, opened 2021-08-03
CB-12701 Add support for registering CCMV2 Jumpgate type clusters

bergerdenes:
Updated cluster-proxy request and response objects according to the latest
image available.
Fixed typo in CCMV2 proto file.
Added fallback for CCM tunnel type.
Changed cluster proxy registration mode depending on CCMv2 or Jumpgate:
for Jumpgate, the Environment CRN must be passed but the CCMv2 config
should not, only the CCMv2 boolean flag must be set.
Old CCMv2 registration should remain the same.

Merge only after proper images are ready for FreeIPA AND base/prewarm

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
@@ -46,7 +46,10 @@ import com.sequenceiq.common.api.type.Tunnel;
 
 @Component
 public class ClusterProxyService {
-    public static final String CB_INTERNAL = "cb-internal";
+
+    private static final String CB_INTERNAL = "cb-internal";
+
+    private static final String CLOUDERA_MANAGER_SERVICE_NAME = "cloudera-manager";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClusterProxyService.class);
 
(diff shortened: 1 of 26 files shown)


lacikaaa:
moved to draft because of the label

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
```diff
@@ -140,22 +140,36 @@ private Environment initializeEnvironment(EnvironmentCreationDto creationDto) {
...
+                experimentalFeatures.setTunnel(tunnel);
+                environment.setExperimentalFeaturesJson(experimentalFeatures);
             }
+            // TODO: if we want the Jumpgate to be default then this logic needs to be added for CCMV2 as well
         }
         LOGGER.info("Environment is initialized with [{}] tunnel.", tunnel);
     }
 
+    private void checkCcmEntitlements(Tunnel tunnel, boolean ccmV2Enabled, boolean ccmV2JumpgateEnabled) {
+        if (Tunnel.CCMV2_JUMPGATE == tunnel && !ccmV2JumpgateEnabled) {
```
lacikaaa, line 165: `ccmV2JumpgateEnabled` requires `ccmV2Enabled` to be true or are these totally different, and a customer can have jumpgate without the `ccmV2Enabled` entitlement?

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
```diff
@@ -115,23 +118,42 @@ public void deregisterCluster(Stack stack) {
...
-        } else if (stack.getTunnel().useCcmV2OrJumpgate()) {
-            requestBuilder.withAccountId(getAccountId(stack)).withServices(serviceConfigsForCcmV2(stack)).withCcmV2Entries(ccmV2Configs(stack));
+            requestBuilder.withAccountId(getAccountId(stack))
+                    .withTunnelEntries(tunnelEntries(stack));
+        } else if (stack.getTunnel().useCcmV2()) {
+            requestBuilder.withAccountId(getAccountId(stack))
+                    .withServices(serviceConfigsForCcmV2(stack))
+                    .withCcmV2Entries(ccmV2Configs(stack));
+        } else if (stack.getTunnel().useCcmV2Jumpgate()) {
+            requestBuilder.withEnvironmentCrn(stack.getEnvironmentCrn())
```
lacikaaa, line 131: while I see the env crn here, I miss it on the FreeIPA side. I thought I would see it during FreeIPA CCMv2 registration in case it's a jumpgate type, or always

environment/src/main/java/com/sequenceiq/environment/environment/service/EnvironmentCreationService.java
bergerdenes: ccmV2JumpgateEnabled is a standalone entitlement and if it is set on customer account it will work

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
bergerdenes: I discussed with @zabhi  and he suggested the way it is coded now. That is: freeipa registration **should not have** the environmentCrn but has the CCMv2config and isCCMv2 enabled. The dl/dh registration **should have** the environmentCrn but no CCMv2Config. The isCCMv2 flag must be enabled too. This is also tested by unit tests.

lacikaaa: ok, I looked it up in the code, so actually CCMv2 registration happens on FreeIPA with an env crn, while it happens without it on CB side. So it must be how it's connected.

bergerdenes: vica versa but yes. freeipa without env crn, CB core with env crn

lacikaaa: I was talking about CCMv2 reg, you are talking about CP reg. but both of us is right :)

bergerdenes: Yes right :) I have not checked the CCMv2 inverting proxy agent registration so that is unchanged and works as you have described.

doktoric (approved):

bergerdenes:
Images are ready

lacikaaa (approved):

PR #11204, opened 2021-08-04
CB-13532: Modified backup-restore scripts to work on a given database.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-13532

This change is primarily focused on extending the existing backup/restore scripts to be usable with a given database. Specifically, I have made it so the scripts take in an optional database name argument and, when the argument is _not_ provided, the default action is to backup the hive and ranger databases (the existing behavior).

Part of this change includes changes to the existing salt files in order to work correctly with the new optional argument. A side effect of this change is that the other optional argument, the log file location, can now only be used when the database name argument is given as well. We believe that this is okay as this argument is not used much in practice.

I have performed general testing of these changes, but not any rigorous QE tests, so I hope that they can be done once this is merged in.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
+++ b/orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
@@ -1,2 +1,4 @@
 disaster_recovery:
   object_storage_url:
+  ranger_admin_group:
+  database_name:
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
@@ -6,7 +6,7 @@ include:
 {% if 'None' != configure_remote_db %}
 backup_postgresql_db:
   cmd.run:
-    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} {{salt['pillar.get']('postgres:remote_db_url')}} {{salt['pillar.get']('postgres:remote_db_port')}} {{salt['pillar.get']('postgres:remote_admin')}} {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}}
+    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} {{salt['pillar.get']('postgres:remote_db_url')}} {{salt['pillar.get']('postgres:remote_db_port')}} {{salt['pillar.get']('postgres:remote_admin')}} {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}} {{salt['pillar.get']('disaster_recovery:database_name') or ''}}
     - require:
       - sls: postgresql.disaster_recovery
 
@@ -23,7 +23,7 @@ add_root_role_to_database:
 
 backup_postgresql_db:
   cmd.run:
-    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} "" "" "" {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}} "/var/lib/pgsql"
+    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} "" "" "" {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}} {{salt['pillar.get']('disaster_recovery:database_name') or ''}} "/var/lib/pgsql"
     - require:
       - cmd: add_root_role_to_database
 {% endif %}
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -8,15 +8,16 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ $# -ne 5 && $# -ne 6 ]]; then
+if [[ $# -lt 5 || $# -gt 7 ]]; then
   echo "Invalid inputs provided"
-  echo "Script accepts 5 inputs:"
+  echo "Script accepts at least 5 and at most 7 inputs:"
   echo "  1. Object Storage Service url to place backups."
   echo "  2. PostgreSQL host name."
   echo "  3. PostgreSQL port."
   echo "  4. PostgreSQL user name."
   echo "  5. Ranger admin group."
-  echo "  6. (optional) Log file location."
+  echo "  6. (optional) Name of the database to backup. If not given, will backup ranger and hive databases."
+  echo "  7. (optional) Log file location. Must be provided along with a database name."
   exit 1
 fi
 
@@ -25,7 +26,8 @@ HOST="$2"
 PORT="$3"
 USERNAME="$4"
 RANGERGROUP="$5"
-LOGFILE=${6:-/var/log/}/dl_postgres_backup.log
+DATABASENAME="${6-}"
+LOGFILE=${7:-/var/log}/dl_postgres_backup.log
 echo "Logs at ${LOGFILE}"
 
 {%- from 'postgresql/settings.sls' import postgresql with context %}
@@ -142,8 +144,14 @@ run_backup() {
   DATE_DIR=${BACKUPS_DIR}/$(date '+%Y-%m-%dT%H:%M:%SZ')
   mkdir -p "$DATE_DIR" || error_exit "Could not create local directory for backups."
 
-  backup_database_for_service "hive"
-  backup_database_for_service "ranger"
+  if [[ -z "$DATABASENAME" ]]; then
+    echo "No database name provided. Will backup hive and ranger databases."
+    backup_database_for_service "hive"
+    backup_database_for_service "ranger"
+  else
+    echo "Backing up ${DATABASENAME}."
+    backup_database_for_service "${DATABASENAME}"
+  fi
 
   rmdir -v "$DATE_DIR" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2)
 }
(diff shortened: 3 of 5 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -8,15 +8,16 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ $# -ne 5 && $# -ne 6 ]]; then
+if [[ $# -lt 5 || $# -gt 7 ]]; then
```
brycederriso, line 11: **praise:** Thanks for using ranges instead of the equality with `&&`.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/backup.sls
```diff
@@ -23,7 +23,7 @@ add_root_role_to_database:
 
 backup_postgresql_db:
   cmd.run:
-    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} "" "" "" {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}} "/var/lib/pgsql"
+    - name: /opt/salt/scripts/backup_db.sh {{salt['pillar.get']('disaster_recovery:object_storage_url')}} "" "" "" {{salt['pillar.get']('disaster_recovery:ranger_admin_group')}} {{salt['pillar.get']('disaster_recovery:database_name')}} "/var/lib/pgsql"
```
brycederriso, line 26: I think this'll cause an issue when we don't provide the `database_name`. IIRC when that's not provided, the actual value here will be a space, and we'll have 6 arguments.

In that case, the 6th argument will be the log location base, and `DATABASENAME="${6-}"` in the backup script will evaluate to `/var/lib/pgsql`.

A possible way out of this might be to use named argument parameters rather than the positional ones, so that we wouldn't have to worry about ordering.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -8,15 +8,16 @@ set -o nounset
...
   echo "Invalid inputs provided"
-  echo "Script accepts 5 inputs:"
+  echo "Script accepts at least 5 and at most 7 inputs:"
   echo "  1. Object Storage Service url to place backups."
   echo "  2. PostgreSQL host name."
   echo "  3. PostgreSQL port."
   echo "  4. PostgreSQL user name."
   echo "  5. Ranger admin group."
-  echo "  6. (optional) Log file location."
+  echo "  6. (optional) Name of the database to backup. If not given, will backup ranger and hive databases."
```
brycederriso, line 19: **suggestion:** I think we should document the special behavior of `None` being passed here. Namely that it causes us to backup the `hive` and `ranger` dbs as normal.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -7,15 +7,16 @@ set -o nounset
...
   echo "Invalid inputs provided"
-  echo "Script accepts 5 inputs:"
+  echo "Script accepts at least 5 and at most 7 inputs:"
   echo "  1. Object Storage Service url to retrieve backups."
   echo "  2. PostgreSQL host name."
   echo "  3. PostgreSQL port."
   echo "  4. PostgreSQL user name."
   echo "  5. Ranger admin group."
-  echo "  6. (optional) Log file location."
+  echo "  6. (optional) Name of the database to restore. If not given, will restore ranger and hive databases."
```
brycederriso, line 18: suggestion: I think we should document the special behavior of None being passed here. Namely that it causes us to restore the hive and ranger dbs as normal.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/restore_db.sh
```diff
@@ -138,8 +140,14 @@ run_restore() {
 
   hdfs dfs -copyToLocal -f "$BACKUP_LOCATION" "$BACKUPS_DIR" > >(tee -a $LOGFILE) 2> >(tee -a $LOGFILE >&2) || errorExit "Could not copy backups from ${BACKUP_LOCATION}."
 
-  restore_db_from_local "hive"
-  restore_db_from_local "ranger"
+  if [[ -z "$DATABASENAME" || "$DATABASENAME" == "None" ]]; then
+    echo "No database name provided. Will restore hive and ranger databases."
```
brycederriso, line 144: **praise:** This is the punchline of the changes, and it looks good to me.

brycederriso (changes requested):
Just need to comment that `None` behaves strangely as an argument. Otherwise the changes look good.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
kkalvagadda1: @brycederriso Where do you suggest we document?

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
```diff
@@ -1,2 +1,4 @@
 disaster_recovery:
   object_storage_url:
+  ranger_admin_group:
```
kkalvagadda1, line 3: @brycederriso Why was ranger_admin_group missing here?

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
brycederriso: The `.sls` files use Python Jinja to plugin values, so if there's no value it actually does plug in `None`. Santiago tested this and that was what he was getting back.

I did find [this stack overflow](https://stackoverflow.com/questions/19614027/jinja2-template-variable-if-none-object-set-a-default-value) that seems to provide a way to provide an empty value rather than the `None` value. @sxxgrc.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -8,15 +8,16 @@ set -o nounset
...
-  echo "Script accepts 5 inputs:"
+  echo "Script accepts at least 5 and at most 7 inputs:"
   echo "  1. Object Storage Service url to place backups."
   echo "  2. PostgreSQL host name."
   echo "  3. PostgreSQL port."
   echo "  4. PostgreSQL user name."
   echo "  5. Ranger admin group."
-  echo "  6. (optional) Log file location."
+  echo "  6. (optional) Name of the database to backup. If not given, or if 'None' is given, will backup ranger and hive databases."
+  echo "  7. (optional) Log file location. Must be provided along with a database name."
```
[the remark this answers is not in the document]
brycederriso: We had a request to include this in the backup IIRC, I'm not sure where it's actually being used, but it was a request I had gotten to include in the scripts.

Maybe we should open a ticket to investigate if it's necessary, I'm not sure how to check where it's being used.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -8,15 +8,16 @@ set -o nounset
...
   echo "Invalid inputs provided"
-  echo "Script accepts 5 inputs:"
+  echo "Script accepts at least 5 and at most 7 inputs:"
   echo "  1. Object Storage Service url to place backups."
   echo "  2. PostgreSQL host name."
   echo "  3. PostgreSQL port."
   echo "  4. PostgreSQL user name."
   echo "  5. Ranger admin group."
-  echo "  6. (optional) Log file location."
+  echo "  6. (optional) Name of the database to backup. If not given, or if 'None' is given, will backup ranger and hive databases."
```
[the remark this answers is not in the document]
sxxgrc: As Bryce mentioned above, None is passed through Salt when the sls file contains nothing. I am going to quickly try to make it send over an empty string but not sure if I will be able to do that easily.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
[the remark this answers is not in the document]
sxxgrc: I think it would be good to open a ticket to investigate if it's used, but leave it as it is for now.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
sxxgrc: Thanks for the stack overflow link Bryce, I think I'll be able to get it working without any of the "None" stuff.

sxxgrc:
I have modified the backup and restore salt scripts so as to return an empty string instead of None when the primary sls file has nothing for the database name argument. I have verified this locally. 
I will test the local postgres next.

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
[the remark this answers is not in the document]
kkalvagadda1: I'm fine with it. please open a Jira accordingly.
Fact that you changed the logic to pass an empty string for the database name, please update the Help message accordingly.

orchestrator-salt/src/main/resources/salt/pillar/postgresql/disaster_recovery.sls
brycederriso: The values in this pillar aren't actually needed from what I can tell, so I forgot to add this when we did the first pass.

When the salt pillar update related code in CB runs, it _completely replaces_ the values in this file with the appropriate JSON values.

I think the intent of this file is to provide reasonable defaults so that we can make salt calls without first running salt pillar updates, but we're not doing that for backup/restore.

We can also kinda-sorta use these properties as documentation to show what we _will_ expect, and it makes it somewhat easier to experiment with salt scripts on a CB managed node -- you can manually edit the salt pillar and you have a starting key at least.

brycederriso (approved):
Based on our conversations, I'm happy with this.

Thanks Santiago!

kkalvagadda1 (approved):

PR #11106, opened 2021-07-20
CB-13110:Add the encryption Arn parameter passed from CLI to AWS env creation in Cloudbreak

satyabollineni:
Created a Environment with the encryption Key Arn from CLI and verified that the parameter encryption-key-arn is added to the DB:Screenshot 2021-07-29 at 5 21 34 PM

PR #11205, opened 2021-08-05
CB-13109:Add cloudbreak validations for Encryption Arn passed from CLI

satyabollineni:
Once the PR for CB-13110 https://github.com/hortonworks/cloudbreak/pull/11106 is merged to master I will remove all those changes from this PR and just leave the validation changes in this PR

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
@@ -1,0 +1,28 @@
+package com.sequenceiq.environment.environment.validation.validators;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.validation.ValidationResult;
+
+@Component
+public class EncryptionKeyArnValidator {
+
+    private static final Pattern ENCRYPTION_KEY_ARN_PATTERN = Pattern.compile("^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:" +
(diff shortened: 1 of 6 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.regex.Pattern;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.cloudbreak.validation.ValidationResult;
+
+@Component
+public class EncryptionKeyArnValidator {
+
+    private static final Pattern ENCRYPTION_KEY_ARN_PATTERN = Pattern.compile("^arn:(aws|aws-cn|aws-us-gov):kms:[a-zA-Z0-9-]+:[0-9]+:" +
```
juhi-09, line 13: I think, for AWS we should check the existence of the key directly - as APIs are available. We can check if the keyArn provided by user is present in `aws kms list-keys`.
A little background, For Azure we validated the format of key because the existence validation is not possible(at least I am not aware of) currently - CB-13810

juhi-09 (changes requested):

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
satyabollineni: A separate jira is created and being worked on to validate the key existence in AWS https://jira.cloudera.com/browse/CB-13464?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

juhi-09 (commented):
Encryption key validation changes LGTM, Thanks for the fixes.

PR #11210, opened 2021-08-05
CB-13680: Custom catalog list returns the cloudbreak-default

jszeles:
Custom catalog 'list' cli command also returns cloudbreak-default which unfortunately is not a custom catalog.

### DIFF
--- a/core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
+++ b/core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
@@ -121,11 +121,15 @@ public class ImageCatalogServiceTest {
 
     private static final String CDP_DEFAULT_CATALOG_NAME = "cdp-default";
 
+    private static final String CATALOG_NAME = "test-catalog";
+
     private static final String CUSTOM_CATALOG_NAME = "custom-catalog";
 
     private static final String CUSTOM_BASE_PARCEL_URL = "https://myarchive.test.com";
 
     private static final String CUSTOM_IMAGE_ID = "customImageId";
+
+    private static final String USER_CRN = "crn:altus:iam:us-west-1:1111:user:1111";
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
```diff
@@ -121,12 +122,18 @@
...
 
+    private static final String CATALOG_NAME = "test-catalog";
+
     private static final String CUSTOM_CATALOG_NAME = "custom-catalog";
 
     private static final String CUSTOM_BASE_PARCEL_URL = "https://myarchive.test.com";
 
     private static final String CUSTOM_IMAGE_ID = "customImageId";
 
+    private static final String UUID = java.util.UUID.randomUUID().toString();
```
pkedvessy, line 133: It is not needed to generate a real UUID, you can hardcode a static one into the USER_CRN or simply copy what we have in the EnvironmentApiConverterTest.

core/src/test/java/com/sequenceiq/cloudbreak/service/image/ImageCatalogServiceTest.java
```diff
@@ -877,6 +891,60 @@ public void testGetImageShouldLookupCustomImageInCaseOfNullImageCatalogUrl()
         assertEquals(statedImage.getImage(), actual.getImage());
     }
 
+    @Test
+    public void testFindAllByIdsWithDefaultsInCaseOfCustomCatalogsOnly() {
+        ImageCatalog customImageCatalog = new ImageCatalog();
```
pkedvessy, line 896: This first "block" is duplicated in the next test case. Please create a method like getImageCatalogs and you can use it in both test cases.

pkedvessy (changes requested):

pkedvessy (approved):

PR #11211, opened 2021-08-05
DISTX-679: Turn off hive.privilege.synchronizer by default in templates

juhi-09:
DISTX-679: Turn off hive.privilege.synchronizer by default in templates

This change turns off "hive.privilege.synchronizer" by default. This property is used mainly for information schema population which users are not using actively. This can be enabled by users, in case they want to use information_schema.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
@@ -153,7 +153,7 @@
             "name": "hive_service_config_safety_valve",
             "value": "<property><name>fs.s3a.ssl.channel.mode</name><value>openssl</value></property><property><name>hive.txn.acid.dir.cache.duration</name><value>0</value></property><property><name>hive.server2.tez.session.lifetime</name><value>30m</value></property>
                         <property><name>hive.blobstore.supported.schemes</name><value>s3,s3a,s3n,abfs,gs</value></property><property><name>hive.orc.splits.include.fileid</name><value>false</value></property><property><name>hive.hook.proto.events.clean.freq</name><value>1h</value></property>
-                        <property><name>hive.metastore.try.direct.sql.ddl</name><value>true</value></property>"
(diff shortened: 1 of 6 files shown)


amitagarwal06:
@rbalamohan Juhi is setting this in a safety valve of Hive service. But should this config go to HMS service or Hive service?

rbalamohan:
Setting is for HMS service trying to communicate with DB in native SQL mode.

sidseth:
@juhi-09  - could you please change this to be HMS specific, similar to the directsql change.

rbalamohan:
I guess my comment got mixed up for direct SQL. 

Priv sync should be in HS2 setting. Sorry about the confusion.

juhi-09:
> Priv sync should be in HS2 setting. 

@rbalamohan I have updated the property to be present in config of `hive-HIVESERVER2-BASE`, however it is not visible in UI. Please review.

core/src/main/resources/defaults/blueprints/7.2.11/cdp-data-engineering-ha.bp
```diff
@@ -182,6 +182,10 @@
               {
                 "name": "hiveserver2_idle_session_timeout",
                 "value": 14400000
+              },
+              {
+                "name": "hive.privilege.synchronizer",
```
amitagarwal06, line 187: Setting it this way is not correct. Its not applied to HS2. If you open hive-site.xml from UI, you wont see it getting set anywhere. You wont see it getting set via beeline either

0: jdbc:hive2://perf-test2-master0.perf-aws.x> set hive.privilege.synchronizer;
+-----------------------------------+
|                set                |
+-----------------------------------+
| hive.privilege.synchronizer=true  |
+-----------------------------------+

Right way to set this is as you did it first time. Add this property to hive_service_config_safety_valve

amitagarwal06 (changes requested):
@juhi-09 Pls make the changes as suggested in the review and re-submit.

cegganesh84:
@juhi-09 Please pull in the recent changes to these files, otherwise, you will not be able to merge the changes.

amitagarwal06 (approved):
Have approved it. Not sure of whether failing swagger tests matter.

cegganesh84:
I guess it will go away if the code is rebased with the latest changes in the master.

PR #11213, opened 2021-08-05
CB-13646 In case of stopped clusters, do not try to terminate CM and …

bergerdenes:
…hosts

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/Status.java
+++ b/core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/Status.java
@@ -93,10 +93,8 @@ public enum Status {
 
     public boolean isStopState() {
         return Status.STOPPED.equals(this)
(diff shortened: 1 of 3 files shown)


core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/common/Status.java
```diff
@@ -93,10 +93,8 @@ public boolean isStartState() {
 
     public boolean isStopState() {
         return Status.STOPPED.equals(this)
-                || UPDATE_IN_PROGRESS.equals(this)
                 || Status.STOP_IN_PROGRESS.equals(this)
-                || Status.STOP_REQUESTED.equals(this)
-                || Status.STOP_FAILED.equals(this);
```
doktoric, line 99: why we removed STOP_REQUESTED ?

bergerdenes: we did not

doktoric (approved):

PR #11223, opened 2021-08-06
CB-13514 - Add updated load balancer metadata collection to AwsNative classes

hreeve-cloudera:
Adds a new AwsNativeLbMetadataCollector to collect the AWS LB specific metadata to later
be persisted to the database.

Tested with unit tests.


yg-gd:
One thought I had with this, is that these are the values from when it was created. if some detail of a targetgroup/listener changed out from under us the metadata wouldn't be updated.

hreeve-cloudera:
> One thought I had with this, is that these are the values from when it was created. if some detail of a targetgroup/listener changed out from under us the metadata wouldn't be updated.

At the moment, there's no way for that to happen. Listeners and target groups are static. Instances can be added to or removed from the target group, but the target group construct exists either way, and does not change. Note that this is true for AWS; I'm not sure if the other cloud providers have metadata that can change on us.

If we ever add logic that deletes/recreates listeners or target groups, we would have to add new code to update the metadata appropriately.

yg-gd (approved):

biharitomi (approved):
LGTM

PR #11224, opened 2021-08-07
CB-13638: Restart CM management services before running CDH runtime upgrade

tiborpopovics:


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/ClouderaManagerModificationService.java
```diff
@@ -215,10 +215,11 @@ public void upgradeClusterRuntime(Set<ClusterComponent> components, boolean patc
                 downloadAndActivateParcels(products, parcelResourceApi, true);
                 restartServices(clustersResourceApi);
             } else {
+                restartMgmtServices();
```
keyki, line 218: This should be moved out from the if condition

keyki (approved):

PR #11235, opened 2021-08-10
CB-13619 decreasing list-cluster-template operation runtime

afarsang:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/init/blueprint/BlueprintLoaderService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/init/blueprint/BlueprintLoaderService.java
@@ -15,6 +15,7 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
+import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/init/blueprint/BlueprintLoaderService.java
```diff
@@ -42,17 +43,23 @@
     @Inject
     private ClusterTemplateService clusterTemplateService;
 
+    private Map<String, Blueprint> defaultBlueprints;
+
+    @PostConstruct
+    public void getBlueprintsFromCache() {
+        defaultBlueprints = Map.copyOf(defaultBlueprintCache.defaultBlueprints());
```
doktoric, line 50: defaultBlueprints are also initialized there in postconstruct. Can you please explain why it helps ?

afarsang: getting defaultBlueprints is kind of slow due to :
        defaultBlueprints.forEach((key, value) -> result.put(key, SerializationUtils.clone(value)));

but this list is not meant to change, so it is okay to get it only once (at startup), right?

doktoric: yep

doktoric (approved):

PR #11238, opened 2021-08-10
CB-13733 Cluster stays in UPDATE_IN_PROGRESS state after validation e…

lacikaaa:
…rror

- set stack and cluster status to `AVAILABLE` after validation failure
- add logs
- smaller refactors

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/upgrade/validation/ClusterUpgradeImageValidationEvent.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/upgrade/validation/ClusterUpgradeImageValidationEvent.java
@@ -35,4 +35,12 @@ public class ClusterUpgradeImageValidationEvent extends ClusterUpgradeValidation
         return cloudContext;
     }
 
+    @Override
+    public String toString() {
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/datalake/upgrade/validation/ClusterUpgradeImageValidationEvent.java
```diff
@@ -35,4 +35,12 @@ public CloudContext getCloudContext() {
         return cloudContext;
     }
 
+    @Override
+    public String toString() {
+        return "ClusterUpgradeImageValidationEvent{" +
+                "cloudCredential=" + cloudCredential +
```
keyki, line 41: Have you checked if we're logging any sensitive here?

lacikaaa: haven't until now: https://github.com/hortonworks/cloudbreak/blob/master/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/CloudCredential.java#L60
so it should be safe

PR #11244, opened 2021-08-13
Cb 13658 reattach volumes

topolyai5:


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpResourceConnector.java
```diff
@@ -73,36 +68,8 @@ public String getDBStackTemplate() throws TemplatingNotSupportedException {
     }
 
     @Override
-    public List<CloudResourceStatus> upscale(AuthenticatedContext auth, CloudStack stack, List<CloudResource> resources) {
```
doktoric, line 76: why we removed this?

topolyai5: I moved this to abstract

topolyai5: I have reused the abstract method. The upscale is almost the same, the only difference is the disk reattachment. I refactored the disk reattachment into a protected method and this method can easily turn off with an override. 

However I have checked, the Azure use a different implementation, the AWS and GCP use the same implementation included the disk reattachment

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
```diff
@@ -180,9 +185,29 @@ public ExternalDatabaseStatus getDatabaseServerStatus(AuthenticatedContext authe
...
+            CloudResource cloudResource = diskSets.get(i);
+            Optional.ofNullable(scalingGroup.getInstances().get(i))
+                    .map(CloudInstance::getTemplate)
+                    .map(InstanceTemplate::getPrivateId)
+                    .ifPresent(privateId -> context.addComputeResources(privateId, List.of(cloudResource)));
+        }
+    }
+
+    protected ResourceType getDisResourceType() {
+        throw new IllegalArgumentException("Please override the disk resource type or implement the provider specified disk reattachment");
```
daszabo, line 208: Can you you delegate the responsibilty to implenetors by using an abstract method here?

cloud-aws-common/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/common/resource/volume/AwsVolumeResourceBuilder.java
```diff
@@ -238,7 +238,7 @@ private String getAvailabilityZoneFromSubnet(AuthenticatedContext auth, String s
         }
         LOGGER.debug("Volume creation requests sent");
 
-        return buildableResource.stream()
+        return requestedResources.stream()
```
daszabo, line 241: I am not sure if this is correct. During repair, the context is populated with the already existing disks. I think the status of those is not REQUESTED, rather DETACHED or stg like that. During repair, those are just passed through this builder without modification and the attachment builder will be responsible to attach those disks to an instance. Have you tried repair with this change?

daszabo (changes requested):

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
topolyai5: It is mandatory. We need to force them to implement if they want to use the disk reattachment out of the box. As the message says.

cloud-aws-common/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/common/resource/volume/AwsVolumeResourceBuilder.java
topolyai5: I'm implementing the repair with multi-az.
I have found a bug if any errors occur during the repair flow, the DETACHED volumes statuses will be set to CREATED. After the restart, the resource builder won't find the disk because we try to find it with DETACHED status and the volumes won't reattach but create attached disks again.

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
daszabo: I understand that, what I meant is people need to remember to implement this. The exception will only come in runtime, however if this would be an abstract method, it would not even compile.

topolyai5: Unfortunately, I need to implement in Azure, Yarn and Mock as well if I transform to an abstract method and I can return with null only and in that case, only an NPE will occur. It will happen if some try to implement in these 3 other providers.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/InstanceMetaDataService.java
```diff
@@ -100,13 +110,39 @@ public Stack saveInstanceAndGetUpdatedStack(Stack stack, List<CloudInstance> clo
...
     }
 
-    private Map<String, String> getSubnetAzPairs(String environmentCrn) {
+    String getAzFromDiskOrNullIfRepair(Stack stack, boolean repair) {
+        String availabilityZone = null;
+        if (repair) {
+            ResourceType resourceType = getSupportedReattachableDiskType(stack);
+            if (resourceType != null) {
+                Optional<CloudResource> reattachableDiskResource = resourceRetriever.findFirstByStatusAndTypeAndStack(CommonStatus.DETACHED,
+                        resourceType, stack.getId());
```
lajosrodek, line 119: Sorry, maybe I am missing something but I think it is possible to have multiple `DETACHED` disk resources for a given stack that span multiple AZs (even if they all belong to the same instance group). In this case we are always picking the first such resource and using its AZ for all the upscaled instances later on.

Moreover, what if the `DETACHED` disk resources belong to (terminated) instances of different instance groups? Those IGs could be also associated with totally different AZs. The `resourceRetriever` call here does not consider IG association, so I do not see if the choice made will be always appropriate.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/upgrade/recover/DatalakeRecoverySetupNewInstancesHandler.java
```diff
@@ -81,7 +81,7 @@ private void setupNewInstances(Stack stack) {
         for (InstanceGroup instanceGroup : instanceGroups) {
             List<CloudInstance> newInstance =
                     stackUpscaleService.buildNewInstances(stack, instanceGroup.getGroupName(), instanceGroup.getInitialNodeCount());
-            instanceMetaDataService.saveInstanceAndGetUpdatedStack(stack, newInstance, true, Collections.emptySet());
+            instanceMetaDataService.saveInstanceAndGetUpdatedStack(stack, newInstance, true, Collections.emptySet(), false);
```
lajosrodek, line 84: Just a sanity check: I am not familiar with the DL upgrade recovery process; does it always discard former volumes when creating replacement instances?

core/src/test/java/com/sequenceiq/cloudbreak/service/multiaz/MultiAzCalculatorServiceTest.java
```diff
@@ -77,6 +77,22 @@ public void testPrepareSubnetAzMap() {
         Assertions.assertEquals(result.get(cloudSubnetName(2)), cloudSubnetAz(2));
     }
 
+    @Test
+    public void testPrepareMultiAzWhenAvailabilityZoneNotNull() {
```
lajosrodek, line 81: Perhaps rename to `testPrepareSubnetAzMapWhenAvailabilityZoneNotNull`.

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/cluster/upgrade/recover/DatalakeRecoverySetupNewInstancesHandler.java
topolyai5: It is the original functionality. I will check the DL upgrade in the next phase

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/InstanceMetaDataService.java
topolyai5: Yes, you are right. It is a potential bug but I think, we need a refactor here because I don't know about the downscaled instance. Hopefully, we repair only one instance at once.

core/src/main/java/com/sequenceiq/cloudbreak/service/multiaz/MultiAzCalculatorService.java
```diff
@@ -44,21 +44,34 @@ public String determineRackId(String subnetId, String availabilityZone) {
...
             for (Map.Entry<String, CloudSubnet> entry : environment.getNetwork().getSubnetMetas().entrySet()) {
                 CloudSubnet value = entry.getValue();
-                if (!isNullOrEmpty(value.getName())) {
+                if (!isNullOrEmpty(value.getName()) && needToAddAZ(value, availabilityZone)) {
                     subnetAzPairs.put(value.getName(), value.getAvailabilityZone());
                 }
-                if (!isNullOrEmpty(value.getId())) {
+                if (!isNullOrEmpty(value.getId()) && needToAddAZ(value, availabilityZone)) {
                     subnetAzPairs.put(value.getId(), value.getAvailabilityZone());
                 }
```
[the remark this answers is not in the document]
topolyai5: I believe in the test coverage, so ok :D

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
```diff
@@ -180,9 +185,29 @@ public ExternalDatabaseStatus getDatabaseServerStatus(AuthenticatedContext authe
...
 
+    protected void diskReattachment(List<CloudResource> resources, Group scalingGroup, ResourceBuilderContext context) {
+        List<CloudResource> diskSets = resources.stream()
+                .filter(cloudResource -> scalingGroup.getName().equalsIgnoreCase(cloudResource.getGroup()))
+                .filter(cloudResource -> getDisResourceType().equals(cloudResource.getType()))
+                .filter(cloudResource -> StringUtils.isEmpty(cloudResource.getInstanceId()) || CommonStatus.DETACHED.equals(cloudResource.getStatus()))
+                .collect(Collectors.toList());
+        for (int i = 0; i < diskSets.size(); i++) {
+            CloudResource cloudResource = diskSets.get(i);
+            Optional.ofNullable(scalingGroup.getInstances().get(i))
```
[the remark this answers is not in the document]
topolyai5: We cannot guaranteed, we should refactor this code part

cloud-aws-common/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/common/resource/volume/AwsVolumeResourceBuilder.java
topolyai5: It worked until an error because we fetched at the beginning of the flow step and we worked with the "outdated" entity. I mean we only updated in the DB the status of the resource but in the memory, we used it with the old status

cloud-template/src/main/java/com/sequenceiq/cloudbreak/cloud/template/AbstractResourceConnector.java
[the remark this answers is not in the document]
lajosrodek: Okay. I vaguely remember some service (HBase perhaps) storing host-dependent data on HDFS (or local FS, cannot remember exactly), so that re-attaching the wrong volumeset to an instance will result in a conflict due to the instance FQDN and disk data mismatch.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/InstanceMetaDataService.java
lajosrodek: I would be more relaxed if we found a robust approach for this. Anyway, the fix for that (along with the related comment in `AbstractResourceConnector`) may come in a later PR as well.

lajosrodek (approved):
LGTM. A follow-up PR for the FQDN-aware volume reattach refactoring would be welcome nevertheless.

daszabo (approved):

PR #11245, opened 2021-08-16
CB-13796 Hard limit of 500 nodes for the cluster size for CB

horadla23:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/NodeCountLimitValidator.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/NodeCountLimitValidator.java
@@ -1,0 +1,39 @@
+package com.sequenceiq.cloudbreak.service;
+
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.base.InstanceGroupV4Base;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/NodeCountLimitValidator.java
```diff
@@ -0,0 +1,43 @@
...
+    public void validateScale(Long stackId, Integer scalingAdjustment) {
+        if (scalingAdjustment > 0) {
+            Integer currentNodeCount = instanceMetaDataService.countByStackId(stackId).getInstanceCount();
+            validateNodeCount(currentNodeCount + scalingAdjustment);
+        }
+    }
+
+    public void validateProvision(StackV4Request stackRequest) {
+        AtomicInteger nodeCounter = new AtomicInteger(0);
+        stackRequest.getInstanceGroups().stream().map(InstanceGroupV4Base::getNodeCount).forEach(igNodeCount -> nodeCounter.addAndGet(igNodeCount));
```
lturcsanyi, line 33: I think it could be simpler, for example:
`Integer nodeCount = stackRequest.getInstanceGroups().stream().mapToInt(InstanceGroupV4Base::getNodeCount).sum();`

lturcsanyi (approved):

PR #11246, opened 2021-08-16
CB-11919. Support multi region CP for databus clients

oleewere:
details:
- add new configuration (that should be passed in cloud-services-infra + dps-k8s) -> as thunderhad-sigmadbus is part of CSI, pass the s3 bucket name for dbus.
- use env region option for fluentd dbus plugin, if region is not used (or us-west-1 is passed), it will use the old dbus api, otherwise it will use the new /api/v1/dbus one
- update Uts

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/diagnostics/DiagnosticsFlowService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/diagnostics/DiagnosticsFlowService.java
@@ -56,6 +56,8 @@ public class DiagnosticsFlowService {
 
     private static final String AWS_METADATA_SERVER_V2_SUPPORT_VERSION = "0.4.9";
 
+    private static final String MULTI_CP_REGION_SUPPORT_VERSION = "0.4.12";
+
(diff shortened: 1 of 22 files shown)


lnardai:
@lturcsanyi Please work with @oleewere and get this merged, since it's needed for multi region control plane.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/diagnostics/DiagnosticsFlowService.java
```diff
@@ -56,6 +56,8 @@
 
     private static final String AWS_METADATA_SERVER_V2_SUPPORT_VERSION = "0.4.9";
 
+    private static final String MULTI_AZ_SUPPORT_VERSION = "0.4.12";
```
lturcsanyi, line 59: Is it MULTI_AZ_SUPPORT_VERSION or it should be MULTI_REGION_CP_VERSION?

lturcsanyi:
This is an old PR, have you tested it recently?

oleewere:
> This is an old PR, have you tested it recently?
it was done today, it seems OK

lturcsanyi (approved):

PR #11249, opened 2021-08-16
[CB-13049] collect metadata for GCP loadbalancers

yg-gd:
requires other changes from CB-13426 to establish GCP loadbalancing scheme, and Forwarding Rules type.

Gathers the just IP address that has been assigned to a loadbalancer. Other details around backing instances will be a part of the CLI required changes. 

test coverage similar to AWS, Azure loadbalancer metadata collectors

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
@@ -2,11 +2,12 @@ package com.sequenceiq.cloudbreak.cloud.gcp;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
@@ -14,11 +15,17 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
+import com.google.api.services.compute.Compute;
 import com.google.api.services.compute.model.AccessConfig;
+import com.google.api.services.compute.model.ForwardingRule;
+import com.google.api.services.compute.model.ForwardingRuleList;
 import com.google.api.services.compute.model.NetworkInterface;
 import com.sequenceiq.cloudbreak.cloud.MetadataCollector;
 import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
+import com.sequenceiq.cloudbreak.cloud.gcp.client.GcpComputeFactory;
+import com.sequenceiq.cloudbreak.cloud.gcp.loadbalancer.GcpLoadBalancerTypeConverter;
 import com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil;
+import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstanceMetaData;
 import com.sequenceiq.cloudbreak.cloud.model.CloudLoadBalancerMetadata;
@@ -40,6 +47,12 @@ public class GcpMetadataCollector implements MetadataCollector {
 
     @Inject
     private GcpStackUtil gcpStackUtil;
+
+    @Inject
+    private GcpComputeFactory gcpComputeFactory;
+
+    @Inject
+    private GcpLoadBalancerTypeConverter gcpLoadBalancerTypeConverter;
 
     @Override
     public List<CloudVmMetaDataStatus> collect(AuthenticatedContext authenticatedContext, List<CloudResource> resources, List<CloudInstance> vms,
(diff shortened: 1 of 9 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpLoadBalancerScheme.java
```diff
@@ -32,4 +33,20 @@ public static GcpLoadBalancerScheme getScheme(CloudLoadBalancer cloudLoadBalance
                 return INTERNAL;
         }
     }
+
+    public static GcpLoadBalancerScheme getScheme(String gcpType) {
```
hreeve-cloudera, line 37: Could me move the conversion logic to the `LoadBalancerTypeConverter` class? This is getting weirdly busy for an enum class.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpLoadBalancerScheme.java
```diff
@@ -32,4 +33,20 @@ public static GcpLoadBalancerScheme getScheme(CloudLoadBalancer cloudLoadBalance
...
+        switch (gcpType) {
+            case Constants.EXTERNAL_STRING:
```
hreeve-cloudera, line 39: Having a static class of constants in an enum class, which is already a class of constants, just feels odd. Could you not do something like:
```
if (INTERNAL.gcpType.equals(gcpType)) {
    return INTERNAL; 
} else etc...
```
That way you wouldn't need to reference the "INTERNAL" and "EXTERNAL" strings explicitly in two different places.

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/GCPLoadBalancerMetadataCollectorTest.java
```diff
@@ -0,0 +1,210 @@
...
+        when(forwardingRuleListResponse.getItems()).thenReturn(List.of(publicFowardingRule, publicFowardingRule2));
+
+        List<CloudLoadBalancerMetadata> result = underTest.collectLoadBalancer(authenticatedContext, List.of(LoadBalancerType.PUBLIC), resources);
+
+        assertEquals(2, result.size());
+        String actualIp = result.get(0).getIp();
+        String actualIp2 = result.get(1).getIp();
+        assertTrue((actualIp.equals(PUBLIC_IP2) && actualIp2.equals(PUBLIC_IP)) ||
+                (actualIp.equals(PUBLIC_IP) && actualIp2.equals(PUBLIC_IP2)));
+        assertEquals(PUBLIC_IP, actualIp);
```
hreeve-cloudera, line 110: In the first assert here, you have as an option `(actualIp.equals(PUBLIC_IP2) && actualIp2.equals(PUBLIC_IP))`. But the second assert says `assertEquals(PUBLIC_IP, actualIp)`. So actualIp will never actually equal PUBLIC_IP2, correct?

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollectorTest.java
```diff
@@ -59,25 +65,21 @@
 
     private static final List<CloudInstance> KNOWN_INSTANCES = Collections.emptyList();
 
-    private static final String PUBLIC_IP = "192.168.1.1";
-
-    private static final String PRIVATE_IP = "10.10.1.1";
-
     @InjectMocks
-    private GcpMetadataCollector underTest;
+    GcpMetadataCollector underTest;
```
hreeve-cloudera, line 69: Why did all these become not private?

hreeve-cloudera (changes requested):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
```diff
@@ -129,8 +139,35 @@ private CloudVmMetaDataStatus getCloudVmMetaDataStatus(CloudResource cloudResour
     @Override
     public List<CloudLoadBalancerMetadata> collectLoadBalancer(AuthenticatedContext ac, List<LoadBalancerType> loadBalancerTypes,
             List<CloudResource> resources) {
+        CloudCredential credential = ac.getCloudCredential();
+        Compute compute = gcpComputeFactory.buildCompute(credential);
+        String projectId = gcpStackUtil.getProjectId(credential);
+        String region = ac.getCloudContext().getLocation().getRegion().getRegionName();
+        List<CloudLoadBalancerMetadata> results = new ArrayList<>();
+        Set<String> names = resources.stream().map(CloudResource::getName)
```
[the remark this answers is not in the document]
yg-gd: Got it, for now filtering on just forwarding rules, will like need others for cli work

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpLoadBalancerScheme.java
yg-gd: Done, I think I've used a pattern where Enums can convert themselves, but cloudbreak seems to like more converters than i've seen before.

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollectorTest.java
yg-gd: was at one point a subclass of the new test, not needed now.

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/loadbalancer/GcpLoadBalancerScheme.java
hreeve-cloudera: They do love their converters. When I suggested this, I thought that `LoadBalancerTypeConverter` was a generic converter, and you could just add your logic to it. I now see the class I was thinking of is in the `cloud-aws-common` module, so not a shared resource. So that was my bad, but I do think having a separate converter does match the CB model better.

hreeve-cloudera (approved):

PR #11269, opened 2021-08-19
CB-10272 Move mock imagecatalog to mock-infrastructure 

lnardai:
Since we have mock-infrastructure needed for mock tests it makes sense to move imagecatalog mocks there.
thunderhead-mock was just a temporary location since we haven't had the time to introduce a new service back then.
Also solves the different ports between 8080 vs 10080.


horadla23 (approved):
LGTM

PR #11270, opened 2021-08-19
CB-13836 - Upgrade endpoint should contains parcel info

pdarvasi:
Parcel info is returned on the Upgrade Endpoint:
image

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ComponentVersionProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ComponentVersionProvider.java
@@ -1,14 +1,24 @@
 package com.sequenceiq.cloudbreak.service.upgrade;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ComponentVersionProvider.java
```diff
@@ -1,22 +1,53 @@
...
-                osPatchLevel);
+                osPatchLevel,
+                getParcelInfoResponse(packageVersions));
+    }
+
+    private List<ParcelInfoResponse> getParcelInfoResponse(Map<String, String> packageVersions) {
+        List<ParcelInfoResponse> parcelInfoResponses = packageVersions.entrySet()
+                .stream()
+                .filter(this::imagePackageVersionExists)
+                .map(entry -> new ParcelInfoResponse(
```
tiborpopovics, line 38: minor: I think it's cleaner to extract the ParcelInfoResponse creation into a private method.

tiborpopovics (approved):

PR #11272, opened 2021-08-19
CB-13867 Registering 443 as ccmv2config into CP

horadla23:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
@@ -189,11 +189,13 @@ public class ClusterProxyService {
     }
 
     private List<CcmV2Config> ccmV2Configs(Stack stack) {
-        return stack.getNotTerminatedGatewayInstanceMetadata().stream().map(instanceMetaData -> new CcmV2Config(
-                instanceMetaData.getPublicIpWrapper(),
-                ServiceFamilies.GATEWAY.getDefaultPort(),
-                String.format(CCMV2_BACKEND_ID_FORMAT, stack.getCcmV2AgentCrn(), instanceMetaData.getInstanceId()),
-                CLOUDERA_MANAGER_SERVICE_NAME)).collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


bergerdenes (approved):

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
```diff
@@ -189,11 +189,13 @@ private String getAccountId(Stack stack) {
...
-                ServiceFamilies.GATEWAY.getDefaultPort(),
-                String.format(CCMV2_BACKEND_ID_FORMAT, stack.getCcmV2AgentCrn(), instanceMetaData.getInstanceId()),
-                CLOUDERA_MANAGER_SERVICE_NAME)).collect(Collectors.toList());
+        return stack.getNotTerminatedGatewayInstanceMetadata().stream().map(instanceMetaData -> {
+            CcmV2Config gatewayConfig = new CcmV2Config(instanceMetaData.getPublicIpWrapper(), ServiceFamilies.GATEWAY.getDefaultPort(),
+                    String.format(CCMV2_BACKEND_ID_FORMAT, stack.getCcmV2AgentCrn(), instanceMetaData.getInstanceId()), CLOUDERA_MANAGER_SERVICE_NAME);
+            CcmV2Config knoxConfig = new CcmV2Config(instanceMetaData.getPublicIpWrapper(), ServiceFamilies.KNOX.getDefaultPort(),
+                    String.format(CCMV2_BACKEND_ID_FORMAT, stack.getCcmV2AgentCrn(), instanceMetaData.getInstanceId()), CLOUDERA_MANAGER_SERVICE_NAME);
+            return List.of(gatewayConfig, knoxConfig);
+        }).flatMap(List::stream).collect(Collectors.toList());
```
zabhi, line 198: For this to work in CCM v2, you would need another agent to forward calls to port 443 which is not happening. The fix should have been to revert the port number to 9443.

Did it work during your dev tests?

horadla23: since CB does not use 443 to communicate with CM, I couldn't test this scenario, so my thought was that we only need to add 443 into ccmv2configs, based on the ticket and comments. So will this not work at all?

PR #11275, opened 2021-08-20
ENGESC-9803. Add preflight check for VM level checks before diagnosti…

oleewere:
…cs collection.

details:
- create a new vm preflight check step for freeipa/core diagnostics
- the new step uploads a preflight-check.sh script to the master node and executes it
- on failure or if the result is unexpected it will interrupt the diagnostics process
- any error could be fine here as that also interrupts the diagnostics
- check the non rotated or relatively large files (salt logs are not rotated, syslog and krb5kdc log are, but still can grow large as the file limit is large for them and rotation happens daily)
- the new script only running on the master nodes - and sends commands to the minions
- the script does not case with the unresponsive minions, as collection won't happen on those nodes


oleewere:
some methods could be duplicated, it will need some refactoring, but that's i could implement for hotfix

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/scripts/preflight-check.sh
```diff
@@ -0,0 +1,338 @@
...
+  done
+
+  # cleanup
+  rm -rf /srv/salt/distribution
+
+  if [[ "$error_msg" != "" ]];then
+    local full_error_msg=$(cat << EOF
+PreFlight diagnostics check FAILED
+Diagnostics log collection has been interrupted in order to avoid memory pressure on VM nodes that could shut down the machines , as some files are too large to process.
+Please upgrade cdp-telemetry binary to workaround the issue (at least to version 0.4.13) or filter out affected hosts/labals (in case of CDP cli was used for the bundle collection).
```
aakanser, line 227: using host filters is also an option on the UDX UI now

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/scripts/preflight-check.sh
```diff
@@ -0,0 +1,338 @@
...
+The following logs/nodes are affected: $error_msg
+
+You can upgrade the cdp-telemetry binary by using --update-package option by CDP cli (for environment/datalake/datahub collect-XX-diagnostics commands). But note that requires network access for https://cloudera-service-delivery-cache.s3.amazonaws.com (on the VM nodes).
+Alternatively, the binary can be updated manually by the following steps (e.g. for version $EXPECTED_MIN_VERSION):
+1. Download rpm from https://cloudera-service-delivery-cache.s3.amazonaws.com/telemetry/cdp-telemetry/cdp_telemetry-$EXPECTED_MIN_VERSION.x86_64.rpm
+2. Scp the rpm file to a salt-master node. (e.g. into /home/cloudbreak folder)
+3. Execute the following commands as root on the salt-master node:
```
aakanser, line 234: when you say `salt-master` node, do you mean the master node of datalake, freeipa etc.?

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/scripts/preflight-check.sh
```diff
@@ -0,0 +1,338 @@
+#!/bin/sh
+
+EXPECTED_MIN_VERSION="0.4.13"
+# ~300MB
+MAX_FILE_SIZE=314572800
```
aakanser, line 5: can be MAX_FILE_SIZE_BM

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/scripts/preflight-check.sh
oleewere: mostly, but not always true, like can be called gateway as well

orchestrator-salt/src/main/resources/salt-common/salt/filecollector/scripts/preflight-check.sh
oleewere: actually that's in bytes, so will keep as is

aakanser: then it can be MAX_FILE_BYTES so that there is no confusion about units

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/EntitlementService.java
```diff
@@ -280,6 +281,10 @@ public boolean isAWSDiskEncryptionWithCMKEnabled(String accountId) {
         return isEntitlementRegistered(accountId, CDP_CB_AWS_DISK_ENCRYPTION_WITH_CMK);
     }
 
+    public boolean isDiagnosticsServiceUnderMaintenance(String accountId) {
+        return isEntitlementRegistered(accountId, CDP_DIAGNOSTICS_SERVICE_MAINTENANCE);
```
aakanser, line 285: I think making this a positive flag fits better under entitlement structure. Straightforward: `Entitlement.CDP_VM_DIAGNOSTICS`

If it's not granted, than it's under maintenance, or there is some other issue, etc.

oleewere: ok that make sense, also would not need UMS deploy to disable the API

oleewere:
removed entitlement, moved that to an another PR: https://github.com/hortonworks/cloudbreak/pull/11277
+ a made this draft, could be that change can be ok for 2.47 , if diagnostics will be disabled for longer time (~1 week)

oleewere:
change will go to 2.47.0, see: https://github.com/hortonworks/cloudbreak/pull/11284

PR #11276, opened 2021-08-20
[CB-13876] issue with creating forwarding rule for shared vpcs

yg-gd:
needs to be validated in mowdev to networking issues on local set up
unit tests expanded to cover multiple vpc settings


hreeve-cloudera (approved):

biharitomi (approved):
LGTM

PR #11278, opened 2021-08-20
CB-13859. Enable log rotation for salt logs.

oleewere:

### DIFF
--- a/orchestrator-salt/src/main/resources/salt-common/salt/logrotate/conf/salt
+++ b/orchestrator-salt/src/main/resources/salt-common/salt/logrotate/conf/salt
@@ -1,0 +1,15 @@
+/var/log/salt/master
+/var/log/salt/api
+/var/log/salt/key
+/var/log/salt/minion
+{
+    missingok
+    notifempty
+    copytruncate
+    daily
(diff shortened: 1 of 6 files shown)


daszabo (approved):
LGTM

orchestrator-salt/src/main/resources/salt-common/salt/logrotate/conf/salt
```diff
@@ -0,0 +1,15 @@
...
+/var/log/salt/api
+/var/log/salt/key
+/var/log/salt/minion
+{
+    missingok
+    notifempty
+    copytruncate
+    daily
+    size 50M
+    rotate 14
```
sidseth, line 11: What does this set of configs translate to in terms of the total salt log size that will be available? 14*50MB?

oleewere: yes. it's true it can be reached more quickly (few days) for the salt master log for larger DH clusters so created PR for that: https://github.com/hortonworks/cloudbreak/pull/11348 (currently this on master anyway)

PR #11280, opened 2021-08-20
CB-13805: Make necessary changes to make the DL resizing transparent to experiences and UI

kkalvagadda1:
1. Introduced new field in SDX cluster to store display name.
2. In the case of resizing, the display name of the resized data lake is the same as the original one.
3. Same is passed to the stack. Display name of the stack created for resized data lake has the name of the original data lake.
4. Once the resizing is complete, the CRN of the new data lake is replaced with the CRN of the original data lake. Without which re-wiring with existing experiences becomes impossible.

There is a follow-up jira CB-13515. It needs a lot of testing. That is why I separated it.


datalake/src/main/java/com/sequenceiq/datalake/flow/delete/SdxDeleteActions.java
```diff
@@ -147,6 +152,13 @@ protected void doExecute(SdxContext context, RdsDeletionSuccessEvent payload, Ma
                 if (sdxCluster != null) {
                     metricService.incrementMetricCounter(MetricType.SDX_DELETION_FINISHED, sdxCluster);
                 }
+                // If the sdx that is deleted is detached, we need to replace CRN with the CRN of the original SDX
```
lnardai, line 155: Wouldn't it make more sense to do the swapping during the last step of the creation of the replacement? 
There could be a timeframe where someone can issue operations on the newly created but not swapped DL.
I'm asking because what if the old DL get's stuck for some reason, like someone does a repair in the background, or any other operations happens?

horadla23: why do we need to replace the CRN with original one? or other question is that why does CRN change at all during resize? second question may was answered in the past but I cannot remember

kkalvagadda1: Laszlo,
Ideally yes. If I do it there will be having two non-deleted data lakes using the CRN. As part of CB-13515, we will be updating the name at the same time. That is why picked the point where the old data lake is deleted.


Adam,
WIth resizing we will be creating a parallel SDX before we delete the older one. This makes recovery simple.
As we creating a new SDX, the CRN of the newly created SDX will be different. 
All the experiences use the CRN of the data lake to pull the updated configuration. It is critical to maintain the same CRN after resizing.


https://docs.google.com/document/d/1KbPCF6Sem5YkWIbAk5GkbG2vM_cqCzhFRGJfR10aNQg/edit?usp=sharing

datalake-api/src/main/java/com/sequenceiq/sdx/api/model/SdxClusterResponse.java
```diff
@@ -15,6 +15,8 @@
 
     private String name;
 
+    private String displayName;
```
lnardai, line 18: If we add this  we can't remove it because it will be a breaking change.

PR #11284, opened 2021-08-22
ENGESC-9803. Add on the fly telemetry upgrade + vm level preflight ch…   …eck before doing diagnostics collections

oleewere:
details:
- add 2 new flow steps for both core and freeipa diagnostics: telemetry upgrade & vm level preflight check
- create 3 scripts that will be uploaded during these steps: cdp_telemetry_deployer.sh + preflight_check.sh + filecollector_minion_check.py
- cdp_telemetry_deployer.sh: it will try to upgrade cdp-telemetry from the master node (primary) - failure is fine here (also checks proxy settings + network availability)
- preflight_check.sh: that will run on the primary master node, it will execute filecollector_minion_check.py script on every minion, it will print different outputs based on the minion results
- if preflight fails or we got an unexpected response, that will interrupt the collection
- preflight checks available memory + actuaal cdp-telemetry version + it also reconfigure the filecollector (if cdp-telemetry has older version) -> in order to avoid memory pressure (of course it will slow down the process)
- preflight check can reconfigure filecollector configs in a way that will remove ceertain (too large logs) to process. it won't skip salt or syslogs. if those are too large and cdp-telemetry version is too low, the check will fail


daszabo (approved):
LGTM

akanto:
Make sure that we have appropriate test coverage.

PR #11296, opened 2021-08-23
CB-13178 - Gather Azure medium duty datalake load balancer configs and return via cdp cli

hreeve-cloudera:
The Azure companion change to CB-13176. Gathers the Azure load balancer name and the availability set
name associated with each target group. That information is persisted to the database and returned
as part of the describe datalake output.

Tested with unit tests.

### DIFF
--- a/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AvailabilitySetNameService.java
+++ b/cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AvailabilitySetNameService.java
@@ -1,0 +1,11 @@
+package com.sequenceiq.cloudbreak.cloud.model.instance;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public class AvailabilitySetNameService {
+
+    public String generateName(String prefix, String instanceGroupName) {
+        return String.format("%s-%s-as", prefix, instanceGroupName);
+    }
+}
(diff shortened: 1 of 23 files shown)


cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AvailabilitySetNameService.java
```diff
@@ -0,0 +1,11 @@
+package com.sequenceiq.cloudbreak.cloud.model.instance;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public class AvailabilitySetNameService {
```
sxxgrc, line 6: Is this going to be used for other cloud providers as well? Just wondering why you didn't put it with the other Azure stuff.

sxxgrc (approved):
Looks good to me! Thanks for the changes.

cloud-api/src/main/java/com/sequenceiq/cloudbreak/cloud/model/instance/AvailabilitySetNameService.java
hreeve-cloudera: This logic used to be hardcoded in `LoadBalancerConfigService`. I needed to be able to generate the AS name in `AzureLoadBalancerMetadataCollector` as well, but due to module dependencies I couldn't import `LoadBalancerConfigService` into that class. So I moved it here so that both of those classes could import it and use the logic.

sxxgrc: Ah okay that makes sense, thanks for letting me know.

PR #11301, opened 2021-08-24
CB-13880: Move DataHub OS upgrade behind CDP_OS_UPGRADE_DATAHUB entitlement.

tiborpopovics:


mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
```diff
@@ -293,6 +294,9 @@
     @Value("${auth.mock.datahub.runtime.upgrade.enable}")
     private boolean datahubRuntimeUpgradeEnabled;
 
+    @Value("${auth.mock.datahub.os.upgrade.enable}")
```
keyki, line 297: this should be introduced in the app.yml with a true value

keyki (changes requested):
When the UI initiates the OS upgrade process it only sends the image ID thus it's not recognized as an OS upgrade and will pass the validation even if you don't have the entitlement. 

The second problem is when you don't have the `CDP_RUNTIME_UPGRADE_DATAHUB` and initiate the OS upgrde from the CLi with --lock-components and incorrect message will be returned as the status reason is overwritten saying there are no maintenance candidates available. This is probably cause by my recent change for maintenanc eupgrade.

PR #11303, opened 2021-08-25
CB-13649 Support of multi-AZ on native AWS variant for Datalakes. The…

topolyai5:
… medium_duty_ha.json is used and it decorates with the multi-AZ values. Multi-AZ can enable by entitlement. Entitlment added as well.
the subnet of DL's instance will be selected from the environment's public subnets


biharitomi (approved):
LGTM

PR #11307, opened 2021-08-26
CB-13731 Datalake - CB core flow to flow interaction is not idempotent

foldik:
Ability to mark certain flow's as idempotent, so when a flow or flow chain is running and someone would try to start the exact some process the flow engine returns the original flow / flow chain id.

# Improvement:

Since the correctness of the flow module largely depends on the interaction of different classes such as Flow2Handler, FlowChainHandler, FlowChains, FlowRegister, FlowStateCache a new integration test class called `FlowIntegrationTest` was created.

# Bugfixes:

During testing various bugs came out and was fixed. See details below. Now they are covered by integration tests.

## Parts of flow chain can be lost

If a new flow chain's first event is to create a new nested flow chain then the parent flow chain won't be saved until the child flow chain doesn't finish. If the application stops and restarts, parts of the flow chain can be lost since those were not saved to the database (they were present only in memory).

## Memory leak

If a new flow is part of a flow chain but there is already a running flow on the resource then the flow chain won't be cleaned up from inmemory maps.

## Retried but succeeded flow can be retried

If a flow failed but after a retry it succeeded the flow module still allowed it too be retried.


foldik:
Manual tested:
- create datalake / datahub,
- repair datalake (just master, master and idbroker),
- upscale/downscale datahub's compute node,
- repair datahub's compute node.

flow/src/test/java/com/sequenceiq/flow/integration/sleep/event/SleepStartEvent.java
```diff
@@ -0,0 +1,92 @@
...
+import com.sequenceiq.cloudbreak.common.event.AcceptResult;
+import com.sequenceiq.cloudbreak.common.event.IdempotentEvent;
+
+import reactor.rx.Promise;
+
+public class SleepStartEvent implements IdempotentEvent<SleepStartEvent> {
+
+    public static final LocalDateTime NEVER_FAIL = LocalDateTime.MIN;
+
+    public static final LocalDateTime ALAWAYS_FAIL = LocalDateTime.MAX;
```
horadla23, line 16: typo

horadla23 (approved):
LGTM

sodre90 (approved):

PR #11327, opened 2021-09-01
CB-13983 Do not set nifi.ui.knox.url property automatically. Removed …

vbrodetskyi:
…all properties that were set for NiFi Knox automatically, but didn't remove a class, because in future it should be useful.


lajosrodek:
@vbrodetskyi 
In general, I would rather suggest dropping the whole class. It makes no sense to keep an almost empty ConfigProvider, it can be re-added later anytime if deemed useful.

One q: Is this change compatible with all currently supported runtime versions? :smile: If not, the legacy logic of the class has to be kept and wrapped in a conditional block. (Or simply `isConfigurationNeeded()` be changed by adding another condition for the runtime version check.)

vbrodetskyi:
@lajosrodek great thanks for your review, looks like i really should keep this class and add condition for correct support of older versions.

vbrodetskyi:
@lajosrodek @doktoric could you please review current patch?

doktoric (approved):

PR #11329, opened 2021-09-02
CB-13878 Replace swagger-diff with swagger_spec_compatibility for better compatibility checking

doktoric:


integration-test/scripts/swagger-check.sh
```diff
@@ -51,160 +86,31 @@ IFS=,
...
+  for incompat in "${INCOMPATIBLE_CHANGES[@]}"; do
+    echo "- $incompat"
+  done
+  if (( NUM_INCOMPATIBLE_CHANGES > 255 )); then
+    NUM_INCOMPATIBLE_CHANGES=255
+  fi
+fi
 
-exit $swaggerdiffresult && $swaggervalidationresult
+exit $NUM_INCOMPATIBLE_CHANGES
```
lajosrodek, line 116: Out of curiosity, why the need for the funny exit code? Code 0 means success, codes 1...127 are regular errors, while codes C 128...255 typically denote that the process exited due to catching / receiving signal number (C-128). (Only a subset of the latter range is valid as per the permitted signal codes; see `man 7 signal`.) Quoting the `bash` manual:
```
       For the shell's purposes, a command which exits with a zero exit status has succeeded.  An
       exit  status of zero indicates success.  A non-zero exit status indicates failure.  When a
       command terminates on a fatal signal N, bash uses the value of 128+N as the exit status.

       If a command is not found, the child process created to execute it  returns  a  status  of
       127.  If a command is found but is not executable, the return status is 126.
```

doktoric: I think the important thing is that we should not return 0. This is coming from the thunderhead codebase
https://github.com/hortonworks/cloudbreak/pull/11329/files#diff-ffa4a6b378a979862d274632d729be771ea467e80c00fe5f19fdb4048f6c567fR111

lajosrodek: Okay, but sticking to just two values (0 and 1, say) would suffice. I suggest replacing lines 111..113
```
  if (( NUM_INCOMPATIBLE_CHANGES > 255 )); then
    NUM_INCOMPATIBLE_CHANGES=255
  fi
```
with
```
  NUM_INCOMPATIBLE_CHANGES=1
```

cegganesh84:
+1. This is great @doktoric, thanks for doing this. I was afraid of RES-E003 being a hindrance for adoption but it is nicely excluded from the check.

integration-test/scripts/swagger-check.sh
doktoric: done

lajosrodek (approved):
Thanks for the fix! LGTM

PR #11331, opened 2021-09-06
CB-13992 Create the environment with network by default. We need to e…

topolyai5:
…nforce the client to add the network with a constraint on the environment request. Mock test added

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/network/EnvironmentNetworkValidator.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/network/EnvironmentNetworkValidator.java
@@ -62,4 +62,11 @@ public interface EnvironmentNetworkValidator {
         return diff;
     }
 
+    default void checkNullable(CloudPlatform cloudPlatform, NetworkDto networkDto, ValidationResultBuilder resultBuilder) {
+        boolean canBeNull = cloudPlatform == CloudPlatform.YARN || networkDto != null;
+        if (!canBeNull) {
+            resultBuilder.error("Environment network cannot be null");
+        }
+    }
+
 }
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/NetworkCreationValidator.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.sequenceiq.cloudbreak.common.mappable.CloudPlatform;
 import com.sequenceiq.cloudbreak.validation.ValidationResult.ValidationResultBuilder;
 import com.sequenceiq.environment.environment.domain.Environment;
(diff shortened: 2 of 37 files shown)


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentRequest.java
```diff
@@ -58,6 +58,7 @@
 
     @Valid
     @ApiModelProperty(EnvironmentModelDescription.NETWORK)
+    @NotNull(message = "Environment network cannot be null")
     private EnvironmentNetworkRequest network;
```
doktoric, line 62: I think this will broke qaas tests in case of ycloud.

core/src/main/java/com/sequenceiq/cloudbreak/service/environment/credential/CredentialClientService.java
```diff
@@ -60,4 +62,16 @@ public Credential getByEnvironmentCrn(String envCrn) {
             throw new CloudbreakServiceException(message, e);
         }
     }
+
+    @Override
+    public String getCloudPlatformByCredential(String credentialName, String accountId, CredentialType type) {
+        try {
+            CredentialResponse credentialResponse = credentialEndpoint.getByName(credentialName);
+            return credentialResponse.getCloudPlatform();
+        } catch (WebApplicationException | ProcessingException | IllegalStateException e) {
```
lacikaaa, line 71: we usually handle `WebApplicationException` separately to extract the real reason

datalake/src/main/java/com/sequenceiq/datalake/service/CredentialService.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final Logger LOGGER = getLogger(CredentialService.class);
+
+    @Inject
+    private CredentialEndpoint credentialEndpoint;
+
+    @Inject
+    private SecretService secretService;
+
+    @Inject
+    private WebApplicationExceptionMessageExtractor webApplicationExceptionMessageExtractor;
```
lacikaaa, line 32: these are not used

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/validator/EnvironmentNetwork.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.Target;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Documented
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = EnvironmentNetworkValidator.class)
+public @interface EnvironmentNetwork {
```
lacikaaa, line 16: I think the name is not meaningful enough. something like `ValidEnvironmentNetwork` is indicating more what is the purpose of this class

freeipa/src/main/java/com/sequenceiq/freeipa/service/CredentialService.java
```diff
@@ -44,4 +53,16 @@ public Credential getCredentialByEnvCrn(String envCrn) {
         String attributes = secretService.getByResponse(secretResponse);
         return new Credential(credentialResponse.getCloudPlatform(), credentialResponse.getName(), attributes, credentialResponse.getCrn());
     }
+
+    @Override
+    public String getCloudPlatformByCredential(String credentialName, String accountId, CredentialType type) {
```
lacikaaa, line 58: this method and the implementation seems to be duplicated several times. I think it should be refactored, maybe to a default implementation in the interface or similar

topolyai5: I cannot create a common implementation, because we need implementation in every microservice. If I create a default implementation in the env-api, the env service won't start (and every other microservice where a custom impl will be created).

But I think I will do it another way. The annotation is nicer but more complicated because the env network validator should retrieve the credential because of the cloud platform. And If I looking for a common way of implementation, it will increase the complexity of the whole task exponentially. `@NotNull` annotation could not be used, because the yarn provider is accepted without a network and the qaas test will fail.

environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/validator/EnvironmentNetwork.java
topolyai5: I have removed it

datalake/src/main/java/com/sequenceiq/datalake/service/CredentialService.java
topolyai5: i have removed it

core/src/main/java/com/sequenceiq/cloudbreak/service/environment/credential/CredentialClientService.java
topolyai5: i have removed it

biharitomi (approved):
LGTM

lnardai (approved):

doktoric (approved):

PR #11332, opened 2021-09-06
CB-12707 - Put original repaired node back to DELETED_ON_PROVIDER_SIDE if repair fails if stack upscale failed

sodre90:
Node can disappear at repair. It can happen if stack upscale fails at repair. In this scenario the old node is deleted, but the new node is in REQUESTED status. But requested nodes without instance id are cleaned up in many ways. For example if upscale fails, the upscale fail handling logic deletes these kind of nodes. It results the repaired node will disappear.

My solution: if a stack upscale fails, then we take back the original deleted node to DELETED_ON_PROVIDER_SIDE.
In this case the retry will not work. Reason is because retry will retry the upscale process and it would uspcale a new node next to the original node which is in DELETED_ON_PROVIDER_SIDE status. It is handled, retry (upscale) will fail if there is an another node with the same FQDN at upscale (e.g. the original node in DELETED_ON_PROVIDER_SIDE status).

Other solution would be if we just don't remove the upscaled node if it fails, but it is very hard to figure out which status we use in this case. As I experienced UI can repair SERVICES_UNHEALTHY nodes only and it will cause much trouble if user don't use retry but the user initiates a new repair for this failed upscaled node. Another thing is we ha cleanup codes for instances without instance ids as I mentioned before.

First I tried to implement the second solution but it was very hard to cover every edge cases and it leads to very complicated codes. The solution I choose looked more cleaner and easier to understand.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
@@ -341,9 +341,11 @@ public class AzureResourceConnector extends AbstractResourceConnector {
         Collection<CloudResource> result = new ArrayList<>();
         for (CloudInstance instance : instances) {
             String instanceId = instance.getInstanceId();
-            for (CloudResource resource : resources) {
-                if (instanceId.equalsIgnoreCase(resource.getName()) || instanceId.equalsIgnoreCase(resource.getInstanceId())) {
-                    result.add(resource);
+            if (instanceId != null) {
+                for (CloudResource resource : resources) {
+                    if (instanceId.equalsIgnoreCase(resource.getName()) || instanceId.equalsIgnoreCase(resource.getInstanceId())) {
+                        result.add(resource);
+                    }
                 }
             }
         }
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/MockSdxRepairTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/MockSdxRepairTests.java
@@ -6,6 +6,7 @@ import static com.sequenceiq.it.cloudbreak.context.RunningParameter.key;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.function.Consumer;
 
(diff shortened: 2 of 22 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/mock/MockSdxRepairTests.java
```diff
@@ -75,6 +77,59 @@ public void repairTerminatedMaster(MockedTestContext testContext) {
                 SdxClusterStatusResponse.CLUSTER_AMBIGUOUS);
     }
 
+    @Test(dataProvider = TEST_CONTEXT_WITH_MOCK)
```
foldik, line 80: Good stuff!

foldik:
Seems good but can you please write down what cases did you try out? If affects datahub repair as well right? Did you test downscale / upscale / repair operations with various setups?

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
```diff
@@ -341,9 +341,11 @@ private void checkTemplateDeployment(List<CloudResourceStatus> result, AzureClie
         Collection<CloudResource> result = new ArrayList<>();
         for (CloudInstance instance : instances) {
             String instanceId = instance.getInstanceId();
-            for (CloudResource resource : resources) {
-                if (instanceId.equalsIgnoreCase(resource.getName()) || instanceId.equalsIgnoreCase(resource.getInstanceId())) {
-                    result.add(resource);
+            if (instanceId != null) {
```
foldik, line 344: Do other cloud platforms handle it differently?

sodre90:
> Seems good but can you please write down what cases did you try out? If affects datahub repair as well right? Did you test downscale / upscale / repair operations with various setups?

You are right, I did testing only with DLs. I will do some tests with DH also.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureResourceConnector.java
sodre90: yes, they do it in different ways

lturcsanyi:
Please send notification event about node status restore after a failed upscale.

sodre90:
> Please send notification event about node status restore after a failed upscale.

done

lnardai (approved):

PR #11337, opened 2021-09-06
CB-13991 Not allowed to scale streaming 7.2.12 clusters in mow-dev. A…

doktoric:
…dding the scale limit to recommendation endpoint


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateProcessor.java
```diff
@@ -298,10 +300,16 @@ public AutoscaleRecommendation recommendAutoscale() {
...
-                        Entitlement entitledFor = ((EntitledForServiceScale) enumValue.get()).getEntitledFor();
+                        EntitledForServiceScale entitledForServiceScale = (EntitledForServiceScale) enumValue.get();
+                        Entitlement entitledFor = entitledForServiceScale.getEntitledFor();
                         if (!entitlements.contains(entitledFor.name())) {
-                            foundBlackListItem = true;
-                            break;
+                            if (isVersionEnablesScaling(version, entitledForServiceScale)) {
+                                foundBlackListItem = false;
+                                break;
+                            } else {
```
lajosrodek, line 309: 1. `foundBlackListItem` is always `false` at this point. Any time it is set to `true` the loop is exited by a subsequent `break`.
2. I wonder if doing a `break` here is indeed the right thing, instead of either `continue` or simply not having any specific action (which is basically equivalent with `continue` as there are no other statements after the enclosing `if`). What if some `service1` is found okay due to `!entitlements.contains(entitledFor.name()) && isVersionEnablesScaling(version, entitledForServiceScale)` (and hence executing the statements inside this `if`), but some other `service2` occurring later (with respect to iteration order) in the `hostGroupServices.getValue()` collection would be blacklisted? With the current logic, the check for validity of `service2` is never made due to the `break`.

mock-thunderhead/src/main/java/com/sequenceiq/thunderhead/grpc/service/auth/MockUserManagementService.java
```diff
@@ -817,7 +817,6 @@ public void getAccount(GetAccountRequest request, StreamObserver<GetAccountRespo
                                 .addEntitlements(createEntitlement(LOCAL_DEV))
                                 .addEntitlements(createEntitlement(DATAHUB_FLOW_SCALING))
                                 .addEntitlements(createEntitlement(CDP_SHOW_CLI))
-                                .addEntitlements(createEntitlement(DATAHUB_STREAMING_SCALING))
```
lajosrodek, line 820: Is this intentional, or perhaps made only for sake of your local testing?

lajosrodek (commented):
One general comment: The JIRA ticket appearing in the PR title, commit summary and branch name are all wrong. In particular, I think this PR is supposed to be for CB-13964. :smile:

PR #11342, opened 2021-09-07
CB-13731 Mark datalake to cloudbreak flow trigger events as idempotent.

foldik:

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/event/ClusterCertificatesRotationTriggerEvent.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/event/ClusterCertificatesRotationTriggerEvent.java
@@ -1,5 +1,7 @@
 package com.sequenceiq.cloudbreak.core.flow2.event;
 
+import java.util.Objects;
+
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/event/ClusterCertificatesRotationTriggerEvent.java
```diff
@@ -23,4 +25,13 @@ public ClusterCertificatesRotationTriggerEvent(String event, Long resourceId, Pr
     public CertificateRotationType getCertificateRotationType() {
         return certificateRotationType;
     }
+
+    @Override
+    public boolean equalsEvent(StackEvent other) {
+        if (!isClassAndEqualsEvent(ClusterCertificatesRotationTriggerEvent.class, other)) {
+            return false;
+        }
```
horadla23, line 33: this part is very similar in every `equalsEvent` method, maybe it is worth to extract it in `StackEvent`

horadla23: or event specific comparisons should be extracted into different overridden method

horadla23 (approved):

PR #11351, opened 2021-09-08
CB-14023 Create weekly job for high node count scale e2e test

horadla23:
- cleanup scale test parameters class
- remove unused methods from AbstractIntegrationTest
- cleanup DistroxScaleTest
- initiate env and datalake creation at the same time in DistroxScaleTest
- create freeipa along with environment creation, not separately

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
@@ -242,6 +242,14 @@ public abstract class AbstractIntegrationTest extends AbstractMinimalTest {
                 .validate();
     }
 
+    protected void createEnvironmentWithFreeIpaAndDatalake(TestContext testContext) {
+        initiateEnvironmentCreation(testContext);
+        initiateDatalakeCreation(testContext);
+        waitForEnvironmentCreation(testContext);
+        waitForUserSync(testContext);
+        waitForDatalakeCreation(testContext);
+    }
+
     protected void createDefaultUser(TestContext testContext) {
         testContext.as();
     }
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
@@ -11,10 +11,7 @@ import org.testng.ITestResult;
 import org.testng.annotations.BeforeMethod;
 
 import com.sequenceiq.common.api.type.Tunnel;
-import com.sequenceiq.environment.api.v1.environment.model.EnvironmentNetworkMockParams;
-import com.sequenceiq.environment.api.v1.environment.model.request.EnvironmentNetworkRequest;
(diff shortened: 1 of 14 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
```diff
@@ -242,56 +228,24 @@ protected void createEnvironmentWithFreeIpa(TestContext testContext) {
                     .withNetwork()
                     .withTelemetry("telemetry")
                     .withTunnel(Tunnel.CLUSTER_PROXY)
-                    .withCreateFreeIpa(Boolean.FALSE)
+                    .withCreateFreeIpa(Boolean.TRUE)
+                    .withFreeIpaImage(commonCloudProperties().getImageValidation().getFreeIpaImageCatalog(),
+                            commonCloudProperties().getImageValidation().getFreeIpaImageUuid())
```
topolyai5, line 233: Are these lines always setup together? If yes, I think would be better as a dedicated method, like `enableFreeIpaCreation` or stuff like this

topolyai5: And I would hold the original `.withCreateFreeIpa(Boolean.TRUE)` method as well

horadla23: the original logic is that we are creating freeipa always, but after environment creation was success without actual freeipa, which does not make any sense from product point of view

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/distrox/DistroXScaleTest.java
```diff
@@ -1,72 +1,54 @@
...
         createDefaultUser(testContext);
         createDefaultCredential(testContext);
         initializeDefaultBlueprints(testContext);
-        createEnvironmentWithFreeIpa(testContext);
-        createDatalake(testContext);
+        initiateEnvironmentCreation(testContext);
+        initiateDatalakeCreation(testContext);
+        waitForEnvironmentCreation(testContext);
+        waitForUserSync(testContext);
+        waitForDatalakeCreation(testContext);
```
topolyai5, line 31: It would be better in AbstractE2ETest as a standalone method. Maybe it could be the default process to create the env and sdx together (I think it is much quicker with external db)

horadla23: yes, that would be the plan, but first I would enable it only for this test first, to not break anything in worse case

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/AbstractIntegrationTest.java
horadla23: discussed internally, we can review/remove/refactor image settings of freeipa within different task

horadla23:
retest this please

lnardai (approved):

horadla23:
at this point I would wait until 2.48 branch is created

PR #11360, opened 2021-09-09
CB-14055: Failure on database backup with failure "File exists"

kkalvagadda1:
Made sure the existing file is replaced instead of throwing File exists error.


brycederriso (approved):
This change does what we expect it to.

I'm concerned that we may not want to overwrite existing backups though. Approving since I think this is causing a lot of pain at the moment with people trying to rerun failed backups.

kkalvagadda1:
> This change does what we expect it to.
> 
> I'm concerned that we may not want to overwrite existing backups though. Approving since I think this is causing a lot of pain at the moment with people trying to rerun failed backups.

There are few things to clarify
1. This is not an issue when the backup is performed manually. We are observing issues when database backup is retried when by datalake-dr due to some internal failures.
2. Each backup is associated with a backup-id. we are not replacing other backups. we are making sure there is another backup file associated with the same backup, we will be replacing it.

PR #11362, opened 2021-09-09
CB-13963 This commit reduces the namber of cases when we use "ambigou…

akanto:
…s" as a state. I have split the ambigous state into two different states, that are aligned with FreeIPA states as well. The two new states are:

- Unreachable -> when CM is not reachable e.g. CM is down, Nginx / cluster / proxy etc is down
- Node Failure ->  when we have node failure (e.g CM is down), or when we have service failure. e.g after CR 7.2.12.

There are oppurtunities to separate Node Failure into Agent Failure / Service failure states, but for that the next step is to refactor the state management in the InstanceMetadata.

### DIFF
--- a/datalake/src/main/java/com/sequenceiq/datalake/job/SdxClusterStatusCheckerJob.java
+++ b/datalake/src/main/java/com/sequenceiq/datalake/job/SdxClusterStatusCheckerJob.java
@@ -64,6 +64,8 @@ public class SdxClusterStatusCheckerJob extends StatusCheckerJob {
         handleSdxStatusChange();
     }
 
+    //It is hard to refactor something simpler than a switch case
+    //CHECKSTYLE:OFF: checkstyle:CyclomaticComplexity
     private void handleSdxStatusChange() {
         Optional<SdxCluster> cluster = sdxClusterRepository.findById(Long.valueOf(getLocalId()));
         cluster.ifPresent(sdx -> {
@@ -78,7 +80,9 @@ public class SdxClusterStatusCheckerJob extends StatusCheckerJob {
                     handleStoppedSdx(stack, sdx);
                     break;
                 case CLUSTER_AMBIGUOUS:
-                    handleAmbiguousSdx(stack, sdx);
+                case CLUSTER_UNREACHABLE:
+                case NODE_FAILURE:
+                    switchBackToAvailableIfClusterIsAvailable(stack, sdx, status.getStatus());
                     break;
                 case SYNC_FAILED:
                 case DATALAKE_UPGRADE_FAILED:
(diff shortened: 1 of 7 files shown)


datalake/src/main/java/com/sequenceiq/datalake/job/SdxClusterStatusCheckerJob.java
```diff
@@ -172,14 +177,27 @@ private void handleRunningSdx(StackStatusV4Response stack, SdxCluster sdx) {
...
+            logStateChange(DatalakeStatusEnum.RUNNING, DatalakeStatusEnum.CLUSTER_UNREACHABLE);
+        } else if (isNodeFailure(stack.getStatus(), stack.getClusterStatus())) {
+            String statusReason = stack.getStatus() == Status.NODE_FAILURE ? stack.getStatusReason() : stack.getClusterStatusReason();
+            sdxStatusService.setStatusForDatalakeAndNotify(DatalakeStatusEnum.NODE_FAILURE, ResourceEvent.CLUSTER_AMBARI_CLUSTER_SYNCHRONIZED,
+                    Collections.singleton(sdx.getClusterName()), statusReason, sdx);
+            logStateChange(DatalakeStatusEnum.RUNNING, DatalakeStatusEnum.NODE_FAILURE);
         }
     }
 
+    private boolean isUnreachable(Status stackStatus, Status clusterStatus) {
```
lnardai, line 193: So, here:
- stack could be: NODE_FAILURE, UNREACHABLE
- cluster could be: NODE_FAILURE, UNREACHABLE

In case of stack what is the difference between NODE_FAILURE and UNREACHABLE? Shouldn't it only be REACHABLE or UNREACHABLE? Or are we partitioning stackStatus like:

 UNREACHABLE: problem thorough provider API
 NODE_FAILURE: CCM or nginx does no work?

lnardai: I see that this comes from the older implementation as probably both statuses were set in conjunction.

sodre90 (approved):

datalake/src/main/java/com/sequenceiq/datalake/job/SdxClusterStatusCheckerJob.java
akanto: UNREACHABLE: -> We cannot reach, maybe CM, NGNIX, CP failed
NODE_FAILURE: -> CM, NGNIX, CP works, but it is still not working properly e.g. CM agent failed, Entire VM failed, or even just a service failed.

PR #11364, opened 2021-09-09
ENGESC-10264 backport CB-13027

sodre90:


gregito:
swagger validation is broken only which is reasonable hence the change itself LGTM

gregito (approved):

PR #11366, opened 2021-09-09
CB-13796 Autoscaling should adjust upscale request for CB Supported M…

smaniraju:
…ax Cluster NodeCount.

### DIFF
--- a/autoscale/src/main/java/com/sequenceiq/periscope/controller/validation/AlertValidator.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/controller/validation/AlertValidator.java
@@ -14,6 +14,7 @@ import org.springframework.stereotype.Component;
 import com.google.common.annotations.VisibleForTesting;
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
 import com.sequenceiq.cloudbreak.message.CloudbreakMessagesService;
+import com.sequenceiq.periscope.api.model.AdjustmentType;
 import com.sequenceiq.periscope.api.model.AlertType;
 import com.sequenceiq.periscope.api.model.DistroXAutoscaleClusterRequest;
 import com.sequenceiq.periscope.api.model.LoadAlertRequest;
@@ -25,6 +26,7 @@ import com.sequenceiq.periscope.domain.Cluster;
 import com.sequenceiq.periscope.service.AutoscaleRecommendationService;
 import com.sequenceiq.periscope.service.DateService;
 import com.sequenceiq.periscope.service.EntitlementValidationService;
+import com.sequenceiq.periscope.service.configuration.LimitsConfigurationService;
 
 @Component
 public class AlertValidator {
@@ -43,6 +45,9 @@ public class AlertValidator {
 
     @Inject
     private AutoScaleClusterCommonService asClusterCommonService;
+
+    @Inject
+    private LimitsConfigurationService limitsConfigurationService;
 
     public void validateEntitlementAndDisableIfNotEntitled(Cluster cluster) {
         if (!entitlementValidationService.autoscalingEntitlementEnabled(ThreadBasedUserCrnProvider.getAccountId(), cluster.getCloudPlatform())) {
@@ -80,6 +85,8 @@ public class AlertValidator {
             Set<String> timeAlertHostGroups = distroXAutoscaleClusterRequest.getTimeAlertRequests().stream()
                     .map(timeAlertRequest -> {
                         validateSchedule(timeAlertRequest);
+                        validateClusterLimit(timeAlertRequest.getScalingPolicy().getAdjustmentType(),
+                                timeAlertRequest.getScalingPolicy().getScalingAdjustment());
                         return timeAlertRequest;
                     })
                     .map(TimeAlertRequest::getScalingPolicy)
@@ -90,10 +97,23 @@ public class AlertValidator {
 
         if (!distroXAutoscaleClusterRequest.getLoadAlertRequests().isEmpty()) {
             Set<String> loadAlertHostGroups = distroXAutoscaleClusterRequest.getLoadAlertRequests().stream()
+                    .map(loadAlertRequest -> {
+                        validateClusterLimit(loadAlertRequest.getScalingPolicy().getAdjustmentType(),
+                                loadAlertRequest.getLoadAlertConfiguration().getMaxResourceValue());
+                        return loadAlertRequest;
+                    })
                     .map(LoadAlertRequest::getScalingPolicy)
                     .map(ScalingPolicyBase::getHostGroup)
                     .collect(Collectors.toSet());
             validateSupportedHostGroup(cluster, loadAlertHostGroups, AlertType.LOAD);
+        }
+    }
+
+    public void validateClusterLimit(AdjustmentType adjustmentType, Integer configuredMax) {
+        if ((adjustmentType == AdjustmentType.LOAD_BASED || adjustmentType == AdjustmentType.EXACT)
+                && configuredMax > limitsConfigurationService.getMaxNodeCountLimit()) {
+            throw new BadRequestException(messagesService.getMessage(MessageCode.AUTOSCALING_CLUSTER_LIMIT_EXCEEDED,
+                    List.of(limitsConfigurationService.getMaxNodeCountLimit())));
         }
     }
 
--- a/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ScalingPolicyTargetCalculator.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ScalingPolicyTargetCalculator.java
@@ -40,8 +40,6 @@ public class ScalingPolicyTargetCalculator {
                 desiredAbsoluteHostGroupNodeCount = hostGroupNodeCount;
         }
         int minSize = ScalingConstants.DEFAULT_HOSTGROUP_MIN_SIZE;
-        int maxSize = ScalingConstants.DEFAULT_HOSTGROUP_MAX_SIZE;
-
-        return desiredAbsoluteHostGroupNodeCount < minSize ? minSize : desiredAbsoluteHostGroupNodeCount > maxSize ? maxSize : desiredAbsoluteHostGroupNodeCount;
+        return desiredAbsoluteHostGroupNodeCount < minSize ? minSize : desiredAbsoluteHostGroupNodeCount;
     }
 }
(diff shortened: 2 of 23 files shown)


autoscale/src/main/java/com/sequenceiq/periscope/controller/validation/AlertValidator.java
```diff
@@ -90,13 +97,26 @@ public void validateDistroXAutoscaleClusterRequest(Cluster cluster,
...
                     .collect(Collectors.toSet());
             validateSupportedHostGroup(cluster, loadAlertHostGroups, AlertType.LOAD);
         }
     }
 
+    public void validateClusterLimit(AdjustmentType adjustmentType, Integer configuredMax) {
+        if ((adjustmentType == AdjustmentType.LOAD_BASED || adjustmentType == AdjustmentType.EXACT)
+                && configuredMax > limitsConfigurationService.getMaxNodeCountLimit()) {
+            throw new BadRequestException(messagesService.getMessage(MessageCode.AUTOSCALING_CLUSTER_LIMIT_EXCEEDED,
+                    List.of(limitsConfigurationService.getMaxNodeCountLimit())));
```
sidseth, line 116: This check will be at the hostGroup level, correct? The limit on the cluster size spans across all hostGroups in Cloudbreak? - so this would allow the cluster to be configured to a size which it cannot actually scale out to?

Should the existing cluster size be checked before configuring this property? (and factor in the min/max that is being configured on the hostGroup being configured, and the current size of that hostGroup?)

Even after this, it is possible that a different hostGroup is increased in size - in which case - the cluster will not actually be able to reach the configured max capacity (if another hostGroup is grown). If such a scenario is hit during actual scaling - I think it will be very useful to include this in the message that shows up on the console. Something like
"Cannot scale up to X nodes, Will exceed maximum cluster node count. Instead scaling to ${Actual Node Count}" (Fix text to be more aligned with the current scaling message)

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/LimitsConfigurationService.java
```diff
@@ -0,0 +1,28 @@
...
+public class LimitsConfigurationService {
+
+    @Inject
+    private CloudbreakCommunicator cbCommunicator;
+
+    @Cacheable(cacheNames = LIMITS_CONFIGURATION_CACHE)
+    public Integer getMaxNodeCountLimit() {
+        return Optional.ofNullable(cbCommunicator.getLimitsConfiguration())
+                .map(LimitsConfigurationResponse::getMaxNodeCountLimit)
+                .orElse(ScalingConstants.DEFAULT_CLUSTER_MAX_SIZE);
```
sidseth, line 26: if the orElse is ever triggered ... can we please log this?
Also, Given this gets refreshed every 10 minutes - generally log the limit being used. Will make debugging easier.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ScalingPolicyTargetCalculator.java
```diff
@@ -40,8 +40,6 @@ public Integer getDesiredAbsoluteNodeCount(ScalingEvent event, int hostGroupNode
                 desiredAbsoluteHostGroupNodeCount = hostGroupNodeCount;
         }
         int minSize = ScalingConstants.DEFAULT_HOSTGROUP_MIN_SIZE;
-        int maxSize = ScalingConstants.DEFAULT_HOSTGROUP_MAX_SIZE;
-
-        return desiredAbsoluteHostGroupNodeCount < minSize ? minSize : desiredAbsoluteHostGroupNodeCount > maxSize ? maxSize : desiredAbsoluteHostGroupNodeCount;
+        return desiredAbsoluteHostGroupNodeCount < minSize ? minSize : desiredAbsoluteHostGroupNodeCount;
```
sidseth, line 43: Although unlikely, should this be checking the current limit as well (in case it has been reduced from the time the policy was registered)?
This is also a behaviour question ... in case such a reduction were to happen, what should the user experience look like? - Fail all scaling operations which would otherwise hit the limit OR go to the current limit?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -81,35 +85,47 @@
...
         try {
-            int scalingAdjustment = desiredNodeCount - totalNodes;
+            int scalingAdjustment = desiredHostGroupNodeCount - existingHostGroupNodeCount;
             if (!decommissionNodeIds.isEmpty()) {
                 scaleDownByNodeIds(decommissionNodeIds);
             } else if (scalingAdjustment > 0) {
-                scaleUp(scalingAdjustment, totalNodes);
+                int cbSupportedMaxClusterSize = limitsConfigurationService.getMaxNodeCountLimit();
+                if ((existingClusterNodeCount + scalingAdjustment) > cbSupportedMaxClusterSize) {
+                    LOGGER.info("Upscale requested for '{}' nodes. Cluster can have maximium '{}' node(s)", scalingAdjustment, cbSupportedMaxClusterSize);
```
sidseth, line 111: Would be useful to log current node count (ideally with hostGroup break-up - which, am not sure is available)

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -81,35 +85,47 @@
...
             LOGGER.error("Error while executing ScaleRequest", e);
         }
     }
 
-    private void scaleUp(int scalingAdjustment, int totalNodes) {
+    private void scaleUp(int scalingAdjustment, int hostGroupNodeCount) {
         metricService.incrementMetricCounter(MetricType.CLUSTER_UPSCALE_TRIGGERED);
         if (scalingHardLimitsService.isViolatingAutoscaleMaxStepInNodeCount(scalingAdjustment)) {
-            LOGGER.debug("Upscale requested for '{}' nodes. Upscaling with the maximum allowed of '{}' node(s)",
+            LOGGER.info("Upscale requested for '{}' nodes. Upscaling with the maximum allowed of '{}' node(s)",
```
sidseth, line 128: Add step-size to log line?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -137,10 +153,10 @@ private void scaleUp(int scalingAdjustment, int totalNodes) {
             scalingStatus = ScalingStatus.FAILED;
             statusReason = getMessageForCBException(e);
             LOGGER.error("Couldn't trigger upscaling for host group '{}', cluster '{}', desiredNodeCount '{}', error '{}' ",
-                    hostGroup, cluster.getStackCrn(), desiredNodeCount, statusReason, e);
+                    hostGroup, cluster.getStackCrn(), desiredHostGroupNodeCount, statusReason, e);
             metricService.incrementMetricCounter(MetricType.CLUSTER_UPSCALE_FAILED);
         } finally {
-            processAutoscalingTriggered(scalingAdjustment, totalNodes, statusReason, scalingStatus);
+            processAutoscalingTriggered(scalingAdjustment, hostGroupNodeCount, statusReason, scalingStatus);
```
sidseth, line 159: Mentioned elsewhere. The decision to scale up, if affected by max-size, needs to be explicitly logged - to avoid confusion, and questions around "Why did this not scale-up to the configured size"

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
```diff
@@ -81,35 +85,47 @@
...
     public void run() {
         LoggingUtils.buildMdcContext(cluster);
         try {
-            int scalingAdjustment = desiredNodeCount - totalNodes;
+            int scalingAdjustment = desiredHostGroupNodeCount - existingHostGroupNodeCount;
             if (!decommissionNodeIds.isEmpty()) {
                 scaleDownByNodeIds(decommissionNodeIds);
             } else if (scalingAdjustment > 0) {
-                scaleUp(scalingAdjustment, totalNodes);
+                int cbSupportedMaxClusterSize = limitsConfigurationService.getMaxNodeCountLimit();
```
sidseth, line 109: Asked earlier also. Looks like the check is in place... 
Goes to the second question: Should the check be in place? (If CB reduces the limit - should the behaviour be to fail (user expectations not met) vs scale to the reduced limit.
This gets even more complicated since the same scenario can be hit if a different hostGroup is resized up (without the CB limit changing).

autoscale/src/main/java/com/sequenceiq/periscope/controller/validation/AlertValidator.java
smaniraju: CB does not support hostGroup level max (only total cluster limit) but for autoscaling we have to define hostgroup level max, hence this validation. 
But during actual scaling action, scale-up node count will be adjusted  based on actual cluster size at time of trigger.
Also added message to console indicating the same.

autoscale/src/main/java/com/sequenceiq/periscope/service/configuration/LimitsConfigurationService.java
smaniraju: ok

autoscale/src/main/java/com/sequenceiq/periscope/monitor/evaluator/ScalingPolicyTargetCalculator.java
smaniraju: Current CB supported max limit is checked just before triggering scale-up in ScalingRequest. 
Considering we are requesting partial scale-ups in CB, it would be better to adjust it to max allowed than fail.
For Ex : Scale-up is requested for 100 nodes when cluster size is 302, failing 98 node scale-up just because 2 nodes  would exceed limit, does not seem correct and also makes Autoscaling less effective.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: logged the current node count.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: done. logged to our logs and added to console also

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ScalingRequest.java
smaniraju: answered above also. just before scale-up is requested, node count is adjusted for max cluster size.

sidseth (approved):

PR #11367, opened 2021-09-09
[CB-13177] collect information from GCP

yg-gd:
continuation of CB-13177. makes the calls to GCP to collect needed info for CLI reporting only

### DIFF
--- a/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
+++ b/cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
@@ -17,6 +17,7 @@ import org.springframework.stereotype.Service;
 
 import com.google.api.services.compute.Compute;
 import com.google.api.services.compute.model.AccessConfig;
+import com.google.api.services.compute.model.BackendService;
 import com.google.api.services.compute.model.ForwardingRule;
 import com.google.api.services.compute.model.ForwardingRuleList;
 import com.google.api.services.compute.model.NetworkInterface;
@@ -25,6 +26,7 @@ import com.sequenceiq.cloudbreak.cloud.context.AuthenticatedContext;
 import com.sequenceiq.cloudbreak.cloud.gcp.client.GcpComputeFactory;
 import com.sequenceiq.cloudbreak.cloud.gcp.loadbalancer.GcpLoadBalancerTypeConverter;
 import com.sequenceiq.cloudbreak.cloud.gcp.util.GcpStackUtil;
+import com.sequenceiq.cloudbreak.cloud.gcp.view.GcpLoadBalancerMetadataView;
 import com.sequenceiq.cloudbreak.cloud.model.CloudCredential;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstance;
 import com.sequenceiq.cloudbreak.cloud.model.CloudInstanceMetaData;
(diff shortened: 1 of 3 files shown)


cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
```diff
@@ -159,10 +161,25 @@ private CloudVmMetaDataStatus getCloudVmMetaDataStatus(CloudResource cloudResour
...
                 LoadBalancerType itemType = gcpLoadBalancerTypeConverter.getScheme(item.getLoadBalancingScheme()).getCbType();
                 if (names.contains(item.getName()) && loadBalancerTypes.contains(itemType)) {
+                    Map<String, Object> params = new HashMap<>();
+                    List<String> ports = item.getPorts();
+                    if (ports == null || ports.size() != 1) {
+                        LOGGER.warn("Unexpected port count on {}, {}", item.getName(), ports);
+                    }
+                    try {
+                        String backendService = item.getBackendService();
+                        params.put(GcpLoadBalancerMetadataView.getBackendServiceParam(ports.get(0)), backendService);
```
hreeve-cloudera, line 171: You have a null check above that only lists a warning, but here you access `ports` directly. Should there be some logic to handle the case of `ports` being null?

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/view/GcpLoadBalancerMetadataView.java
```diff
@@ -32,7 +32,15 @@ public static String getInstanceGroupParam(int port) {
         return INSTANCE_GROUP_PREFIX + port;
```
hreeve-cloudera, line 32: Instead of having two implementations for these functions (simple as I know the implementations are), could you do something like
```
return getInstanceGroupParam(String.valueOf(port));
```

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/GCPLoadBalancerMetadataCollectorTest.java
```diff
@@ -226,9 +272,20 @@ private ForwardingRule createPrivateFowardingRule() {
         forwardingRule.setLoadBalancingScheme("INTERNAL");
         forwardingRule.setName(FORWARDING_RULE_NAME_2);
         forwardingRule.setIPAddress(PRIVATE_IP);
+        forwardingRule.setPorts(List.of(PORT2));
+        forwardingRule.setBackendService(BACKEND_SERVICE_NAME);
         return forwardingRule;
     }
```
hreeve-cloudera, line 279: If you don't have a test case for when `ForwardingRule.getPorts()` is null, you should add one. That's what tripped me up in CB-14040.

hreeve-cloudera (changes requested):

cloud-gcp/src/main/java/com/sequenceiq/cloudbreak/cloud/gcp/GcpMetadataCollector.java
yg-gd: IT was being handled with the catch, but better to do it explicitly

cloud-gcp/src/test/java/com/sequenceiq/cloudbreak/cloud/gcp/GCPLoadBalancerMetadataCollectorTest.java
yg-gd: good idea!

hreeve-cloudera (approved):

PR #11368, opened 2021-09-10
CB-13685 Implement getFreeIpaUpgradeOptions in FreeIpaUpgradeV1Contro…

lacikaaa:
…ller

- `ImageProvider` interface an related classes has been added the `getImages` method for fetching all available images
- all FreeIPA image is returned which is present in the catalog `versions` part, and have date field filled, and is newer than the curren image and the ID is different
- `ImageEntity` is enriched with a string date field to enable returning the date of the current image on the API
- upgrade and fetching upgrade options would use the current image's catalog instead of the default if not defined in the request
- `FreeIpaUpgradeOptions` contains the current image
- added `toString` implementation to API model classes

### DIFF
--- a/freeipa/src/main/java/com/sequenceiq/freeipa/service/image/CoreImageProvider.java
+++ b/freeipa/src/main/java/com/sequenceiq/freeipa/service/image/CoreImageProvider.java
@@ -1,17 +1,24 @@
 package com.sequenceiq.freeipa.service.image;
 
-import com.sequenceiq.cloudbreak.api.endpoint.v4.imagecatalog.ImageCatalogV4Endpoint;
-import com.sequenceiq.cloudbreak.api.endpoint.v4.imagecatalog.responses.ImageV4Response;
-import com.sequenceiq.freeipa.api.model.image.Image;
-import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.image.ImageSettingsRequest;
-import com.sequenceiq.freeipa.dto.ImageWrapper;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
-import javax.inject.Inject;
-import java.util.Optional;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.imagecatalog.ImageCatalogV4Endpoint;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.imagecatalog.responses.ImageV4Response;
+import com.sequenceiq.cloudbreak.api.endpoint.v4.imagecatalog.responses.ImagesV4Response;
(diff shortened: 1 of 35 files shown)


freeipa/src/main/java/com/sequenceiq/freeipa/service/upgrade/UpgradeImageService.java
```diff
@@ -35,15 +50,65 @@ public ImageInfoResponse selectImage(Stack stack, ImageSettingsRequest imageSett
...
         return imageInfoResponse;
     }
+
+    public List<ImageInfoResponse> findTargetImages(Stack stack, ImageSettingsRequest imageSettingsRequest, ImageInfoResponse currentImage) {
+        Optional<String> currentImageDate = getCurrentImageDate(stack, currentImage);
+        List<Pair<ImageWrapper, String>> imagesWrapperAndName = imageService.fetchImagesWrapperAndName(stack, imageSettingsRequest);
+        return imagesWrapperAndName.stream()
+                .filter(imageWrapperAndName -> currentImageDate.isPresent()
+                        && isCandidateImageNewerThanCurrent(imageWrapperAndName.getLeft(), currentImageDate.get()))
+                .filter(imageWrapperAndName -> !currentImage.getId().equals(imageWrapperAndName.getLeft().getImage().getUuid()))
```
pkedvessy, line 72: As far as I can see isCandidateImageNewerThanCurrent makes it sure that the candidate is different from the current image. Is this filtering here to make it sure it wont return the current one?

lacikaaa: you are right, it's to make sure we don't include it. theoretically it's not necessary

freeipa/src/main/java/com/sequenceiq/freeipa/service/image/CoreImageProvider.java
```diff
@@ -41,6 +51,29 @@
...
+    public List<ImageWrapper> getImages(ImageSettingsRequest imageSettings, String region, String platform) {
+        try {
+            ImagesV4Response imagesV4Response = imageCatalogV4Endpoint.getImagesByName(WORKSPACE_ID_DEFAULT, imageSettings.getCatalog(), null, platform, null,
+                    null);
+            LOGGER.debug("Images received: {}", imagesV4Response);
+            return Optional.ofNullable(imagesV4Response.getFreeipaImages()).
+                    orElseGet(List::of).stream()
+                    .map(this::convert)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
```
pkedvessy, line 64: Instead of these two lines I'd recommend to use `.flatMap(Optional::stream)`

pkedvessy (approved):
LGTM with a minor comment.

freeipa/src/main/java/com/sequenceiq/freeipa/service/upgrade/UpgradeImageService.java
pkedvessy: thanks

freeipa/src/main/java/com/sequenceiq/freeipa/service/image/CoreImageProvider.java
lacikaaa: thanks

PR #11371, opened 2021-09-10
CB-13732. Improve time based flow progress calculation & stop using s…

oleewere:
…tat caches

### DIFF
--- a/flow/src/main/java/com/sequenceiq/flow/core/stats/FlowOperationStatisticsService.java
+++ b/flow/src/main/java/com/sequenceiq/flow/core/stats/FlowOperationStatisticsService.java
@@ -1,15 +1,19 @@
 package com.sequenceiq.flow.core.stats;
 
+import static java.util.stream.Collectors.groupingBy;
+
 import java.text.DecimalFormat;
+import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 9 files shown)


flow/src/main/java/com/sequenceiq/flow/core/stats/FlowOperationStatisticsService.java
```diff
@@ -41,37 +51,66 @@
...
+            response.put(entry.getKey(), progressResponse);
+            createdMap.put(progressResponse.getCreated(), entry.getKey());
+        }
+        for (Map.Entry<Long, String> entry : createdMap.entrySet()) {
+            typeOrderList.add(entry.getValue());
+        }
+        if (!createdMap.isEmpty()) {
+            Long operationCreated = createdMap.entrySet().stream()
+                    .min(Comparator.comparingLong(Map.Entry::getKey))
+                    .get().getKey();
```
nadamdb, line 101: No need to stream the whole entrySet, streaming .keySet() is enough

nadamdb (changes requested):
I know the commented parts are only relocated, but it can be optimized a little by reducing unnecessary operations. Otherwise LGTM

nadamdb (approved):

PR #11372, opened 2021-09-10
CB-13986 Node is considered unreachable and cluster upscale fails whe…

lturcsanyi:
…n the Salt `network.ip_addrs` fun takes more time

- Salt timeout increased to 15 sec for network.ipaddrs calls
- Retry requests if unreachable nodes found

### DIFF
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/SaltConnector.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/SaltConnector.java
@@ -153,8 +153,12 @@ public class SaltConnector implements Closeable {
         return run(null, fun, clientType, clazz, arg);
     }
 
-    @Measure(SaltConnector.class)
     public <T> T run(Target<String> target, String fun, SaltClientType clientType, Class<T> clazz, String... arg) {
+        return run(target, fun, clientType, clazz, null, arg);
+    }
+
+    @Measure(SaltConnector.class)
+    public <T> T run(Target<String> target, String fun, SaltClientType clientType, Class<T> clazz, Long timeout, String... arg) {
         Form form = new Form();
         form = addAuth(form)
                 .param("fun", fun)
@@ -162,6 +166,9 @@ public class SaltConnector implements Closeable {
         if (target != null) {
             form = form.param("tgt", target.getTarget())
                     .param("tgt_type", target.getType());
+        }
+        if (timeout != null) {
+            form = form.param("t", timeout.toString());
         }
         if ("state.show_sls".equals(fun)) {
             form.param("full_return", "True");
--- a/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/domain/MinionIpAddressesResponse.java
+++ b/orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/domain/MinionIpAddressesResponse.java
@@ -20,9 +20,19 @@ public class MinionIpAddressesResponse {
     public List<String> getAllIpAddresses() {
         return result.stream()
                 .flatMap(result -> result.entrySet().stream())
+                .filter(entry -> entry.getValue() != null)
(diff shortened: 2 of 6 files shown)


orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/domain/MinionIpAddressesResponse.java
```diff
@@ -26,6 +26,15 @@
                 .collect(Collectors.toList());
     }
 
+    @JsonIgnore
+    public List<String> getUnreachableNodes() {
+        return result.stream()
+                .flatMap(result -> result.entrySet().stream())
+                .filter(entry -> "false".equals(entry.getValue().asText()))
```
horadla23, line 33: is it possible that `entry.getValue()` is null? if yes, then maybe it is worth to filter them out

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
```diff
@@ -204,19 +207,41 @@ private static void validateRunningInfoResultNotNull(RunningJobsResponse running
...
     }
 
-    public static MinionIpAddressesResponse collectMinionIpAddresses(Retry retry, SaltConnector sc) {
-        return retry.testWith1SecDelayMax5Times(() -> {
-            try {
-                return collectMinionIpAddresses(sc);
-            } catch (RuntimeException e) {
-                LOGGER.error("Collecting minion IP addresses failed", e);
-                throw new Retry.ActionFailedException("Collecting minion IP addresses failed", e);
+    public static Set<String> collectMinionIpAddresses(Retry retry, SaltConnector sc) {
```
horadla23, line 210: This method is quite large and has several catches, conditions, etc, so maybe it is better to split into at least 2 methods to increase readability

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/client/SaltConnector.java
```diff
@@ -153,8 +153,12 @@ public GenericResponses action(SaltAction saltAction) {
         return run(null, fun, clientType, clazz, arg);
     }
 
-    @Measure(SaltConnector.class)
     public <T> T run(Target<String> target, String fun, SaltClientType clientType, Class<T> clazz, String... arg) {
+        return run(target, fun, clientType, clazz, null, arg);
```
horadla23, line 157: based on description is this mean that the default timeout is 15 secs?

lturcsanyi: The default timeout for salt is 5 sec. Only the network.ipaddrs calls will use 15 sec timeout.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/domain/MinionIpAddressesResponse.java
lturcsanyi: Hopefully not, the same logic is used in the above method, it should have failed already.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/states/SaltStates.java
lturcsanyi: You are right, I will refactor this.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/domain/MinionIpAddressesResponse.java
horadla23: Ok, it makes sense

horadla23 (approved):
LGTM

PR #11378, opened 2021-09-11
CB-14012 Always register private IP to Cluster Proxy when using CCM

lajosrodek:
* Up to now, Cluster Proxy registrations used either public or private IPs, depending on the source subnet, even if CCM was active. This has been changed to always pick the VM private IP when any variant of CCM is employed.
  * Applies to both FreeIPA and DL / DH stacks, and all kinds of Cluster Proxy service, tunnel and CCMv2 endpoint entries.
  * The former FreeIPA logic was checking for `stack.getSecurityConfig().isUsePrivateIpToTls()` to decide if the private IP was preferable. That condition is a legacy remnant from 2.9 CB and, thus, not applicable to CDP deployments (where it will always evaluate to `false`).
* `InstanceMetaData` (`core-model`): Added a helper, `getIpWrapper(boolean)`, that returns either the private IP or the result of `getPublicIpWrapper()`, depending on the `boolean` argument.
* Testing: Added new UT & adapted existing ones.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
@@ -54,6 +54,8 @@ public class ClusterProxyService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClusterProxyService.class);
 
+    private static final boolean ALWAYS_PRIVATE_IP = true;
+
     @Inject
     private StackService stackService;
 
@@ -107,7 +109,7 @@ public class ClusterProxyService {
     }
 
     private void registerGateway(Stack stack) {
-        String knoxUrl = stack.getTunnel().useCcmV2OrJumpgate() ? knoxUrlForCcmV2(stack) : knoxUrl(stack);
+        String knoxUrl = stack.getTunnel().useCcmV2OrJumpgate() ? knoxUrlForCcmV2(stack) : knoxUrlForNoCcmAndCcmV1(stack);
         ConfigUpdateRequest request = new ConfigUpdateRequest(stack.getResourceCrn(), knoxUrl);
         clusterProxyRegistrationClient.updateConfig(request);
     }
@@ -120,7 +122,7 @@ public class ClusterProxyService {
     private ConfigRegistrationRequest createProxyConfigRequest(Stack stack) {
         ConfigRegistrationRequestBuilder requestBuilder = new ConfigRegistrationRequestBuilder(stack.getResourceCrn())
                 .withAliases(singletonList(clusterId(stack.getCluster())))
-                .withServices(serviceConfigs(stack)).withAccountId(getAccountId(stack));
+                .withServices(serviceConfigsForNoCcmAndCcmV1(stack)).withAccountId(getAccountId(stack));
         if (stack.getTunnel().useCcmV1()) {
             requestBuilder.withTunnelEntries(tunnelEntries(stack));
         } else if (stack.getTunnel().useCcmV2()) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
```diff
@@ -146,35 +148,37 @@ private ConfigRegistrationRequest createProxyConfigReRegisterRequest(Stack stack
         return requestBuilder.build();
     }
 
+    // For CCM disabled and CCMv1 only.
```
biharitomi, line 151: Next time please rename the function instead of the comment

biharitomi (approved):
LGTM

horadla23:
can you add to the description how you tested this locally and how you made sure if all service configs (all scenarios) are tested?
we had couple of issues regarding CP registrations weeks ago, those affected datalake HA repair, experiences (like DWX, they are using cloudera-manager registration to access CM), CCMv1 and CCMv2 cluster provisions, etc. and I would like to make sure if everything is still operating correctly, since unfortunately we do not have e2e tests for CCM based environments and QE also have limited tests for that as far as I understand.

biharitomi:
> can you add to the description how you tested this locally and how you made sure if all service configs (all scenarios) are tested?
> we had couple of issues regarding CP registrations weeks ago, those affected datalake HA repair, experiences (like DWX, they are using cloudera-manager registration to access CM), CCMv1 and CCMv2 cluster provisions, etc. and I would like to make sure if everything is still operating correctly, since unfortunately we do not have e2e tests for CCM based environments and QE also have limited tests for that as far as I understood.

You can find more details of [the investigated issue](https://jira.cloudera.com/browse/CB-14012) as it was reproduced on mow-dev and it seems to be a logic issue as it doesn't make sense to use public IPs with CCM by design.

horadla23:
> > can you add to the description how you tested this locally and how you made sure if all service configs (all scenarios) are tested?
> > we had couple of issues regarding CP registrations weeks ago, those affected datalake HA repair, experiences (like DWX, they are using cloudera-manager registration to access CM), CCMv1 and CCMv2 cluster provisions, etc. and I would like to make sure if everything is still operating correctly, since unfortunately we do not have e2e tests for CCM based environments and QE also have limited tests for that as far as I understood.
> 
> You can find more details of [the investigated issue](https://jira.cloudera.com/browse/CB-14012) as it was reproduced on mow-dev and it seems to be a logic issue as it doesn't make sense to use public IPs with CCM by design.

It's fine, I just want to make sure if every scenarios are tested regarding CP registration and calling CM or any gateway node through CP is working in case of CCMv1 and CCMv2

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/provision/service/ClusterProxyService.java
lajosrodek: Thanks, fixed.

lajosrodek:
@horadla23 I understand your concerns. Only a minimal amount of manual verification has taken place so far due to my last-minute vacation yesterday and today.

Just to clarify, the fixes in the present ticket only affect setups that involve CCM (of any variant) with _public_ subnets. The related service endpoints and other Cluster Proxy registration entries used to be registered with their public IPs so far. Doing so did not make a difference when using CCMv1 or regular CCMv2. With Jumpgate / CCMv2.5, however, communication with DL & DH got broken due to the usage of public IPs. (More details in in the [associated Jira ticket](https://jira.cloudera.com/browse/CB-14012).)

The above behavior has now been changed to always pick private IPs when CCM is enabled, no matter whether using a public or private subnet. Actually, the combination of "CCM+public subnet" does seem too realistic in practice, but we shall anyway support it equally well as the private subnet case. The changes here will make "CCM+public subnet" behave identically as "CCM+private subnet", the latter of which remains unchanged.

PR #11387, opened 2021-09-13
CB-14102: Fix ipa dns record errors so that it retries on more errors

jamisonbennett:
Fix the DNS record add script so that it retries when the following
error happens with err code 1:
ipa: ERROR: Configured time limit exceeded

Only retry when the error code is 1 and the error message is:
ipa: ERROR: no modifications to be performed

This was manually tested.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
@@ -22,9 +22,17 @@ for attempt in {1..3}
 do
   sleep 10
   echo "add dns a-record hostname for ${HOSTNAME}, attempt ${attempt}"
+  IPA_STDERR=$(mktemp)
   retVal=0
-  ipa dnsrecord-mod {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}" "--a-rec=${IPADDR}" --ttl {{ pillar['sssd-ipa']['dns_ttl'] }} || retVal=$?
-  if [[ "$retVal" -le 1 ]]; then
+  alreadyExists=0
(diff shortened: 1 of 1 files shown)


cegganesh84 (approved):
LGTM, thanks!

orchestrator-salt/src/main/resources/salt/salt/sssd/template/add_dns_record.j2
```diff
@@ -22,9 +22,16 @@ for attempt in {1..3}
...
   echo "add dns a-record hostname for ${HOSTNAME}, attempt ${attempt}"
+  IPA_STDERR=$(mktemp)
   retVal=0
-  ipa dnsrecord-mod {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}" "--a-rec=${IPADDR}" --ttl {{ pillar['sssd-ipa']['dns_ttl'] }} || retVal=$?
-  if [[ "$retVal" -le 1 ]]; then
+  alreadyExists=0
+  ipa dnsrecord-mod {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}" "--a-rec=${IPADDR}" --ttl {{ pillar['sssd-ipa']['dns_ttl'] }} 2>"${IPA_STDERR}" || retVal=$?
+  cat "${IPA_STDERR}" >&2
+  if grep "ERROR: no modifications to be performed" ${IPA_STDERR} &>/dev/null; then
+    alreadyExists=1
```
lacikaaa, line 31: I think some log would be useful here. Could you please add it?

jamisonbennett: Done

lacikaaa (approved):

PR #11389, opened 2021-09-14
CB-13736 Environment deletion failed on mow-int-eu1 control plane

horadla23:
- during sdx creation flow, user initiated environment deletion
- during sdx deletion, sdx set status to DELETE_REQUESTED and started to terminate related RDS
- because of RDS termination, sdx creation flow failed
- during sdx creation flow failure handling, sdx tries to set status to PROVISION_FAILED but it has thrown an unhandled exception, since we cannot set status from any "delete in progress" to PROVISION_FAILED
- unhandled exception will go to general sdx failure handling, which (SdxFlowInformation) tries to set status based on current status, since it was EXTERNAL_DATABASE_DELETION_IN_PROGRESS, the failure handler will set status to DELETE_FAILED
- because of DELETE_FAILED status, environment deletion will fail also, freeipa will continue to provision ipa server and it will cause extra costs for users
- solution: catch unhandled exception regarding status set from any "delete in progress" status to PROVISION_FAILED, thus there will be no unhandled exception and sdx failure handling will not ste status to DELETE_FAILED


foldik (approved):
Very good description! :)

PR #11392, opened 2021-09-14
CB-13464:Add validation to check if the mentioned Key exists in AWS

satyabollineni:
Verified he change with invalid (correct format but wrong ARN(the ARN of the key which is not present on AWS on the specified region)) and the env creation failed as belowScreenshot 2021-09-14 at 1 42 51 PM

Also validated the successful env creation with valid key.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentFlowValidatorService.java
+++ b/environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentFlowValidatorService.java
@@ -14,6 +14,7 @@ import com.sequenceiq.environment.environment.validation.cloudstorage.Environmen
 import com.sequenceiq.environment.environment.validation.cloudstorage.EnvironmentBackupStorageLocationValidator;
 import com.sequenceiq.environment.environment.validation.cloudstorage.EnvironmentLogStorageConfigurationValidator;
 import com.sequenceiq.environment.environment.validation.cloudstorage.EnvironmentLogStorageLocationValidator;
+import com.sequenceiq.environment.environment.validation.validators.EncryptionKeyArnValidator;
 import com.sequenceiq.environment.environment.validation.validators.EnvironmentAuthenticationValidator;
 import com.sequenceiq.environment.environment.validation.validators.EnvironmentNetworkProviderValidator;
 import com.sequenceiq.environment.environment.validation.validators.EnvironmentParameterValidator;
@@ -41,6 +42,8 @@ public class EnvironmentFlowValidatorService {
 
     private final EnvironmentAuthenticationValidator environmentAuthenticationValidator;
 
+    private final EncryptionKeyArnValidator encryptionKeyArnValidator;
+
     public EnvironmentFlowValidatorService(
             EnvironmentRegionValidator environmentRegionValidator,
             EnvironmentNetworkProviderValidator environmentNetworkProviderValidator,
@@ -49,7 +52,8 @@ public class EnvironmentFlowValidatorService {
             EnvironmentParameterValidator environmentParameterValidator,
             EnvironmentAuthenticationValidator environmentAuthenticationValidator,
             EnvironmentLogStorageConfigurationValidator logStorageConfigurationValidator,
-            EnvironmentBackupStorageConfigurationValidator backupStorageConfigurationValidator) {
+            EnvironmentBackupStorageConfigurationValidator backupStorageConfigurationValidator,
+            EncryptionKeyArnValidator encryptionKeyArnValidator) {
         this.environmentRegionValidator = environmentRegionValidator;
         this.environmentNetworkProviderValidator = environmentNetworkProviderValidator;
         this.logStorageLocationValidator = logStorageLocationValidator;
@@ -58,6 +62,7 @@ public class EnvironmentFlowValidatorService {
         this.environmentAuthenticationValidator = environmentAuthenticationValidator;
         this.logStorageConfigurationValidator = logStorageConfigurationValidator;
         this.backupStorageConfigurationValidator = backupStorageConfigurationValidator;
+        this.encryptionKeyArnValidator = encryptionKeyArnValidator;
     }
 
     public ValidationResult.ValidationResultBuilder validateRegionsAndLocation(String location, Set<String> requestedRegions,
@@ -97,6 +102,14 @@ public class EnvironmentFlowValidatorService {
 
     public ValidationResult validateAuthentication(EnvironmentValidationDto environmentValidationDto) {
         return environmentAuthenticationValidator.validate(environmentValidationDto);
+    }
+
+    public ValidationResult validateAwsKeysPresent(EnvironmentValidationDto environmentValidationDto, ParametersDto parametersDto) {
+        ValidationResult.ValidationResultBuilder validationResultBuilder = ValidationResult.builder();
+        if (environmentValidationDto.getEnvironmentDto().getCloudPlatform().equals("AWS")) {
+            return encryptionKeyArnValidator.validate(environmentValidationDto, parametersDto);
+        }
+        return validationResultBuilder.build();
     }
 
 }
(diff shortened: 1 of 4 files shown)


environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
```diff
@@ -25,4 +52,34 @@ public ValidationResult validateEncryptionKeyArn(String encryptionKeyArn) {
...
+        CloudEncryptionKeys encryptionKeys = new CloudEncryptionKeys(new HashSet<>());
+
+        try {
+            CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(
+                    Platform.platform(environmentDto.getCloudPlatform()), null);
+            encryptionKeys = cloudPlatformConnectors.get(cloudPlatformVariant)
+                    .platformResources().encryptionKeys(cloudCredential, region, Collections.emptyMap());
+        } catch (RuntimeException e) {
+            LOGGER.error("Failed to fetch AWS Cloud keys", e.getMessage());
+
```
cegganesh84, line 72: We need to give different error message to the user right? This is not a user input problem yet, just that we are not able to validate it, may be due to many reasons like API throttling, network error etc.

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
```diff
@@ -25,4 +52,34 @@ public ValidationResult validateEncryptionKeyArn(String encryptionKeyArn) {
...
+        }
+
+        if (encryptionKeyArn != null && !encryptionKeyArn.isEmpty()) {
+
+            if (encryptionKeys.getCloudEncryptionKeys().stream().map(CloudEncryptionKey::getName)
+                    .noneMatch(s -> s.equals(encryptionKeyArn))) {
+                validationResultBuilder.error("The provided encryption key does not exists in the given region's encryption key list for this credential.");
```
cegganesh84, line 79: nit: exists -> exist.

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
satyabollineni: `"An unexpected error occurred while trying to fetch the AWS cloud keys", e.getMessage()` Could you please confirm if this error message makes more sense to the user than the previous one as the e.getmessage would have the error details for user to refer to ?

cegganesh84: @satyabollineni We are swallowing the exception here. This means the error message we are sending to the user will have "The provided encryption key does not exist in the given region's encryption key list for this credential.". from the code below. So we are going to give an error message that says two things, what the problem was and what the user could do next. For example, including a phrase like please retry the operation will help the user for this case. Also, according to https://jira.cloudera.com/browse/CB-12758?focusedCommentId=4749072&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4749072 these types of validation failures (AWS throttling) should result in cluster creation to proceed and the user gets an indication in events that the validation was not fully complete. CC: @lnardai

satyabollineni: @cegganesh84 Thanks for elaborating on your comment above. I agree with you on the retry part and I would introduce the retry logic If we face any failures from AWS end like AWS throttling etc. But I have a slightly different view on the second part i.e **"these types of validation failures (AWS throttling) should result in cluster creation to proceed and the user gets an indication in events that the validation was not fully complete."**. If a user passes the encryption keys, User is expecting to encrypt the data but for any reason If the validation fails and we proceed to cluster creation without encryption then that would be contrast to the user expectation of encrypting the data on cluster.

Lets say we proceed without validating the key (if validation failed due to AWS throttling etc) and try to encrypt the data then at the later point If the key is invalid we fail to encrypt which is not expected for the end user. (of course things will go good If luckily the key is valid). In my opinion failing would be more appropriate in these cases rather than proceeding. But I would be happy to know If I am missing any point here. cc @lnardai

cegganesh84: Yes, we should always fail if the encryption parameter is given but we could not encrypt, there is no different viewpoint there. AWS will give an error message if you pass an invalid encryption arn, and we will rely on it. The issue is only during the validation phase (which is to give an upfront error message) about what to do with transient issues because of which we do not know if the key is valid or not.

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
```diff
@@ -25,4 +59,32 @@ public ValidationResult validateEncryptionKeyArn(String encryptionKeyArn) {
...
+        CloudEncryptionKeys encryptionKeys;
+        CloudPlatformVariant cloudPlatformVariant = new CloudPlatformVariant(
+                Platform.platform(environmentDto.getCloudPlatform()), null);
+        try {
+            encryptionKeys =  retryService.testWith2SecDelayMax15Times(() -> cloudPlatformConnectors.get(cloudPlatformVariant).
+                                                                            platformResources().encryptionKeys(cloudCredential, region, Collections.emptyMap()));
+        } catch (Retry.ActionFailedException | AmazonServiceException e) {
+            LOGGER.error("An unexpected error occurred while trying to fetch the KMS keys from AWS");
+            throw e;
+        }
```
cegganesh84, line 79: Let's go with this for now, but needs unit testing.

satyabollineni: Added the unit tests

cegganesh84 (approved):
LGTM. Thanks for the changes.

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
```diff
@@ -25,4 +61,32 @@ public ValidationResult validateEncryptionKeyArn(String encryptionKeyArn) {
         }
         return validationResultBuilder.build();
     }
+
+    public ValidationResult validate(EnvironmentValidationDto environmentValidationDto, ParametersDto parametersDto) {
+        EnvironmentDto environmentDto = environmentValidationDto.getEnvironmentDto();
+        ValidationResult.ValidationResultBuilder validationResultBuilder = ValidationResult.builder();
+        String encryptionKeyArn = parametersDto.getAwsParametersDto().getAwsDiskEncryptionParametersDto().getEncryptionKeyArn();
```
juhi-09, line 68: Validate the key existence only when the `encryptionKeyArn` is present - Need not to make the cloud call if no encryption key is provided by user.

juhi-09 (changes requested):

juhi-09 (approved):
LGTM, Thanks for the changes.

cegganesh84:
@satyabollineni Could you please squash the commits into one with the latest commit message?

environment/src/main/java/com/sequenceiq/environment/environment/validation/EnvironmentFlowValidatorService.java
```diff
@@ -99,4 +104,12 @@ public ValidationResult validateAuthentication(EnvironmentValidationDto environm
         return environmentAuthenticationValidator.validate(environmentValidationDto);
     }
 
+    public ValidationResult validateAwsKeysPresent(EnvironmentValidationDto environmentValidationDto, ParametersDto parametersDto) {
+        ValidationResult.ValidationResultBuilder validationResultBuilder = ValidationResult.builder();
+        if (environmentValidationDto.getEnvironmentDto().getCloudPlatform().equals("AWS")) {
```
biharitomi, line 109: Why a simple string uses a constant is minimum I think, but using the com.sequenceiq.cloudbreak.common.mappable.CloudPlatform would take much more sense and be safer.

environment/src/main/java/com/sequenceiq/environment/environment/validation/validators/EncryptionKeyArnValidator.java
biharitomi: I think it would be good to validate that even the AwsParametersDto and the AwsDiskEncryptionParametersDto are not null in this call chain.

PR #11396, opened 2021-09-14
CB-13044 REST endpoint to execute sync in SDX - part of epic: CB-1273…

gergopapi2:
…6, After a failed DL/DH upgrade sync runtime versions

Customer should be able to manually issue a cdp-cli command to execute the syncing of
- CM versions stored in CB Component and ClusterComponent tables to that of CM, and
- SDX version in SdxCluster table.

For that this commit introduces
- an endpoint on sdx for sync cm in sdx
- a flow to call sync cm in cloudbreak
- and set the SDX cluster version based on the CDH parcel version

Besides, this commit introduces package sdx.flowwait, a more generic way for polling tasks waiting for flows in cloudbreak, to enhance reusability.

Also, within flowwait, SdxWaitTaskForCloudbreakFlow introduces a wait task that waits only on the cloudbreak flow state without checking cluster state.


datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncFlowConfig.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public SdxCmSyncEvent[] getInitEvents() {
+        return new SdxCmSyncEvent[]{
+                SDX_CM_SYNC_START_EVENT
+        };
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Sync CM version and parcels and update SDX version";
```
foldik, line 66: You could have used Data Lake instead of SDX.

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncFlowConfig.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.List;
+
+import org.springframework.stereotype.Component;
+
+import com.sequenceiq.flow.core.config.AbstractFlowConfiguration;
+import com.sequenceiq.flow.core.config.AbstractFlowConfiguration.Transition.Builder;
+
+@Component
+public class SdxCmSyncFlowConfig extends AbstractFlowConfiguration<SdxCmSyncState, SdxCmSyncEvent> {
```
foldik, line 25: It's not retryable. Currently it's not a problem since the user can just trigger it again, it doesn't do much.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudbreakFlowService.java
```diff
@@ -42,6 +43,11 @@ public FlowLogResponse getLastCloudbreakFlowChainId(SdxCluster sdxCluster) {
                 flowEndpoint.getFlowLogsByFlowId(flowId));
     }
 
+    public boolean getFlowResultByFlowId(SdxCluster sdxCluster) {
```
foldik, line 46: Generally what we need to do is to check the flow state. Did it complete, did it fail. SdxCluster has

```java
    private String lastCbFlowChainId;

    private String lastCbFlowId;
```

columns. Based on these we should be able to use 

- `com.sequenceiq.flow.api.FlowEndpoint#hasFlowRunningByChainId`
- `com.sequenceiq.flow.api.FlowEndpoint#hasFlowRunningByFlowId`

endpoints and figure out the state (completed / failed). We should not poll the entire flow log and make decisions based on it. FlowEndpoint already provides a solution to do it better.

Additionally the below logic wont work if the flow first failed, then retried (there will be FAILED states).

```java
return flowLogsByFlowId.stream().allMatch(fl -> StateStatus.SUCCESSFUL.equals(fl.getStateStatus()));
```

foldik (changes requested):

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/CloudbreakFlowService.java
gergopapi2: Yes, you are right, will change it.

I used it because of a special use-case: it is possible that a flow state fails but the whole flow is not failed. But that is a special thing, should not be generalized.

foldik: Yes after a retry there will be failed flow states but it can be successful flow so it's better to handle it. I afraid that developers wont go into the details just assume it's working but actually it's not. Other thing is that it would be better to give it a name like isLastFlowSuccessful(cluster) or something.

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncFlowConfig.java
gergopapi2: I did not change it. This will be invokeable from cdp-cli, so retry does not make much sense. If you like I can make it retryable.

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncFlowConfig.java
gergopapi2: Changed, thank you!

tiborpopovics (approved):

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxUpgradeController.java
```diff
@@ -39,6 +40,9 @@
     @Inject
     private SdxUpgradeRecoveryService sdxUpgradeRecoveryService;
 
+    @Inject
+    private SdxService sdxService;
```
lacikaaa, line 44: this is not used

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/event/SdxCmSyncStartEvent.java
```diff
@@ -0,0 +1,11 @@
+package com.sequenceiq.datalake.flow.datalake.cmsync.event;
+
+import static com.sequenceiq.datalake.flow.datalake.cmsync.SdxCmSyncEvent.SDX_CM_SYNC_START_EVENT;
+
+import com.sequenceiq.datalake.flow.SdxEvent;
+
+public class SdxCmSyncStartEvent extends SdxEvent {
+    public SdxCmSyncStartEvent(Long sdxId, String sdxName, String userId) {
+        super(SDX_CM_SYNC_START_EVENT.event(), sdxId, sdxName, userId);
```
lacikaaa, line 9: why? I mean `SdxEvent` default selector would be the same

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/handler/SdxCmSyncWaitHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    private SdxService sdxService;
+
+    @Override
+    public String selector() {
+        return EventSelectorUtil.selector(SdxCmSyncWaitEvent.class);
+    }
+
+    @Override
+    protected Selectable defaultFailureEvent(Long resourceId, Exception e, Event<SdxCmSyncWaitEvent> event) {
+        return new SdxCmSyncFailedEvent(resourceId, event.getData().getUserId(), e);
```
lacikaaa, line 49: could you add log?

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/handler/SdxCmSyncWaitHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    protected Selectable doAccept(HandlerEvent<SdxCmSyncWaitEvent> event) {
+        SdxCmSyncWaitEvent sdxCmSyncWaitEvent = event.getData();
+        try {
+            SdxCluster sdxCluster = sdxService.getById(sdxCmSyncWaitEvent.getResourceId());
+            PollingConfig pollingConfig = new PollingConfig(sleepTimeInSec, TimeUnit.SECONDS, durationInMinutes, TimeUnit.MINUTES)
+                    .withStopPollingIfExceptionOccurred(true);
+            sdxWaitService.waitForCloudbreakFlow(sdxCluster, pollingConfig, "Sync cm");
+            return new SdxEvent(SDX_CM_SYNC_FINISHED_EVENT.event(), event.getData().getResourceId(), event.getData().getUserId());
+        } catch (SdxWaitException e) {
+            return new SdxCmSyncFailedEvent(sdxCmSyncWaitEvent.getResourceId(), sdxCmSyncWaitEvent.getUserId(), e);
```
lacikaaa, line 62: could you add log?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxCmSyncService.java
```diff
@@ -0,0 +1,48 @@
...
+    @Inject
+    private StackV4Endpoint stackV4Endpoint;
+
+    @Inject
+    private CloudbreakFlowService cloudbreakFlowService;
+
+    @Inject
+    private WebApplicationExceptionMessageExtractor exceptionMessageExtractor;
+
+    public void callCmSync(Long sdxId) {
```
lacikaaa, line 34: could you add some log?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -557,6 +558,18 @@ public void syncByCrn(String userCrn, String crn) {
...
+    public SdxSyncComponentVersionsFromCmResponse syncComponentVersionsFromCm(String userCrn, NameOrCrn clusterNameOrCrn) {
+        SdxCluster cluster = getByNameOrCrn(userCrn, clusterNameOrCrn);
+        MDCBuilder.buildMdcContext(cluster);
+        SdxStatusEntity sdxStatus = sdxStatusService.getActualStatusForSdx(cluster);
+        if (sdxStatus.getStatus().isStopState()) {
+            return new SdxSyncComponentVersionsFromCmResponse(
+                    "Syncing CM and parcel versions from CM cannot be initiated as the datalake is either stopped or is stopping");
+        }
+        return new SdxSyncComponentVersionsFromCmResponse("Syncing CM and parcel versions from CM initiated",
+                sdxReactorFlowManager.triggerDatalakeSyncComponentVersionsFromCmFlow(cluster));
```
lacikaaa, line 570: could you add logs and `else`?

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/flowwait/task/SdxWaitTaskForCloudbreakFlow.java
```diff
@@ -0,0 +1,59 @@
...
+import com.sequenceiq.datalake.entity.SdxCluster;
+import com.sequenceiq.datalake.service.sdx.PollingConfig;
+import com.sequenceiq.datalake.service.sdx.flowcheck.CloudbreakFlowService;
+import com.sequenceiq.datalake.service.sdx.flowcheck.FlowState;
+
+/**
+ * This wait task will wait for a cloudbreak flow or flow chain to finish, then checks the result of the flow.
+ */
+@Component
+@Scope("prototype")
```
lacikaaa, line 22: I would say it would be better if this would be a pojo and expect all dependency in the constructor, which would be provided by `SdxWaitTaskFactory`

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/flowwait/task/SdxWaitTaskFactory.java
```diff
@@ -0,0 +1,20 @@
...
+import com.sequenceiq.datalake.service.sdx.PollingConfig;
+
+@Service
+public class SdxWaitTaskFactory {
+
+    @Inject
+    private ApplicationContext applicationContext;
+
+    public SdxWaitTaskForCloudbreakFlow createCloudbreakFlowWaitTask(SdxCluster sdxCluster, PollingConfig pollingConfig, String pollingMessage) {
+        return applicationContext.getBean(SdxWaitTaskForCloudbreakFlow.class, sdxCluster, pollingConfig, pollingMessage);
```
lacikaaa, line 18: I think it would be better to use `new` here

datalake/src/main/java/com/sequenceiq/datalake/controller/sdx/SdxUpgradeController.java
gergopapi2: Right, thanks. It might make sense to add a unit test checking that, the same we have for core.

gergopapi2: Asked it in the dev_internal.

datalake/build.gradle
```diff
@@ -87,7 +87,8 @@ dependencies {
   implementation     group: 'com.cloudera.api.swagger',    name: 'cloudera-manager-api-swagger',  version: cmClientVersion
 
   testImplementation group: 'org.springframework.boot',  name: 'spring-boot-starter-test',        version: springBootVersion
-  testImplementation group: 'org.mockito',               name: 'mockito-core',                    version: mockitoVersion
+  testImplementation group: 'org.mockito',               name: 'mockito-inline',                  version: mockitoVersion
```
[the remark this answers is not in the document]
gergopapi2: I wanted to mock a static method:
https://github.com/hortonworks/cloudbreak/blob/4482d8cecde8808fa1a418932585e6fb32e98bc8/datalake/src/test/java/com/sequenceiq/datalake/service/sdx/flowwait/SdxWaitTaskServiceTest.java#L68

That is possible:
https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#static_mocks

In fact I got an error message:
"org.mockito.exceptions.base.MockitoException: 
The used MockMaker SubclassByteBuddyMockMaker does not support the creation of static mocks

Mockito's inline mock maker supports static mocks based on the Instrumentation API.
You can simply enable this mock mode, by placing the 'mockito-inline' artifact where you are currently using 'mockito-core'.
Note that Mockito's inline mock maker is not supported on Android."


I then tried mockito-inline and it worked the same for all the other tests as mockito-core.

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncActions.java
```diff
@@ -0,0 +1,102 @@
...
+            protected void doExecute(SdxContext context, SdxEvent payload, Map<Object, Object> variables) {
+                Optional<String> newVersionOpt = sdxUpgradeService.updateRuntimeVersionFromCloudbreak(payload.getResourceId());
+                LOGGER.info("Sdx cm sync updated sdx version, id: {}, new version: {}", payload.getResourceId(), newVersionOpt);
+                sendEvent(context, SDX_CM_SYNC_FINALIZED_EVENT.event(), payload);
+            }
+        };
+    }
+
+    @Bean(name = "SDX_CM_SYNC_FAILED_STATE")
+    public Action<?, ?> cmSyncFailedAction() {
```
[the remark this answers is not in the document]
gergopapi2: This is on purpose: even if the sync does not work it should not penalize the user with a FAILED state.

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncActions.java
```diff
@@ -0,0 +1,102 @@
...
+
+            @Override
+            protected void doExecute(SdxContext context, SdxCmSyncFailedEvent payload, Map<Object, Object> variables) throws Exception {
+                LOGGER.info("Sdx cm sync failure, id: {}, error: {}", payload.getResourceId(), payload.getException());
+                sendEvent(context, SDX_CM_SYNC_FAILED_HANDLED_EVENT.event(), payload);
+            }
+
+            @Override
+            protected Object getFailurePayload(SdxCmSyncFailedEvent payload, Optional<SdxContext> flowContext, Exception ex) {
+                return null;
```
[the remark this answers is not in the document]
gergopapi2: Yes, we checked that in a freeipa flow: firing the same SdxCmSyncFailedEvent will not trigger loop as it has no failureEvent, it will launch Missing error handling.
Fixed

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/event/SdxCmSyncStartEvent.java
gergopapi2: I did it for getSdxName to work, as no other ctor does set the sdx name. I did not want to refactor here, neither wanted to introduce another ctor.
I used sdxName here. But maybe it is really not needed as the sole id we use in kibana is the datalake crn.
https://github.com/hortonworks/cloudbreak/blob/4482d8cecde8808fa1a418932585e6fb32e98bc8/datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/SdxCmSyncActions.java#L47

datalake/src/main/java/com/sequenceiq/datalake/flow/datalake/cmsync/handler/SdxCmSyncWaitHandler.java
gergopapi2: added

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxCmSyncService.java
gergopapi2: Yes, added.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
gergopapi2: added

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/flowwait/task/SdxWaitTaskFactory.java
gergopapi2: SdxWaitTaskForCloudbreakFlow has an injected field: if using new that would not be injected.

PR #11398, opened 2021-09-14
CB-13938 Extend core_settings configuration provider with new property

patrik-marton:
A new property 'hadoop.security.groups.cache.background.reload' was added to core_settings. If set to true, this enables a small threadpool of (default) 3 threads to fetch the security groups in the background, and this prevents the calling threads from blocking. Since this property is not present in the CSD, a safety valve was used.
The related test class is extended with Mockito to allow dependency injection in the provider that's needed for the unit tests.

Tested:

- Unit test
- Provisioned cluster

Note:

- A ticket was raised [OPSAPS-61614](https://jira.cloudera.com/browse/OPSAPS-61614) to add the property to the CSD. Once it's completed the safety valve can be removed.


doktoric (approved):

lajosrodek:
@patrik-marton Is this setting applicable to all CDH runtime versions supported in CDP PC?

patrik-marton:
> @patrik-marton Is this setting applicable to all CDH runtime versions supported in CDP PC?

@lajosrodek This behaviour was observed in 7.2.10, but I believe it can affect all CDH versions if the environment becomes too busy.

patrik-marton:
retest this please

PR #11401, opened 2021-09-14
CB-14110 we shouldn't limit the custom template upgrades with the DH …

keyki:
…upgrade entitlement

If the CDP_RUNTIME_UPGRADE_DATAHUB entitlement is granted we shouldn't limit the type
of customer templates that can be upgraded. Neither the version nor the type of services
should be limited.


keyki:
This is a backport from master. The 2.48 branch was created slightly faster than it was merged.

foldik (approved):
LGTM

PR #11402, opened 2021-09-14
CB-14075 Environment freeipa poller is not going to exit only the checkStatus in order to handle freeipa deletion

afarsang:


biharitomi (approved):
LGTM

PR #11404, opened 2021-09-14
CB-13732. Improve time based flow progress calculation & stop using s…

oleewere:
…tat caches

this change is a fix for a 2.48 feature, branching happened right before that so merging it back to 2.48


foldik (approved):
LGTM

PR #11406, opened 2021-09-15
CB-14106 REST endpoint to execute cm sync in DistroX - part of epic: …

gergopapi2:
…CB-12736, After a failed DL/DH upgrade sync runtime versions

Customer should be able to manually issue a cdp-cli command to execute the syncing of the CM and active parcel versions from the CM server to CB Component and ClusterComponent tables.
For that end this commit introduces an endpoint in DistroX. The new endpoint calls the already existing sync cm flow.

### DIFF
--- a/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
+++ b/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
@@ -77,6 +77,7 @@ import com.sequenceiq.distrox.api.v1.distrox.model.cluster.DistroXMultiDeleteV1R
 import com.sequenceiq.distrox.api.v1.distrox.model.diagnostics.docs.DiagnosticsOperationDescriptions;
 import com.sequenceiq.distrox.api.v1.distrox.model.diagnostics.model.CmDiagnosticsCollectionV1Request;
 import com.sequenceiq.distrox.api.v1.distrox.model.diagnostics.model.DiagnosticsCollectionV1Request;
+import com.sequenceiq.distrox.api.v1.distrox.model.upgrade.DistroXSyncCmV1Response;
 import com.sequenceiq.flow.api.model.FlowIdentifier;
 import com.sequenceiq.flow.api.model.FlowProgressResponse;
 import com.sequenceiq.flow.api.model.operation.OperationView;
@@ -437,4 +438,17 @@ public interface DistroXV1Endpoint {
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = ROTATE_CERTIFICATES, nickname = "rotateAutoTlsCertificatesByCrn")
     CertificatesRotationV4Response rotateAutoTlsCertificatesByCrn(@PathParam("crn") String crn, @Valid CertificatesRotationV4Request rotateCertificateRequest);
+
+    @POST
+    @Path("{name}/sync_cm")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "syncs from distrox cluster CM the CM and parcel versions", produces = MediaType.APPLICATION_JSON, nickname = "syncDistroxCm")
+    DistroXSyncCmV1Response syncComponentVersionsFromCmByName(@PathParam("name") String name);
+
+    @POST
+    @Path("crn/{crn}/sync_cm")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "syncs from distrox cluster CM the CM and parcel versions", produces = MediaType.APPLICATION_JSON, nickname = "syncDistroxCmByCrn")
+    DistroXSyncCmV1Response syncComponentVersionsFromCmByCrn(@PathParam("crn") String crn);
+
 }
--- a/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/upgrade/DistroXSyncCmV1Request.java
+++ b/core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/upgrade/DistroXSyncCmV1Request.java
@@ -1,0 +1,29 @@
+package com.sequenceiq.distrox.api.v1.distrox.model.upgrade;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModelProperty;
+
+public class DistroXSyncCmV1Request {
+
+    @ApiModelProperty(ModelDescriptions.CmSyncRequest.IMAGE_IDS)
+    private Set<String> candidateImageUuids = new HashSet<>();
+
+    public Set<String> getCandidateImageUuids() {
+        return candidateImageUuids;
+    }
+
+    public void setCandidateImageUuids(Set<String> candidateImageUuids) {
+        this.candidateImageUuids = candidateImageUuids;
+    }
+
+    @Override
+    public String toString() {
+        return "DistroXSyncCmV1Request{" +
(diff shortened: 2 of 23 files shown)


core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXUpgradeV1Endpoint.java
```diff
@@ -51,4 +53,16 @@ DistroXUpgradeV1Response upgradeClusterByNameInternal(@PathParam("name") String
     DistroXUpgradeV1Response upgradeClusterByCrnInternal(@PathParam("crn") String crn, @Valid DistroXUpgradeV1Request distroxUpgradeRequest,
             @QueryParam("initiatorUserCrn") String initiatorUserCrn);
 
+    @POST
+    @Path("{name}/sync_cm")
```
doktoric, line 57: is this an upgrade related thing?

gergopapi2: When an upgrade fails, versions in Component, ClusterComponent table may slip apart from what is on the cluster. Further, if a customer tries to fix the cluster by hand and installs the newer version of a parcel, then the cluster might be ok but that is not reflected in the aformentioned DB tables.

The purpose of this sync_cm command is to sync versions in the cbdb tables to what is really on the CM. So in this regard this is an upgrade related command.

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/upgrade/DistroXSyncCmV1Request.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Set;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModelProperty;
+
+public class DistroXSyncCmV1Request {
+
+    @ApiModelProperty(ModelDescriptions.CmSyncRequest.IMAGE_IDS)
+    private Set<String> candidateImageUuids = new HashSet<>();
```
pdarvasi, line 13: why do we want to expose candidateImageUuids? It is only because of the additional time to run if we would omit it?
I am asking because 
- you've made some measures that looping across all images doesn't take that much time
- I find it difficult to imagine that any end user would call it with candidateImageUuids set.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/controller/DistroXUpgradeV1Controller.java
```diff
@@ -80,6 +84,23 @@ public DistroXUpgradeV1Response upgradeClusterByCrnInternal(@ResourceCrn String
                 dataHubOsUpgradeEntitled));
     }
 
+    @Override
+    @CheckPermissionByResourceName(action = AuthorizationResourceAction.UPGRADE_DATAHUB)
```
pdarvasi, line 88: why do you tie to upgrade here? Isn't it meant to be a general endpoint in the end? I would rather place it in the general controller and maybe introduce a new action, as I couldn't find an appropriate for the first sight

core/src/main/java/com/sequenceiq/distrox/v1/distrox/service/upgrade/DistroXUpgradeService.java
```diff
@@ -70,6 +73,15 @@ public UpgradeV4Response triggerUpgrade(NameOrCrn cluster, Long workspaceId, Str
         return initUpgrade(request, upgradeV4Response, cluster, workspaceId);
     }
 
+    public DistroXSyncCmV1Response triggerSyncCm(NameOrCrn cluster, Long workspaceId, Set<String> candidateImageUuids) {
+        Stack stack = stackService.getByNameOrCrnInWorkspace(cluster, workspaceId);
+        if (stack.getStackStatus().getStatus().isStopState()) {
```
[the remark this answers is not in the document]
gergopapi2: I was most worried about the stopped state. In starting state it is not possible to trigger it as a flow is under operation, so all *_IN_PROGRESS states should be covered, meaning interim states should not be a problem.

I will change it to stopped state.

core/src/main/java/com/sequenceiq/distrox/v1/distrox/controller/DistroXUpgradeV1Controller.java
gergopapi2: I can, but this action should mostly be used when an upgrade fails. That is, only persons who are eligible for upgrade should be using it.

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/upgrade/DistroXSyncCmV1Request.java
gergopapi2: A not so straightforward usecase would be: 
upgrade fails and customer installs parcels by hand. But, for some reason he installs a different version of a parcel than stated in the package version on the target image. In this case cloudbreak would not sync the new version.

However, there could be other ways to do that, and maybe we should not be so restrictive.

Anyways, I removed it now.

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/model/upgrade/DistroXSyncCmV1Request.java
```diff
@@ -0,0 +1,29 @@
+package com.sequenceiq.distrox.api.v1.distrox.model.upgrade;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.sequenceiq.cloudbreak.doc.ModelDescriptions;
+
+import io.swagger.annotations.ApiModelProperty;
+
+public class DistroXSyncCmV1Request {
```
pdarvasi, line 10: nit: this is not being used anymore, anywhere

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
```diff
@@ -437,4 +438,17 @@ CertificatesRotationV4Response rotateAutoTlsCertificatesByName(@PathParam("name"
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = ROTATE_CERTIFICATES, nickname = "rotateAutoTlsCertificatesByCrn")
     CertificatesRotationV4Response rotateAutoTlsCertificatesByCrn(@PathParam("crn") String crn, @Valid CertificatesRotationV4Request rotateCertificateRequest);
+
+    @POST
+    @Path("{name}/sync_cm")
```
pdarvasi, line 443: thanks for the rename, pls rename Path, too

core-api/src/main/java/com/sequenceiq/distrox/api/v1/distrox/endpoint/DistroXV1Endpoint.java
```diff
@@ -437,4 +438,17 @@ CertificatesRotationV4Response rotateAutoTlsCertificatesByName(@PathParam("name"
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "syncs from distrox cluster CM the CM and parcel versions", produces = MediaType.APPLICATION_JSON, nickname = "syncDistroxCm")
+    DistroXSyncCmV1Response syncComponentVersionsFromCmByName(@PathParam("name") String name);
+
+    @POST
+    @Path("crn/{crn}/sync_cm")
```
pdarvasi, line 449: thanks for the rename, pls rename Path, too

pdarvasi:
Please resolve conflicting files, thank you!

PR #11408, opened 2021-09-15
CB-13980 - policy_validate API call is failing for mlx and liftie service input

gregito:


cloud-aws-common/src/main/java/com/sequenceiq/cloudbreak/cloud/aws/common/AwsCredentialConnector.java
```diff
@@ -154,14 +155,15 @@ public CredentialPrerequisitesResponse getPrerequisites(CloudContext cloudContex
...
         CDPServicePolicyVerificationResponses credentialStatus;
         Map<String, String> servicesWithPolicies = new HashMap<>();
-        services.forEach(service -> {
-            String policy = experiencePrerequisites.get(service.toUpperCase());
-            servicesWithPolicies.put(service, policy);
-        });
+        services.forEach(service -> experiencePrerequisites.keySet()
+                .stream()
+                .filter(AwsCredentialConnector::isPolicyServiceMatchesForName)
+                .findFirst()
```
biharitomi, line 165: Do we really need the `first` item form the filtered result?
The original logic added all the matched services policies. On the other hand if we had a unit test to cover these cases it would cover these cases...

biharitomi (changes requested):

doktoric (approved):

PR #11410, opened 2021-09-15
CB-13810: Validate the existence of key url provided by user for Azure cmk

juhi-09:
User provides the encryption key url during creation of environment to enable disk encryption with CustomerManagedKey. The existence of key on Azure cloud is checked as part of this PR.
The getKey method in AzureClient either returns the keyBundle if the key exists or returns null.

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
@@ -16,6 +16,7 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.microsoft.azure.keyvault.models.KeyBundle;
 import com.microsoft.azure.management.compute.implementation.DiskEncryptionSetInner;
 import com.microsoft.azure.management.resources.ResourceGroup;
 import com.sequenceiq.cloudbreak.cloud.EncryptionResources;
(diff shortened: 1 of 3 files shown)


cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/AzureEncryptionResources.java
```diff
@@ -46,6 +46,10 @@
 
     private static final Pattern ENCRYPTION_KEY_URL_VAULT_NAME = Pattern.compile("https://([^.]+)\\.vault.*");
 
+    private static final Pattern ENCRYPTION_VAULT_URI = Pattern.compile("(.*)\\/keys\\/.*");
+
+    private static final Pattern ENCRYPTION_KEY_NAME = Pattern.compile("https:.*\\/keys\\/([^.]+)\\/.*");
+
```
cegganesh84, line 52: Suggestion to consider : Is it possible to use single regex like `"(https://([^.]+).vault.[^/]+)/keys/([^/]+)/.*"` where group 1 will be vault uri, 2 will be vault name and 3 will be key name?

Also `\\/` I guess these double escaping is not needed, it should work without those.

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/client/AzureClient.java
```diff
@@ -1021,4 +1022,17 @@ public void deleteDiskEncryptionSet(String resourceGroup, String diskEncryptionS
         return azureClientCredentials.getAccesToken();
     }
 
+    public boolean checkEncryptionKeyExistenceOnCloud(String vaultUri, String keyName, String keyVaultName, String keyVaultResourceGroupName) {
+        return handleAuthException(() -> {
+            PagedList<KeyItem> keys = azure.vaults()
+                    .getByResourceGroup(keyVaultResourceGroupName, keyVaultName)
+                    .client()
+                    .listKeys(vaultUri);
```
cegganesh84, line 1030: Is my understanding right that the results are paginated and if the key is not present on the first page we will return `false`?

juhi-09: Thank you for pointing this out. Fixing this.

cegganesh84 (approved):
Just a nit: does not exists -> does not exist

PR #11413, opened 2021-09-15
CB-13692: Encrypt datalake disks with CMEK based on encryption key parameter set in create-gcp-environment call

juhi-09:
Encrypting the datalake disks with GCP CMEK passed during gcp environment creation.


cegganesh84 (approved):
LGTM

PR #11418, opened 2021-09-15
CB-14120 Until OPSAPS-61751 is not fixed we can not test cert rotatio…

doktoric:
…n on 7.2.11 or higher.


biharitomi (approved):

lajosrodek (approved):
LGTM

PR #11420, opened 2021-09-16
CB-13694 Encrypt cloud-sql disks with CMEK based on encryption key pa…

cegganesh84:
…rameter set in create-gcp-environment call

1. Tested and found one missing step in the user onboarding guide, that is the SQL service account needs encrypter and descrypter permission https://cloud.google.com/sql/docs/mysql/configure-cmek#grantkey
2. Just supporting KMS key encryption to begin with as that is the only mode of encryption supported via create-gcp-environment API.

Unit test and end to end test done.


redbeams/src/main/java/com/sequenceiq/redbeams/converter/spi/DBStackToDatabaseStackConverter.java
```diff
@@ -154,29 +156,42 @@ private boolean determineSslEnforcement(DBStack dbStack) {
...
         }
         return params;
     }
 
+    private DetailedEnvironmentResponse getDetailedEnvironmentResponse(DBStack stack) {
+        return measure(() -> environmentService.getByCrn(stack.getEnvironmentId()),
+                LOGGER, "Environment properties were queried under {} ms for environment {}", stack.getEnvironmentId());
+    }
+
+    private Map<String, Object> getMergedMap(Map<String, Object> params, Map<String, Object> cloudParams) {
```
juhi-09, line 185: Thank you for this change. I missed this previously with Azure changes.

juhi-09 (approved):
LGTM

PR #11422, opened 2021-09-16
CB-14141 Remove jcenter() as it is not working (403) and is already obsoleted

bergerdenes:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ gradle
 *.iws
 .shelf/
 
-.vscode
+.jpb/
 .nb-gradle-properties
 .nb-gradle
 *.iml
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -5,11 +5,9 @@ buildscript {
         maven { url = "$repoUrl" }
         maven { url = "$springRepoUrl" }
         mavenCentral()
-        jcenter()
     }
     dependencies {
         classpath "org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}"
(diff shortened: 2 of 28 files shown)


foldik (approved):
Cool that someone drops out garbage.

core/build.gradle
```diff
@@ -5,11 +5,9 @@ buildscript {
         maven { url = "$repoUrl" }
         maven { url = "$springRepoUrl" }
         mavenCentral()
-        jcenter()
     }
     dependencies {
         classpath "org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}"
-        classpath "com.smokejumperit.gradle.license:Gradle-License-Report:0.0.2"
         classpath "gradle.plugin.com.github.jk1:gradle-license-report:0.3.4"
```
lajosrodek, line 11: What about this one? Is this possibly related to the one removed above?

bergerdenes: it is related to the jcenter removal (and it is used nowhere btw)

lajosrodek: Let me clarify. I was referring to this line:
```
classpath "gradle.plugin.com.github.jk1:gradle-license-report:0.3.4"
```
Should it be also removed in addition to line 12?

bergerdenes: it is used in core

.gitignore
```diff
@@ -14,7 +14,7 @@ gradle
 *.iws
 .shelf/
 
-.vscode
```
lacikaaa, line 17: why is this removed?

bergerdenes: it was a duplicate of Line 8

PR #11430, opened 2021-09-17
CB-14109 Backup location base should be validated

bbihari:

Adds validation to the backup cloud storage.
Switches on the aws cloud storage validation for local development.


lturcsanyi (approved):
I think the new fields should be added to the CDP CLI as well.

PR #11433, opened 2021-09-17
CB-14171: Retry joining IPA if the DNS record is missing

jamisonbennett:
Fix the ipa registration for datalake and datahub nodes sot that
during the initial ipa client registration the operations are retried
if the DNS entry is not added. The existence of a DNS record (any
type of record) is validated. If there isn't a record then the salt
operation fails which will cause it to be retried. On the retry, the
ipa client is uninstalled then reinstalled and then the DNS record is
rechecked.

At a later point, the DNS record will be modified to ensure that the
A record exsits. However the later modification requires the a DNS
record (of any type) to already exist.

This was validated manually on a local deployment of cloudbreak using
a light duty datalake.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/sssd/template/join_ipa.j2
+++ b/orchestrator-salt/src/main/resources/salt/salt/sssd/template/join_ipa.j2
@@ -28,4 +28,21 @@ echo "$PW" | kinit {{ pillar['sssd-ipa']['principal'] }}
 
 ipa env
 
+HOSTNAME=$(hostname)
+
+# check the dns record 3 times with a 10 seconds interval (see CB-14171)
+for attempt in {1..3}
+do
+  sleep 10
+  echo "checking dns record hostname for ${HOSTNAME}, attempt ${attempt}"
+  if ipa dnsrecord-show {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}"; then
+    echo "ipa-client-install created the DNS record for ${HOSTNAME}"
+    break
+  fi
+  if [[ "$attempt" -eq 3 ]]; then
+    echo "Failed to add DNS record for ${HOSTNAME}"
+    false
+  fi
+done
+
 set +e


jamisonbennett:
I added the DO_NOT_MERGE tag because I specifically want @cegganesh84 to review this before we commit it and it is related to https://github.com/hortonworks/cloudbreak/pull/11077.

orchestrator-salt/src/main/resources/salt/salt/sssd/template/join_ipa.j2
```diff
@@ -28,4 +28,21 @@ echo "$PW" | kinit {{ pillar['sssd-ipa']['principal'] }}
...
+# check the dns record 3 times with a 10 seconds interval (see CB-14171)
+for attempt in {1..3}
+do
+  sleep 10
+  echo "checking dns record hostname for ${HOSTNAME}, attempt ${attempt}"
+  if ipa dnsrecord-show {{ pillar['sssd-ipa']['domain'] }}. "${HOSTNAME}"; then
+    echo "ipa-client-install created the DNS record for ${HOSTNAME}"
+    break
+  fi
+  if [[ "$attempt" -eq 3 ]]; then
```
lacikaaa, line 42: I think this could be `elif`, but sure the logic won't change

lacikaaa:
I'm not sure this is necessary. The condition to run this state looks like:
```
    - unless: test -f /var/log/ipa-join-executed && test -f /var/log/dnsrecord-add-executed
```
So if the DNS add script fails, because it cannot modify the records, when the highstate retries it would be the same effect as now. Am I missing something?

cegganesh84:
We could remove `    - unless: test -f /var/log/ipa-join-executed && test -f /var/log/dnsrecord-add-executed` it was there for this reason.

30 seconds is definitely too small for the sandbox tenant which is the only place where it will fail. Instead of doing the whole retry with uninstalling and installing again, we could increase the retries or retry interval. I would suggest an effective wait interval of 3 minutes.

jamisonbennett:
I didn't notice the `test -f /var/log/dnsrecord-add-executed` part of this check. I agree that this is not the fix we want, but let me think about what to do. I will probably pull down this PR.

PR #11434, opened 2021-09-18
CB-14152. Update cdp-telemetry desired version to avoid including roo…

oleewere:
…t folder in compressed bundles

### DIFF
--- a/common/src/main/resources/common-config.yml
+++ b/common/src/main/resources/common-config.yml
@@ -10,6 +10,6 @@ altus:
         endpoint: https://cloudera-dbus-stage.s3.amazonaws.com
 telemetry:
   cdpTelemetry:
-    desiredVersion: 0.4.14
+    desiredVersion: 0.4.15
   cdpLoggingAgent:
     desiredVersion: 0.2.11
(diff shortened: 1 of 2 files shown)


lnardai:
I don't see the urgency here, or did we introduce this as a bug in 2.47? If it was always working like this and there is no function impatirty we should only merge this to master.

oleewere:
@lnardai this is regression that is fixed in that 0.4.15 version (introduced in 0.4.14 - that version is available from 2.48), not worked like this, EDH won't be able to process bundles, so.diagnostics cannot be re-enabled without this in 2.48

lnardai (approved):

lnardai:
QA just started validation, but according to JIRA the ticket was opened on last Friday but it's not marked as blocking. So we knew the release was not ready. Now we need to explain to QA again why the version they validated is not the correct one. These are the things that make the release long.
Please link it to corresponding PROM ticket(PROM-6) ticket next time.

oleewere:
@lnardai note that i will need to enable the diagnostics entitlement on mow-int so re run tests will be needed anyway. (or will expect different results, as some of the test failures were expected because of disabling diagnostics)

lnardai:
Is this behind entitlement on Prod as well?

PR #11442, opened 2021-09-20
CB-18305: Make necessary changes to make the DL resizing transparent to experiences and UI

kkalvagadda1:
Here is the summary of the changes:

	⁃	Logic to mark the stack detached
	⁃	New CB API’s to detach the stack and to delete a detached stack.
	⁃	Changes to detach the stack.
	⁃	Change to the CB database indexes
	⁃	Changes to StackRepository.
	⁃	Changes to Stack View
	⁃	Make sure that CB does not de-register with cluster proxy when the stack being deleted is detached.

Test performed:

- Made sure that CRN and name are unchanged after resize is performed.
- Made sure that data lake can be stopped/deleted from CLI after resizing.
- Made sure that the environment can be stopped/started after resize.
- Made sure data lake can be upgraded after resize.


lajosrodek:
@kkalvagadda1 Nit: Could you please correct the Jira ticket ID in the commit msg (CB-13805 vs CB-18305). :smile:

kkalvagadda1:
> @kkalvagadda1 Nit: Could you please correct the Jira ticket ID in the commit msg (CB-13805 vs CB-18305). 😄

done :-)

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/detach/StackDetachService.java
```diff
@@ -0,0 +1,51 @@
...
+    @Inject
+    private Clock clock;
+
+    public Stack detachStack(Stack stack) {
+        // While detaching the stack, name of the stack is renamed to avoid conflicts.
+        MDCBuilder.buildMdcContext(stack);
+        stack.setDetached(clock.getCurrentTimeMillis());
+        Stack savedStack = measure(() -> stackService.save(stack),
+                LOGGER, "Stack save took {} ms for stack {}", stack.getName());
+        Optional<Cluster> cluster = clusterService.findOneWithLists(savedStack.getCluster().getId());
```
foldik, line 43: Can you please change it so it wont query the database like:

```sql
SELECT c FROM Cluster c LEFT JOIN FETCH c.gateway LEFT JOIN FETCH c.hostGroups LEFT JOIN FETCH c.containers LEFT JOIN FETCH c.components LEFT JOIN FETCH c.rdsConfigs WHERE c.id= :id"
```

just updates the name?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -41,50 +41,82 @@
...
 
+    @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig LEFT JOIN FETCH ig.instanceMetaData LEFT JOIN FETCH ig.template "
+            + "LEFT JOIN FETCH ig.securityGroup LEFT JOIN FETCH ig.instanceGroupNetwork WHERE s.resourceCrn= :crn " +
+            "AND s.workspace.id= :workspaceId AND s.type = :type AND s.detached is not null AND " +
+            SHOW_TERMINATED_CLUSTERS_IF_REQUESTED)
+    Optional<Stack> findByCrnAndWorkspaceIdWithListsAndDetached(@Param("crn") String crn, @Param("type") StackType type, @Param("workspaceId") Long workspaceId,
+            @Param("showTerminated") Boolean showTerminated, @Param("terminatedAfter") Long terminatedAfter);
+
     @Query("SELECT s FROM Stack s LEFT JOIN FETCH s.instanceGroups ig LEFT JOIN FETCH ig.instanceMetaData LEFT JOIN FETCH ig.template " +
             "LEFT JOIN FETCH ig.securityGroup LEFT JOIN FETCH ig.instanceGroupNetwork WHERE s.id= :id AND (s.type is not 'TEMPLATE' OR s.type is null)")
```
foldik, line 121: Shouldn't this be changed as well?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -207,23 +240,23 @@
     @Query("SELECT s.id as id, s.name as name FROM Stack s WHERE s.network = :network")
```
foldik, line 240: Shouldn't we change this as well?

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
```diff
@@ -207,23 +240,23 @@
...
 
-    @Query("SELECT s.environmentCrn FROM Stack s WHERE s.resourceCrn = :crn")
+    @Query("SELECT s.environmentCrn FROM Stack s WHERE s.resourceCrn = :crn AND s.detached = null")
     Optional<String> findEnvCrnByResourceCrn(@Param("crn") String crn);
 
-    @Query("SELECT s.resourceCrn as resourceCrn, s.environmentCrn as environmentCrn FROM Stack s WHERE s.resourceCrn IN (:crns)")
+    @Query("SELECT s.resourceCrn as resourceCrn, s.environmentCrn as environmentCrn FROM Stack s WHERE s.resourceCrn IN (:crns) AND s.detached = null")
     List<StackCrnView> findAllByResourceCrn(@Param("crns") Set<String> crn);
 
     @Query("SELECT s.id as id, s.name as name, s.resourceCrn as crn FROM Stack s WHERE s.datalakeCrn  = :crn")
```
foldik, line 262: What about these queries?

core/src/main/java/com/sequenceiq/cloudbreak/repository/snippets/ShowDetachedClustersSnippets.java
```diff
@@ -0,0 +1,9 @@
+package com.sequenceiq.cloudbreak.repository.snippets;
+
+public class ShowDetachedClustersSnippets {
```
foldik, line 3: I don't see the usage of this. Can you please remove this if not needed?

core/src/main/java/com/sequenceiq/distrox/v1/distrox/StackOperations.java
```diff
@@ -80,9 +81,14 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(StackOperations.class);
 
+    private static final String DELIMITER = "_";
```
foldik, line 84: You don't seem to use it. Please remove.

core/src/test/java/com/sequenceiq/cloudbreak/repository/StackRepositoryTest.java
```diff
@@ -32,15 +34,18 @@
 @EntityScan(basePackages = {
         "com.sequenceiq.cloudbreak.repository",
         "com.sequenceiq.cloudbreak.domain",
-        "com.sequenceiq.cloudbreak.workspace.repository",
+        "com.sequenceiq.cloudbreak.workspace.repository.workspace",
```
foldik, line 37: Why it did not cause issue previously?

datalake/src/main/resources/schema/app/20210701164636899_CB-13806_Make_necessary_changes_to_make_the_DL_resizing_transparent.sql
```diff
@@ -0,0 +1,20 @@
...
+
+
+-- //@UNDO
+-- SQL to undo the change goes here.
+ALTER TABLE sdxcluster DROP COLUMN clusterDisplayName;
+DROP INDEX IF EXISTS unq_index_sdxcluster_accountid_crn_deleted_is_not_null_detached_false;
+CREATE UNIQUE INDEX unq_index_sdxcluster_accountid_crn_deleted_is_not_null ON sdxcluster (accountid, crn, deleted) WHERE deleted IS NOT NULL;
+
+DROP INDEX IF EXISTS unq_index_sdxcluster_name_actid_deleted_is_not_null_detached_false;
+create unique index if not exists unq_index_sdxcluster_clustername_accountid_deleted_is_not_null
```
foldik, line 18: Can you make the sql keywords uppercase like `CREATE UNIQUE INDEX IF`

datalake/src/main/resources/schema/app/20210701164636899_CB-13806_Make_necessary_changes_to_make_the_DL_resizing_transparent.sql
```diff
@@ -0,0 +1,20 @@
...
+	on sdxcluster (clustername, accountid, deleted)
+	where (deleted IS NOT NULL);
```
foldik, line 20: Why the parenthesis?

datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxUpgradeServiceTest.java
```diff
@@ -55,7 +56,7 @@ public void setUp() {
     public void testUpdateRuntimeVersionFromCloudbreak() {
         when(sdxService.getById(1L)).thenReturn(sdxCluster);
         StackV4Response stackV4Response = getStackV4Response();
-        when(stackV4Endpoint.get(eq(0L), eq("test-sdx-cluster"), eq(Set.of()), anyString()))
```
foldik, line 58: Please use `eq(true)` or `eq(false)` instead of `anyBoolean()` since this is important here.

foldik (changes requested):
Going through this PR it was painful to see how you needed to drive a single basically somewhere long and somewhere boolean parameter through the entire code (API calls, DB queries, ...etc.). I think this change will make the maintenance of this code harder but that's the way it is for now.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
```diff
@@ -74,8 +74,13 @@ public void startStackDeletion(Long id, boolean forced) {
         SdxCluster sdxCluster = sdxService.getById(id);
         try {
             String initiatorUserCrn = ThreadBasedUserCrnProvider.getUserCrn();
+            if (sdxCluster.isDetached()) {
             ThreadBasedUserCrnProvider.doAsInternalActor(() ->
```
sodre90, line 78: indent please

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
```diff
@@ -428,6 +429,13 @@ public SdxCluster getByNameInAccount(String userCrn, String name) {
 
         newSdxCluster.setDatabaseAvailabilityType(sdxCluster.getDatabaseAvailabilityType());
         StackV4Request stackRequest = getStackRequest(shape, sdxCluster.isRangerRazEnabled(), null, cloudPlatform, sdxCluster.getRuntime(), null);
+        if (shape == SdxClusterShape.MEDIUM_DUTY_HA) {
+            // This is added to make sure the host name used by Light and Medium duty are not the same.
+            CustomDomainSettingsV4Request customDomainSettingsV4Request = new CustomDomainSettingsV4Request();
+            customDomainSettingsV4Request.setHostname(sdxCluster.getClusterName() + SDX_RESIZE_NAME_SUFFIX);
```
sodre90, line 435: maybe it won't cause any problem, but did you try it with attached DHs? I am not aware of if DHs joining with FQDN to the SDX master but if they do it then it can cause issue.

sodre90 (changes requested):

sodre90:
Anyway, what's happening if I delete an SDX while it is in a resizing process? Will it delete both stacks?

kkalvagadda1:
> Anyway, what's happening if I delete an SDX while it is in a resizing process? Will it delete both stacks?

I don't think. :-( It should be addressed.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/SdxService.java
kkalvagadda1: With attached DHs, remote data context has to be updated regardless. It should take care of updating the new configuration with new host names.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/detach/StackDetachService.java
kkalvagadda1: This can be optimized. will update.

core/src/main/java/com/sequenceiq/cloudbreak/repository/StackRepository.java
kkalvagadda1: I did not want to change the query that's already there. I re-used the same and extended it to include detached check.

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/detach/StackDetachService.java
```diff
@@ -0,0 +1,51 @@
...
+    private ClusterService clusterService;
+
+    @Inject
+    private StackService stackService;
+
+    @Inject
+    private Clock clock;
+
+    public Stack detachStack(Stack stack) {
+        // While detaching the stack, name of the stack is renamed to avoid conflicts.
```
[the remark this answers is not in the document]
kkalvagadda1: There is a type in the comment. It was with the cluster name. Even the cluster name should be unique. When a new stack is created with the same name it would fail because of the existing cluster entry.

core/src/main/resources/schema/app/20210907132719_CB-13806_Make_necessary_changes_to_make_the_DL_resizing_transparent.sql
```diff
@@ -0,0 +1,34 @@
+-- // CB-13806: Make necessary changes to make the DL resizing transparent.
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE stack ADD COLUMN IF NOT EXISTS detached BIGINT;
+
+ALTER TABLE stack
+    DROP CONSTRAINT IF EXISTS stack_crn_uq;
+
+ALTER TABLE stack
+    ADD CONSTRAINT stack_crn_uq UNIQUE (resourcecrn, detached);
```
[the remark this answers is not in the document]
kkalvagadda1: Changed the solution approach. Also changed the detached from timestamp to a boolean.

kkalvagadda1:
> > Anyway, what's happening if I delete an SDX while it is in a resizing process? Will it delete both stacks?
> 
> I don't think. :-( It should be addressed.

It is addressed with the new patch.

datalake/src/main/java/com/sequenceiq/datalake/service/sdx/ProvisionerService.java
```diff
@@ -74,8 +74,13 @@ public void startStackDeletion(Long id, boolean forced) {
...
         try {
             String initiatorUserCrn = ThreadBasedUserCrnProvider.getUserCrn();
-            ThreadBasedUserCrnProvider.doAsInternalActor(() ->
-                    stackV4Endpoint.deleteInternal(0L, sdxCluster.getClusterName(), forced, initiatorUserCrn));
+            if (sdxCluster.isDetached()) {
+                ThreadBasedUserCrnProvider.doAsInternalActor(() ->
+                        stackV4Endpoint.deleteDetached(0L, sdxCluster.getClusterName(), forced, initiatorUserCrn));
+            } else {
+                ThreadBasedUserCrnProvider.doAsInternalActor(() ->
+                        stackV4Endpoint.deleteInternal(0L, sdxCluster.getClusterName(), forced, initiatorUserCrn));
```
yg-gd, line 82: This seems a little strange. I'd hope that delete internal could be able to realize that is was supposed to delete a detached cluster

core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/stack/detach/StackDetachService.java
```diff
@@ -0,0 +1,75 @@
...
+
+    String getNameToDetach(String stackName) {
+        return stackName + DELIMITER + DETACHED;
+    }
+
+    String getNameToAttach(String stackName) {
+        return StringUtils.removeEnd(stackName, DELIMITER + DETACHED);
+    }
+
+    String getCrnToDetach(String resourceCrn) {
```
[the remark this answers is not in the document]
horadla23: I would be also careful with this solution, here are some thoughts of mine which should be considered:
- CRN is used across CDP and this solution would increase the length of it. Since we are using it in databases and also in URLs, I am not sure if every table and every solution is prepared for the increased length
- with this, searching for DL CRN would result to return with logs regarding current DL and also logs regarding detached DL, I am not sure if it is a pro or a con

I have seen Peter asked in another comment if `detached` field is needed or not. The question here is the same I think. We can have this modified CRN (and modified name) with delimiter or we have can only the detached flag and optimize queries and constraints on database level for that field. One of them possibly would make possible to distinguish detached and attached datalake in CB service and in datalake service.

lajosrodek:
For the record: The ticket number in the commit msg as well as PR title is wrong. This change belongs to [CB-13805](https://jira.cloudera.com/browse/CB-13805).

PR #11443, opened 2021-09-21
CB-14057 Created more user friendly error reporting about CM command …

foldik:
…failures. Add more detailed logging.


cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/task/AbstractClouderaManagerCommandListCheckerTask.java
```diff
@@ -54,19 +57,15 @@ protected boolean doStatusCheck(T pollerObject, CommandsResourceApi commandsReso
     private void validateApiCommandResults(List<ApiCommand> apiCommands, CommandsResourceApi commandsResourceApi) {
         List<ApiCommand> failedCommands = apiCommands.stream().filter(cmd -> !cmd.getSuccess()).collect(Collectors.toList());
         if (!failedCommands.isEmpty()) {
-            String message = failedCommands.stream().map(cmd -> createFailedCommandResultString(cmd, commandsResourceApi)).collect(Collectors.joining(","));
+            List<CommandDetails> failedCommandDetails = new LinkedList<>();
+            failedCommands.forEach(cmd -> failedCommandDetails.addAll(ClouderaManagerCommandUtil.getFailedOrActiveCommands(cmd, commandsResourceApi)));
+            String message = CommandDetailsFormatter.formatFailedCommands(failedCommandDetails);
             LOGGER.info(message);
+            LOGGER.debug("");
```
horadla23, line 64: pls remove

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/util/ClouderaManagerCommandUtil.java
```diff
@@ -46,23 +47,19 @@ private ClouderaManagerCommandApiErrorParserUtil() {
...
             Optional.ofNullable(apiCommand.getChildren()).map(ApiCommandList::getItems).orElse(List.of()).stream()
                     .filter(commandFailed())
                     .map(readCommand(commandsResourceApi))
-                    .forEach(cmd -> errors.addAll(getErrors(cmd, commandsResourceApi)));
-
-            if (errors.isEmpty() && StringUtils.isNotEmpty(apiCommand.getResultMessage())) {
-                String reason = String.format("Command [%s], with id [%.0f] failed: %s",
-                        apiCommand.getName(), apiCommand.getId(), apiCommand.getResultMessage());
-                errors.add(reason);
+                    .forEach(cmd -> failedCommands.addAll(getFailedOrActiveCommands(cmd, commandsResourceApi)));
```
horadla23, line 56: this looks like a recursive method call with a CM call in it, I would avoid this if possible, this could lead to enormous amount of call to CM

horadla23 (changes requested):
pls add an example to the description:
before the change, the message was like ...
after the change, the message is like ...

horadla23 (approved):
there is no easy solution for my comments, so I approve

PR #11445, opened 2021-09-21
CB-14185 Add entitlement and override checking for CCMv2 TLS mode

bergerdenes:
2 commits, the one is just nit fixes from a previous, already merged PR. 
Please check the 2 commits separately. Thanks.


environment-api/src/main/java/com/sequenceiq/environment/api/v1/environment/model/request/EnvironmentRequest.java
```diff
@@ -301,6 +304,22 @@ public void setEnvironmentServiceVersion(String environmentServiceVersion) {
...
+
+    public void setGcp(GcpEnvironmentParameters gcp) {
+        this.gcp = gcp;
+    }
+
+    public CcmV2TlsType getCcmV2TlsType() {
+        return ccmV2TlsType;
+    }
+
+    public void setCcmV2TlsType(CcmV2TlsType ccmV2TlsType) {
```
doktoric, line 319: should we just use tlstype? and in case of ccmv1 we just using two_way?

bergerdenes: only ccmv2, moreover it is applied only for the jumpgate version...

doktoric (approved):

topolyai5 (approved):

PR #11448, opened 2021-09-21
ENGESC-10456 Fix gen credential script to handle freeipa repair

lacikaaa:

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/kerberos/cm-keytab.sls
+++ b/orchestrator-salt/src/main/resources/salt/salt/kerberos/cm-keytab.sls
@@ -18,8 +18,14 @@ configure_cm_principal:
 replace_ipa_env_host_to_server:
   file.replace:
     - name: /opt/cloudera/cm/bin/gen_credentials_ipa.sh
-    - pattern: "ipa env host"
-    - repl: "ipa env server"
+    - pattern: "IPA_HOST=.*"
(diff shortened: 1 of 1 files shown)


hreeve-cloudera:
I checked out the new ipa commands, and it all looks good to me. Thanks for jumping on this so quickly!

jamisonbennett (approved):

hreeve-cloudera (changes requested):
@lacikaaa I was taking another look at the gen_credentials_ipa.sh script, and I realized this isn't a complete solution. This add lgoic to explicitly define the server in the `ipa-getkeytab --principal=$PRINCIPAL --keytab=$KEYTAB_PATH` command. However, there are multiple `ipa` commands that occur earlier in the script where we're not specifying the server to connect to. I think if there are multiple FreeIPA instances, then we could run into trouble if the `ipa` commands pick different servers to talk to. So we either need to specify the server explicitly in all `ipa` commands, or update the `/etc/ipa/default.conf` file after a FreeIPA repair to a valid server so the `ipa` commands aren't randomly picking on each time.

I've opened CB-14212 to describe the problem in detail, and I've included the steps where I think this fix would be insufficient. Take a look and let me know if it makes sense.

jamisonbennett:
I agree with @hreeve-cloudera

lacikaaa:
right, thanks @hreeve-cloudera , I'm going to update the `default.conf` and see how it works. In the meanwhile I'll close this

PR #11449, opened 2021-09-21
CB-7858 CB should provide a link to CM command UI in case of timeout …

horadla23:
…or failure

- removing redundant cm command listener task classes
- generalize success message of cm command listener tasks
- generalize timeout handling of cm command listener tasks
- extracting CM server fqdn and knox based CM basepath and provide link to the failed command, if possible
- sending command link with a notification, thus user can use it to check the command on CM UI


horadla23:
Screenshot from 2021-09-23 18-08-35

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CmCommandLinkProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sequenceiq.cloudbreak.service.ServiceEndpointCollector;
+import com.sequenceiq.cloudbreak.util.StackUtil;
+
+@Service
+public class CmCommandLinkProvider {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CmCommandLinkProvider.class);
+
+    @Value("${distrox.gateway.topology.name}")
+    private String defaultTopologyName;
```
lajosrodek, line 28: Is this really relevant for DH only? If not (i.e. it also applies for DL stacks), I wonder if the `distrox` category is indeed appropriate.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CmCommandLinkProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.stereotype.Service;
+
+import com.sequenceiq.cloudbreak.api.endpoint.v4.stacks.response.cluster.gateway.topology.ClusterExposedServiceV4Response;
+import com.sequenceiq.cloudbreak.api.service.ExposedServiceCollector;
+import com.sequenceiq.cloudbreak.domain.stack.Stack;
+import com.sequenceiq.cloudbreak.service.ServiceEndpointCollector;
+import com.sequenceiq.cloudbreak.util.StackUtil;
+
+@Service
+public class CmCommandLinkProvider {
```
lajosrodek, line 23: Please add UT if possible. :smile:

lajosrodek (commented):
Have not checked the whole PR, just two minor comments.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CmCommandLinkProvider.java
horadla23: no, it is relevant for datalake too and it is applied for datalake too, the config name is misleading, but I do not want to change it in this PR

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/CmCommandLinkProvider.java
horadla23: sure, good point

core/src/main/java/com/sequenceiq/cloudbreak/service/events/DefaultCloudbreakEventService.java
```diff
@@ -108,6 +121,19 @@ public void fireCloudbreakInstanceGroupEvent(Long stackId, String eventType, Str
         fireEventWithPayload(stackId, eventType, resourceEvent, eventMessageArgs, eventMessage, instanceGroupName);
     }
 
+    @Override
+    public void fireClusterManagerEvent(Stack stack, ResourceEvent resourceEvent, String eventName, Optional<BigDecimal> clusterManagerEventId) {
+        if (clusterManagerEventId.isPresent()) {
+            Optional<String> link = cmCommandLinkProvider.getCmCommandLink(stack, clusterManagerEventId.get().toString());
+            if (link.isPresent()) {
```
lturcsanyi, line 128: Is it intentional not to send event if no link is provided? The _Cloudera Manager command failed ..._ part of error message seems still valid for me in this case.

horadla23: Yes, this is intentional. Unfortunately I had to send an extra notification with the link, there is always a notification about the failure, but at that time the command id is not present, so I had to add the logic to send an extra event with the link, so without the link the extra notification isn't needed.

horadla23: Unfortunately there are various error handling logics implemented for exceptions from CM command pollers in various places (flow, service layer, etc.), so there is no easy way to add the link to the existing events, as far as I have seen it in the code.

cluster-cm/src/main/java/com/sequenceiq/cloudbreak/cm/polling/ClouderaManagerCommandPollerObject.java
```diff
@@ -7,15 +7,12 @@
...
-    private final Stack stack;
-
     private final ApiClient apiClient;
 
     private final BigDecimal id;
 
     public ClouderaManagerCommandPollerObject(Stack stack, ApiClient apiClient, BigDecimal id) {
         super(stack, apiClient);
-        this.stack = stack;
         this.apiClient = apiClient;
```
foldik, line 16: You could have add some null checks here.

horadla23: I am not sure if it is a bad behaviour if the stack is null, so I checked the stack for null in the event service

horadla23: https://github.com/hortonworks/cloudbreak/pull/11449/files#diff-03a2afa3da231e8ffae47ea036184e2b9590359d2bb6b767abc3309864e38e88R126

foldik (approved):
LGTM

lturcsanyi (approved):

PR #11452, opened 2021-09-22
CB-14216 missing converters now working. Also deleted the last conver…

doktoric:
…sionservices.

### DIFF
--- a/environment/src/main/java/com/sequenceiq/environment/platformresource/v1/CredentialPlatformResourceController.java
+++ b/environment/src/main/java/com/sequenceiq/environment/platformresource/v1/CredentialPlatformResourceController.java
@@ -6,14 +6,12 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
-import javax.inject.Named;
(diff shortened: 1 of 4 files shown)


environment/src/main/java/com/sequenceiq/environment/platformresource/v1/CredentialPlatformResourceController.java
```diff
@@ -59,23 +57,67 @@
...
-    @Named("conversionService")
-    private ConversionService convertersionService;
-
     @Inject
     private PlatformParameterService platformParameterService;
 
     @Inject
     private CommonPermissionCheckingUtils commonPermissionCheckingUtils;
 
+    @Inject
```
foldik, line 85: What was changed that now we can't use the conversion service but we have to import all converters separately?

doktoric: and you need to invoke them directly.

biharitomi (approved):

foldik (approved):
LGTM

PR #11455, opened 2021-09-22
CB-14212 IPA server is not updated on CM nodes after FreeIPA repair

lacikaaa:
`/etc/ipa/default.conf` should always point to a valid FreeIPA server, so every `ipa` command would reach the same server
- during provisioning and when there is a change on FMS side the `default.conf` would be updated to a current server
- the choosing logic queries FMS for all instances and filters for ones in `CREATED` state and choose the first in alphabetical order
- if there is no instance in created choose the first one
- the file only edited if the FreeIPA join is successful and FMS is returning a response which is not 404

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
@@ -24,6 +24,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
@@ -102,6 +103,7 @@ import com.sequenceiq.cloudbreak.service.blueprint.ComponentLocatorService;
 import com.sequenceiq.cloudbreak.service.cluster.ClusterApiConnectors;
 import com.sequenceiq.cloudbreak.service.cluster.flow.recipe.RecipeEngine;
 import com.sequenceiq.cloudbreak.service.environment.EnvironmentConfigProvider;
+import com.sequenceiq.cloudbreak.service.freeipa.FreeipaClientService;
 import com.sequenceiq.cloudbreak.service.hostgroup.HostGroupService;
 import com.sequenceiq.cloudbreak.service.idbroker.IdBrokerService;
 import com.sequenceiq.cloudbreak.service.proxy.ProxyConfigProvider;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
```diff
@@ -451,23 +459,52 @@ private VirtualGroupRequest getVirtualGroupRequest(String virtualGroupsEnvironme
...
+                    .flatMap(ig -> ig.getMetaData().stream());
+            InstanceMetaDataResponse instanceMetaDataResponse = instanceMetaDataStream
+                    .filter(im -> InstanceStatus.CREATED == im.getInstanceStatus())
+                    .min(Comparator.comparing(InstanceMetaDataResponse::getDiscoveryFQDN))
+                    .orElseGet(() -> instanceMetaDataStream
+                            .min(Comparator.comparing(InstanceMetaDataResponse::getDiscoveryFQDN))
+                            .orElse(null));
+            LOGGER.info("Chosen instance for default FreeIPA server: {}", instanceMetaDataResponse);
+            if (instanceMetaDataResponse == null) {
+                LOGGER.debug("No FreeIPA instance available");
```
jamisonbennett, line 499: [nit] maybe this should be "No healthy FreeIPA instance available".

jamisonbennett (approved):

core/src/main/java/com/sequenceiq/cloudbreak/core/bootstrap/service/host/ClusterHostServiceRunner.java
lacikaaa: as we include unhealthy ones if we can't find a healthy one this would be not true :)

hreeve-cloudera:
The changes look good to me. Just to make sure, is the logic to update the default.conf file going to be triggered all nodes when a FreeIPA repair is done, or just the new FreeIPA nodes?

lacikaaa:
> The changes look good to me. Just to make sure, is the logic to update the default.conf file going to be triggered all nodes when a FreeIPA repair is done, or just the new FreeIPA nodes?

it would be triggered on cluster instances, DL and DH, not FreeIPA. I haven't checked FreeIPA instances regarding this config file, but I don't think it's really necessary for us to modify that file.

This flow is triggered on freeipa repair: https://github.com/hortonworks/cloudbreak/blob/CB-2.48.0/core/src/main/java/com/sequenceiq/cloudbreak/core/flow2/cluster/config/update/PillarConfigUpdateActions.java#L25
And also when a cluster is started. The original motivation was to update unbound configuration after repair, but it would suit our case too.

Please note this change would make effect only for new or upgraded clusters.

hreeve-cloudera (approved):
@lacikaaa Makes sense to me. I also don't think we need to update the config on the FreeIPA instances. I just wasn't entirely sure when the ClusterHostServiceRunner was triggered. Thanks for working on this!

PR #11456, opened 2021-09-22
CB-14143 Cert renewal pops up randomly

lacikaaa:
- cert health check should be considered unhealthy only if CM reports `BAD` or `CONCERNING` state
- host health reason summary includes only the reason in case it's unhealthy
- cert check reason contains the original health summary from CM


horadla23 (approved):
LGTM

PR #11458, opened 2021-09-22
CB-14223 Install NiFi Registry Gateway on all hosts with NiFi service…

vbrodetskyi:
…. Added NiFi Registry Gateway role to Flow templates 7.2.12. New role will be installed on hosts with NiFi, we need that to get NiFi Registry url from NiFi.


vbrodetskyi:
@doktoric could you please review/merge this PR? We need it asap, because it's release blocker. Thank you.

PR #11460, opened 2021-09-23
CB-14183 Add a variant field to InstanceMetaData. We need to know tha…

topolyai5:
…t is the instance is AWS_NATIVE or not. It is necessary for the downscale. We will downscale the CF instances first and we will create a new one as a native way.

### DIFF
--- a/core/src/main/resources/schema/app/20210922084527_CB-14183_alter_instance_metadata_table_add_variant_column.sql
+++ b/core/src/main/resources/schema/app/20210922084527_CB-14183_alter_instance_metadata_table_add_variant_column.sql
@@ -1,0 +1,17 @@
+-- // CB-13685 alter image table add date column
+-- Migration SQL that makes the change goes here.
+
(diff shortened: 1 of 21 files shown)


core/src/main/resources/schema/app/20210922084527_CB-14183_alter_instance_metadata_table_add_variant_column.sql
```diff
@@ -0,0 +1,9 @@
+-- // CB-13685 alter image table add date column
+-- Migration SQL that makes the change goes here.
+
+ALTER TABLE instancemetadata ADD COLUMN IF NOT EXISTS variant VARCHAR(255);
```
doktoric, line 4: should we also update the current records?

topolyai5: Nope, because we should know the cloudplatform for the default value. I think easier to handle if this value is null. If null is the same as the cloudpaltform name (or anything we want to want)

doktoric (approved):

PR #11465, opened 2021-09-23
CB-13844 only use the v45 client when the CM server is already upgraded

keyki:
The first step of the upgrade process is to deactivate and remove the unused parcels.
However, when the execution get's there we've already updated the database to use
the new image and new components including the CM server version. Due to this it
was incorrectly determined that the v45 API is available. As part of this PR I'm
reverting the change and always use the v31 API except when the post upgrade command
needs to be invoked.


biharitomi (approved):
LGTM

doktoric (approved):

PR #11466, opened 2021-09-23
CB-14148 Validate user and resource account id in authz

foldik:
Validate if the user's account is the same as the requested resources. If not throw an access denied exception. The exception intentionally doesn't contain information about the user crn or the resources crn since it can be a programming error and we don't want to leak resource crns to users in other accounts. When the account validation fails we log it on WARN level.


horadla23 (approved):
LGTM

PR #11469, opened 2021-09-23
CDPD-27965 Atlas is up but Knox throws service connectivity error trying to connect to Atlas

hreeve-cloudera:
Adds the "atlas_ssl_exclude_protocols" : "TLSv1,TLSv1.1" setting to all SDX blueprints.
Tested via creating two datalakes (one 7.2.12, one 7.2.2) and verifying the service configuration
was set as expected, and the Atlas UI was accessible via Knox proxy.

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.1.0/cdp-sdx.bp
+++ b/core/src/main/resources/defaults/blueprints/7.1.0/cdp-sdx.bp
@@ -114,6 +114,10 @@
               {
                 "name": "atlas_authentication_method_file",
                 "value": "false"
+              },
+              {
+                "name": "atlas_ssl_exclude_protocols",
+                "value": "TLSv1,TLSv1.1"
               }
             ],
             "refName": "atlas-ATLAS_SERVER-BASE",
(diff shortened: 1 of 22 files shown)


sxxgrc (approved):

aakanser (approved):
Looks good. It makes sense to exclude TLSv1 and v1.1 since they are deprecated.

Should we also do the same for 7.2.13 now in this change? How about the sdx-micro blueprints?

hreeve-cloudera:
> Looks good. It makes sense to exclude TLSv1 and v1.1 since they are deprecated.
> 
> Should we also do the same for 7.2.13 now in this change? How about the sdx-micro blueprints?

Thanks for pointing that out. I realize now I had made this branch off an old version of master, that didn't have the 7.2.13 or micro duty blueprints. Not including them wouldn't break anything, since the problem this is solving was actually fixed in CM 7.2.6. But I included all blueprints for consistency and to avoid regression problems, so I should also include 7.2.13 and micro duty. Updating with those now.

PR #11470, opened 2021-09-23
CB-14173: Add intermediary `clouderamanager` for backup and restore sls.

brycederriso:
Closes [CB-14173](https://jira.cloudera.com/browse/CB-14173).

I added an additional property to look up the values needed for `.pgpass`.

However, this only uncovered another issue on the DH -- we'll need to add the `smm` table to the `.pgpass` file in order to connect to that database. We can do that, but this obviously doesn't scale long term.

This is a limitation to my original design, and means that we'll probably need to rethink how credentials are supplied to the backup/restore scripts as they connect to DB instances.


sxxgrc (approved):
Glad the bug can be closed and that you were able to figure this out, though saddened that the code is becoming even more wordy and convoluted, but it is what it is haha.

Thanks for the change Bryce.

PR #11475, opened 2021-09-27
CB-14234: Remove unnecessary components from the database in the begining of the upgrade

tiborpopovics:
Before this change when the upgrade flow started we removed the unused parcels from the CM server but those parcels are left in the database. This caused an inconsistent state because the Cloudbreak wanted to activate parcels which are not downloaded to the cluster.  
To solve this issue we need to remove those unused parcels from the clustercomponent table too.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterComponentUpdater.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterComponentUpdater.java
@@ -82,4 +82,23 @@ public class ClusterComponentUpdater {
         LOGGER.debug("Removing unused components: {}", unusedCdhProductDetails);
         clusterComponentConfigProvider.deleteClusterComponents(unusedCdhProductDetails);
     }
+
+    public void removeUnusedCdhProductsFromClusterComponents(Long clusterId, Set<ClusterComponent> clusterComponentsByBlueprint) {
+        Set<ClusterComponent> clusterComponentsFromDb = clusterComponentConfigProvider.getComponentsByClusterId(clusterId);
+        Set<ClusterComponent> componentsToDelete = getUnusedComponents(clusterComponentsByBlueprint, clusterComponentsFromDb);
+        if (componentsToDelete.isEmpty()) {
+            LOGGER.debug("There is no cluster component to be deleted.");
+        } else {
+            LOGGER.debug("The following cluster components will be deleted: {}", componentsToDelete);
+            clusterComponentConfigProvider.deleteClusterComponents(componentsToDelete);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


keyki (approved):
Tested, but haven't checked the code.

core/src/test/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterComponentUpdaterTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void testRemoveUnusedCdhProductsFromClusterComponentsWhenThereIsOneComponentToDelete() {
+        Set<ClusterComponent> clusterComponentsByBlueprint = createComponents(Set.of(CDH));
+        Set<ClusterComponent> clusterComponentsFromDb = createComponents(Set.of(CDH, FLINK));
+
+        when(clusterComponentConfigProvider.getComponentsByClusterId(CLUSTER_ID)).thenReturn(clusterComponentsFromDb);
+
+        underTest.removeUnusedCdhProductsFromClusterComponents(CLUSTER_ID, clusterComponentsByBlueprint);
+
+        verify(clusterComponentConfigProvider).getComponentsByClusterId(CLUSTER_ID);
+        verify(clusterComponentConfigProvider).deleteClusterComponents(any());
```
gergopapi2, line 73: Here testing that the correct component is deleted could be useful

gergopapi2, line 73: It could make sense to check that the right ClusterComponent is removed.

gergopapi2 (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/ClusterComponentUpdater.java
```diff
@@ -82,4 +82,23 @@ private void removeUnusedComponents(Set<Component> targetComponents, Set<Cluster
         LOGGER.debug("Removing unused components: {}", unusedCdhProductDetails);
         clusterComponentConfigProvider.deleteClusterComponents(unusedCdhProductDetails);
     }
+
+    public void removeUnusedCdhProductsFromClusterComponents(Long clusterId, Set<ClusterComponent> clusterComponentsByBlueprint) {
+        Set<ClusterComponent> clusterComponentsFromDb = clusterComponentConfigProvider.getComponentsByClusterId(clusterId);
+        Set<ClusterComponent> componentsToDelete = getUnusedComponents(clusterComponentsByBlueprint, clusterComponentsFromDb);
+        if (componentsToDelete.isEmpty()) {
+            LOGGER.debug("There is no cluster component to delete.");
```
pdarvasi, line 90: nit: to be deleted

gergopapi2 (approved):

keyki:
@tiborpopovics can you please extend the commit message with more details about why this change is needed and how does it fix the problem?

lnardai (approved):
LGTM!

PR #11478, opened 2021-09-28
CB-14272: Avoid limiting the incoming connection while performing backup.

kkalvagadda1:
When the backup is performed with input not to close DB connections, we should not only avoid closing existing connections we should also avoid blocking new connections as well.

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
@@ -123,16 +123,18 @@ limit_incomming_connection() {
 
 backup_database_for_service() {
   SERVICE="$1"
-  limit_incomming_connection $SERVICE 0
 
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -123,16 +123,18 @@ limit_incomming_connection() {
 
 backup_database_for_service() {
   SERVICE="$1"
-  limit_incomming_connection $SERVICE 0
 
   if [[ "$CLOSECONNECTIONS" == "true" ]]; then
+    limit_incomming_connection $SERVICE 0
```
brycederriso, line 128: **nit:** Fix `incomming` -> `incoming`
```suggestion
    limit_incoming_connection $SERVICE 0
```

brycederriso (approved):
Change looks good to me, thanks Kalyan!

I have a nit on spelling, `incomming` -> `incoming`, but it's not a blocker -- it's been that way for a while.

PR #11485, opened 2021-09-29
CB-14274 Integrate with Telemetry solution for GCP CMEK

juhi-09:
CB-14274 Integrate with Telemetry solution for GCP CMEK
Adding a new property to CDPEnvironmentDetails, CDPEnvironmentGcpDetails which has entity bool resourceEncryptionEnabled, this is set only when encryptionKey is present.
As encryptionKey is present only when "CDP_CB_GCP_DISK_ENCRYPTION_WITH_CMEK" is enabled and encryptionKey is provided during environment creation, checking its presence is sufficient to know if user is using "GCP SSE with CMEK".

Thunderhead PR- https://github.infra.cloudera.com/thunderhead/thunderhead/pull/6016


cegganesh84 (approved):
LGTM

PR #11488, opened 2021-09-29
CB-14129 Update tag specifications

agov:
Updated AWS, Azure, GCP and OpenStack tag specifications:
- Reflect the latest restrictions (see links in Jira)
- Fixed '-' issue where it was interpreted as a range
- Leading and trailing spaces are restricted

### DIFF
--- a/environment/src/test/java/com/sequenceiq/environment/environment/validation/validators/TagValidatorTest.java
+++ b/environment/src/test/java/com/sequenceiq/environment/environment/validation/validators/TagValidatorTest.java
@@ -26,18 +26,11 @@ import com.sequenceiq.cloudbreak.cloud.azure.validator.AzureTagValidator;
 import com.sequenceiq.cloudbreak.cloud.init.CloudPlatformConnectors;
 import com.sequenceiq.cloudbreak.cloud.model.Platform;
 import com.sequenceiq.cloudbreak.cloud.model.Variant;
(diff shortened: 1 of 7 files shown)


environment/src/test/java/com/sequenceiq/environment/environment/validation/validators/TagValidatorTest.java
```diff
@@ -35,7 +35,7 @@
 
     public static final String BAD_KEY = "azureprefix";
```
horadla23, line 36: I would extend this or I would add more different bad values to cover more cases mentioned in links in the ticket.

agov: I agree, also not only Azure should be tested. But I'm not sure if it should be within the scope of this ticket.

horadla23: I think if we don't extend it here in this PR, no one will later.

agov: Updated tests as agreed.

horadla23: Thank you!

horadla23 (approved):

PR #11496, opened 2021-09-30
CB-13986 Upscaled Azure VMs are not rolled back in case of CloudExcep…

lturcsanyi:
…tion

- Retry deployment if Azure cannot start VMs in time, but there is a chance it has not yet failed.
- Rollback if other CloudException occurs, e.g. quota limit
- Use the Deployment object to rollback resources.
- Cleanup requested instances (without FQDN) if upscale fails

### DIFF
--- a/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/upscale/AzureUpscaleService.java
+++ b/cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/upscale/AzureUpscaleService.java
@@ -8,6 +8,7 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 
 import org.apache.commons.collections4.ListUtils;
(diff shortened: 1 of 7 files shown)


lnardai (approved):

cloud-azure/src/main/java/com/sequenceiq/cloudbreak/cloud/azure/upscale/AzureUpscaleService.java
```diff
@@ -78,10 +80,12 @@
         List<CloudResource> templateResources = new ArrayList<>();
         List<CloudResource> osDiskResources = new ArrayList<>();
 
+        DateTime now = DateTime.now();
```
sodre90, line 83: can we please rename it with something more meaningful word? :D

sodre90:
if "now" is renamed, it will be good for me

PR #11497, opened 2021-09-30
Revert "CB-12780 - Audit API and new event API should be authorized"

horadla23:
This reverts commit a6070113c8b5b0b33e2ab36eba8d16e4ffbf00ee.


lnardai (approved):
This will probably end up with Compile errors on master. Need to double check.

foldik (changes requested):
EventAuthorizationUtils is used here:

https://github.com/hortonworks/cloudbreak/blob/CB-2.48.0/structuredevent-service-cdp/src/main/java/com/sequenceiq/cloudbreak/structuredevent/rest/controller/CDPStructuredEventV1Controller.java#L38

foldik (approved):
LGTM

PR #11515, opened 2021-10-04
CB-14343 Forbid anonymous access to FreeIPA LDAP server

lacikaaa:
- remove anonymous acces from health agent script as it seems it's not used anymore
- add a new script and state for disabling anonymous access to common install state file, as it has to be done for every instance separately
- verified by installing HA FreeIPA using `ldapsearch`
- tested by provisioning DL and DH, and scaling DH


jamisonbennett (commented):
I investigated the health checks a little deeper to see if they are using this. I have come to the conclusion that we will need to make some changes to the health checks.

The LDAP access is used in the new health check for the following:
1. The existence of replication agreement is used to set the `is_part_of_cluster` variable which is used by the service check plugin to validate the health of cluster services (currently the list is empty). It is also used by the port check plugin to check the cluster service (currently the list is empty).
2. The replication check plugin validates replications. However, this runs in the cluster health check only (we currently do not call this API but it does exist).

Since the API isn't called programmatically (as far as I can tell) and there is no meaningful difference right now, this change will not break anything. However, that doesn't mean it should be done as it will incur technical debt. If we commit this, then I want the health checks to be updated to remove the stuff that will no longer work.

I want to also get the opinion of @wonderslug on this. Do we remove the anonymous access and remove the dependent health checks (as mentioned above they are not being actally invoked)?

jamisonbennett (approved):
@wonderslug and I discussed this. We think there might be a small risk in a workload depending on anonymous ldap to get config, but it should be caught in the integration testing. We also think that if we commit this then lets just create a JIRA to remove the cluster mode health checks.

lacikaaa:
> @wonderslug and I discussed this. We think there might be a small risk in a workload depending on anonymous ldap to get config, but it should be caught in the integration testing. We also think that if we commit this then lets just create a JIRA to remove the cluster mode health checks.

I left in place the part for replication anon access

PR #11518, opened 2021-10-04
CB-14364 Remove Openstack because we dont support that anymore. Remov…

doktoric:
…ed from converters, cloud modules, tests. The api is still there as deprecated because of backward compatibility.


foldik (approved):
LGTM

core/src/main/resources/schema/app/20200824205107_CB-1068_multi_az.sql
```diff
@@ -41,7 +41,7 @@ UPDATE instancegroupnetwork SET attributes=network.attributes
 -- Search subnet-id
 UPDATE instancegroupnetwork SET subnetid=instancegroupnetwork.attributes::json->>'subnetId';
 
--- If AWS, AZURE, GCP, OPENSTACK, MOCK
+-- If AWS, AZURE, GCP, MOCK
```
foldik, line 44: If we don't do hash based comparison of migration scripts than it's fine.

PR #11519, opened 2021-10-04
DISTX-705 Set Azure buffer dir to Yarn local dir

amitagarwal06:
[DISTX-705](https://jira.cloudera.com/browse/DISTX-705)

This is backport to CB-2.48.0 for the change already in master.

Fixes for CDPD-13652 in 7.2.12 introduced disk buffering for abfs streams. This also requires cleaning up the buffer directory as part of failed yarn jobs. Hence configuring following in 7.2.12 DE templates
fs.azure.buffer.dir=${env.LOCAL_DIRS:-${hadoop.tmp.dir}}/abfs

### DIFF
--- a/core/src/main/resources/defaults/blueprints/7.2.12/cdp-data-engineering-ha.bp
+++ b/core/src/main/resources/defaults/blueprints/7.2.12/cdp-data-engineering-ha.bp
@@ -33,7 +33,7 @@
           },
           {
             "name": "core_site_safety_valve",
-            "value": "<property><name>fs.s3a.buffer.dir</name><value>${env.LOCAL_DIRS:-${hadoop.tmp.dir}}/s3a</value></property><property><name>fs.s3a.committer.name</name><value>directory</value></property>"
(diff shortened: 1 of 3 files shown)


amitagarwal06:
@shwethags @cegganesh84 Can one of you pls approve this?
@lnardai This change is already present in master but requires in CB-2.48 too.

cegganesh84 (approved):
LGTM. The raven team has the authority on merging this.

doktoric:
I think this should be only added in case of azure right ?

foldik:
Is this really needed in 2.48? We already took lot of fixes which delayed this release. I would say this can go in 2.48 as a hotfix after the 2.48 version reached prod (if needed).

lturcsanyi:
I agree with @foldik, this does not seem a must have in 2.48.

shwethags:
@foldik  @lturcsanyi 
This is required for 7.2.12 release. Since CB 2.49 won't be release by then, including this in 2.48. 

This doesn't have to block initial CB 2.48 release. But lets merge it so that it can be part of the next build that 2.48 picks up

lnardai (approved):

lnardai:
Discussed with the team, since we need to apply 2.47 hotfix, we'll include this as well.

PR #11522, opened 2021-10-05
CB-14371 PrefferedSubnetIds must be a list when query detailed respon…

doktoric:
…ses.


biharitomi (approved):
LGTM

PR #11524, opened 2021-10-05
CB-14353 disable direct CM UI debug path in NGiNX configuration.

biharitomi:


lacikaaa:
can we do it by moving those parts into separate files and use include directive, so we could comment only that line out? I don't like seeing half of the file commented

biharitomi:
I don't think it's worth to move it and have a test environment and cluster to test that refactor.
On the other hand the conditional rewrite logic needs to be there to have evaluated before the location section.

So if you don't like it this way I can remove the entire part and if somebody would like to use this backdoor then needs to look up for it in a previous line of CB. 
cc: @akanto @keyki @doktoric what do you think?

moresandeep (approved):
I tested the patch in a test cluster and I don't see any 502 or 404s anymore :)

akanto:
I also don't think that refactoring is worth the effort, please go ahead and delete the commented-out lines completely.

PR #11525, opened 2021-10-05
CB-14308 Image catalog change implementation for Datahub. In use imag…

pkedvessy:
…e should be available in the new image catalog and there should not be in progress flow to apply the update.


pkedvessy:
retest this please

pkedvessy:
retest this please

Bajzathd (approved):

PR #11528, opened 2021-10-05
CB-14378 Do no configure ephemeral storage to store customer data

daszabo:
In an earlier commit, volume template objects were created, whenever
magnetic storage was requested but the instance type supported them,
in order to implicitly request the maximum number of ephemeral
volumes from AWS. This had the unintended side effect, that these
ephemeral volumes got configured for various roles like HDFS,
because the logic in RoleConfigProvider classes counted them in
while generating directory paths to configure.

In this change, volume template objects are no longer created additionally,
when the user selects magnetic storage. This restores the original
behaviour, meaning either EBS volumes are configured or ephemeral,
if EBS is not requested.

In order to not break the ephemeral volume caching feature, and
still request the maximum number of ephemeral volumes implicitly,
a temporaryStorageCount field was introduced in the cloud model
class InstanceTemplate, which is used in the AWS module to properly
fill the CloudFormation template.

The ephemeral disk caching e2e test was also extended to check the
HDFS configuration on a 'control' datahub cluster, where the feature
is not enabled.


lnardai:
Please make sure to merge this to 2.48 ASAP since it delays the release.

core/src/main/java/com/sequenceiq/cloudbreak/service/stack/StackStopRestrictionService.java
```diff
@@ -22,7 +22,7 @@ public StopRestrictionReason isInfrastructureStoppable(String cloudPlatform, Set
         if (SUPPORTED_CLOUD_PLATFORM.equals(cloudPlatform)) {
             for (InstanceGroup instanceGroup : instanceGroups) {
                 if (instanceGroup.getTemplate().getVolumeTemplates().stream()
-                        .allMatch(volume -> AwsDiskType.Ephemeral.value().equals(volume.getVolumeType()))) {
+                        .anyMatch(volume -> AwsDiskType.Ephemeral.value().equals(volume.getVolumeType()))) {
                     reason = StopRestrictionReason.EPHEMERAL_VOLUMES;
                     LOGGER.info("Infrastructure cannot be stopped. Instances in group [{}] have ephemeral storage only.", instanceGroup.getGroupName());
```
lturcsanyi, line 27: "Instances in group [{}] have ephemeral storage **_only_**."

Is this still true?

daszabo:
The updated e2e test ran green on my machine. Also checked the new clusters manually. They seemed properly configured.

PR #11531, opened 2021-10-06
CB-14353 disable direct CM UI debug path in NGiNX configuration.

biharitomi:
> CB-14353 disable direct CM UI debug path in NGiNX configuration
It has been cherry-pick-ed from 2.49.0 line of Cloudbreak.


doktoric (approved):

horadla23:
this means, that https://github.com/hortonworks/cloudbreak#how-to-reach-cm-ui-directlynot-through-knox won't work, right? at least we need to change ssl.conf manually, so this should be added to the README

horadla23 (approved):

PR #11533, opened 2021-10-06
CB-14178 enable sonar on every modul except core where we get an exce…

doktoric:
…ption. Now we are using token mechanism for authentication.


biharitomi (commented):
Can we get the version of `sonarqube-gradle-plugin` from the gradle properties?
It would be easier to modify the version next time.

PR #11539, opened 2021-10-06
CB-14394 according to sonar .*\\s.* is not a safe expression

lnardai:
- Added unit tests and removed wildcards from regexp


datalake/src/test/java/com/sequenceiq/datalake/service/sdx/SdxServiceTest.java
```diff
@@ -204,6 +207,109 @@
...
+                {null, true},
+                {"7.2.0", false},
+                {"7.2.1", false},
+                {"7.2.5", false},
+                {"7.2.6", true},
+                {"7.2.7", true},
+        };
+    }
+
+    public static Stream<Arguments> storageBaseLocationsWhiteSpaceValidation() {
```
lnardai, line 304: Formatter was not applied to this file, so marking my code parts.
his is new.

PR #11556, opened 2021-10-08
CDPSDX-3120: Added Profiler Manager Service into a new medium duty sd…

horadla23:
…x blueprint and when CDP_DATA_LAKE_MEDIUM_DUTY_WITH_PROFILER Entitlement is enabled only then the sdx medium duty with profiler would be installed.


horadla23:
created this PR based on @subrat-mishra 's PR, he will test it and then we can merge it if everything is OK

PR #11569, opened 2021-10-11
CB-14323: Save the target image to the DB when the upgrade starts

tiborpopovics:
At the beginning of the upgrade, we need to save the target image for later use in the CM syncer flow.
-Save the target image to the Component table with component type IMAGE and name TARGET_IMAGE
-Remove the target image from the database at the end of the upgrade


keyki:
> Few minor comments and one question about retry related stuff.

@foldik can you please re-review?

PR #11574, opened 2021-10-12
CB-13609 DH Autoscaling should use machine user for polling yarn reco…

smaniraju:
…mmendations

Autoscaling to use environment level internal autoscaling-machine-user with EnvironmentUser Role for yarn polling.
Autoscaling to delete environment level autoscaling-machine-user if Autoscaling is not configured for any cluster in environment.

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
@@ -373,6 +373,35 @@ public class GrpcUmsClient {
     }
 
     /**
+     * Get or Create new machine user for given machineUserName.
+     *
+     * @param machineUserName new machine user name
+     * @param accountId       the accountId
+     * @param requestId       an optional request Id
+     * @return the machineUser
+     */
+    @Retryable(value = UmsOperationException.class, maxAttempts = 10, backoff = @Backoff(delay = 5000))
+    public MachineUser getOrCreateMachineUserWithoutAccessKey(String machineUserName, String accountId, Optional<String> requestId) {
+        try {
+            UmsClient client = makeClient(channelWrapper.getChannel());
+            String generatedRequestId = RequestIdUtil.getOrGenerate(requestId);
+            LOGGER.debug("Creating machine user {} for accountId {} using request ID {}", machineUserName, accountId, generatedRequestId);
+            MachineUser machineUser = client.getOrCreateMachineUserWithoutAccessKey(generatedRequestId, accountId, machineUserName);
+            LOGGER.debug("Machine User retrieved for machineUserName: {}, machineUser: {}", machineUserName, machineUser);
+            return machineUser;
+        } catch (StatusRuntimeException ex) {
+            if (Status.UNAVAILABLE.getCode().equals(ex.getStatus().getCode())) {
+                String errMessage = String.format("Cannot create machinue user '%s' for '%s' as " +
+                        "UMS API is UNAVAILABLE at the moment", machineUserName, accountId);
+                LOGGER.debug(errMessage, ex);
+                throw new UmsOperationException(errMessage, ex);
+            } else {
+                throw ex;
+            }
+        }
+    }
+
+    /**
      * Delete machine user
      *
      * @param machineUserName user name that should be deleted
--- a/autoscale/src/main/java/com/sequenceiq/periscope/init/InitializeMachineUserService.java
+++ b/autoscale/src/main/java/com/sequenceiq/periscope/init/InitializeMachineUserService.java
@@ -1,0 +1,43 @@
+package com.sequenceiq.periscope.init;
+
+import java.util.List;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Component;
+
+import com.google.common.base.Strings;
+import com.sequenceiq.periscope.domain.Cluster;
+import com.sequenceiq.periscope.monitor.handler.CloudbreakCommunicator;
+import com.sequenceiq.periscope.service.AltusMachineUserService;
+import com.sequenceiq.periscope.service.ClusterService;
+
+@Component
+public class InitializeMachineUserService {
+
+    @Inject
+    private ClusterService clusterService;
+
+    @Inject
+    private AltusMachineUserService altusMachineUserService;
+
+    @Inject
+    private CloudbreakCommunicator cloudbreakCommunicator;
+
+    @PostConstruct
+    public void init() {
+        List<Cluster> clusters = clusterService.findByEnvironmentCrnOrMachineUserCrn(null, null);
+        clusters.forEach(cluster ->  {
+            if (Strings.isNullOrEmpty(cluster.getEnvironmentCrn())) {
+                String envCrn = cloudbreakCommunicator.getAutoscaleClusterByCrn(cluster.getStackCrn()).getEnvironmentCrn();
+                clusterService.setEnvironmentCrn(cluster.getId(), envCrn);
+                cluster = clusterService.findById(cluster.getId());
+            }
+
+            if (Strings.isNullOrEmpty(cluster.getMachineUserCrn())) {
+                altusMachineUserService.initializeMachineUserForEnvironment(cluster);
+            }
+        });
+    }
+}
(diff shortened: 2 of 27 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
```diff
@@ -371,6 +371,35 @@ public MachineUser getMachineUserDetails(String userCrn, String accountId, Optio
...
+                String errMessage = String.format("Cannot create machinue user '%s' for '%s' as " +
+                        "UMS API is UNAVAILABLE at the moment", machineUserName, accountId);
+                LOGGER.debug(errMessage, ex);
+                throw new UmsOperationException(errMessage, ex);
+            } else {
+                throw ex;
+            }
+        }
+    }
+
```
horadla23, line 402: what is the difference between this and method `createMachineUser` ? As I see there are common parts, maybe it is worth to create different method for common parts, what do you think?

autoscale/src/main/java/com/sequenceiq/periscope/service/AltusMachineUserService.java
```diff
@@ -0,0 +1,98 @@
...
+    private void syncEnvironment(String accountId, String machineUserCrn, String environmentCrn, Optional<String> deletedWorkloadUserName) {
+        ThreadBasedUserCrnProvider.doAsInternalActor(() -> {
+            SynchronizeAllUsersRequest request = new SynchronizeAllUsersRequest();
+            request.setAccountId(accountId);
+            request.setEnvironments(Set.of(environmentCrn));
+            request.setMachineUsers(Set.of(machineUserCrn));
+            if (deletedWorkloadUserName.isPresent()) {
+                request.setDeletedWorkloadUsers(Set.of(deletedWorkloadUserName.get()));
+            }
+            userV1Endpoint.synchronizeAllUsers(request);
```
horadla23, line 93: are you sure we need this? I would ask @handavid or @gitmfox if there is automatic sync implemented or not, because I would avoid to introduce a new service-to-service relation (autoscale - freeipa), if possible.

core-api/src/main/java/com/sequenceiq/cloudbreak/api/endpoint/v4/stacks/response/AutoscaleStackV4Response.java
```diff
@@ -72,6 +72,9 @@
     @ApiModelProperty(StackModelDescription.TYPE)
     private StackType stackType;
 
+    @ApiModelProperty(StackModelDescription.ENVIRONMENT_CRN)
+    private String environmentCrn;
+
```
horadla23, line 77: is it required to add environmentCrn to the API also? We are using it for internal logic only as I see

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
```diff
@@ -56,6 +61,23 @@ public void onApplicationEvent(ClusterStatusSyncEvent event) {
             clusterService.setState(cluster.getId(), ClusterState.SUSPENDED);
             LOGGER.info("Suspended cluster '{}', CB Stack Status '{}', CB Cluster Status '{}'",
                     cluster.getStackCrn(), statusResponse.getStatus(), statusResponse.getClusterStatus());
+        } else if (cluster.getEnvironmentCrn() == null) {
+            String environmentCrn = cloudbreakCommunicator.getAutoscaleClusterByCrn(cluster.getStackCrn()).getEnvironmentCrn();
+            clusterService.setEnvironmentCrn(cluster.getId(), environmentCrn);
+        } else if (cluster.getMachineUserCrn() == null) {
```
horadla23, line 67: This logic should not belong to this class, the purpose of this class to execute logics based on status of the cluster, not to fill in environmentCrn for old entries in periscope. Have you checked if it is possible to fill in environmentCrn using other solutions like a migration class during application start up? Using this logic will put a high pressure on CB and autoscale at first on application startup, then application will never reach this code again.
If there is no other solution, I would suggest to refactor this code at least into another class, since this class should handle only statuses, not environmentCrns.

horadla23 (changes requested):

autoscale/src/main/java/com/sequenceiq/periscope/service/AltusMachineUserService.java
horadla23: I see there were already another thread about this, although if there is no high priority on this task, I would seek for solution/wait for solution which removes the necessity of this sync, since this will require addition configuration also in CSI repo before merging this change.

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
horadla23: for example it would be better if a separate class could call a list on datahub API, filter for relevant ones and fill the environmentCrn, in this case we would not call CB several times during application startup

doktoric:
do we still need this?

sidseth:
I believe we do. This is almost done. Likely needs to be picked up again soon ... tested and then merged. For now, can we leave it as a pending PR?

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
aakashshankar: Sorry for the long gap between updates.

Addressed this by creating `InitializeMachineUserService` with `@PostConstruct` to initialize environmentCrn and machineUserCrn for older periscope clusters, so that it is executed during application startup. For new clusters, envCrn and machineUserCrn initialization is added to the `create` method which persists a newly created cluster in the persicope DB via `ClusterService`

aakashshankar:
retest this please

autoscale/src/main/java/com/sequenceiq/periscope/service/AltusMachineUserService.java
aakashshankar: Has automatic sync been implemented? A jira has been filed to track the removal of this explicit userSync: [CB-14554](https://jira.cloudera.com/browse/CB-14554)

aszegedi:
retest this please

auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/GrpcUmsClient.java
horadla23: @aakashshankar can you check this comment of mine?

autoscale/src/main/java/com/sequenceiq/periscope/service/AltusMachineUserService.java
horadla23: @aakashshankar it seems that is an old ticket, please check if manual sync is still needed or not

autoscale/src/main/java/com/sequenceiq/periscope/monitor/handler/ClusterStatusSyncHandler.java
horadla23: @aakashshankar this logic still calls CB in a foreach which can result several CB call, we should avoid that

autoscale/src/main/java/com/sequenceiq/periscope/init/InitializeMachineUserService.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Inject
+    private CloudbreakCommunicator cloudbreakCommunicator;
+
+    @PostConstruct
+    public void init() {
+        List<Cluster> clusters = clusterService.findByEnvironmentCrnOrMachineUserCrn(null, null);
+        clusters.forEach(cluster ->  {
+            if (Strings.isNullOrEmpty(cluster.getEnvironmentCrn())) {
+                String envCrn = cloudbreakCommunicator.getAutoscaleClusterByCrn(cluster.getStackCrn()).getEnvironmentCrn();
```
horadla23, line 33: @aakashshankar as mentioned, API call in foreach, please try to avoid that

aakashshankar:
Closing this PR, as PR jobs cannot be triggered for the current owner. Opened #12480, to address the review comments.

cc @horadla23 @sidseth

PR #11598, opened 2021-10-14
CB-14505 Upgrade Cloudbreak project to use Gradle 7

bergerdenes:
Upgrading to Gradle 7.3 rc1 in order to prepare migration to Java 17.
Reorganized gradle files because of removed "compile" and "testCompile" targets.
Bumped version for checkstyle and spotbugs.


lturcsanyi:
Why RC?

bergerdenes:
> Why RC?

that is the latest available for now, and the earliest that supports Java 17

lturcsanyi:
> and the earliest that supports Java 17

Maybe we should wait for a more stable version.

PR #11646, opened 2021-10-22
CB-14645 Adds real ums tests to validate change image catalog permissions.

pkedvessy:

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/ChangeImageCatalogTest.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/ChangeImageCatalogTest.java
@@ -1,0 +1,312 @@
+package com.sequenceiq.it.cloudbreak.testcase.authorization;
+
+import static com.sequenceiq.it.cloudbreak.context.RunningParameter.expectedMessage;
+
+import javax.inject.Inject;
+import javax.ws.rs.ForbiddenException;
+
+import org.testng.annotations.Test;
+
+import com.sequenceiq.environment.api.v1.environment.model.response.EnvironmentStatus;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status;
+import com.sequenceiq.it.cloudbreak.actor.CloudbreakActor;
+import com.sequenceiq.it.cloudbreak.client.CredentialTestClient;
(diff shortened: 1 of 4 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/authorization/ChangeImageCatalogTest.java
```diff
@@ -0,0 +1,174 @@
...
+        useRealUmsUser(testContext, AuthUserKeys.ENV_CREATOR_A);
+    }
+
+    @Test(dataProvider = TEST_CONTEXT_WITH_MOCK)
+    @Description(
+            given = "there is a running Freeipa",
+            when = "a change image catalog request is sent",
+            then = "admin user can perform the operation but env creator user should get forbidden exception")
+    public void testChangeFreeipaImageCatalog(MockedTestContext testContext) {
+        useRealUmsUser(testContext, AuthUserKeys.ACCOUNT_ADMIN);
```
foldik, line 69: Would it make sense to do the following:

- env_creator_a creates all resources (image catalog, credential, freeipa, environment...),
- env_creator_b creates a new image catalog,
- env_creator_b tries to change the image catalog to the one he created and it fails,
- account_admin tries to do the same (change the image catalog to env_creator_b's image catalog) and he succeeds.

foldik (changes requested):
I recognised that we don't authorize if the user has access to the image catalog here:

```java
public class DistroXV1Controller {

    @Override
    @CheckPermissionByResourceName(action = AuthorizationResourceAction.CHANGE_IMAGE_CATALOG_DATAHUB)
    public void changeImageCatalog(@ResourceName String name, @Valid @NotNull ChangeImageCatalogV4Request changeImageCatalogRequest)
```

and here

```java
public class SdxController {

    @CheckPermissionByResourceName(action = AuthorizationResourceAction.CHANGE_IMAGE_CATALOG_DATALAKE)
    public void changeImageCatalog(@ResourceName String name, SdxChangeImageCatalogRequest changeImageCatalogRequest)
```

and here

```java
public class FreeIpaV1Controller {

    @CheckPermissionByResourceCrn(action = ENVIRONMENT_CHANGE_FREEIPA_IMAGE)
    public void changeImageCatalog(@ResourceCrn String environmentCrn, ChangeImageCatalogRequest changeImageCatalogRequest)
```

PR #11656, opened 2021-10-25
CB-14624 Cluster definition uniqueness across the env instead of work…

topolyai5:
…space. Check the cluster definition name and the related stack's env crn.

* if the new and old one's env crn is null -> exception (they are default template)
* if the env crn is the same -> exception (they are in the same env)
* if the new or the old one's env crn is null -> ok


topolyai5:
This need more investigation

PR #11661, opened 2021-10-26
CB-14482 Set the callingServiceName GRPC context for UMS calls

foldik:

### DIFF
--- a/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/config/UmsClientConfig.java
+++ b/auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/config/UmsClientConfig.java
@@ -1,5 +1,10 @@
 package com.sequenceiq.cloudbreak.auth.altus.config;
 
+import static org.glassfish.jersey.internal.guava.Preconditions.checkArgument;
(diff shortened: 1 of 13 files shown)


auth-connector/src/main/java/com/sequenceiq/cloudbreak/auth/altus/config/UmsClientConfig.java
```diff
@@ -23,6 +23,9 @@
     @Value("${altus.ums.client.list_service_principal_cloud_identities_page_size:100}")
     private int listServicePrincipalCloudIdentitiesPageSize;
 
+    @Value("${altus.ums.calling_service_name:cloudbreak}")
```
horadla23, line 26: spring property with underscores looks strange, I would use `caller.service.name` or simply `caller`, what do you think about it?

foldik: Renamed the property to `caller`.

horadla23 (approved):

lnardai (approved):
Can't we add an assertion that ensures the header added properly?

foldik:
@lnardai I can add one to mock ums.

doktoric (approved):

akanto (approved):

PR #11665, opened 2021-10-27
CB-14656 Fix unbound service restart on running clusters

Bajzathd:
A new step was introduced as part of the cluster sync. This step will check package versions if any of them are incorrect, and tries to fix them.

**Note:** This is just a POC, it should be refined to production ready quality


doktoric:
do we still need this?

Bajzathd:
The actual solution is already merged, so I am closing this

PR #11676, opened 2021-10-28
CB-13604: After a failed upgrade the syncer reads parcel and CM repo …

tiborpopovics:
The system might well be in a mixed state (e.g. CM upgraded, CDH not). This setup is not guaranteed to work for the customer, as a given CDH only works with a specific version of parcel X.

There are two main cases when the CB send notification about the mixed package versions:

## 1. Syncer called by the upgrade failure flow handler. 
In this case, the target image is present in the database, so we compare the active parcels to the parcel versions in the target image.

When the package versions are not valid based on the target image we show the following message:
> The current Cloudera Manager 7.4.0, Runtime CDH 7.2.0-1.cdh7.2.0.p3.5579072, combination is not certified by Cloudera therefore their usage is unadvised and might lead to unexpected errors. Please consider upgrading them to a certified combination, like Cloudera Manager 7.4.0,  CDH 7.2.8-1.cdh7.2.8.p0.11560957

When the package versions are not valid because the user installed a higher version than expected:
> The current Cloudera Manager 7.4.0, Runtime CDH 7.2.8-1.cdh7.2.8.p0.11560957, FLINK 1.13.2-csadh1.5.1.0-cdh7.2.13.0-160-18643985 combination is not certified by Cloudera therefore their usage is unadvised and might lead to unexpected errors. The following components are newer Runtime CDH 7.2.8-1.cdh7.2.8.p0.11560957, FLINK 1.13.2-csadh1.5.1.0-cdh7.2.13.0-160-18643985 then the expected versions CDH 7.2.0-1.cdh7.2.0.p3.5579072, FLINK 1.10.0-csa1.2.1.0-cdh7.2.1.0-327-7398223. Please use recovery or contact the Cloudera support.

## 2. Syncer called manually
In this case, the target image is not present in the database so we need to validate the active parcels based on the image catalog.

When the package versions are not valid and multiple possible combinations are found we show the following message:
> The current Cloudera Manager 7.4.0, Runtime CDH 7.2.0-1.cdh7.2.0.p3.5579072, combination is not certified by Cloudera therefore their usage is unadvised and might lead to unexpected errors. Please consider upgrading them to a certified combination, like Cloudera Manager 7.4.0,  CDH 7.2.8-1.cdh7.2.8.p0.11560957. For other possible combinations please visit the image catalog at https://cloudbreak-imagecatalog.s3.amazonaws.com/v3-test-cb-image-catalog.json

When the package versions are not valid and only combination found we show the following message:
> The current Cloudera Manager 7.4.0, Runtime CDH 7.2.0-1.cdh7.2.0.p3.5579072, combination is not certified by Cloudera therefore their usage is unadvised and might lead to unexpected errors. Please consider upgrading them to a certified combination, like Cloudera Manager 7.4.0,  CDH 7.2.8-1.cdh7.2.8.p0.11560957.

When the package versions are not valid and there is no combination found we show the following message:
> The current Cloudera Manager 7.4.0, Runtime CDH 7.2.0-1.cdh7.2.0.p3.5579072, combination is not certified by Cloudera therefore their usage is unadvised and might lead to unexpected errors. Please contact the Cloudera support for further actions.


core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+
+    @Inject
+    private ImageService imageService;
+
+    @Inject
+    private ImageCatalogService imageCatalogService;
+
+    void validatePackageVersions(Long stackId, CmSyncOperationResult syncResult, Set<Image> candidateImages) {
+        Optional<String> activeCmVersion = syncResult.getCmRepoSyncOperationResult().getInstalledCmVersion();
+        if (activeCmVersion.isPresent()) {
```
gergopapi2, line 64: I wonder if it makes sense to do the validation if cm version is present but for some reason there is no parcel info from the server. Not clear, however, how can we have CM version but no parcel versions.
There is a task in phase 2 to read the CM version from yum if CM would not respond. If that get's implemented then it might cause such a situation that only CM version is present.
It might make sense to check CM version alone, but then messages might need to be changed (e.g. "... is not a tested combination" would not make much sense)

common/src/main/resources/messages/messages.properties
```diff
@@ -66,6 +66,10 @@ stack.cm.diagnostics.upload.running=CM based diagnostics uploading in progress.
 stack.cm.diagnostics.cleanup.running=CM based diagnostics cleanup in progress
 stack.cm.diagnostics.collection.finished=CM based Diagnostics collection finished
 stack.cm.diagnostics.collection.failed=CM based diagnostics collection failed. Reason: {0}
+stack.cm.mixed.package.versions.failed=The current Cloudera Manager {0}, {1} are not a tested combination to work together and their usage might lead to errors. Please consider upgrading them to a working combination, like {2}
+stack.cm.mixed.package.versions.failed.no.candidate=The current Cloudera Manager {0}, {1} are not a tested combination to work together and their usage might lead to errors. Please contact the Cloudera support for further actions.
+stack.cm.mixed.package.versions.failed.multiple=The current Cloudera Manager {0}, {1} are not a tested combination to work together and their usage might lead to errors. Please consider upgrading them to a working combination, like {2}. For other possible combinations please see the following link: {3}
```
gergopapi2, line 71: Instead of following link I would add: "Please visit the image catalog at {3}"

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+        return String.format("Cloudera Manager %s, %s", getCmVersion(image.getPackageVersions()), suggestedParcelsMessage);
+    }
+
+    private Set<Image> getImageByCurrentCmVersion(String currentCmVersion, Set<Image> candidateImages) {
+        return candidateImages.stream()
+                .filter(image -> currentCmVersion.equals(getCmVersion(image.getPackageVersions())))
+                .collect(Collectors.toSet());
+    }
+
+    private boolean allActiveParcelVersionMatchesWithCandidateImage(Set<ParcelInfo> activeParcels, String cmVersion, Set<Image> images) {
```
gergopapi2, line 215: I would call it as componentVersionsMatchWithAnyImage, similar to the naming used earlier.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+    @Inject
+    private ClouderaManagerProductTransformer clouderaManagerProductTransformer;
+
+    @Inject
+    private ImageService imageService;
+
+    @Inject
+    private ImageCatalogService imageCatalogService;
+
+    void validatePackageVersions(Long stackId, CmSyncOperationResult syncResult, Set<Image> candidateImages) {
```
gergopapi2, line 62: This class contains many private methods, several of them containing quite some logic. While the unit test of this class does give coverage, but I think it would be better if some of the methods representing basic operations (e.g. comparing parcel and CM versions to images, or message generation) would be extracted to a separate class and would receive unit tests on their own. I would leave only the generic logic here (if matches original image - else if matches target image - else mathces any image).

gergopapi2 (changes requested):

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+                .collect(Collectors.toMap(ParcelInfo::getName, ParcelInfo::getVersion));
+        if (versionComparator.compare(() -> currentCmVersion, () -> getCmVersion(targetImage.getPackageVersions())) > 0) {
+            componentsWithHigherVersion.put("Cloudera Manager", currentCmVersion);
+        }
+        return componentsWithHigherVersion;
+    }
+
+    private boolean compareParcelVersions(ParcelInfo parcelInfo, Set<ClouderaManagerProduct> productsFromImage, VersionComparator versionComparator) {
+        return productsFromImage.stream().filter(product -> parcelInfo.getName().equals(product.getName())).findFirst()
+                .map(version -> versionComparator.compare(parcelInfo::getVersion, version::getVersion) > 0)
```
pdarvasi, line 166: I'd rename `version` to indicate that it is a `ClouderaManagerProduct`

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private boolean allActiveParcelVersionMatchesWithCandidateImage(Set<ParcelInfo> activeParcels, String cmVersion, Set<Image> images) {
+        return images.stream().anyMatch(image -> cmVersion.equals(getCmVersion(image.getPackageVersions())) && matchParcelVersions(activeParcels, image));
+    }
+
+    private boolean matchParcelVersions(Set<ParcelInfo> activeParcels, Image image) {
+        Set<ClouderaManagerProduct> clouderaManagerProducts = getProductsFromImage(image);
+        return activeParcels.stream()
+                .allMatch(parcelInfo -> clouderaManagerProducts.stream()
```
pdarvasi, line 222: again, a `HashMap` of name and version would be better here instead of `Set<clouderaManagerProducts>`

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private boolean componentVersionsAreMatchesWithImage(Image image, Set<ClouderaManagerProduct> productsFromImage, String currentCmVersion,
+            Set<ParcelInfo> activeParcels) {
+        return currentCmVersion.equals(getCmVersion(image.getPackageVersions())) && allActiveParcelVersionMatches(activeParcels, productsFromImage);
+    }
+
+    private boolean allActiveParcelVersionMatches(Set<ParcelInfo> activeParcels, Set<ClouderaManagerProduct> productsFromImage) {
+        return activeParcels.stream()
+                .allMatch(parcelInfo -> productsFromImage.stream()
```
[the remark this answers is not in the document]
pdarvasi: this logic starting with `.allMatch` is present twice in the class, pls extract into a method

core/src/test/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionServiceTest.java
```diff
@@ -0,0 +1,355 @@
...
+import com.sequenceiq.cloudbreak.service.image.StatedImage;
+import com.sequenceiq.cloudbreak.service.parcel.ClouderaManagerProductTransformer;
+import com.sequenceiq.cloudbreak.service.upgrade.sync.common.ParcelInfo;
+import com.sequenceiq.cloudbreak.service.upgrade.sync.operationresult.CmParcelSyncOperationResult;
+import com.sequenceiq.cloudbreak.service.upgrade.sync.operationresult.CmRepoSyncOperationResult;
+import com.sequenceiq.cloudbreak.service.upgrade.sync.operationresult.CmSyncOperationResult;
+import com.sequenceiq.cloudbreak.structuredevent.event.CloudbreakEventService;
+
+@ExtendWith(MockitoExtension.class)
+public class MixedPackageVersionServiceTest {
```
pdarvasi, line 44: pls cover this logic, too:

            if (imagesWithSameCmVersion.isEmpty()) {
                LOGGER.debug("There is no other image in the catalog with the same {} CM version.", currentCmVersion);
                sendNotification(stackId, STACK_CM_MIXED_PACKAGE_VERSIONS_FAILED_NO_CANDIDATE, Collections.emptyList());

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,226 @@
...
+        try {
+            return Optional.of(imageCatalogService.getImage(image.getImageCatalogUrl(), image.getImageCatalogName(), image.getImageId()).getImage());
+        } catch (CloudbreakImageNotFoundException | CloudbreakImageCatalogException e) {
+            LOGGER.warn("There is no image {} found in catalog {}.", image.getImageId(), image.getImageCatalogUrl());
+            return Optional.empty();
+        }
+    }
+
+    private void examinePackageVersionsWithTargetImage(Long stackId, Image targetImage, String activeCmVersion, Set<ParcelInfo> activeParcels) {
+        LOGGER.debug("Comparing active package versions {} and active CM version {} with target image {}", activeParcels, activeCmVersion, targetImage);
```
gergopapi2, line 100: This is also a largish private method, and would add test for it. However, in that case it might make sense to put it to a separate class and add tests for it on its own.

core/src/main/java/com/sequenceiq/cloudbreak/service/upgrade/sync/MixedPackageVersionService.java
```diff
@@ -0,0 +1,226 @@
...
+                sendNotification(stackId, STACK_CM_MIXED_PACKAGE_VERSIONS_NEWER_FAILED, List.of(activeCmVersion, activeParcelsMessage,
+                        createMessageFromMap(newerComponents),
+                        createMessageFromMap(filterTargetPackageVersionsByNewerPackageVersions(targetProducts, targetImage, newerComponents))));
+            }
+        }
+    }
+
+    private void examinePackageVersionsWithAllCandidateImages(Long stackId, Set<Image> candidateImages, String currentCmVersion, Set<ParcelInfo> activeParcels,
+            String imageCatalogUrl) {
+        LOGGER.debug("Target image not found for cluster therefore comparing package versions with other candidate images.");
```
gergopapi2, line 120: This is also a large private method and I would be nice to have tests for the method. It might make sense to put it to a separate class and create tests for it.

gergopapi2 (changes requested):

pdarvasi (changes requested):

pdarvasi (approved):

keyki (approved):

keyki (approved):

keyki:
Gergo's requests have been addressed, but he is on vacation for a while and this PR can't move forward without his approval. Please ignore those requests.

pdarvasi (approved):

gergopapi2 (approved):
I finally had a quick look at the files, LGTM.

PR #11678, opened 2021-10-28
CB-14688: Remove world-readable permissions on cloudbreak webhdfs token

tiborpopovics:
Before my change:
```
ls -l /tmp/cloudbreak-webhdfs.cookies
-rw-r--r-- 1 root root 406 Oct 26 19:02 /tmp/cloudbreak-webhdfs.cookies
```
After my change:
```
ls -l /tmp/cloudbreak-webhdfs.cookies
-rw------- 1 root root 406 Nov  2 13:13 /tmp/cloudbreak-webhdfs.cookies
```

### DIFF
--- a/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
+++ b/orchestrator-salt/src/main/resources/salt/salt/post-recipes/scripts/createuserhome.sh
@@ -53,6 +53,7 @@ webhdfs_command()   {
       -X $method --negotiate -u : \
       "$WEBHDFS_URL$path?$query")
     exec {out_fd}>&-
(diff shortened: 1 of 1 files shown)


orchestrator-salt/src/main/resources/salt/salt/recipes/post-cluster-install.sls
```diff
@@ -37,6 +37,12 @@ run_post_cluster_install_script_{{ script_name }}:
     - makedirs: True
     - mode: 755
 
+/tmp/cloudbreak-webhdfs.cookies:
```
keyki, line 40: - mode: 755 is 2 lines above, why don't we change it there?

lacikaaa (approved):

PR #11680, opened 2021-10-29
CB-14640: Fix yarn_nodemanager_log_dirs to be ephemeral volumes if present

juhi-09:
CB-14640: Fix yarn_nodemanager_log_dirs to be ephemeral volumes if present
If Instance has ephemeral volumes which are mounted, temporary data like `yarn_nodemanager_local_dirs` and `yarn_nodemanager_log_dirs` should use ephemeral volumes.


nadamdb:
> > Is it the expected to store the yarn logs on the ephemeral volumes instead of the permanent ones? Isn't log data need to be permanent in case of some failure?
> The logs get uploaded to cloud storage after an app completes. We can also look at enabling incremental log-aggregation in YARN.

Thanks for clarifying. 
The change LGTM

nadamdb (approved):

doktoric (approved):

PR #11684, opened 2021-10-30
CB-14692: Added knox keytab use to backup/restore scripts in order to resolve authentication issues on DataHubs

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-14692

Simply added the Knox keytab as a choice to the backup/restore scripts.
Verified that this enables successful database backups on Azure and AWS DataHubs.


brycederriso (approved):
This is pretty standard by now, changes look good to me.

Thanks Santiago!

orchestrator-salt/src/main/resources/salt/salt/postgresql/disaster_recovery/scripts/backup_db.sh
```diff
@@ -81,6 +82,8 @@ run_kinit() {
     doLog "Successful kinit using solr principal"
   elif kinit_as atlas "$ATLAS_KEYTAB"; then
     doLog "Successful kinit using atlas principal"
+  elif kinit_as knox "$KNOX_KEYTAB"; then
+    doLog "Successful kinit using knox principal"
```
hreeve-cloudera, line 86: Should we have some final `else` statement that lists the users we failed to kinit as? For troubleshooting if we ever need it.

sxxgrc: The logging done by the **kinit** function takes care of this. When you run the scripts the logs will tell you exactly which kinit failed and what the actual principal used for authentication was.

PR #11693, opened 2021-11-02
CB-14721 added short circuit for SdxClusterStatusCheckerJob

lnardai:
- added short circuit for SdxClusterStatusCheckerJob when there is a running flow for that Datalake to prevent adhoc status changes
- it was causing sideeffect for SDX status WAITING_FOr_EXTERNAL_DATABASE_STOP would be switched to STOP making EnvService saying stop complete


lnardai:
Talked with @doktoric and Rebased to 2.49.

doktoric (approved):

PR #11696, opened 2021-11-02
CB-14697 do not use body for get request -> use query params instead.…

oleewere:
… (addendum)


oleewere:
this is a merge from master to 2.49 (put 2 commits regarding CB-14697 into 1 commit for simplicity) - this is required, as if 2.49 swagger will be checked for master -> that will cause swagger error, validation here is really simple, will be more proper on master, but the endpoints should be updated to this at least

doktoric (approved):

sodre90 (approved):

PR #11697, opened 2021-11-02
CB-14572 Add TunnelType to FreeIPA response

lajosrodek:
* Add new property `DescribeFreeIpaResponse.tunnel` (`freeipa-api`). It is copied as-is from `Stack.tunnel` (`freeipa`).
* `freeipa` Swagger fixes:
  * Add API description.
  * Correct base path so that Swagger UI becomes usable.
  * Add `ApiKeyAuthDefinition` to `FreeIpaApi` (`freeipa`) so that user authn information can be entered in Swagger UI. (Solution has been adapted from `redbeams`.) This extra header is currently only enabled for `FreeIpaV1Endpoint`.
* Fix spelling errors in several places.
* Testing:
  * UT.
  * `freeipa` Swagger UI in local CB.

### DIFF
--- a/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
+++ b/freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
@@ -24,6 +24,7 @@ import com.sequenceiq.cloudbreak.auth.security.internal.AccountId;
 import com.sequenceiq.cloudbreak.jerseyclient.RetryAndMetrics;
 import com.sequenceiq.flow.api.model.FlowIdentifier;
 import com.sequenceiq.flow.api.model.RetryableFlowResponse;
+import com.sequenceiq.freeipa.api.FreeIpaApi;
 import com.sequenceiq.freeipa.api.v1.freeipa.cleanup.CleanupRequest;
 import com.sequenceiq.freeipa.api.v1.freeipa.stack.doc.FreeIpaNotes;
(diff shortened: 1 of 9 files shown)


freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
```diff
@@ -43,11 +44,15 @@
...
 import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.Authorization;
 
 @RetryAndMetrics
 @Path("/v1/freeipa")
 @Consumes(MediaType.APPLICATION_JSON)
-@Api(value = "/v1/freeipa", protocols = "http,https", consumes = MediaType.APPLICATION_JSON)
+@Api(value = "/v1/freeipa",
+        protocols = "http,https",
+        authorizations = {@Authorization(value = FreeIpaApi.CRN_HEADER_API_KEY)},
```
bergerdenes, line 54: will it work with DP CLI?

bergerdenes (approved):
LGTM

freeipa-api/src/main/java/com/sequenceiq/freeipa/api/v1/freeipa/stack/FreeIpaV1Endpoint.java
lajosrodek: I hope so, but will give it a try to be sure. The same approach is already in place for `redbeams`.

lacikaaa (approved):

lajosrodek:
This has been also verified with dp CLI. See https://github.com/hortonworks/cb-cli/pull/791 for the respective changes.

PR #11699, opened 2021-11-03
CB-14656 Fix unbound service restart of existing stacks

Bajzathd:
Introduce a new quartz job that runs for existing stacks, applying fixes.
The first fix implemented is a salt command to replace the faulty unbound service restart on the stacks that use one of the affected images.

### DIFF
--- a/core/src/main/java/com/sequenceiq/cloudbreak/job/existingstackfix/ExistingStackFixerJobService.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/job/existingstackfix/ExistingStackFixerJobService.java
@@ -1,0 +1,89 @@
+package com.sequenceiq.cloudbreak.job.existingstackfix;
+
+import javax.inject.Inject;
+
+import org.quartz.JobBuilder;
+import org.quartz.JobDataMap;
+import org.quartz.JobDetail;
+import org.quartz.JobKey;
+import org.quartz.Scheduler;
+import org.quartz.SchedulerException;
+import org.quartz.SimpleScheduleBuilder;
+import org.quartz.Trigger;
+import org.quartz.TriggerBuilder;
+import org.quartz.impl.matchers.GroupMatcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ExistingStackFixerJobService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExistingStackFixerJobService.class);
+
+    private static final String JOB_GROUP = "existing-stack-fixer-jobs";
+
+    private static final String TRIGGER_GROUP = "existing-stack-fixer-triggers";
+
+    private static final String LOCAL_ID = "localId";
+
+    private static final String REMOTE_RESOURCE_CRN = "remoteResourceCrn";
+
+    @Inject
+    private ExistingStackFixerConfig properties;
+
+    @Inject
+    private Scheduler scheduler;
+
+    public void schedule(ExistingStackFixerJobAdapter resource) {
+        String localId = resource.getLocalId();
+        JobDetail jobDetail = buildJobDetail(localId, resource.getRemoteResourceId());
+        Trigger trigger = buildJobTrigger(jobDetail);
+        try {
+            if (scheduler.getJobDetail(JobKey.jobKey(localId, JOB_GROUP)) != null) {
+                unschedule(localId);
+            }
+            scheduler.scheduleJob(jobDetail, trigger);
+        } catch (SchedulerException e) {
+            LOGGER.error(String.format("Error during scheduling quartz job: %s", jobDetail), e);
+        }
+    }
+
+    public void unschedule(String id) {
+        JobKey jobKey = JobKey.jobKey(id, JOB_GROUP);
+        try {
+            if (scheduler.getJobDetail(jobKey) != null) {
+                scheduler.deleteJob(jobKey);
+            }
+            if (scheduler.getJobKeys(GroupMatcher.groupEquals(JOB_GROUP)).isEmpty()) {
+                LOGGER.info("All existing stacks have been fixed, hooray!");
+            }
+        } catch (SchedulerException e) {
+            LOGGER.error(String.format("Error during unscheduling quartz job: %s", jobKey), e);
+        }
+    }
+
+    private JobDetail buildJobDetail(String stackId, String crn) {
+        JobDataMap jobDataMap = new JobDataMap();
+        jobDataMap.put(LOCAL_ID, stackId);
+        jobDataMap.put(REMOTE_RESOURCE_CRN, crn);
+        return JobBuilder.newJob(ExistingStackFixerJob.class)
+                .withIdentity(stackId, JOB_GROUP)
+                .withDescription("Fixing existing stack: " + crn)
+                .usingJobData(jobDataMap)
+                .storeDurably()
+                .build();
+    }
+
+    private Trigger buildJobTrigger(JobDetail jobDetail) {
+        return TriggerBuilder.newTrigger()
+                .forJob(jobDetail)
+                .withIdentity(jobDetail.getKey().getName(), TRIGGER_GROUP)
+                .withDescription("Trigger for fixing existing stack")
+                .withSchedule(SimpleScheduleBuilder.simpleSchedule()
+                        .withIntervalInHours(properties.getIntervalInHours())
+                        .repeatForever()
+                        .withMisfireHandlingInstructionNextWithRemainingCount())
+                .build();
+    }
+}
--- a/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabProvider.java
+++ b/core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabProvider.java
@@ -8,6 +8,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import com.sequenceiq.cloudbreak.auth.ThreadBasedUserCrnProvider;
+import com.sequenceiq.cloudbreak.auth.crn.Crn;
+import com.sequenceiq.cloudbreak.auth.crn.InternalCrnBuilder;
 import com.sequenceiq.cloudbreak.common.exception.CloudbreakServiceException;
 import com.sequenceiq.cloudbreak.common.exception.WebApplicationExceptionMessageExtractor;
 import com.sequenceiq.cloudbreak.domain.stack.Stack;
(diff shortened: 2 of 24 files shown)


Bajzathd:
The scenario where upscaling a cluster created with the faulty image is not yet working correctly, so moving it back to draft state

Bajzathd:
The fix was to trigger a full salt update, I have added it in a separate commit. Reopening PR.

core/src/main/java/com/sequenceiq/cloudbreak/job/existingstackfix/ExistingStackFixerJobService.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExistingStackFixerJobService.class);
+
+    private static final String JOB_GROUP = "existing-stack-fixer-jobs";
+
+    private static final String TRIGGER_GROUP = "existing-stack-fixer-triggers";
+
+    private static final String LOCAL_ID = "localId";
+
+    private static final String REMOTE_RESOURCE_CRN = "remoteResourceCrn";
```
lacikaaa, line 30: I see these multiple times in our code. I think it worth it to move to somewhere common instead of duplicating them again

core/src/main/java/com/sequenceiq/cloudbreak/reactor/handler/kerberos/KeytabProvider.java
```diff
@@ -33,7 +35,12 @@
     public ServiceKeytabResponse getServiceKeytabResponse(Stack stack, GatewayConfig primaryGatewayConfig) {
         ServiceKeytabRequest request = cmServiceKeytabRequestFactory.create(stack, primaryGatewayConfig);
         try {
-            String accountId = ThreadBasedUserCrnProvider.getAccountId();
+            String accountId;
+            if (InternalCrnBuilder.isInternalCrn(ThreadBasedUserCrnProvider.getUserCrn())) {
+                accountId = Crn.fromString(stack.getResourceCrn()).getAccountId();
+            } else {
+                accountId = ThreadBasedUserCrnProvider.getAccountId();
+            }
```
lacikaaa, line 43: please refactor, move the logic to a method, so it would look like:
String accountId = awesomemethod(stack.getreasourcecrn());
You might add some log to the new method so we would see what happened there

Bajzathd:
Moving the changes to a new PR with squashed commits, more detailed commit message, and to CB-2.49.0 instead of master
https://github.com/hortonworks/cloudbreak/pull/11728

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/GatewayConfigProvider.java
```diff
@@ -0,0 +1,29 @@
...
+import com.sequenceiq.cloudbreak.orchestrator.model.GatewayConfig;
+import com.sequenceiq.cloudbreak.service.GatewayConfigService;
+
+@Component
+public class GatewayConfigProvider {
+
+    @Inject
+    private GatewayConfigService gatewayConfigService;
+
+    public GatewayConfig getGatewayConfig(Stack stack) {
```
lacikaaa, line 19: why don't you use `GatewayConfigService#getPimaryGatewayConfig`?

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/ExistingStackFixService.java
```diff
@@ -0,0 +1,55 @@
...
+    private FlowLogService flowLogService;
+
+    public boolean isStackAlreadyFixed(Stack stack) {
+        return stackFixRepository.findByStackAndType(stack, getStackFixType()).isPresent();
+    }
+
+    public void apply(Stack stack) {
+        if (flowLogService.isOtherFlowRunning(stack.getId())) {
+            throw new IllegalStateException("Another flow is running for stack " + stack.getResourceCrn());
+        }
```
lacikaaa, line 33: please add `else`

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/ExistingStackFixService.java
```diff
@@ -0,0 +1,55 @@
...
+
+        measure(() -> doApply(stack), LOGGER, "Existing stack fixing {} took {} ms for stack: {}.", getStackFixType(), stack.getResourceCrn());
```
lacikaaa, line 35: could you add a log line before this?

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+@Service
+public class UnboundRestartFixService extends ExistingStackFixService {
+
+    static final String AFFECTED_STACK_VERSION = "7.2.11";
+
+    static final Set<String> AFFECTED_IMAGE_IDS = Set.of(
+            "26cd5a65-cd5c-457d-8d48-9caf1a486516",
+            "19cf97b8-56d8-4be9-b317-998eea99d884",
+            "c24acec3-9110-4474-9082-3620deac0910"
+    );
```
lacikaaa, line 37: should be private, and if you need the values in the tests please copy them. if I mess them up here the unit test would pass too, and we want to ensure the test would catch that case

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public StackFix.StackFixType getStackFixType() {
+        return UNBOUND_RESTART;
+    }
+
+    @Override
+    public boolean isAffected(Stack stack) {
+        if (!AFFECTED_STACK_VERSION.equals(stack.getStackVersion())) {
+            return false;
+        }
```
lacikaaa, line 59: please add `else`

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+    @Inject
+    private FlowService flowService;
+
+    @Override
+    public StackFix.StackFixType getStackFixType() {
+        return UNBOUND_RESTART;
+    }
+
+    @Override
+    public boolean isAffected(Stack stack) {
```
lacikaaa, line 56: please add logs

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        FlowIdentifier flowIdentifier = ThreadBasedUserCrnProvider.doAsInternalActor(() -> clusterOperationService.updateSalt(stack));
+        Boolean success = Polling.waitPeriodly(1, TimeUnit.MINUTES)
+                .run(() -> {
+                    FlowCheckResponse flowState = flowService.getFlowState(flowIdentifier.getPollableId());
+                    if (flowState.getHasActiveFlow()) {
+                        return AttemptResults.justContinue();
+                    }
+                    return AttemptResults.finishWith(!flowState.getLatestFlowFinalizedAndFailed());
+                });
```
lacikaaa, line 82: could you move this to a method?

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+    void doApply(Stack stack) {
+        if (!isCmServerReachable(stack)) {
+            LOGGER.info("UnboundRestartFixService cannot run, because CM server is unreachable of stack: {}", stack.getResourceCrn());
+            throw new RuntimeException("CM server is unreachable for stack: " + stack.getResourceCrn());
+        }
+        FlowIdentifier flowIdentifier = ThreadBasedUserCrnProvider.doAsInternalActor(() -> clusterOperationService.updateSalt(stack));
+        Boolean success = Polling.waitPeriodly(1, TimeUnit.MINUTES)
+                .run(() -> {
+                    FlowCheckResponse flowState = flowService.getFlowState(flowIdentifier.getPollableId());
+                    if (flowState.getHasActiveFlow()) {
```
lacikaaa, line 78: you could simplify with a tertiary operator if you don't want to add any logs

core/src/main/java/com/sequenceiq/cloudbreak/service/existingstackfix/UnboundRestartFixService.java
```diff
@@ -0,0 +1,96 @@
...
+                        return AttemptResults.justContinue();
+                    }
+                    return AttemptResults.finishWith(!flowState.getLatestFlowFinalizedAndFailed());
+                });
+        if (!success) {
+            throw new RuntimeException("Failed to update salt for stack " + stack.getResourceCrn());
```
lacikaaa, line 84: some custom exception would be nice

core/src/main/java/com/sequenceiq/cloudbreak/service/freeipa/FreeipaClientService.java
```diff
@@ -49,6 +51,10 @@ public DescribeFreeIpaResponse getByEnvironmentCrn(String environmentCrn) {
 
     public Optional<DescribeFreeIpaResponse> findByEnvironmentCrn(String environmentCrn) {
         try {
+            if (InternalCrnBuilder.isInternalCrn(ThreadBasedUserCrnProvider.getUserCrn())) {
+                String accountId = Crn.fromString(environmentCrn).getAccountId();
+                return Optional.ofNullable(freeIpaV1Endpoint.describeInternal(environmentCrn, accountId));
+            }
```
lacikaaa, line 57: `else`

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
```diff
@@ -819,6 +819,17 @@ public void removeDeadSaltMinions(GatewayConfig gateway) throws CloudbreakOrches
         }
     }
 
+    @Override
+    public Map<String, String> replacePatternInFileOnAllHosts(GatewayConfig gatewayConfig, String file, String pattern, String replace)
+            throws CloudbreakOrchestratorFailedException {
+        try (SaltConnector saltConnector = saltService.createSaltConnector(gatewayConfig)) {
+            return SaltStates.replacePatternInFile(retry, saltConnector, file, pattern, replace);
+        } catch (RuntimeException e) {
+            LOGGER.info("Error occurred during file replace execution in file '{}' while replacing pattern '{}' with '{}'", file, pattern, replace, e);
```
lacikaaa, line 828: I think it should be warn at least

orchestrator-salt/src/main/resources/salt/salt/top.sls
```diff
@@ -13,6 +13,7 @@ base:
     - logrotate
     - ntp
     - postgresql.root-certs
+    - bugfix
```
lacikaaa, line 16: I would place this under `unbound` and give it a more meaningful name

lacikaaa (changes requested):
I'm not a fan of this `fix` naming. I prefer something like `patch`.

core/src/main/java/com/sequenceiq/cloudbreak/service/cluster/GatewayConfigProvider.java
Bajzathd: I have extracted this from InstanceMetadataUpdater, but if both of them are effectively the same, I can use that.

orchestrator-salt/src/main/java/com/sequenceiq/cloudbreak/orchestrator/salt/SaltOrchestrator.java
Bajzathd: All other methods log the error with info, should I replace them all?

lacikaaa: I'm fine with this only, warn suits this better then info in my opinion

PR #11706, opened 2021-11-03
CB-14540 Base distrox OS upgrade L0 test

aszegedi:
Create new data hub (distrox) OS upgrade L0 test(s) at [Cloudbreak API test project](https://github.com/hortonworks/cloudbreak/tree/master/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion).

Test duration must be take into consideration:
- base (1:master, 1:compute, 3:workers) test takes more than 3 hours
- 50 workers test takes around 4 hours
- 200 workers test takes around 7 hours

So I was implemented the base case for the test project.

**NOTE:**
- We need to wait the [OPSAPS-62124](https://jira.cloudera.com/browse/OPSAPS-62124) to be resolved
- In the future if we would like to run this automated test on a daily basis, we need to create a separate `Longrunning` Jenkins job for this. This can be part of a follow up JIRA.

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/DistroXOSUpgradeTests.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/DistroXOSUpgradeTests.java
@@ -1,0 +1,139 @@
+package com.sequenceiq.it.cloudbreak.testcase.e2e.l0promotion;
+
+import static com.sequenceiq.it.cloudbreak.context.RunningParameter.key;
+
+import javax.inject.Inject;
+
+import org.testng.annotations.Ignore;
+import org.testng.annotations.Test;
+
+import com.sequenceiq.distrox.api.v1.distrox.model.upgrade.DistroXUpgradeReplaceVms;
+import com.sequenceiq.freeipa.api.v1.freeipa.stack.model.common.Status;
+import com.sequenceiq.it.cloudbreak.client.DistroXTestClient;
+import com.sequenceiq.it.cloudbreak.client.FreeIpaTestClient;
+import com.sequenceiq.it.cloudbreak.client.SdxTestClient;
+import com.sequenceiq.it.cloudbreak.cloud.v4.CommonClusterManagerProperties;
+import com.sequenceiq.it.cloudbreak.context.Description;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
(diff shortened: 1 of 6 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/DistroXOSUpgradeTests.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    protected void setupTest(TestContext testContext) {
+        testContext.getCloudProvider().getCloudFunctionality().cloudStorageInitialize();
+        createDefaultUser(testContext);
+        initializeDefaultBlueprints(testContext);
+        createDefaultCredential(testContext);
+        createEnvironmentWithFreeIpa(testContext);
+    }
+
+    @Test(dataProvider = TEST_CONTEXT, description = "We need to wait the OPSAPS-62124 to be resolved")
```
lnardai, line 54: Why  did we add @UseSpotInstances?

if this is in the l0promotion won't this be automatically triggered by the L0 tests? or do we have an explicit list there? For this reason I would add a disabled test notation just to be safe that this won't be included accidentally.

lnardai (approved):
Left one improvement, but after that we can merge it.

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/testcase/e2e/l0promotion/DistroXOSUpgradeTests.java
aszegedi: Right now promotion tests are defined suite-by-suite with own YAML file:
- [cloudbreak-ansible-playbooks/jenkins/jenkins_jobs/mow-api-e2e-aws.xml](https://github.com/hortonworks/cloudbreak-ansible-playbooks/blob/master/jenkins/jenkins_jobs/mow-api-e2e-aws.xml#L583)
- [cloudbreak-ansible-playbooks/jenkins/jenkins_jobs/mow-api-e2e-azure.xml](https://github.com/hortonworks/cloudbreak-ansible-playbooks/blob/master/jenkins/jenkins_jobs/mow-api-e2e-azure.xml#L521)

However we can rest assured by the mentioned [@Ignore TestNG annotation](https://testng.org/doc/documentation-main.html#ignore-tests). So I've introduced this.

PR #11708, opened 2021-11-03
CB-14752 Validate Node Manager role count before downscale

sodre90:

### DIFF
--- a/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
+++ b/template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
@@ -6,6 +6,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
@@ -49,10 +50,12 @@ public class CmTemplateValidator implements BlueprintValidator {
     }
 
     @Override
-    public void validateHostGroupScalingRequest(String accountId, Blueprint blueprint, HostGroup hostGroup, Integer adjustment) {
+    public void validateHostGroupScalingRequest(String accountId, Blueprint blueprint, HostGroup hostGroup, Integer adjustment,
+            Collection<InstanceGroup> instanceGroups) {
         CmTemplateProcessor templateProcessor = processorFactory.get(blueprint.getBlueprintText());
         Versioned blueprintVersion = () -> templateProcessor.getVersion().get();
         Set<String> services = templateProcessor.getComponentsByHostGroup().get(hostGroup.getName());
+        validateNodeManagerAdjustment(hostGroup, adjustment, instanceGroups, templateProcessor);
         for (BlackListedDownScaleRole role : BlackListedDownScaleRole.values()) {
             if (services.contains(role.name()) && adjustment < 0) {
                 validateRole(accountId, role, blueprintVersion, templateProcessor);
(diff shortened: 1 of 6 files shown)


template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
```diff
@@ -65,6 +68,21 @@ public void validateHostGroupScalingRequest(String accountId, Blueprint blueprin
         }
     }
 
+    private void validateNodeManagerAdjustment(HostGroup hostGroup, Integer adjustment, Collection<InstanceGroup> instanceGroups,
```
foldik, line 71: We have this field `UpdateClusterV4Request.hostGroupAdjustment.forced`. Does the code checks if it is a force request then proceed without this validation?

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
```diff
@@ -65,6 +68,21 @@ public void validateHostGroupScalingRequest(String accountId, Blueprint blueprin
...
+            CmTemplateProcessor templateProcessor) {
+        Map<String, Set<String>> nodeManagerComponentInGroups = templateProcessor.getComponentsByHostGroup().entrySet().stream()
+                .filter(entry -> entry.getValue().contains("NODEMANAGER"))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        if (nodeManagerComponentInGroups.containsKey(hostGroup.getName())) {
+            instanceGroups.stream().filter(instanceGroup -> nodeManagerComponentInGroups.containsKey(instanceGroup.getGroupName()))
+                    .map(InstanceGroup::getNodeCount).reduce(Integer::sum).ifPresent(nodeManagerCountInGroups -> {
+                if (nodeManagerCountInGroups + adjustment <= 0) {
+                    throw new BadRequestException("At lease one instance required with NODEMANAGER role.");
```
biharitomi, line 80: Please fix the typo in the second word! :)
Maybe one sentence could also be added to be more customer-friendly like `Your request would eliminate all the instances with NODEMANAGER role which is not supported. At least.....`

biharitomi (changes requested):
I have one comment, otherwise LGTM

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
foldik, line 80: @biharitomi as I see this code works based on the blueprint and not based on the actual state in CM. Is it acceptable? If yes then I suggest to change the error message to
```
Scaling adjustment is not allowed, based on the template it would delete all nodes with required Node Manager role.
```

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidatorTest.java
```diff
@@ -3,12 +3,14 @@
...
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Set;
 
 import org.junit.Test;
```
foldik, line 16: Can you please use `org.junit.jupiter.api.Test` instead?

foldik: Also can you please use `@ExtendWith(MockitoExtension.class)` instead of `@RunWith(MockitoJUnitRunner.class)`.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
foldik: @biharitomi I add another option below please check it.

foldik (changes requested):
A few minor changes requested.

template-manager-cmtemplate/src/test/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidatorTest.java
sodre90: It is an existing test class. I would not like to change it, instead, we should create a ticket to remove all junit4 tests and rewrite them to junit5.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
biharitomi: Fine for me.

template-manager-cmtemplate/src/main/java/com/sequenceiq/cloudbreak/cmtemplate/CmTemplateValidator.java
sodre90: It depends, in case of thrash icon the code goes this path, so the validation will be skipped if force is enabled: https://github.com/hortonworks/cloudbreak/blob/7dd704fbee7e72d7141e33e021c4b4f34ce85f8f/core/src/main/java/com/sequenceiq/cloudbreak/service/stack/flow/StackOperationService.java#L116-L120

In case of normal downscale, we don't care about force flag, we will validate:
https://github.com/hortonworks/cloudbreak/blob/3e4b5600c66d548bd472cf9770a9ef90bf58a824/core/src/main/java/com/sequenceiq/cloudbreak/service/ClusterCommonService.java#L137-L160

As we discussed with @biharitomi it looks fine, because every other validation will run in a normal downscale scenario (regardless force flag is enabled or not).

biharitomi (approved):

PR #11710, opened 2021-11-04
CB-14645 Adds a real ums test to validate change image catalog permis…

pkedvessy:
…sions.


foldik (approved):
Very good!

PR #11711, opened 2021-11-04
CB-14758 Added tests:

topolyai5:
* native upgrade
* native repair
Some minor fixes the AWS_NATIVE resource builders.
Environment name check moved into the create action instead of the TestDto's valid method in the IT. This was neccessary because we cannot prepare properly the distrox object. In this case I wnat to use an existing env with name, but I cannot set because the env was required in prepare time instad of the execution

### DIFF
--- a/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/v1/distrox/FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction.java
+++ b/integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/v1/distrox/FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction.java
@@ -1,0 +1,31 @@
+package com.sequenceiq.it.cloudbreak.action.v1.distrox;
+
+import java.util.ArrayList;
+
+import com.sequenceiq.distrox.api.v1.distrox.model.network.InstanceGroupNetworkV1Request;
+import com.sequenceiq.distrox.api.v1.distrox.model.network.aws.InstanceGroupAwsNetworkV1Parameters;
+import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
+import com.sequenceiq.it.cloudbreak.CloudbreakClient;
+import com.sequenceiq.it.cloudbreak.EnvironmentClient;
+import com.sequenceiq.it.cloudbreak.action.Action;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+
+public class FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction implements Action<DistroXTestDto, CloudbreakClient> {
+
+    @Override
+    public DistroXTestDto action(TestContext testContext, DistroXTestDto testDto, CloudbreakClient client) throws Exception {
+        if ("AWS_NATIVE".equals(testDto.getRequest().getVariant())) {
+            EnvironmentClient envClient = testContext.getMicroserviceClient(EnvironmentClient.class);
+            DetailedEnvironmentResponse envResponse = envClient.getDefaultClient()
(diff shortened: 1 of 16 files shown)


integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/v1/distrox/FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction.java
```diff
@@ -0,0 +1,31 @@
...
+import com.sequenceiq.distrox.api.v1.distrox.model.network.InstanceGroupNetworkV1Request;
+import com.sequenceiq.distrox.api.v1.distrox.model.network.aws.InstanceGroupAwsNetworkV1Parameters;
+import com.sequenceiq.environment.api.v1.environment.model.response.DetailedEnvironmentResponse;
+import com.sequenceiq.it.cloudbreak.CloudbreakClient;
+import com.sequenceiq.it.cloudbreak.EnvironmentClient;
+import com.sequenceiq.it.cloudbreak.action.Action;
+import com.sequenceiq.it.cloudbreak.context.TestContext;
+import com.sequenceiq.it.cloudbreak.dto.distrox.DistroXTestDto;
+
+public class FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction implements Action<DistroXTestDto, CloudbreakClient> {
```
afarsang, line 14: this seems as a util for the test, not an actual test action, right?
shouldn't be this a testDto 'withEnironmentSubnets' method rather?

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/distrox/DistroXTestDtoBase.java
```diff
@@ -43,19 +43,17 @@ protected DistroXTestDtoBase(DistroXV1Request request, TestContext testContext)
...
     }
 
     public DistroXTestDtoBase<T> withEnvironmentKey(String environmentKey) {
         EnvironmentTestDto env = getTestContext().get(environmentKey);
-        if (env == null) {
-            throw new IllegalArgumentException("Env is null with given key: " + environmentKey);
+        DistroXTestDtoBase<T> ret = this;
+        if (env != null && env.getResponse() != null) {
+            ret = withEnvironmentName(env.getResponse().getName());
         }
```
afarsang, line 55: I am missing the else branch where a illegalArgumentExcepton is thrown. What is the idea here?

afarsang (changes requested):

afarsang:
is something missing from the commit message? "Added test:" what I see

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/dto/distrox/DistroXTestDtoBase.java
topolyai5: Not missing, just fetch the env name if it exists. Or we set later. After this the env can be null, I have moved the validation into the post action

topolyai5:
> is something missing from the commit message? "Added test:" what I see

the other part in the description

biharitomi (approved):

integration-test/src/main/java/com/sequenceiq/it/cloudbreak/action/v1/distrox/FetchPreferredSubnetForInstanceNetworkIfMultiAzEnabledAction.java
topolyai5: I think it is a mixed one. We fetch from the server via API and set the value. I prefer when we set a direct value via the testDto like your suggestion. We should just set the value because if we set the value (no network, no valid data) we don't need to break the run. Probably, it is not causing any problem here, but if someone uses the `.valid()` we can't override the value later. We should only break the test run via Acton or Assertion.

PR #11713, opened 2021-11-04
CB-14751 Force delete and downscale should work when CM decommission fails

foldik:


biharitomi (approved):
LGTM, Shouldn't it go to 2.49 at least?

schfeca75 (approved):

